PR #897, opened 2020-10-17
Added example on how to execute RPC calls in javascript

theotow:
see changeset

### DIFF
--- a/docs/buidler-evm/README.md
+++ b/docs/buidler-evm/README.md
@@ -251,6 +251,23 @@ To customise it, take a look at [the configuration section](/config/#buidler-evm
 - `eth_submitHashrate`
 - `eth_submitWork`
 
+### Example executing a method in Javascript
+
+Without params
+
+```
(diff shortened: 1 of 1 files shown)


alcuadrado:
Hey @theotow,

Thanks for this PR. Unfortunately, not everyone using Buidler uses ethers, so it would be confusing to use ethers in that section of the docs.

theotow:
@alcuadrado what would be a good place to place this in your opinion? I just started to use buidler and dont have a full grasp of the project jet. Mainly choose it since the unit-test are speedy and i get rapid feedback, i can imagine with more examples it will be much easier to onboard to buidler

alcuadrado:
We are in the process of rebranding Buidler, and that includes lots of changes to the docs, so it's hard to answer that right now.

I agree with you that more examples are needed. Thanks for the feedback.

You should join [our discord](https://invite.gg/BuidlerSupport), which complements the docs pretty well.

PR #900, opened 2020-10-19
Prepare RC versions

alcuadrado:
Just a upgrades 2 dev dependencies and bumps every version.

### DIFF
--- a/packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
+++ b/packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
@@ -70,8 +70,12 @@ export function useProvider(
   });
 
   afterEach("Remove provider", async function () {
-    delete this.provider;
-    delete this.hardhatNetworkProvider;
+    // These two deletes are unsafe, but we use this properties
+    // in very locally and are ok with the risk.
+    // To make this safe the properties should be optional, which
+    // would be really uncomfortable for testing.
+    delete (this as any).provider;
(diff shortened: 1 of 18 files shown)


packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
```diff
@@ -70,8 +70,12 @@ export function useProvider(
...
 
   afterEach("Remove provider", async function () {
-    delete this.provider;
-    delete this.hardhatNetworkProvider;
+    // These two deletes are unsafe, but we use this properties
+    // in very locally and are ok with the risk.
+    // To make this safe the properties should be optional, which
+    // would be really uncomfortable for testing.
+    delete (this as any).provider;
+    delete (this as any).hardhatNetworkProvider;
```
fvictorio, line 78: `delete {}.foo` doesn't throw, so it's not a big deal anyway, isn't it?

fvictorio (approved):

packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
alcuadrado: Yeah, I think it's fine. But it makes sense from TS to forbid it unless the prop is optional.

alcuadrado: The unsafe part is the rest of the code, not the `delete`. Places that expect that property to exist because of its type.

PR #906, opened 2020-10-19
Internal Improvement: sync `package-lock`s

CruzMolina:
- Ensures `package-lock` files updated across monorepo


nicholasjpaterno (approved):
LGTM!
:shipit:

fvictorio:
Hi @CruzMolina, thank you for this! Sadly, we are moving to yarn workspaces, which has a single `yarn.lock` at the root, so this change won't be necessary (and it will be impossible to merge :sweat_smile:) I'm going to close the PR for bookkeeping, thanks again.

PR #910, opened 2020-10-20
Update ethers & waffle guide

alcuadrado:


docs/guides/waffle-testing.md
```diff
@@ -31,7 +35,7 @@ Welcome to Hardhat v2.0.0
   Quit
 ```
 
-Select `Create a sample project`. This will create some files but also install the necessary packages.
+Select `Create a sample project`. This will create some files, install the `hardhat-ethers` and `hardhat-waffle` plugins, and other necessary packages.
```
alcuadrado, line 38:
```suggestion
Select `Create a sample project`. This will create some files and install the `@nomiclabs/hardhat-ethers`, `@nomiclabs/hardhat-waffle` plugins, and other necessary packages.
```

alcuadrado: I think it should be like this, @fvictorio. Shouldn't it?

alcuadrado:
@fzeoli also ready

fzeoli:
lgtm

PR #918, opened 2020-10-21
Adapt tutorial

fvictorio:


alcuadrado:
I removed the intructions to setup node-gyp.

We need to update the boilerplate repo, and there's an outdated screenshot of it here.

PR #920, opened 2020-10-21
Update compilation guide to Hardhat

fzeoli:


docs/guides/compile-contracts.md
```diff
@@ -9,19 +9,91 @@ Compiled 1 contract successfully
...
-After the initial compilation, if the contract code hasn't changed then Hardhat will skip compilation when running the `compile` task:
+After the initial compilation, Hardhat will try to do the least amount of work possible the next time you compile. For example, if you didn't change any file since the last compilation, nothing will be compiled. If you only modified one file, only that file and others affected by it will be recompiled.
+
 ```
 $ npx hardhat compile
-All contracts have already been compiled, skipping compilation.
+Nothing to compile
 ```
 
 To force a compilation you can use the `--force` argument, or run `npx hardhat clean` to clear the caches and delete the artifacts.
```
fvictorio, line 19:
```suggestion
To force a compilation you can use the `--force` argument, or run `npx hardhat clean` to clear the cache and delete the artifacts.
```

docs/guides/compile-contracts.md
```diff
@@ -9,19 +9,91 @@ Compiled 1 contract successfully
...
 
+## Configuring the compiler
+
+If you need to customize the `solc` compiler options, then you can do so through the `solidity` config field in your `hardhat.config.js`. The simplest way to use this field is the shorthand for setting the compiler version, which we recommend always doing:
```
fvictorio, line 23: > which we recommend always doing

It's just me or this can be misleading? It sounds like we recommend to always configure the compiler this way, instead of "we recommend always setting some version of the compiler".

docs/guides/compile-contracts.md
```diff
@@ -9,19 +9,91 @@ Compiled 1 contract successfully
...
+
+### Multiple Solidity versions
+
+Hardhat supports projects that use different, incompatible versions of solc. For example, if you have a project where some files use Solidity 0.5 and others use 0.6, you can configure Hardhat to use compiler versions compatible with those files like this:
+
+```js
+module.exports = {
+  solidity: {
+    compilers: [
+      {version: "0.5.5"},
```
fvictorio, line 63:
```suggestion
      {
        version: "0.5.5"
      },
```

docs/guides/compile-contracts.md
alcuadrado, line 23:
```suggestion
If you need to customize the Solidity compiler options, then you can do so through the `solidity` config field in your `hardhat.config.js`. The simplest way to use this field is the shorthand for setting the compiler version, which we recommend always doing:
```

Using `solc` will be confusing, as that term is not present in the config.

docs/guides/compile-contracts.md
```diff
@@ -35,57 +107,52 @@ Each artifact consists of a json with the following properties:
...
+### Reading artifacts
 
-## Solidity 5 and Solidity 6 contracts in the same project
+The [HRE] has an `artifacts` object with helper methods. For example, you can get a list with the paths to all artifacts by calling `hre.artifacts.getArtifactPaths()`.
 
-Hardhat can handle scenarios where you need to compile Solidity 5 and Solidity 6 contracts in the same project. An example of this is when there are deployed contracts that have been written in Solidity 5 but your new contracts that depend on the old ones are written in Solidity 6, or different contract dependencies use different Solidity versions and you need both of these to play along to run your tests.
+You can also read an artifact using the name of the contract by calling `hre.artifacts.readArtifact("Bar")` and that will give us the content of the artifact for the `Bar` contract in `Bar.sol`. This works because there is only one contract `Bar` in the whole project, but calling `hre.artifacts.readArtifact("Foo")`, would yield an error because there are two `Foo` contracts. To disambiguate, you have to use the **Fully Qualified Name** of the contract: `hre.artifacts.readArtifact("contracts/Foo.sol:Foo")`.
```
fvictorio, line 122: This doesn't make a lot of sense here because the contracts files structure is below, so the reader doesn't know where those contracts came from.

docs/guides/compile-contracts.md
alcuadrado, line 63:
```suggestion
      { version: "0.5.5" },
```

docs/guides/compile-contracts.md
alcuadrado, line 122:
```suggestion
You can also read an artifact using the name of the contract by calling `hre.artifacts.readArtifact("Bar")` and that will give us the content of the artifact for the `Bar` contract in `Bar.sol`. This works because there is only one contract `Bar` in the whole project, but calling `hre.artifacts.readArtifact("Foo")`, would throw an error because there are two `Foo` contracts. To disambiguate, you have to use the **Fully Qualified Name** of the contract: `hre.artifacts.readArtifact("contracts/Foo.sol:Foo")`.
```

alcuadrado, line 122: This is explained based on a few contracts that haven't been mentioned before.

alcuadrado: Oh, I guess they are the ones from the following section, but it's confusing.

docs/guides/compile-contracts.md
```diff
@@ -35,57 +116,52 @@ Each artifact consists of a json with the following properties:
...
+### Reading artifacts
 
-## Solidity 5 and Solidity 6 contracts in the same project
+The [HRE] has an `artifacts` object with helper methods. For example, you can get a list with the paths to all artifacts by calling `hre.artifacts.getArtifactPaths()`.
 
-Hardhat can handle scenarios where you need to compile Solidity 5 and Solidity 6 contracts in the same project. An example of this is when there are deployed contracts that have been written in Solidity 5 but your new contracts that depend on the old ones are written in Solidity 6, or different contract dependencies use different Solidity versions and you need both of these to play along to run your tests.
+You can also read an artifact using the name of the contract by calling `hre.artifacts.readArtifact("Bar")` and that will give us the content of the artifact for the `Bar` contract. This would only work if there was just one contract `Bar` in the whole project, but calling `hre.artifacts.readArtifact("Foo")`, would yield an error if there were two `Foo` contracts. To disambiguate this case, you would have to use the **Fully Qualified Name** of the contract: `hre.artifacts.readArtifact("contracts/Foo.sol:Foo")`.
```
alcuadrado, line 131:
```suggestion
You can also read an artifact using the name of the contract by calling `hre.artifacts.readArtifact("Bar")` and that will give us the content of the artifact for the `Bar` contract. This would only work if there was just one contract `Bar` in the whole project, but calling `hre.artifacts.readArtifact("Foo")`, would throw an error if there were two `Foo` contracts. To disambiguate this case, you would have to use the **Fully Qualified Name** of the contract: `hre.artifacts.readArtifact("contracts/Foo.sol:Foo")`.
```

alcuadrado (approved):

docs/guides/compile-contracts.md
```diff
@@ -9,19 +9,100 @@ Compiled 1 contract successfully
 
 The compiled artifacts will be saved in the `artifacts/` directory by default, or whatever your configured artifacts path is. Look at the [paths configuration section](../config) to learn how to change it. This directory will be created if it doesn't exist.
```
scnale, line 10:
```suggestion
The compiled artifacts will be saved in the `artifacts/` directory by default, or whatever your configured artifacts path is. Look at the [paths configuration section](../config/README.md#path-configuration) to learn how to change it. This directory will be created if it doesn't exist.
```

docs/guides/compile-contracts.md
```diff
@@ -9,19 +9,100 @@ Compiled 1 contract successfully
...
+    }
+  }
+}
+```
+
+`settings` has the same schema as the `settings` entry in the [Input JSON](https://solidity.readthedocs.io/en/v0.7.2/using-the-compiler.html#input-description) that can be passed to the compiler. Some commonly used settings are:
+
+- `optimizer`: an object with `enabled` and `runs` keys. Default value: `{ enabled: false, runs: 200 }`.
+
+- `evmVersion`: a string controlling the target evm version. One of `homestead`, `tangerineWhistle`, `spuriousDragon`, `byzantium`, `constantinople`, `petersburg`, and `instanbul`. Default value: managed by `solc`.
```
scnale, line 53:
```suggestion
- `evmVersion`: a string controlling the target evm version. One of `homestead`, `tangerineWhistle`, `spuriousDragon`, `byzantium`, `constantinople`, `petersburg`, `istanbul`, and `berlin`. Default value: managed by `solc`. 
```

PR #923, opened 2020-10-21
Fix the hardhat_setLoggingEnabled RPC method

alcuadrado:
@fvictorio

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
@@ -272,7 +272,7 @@ export class HardhatNetworkProvider extends EventEmitter
       node,
       this._throwOnTransactionFailures,
       this._throwOnCallFailures,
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -284,6 +284,7 @@ export class HardhatNetworkProvider extends EventEmitter
       this._reset.bind(this),
       (loggingEnabled: boolean) => {
         this._loggingEnabled = loggingEnabled;
```
fvictorio, line 286: Feels like this variable can be removed. But I'm fine with doing it later.

fvictorio (approved):

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
alcuadrado: Probably right. But let's keep this small for now.

PR #926, opened 2020-10-22
Make buidler-evm's tracing compatible with mapping function arguments.

adhusson:
`typeDescriptions.typeString` should be correct for mapping types as well!


alcuadrado:
Thanks for this PR, @adhusson! We'll check it out soon.

Do you think you can add a test for this?

fvictorio:
Hey @adhusson, I implemented your changes with tests in #942, but for hardhat. But if it's not possible for you to [migrate](https://hardhat.org/guides/migrate-from-buidler.html) soon, I think we can backport it in buidler too.

PR #943, opened 2020-10-22
fix: do not prompt for consent in ci environments

fubhy:
There is already a check that disables telemetry inside of `analytics.ts`, however, the prompt is still shown. This is problematic in CI environments especially because there is currently no way to skip this check (e.g. with `--yes` // `--no`). As a result, CI currently gets stuck infinitely.

### DIFF
--- a/packages/hardhat-core/src/internal/cli/cli.ts
+++ b/packages/hardhat-core/src/internal/cli/cli.ts
@@ -18,6 +18,7 @@ import { isCwdInsideProject } from "../core/project-structure";
 import { Environment } from "../core/runtime-environment";
 import { loadTsNode, willRunWithTypescript } from "../core/typescript-support";
 import { Reporter } from "../sentry/reporter";
+import { isRunningOnCiServer } from "../util/ci-detection";
 import {
   hasConsentedTelemetry,
   writeTelemetryConsent,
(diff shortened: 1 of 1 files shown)


alcuadrado:
Thanks for submitting this PR, it looks like the right change. Unfortunately, something seems to have broken on our CI, and I'd rather fix it and publish tomorrow morning.

I hope the delay doesn't cause you much trouble.

fubhy:
> Thanks for submitting this PR, it looks like the right change. Unfortunately, something seems to have broken on our CI, and I'd rather fix it and publish tomorrow morning.
> 
> I hope the delay doesn't cause you much trouble.

No problem. The tests pass locally so maybe it's just a random failure. My CI is happy for now... Did this in case anyone runs into the same issue in the meantime:

```
        # TODO: Remove this again once it's fixed in hardhat
      - name: Skip telemetry consent
        run: |
          mkdir -p ~/.config/hardhat-nodejs
          echo { \"consent\": false } >> ~/.config/hardhat-nodejs/telemetry-consent.json
```

alcuadrado:
I think the issue here is that github doesn't share out GH Action secrets with you. I'll merge this and wait for the CI to run in `master` before publishing it.

PR #948, opened 2020-10-23
feat: add support for EIP-1898 block tags

fubhy:
https://eips.ethereum.org/EIPS/eip-1898

It's currently not possible to specify block hashes / numbers using the EIP-1898 spec.

This spec is implemented by e.g. thegraph's graph-node which is a pretty core architectural piece of ethereum development these days! With this PR, graph-node works properly and hardhat can be used to develop subgraphs.


fubhy:
Whoops, accidentally included some code from debugging. Solved now.

fubhy:
Are these test failures the same that we saw in my other PR (due to lack of access to the configured secrets) or are they legit?

fvictorio:
Probably the secrets thing, I'm going to merge master to see if it fixes it.

fvictorio:
They seem like errors from this PR, at least I get them locally too.

tynes:
I'd definitely like to see this functionality, would be very useful for what I'm working on right now. I added this functionality to ganache but we'd like to move to hardhat

fubhy:
@fvictorio Do you generally agree with this PR though? If yes, I am happy to fix the tests and add test coverage for the newly added feature.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -590,9 +592,18 @@ export class EthModule {
       filter.toBlock = new BN(block.header.number);
     }
 
+    const [fromBlock, toBlock] = await Promise.all([
```
fvictorio, line 595: I don't understand this change. It isn't related to the rest of the changes, is it? (That's ok by me if it's an improvement; just trying to understand it).

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
```diff
@@ -28,6 +28,25 @@ const isRpcDataString = (u: unknown) =>
...
   typeof u === "string" && u.length === 66 && isRpcDataString(u);
 
+interface BlockTagObject {
+  blockHash?: string | Buffer;
+  blockNumber?: string | number | BN;
+}
+
+const isBlockTagObject = (u: any): u is BlockTagObject => {
+  if (typeof u === "object") {
+    if (u?.blockHash != null && u?.blockNumber == null) {
```
fvictorio, line 38: @alcuadrado should we have a tslint rule that checks that `===` and `!==` are used?

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
```diff
@@ -77,8 +96,28 @@ export const logTopics = t.union([
...
+  (u): u is BN | Buffer => BN.isBN(u) || Buffer.isBuffer(u),
+  (u, c) => {
+    if (isBlockTagObject(u)) {
+      if (u.blockHash != null) {
+        return t.success(toBuffer(u.blockHash));
+      }
+
+      if (u.blockNumber != null) {
+        return t.success(new BN(toBuffer(u.blockNumber)));
+      }
```
fvictorio, line 110: Does the spec say what should happen if both `blockHash` and `blockNumber` are specified?

fvictorio: Oh, I see that `isBlockTagObject` kind of checks that. I don't think we should do that though: an object with both values specified *is* a `BlockTagObject` (according to the interface). Throwing an error because that's invalid should be done elsewhere, maybe here. The result will be the same, it's just that the code will be a little clearer.

fvictorio:
Hey @fubhy, sorry for taking so long to review this! I left some initial comments.

Besides that, I guess my main concern is that this EIP isn't finished, is it? I know they can take a long time to be finished, but it's also a pain to have to update the code every time the spec changes. @alcuadrado what do you think?

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
fubhy: @fvictorio this is `==` on purpose to check both `undefined` and `null`, because `null === undefined // false` but `null == undefined // true`

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
fubhy: It's related. This used to be synchronous so it was fine to call the function serially. However, now that it's asynchronous I refactored it so it's called asynchronously in parallel for performance reasons.

fubhy:
> Hey @fubhy, sorry for taking so long to review this! I left some initial comments.
> 
> Besides that, I guess my main concern is that this EIP isn't finished, is it? I know they can take a long time to be finished, but it's also a pain to have to update the code every time the spec changes. @alcuadrado what do you think?

True, it's not yet finished but very likely to be accepted in the current shape or at least very close to it. It's also already widely adopted (infura, alchemy, geth, parity, ganache, etc.). Also, since https://thegraph.com/ uses it already in their graph-node and graphprotocol is already a core part of most solidity development that I am aware of, usage of Hardhat is currently impossible for subgraph development.

Hence, I think it's worth to adopt it even if it's still in Draft state.

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
fubhy: Right. I am validating that currently by disallowing `blockHash` and `blockNumber` to be defined together. Only one is allowed as you observed. I didn't see an example of both `blockHash` and `blockNumber` in the same request payload in the EIP. Hence, my assumption is that it's not valid to set both in the same request payload.

I don't have a strong opinion on where the error should be thrown.

fvictorio:
> True, it's not yet finished but very likely to be accepted in the current shape or at least very close to it. It's also already widely adopted (infura, alchemy, geth, parity, ganache, etc.).

Oh, that should be more than enough. In that case, I think you can start adding tests. They aren't going to be tied to the implementation, so the details about how it's done shouldn't matter a lot.

Another question I have is: should we support the `requireCanonical` flag? If some client uses it, this will reject the request, so maybe we should at least accept it (and ignore it for now, since all blocks are going to be canonical). What do you think?

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
fvictorio: Oh, we don't usually do that. If both things can happen, we explicitly check both.

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
fvictorio: Yes, I agree that an error should be thrown. It's just a minor comment about who is responsible for doing that: I think `isBlockTagObject` should return `true` if both are defined and then the decoder should fail if they are.

But this is _very_ minor and a personal preference, feel free to ignore it :sweat_smile:

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
fubhy: For nullish value assertions I'd argue that `==` is the preferred and generally accepted operator in JavaScript / TypeScript and also what powers the nullish coalescing (`.?`) operator (undefined & null are treated equally here too).

For any other scalar comparisons I'd agree.

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
fubhy: Gotcha. I mean you are right... In terms of "type" compliance with this current interface it would be fair to simply let it slide here because, yes, it's compliant with the TS interface if both are defined too. Maybe there is a third option though! We could create two different types (one for blockHash & one for blockNumber) and type them out explicitly. Would you prefer that or keep it simple?

fubhy:
> > True, it's not yet finished but very likely to be accepted in the current shape or at least very close to it. It's also already widely adopted (infura, alchemy, geth, parity, ganache, etc.).
> 
> Oh, that should be more than enough. In that case, I think you can start adding tests. They aren't going to be tied to the implementation, so the details about how it's done shouldn't matter a lot.
> 
> Another question I have is: should we support the `requireCanonical` flag? If some client uses it, this will reject the request, so maybe we should at least accept it (and ignore it for now, since all blocks are going to be canonical). What do you think?

Yes, that makes sense (adding `requireCanonical`). Will do!

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
fvictorio: That makes sense, but I think internal consistency is more important in this case. If you make a search, you'll see that there are only a couple of cases where `==` and `!=` are used, and those were probably something that went unnoticed in a code review. I'm surprised we don't have a linter rule for this :sweat_smile:

packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
fvictorio: I think just handling it in the decoder is the balanced approach, the intention is clearer that way.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
fvictorio: Oh, I see :+1:

fubhy:
@fvictorio I added a minimal test for eth_call and fixed the failing tests by re-introducing _extractBlock (apparently it's meant to be a shortcut for eth_getLogs to allow for out-of-bounds block numbers in this case).

fvictorio (approved):
I added some tests and fixed one issue. I think it can be merged now, but I would like for @alcuadrado to take a look at it first.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -1041,7 +1048,13 @@ export class EthModule {
...
+    let block: Block | undefined;
+    if (BN.isBN(blockTag)) {
+      block = await this._node.getBlockByNumber(blockTag);
+    } else if (Buffer.isBuffer(blockTag)) {
+      block = await this._node.getBlockByHash(blockTag);
+    }
+
     if (block === undefined) {
       const latestBlock = await this._node.getLatestBlockNumber();
       throw new InvalidInputError(
```
alcuadrado, line 1060: This error message is now outdated, isn't it?

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -1052,18 +1065,26 @@ export class EthModule {
...
 
-  private _extractBlock(blockTag: OptionalBlockTag): BN {
+  private async _extractBlock(blockTag: OptionalBlockTag): Promise<BN> {
+    if (BN.isBN(blockTag)) {
+      return blockTag;
+    }
+
+    if (Buffer.isBuffer(blockTag)) {
+      const block = await this._node.getBlockByHash(blockTag);
+      return block === undefined ? LATEST_BLOCK : new BN(block.header.number);
```
alcuadrado, line 1075: Won't this mean that sending a non-existent block hash will be interpreted as `latest`?

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
fubhy: Hmm, why would it be outdated? `block` can still be `undefined` if the given blockTag is invalid for whatever reason.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
fubhy: True. This could be changed to throw an exception instead. This function is currently only used for logs though. I don't h ave a strong opinion about this.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
alcuadrado: It assumes that the block tag is a number. The wording, I mean.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
alcuadrado: I think the eip says that a resource not found error should be returned. It's a bit ambiguous on this front though.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
fubhy: Ah okay. Yeah I can update that.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
fubhy: Yeah It's not the most detailed EIP I have to say. It's also quite ambiguous in terms of input validation in general. I think this implementation should be fine though. I will update the PR to throw an error.

alcuadrado:
Hey @fubhy, I fixed those things myself so we can publish this today. Thanks for your work on this PR!

PR #951, opened 2020-10-23
Export extended ethers Signer

fvictorio:
Using the extended signer in a test is cumbersome because you need to do this:

```typescript
import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/dist/src/signer-with-address";

describe(() => {
  let signers: SignerWithAddress[]

  beforeEach(async () => {
    signers = await ethers.getSigners()
```

to be able to type-check something like `signers[0].address` .

With this PR, the import can be replaced with:

```typescript
import { Signer } from "@nomiclabs/hardhat-ethers";
```

### DIFF
--- a/packages/hardhat-ethers/src/index.ts
+++ b/packages/hardhat-ethers/src/index.ts
@@ -26,3 +26,5 @@ extendEnvironment((hre) => {
     };
   });
 });
+
+export { ExtendedEthersSigner as Signer } from "./types";
--- a/packages/hardhat-ethers/src/signer-with-address.ts
+++ b/packages/hardhat-ethers/src/signer-with-address.ts
@@ -1,6 +1,9 @@
 import { ethers } from "ethers";
 
-export class SignerWithAddress extends ethers.Signer {
+import { ExtendedEthersSigner } from "./types";
+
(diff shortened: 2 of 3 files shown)


alcuadrado:
Oh, this is really interesting. I just wouldn't name it `Signer`, cause that can lead to very uncomfortable patterns because of names clashing. Maybe keep `SignerWithAddress`? Or `HardhatSigner`?

alcuadrado:
> ```ts
> ```ts
> import { Signer } from "@nomiclabs/hardhat-ethers";
> ```
> 
> 
>       
> ```

Now that I think of, this can be dangerous, as you can accidentally import the plugin before hardhat is initialized.

fvictorio:
Two ideas:

- Recommend users to do `import type` for this
- Bite the bullet and try to change the way things are built so that users can do `import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/types'`

fvictorio:
- Use two tsconfig, one for src and one for test. Output of src should be `.`, like in core.
- Keep a `src/dist/etc` so that this is not a breaking change for people importing the current type.

alcuadrado:
This was fixed in another PR

PR #953, opened 2020-10-24
Buildinfo id exposed

wighawag:
This expose the id of the buildInfo so plugin can identify same buildinfo and benefit from the space optimization hardhat do by sharing same buildinfo across file


alcuadrado:
@wighawag have you thought about what would happen when reading a build info that doesn't have the id? I think this change can be a bit dangerous. You are probably the only one consuming this though, so the risk is not too big to stop it.

wighawag:
I intend to use this for caching purpose.
for undefined id, I can simply do not use cache for these.

But as users move to higher version of hardhat, this won't be a problem. I could even add a warning to let user know to upgrade.

PR #954, opened 2020-10-26
feat: allow to disable collection of telemetry data through config

fubhy:
Currently, when running `hardhat` in a lerna setup and running e.g. `hardhat compile` in many packages at once, the prompt for consent silently fails but doesn't exit the process, thus leaving the user with no way to realize that something went wrong at all.

To fix this for them, the user then has to `cd` into one of the packages manually and run `hardhat compile` in there specifically to answer the consent prompt before being able to run it at the root of the `lerna` repository. This all obviously first requires them to realize what happened and why the previous command silently exited.

I am not sure what the best way to fix this is... I personally prefer not having prompts like this at all, because they are quite intrusive, or having a way to skip them through configuration in hardhat.config.js entirely.

This PR allows to skip it both through config and/or through a cli argument.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -9,11 +9,12 @@ The entirety of your Hardhat setup (i.e. your config, plugins and custom tasks) 
 
 To set up your config, you have to export an object from `hardhat.config.js`.
 
-This object can have the following entries: `defaultNetwork`, [`networks`](#networks-configuration), [`solidity`](#solidity-configuration), and [`paths`](#path-configuration). For example:
+This object can have the following entries: `defaultNetwork`, `disableTelemetry` [`networks`](#networks-configuration), [`solidity`](#solidity-configuration), and [`paths`](#path-configuration). For example:
 
 ```js
 module.exports = {
   defaultNetwork: "rinkeby",
+  disableTelemetry: true,
   networks: {
     hardhat: {
     },
--- a/docs/getting-started/README.md
+++ b/docs/getting-started/README.md
@@ -77,6 +77,7 @@ GLOBAL OPTIONS:
   --max-memory          The maximum amount of memory that Hardhat can use.
   --network             The network to connect to.
   --show-stack-traces   Show stack traces.
+  --disable-telemetry   Disable collection of telemetry data.
   --tsconfig            Reserved hardhat argument -- Has no effect.
   --verbose             Enables Hardhat verbose logging
   --version             Shows hardhat's version.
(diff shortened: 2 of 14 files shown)


fubhy:
Btw. just as a heads-up: With the default vscode config with prettier configured in visual studio, writing PRs for hardhat is currently quite cumbersome because it tries to re-format each file I am editing. I could of course change my vscode config for this project but just wanted to bring this up... Maybe you want to consider also adding a prettier config at the root?

alcuadrado:
Thanks for submitting this, @fubhy. Can you help us reproduce the problem first? We'd rather understand what's happening before going for any kind of solution.

alcuadrado:
> Btw. just as a heads-up: With the default vscode config with prettier configured in visual studio, writing PRs for hardhat is currently quite cumbersome because it tries to re-format each file I am editing. I could of course change my vscode config for this project but just wanted to bring this up... Maybe you want to consider also adding a prettier config at the root?

I don't understand this. We use prettier without any special config on purpose, so the default should just work.

fubhy:
> > Btw. just as a heads-up: With the default vscode config with prettier configured in visual studio, writing PRs for hardhat is currently quite cumbersome because it tries to re-format each file I am editing. I could of course change my vscode config for this project but just wanted to bring this up... Maybe you want to consider also adding a prettier config at the root?
> 
> I don't understand this. We use prettier without any special config on purpose, so the default should just work.

I've got `format-on-save` configured and, at least in my vscode setup, prettier constantly tries to re-format stuff when I edit hardhat files. I can check the reason for that on my next PR. It might well be that it's a version mismatch. Will check next time.

> Thanks for submitting this, @fubhy. Can you help us reproduce the problem first? We'd rather understand what's happening before going for any kind of solution.

The problem appears with any `lerna` project that hosts multiple packages that e.g. run `hardhat compile`. Now, if I invoke those with `lerna run compile` from the root without `--stream` it simply gets stuck forever without any visual feedback. However, if run with `--stream` lerna just runs past the prompt. When using `--stream`, one can at least see the text but but it's not waiting for me to fill in the prompt.

alcuadrado:
Does that happen in your CI, or is the problem present in your workstation?

fubhy:
> Does that happen in your CI, or is the problem present in your workstation?

On the CI this is mitigated by the new check. This only happens on the workstation of someone who has previously NOT worked with hardhat (and thus confirmed the dialog). It occurred in all of my colleagues dev environments after I switched our repository over to hardhat.

fubhy:
So if you want to reproduce it you have to clear your recorded hardhat config in `~/.config/hardhat-nodejs/telemetry-consent.json`

alcuadrado:
Thanks for the info, @fubhy.

I implemented an alternative fix in #956. While the hardhat argument is nice, it has the disadvantage of not being immediately obvious when to use it or not. The normal flow would be users having their processes hang, spending some time figuring out what's going on, and just then using it. 

Disabling it in the project config is not something we want to implement, as giving or not consent is a personal decision, and a flag in the config changes that.

PR #962, opened 2020-10-27
Update buidler-typechain migration status in guide.

scnale:


alcuadrado:
Nice! Thanks @scnale !

PR #966, opened 2020-10-28
Create classes handling the transaction pool logic

msieczko:
This PR introduces two new classes:

- `TxPool` - responsible for validation, storage and retrieval of not-yet-mined transactions. It divides the incoming transactions into two groups: _pending_ and _queued_. **Pending** transactions are the ones that, given the current sender `nonce`, could be executed one by one (there are no gaps in the nonce sequence). The transactions which nonces are too high are **queued**. To save the order in which transactions are added each of them is assigned an `orderId`.
- `TxPriorityHeap` - a structure which implements the sorting strategy of transactions in a new block. It allows to retrieve the next processable transaction with the highest gas price and the lowest order id. It will be used by `HardhatNode.mineBlock` function later on.


config/typescript/@types/mnemonist/index.d.ts
```diff
@@ -0,0 +1,23 @@
+declare module "mnemonist/heap" {
```
fvictorio, line 1: I guess this is because mnemonist types are wrong. Maybe it's worth it opening a PR fixing them in their repo.

packages/hardhat-core/test/internal/hardhat-network/helpers/assertEqualTransactionMaps.ts
```diff
@@ -0,0 +1,238 @@
+import { assert } from "chai";
```
fvictorio, line 1: What is this file testing exactly? It seems to create two helpers that aren't used outside of this file and then it tests them. Are they meant to be used in the future but aren't being used right now?

fvictorio (commented):
This is not a proper review, I was just skimming the PR and noticed a couple of things.

config/typescript/@types/mnemonist/index.d.ts
msieczko: Yes, that's the reason. Good point.

packages/hardhat-core/test/internal/hardhat-network/helpers/assertEqualTransactionMaps.ts
msieczko: The `assertEqualTransactionMaps` function is actually used in the tests for `TxPool` class right now. For instance [here](https://github.com/EthWorks/buidler/blob/transaction-pool/packages/hardhat-core/test/internal/hardhat-network/provider/TxPool.ts#L501).

fvictorio: Oh, my bad, that file was not shown in the diff because it was too big and so I didn't catch it doing ctrl-f. Thanks!

fvictorio:
Closing this in favor of #1087.

PR #969, opened 2020-10-29
Warn against standalone types in hardhat-ethers type extensions.

scnale:
This removes standalone types in `hardhat-ethers` type extensions.


scnale:
Removing these types could break some projects that imported `Libraries` or `FactoryOptions` already. I'll add a few comments to warn that this isn't a good practice instead.

PR #974, opened 2020-10-30
Fix config link

morelazers:


alcuadrado:
Thanks @morelazers! I'll merge this and add a redirect.

PR #979, opened 2020-11-02
Set relative block timestamp and proper nonces in forked network

fubhy:
This is important for tests that check block.timestamp (e.g. staleness of price oracle data / lock period, etc.).

Let me know if this approach is how you'd imagine this to work. I am happy to add tests if this is the path you want to take for this.


fubhy:
@fvictorio @alcuadrado Anything I can do to help get a review on this? If you agree that this is the way you want to have it implemented, I'd be happy to add test coverage to get this merged.

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
```diff
@@ -57,6 +63,13 @@ Please use block number ${lastSafeBlock} or wait for the block to get ${
     forkBlockNumber = new BN(lastSafeBlock);
   }
 
+  const block = (await provider.request({
```
fvictorio, line 66: Let's move this to a function, similar to how `getNetworkId` and `getLatestBlockNumber` are done.

fvictorio (commented):
LGTM, but I'm not very familiar with this part of the code. I would wait for @alcuadrado's review, but I think you can start adding tests anyway.

fubhy:
@alcuadrado @fvictorio Hey guys. Any update regarding this? Last time we spoke you said you had plans to tackle this potentially as a part of a broader refactoring of how the fork config works? If this is further out in the future, can we get the ball rolling for this pull request in the meantime? :)

fubhy:
Ping

crazyrabbitLTC:
Another Ping. Also interested in this.

fvictorio:
Hey, sorry for taking so long to look at this, I will revisit it this week.

fvictorio:
@alcuadrado this is ready for review

packages/hardhat-core/src/builtin-tasks/node.ts
```diff
@@ -93,12 +93,26 @@ subtask(TASK_NODE_GET_PROVIDER)
...
+          ERRORS.BUILTIN_TASKS.NODE_FORK_BLOCK_NUMBER_WITHOUT_URL
+        );
+      }
+
+      // if the url or the block is different to the one in the configuration,
+      // we use hardhat_reset to set the fork
+      if (
+        forkUrl !== forkUrlConfig ||
+        forkBlockNumber !== forkBlockNumberConfig
+      ) {
```
fvictorio, line 115: This is the fix for the double-initialization. I might be being too clever here, but I'm not checking that `forkUrl` and `forkBlockNumber` aren't undefined because they can't be both undefined AND different to the config. Maybe it's worth checking that explicitly, I don't know.

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
fvictorio: Ugh, forgot my own comment here. But this is a low-risk change, so I think this PR can still be reviewed.

packages/hardhat-core/test/internal/hardhat-network/provider/forked-provider.ts
```diff
@@ -539,6 +556,38 @@ describe("Forked provider", () => {
           });
         });
       });
+
+      describe("blocks timestamps", () => {
+        it("should use a timestamp relative to the forked block timestamp", async function () {
```
fvictorio, line 561: I don't know if there is a better way of testing this. I'm using the first block of 2021-01-01, mining a new one, and checking that this new block has a timestamp on that day.

alcuadrado (approved):

PR #981, opened 2020-11-02
Add example for using an impersonated account

fvictorio:


mmd-mostafaee:
Can you also add an example to use impersonated account out of hardhat?
For example in python or using rpc url directly

PR #986, opened 2020-11-03
Return block in callbackified blockchain

fvictorio:
This fix is literally one line, but I also modified a bunch of typescript stuff to make typescript detect the previous issue. I'm not sure at all about those changes.

I also added some minimal tests for the callbackified version of the `Blockchain` class. One of them is skipped for reasons I will explain in the comments.


packages/hardhat-core/src/internal/hardhat-network/provider/types/Callback.ts
```diff
@@ -1,2 +1 @@
-export type Callback<T = void> = ((error: NonNullable<any>) => void) &
-  ((error: null | undefined, value: T) => void);
+export type Callback<T = void> = (error: any, value: T) => void;
```
fvictorio, line 1: This is the type change that triggers all the other typescript changes. This makes sense to me, but I have no idea what I'm doing.

packages/hardhat-core/test/internal/hardhat-network/provider/HardhatBlockchain.ts
```diff
@@ -54,6 +61,17 @@ describe("HardhatBlockchain", () => {
       assert.equal(await blockchain.getBlock(one.hash()), one);
     });
 
+    it("can get existing block by hash (callbackified)", async () => {
+      const genesis = createBlock(0);
+      const one = createBlock(1);
+      await promisify<Block>(callbackifiedBlockchain.putBlock)(genesis);
+      await promisify<Block>(callbackifiedBlockchain.putBlock)(one);
```
fvictorio, line 68: This is ugly, I'd rather have a `promisifiedCallbackifiedBlockchain` (without that name, I swear) and just duplicate the previous test as is.

packages/hardhat-core/test/internal/hardhat-network/provider/HardhatBlockchain.ts
```diff
@@ -146,6 +176,34 @@ describe("HardhatBlockchain", () => {
       assert.equal(await blockchain.getBlock(blockThree.hash()), undefined);
     });
 
+    it.skip("removes the block and all subsequent ones (callbackified)", async () => {
```
fvictorio, line 179: This test is a duplicated version of the previous one, like the other added tests. But this one fails because of this line:

https://github.com/nomiclabs/hardhat/blob/7b606216090859ac0edbca105d6cbd49113c209a/packages/hardhat-core/src/internal/hardhat-network/provider/types/PBlockchain.ts#L34

I don't know if that line is fine, and so this tests needs to be different, or if that undefined check should be removed

PR #988, opened 2020-11-04
Improve anonymizer.

scnale:


packages/hardhat-core/src/internal/sentry/anonymizer.ts
```diff
@@ -93,6 +93,14 @@ export class Anonymizer {
   }
 
   public anonymizeErrorMessage(errorMessage: string): string {
+    const mnemonicPhraseLengthThreshold = 7;
+    const minimumAmountOfWordsToAnonymize = 4;
```
fvictorio, line 97: These can be module-level constants, no need to parameterize a private function.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
```diff
@@ -226,4 +234,80 @@ export class Anonymizer {
...
+        require("ethereum-cryptography/bip39/wordlists/japanese"),
+        require("ethereum-cryptography/bip39/wordlists/korean"),
+        require("ethereum-cryptography/bip39/wordlists/simplified-chinese"),
+        require("ethereum-cryptography/bip39/wordlists/spanish"),
+        require("ethereum-cryptography/bip39/wordlists/traditional-chinese"),
+      ].map((wordlistModule) => wordlistModule.wordlist)
+    );
+    for (const [fragmentIndex, fragment] of fragments.entries()) {
+      // TODO: split on whitespace instead of spaces?
+      const words = fragment.split(" ");
```
fvictorio, line 261: I think there's no downside on using `/\s+/` here.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
```diff
@@ -226,4 +234,80 @@ export class Anonymizer {
...
+        require("ethereum-cryptography/bip39/wordlists/french"),
+        require("ethereum-cryptography/bip39/wordlists/italian"),
+        require("ethereum-cryptography/bip39/wordlists/japanese"),
+        require("ethereum-cryptography/bip39/wordlists/korean"),
+        require("ethereum-cryptography/bip39/wordlists/simplified-chinese"),
+        require("ethereum-cryptography/bip39/wordlists/spanish"),
+        require("ethereum-cryptography/bip39/wordlists/traditional-chinese"),
+      ].map((wordlistModule) => wordlistModule.wordlist)
+    );
+    for (const [fragmentIndex, fragment] of fragments.entries()) {
```
fvictorio, line 259: Why entries? `fragmentIndex` is not used.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
```diff
@@ -226,4 +234,80 @@ export class Anonymizer {
...
+  }
+}
+
+function getMaximalMnemonicPhrase(
+  words: string[],
+  startIndex: number,
+  wordlist: string[]
+): string[] {
+  const maximalPhrase: string[] = [];
+  for (let endIndex = startIndex; endIndex < words.length; endIndex++) {
```
fvictorio, line 304: Why `endIndex`? I think this would be way clearer if it was just `i`.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
```diff
@@ -226,4 +234,80 @@ export class Anonymizer {
...
+    );
+    for (const [fragmentIndex, fragment] of fragments.entries()) {
+      // TODO: split on whitespace instead of spaces?
+      const words = fragment.split(" ");
+      if (words.length >= mnemonicPhraseLengthThreshold) {
+        // Determine all mnemonic phrase maximal fragments.
+        // We check sequences of n consecutive words just in case there is a typo
+        for (
+          let wordIndex = 0;
+          wordIndex < words.length - minimumAmountOfWordsToAnonymize + 1;
```
fvictorio, line 267: I understand why the `- minimum...` is here, but it seems like a negligible optimization that only adds complexity. I'm ok with keeping it, but I would remove it.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
fvictorio: Besides, this means that the heuristic won't work if the mnemonic is separated by line breaks, right? Not sure how likely that is, but given what's at stake I'd rather be over-paranoid.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
```diff
@@ -226,4 +234,80 @@ export class Anonymizer {
...
+        ) {
+          const maximalPhrase = getMaximalMnemonicPhrase(
+            words,
+            wordIndex,
+            wordlist
+          );
+
+          if (maximalPhrase.length >= minimumAmountOfWordsToAnonymize) {
+            // Mark these words
+            fragmentsToBeAnonymized.push(maximalPhrase.join(" "));
```
fvictorio, line 278: This will need to be re-think if we use `/\s+/` as separator.

packages/hardhat-core/test/internal/sentry/anonymizer.ts
```diff
@@ -299,5 +299,79 @@ describe("Anonymizer", () => {
...
+      assert.equal(
+        anonymizedErrorMessage,
+        "My mnemonic phrase is <mnemonic> tset <mnemonic>. This is an error message."
+      );
+    });
+
+    it("should hide mnemonic phrases with ideograms", () => {
+      const anonymizer = new Anonymizer();
+      const errorMessage =
+        "My mnemonic phrase is 就 就 就 就 就 就 就 就 就 就 就 就. This is an error message.";
```
fvictorio, line 334: This is great

packages/hardhat-core/src/internal/sentry/anonymizer.ts
fvictorio: By the way, I don't understand why we need two constants? It feels like this could be done with just one.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
scnale: That's right. The problem is that I need to reconstruct the string as it originally was. I think `String.prototype.split` doesn't help much here...

scnale: Actually, now that I think about it, it might be fine to mangle the string a bit just to cover these additional cases.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
scnale: Obsolete leftover :P

packages/hardhat-core/src/internal/sentry/anonymizer.ts
fvictorio: Yes, I'm afraid doing this requires some careful usage of regex. I'm still inclined to do it, but I'm not 100% sure. @alcuadrado wdyt?

packages/hardhat-core/src/internal/sentry/anonymizer.ts
alcuadrado: We discussed this with @scnale in a call.

The first one is just an optimization, to avoid running the mnemonic detection algorithm on every string.

The second one is which is the minimum number of consecutive mnemonic words to clear. The reason for this is that we still want to anonymize most of it if there's a typo in one of its words. This way, we would anonymize `minimumAmountOfWordsToAnonymize+` words on each side of the typo.

The first one is just a performance optimization. The second one is actually how many words we anonymize at once. TH

alcuadrado: I agree about the module level constants

packages/hardhat-core/src/internal/sentry/anonymizer.ts
```diff
@@ -226,4 +234,80 @@ export class Anonymizer {
...
   }
+
+  private _anonymizeMnemonic(
+    errorMessage: string,
+    mnemonicPhraseLengthThreshold: number,
+    minimumAmountOfWordsToAnonymize: number
+  ): string {
+    const fragments = errorMessage.split(/[!$%^&*()_+|~\-=`{}[\]:;<>?"',.]/gu);
+    const fragmentsToBeAnonymized: string[] = [];
+    // If there are 10 consecutive words, there's a good chance of there being a mnemonic phrase
```
alcuadrado, line 245: what's the 10 here?

scnale: Hm, that's old. It's a reference to the `mnemonicPhraseLengthThreshold` constant.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
alcuadrado: Maybe split with `/\b/`, and then make the algorithm ignore the strings that match `/\s+/`?

alcuadrado: `\b` is word boundary, so for example

```js
"asd asd asd".split(/\b/)
```

is

```js
[ 'asd', ' ', 'asd', ' ', 'asd' ]
```

scnale: Sadly, it seems it's not enough:
```js
> '就 就 就 就 就 就 就 就 就 就 就 就'.split(/\b/u)
[ '就 就 就 就 就 就 就 就 就 就 就 就' ]
```

fvictorio: I forgot to update this here, but we talked about using `/\p{Letter}+/u` to get the list of words of a string. It seems robust enough.

alcuadrado: I'd split it manually rather than using something that obscure. Something like this:

```js
function splitByWordBoundaries(str) {
  const parts = [];
  let withinWord = false;

  let current = "";
  for (const codePoint of str) {
    if (codePoint.match(/\s/u)) {
      if (withinWord) {
        if (current !== "") {
          parts.push(current);
        }
        current = codePoint;
        withinWord = false;
      } else {
        current += codePoint;
      }
    } else {
      if (withinWord) {
        current += codePoint;
      } else {
        if (current !== "") {
          parts.push(current);
        }
        current = codePoint;
        withinWord = true;
      }
    }
  }

  if (current !== "") {
    parts.push(current);
  }

  return parts;
}
```

packages/hardhat-core/src/internal/sentry/anonymizer.ts
```diff
@@ -226,4 +236,108 @@ export class Anonymizer {
...
+          const nextWordIndex =
+            wordIndex + maximalPhrase.length < matches.length
+              ? matches[wordIndex + maximalPhrase.length].index
+              : errorMessage.length;
+          const sliceUntilNextWord = errorMessage.slice(
+            lastAnonymizedWord.index + lastAnonymizedWord.word.length,
+            nextWordIndex
+          );
+          anonymizedMessage += `${ANONYMIZED_MNEMONIC}${sliceUntilNextWord}`;
+          wordIndex += maximalPhrase.length - 1;
```
fvictorio, line 281: The `-1` is here to cancel the `++` of the for loop, right? That makes me think that this shoul be a `while` instead.

fvictorio:
Added a couple of failing test cases, and the fix.

packages/hardhat-core/src/internal/sentry/anonymizer.ts
scnale: That's right. I was actually thinking the same.

PR #989, opened 2020-11-04
Improve compilation job error messages

fvictorio:
This PR improves the errors shown when the files can't be compiled.

### DIFF
--- a/packages/hardhat-core/src/builtin-tasks/compile.ts
+++ b/packages/hardhat-core/src/builtin-tasks/compile.ts
@@ -29,13 +29,12 @@ import { ResolvedFile, Resolver } from "../internal/solidity/resolver";
 import { glob } from "../internal/util/glob";
 import { getCompilersDir } from "../internal/util/global-dir";
 import { pluralize } from "../internal/util/strings";
-import { unsafeObjectEntries, unsafeObjectKeys } from "../internal/util/unsafe";
 import { Artifacts, CompilerInput, CompilerOutput, SolcBuild } from "../types";
 import * as taskTypes from "../types/builtin-tasks";
 import {
   CompilationJob,
   CompilationJobCreationError,
-  CompilationJobsCreationErrors,
+  CompilationJobCreationErrorReason,
   CompilationJobsCreationResult,
 } from "../types/builtin-tasks";
 import { getFullyQualifiedName } from "../utils/contract-names";
@@ -272,22 +271,15 @@ subtask(TASK_COMPILE_SOLIDITY_GET_COMPILATION_JOBS)
         )
       );
 
-      const compilationJobsCreationResult = compilationJobsCreationResults.reduce(
-        (acc, { jobs, errors }) => {
-          acc.jobs = acc.jobs.concat(jobs);
-          for (const [code, files] of unsafeObjectEntries(errors)) {
-            acc.errors[code] = acc.errors[code] ?? [];
-            acc.errors[code] = acc.errors[code]!.concat(files!);
-          }
-          return acc;
-        },
-        {
-          jobs: [],
-          errors: {},
-        }
-      );
+      let jobs: CompilationJob[] = [];
+      let errors: CompilationJobCreationError[] = [];
 
-      return compilationJobsCreationResult;
+      for (const result of compilationJobsCreationResults) {
+        jobs = jobs.concat(result.jobs);
+        errors = errors.concat(result.errors);
+      }
+
+      return { jobs, errors };
     }
   );
 
@@ -446,7 +438,6 @@ subtask(TASK_COMPILE_SOLIDITY_LOG_DOWNLOAD_COMPILER_START)
   .setAction(
     async ({
       isCompilerDownloaded,
-      quiet,
       solcVersion,
     }: {
       isCompilerDownloaded: boolean;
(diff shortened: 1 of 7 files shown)


packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -981,13 +972,11 @@ subtask(TASK_COMPILE_SOLIDITY_HANDLE_COMPILATION_JOBS_FAILURES)
       {
         compilationJobsCreationErrors,
       }: {
-        compilationJobsCreationErrors: CompilationJobsCreationErrors;
+        compilationJobsCreationErrors: CompilationJobCreationError[];
       },
       { run }
     ) => {
```
fvictorio, line 978: This function is *huge*. I think we should move it to another module, but I didn't do it yet so that the PR it's easier to review.

Also, there's a lot of duplication here, especially between the "invalid imports" and the "invalid indirect imports" parts. Maybe we can abstract that, but I'm inclined to leave this as straightforward as possible.

packages/hardhat-core/src/types/builtin-tasks/compile.ts
```diff
@@ -64,18 +64,21 @@ export interface DependencyGraph {
...
 }
 
-export type CompilationJobsCreationErrors = {
-  [error in CompilationJobCreationError]?: string[];
-};
+export interface CompilationJobCreationError {
+  reason: CompilationJobCreationErrorReason;
+  file: ResolvedFile;
+  extra?: any;
+}
```
fvictorio, line 74: This is the main type change. The `CompilationJobsCreationError` type, that was a map from an error tipe to a list of source names with that error, is gone. Instead I used a list of errors. Each error has a reason (this is the extendable enum previously called CompilationJobCreationError), a file, and an optional "extra" type, that can have more information about the error.

I don't love having an `any` for the extra field, but I'm not sure it's worth it to do something trickier here.

packages/hardhat-core/test/internal/solidity/compilation-job.ts
```diff
@@ -6,7 +6,10 @@ import {
 } from "../../../src/internal/solidity/compilation-job";
 import { ResolvedFile } from "../../../src/internal/solidity/resolver";
 import * as taskTypes from "../../../src/types/builtin-tasks";
-import { CompilationJobCreationError } from "../../../src/types/builtin-tasks";
+import {
+  CompilationJobCreationError,
+  CompilationJobCreationErrorReason,
+} from "../../../src/types/builtin-tasks";
```
fvictorio, line 12: This file needs more tests, but I will add them after the new types are definitive.

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -1016,78 +1005,186 @@ subtask(TASK_COMPILE_SOLIDITY_GET_COMPILATION_JOBS_FAILURE_REASONS)
...
+            `File ${sourceName} imports files ${incompatibleImportsFiles
+              .map((x) => x.sourceName)
+              .join(", ")} that use an incompatible version of Solidity`
+          );
+
+          let importsText = "";
+          if (incompatibleImports.length === 1) {
+            importsText = ` imports ${incompatibleImports[0]}`;
+          } else if (incompatibleImports.length === 2) {
+            importsText = ` imports ${incompatibleImports[0]} and ${incompatibleImports[1]}`;
```
alcuadrado, line 1114: Are these direct imports?

alcuadrado: ah, yes. They are. Maybe add a "direct" to the var names?

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -1016,78 +1005,186 @@ subtask(TASK_COMPILE_SOLIDITY_GET_COMPILATION_JOBS_FAILURE_REASONS)
...
+          const incompatibleImportsFiles: ResolvedFile[] =
+            error.extra?.incompatibleIndirectImports ?? [];
+
+          const incompatibleImports = incompatibleImportsFiles.map(
+            (x: ResolvedFile) =>
+              `${x.sourceName} (${x.content.versionPragmas.join(" ")})`
+          );
+
+          log(
+            `File ${sourceName} depends on files ${incompatibleImportsFiles
```
alcuadrado, line 1150: We could add the entire import path here, can't we?

Something like a second line with "Import path: A.sol > B.sol > Incompatible.sol"

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -1016,78 +1005,186 @@ subtask(TASK_COMPILE_SOLIDITY_GET_COMPILATION_JOBS_FAILURE_REASONS)
...
+        errorMessage += `These files and its dependencies cannot be compiled with your config:
 
-${other.map((x) => `* ${x}`).join("\n")}
+${other.map((x) => `* ${x.file.sourceName}`).join("\n")}
 
 `;
       }
 
-      reasons += `Learn more about compiler configuration at https://hardhat.org/config
+      errorMessage += `Learn more about compiler configuration at https://hardhat.org/config
```
alcuadrado, line 1184:
```suggestion
      errorMessage += `This can be because they have incompatible Solidity pragmas, or don't match any of your configured Solidity compilers. Learn more about compiler configuration at https://hardhat.org/config
```

fvictorio: Couldn't that message be misleading? If there are no matching compilers or if there are incompatible imports, we should have detected it and reported it as a different error reason.

alcuadrado: I believe this error has two reasons:

* The range described by the concatenation of all the pragmas is empty.
* The range described by the concatenation of all the pragmas doesn't include any of the configured compilers, even though each individual pragma does match at least one of the compilers. 

Is that right?

If so, we should aim to explain that without too many details.

PR #1002, opened 2020-11-09
[@nomiclabs/hardhat-etherscan] Only upload related sources via --related-sources flag

kartojal:
Fixes https://github.com/nomiclabs/hardhat/issues/804

### DIFF
--- a/packages/hardhat-etherscan/package.json
+++ b/packages/hardhat-etherscan/package.json
@@ -37,6 +37,7 @@
     "@ethersproject/address": "^5.0.2",
     "cbor": "^5.0.2",
     "fs-extra": "^7.0.1",
+    "lodash": "^4.17.20",
     "node-fetch": "^2.6.0",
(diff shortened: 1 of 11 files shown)


kartojal:
Having some issues running `@nomiclabs/hardhat-etherscan` tests, in local and CI, but don't see why it fails, due my changes are minimal and followed the contributing guidelines. Do i miss any step? Can you have a look? Thanks.

https://github.com/nomiclabs/hardhat/pull/1002/checks?check_run_id=1374557005#step:9:6445

alcuadrado:
Hey @kartojal, thanks for creating this PR. Unfortunately, we were already working on this same thing, so this is not mergeable.

The good news is that a (different) fix to the same problem will be released this week.

We also added a section to our `CONTRIBUTING.md` file suggesting to create an issue before sending a PR, so this situation can be avoided in the future.

PR #1006, opened 2020-11-10
Use content-hash to invalidate cache

fvictorio:


packages/hardhat-core/src/builtin-tasks/utils/solidity-files-cache.ts
```diff
@@ -52,30 +57,26 @@ export class SolidityFilesCache {
...
-      await solidityFilesCache.removeModifiedFiles();
+      await solidityFilesCache.removeNonExistingFiles();
       return solidityFilesCache;
     }
 
-    // tslint:disable-next-line only-hardhat-error
-    throw new Error("Couldn't read cache file, try running the clean task"); // TODO use HardhatError
+    log("There was a problem reading the cache");
+
+    return new SolidityFilesCache({
```
alcuadrado, line 66: This will clean the cache as soon as the user upgrades, right?

alcuadrado: I mean, because of the change in the cache file format.

packages/hardhat-core/src/internal/solidity/parse.ts
```diff
@@ -55,12 +59,28 @@ export class Parser {
...
+   * Get parsed data from the internal cache, or from the solidity files cache.
+   *
+   * Returns null if cannot find it in either one.
+   */
+  private _getFromCache(
+    absolutePath: string,
+    contentHash: string
+  ): ParsedData | null {
+    if (this._solidityFilesCache === undefined) {
+      return this._cache.get(absolutePath) ?? null;
```
alcuadrado, line 72: What if we use the content hash instead of the absolute path here? I think it will be more robust. I can imagine using the `absolutePath` bringing issues for things like the `hardhat-watcher` plugin.

packages/hardhat-core/src/internal/solidity/parse.ts
```diff
@@ -55,12 +59,28 @@ export class Parser {
...
+    }
+
+    const cacheEntry = this._solidityFilesCache.getEntry(absolutePath);
+
+    if (cacheEntry === undefined) {
+      return this._cache.get(absolutePath) ?? null;
```
alcuadrado, line 78: Same

alcuadrado (approved):
This turned out to be a very elegant solution. 

The only thing I'd change is the Parser's in-memory cache.

packages/hardhat-core/src/builtin-tasks/utils/solidity-files-cache.ts
fvictorio: Yes, I forgot to add a comment about this.

I think in an ideal world we would handle changes in the files format in a graceful way. For example: detecting that the file has the format v1 but the code expects format v2, and acting accordingly. But in this case, just ignoring everything and starting over is both easier and (I think) correct. We might need to change this approach in the future, but for now I think this is enough.

fvictorio:
> The only thing I'd change is the Parser's in-memory cache.

You mean we keep the `[absolutePath: string]: CacheEntry` structure, but we use the hash for indexing only for the in memory object that represents the cache?

packages/hardhat-core/src/builtin-tasks/utils/solidity-files-cache.ts
alcuadrado: agreed

PR #1016, opened 2020-11-12
Fix how the state of the fork block is handled

### LINKED ISSUE #1004: eth_getBalance with block parameter returns incorrect value in fork mode
Steps for reproducing:
- `npx hardhat node --fork https://fork_node`
- `curl --data '{"id":6185245701164,"jsonrpc":"2.0","params":[{"from":"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266","to":"0x70997970c51812dc3a010c7d01b50e0d17dc79c8","gas":"0x5208","gasPrice":"0x3B9ACA00","value":"0x1","data":"0x","nonce": "0x0"}],"method":"eth_sendTransaction"}' -H "Content-Type: application/json"`
- `curl --data '{"method":"eth_getBalance","params":["0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266", "{current_block - 1}"],"id":1,"jsonrpc":"2.0"}' -H "Content-Type: application/json" -X POST http://127.0.0.1:8545`

The third step should return `0x21e19e0c9bab2400000` but instead, the result is `0x0`, another odd thing is that the third step will return the correct value if called before the second step, even with the block parameter.

This issue is not reproducible in normal mode.

fvictorio:
Closes #1004.


packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
```diff
@@ -50,6 +51,15 @@ export class ForkStateManager implements PStateManager {
...
   }
 
+  /**
+   * The forked block needs special handling because the genesis accounts are added to it.
+   * This is meant to be called when the initial state is final, so that _initialStateRoot
+   * points to this starting fork state.
+   */
+  public updateInitialStateRoot() {
+    this._stateRootToState.set(this._initialStateRoot, this._state);
+  }
```
msieczko, line 61: Instead of adding this function and a long comment explaining its purpose I would put all of the initialization logic inside the FSM constructor. In other words, move this:
https://github.com/nomiclabs/hardhat/blob/b21474ce1df891edd7fb7b94f8bbc079c652b85a/packages/hardhat-core/src/internal/hardhat-network/provider/utils/putGenesisAccounts.ts#L10-L13
and make `genesisAccounts` a constructor parameter.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
```diff
@@ -319,6 +329,10 @@ export class ForkStateManager implements PStateManager {
     if (this._stateCheckpoints.length !== 0) {
       throw checkpointedError("setBlockContext");
     }
+    if (blockNumber.eq(this._forkBlockNumber)) {
+      this._setStateRoot(toBuffer(this._initialStateRoot));
+      return;
+    }
```
msieczko, line 335: I would add a test covering the new if statement. Also this describe name in existing tests is no longer valid:
https://github.com/nomiclabs/hardhat/blob/b21474ce1df891edd7fb7b94f8bbc079c652b85a/packages/hardhat-core/test/internal/hardhat-network/provider/fork/ForkStateManager.ts#L613

msieczko, line 335: Instead of adding this if statement, we could make the `this._initialStateRoot` equal to the actual `stateRoot` of the fork block. However this would require fetching the block before FSM construction and would slow down the HardhatNode creation. Having considered this I would leave this the way it is.

msieczko (changes requested):
Overall I think the fix is adequate and I would do it the same way 👍

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
msieczko: Btw, `putAccount` is async only for the purpose of compliance with the `PStateManager` interface. I'd add a private synchronous `_putAccount` method and call it in the constructor and in the original public `putAccount`.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
fvictorio: I did exactly the same reasoning, I'm glad that you agree :sweat_smile:

fvictorio: I added this one that I used while developing this: https://github.com/nomiclabs/hardhat/blob/b21474ce1df891edd7fb7b94f8bbc079c652b85a/packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth.ts#L701

msieczko (approved):

alcuadrado (approved):

PR #1024, opened 2020-11-16
Reset ethers provider after a hardhat_reset

### LINKED ISSUE #1015: hardhat-ethers getBlockNumber is wrong after a call to hardhat_reset.
This bug was reported on discord.

If you use `hardhat-ethers`, for example, the sample project, and run:

```ts
await ethers.provider.send("evm_mine");
await ethers.provider.send("evm_mine");
await ethers.provider.getBlockNumber(); // 2

await ethers.provider.send("hardhat_reset");
await ethers.provider.getBlockNumber(); // 2 -- should be 0
await network.provider.send("eth_blockNumber") // '0x0'
```

fvictorio:
Closes #1015 

There are a lot of changes in the test file, but they are mostly whitespace because of an extra `describe` block around them.


packages/hardhat-core/src/internal/core/providers/http.ts
```diff
@@ -58,6 +58,10 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
       throw error;
     }
 
+    if (args.method === "hardhat_reset") {
+      this.emit("hardhat_reset");
```
fvictorio, line 62: Not sure about the name of this event.

alcuadrado: maybe `hardhatNetwokReset`? I'd create a constant.

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -94,11 +94,18 @@ export class HardhatNetworkProvider extends EventEmitter
...
+      let result;
       if (this._loggingEnabled && !PRIVATE_RPC_METHODS.has(args.method)) {
-        return await this._sendWithLogging(args.method, args.params);
+        result = await this._sendWithLogging(args.method, args.params);
+      } else {
+        result = await this._send(args.method, args.params);
+      }
+
+      if (args.method === "hardhat_reset") {
+        this.emit("hardhat_reset");
```
alcuadrado, line 105: Shouldn't we emit it in the node and just fwd it in the provider?

fvictorio: As far as I understand, a new node is created on a reset, so that wouldn't be possible, would it?

alcuadrado (approved):

PR #1025, opened 2020-11-16
Add warning about metamask issue

fvictorio:


packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -851,6 +852,11 @@ export class EthModule {
       }
 
       if (error.message.includes("EIP155")) {
+        console.warn(
```
alcuadrado, line 855: I'd just add it to the normal log flow. This `console.warn` can bring problems, like ignoring the loggingEnabled setting.

alcuadrado (changes requested):

fvictorio:
Looks like this:

image

Not sure about the indentation.

PR #1027, opened 2020-11-17
Improve autocomplete

fvictorio:


packages/hardhat-core/test/internal/cli/autocomplete.ts
```diff
@@ -33,18 +34,22 @@ const coreTasks = [
...
+  "--network",
+  "--show-stack-traces",
   "--tsconfig",
   "--verbose",
+  "--version",
 ];
 
-describe("autocomplete", () => {
+describe("autocomplete", function () {
+  if (os.type() === "Windows_NT") {
```
fvictorio, line 49: I did this mainly to avoid having to use `path.sep` in a lot of places, since bash/zsh/fish don't work on Windows, right? (And if they are emulated or something, they use forwards slashes AFAIK.)

packages/hardhat-core/src/internal/cli/autocomplete.ts
```diff
@@ -97,7 +102,58 @@ export async function complete({
...
+
+  // dirty trick when we match a single directory:
+  // we don't want to have a single suggestion because otherwise the shell
+  // will add an space at the end, which is annoying,
+  // so we suggest both the dir, and the dir with a trailing path separator
+  if (suggestions.length === 1) {
+    const [suggestion] = suggestions;
+
+    if (isDirectory(suggestion)) {
+      return [suggestion, path.join(suggestion, path.sep)];
```
fvictorio, line 143: I'm not proud of this, but it works surprisingly well.

packages/hardhat-core/src/internal/cli/autocomplete.ts
```diff
@@ -1,28 +1,59 @@
...
 }
 
+interface CompletionData {
+  networks: string[];
+  tasks: {
+    [taskName: string]: {
+      name: string;
+      paramDefinitions: {
+        [paramName: string]: {
+          name: string;
```
alcuadrado, line 25: is the name stored twice? don't we need to store the type too? or maybe this is the typename?

packages/hardhat-core/src/internal/cli/autocomplete.ts
```diff
@@ -1,28 +1,59 @@
+import findup from "find-up";
+import * as fs from "fs-extra";
+import * as path from "path";
+
 import { HARDHAT_PARAM_DEFINITIONS } from "../core/params/hardhat-params";
 import type hardhat from "../lib/hardhat-lib";
+import { getCacheDir } from "../util/global-dir";
```
alcuadrado, line 7: Should we move it to the `node_modules` dir? It will also have some edge cases, like linked projects.

Maybe we can use the global cache, but instead of saving a single file we can save one per project. What about a hash of the absolute path to the project as filename?

packages/hardhat-core/src/internal/cli/autocomplete.ts
```diff
@@ -62,50 +95,226 @@ export async function complete({
...
     }
   }
 
   // if there's no task, we complete either tasks or params
-  if (task === undefined || hre.tasks[task] === undefined) {
-    const tasks = Object.values(hre.tasks)
+  if (task === undefined || tasks[task] === undefined) {
+    const taskNames = Object.values(tasks)
       .map((x) => x.name)
       .filter((x) => !x.includes(":"));
```
alcuadrado, line 116: maybe store if the task is internal or not and filter based on that?

fvictorio: Not sure I understood this. You mean changing this in core to have that information here?

packages/hardhat-core/src/internal/cli/autocomplete.ts
fvictorio: I think the name is stored twice. Not sure though.

We aren't using the type right now, but maybe we'll do in the future? So yeah, maybe it's a good idea to add it to this interface.

packages/hardhat-core/src/internal/cli/autocomplete.ts
```diff
@@ -62,50 +95,226 @@ export async function complete({
...
+    if (arePreviousMtimesCorrect(cachedCompletionData.mtimes)) {
+      return cachedCompletionData.completionData;
+    }
+  }
+
+  const filesBeforeRequire = Object.keys(require.cache);
+  let hre: typeof hardhat;
+  try {
+    process.env.TS_NODE_TRANSPILE_ONLY = "1";
+    require("../../register");
```
alcuadrado, line 157: I still don't understand why this is necessary. There's a double initialization going on here, which makes things slower.


What's the error you get if you do `hre = require("../..")`?

packages/hardhat-core/src/internal/cli/autocomplete.ts
alcuadrado: lol

packages/hardhat-core/src/internal/cli/autocomplete.ts
alcuadrado: The main reason I'd store them individually is that the cache can grow forever, making reading it slow.

packages/hardhat-core/src/internal/cli/autocomplete.ts
alcuadrado: I think the `includes(":")` is used to decide if a task name should be suggested or not, as a proxy to whether a task is internal or not. That info, being internal or not, is already present in the task definition object.

alcuadrado: For example, the help printer does something very similar.

fvictorio: Got it. Fixed in a508a9c76801ea391cc6ccb1012993ab23fe3698

packages/hardhat-core/src/internal/cli/autocomplete.ts
fvictorio:
```
SyntaxError: Unexpected token export
    at Module._compile (internal/modules/cjs/loader.js:723:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at importCsjOrEsModule (.../hardhat/packages/hardhat-core/internal/core/config/config-loading.js:19:22)
    at Object.loadConfigAndTasks (.../hardhat/packages/hardhat-core/internal/core/config/config-loading.js:48:22)
    at Object.<anonymous> (.../hardhat/packages/hardhat-core/internal/lib/hardhat-lib.js:30:37)
```

packages/hardhat-core/src/internal/cli/autocomplete.ts
fvictorio: Done in 7fbbf4782500c4f8480b3accbc0c3014dc24f4ef

PR #1039, opened 2020-11-19
solhint: add solhint package as a peer and dev dependency

### LINKED ISSUE #799: [buidler-solhint] solhint as peerDependency
use `solhint` as a `peerDependency`, allowing users to choose between incompatible versions, ex: 2.x vs 3.x
currently `buidler-solhint` uses `"solhint": "^2.0.0"` as a dependency.

shalzz:
fixes #799

### DIFF
--- a/packages/hardhat-solhint/README.md
+++ b/packages/hardhat-solhint/README.md
@@ -12,7 +12,7 @@ This plugin runs solhint on the project's sources and prints the report.
 ## Installation
 
 ```bash
-npm install --save-dev @nomiclabs/hardhat-solhint
+npm install --save-dev @nomiclabs/hardhat-solhint solhint
 ```
 
 And add the following statement to your `hardhat.config.js`:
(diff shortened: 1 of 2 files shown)


fvictorio:
This seems to work, but it's a product question: should installing the plugin "just work" or should anyone be able to select the version of solhint they want, at the cost of an extra step?

Btw, if we do this, the readme should be updated.

@alcuadrado @fzeoli what do you think?

alcuadrado:
Considering that npm7, which has already been released, automatically installs peer dependencies, I think it's fine.

fvictorio (approved):
Note for us: we'll have to use a new major when we release this.

alcuadrado:
Hey @shalzz, we just discussed this internally and decided not to merge it. Take a look at the issue you mentioned to understand the reasoning.

Thanks anyway for submitting this PR!

PR #1041, opened 2020-11-19
Fix getContractAt for unlinked contracts in hardhat-ethers.

### LINKED ISSUE #950: [hardhat-ethers] `getContractAt` doesn't work for unlinked contracts
Calling the `getContractAt` helper currently fails if the contract needs library linking.

```js
const token = await ethers.getContractAt("Token", address);
```

This can be worked around like this for now:
```js
const tokenArtifact = await artifacts.readArtifact("Token");
const token = new ethers.Contract(address, tokenArtifact.abi, ethers.provider);
```

scnale:
Fixes #950.

### DIFF
--- a/packages/hardhat-ethers/src/helpers.ts
+++ b/packages/hardhat-ethers/src/helpers.ts
@@ -235,7 +235,7 @@ ${missingLibraries}`
   return linkBytecode(artifact, [...linksToApply.values()]);
 }
 
-export async function getContractFactoryByAbiAndBytecode(
+async function getContractFactoryByAbiAndBytecode(
   hre: HardhatRuntimeEnvironment,
   abi: any[],
   bytecode: ethers.utils.BytesLike,
--- a/packages/hardhat-ethers/src/helpers.ts
+++ b/packages/hardhat-ethers/src/helpers.ts
@@ -107,9 +107,32 @@ async function getContractFactoryByName(
   contractName: string,
   signerOrOptions?: ethers.Signer | FactoryOptions
 ) {
-  const { utils } = require("ethers") as typeof ethers;
-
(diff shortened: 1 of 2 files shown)


packages/hardhat-ethers/src/helpers.ts
```diff
@@ -248,6 +244,7 @@ export async function getContractFactoryByAbiAndBytecode(
     abi
   );
 
+  const { ContractFactory } = require("ethers") as typeof ethers;
```
fvictorio, line 247: I think the point of dynamic requires is just to make Hardhat's startup faster. IMO moving this `require` this way is a micro-optimization that hurts readability and consistency. It's not _bad_, but it feels like an unnecessary change.

packages/hardhat-ethers/src/helpers.ts
```diff
@@ -257,10 +254,14 @@ export async function getContractAt(
...
-
   if (typeof nameOrAbi === "string") {
-    const factory = await getContractFactoryByName(hre, nameOrAbi, signer);
+    const artifact = await hre.artifacts.readArtifact(nameOrAbi);
+    const factory = await getContractFactoryByAbiAndBytecode(
+      hre,
+      artifact.abi,
+      "0x",
+      signer
+    );
```
fvictorio, line 264: This seems like the only change that this PR needs. What do the rest of the changes do? They seem to inline some code that was in the `getContractFactoryByName` function, so it doesn't really seem like an improvement in readability. Am I missing something here?

scnale: I just added the `getContractFactoryByName` function back and restructured this into several commits so it's clearer what it is doing.

PR #1059, opened 2020-11-26
Fix a family of slowdown issues with Hardhat providers in ethers.

### LINKED ISSUE #1037: [hardhat-ethers] Library linking hangs when a contract instance is used instead of an address
When running this test it just hangs when deploying the last contract

https://github.com/tellor-io/lens/blob/6b2e973e9adda48de80a7e7a6dee5401ff0be557/test/tests.js#L76

scnale:
Closes #1037.

This case was being handled already but I added a test for it since it is quite easy to make this mistake.

Update:
The test wasn't reproducing the issue correctly. There is no way to reproduce it in the plugin test environment that I know of.
#1037 is only one specific way of triggering this slowdown. With this fix, it should be hard to trigger it now.

### DIFF
--- a/packages/hardhat-ethers/test/index.ts
+++ b/packages/hardhat-ethers/test/index.ts
@@ -387,6 +387,41 @@ describe("Ethers plugin", function () {
             );
           });
 
+          it("should fail to create a contract factory when incorrectly linking a library with an ethers.Contract", async function () {
+            const libraryFactory = await this.env.ethers.getContractFactory(
+              "TestLibrary"
+            );
+            const library = await libraryFactory.deploy();
+
+            try {
+              const contractFactory = await this.env.ethers.getContractFactory(
+                "TestContractLib",
+                { libraries: { TestLibrary: library as any } }
+              );
+            } catch (reason) {
+              assert.instanceOf(
+                reason,
+                NomicLabsHardhatPluginError,
+                "getContractFactory should fail with a hardhat plugin error"
+              );
+              assert.isTrue(
+                reason.message.includes(
+                  "invalid address",
+                  "getContractFactory should report the invalid address as the cause"
+                )
+              );
+              assert.isTrue(
+                reason.message.length <= 400,
+                "getContractFactory should fail with an error message that isn't too large"
+              );
+              return;
+            }
+
+            assert.fail(
+              "getContractFactory should fail to create a contract factory if there is an invalid address"
+            );
+          });
+
           it("Should be able to send txs and make calls", async function () {
             const Greeter = await this.env.ethers.getContractFactory("Greeter");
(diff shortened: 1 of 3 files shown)


packages/hardhat-ethers/test/index.ts
```diff
@@ -387,6 +387,41 @@ describe("Ethers plugin", function () {
...
+                NomicLabsHardhatPluginError,
+                "getContractFactory should fail with a hardhat plugin error"
+              );
+              assert.isTrue(
+                reason.message.includes(
+                  "invalid address",
+                  "getContractFactory should report the invalid address as the cause"
+                )
+              );
+              assert.isTrue(
```
alcuadrado, line 413: I don't get this

scnale: I added this assert to make sure the resulting string doesn't contain a stringification of an enormous object. The error attempts to be friendly in common cases by printing the passed in argument.

alcuadrado: Can you add a comment explaining why it's there?

packages/hardhat-ethers/src/helpers.ts
```diff
@@ -148,7 +148,13 @@ async function collectLibrariesAndLink(
   for (const [linkedLibraryName, linkedLibraryAddress] of Object.entries(
     libraries
   )) {
-    if (!utils.isAddress(linkedLibraryAddress)) {
+    // ethers.utils.isAddress may end up being very slow if you pass a complex object to it.
+    // We avoid that by adding this type check.
+    // TODO: add link to issue tracking this.
```
fvictorio, line 153: No need for a link.

fvictorio:
We should rename this PR's title.

packages/hardhat-ethers/src/helpers.ts
fvictorio: Oh, you mean the issue in ethers? I think we won't remove this check even if it's fixed in ethers, so I don't think it's worth it. But I don't mind adding it.

scnale:
Yeah, I'm about to try another, more general, fix. I'll rename once I determine if it works or not.

scnale:
OK, I couldn't reproduce this with a test, but I did manage to reproduce it with a standalone project. ethers was stringifying a big Hardhat provider so I added a couple of `toJSON` methods to some of the ethers wrapper classes in this plugin. We might want to take a look at this in the core itself later though.

packages/hardhat-ethers/src/helpers.ts
scnale: Yeah, I don't think it's worth opening an issue in ethers about this, currently.

alcuadrado (approved):

PR #1062, opened 2020-11-27
Unpad values in FSM's contract storage

### LINKED ISSUE #990: Gas usage does not match that of mainnet
See the linked repository for reproduction steps.

The test here indicates that the gas usage for the `AirdropToken#airdrop` function uses about 5 million gas for the given inputs:
https://github.com/jusdefi/jusdefi/blob/2c32d238559773e9b5007da3981502812189909d/test/airdrop_token.js#L106

The same estimate is returned on both Hardhat Network and Ganache.

The script here would not run because the estimated gas requirement, for the same inputs:
https://github.com/jusdefi/jusdefi/blob/master/scripts/airdrop.js

Reducing the size of the inputs (`require('airdrop-ledger/airdrops.json').slice(0, 130)`) yields a gas estimate of about 1.5 million.  However, the mainnet transaction cost about 3.3 million: https://etherscan.io/tx/0x11f1950bf18553a7d39e13cd117b3958946507f3d622e7b0981e6ac36ca70a65

The same transaction on RInkeby also used about 3.3 million gas: https://rinkeby.etherscan.io/tx/0x6759621ab32fd7e0925f5e654a8b54e2c53022a9d4a8574a7227c8e5a3218da5

### LINKED ISSUE #1043: Different gas used in hardhat network with fork, without fork, and rinkeby
This is a simplified version of #990, but it's different enough to merit its own issue. Maybe fixing this (assuming there's something to be fixed) fixes that one too, but maybe not.

### Reproduction steps

```
git clone https://github.com/fvictorio/hardhat-gas-used-bug
cd hardhat-gas-used-bug
export FORK_URL="https://mainnet.infura.io/v3/<project-id>"
export RINKEBY_URL="https://rinkeby.infura.io/v3/<project-id>"
export RINKEBY_PRIVATE_KEY="<rinkeby-account-private-key>"

npx hardhat run script.js # 43317
FORK=true npx hardhat run script.js # 43086
npx hardhat run script.js --network rinkeby # 58317
```

The three runs give a different gas used. To the best of my knowledge, the contracts don't use any on-chain state, so that shouldn't be the cause. But maybe I'm missing something here.

fvictorio:
Closes #990.
Closes #1043.


packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
```diff
@@ -372,3 +386,12 @@ export class ForkStateManager implements PStateManager {
     this._state = state;
   }
 }
+
+function unpadBuffer(buffer: Buffer): Buffer {
```
fvictorio, line 390: The version of `ethereumjs-util` we are using doesn't seem to have this function.

Issue #1061, opened 2020-11-27
bug hardhat-waffle: changeEtherBalance not available


gitpusha:
I get an `Error: Invalid Chai propert: changeEtherBalance` 

```json
    "@nomiclabs/hardhat-ethers": "2.0.0",
    "@nomiclabs/hardhat-waffle": "2.0.0",
    "ethereum-waffle": "3.2.1",
    "ethers": "5.0.23",
    "chai": "4.2.0",
    "hardhat": "2.0.3",
```

dahifi:
add the following two lines to node_modules/@nomiclabs/hardhat-waffle/dist/src/waffle-chai.js
```
const { supportChangeEtherBalance, } = require(`${waffleChaiPath}/matchers/changeEtherBalance`);
supportChangeEtherBalance(chai.Assertion);
```

alcuadrado:
This has already been fixed. Thanks for workin on it, @dahifi

Issue #1020, opened 2020-11-16
hardhat-waffle: changeTokenBalance matcher not available


ItsNickBarry:
Attempting to use the Waffle chai matcher `changeTokenBalance` returns this error: 

```
Invalid Chai property: changeTokenBalance
```

For example:

```javascript
const { expect } = require('chai');

await expect(
  () => instance['mint(address,uint256)'](receiver.address, amount)
).to.changeTokenBalance(
  instance, receiver.address, amount
);
```

`@nomiclabs/hardhat-waffle` 2.0.0
`ethereum-waffle` 3.2.0

dahifi:
Having same issue with `changeEtherBalance`. Same version of waffle.

dahifi:
For those wondering how to get ether balance, there are two ways to do it:

For contracts, you'll have to use `ethers.provider.getBalance(contract.address));`. Addresses that are signers objects from `ether.getSigners()` can use the `getBalance()` method, as in:

```
const {wallet1, wallet2} = await ethers.getSigners();
console.log("Wallet1 balance: ", await wallet1.getBalance());
```

Until this bug is fixed, you'll have to query the wallet balance before the changing call, then again after, then compare the two. 

```
const before = await ethers.provider.getBalance(auction.address);
console.log("Before: ", before.toHexString());
await expect(auction.deposit_pot({value: 1}));
const after = await ethers.provider.getBalance(auction.address);
console.log("After: ", after.toHexString());
console.log("Difference: ", before.sub(after).toHexString());
>Before:  0x00
>After:  0x01
>Difference:  -0x01
```
Note that the balances changes on wallets will also include gas. 

I really hope there's a resolution on this soon.

Skyge:
Have a same error: `Error: Invalid Chai property: changeTokenBalance`
```js
const { expect } = require("chai");
const { ethers, waffle } = require("hardhat");
const { deployContract } = waffle;

const Token = require("../artifacts/contracts/mock/Token.sol/Token.json");

describe('DLToken', () => {
    let token;
    let user1, user2;

    before(async () => {
        const [owner, ...accounts] = await ethers.getSigners();
        user1 = accounts[0];
        user2 = accounts[1];
        token = await deployContract(owner, Token, ["Mock Token", "MT"]);
    });

    it('mint', async () => {
        expect(await token.balanceOf(user1.address)).to.equal(0);
[...]

PR #1064, opened 2020-11-28
Bugfix for #1020 and #1061

dahifi:
Looks like these functions weren't explicitly imported. I went ahead and added the four functions for ether and token balance changes. I would like to add some unit tests, but I don't know enough metaprogramming to check the matchers without actually writing a test against the smart contract.

### DIFF
--- a/packages/hardhat-waffle/test/fixture-projects/hardhat-project/contracts/Token.sol
+++ b/packages/hardhat-waffle/test/fixture-projects/hardhat-project/contracts/Token.sol
@@ -1,0 +1,92 @@
+// Example taken from https://gist.github.com/jonchurch/c71de291b3ffa49ae17e416393f87135 and modified
+
+pragma solidity ^0.7.0;
+
+contract Token {
+  string public symbol = "FIXED";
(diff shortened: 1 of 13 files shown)


dahifi:
Anything else needed from me on this? I got a code review notification, assume no action needed on my part?

packages/hardhat-waffle/test/fixture-projects/hardhat-project/contracts/Token.sol
```diff
@@ -0,0 +1,92 @@
+// Example taken from https://gist.github.com/jonchurch/c71de291b3ffa49ae17e416393f87135 and modified
```
alcuadrado, line 1: That code is MIT, so this is fine.
```suggestion
// Example taken from https://gist.github.com/jonchurch/c71de291b3ffa49ae17e416393f87135 and modified
// Original code's license: BokkyPooBah 2017. The MIT Licence.
```

alcuadrado (approved):
Thanks a lot for this PR, @dahifi

PR #1069, opened 2020-11-29
Hardhat ethers address signer

wighawag:
- Allow to pass address as string instead of signer for the various contract/factory functions

- For `getContractAt` it also fallback to a read-only provider if no accounts[0] is available


fvictorio:
Thanks @wighawag, we'll take a look at this as soon as we can.

Note for ourselves: we need to update the docs with these changes.

fvictorio:
Closing this in favor of `@nomiclabs/hardhat-ethers@2.0.2-rc.1`

PR #1072, opened 2020-11-30
Add fully qualified name parameter to hardhat-etherscan

scnale:
This adds an optional parameter to the `verify` task of hardhat-etherscan that allows the selection of a particular contract when verifying. In practice, this is useful to disambiguate cases where there is more than one contract in the project that matches with the retrieved bytecode.


fvictorio (approved):

packages/hardhat-etherscan/src/network/prober.ts
```diff
@@ -24,30 +24,33 @@ const networkIDtoEndpoint: NetworkMap = {
...
+): Promise<string> {
   const chainID = parseInt(await provider.send("eth_chainId"), 16) as NetworkID;
 
   const endpoint = networkIDtoEndpoint[chainID];
-  if (endpoint !== null && endpoint !== undefined) {
-    // Beware: this delays URL validation until it is effectively "used".
-    // Tests should take this into account.
-    return new URL(endpoint);
+
+  if (endpoint === null || endpoint === undefined) {
```
alcuadrado, line 35:
```suggestion
  if (endpoint === undefined) {
```

this can't be `null`

packages/hardhat-etherscan/src/solc/metadata.ts
```diff
@@ -1,55 +1,132 @@
...
-    super(pluginName, message, undefined, true);
-    Object.setPrototypeOf(this, MetadataAbsentError.prototype);
-  }
-}
+export const METADATA_LENGTH_SIZE = 2;
+export const METADATA_PRESENT_SOLC_NOT_FOUND_VERSION_RANGE = "0.4.7 - 0.5.8";
+export const METADATA_ABSENT_VERSION_RANGE = "<0.4.7";
 
-export async function readSolcVersion(bytecode: Buffer): Promise<string> {
+const log = debug("hardhat-etherscan:metadata-decoder");
```
alcuadrado, line 16: I think we use `hardhat:plugin:plugin-name:module` in the rest of the plugins.

alcuadrado (approved):

PR #1073, opened 2020-11-30
Handle overriden tasks in autocomplete

fvictorio:
This is not the correct solution, but I'm opening this PR so that it's easier to discuss what's the best approach here. See inline comment.

### DIFF
--- a/packages/hardhat-core/src/internal/cli/autocomplete.ts
+++ b/packages/hardhat-core/src/internal/cli/autocomplete.ts
@@ -134,6 +134,7 @@ export async function complete({
 }
 
 async function getCompletionData(): Promise<CompletionData | undefined> {
+  const { default: _ } = await import("lodash");
   const projectId = getProjectId();
 
   if (projectId === undefined) {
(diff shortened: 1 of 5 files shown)


packages/hardhat-core/src/internal/core/tasks/task-definitions.ts
```diff
@@ -560,6 +560,14 @@ export class OverriddenTaskDefinition implements TaskDefinition {
     this.parentTaskDefinition = parentTaskDefinition;
   }
 
+  public toJSON() {
+    return {
+      name: this.name,
+      isSubtask: this.isSubtask,
+      paramDefinitions: this.paramDefinitions,
+    };
+  }
```
fvictorio, line 569: The root of the problem is that, for an `OverridenTaskDefinition`, the `name` field exists but, since it's a getter, it's not serialized.

An obvious solution is to remove those getters (and be overly cautious with getters in the future).

A more type-safe (but complex) solution is to add a `SerializableTaskDefinition` interface, extend that from `TaskDefinition` and use it in the `autocomplete` module, and then return that type here in the `toJSON` method.

alcuadrado (approved):

PR #1083, opened 2020-12-04
Use configured chainId in forked network

fvictorio:


packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkTransaction.ts
```diff
@@ -0,0 +1,71 @@
...
+  public verifySignature(): boolean {
+    return true;
+  }
+
+  public getChainId(): number {
+    return this._chainId;
+  }
+
+  public sign() {
+    throw new Error("Not implemented on ForkTransaction");
```
fvictorio, line 37: This is temporary. Not sure which Hardhat Error to use here.

alcuadrado: I'd throw a ProviderError insteead of a hardhat error.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
```diff
@@ -277,10 +298,17 @@ export class ForkBlockchain implements PBlockchain {
     ) {
       return undefined;
     }
-    const transaction = new Transaction(rpcToTxData(rpcTransaction), {
+
+    const chainId = this._common.chainId();
+    const txData = rpcToTxData(rpcTransaction);
+    const senderPubKey = new Transaction(txData).getSenderPublicKey();
```
alcuadrado, line 304: I don't think this would work. If the tx also uses eip155 this will fail, as it's default to a mainnet common.

alcuadrado: We should get the sender pub key from txData and the chainId.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
```diff
@@ -277,10 +298,17 @@ export class ForkBlockchain implements PBlockchain {
...
+
+    const transaction = new ForkTransaction(chainId, senderPubKey, txData, {
```
alcuadrado, line 306: This `chainId` is not the right one. Here we need the remote network's chain id, but we are using the local one here.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
alcuadrado: It should look something like this:

```ts
const msgHash = this.hash(false);
const v = bufferToInt(this.v);

return ecrecover(
  msgHash,
  v,
  this.r,
  this.s,
  this. _implementsEIP155() ? this.getChainId() : undefined,
);
```

alcuadrado (approved):

PR #1094, opened 2020-12-10
Use forked tabtab and add native fs completion

fvictorio:
This fixes the issue with tasks with a colon in their name, and adds support for native autocomplete. It uses a forked version of (the forked version of) tabtab until this https://github.com/pnpm/tabtab/pull/5 is merged and published.

**Note**: if you test this in zsh, it will work fine. If you then open a bash shell and try it, the colons will be wrongly escaped. This happens because, when you open bash from zsh, the `SHELL` variable is not modified, and `tabtab` uses that to detect the current shell. This makes it think you are still in zsh and does the wrong thing.

The bottom line is, to test this in bash from zsh:

```
$ bash
$ SHELL=/usr/bin/bash
$ hh <tab>
```

and in fish:

```
$ fish
$ set SHELL /usr/bin/fish
$ hh <tab>
```


alcuadrado (approved):

packages/hardhat-launcher/src/completion.ts
```diff
@@ -0,0 +1,139 @@
...
+  console.error(
+    `Unrecognized command "${cmd}". You can install Hardhat completion with the "install" command.`
+  );
+  process.exit(1);
+}
+
+function logWarningWithThrottling(message: string) {
+  // we need to write a file to disk to do the throttling because zsh calls the
+  // autocomplete function several times, and these are separate processes
+  const throttleFile = path.join(__dirname, ".hh-throttle-file");
```
alcuadrado, line 107: This is saving this into a `node_modules`. I think it can lead to problems.

alcuadrado: maybe `process.cwd()` instead of `__dirname`? This errors shouldn't be common.

fvictorio: I think using `process.cwd()` will mean that this file can be created in (multiple) places of the user's repository, which seems annoying (I would find it annoying at least, even if I can .gitignore it).

What problems can we run into if we use `node_modules`?

alcuadrado: It wouldn't work with yarn2, it wouldn't work for the people that incorrectly use sudo to install global packages, and there's probably more issues.

I wouldn't assume that there's a global node modules nor that it's writable.

alcuadrado: We can do something similar to hardhat's global dirs

alcuadrado:
Should we merge this, @fvictorio ?

fvictorio:
Yes, I think we can.

PR #1100, opened 2020-12-13
Fix eth_call and HarhatProvider's logging initialization

alcuadrado:
This PR fixes:

1. How HardhatProvider's logging is initialized. `console.log`s weren't going through the logging infra if `loggingEnabled` was true in the `hardhat` network.
2. Fix how `eth_call` handled `blockTag`, actually running the calls in the block's context.
3. Let `eth_call` use gas limit's higher than the block's one. This is implemented with a temporal workaround which has been documented in the code.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -21,6 +21,7 @@ import EventEmitter from "events";
 
 import { CompilerInput, CompilerOutput } from "../../../types";
 import { HARDHAT_NETWORK_DEFAULT_GAS_PRICE } from "../../core/config/default-config";
+import { assertHardhatInvariant } from "../../core/errors";
 import { Reporter } from "../../sentry/reporter";
 import { getDifferenceInSeconds } from "../../util/date";
 import { createModelsAndDecodeBytecodes } from "../stack-traces/compiler-to-model";
@@ -71,7 +72,6 @@ import { makeForkClient } from "./utils/makeForkClient";
 import { makeForkCommon } from "./utils/makeForkCommon";
 import { makeStateTrie } from "./utils/makeStateTrie";
 import { putGenesisBlock } from "./utils/putGenesisBlock";
-import { assertHardhatInvariant } from "../../core/errors";
 
 const log = debug("hardhat:core:hardhat-network:node");
 
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -1351,10 +1351,6 @@ If you are using a wallet or dapp, try resetting your wallet's accounts.`
   /**
    * This function runs a transaction and reverts all the modifications that it
    * makes.
-   *
-   * If throwOnError is true, errors are managed locally and thrown on
-   * failure. If it's false, the tx's RunTxResult is returned, and the vmTracer
-   * inspected/reset.
    */
   private async _runTxAndRevertMutations(
     tx: Transaction,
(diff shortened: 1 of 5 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/types/Block.ts
```diff
@@ -10,11 +10,12 @@ export const Block: BlockConstructor = EthBlock;
 
 // tslint:disable-next-line:no-misused-new
 type BlockConstructor = new (
-  data?: BlockData | null,
+  data?: BlockData | null | [Buffer[], Buffer[], Buffer[]],
   chainOptions?: { common: Common }
 ) => Block;
 
 export interface Block {
+  readonly raw: [Buffer[], Buffer[], Buffer[]];
```
fvictorio, line 18: I don't understand the changes in this file.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1358,14 +1358,18 @@ If you are using a wallet or dapp, try resetting your wallet's accounts.`
    */
```
fvictorio, line 1358: Unrelated to this PR, but is this comment up to date?

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1383,9 +1387,29 @@ If you are using a wallet or dapp, try resetting your wallet's accounts.`
...
+
+          if (txGasLimit.gt(blockGasLimit)) {
+            blockContext.header.gasLimit = tx.gasLimit;
+          }
+        }
       } else {
-        // if the context is to run calls with the latest block
-        blockContext = await this.getLatestBlock();
+        // if the context is to run calls with a block
+        // We know that this block number exists.
```
fvictorio, line 1401: How do we know that the block number exists?

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1383,9 +1387,29 @@ If you are using a wallet or dapp, try resetting your wallet's accounts.`
...
+        blockContext = (await this.getBlockByNumber(blockNumber))!;
+
+        if (workaroundEthCallGasLimitIssue) {
```
fvictorio, line 1404: Wouldn't it be better to have this extracted after the if? The `previousGasLimit` would be set unnecessarily when the latest block is being used, but seems like a good trade-off to me.

fvictorio: (I'm actually on the fence on this anyway, so I'm also ok with leaving it as it is)

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: Good catch. Deleted it.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: I'll add a better explanation.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: Oh, good point! I didn't realize that, as I initially only implemented it in the `else`.

packages/hardhat-core/src/internal/hardhat-network/provider/types/Block.ts
alcuadrado: I copy-pasted the missing bits from the upstream definition of Block. I defined it upstream IIRC.

alcuadrado: Oh, this isn't used in this PR anymore!

alcuadrado: I reverted the change.

fvictorio (approved):

PR #1106, opened 2020-12-15
Reduce CI false negatives

fvictorio:


packages/hardhat-core/src/internal/hardhat-network/jsonrpc/client.ts
```diff
@@ -322,11 +322,16 @@ export class JsonRpcClient {
   }
 
   private _shouldRetry(isRetryCall: boolean, err: any) {
+    const isRetriableError =
+      err.message.includes("header not found") ||
+      err.message.includes("connect ETIMEDOUT");
+
+    const isServiceUrl =
+      this._httpProvider.url.includes("infura") ||
+      this._httpProvider.url.includes("alchemyapi");
```
fvictorio, line 331: @alcuadrado I don't know why only `infura` was being considered here. I added alchemy just to try it out, but let me know if this should be reverted.

alcuadrado: lgtm

PR #1109, opened 2020-12-16
Fix Vyper cached compilation

Dylan-Kerler:
Fixes: https://github.com/nomiclabs/hardhat/issues/1107

Fixed the path lookup and the ctime comparison between artifact and source files.

### DIFF
--- a/packages/hardhat-vyper/src/compilation.ts
+++ b/packages/hardhat-vyper/src/compilation.ts
@@ -52,11 +52,11 @@ export async function compile(
     const pathFromSources = path.relative(paths.sources, file);
 
     if (await isAlreadyCompiled(file, paths, vyperVersion, files)) {
-      console.log(pathFromCWD, "is already compiled");
+      console.log("✓", pathFromCWD, "is already compiled");
       continue;
     }
 
-    console.log("Compiling", pathFromCWD);
+    console.log("⌛ Compiling", pathFromCWD);
 
     const processResult = await handleCommonErrors(
       compileWithDocker(file, docker, dockerImage, paths)
--- a/packages/hardhat-vyper/src/compilation.ts
+++ b/packages/hardhat-vyper/src/compilation.ts
@@ -100,7 +100,9 @@ async function isAlreadyCompiled(
   }
 
   const contractName = pathToContractName(sourceFile);
-  const artifactPath = path.join(paths.artifacts, `${contractName}.json`);
+  const contractPath = `/contracts/${sourceFile.split("/contracts")[1]}`;
+  const artifactPath = path.join(paths.artifacts, `${contractPath}/${contractName}.json`);
+
   if (!(await fsExtra.pathExists(artifactPath))) {
     return false;
   }
--- a/packages/hardhat-vyper/src/compilation.ts
+++ b/packages/hardhat-vyper/src/compilation.ts
@@ -100,18 +100,20 @@ async function isAlreadyCompiled(
   }
 
   const contractName = pathToContractName(sourceFile);
-  const contractPath = `/contracts/${sourceFile.split("/contracts")[1]}`;
+  const contractPath = `/contracts${sourceFile.split("/contracts")[1]}`;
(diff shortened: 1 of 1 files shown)


packages/hardhat-vyper/src/compilation.ts
```diff
@@ -52,11 +52,11 @@ export async function compile(
     const pathFromSources = path.relative(paths.sources, file);
 
     if (await isAlreadyCompiled(file, paths, vyperVersion, files)) {
-      console.log(pathFromCWD, "is already compiled");
+      console.log("✓", pathFromCWD, "is already compiled");
       continue;
     }
 
-    console.log("Compiling", pathFromCWD);
+    console.log("⌛ Compiling", pathFromCWD);
```
fvictorio, line 59: I would avoid doing this, since some terminals don't handle emojis well.

packages/hardhat-vyper/src/compilation.ts
```diff
@@ -100,17 +100,21 @@ async function isAlreadyCompiled(
...
+  const artifactPath = path.join(paths.artifacts, `${contractPath}/${contractName}.json`);
+
   if (!(await fsExtra.pathExists(artifactPath))) {
     return false;
   }
+  
+  const sourceFilePath = sources.find(s => s.includes(contractPath));
+  if (!sourceFilePath) {
+    return false;
+  }
```
fvictorio, line 117: I don't understand why this is necessary, and why not just do

```
const lastSourcesCtime = (await fsExtra.stat(sourceFile)).ctimeMs;
```

below.

fvictorio:
@Dylan-Kerler I fixed some linter errors and changed the way the `sourceName` is obtained (the way it was done wasn't taking into account Windows, or users that change the `sources` path). I also left a couple of comments.

packages/hardhat-vyper/src/compilation.ts
Dylan-Kerler: Ok yes this is better. I was just modified the old implementation which mapped over the sources array.

packages/hardhat-vyper/src/compilation.ts
Dylan-Kerler: It's hard to distinguish between what is compiling and what already has compiled without the emoji - It's a huge wall of white text. I made the change based on the way mocha displays green ticks next to passed tests.

fvictorio: I understand, but the problem is that it looks like this in a terminal that doesn't support emojis:
image

Maybe making the "Nothing to compile" output greent would help? (`chalk` handles correctly terminals without color support, and in any case that's less common that not having emoji support)

@alcuadrado what do you think?

alcuadrado:
I think this is a great opportunity to make this plugin also work with solidity. The required change is really small. If you are interred on doing it I'll provide some guidance tomorrow.

Dylan-Kerler:
In the future we'll def need that functionality too as there are solidity contracts that we would like to interface with, however I am quite busy working on our current contract's features. So I won't be able to take a look until a couple of weeks.

packages/hardhat-vyper/src/compilation.ts
alcuadrado: We have a function, `emoji`, that can be used for this. Here's an example: https://github.com/nomiclabs/hardhat/blob/aaa9cfe6a590ef578432c0d66e7dc58517ef65bb/packages/hardhat-core/src/internal/cli/project-creation.ts#L84

fvictorio:
Thanks again @Dylan-Kerler and sorry for taking so long to revisit this.

We can't merge this yet since #1249 caused this PR to stop working.

The reason is that the Solidity compilation task removes the artifacts that weren't generated by the current `.sol` files. The fix seems simple: ignore the artifact directories that are not `.sol`. But that means a change in core and that will have to wait a little. I'll try to make that extra change as soon as possible.

fvictorio:
Blocked by #1335.

spape:
+1

atvanguard:
Bumping this up folks. Having vyper files compiled every time eats up a lot of dev time. Thank you.

PR #1114, opened 2020-12-17
Add libraries parameter to hardhat-etherscan

### LINKED ISSUE #945: [hardhat-etherscan] Programatically pass complex arguments to verify task
How to verify if my contract constructor has array params?

Let say i have a contract that have constructor accepting 2 params, string and array

````
constructor(string memory param1, address[] memory param2) public {
}
````

I want to verify it using etherscan verify plugin. 

Tried code below it doesn't work.
````
const bre = require("@nomiclabs/buidler");
await bre.run('verify', {
        address: '0xXXXXXX,
        constructorArguments: ['Some params', ['test']],
    });
````
This will result in error `Invalid value  for argument constructorArguments of type string`.

### PR DESCRIPTION
**Pull request by scnale**
Fixes #945.

This adds:
- An option to pass a `libraries` parameter to the hardhat-etherscan task, thus adding support for libraries that are only used in the deploy bytecode of a contract.
- A subtask with a friendly API for users that want to verify a contract within a script or task.

### CODE REVIEW

File: packages/hardhat-etherscan/src/solc/bytecode.ts [hunk 1]
```diff
@@ -39,6 +40,7 @@ export interface ContractInformation extends BytecodeExtractedData {
   sourceName: SourceName;
   contractName: ContractName;
   contract: CompilerOutput["contracts"][SourceName][ContractName];
+  undetectableLibraries?: LibraryNames;
```
fvictorio, line 43: This, and the use of `ContractInformation` in some places and `Required<ContractInformation>` in others, seems like a code smell to me. I don't know how to improve it though, probably some re-design is needed.

packages/hardhat-etherscan/src/index.ts
```diff
@@ -124,6 +131,10 @@ const verify: ActionType<VerificationArgs> = async (
     }
   );
 
+  const libraries: Libraries = await run(TASK_VERIFY_GET_LIBRARIES, {
```
fvictorio, line 134: This isn't used anywhere. Plus, it has a `Libraries` type, but the task can return undefined.

fvictorio (changes requested):
This doesn't seem to work. I created a contract that uses a library only in the constructor and tried to verify it, and I correctly got the error message. But I then created the `libraries.js` module and passed it as a parameter, but got the exact same error again.

fvictorio:
I paused the review since I don't think it makes sense to review it if it's indeed not working. Will continue it later.

packages/hardhat-etherscan/src/index.ts
fvictorio: Maybe this is the reason it's not working.

scnale: Indeed. I missed this completely. I was about to test it manually since it's a bit impractical to write a test just for this.

packages/hardhat-etherscan/src/index.ts
```diff
@@ -93,6 +98,7 @@ interface GetContractInformationArgs {
   contractFQN: string;
   deployedBytecode: Bytecode;
   matchingCompilerVersions: string[];
+  libraries?: Libraries;
```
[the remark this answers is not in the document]
scnale: Hm, makes sense.

packages/hardhat-etherscan/src/index.ts
```diff
@@ -272,12 +288,20 @@ Possible causes are:
...
   // - "Unable to locate ContractCode at <address>"
-  // - Address of library used in constructor is wrong
-  throw new NomicLabsHardhatPluginError(
-    pluginName,
-    `The contract verification failed.
-Reason: ${verificationStatus.message}`
-  );
+  let errorMessage = `The contract verification failed.
+Reason: ${verificationStatus.message}`;
+  if (contractInformation.undetectableLibraries.length > 0) {
```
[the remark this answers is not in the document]
scnale: Ah, yes; this is necessary to add it as a possible cause. Basically, if your contract uses libraries whose addresses are undetectable, we have no way of knowing whether the addresses provided by the user are okay until Etherscan verifies the contract or not. If the contract is verified, they are correct. But if the verification fails, the API itself isn't very informative about the cause of the failure. That's why I think it is necessary to include this reason in the error message conditionally. There is no corresponding check in `getLibraryLinks` for this since we have to assume these kinds of links correct.

packages/hardhat-etherscan/src/index.ts
```diff
@@ -329,6 +353,43 @@ Reason: ${error.message}`,
...
+      if (typeof librariesModule !== "string") {
+        return;
+      }
+
+      const librariesModulePath = path.resolve(process.cwd(), librariesModule);
+
+      try {
+        const libraries = (await import(librariesModulePath)).default;
+
+        if (typeof libraries !== "object") {
```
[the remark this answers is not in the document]
scnale: I had forgotten that arrays would be `"object"` :sweat_smile:

packages/hardhat-etherscan/src/index.ts
```diff
@@ -515,9 +576,10 @@ subtask(TASK_VERIFY_GET_CONTRACT_INFORMATION)
         contractFQN,
         deployedBytecode,
         matchingCompilerVersions,
+        libraries,
```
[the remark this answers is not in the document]
scnale: Indeed.

packages/hardhat-etherscan/src/solc/bytecode.ts
scnale: Indeed. The `ContractInformation` type is a bit ad hoc. However, since it is not obvious how to address it, we might want to write it down in a comment or open up an issue here in github. I think opening an issue is overkill for something like this though.

PR #1118, opened 2020-12-20
Hardhat network configuration should not contain url property

### LINKED ISSUE #1051: Hardhat network can contain an url field, confusing the provider initialization
If an user adds an `url` field to the Hardhat Network config, we don't detect it during the config validation, but rather leave it there.

Then, when we initialize the network provider, Hardhat gets confused and treats it as an HTTP network config.

This should help reproduce it:

```js
networks: {
    hardhat: {
      url: "valid node url",
    },
  },
```

sisco0:
Closes #1051.

This PR modified the _configuration validator_ code in order to disallow `url` property of the `hardhat` object. This issue was caused because the modified conditional clause was testing `url` against `undefined`, which is a falsy value, but the property could be present after all. In order not to tackle with validations against this property in a subsequent step of the execution (which indeed is the **purpose** of the _configuration validator_) this conditional clause now uses `.hasOwnProperty` instead of comparing the value with `undefined`, where this property could be present with an undefined value.

Test cases have been added as well for the case of `hardhat` object containing an `url` property with value equal or not equal to `undefined`.

I am happy to bring my efforts to this project by this first contribution and want to thank @alcuadrado for his accompaniment through this task accomplishment.

### DIFF
--- a/packages/hardhat-core/src/internal/core/config/config-validation.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-validation.ts
@@ -224,8 +224,8 @@ export function getValidationErrors(config: any): string[] {
   // These can't be validated with io-ts
   if (config !== undefined && typeof config.networks === "object") {
     const hardhatNetwork = config.networks[HARDHAT_NETWORK_NAME];
-    if (hardhatNetwork !== undefined) {
-      if (hardhatNetwork.hasOwnProperty("url")) {
+    if (hardhatNetwork !== undefined && typeof hardhatNetwork === "object") {
+      if ("url" in hardhatNetwork) {
         errors.push(
           `HardhatConfig.networks.${HARDHAT_NETWORK_NAME} can't have an url`
         );
--- a/packages/hardhat-core/src/internal/core/config/config-validation.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-validation.ts
@@ -225,7 +225,7 @@ export function getValidationErrors(config: any): string[] {
   if (config !== undefined && typeof config.networks === "object") {
     const hardhatNetwork = config.networks[HARDHAT_NETWORK_NAME];
     if (hardhatNetwork !== undefined) {
-      if (hardhatNetwork.url !== undefined) {
+      if (hardhatNetwork.hasOwnProperty("url")) {
         errors.push(
           `HardhatConfig.networks.${HARDHAT_NETWORK_NAME} can't have an url`
         );
--- a/packages/hardhat-core/test/internal/core/config/config-validation.ts
+++ b/packages/hardhat-core/test/internal/core/config/config-validation.ts
@@ -689,6 +689,20 @@ describe("Config validation", function () {
             assert.isEmpty(errors);
           });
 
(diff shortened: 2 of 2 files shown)


packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -225,7 +225,7 @@ export function getValidationErrors(config: any): string[] {
   if (config !== undefined && typeof config.networks === "object") {
     const hardhatNetwork = config.networks[HARDHAT_NETWORK_NAME];
     if (hardhatNetwork !== undefined) {
-      if (hardhatNetwork.url !== undefined) {
+      if (hardhatNetwork.hasOwnProperty("url")) {
```
fvictorio, line 228: We usually use `"url" in hardhatNetwork` for this. I know there's a subtle difference between the two things in javascript, but I would do it that way just for consistency.

sisco0: I understand @fvictorio , from what you specify it should be valid to have a `url` property under the `hardhat` network object with value equal to `undefined`. Would it be correct?

fvictorio: Hmm, no, it wouldn't. But (AFAIK) the only difference between the idiom I suggested and the one you used is that `hasOwnProperty` doesn't check the prototype chain, but that doesn't really matter in this scenario, that's why I think consistency is more important here.

sisco0: What any other clause could be used instead to check if `url` property is present or not? (Even if it is set to a value or as undefined).

fvictorio: I think I'm not explaining myself correctly :sweat_smile: All I'm saying is that I'd rather do

```js
if ("url" in hardhatNetwork) {
```

instead of

```js
if (hardhatNetwork.hasOwnProperty("url")) {
```

They are very similar, and in many cases the latter one is the one you want. Since in this case they are pretty much the same, I'd use the first one since it's the one we use in the rest of the codebase.

sisco0: I understand now, thank you for your efforts on this PR review as I want to be an active contributor of this repository. I would do the changes adding a new commit to this PR and would ensure that new tests continue passing.

In this way, `url` parameter could not be a property under `hardhat` object.

sisco0:
I uploaded the changes requested in order to comply with coding standards for this project. Even more, a validation conditional clause has been added as well in order to ensure validity of the `hardhat` network object.

packages/hardhat-core/src/internal/core/config/config-validation.ts
alcuadrado: My 2cents to avoid future confusion:

In this case we do want to use `in`. If we were to use `hasOwnProperty` to validate the absence of a property, all the consumers of the config should be aware of this, and use the config's own properties instead of just doing `config.value`.

alcuadrado:
Thanks for submitting this PR, @sisco0. It looks good now.

PR #1119, opened 2020-12-20
Hint the file name

howardpen9:


fvictorio:
Hi @howardpen9, thanks for submitting this PR. I don't think this is necessary though, since the paragraph above the snippet already mentions this file. Plus, we don't do this in any other snippet, so I think it would be inconsistent with the rest of the docs and maybe a little noisy.

If you have any other feedback on the docs, please feel free to open an issue or letting us know in our discord server.

PR #1122, opened 2020-12-21
Mine several blocks in a single call

### LINKED ISSUE #1112: Mine several blocks in a single call

mul1sh:
I added a new method to allow mining of multiple blocks in a single call.

Fixes #1112

### DIFF
--- a/docs/hardhat-network/README.md
+++ b/docs/hardhat-network/README.md
@@ -268,6 +268,7 @@ To customise it, take a look at [the configuration section](/config/README.md#ha
 
 - `evm_increaseTime` – same as Ganache.
 - `evm_mine` – same as Ganache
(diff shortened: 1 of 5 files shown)


mul1sh:
@bartjman well it will have to mine all those blocks, so the speed is dependent on how fast the hardhat network mines an empty block, which is usually like 1 second from my tests so far

bartjman:
I see, that doesn't really solve our problem. What would happen is you simply jumped 100.000 block numbers?

mul1sh:
@bartjman ahh ok let me try that🙂

mul1sh:
Closing for now

toshiSat:
+1

alcuadrado:
This is close to being released, @toshiSat. See https://github.com/nomiclabs/hardhat/pull/2032

PR #1123, opened 2020-12-21
Add a types module for hardhat-ethers

scnale:
This pull request:
- Adds a `types` module that defines the interface for additional Hardhat fields added to the `ethers` object.
- Separates out compilation projects for the plugin and its tests while maintaining backward compatibility in module resolution.


fvictorio (approved):

alcuadrado:
This PR looks good, but I fixed a few things:

1. The `clean` npm script of `hardhat-ethers` was kinda fragile, as it only deleted some files. I used a wildcard instead.
2. The top-level `package.json`'s scripts hadn't been adapted.
3. Renamed `HardhatEthers` to `HardhatEthersHelpers`, as that expresses its intention. The older name clashed with the plugin's name.
4. Changed the way the type-extension was implemented. There was an implicit assumption that the type fo the module `ethers` matched with the one of its `ethers` property. That's not necessary true, nor I see the reason for doing it like that.
5. Renamed `signer-with-address.ts` to `signers.ts`. The reason for this is that this file name is now part of the plugin's API, so it should have a practical and future-proof name.

PR #1126, opened 2020-12-22
Add error message for one case in hardhat-etherscan.

scnale:
The Etherscan backend may not see the deployed bytecode if it was deployed recently. This adds an error message that better describes this issue and recommends waiting for five confirmations at least.

Apart from that, this pull request improves some messages and prints the URL to the verified contract in the console when the verification is successful.

### DIFF
--- a/packages/hardhat-etherscan/src/etherscan/EtherscanService.ts
+++ b/packages/hardhat-etherscan/src/etherscan/EtherscanService.ts
@@ -1,4 +1,5 @@
 import { NomicLabsHardhatPluginError } from "hardhat/plugins";
+import type { Response } from "node-fetch";
 
 import { pluginName } from "../constants";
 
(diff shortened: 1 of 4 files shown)


packages/hardhat-etherscan/src/etherscan/EtherscanService.ts
```diff
@@ -54,6 +38,40 @@ Reason: ${error.message}`,
       error
     );
   }
+
+  if (!response.ok) {
```
alcuadrado, line 42: I don't think this passes our linter rules. Note that the CI wasn't run in this PR because of its base branch, but otherwise it would have failed.

alcuadrado (approved):
This looks good, but the CI wasn't run yet. We should first merge its base branch.

packages/hardhat-etherscan/test/integration/PluginTests.ts
```diff
@@ -6,6 +6,8 @@ import {
 import type { task as taskT } from "hardhat/config";
 import { NomicLabsHardhatPluginError } from "hardhat/plugins";
 import type { CompilerInput } from "hardhat/types";
+// tslint:disable-next-line: no-implicit-dependencies
```
scnale, line 9:
```suggestion
```

PR #1127, opened 2020-12-23
Please review if this Closes #1075

### LINKED ISSUE #1075: config.networks.hardhat.loggingEnabled = false is ignored by the node task

sisco0:
Closes #1075 

Please kindly review if this task conditional clause for logging has been fixed properly.
I am new to the tasks and I have seen this detail, where a hardcoded `true` value was set for logging at this task.
How could be tested if this fixes the issue? Thank you very much, I would be happy to change the code as needed by your comments in the review.

### DIFF
--- a/packages/hardhat-core/src/builtin-tasks/node.ts
+++ b/packages/hardhat-core/src/builtin-tasks/node.ts
@@ -118,10 +118,10 @@ subtask(TASK_NODE_GET_PROVIDER)
         );
       }
 
-      // enable logging
+      // enable logging conditionally
       await provider.request({
         method: "hardhat_setLoggingEnabled",
-        params: [true],
+        params: [hardhatNetworkConfig.loggingEnabled ? true : false],
       });
 
       return provider;


alcuadrado:
Hey @sisco0, thanks for creating this PR. I'm not sure if this enables it, as the intended behavior is for it to be enabled by default in `hardhat node`, and disabled everywhere else, and by the time you check the config, it has already been populated with a `false`.

alcuadrado:
@fvictorio do you have any idea how this can be fixed?

fvictorio:
I'm not sure. The problem is that we can't tell apart between a `false` loggingEnabled set by the user and a default one, right?

One option (that I think we've discussed before?) is to expose the original user config *and* the resolved one as different properties of the HRE. But I don't know if that will always work, especially with respect to plugins that override some config value. For example: if the user doesn't set the `loggingEnabled`, then it will be undefined, but what if a plugin wants to override this value to make it *explicitly* false? How would that work?

Another alternative is to just keep this value as undefined instead of defaulting to `false`, but that sounds like a breaking change. Not a big one, but not negligible either.

alcuadrado:
As discussed, this doesn't fix the issue and doesn't seem to be fixable right now.

PR #1128, opened 2020-12-23
Added multiple blocks mining

### LINKED ISSUE #1112: Mine several blocks in a single call

sisco0:
Closes #1112

By using this approach multiple blocks could be mined with a single call that could specify as parameters:

1. The number of blocks to be mined
2. An array of timestamps to be set for the first mined blocks (optional)

It should be taken into account that naive `_mineEmptyBlock` function is called, where, the inner components of it are not modified.
I would be happy to collaborate by modifying needed code sections as now I am more integrated in the project in order to understand better the source code and its architecture.

### DIFF
--- a/docs/hardhat-network/README.md
+++ b/docs/hardhat-network/README.md
@@ -268,6 +268,7 @@ To customise it, take a look at [the configuration section](/config/README.md#ha
 
 - `evm_increaseTime` – same as Ganache.
 - `evm_mine` – same as Ganache
+- `evm_mineMultiple` – this method works like `evm_mine`, but it forces a group of consecutive blocks to be mined. It takes one mandatory and one optional parameter, which are the number of blocks to be mined and the array of timestamps for the mining time of the first k-blocks.
 - `evm_revert` – same as Ganache.
 - `evm_snapshot` – same as Ganache.
 - `evm_setNextBlockTimestamp` - this method works like `evm_increaseTime`, but takes the exact timestamp that you want in the next block, and increases the time accordingly.
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/input.ts
@@ -326,6 +326,12 @@ export function validateParams(params: any[], num: typeof t.number): [number];
 
 export function validateParams(
   params: any[],
+  num: typeof t.number,
+  numarr: typeof t.Array
+): [number, number[]];
+
+export function validateParams(
+  params: any[],
   hash: typeof rpcHash,
   bool: typeof t.boolean
 ): [Buffer, boolean];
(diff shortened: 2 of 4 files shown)


bartjman:
How fast would this mine 100.000 blocks?

sisco0:
> How fast would this mine 100.000 blocks?

Have a good day @bartjman , I am testing it right now.

sisco0:
By the definition of an experiment where `evm_mine` is compared against `evm_mineMultiple` without the use of `timestamp` for any of these methods the results are these for each power of ten from 1e2 to 1e5 (the case you pointed out):

```
Benchmark results (prettier):

100 iterations
[evm_mine] Time taken: 1087.198212ms
[evm_mineMultiple] Time taken: 272.138408ms

1000 iterations
[evm_mine] Time taken: 1985.178809ms
[evm_mineMultiple] Time taken: 1049.418932ms

10000 iterations
[evm_mine] Time taken: 13055.387341ms
[evm_mineMultiple] Time taken: 8577.462084ms

100000 iterations
[evm_mine] Time taken: 135421.579329ms
[evm_mineMultiple] Time taken: 112354.947162ms
```

These tests have been carried out under a replicable AWS `t2.2xlarge` (8 vCPUs) cloud instance, where, from the results it could be inferred the next **points**:

1. The acceleration factor is **always greater than 100%**, so `evm_mineMultiple` method is always faster than using `evm_mine` in a loop.
2. The acceleration factor goes from ~**4 times faster** for the proposed method when 1e2 iterations are performed and ~120.5% for the proposed method when 1e5 iterations are performed.

As the **conclusion**, the proposed `evm_mineMultiple` method behavior is better than using the `evm_mine` in a loop, with a noticeable boost up to ~4x for the benchmark run. Even more, it must be pointed out that the new `evm_mineMultiple` also accepts a `timeline` vector optionally for the new empty blocks mined.

Attached script to use with PR version by using `npx hardhat run scripts/benchmark.js`:

```javascript
async function mineSerial(iterations) {
  let hrTime = process.hrtime();
  for (let k = 0; k < iterations; k++) {
    await ethers.provider.send("evm_mine");
  }
  hrTime = process.hrtime(hrTime);
  console.log(`[evm_mine] Time taken: ${(hrTime[0] * 1e9 + hrTime[1])/1e6}ms`);
}
async function mineParallel(iterations) {
  let hrTime = process.hrtime();
  await ethers.provider.send("evm_mineMultiple", [iterations, []]);
  hrTime = process.hrtime(hrTime);
  console.log(`[evm_mineMultiple] Time taken: ${(hrTime[0] * 1e9 + hrTime[1])/1e6}ms`);
}
async function benchmarkTimes(iterations, measureSerial = true) {
  if (measureSerial) {
    await mineSerial(iterations);
  }
  await mineParallel(iterations);
}
(async () => {
  console.log('100 iterations');
  await benchmarkTimes(100);
  console.log('1000 iterations');
  await benchmarkTimes(1000);
  console.log('10000 iterations');
  await benchmarkTimes(10000);
  console.log('100000 iterations');
  await benchmarkTimes(100000);
})();
```

alcuadrado:
Hi @sisco0,

As I mentioned in #1112, we are not ready to merge a change related to this just yet. If you want, try resubmitting a PR once #1087 has landed.

Regarding this implementation's performance, I think it's still extremely slow. It takes about 2 minutes to increase less than 2 weeks worth of blocks. It's prohibitively slow for testing long-term block-based conditions.

PR #1131, opened 2020-12-24
Upgrade solidity parser

### LINKED ISSUE #1130: HardHat 2.0.6 does not compile legit contracts (not a Solidity bug)
https://github.com/vporton/hardhat-solc-versions-bug

```
$ npx hardhat compile
Error HH606: The project cannot be compiled, see reasons below.

The Solidity version pragma statement in these files don't match any of the configured compilers in your config. Change the pragma or configure additional compiler versions in your hardhat config.

  * abdk-libraries-solidity/ABDKMath64x64.sol (^0.5.0)

These files import other files that use a different and incompatible version of Solidity:

  * contracts/import.sol (^0.7.5) imports abdk-libraries-solidity/ABDKMath64x64.sol (^0.5.0)

To learn more, run the command again with --verbose

Read about compiler configuration at https://hardhat.org/config


For more info go to https://hardhat.org/HH606 or run Hardhat with --show-stack-traces
```

fvictorio:
Closes #1130

### DIFF
--- a/packages/hardhat-core/package.json
+++ b/packages/hardhat-core/package.json
@@ -77,7 +77,7 @@
   "dependencies": {
     "@nomiclabs/ethereumjs-vm": "^4.1.1",
     "@sentry/node": "^5.18.1",
-    "@solidity-parser/parser": "^0.10.1",
+    "@solidity-parser/parser": "^0.10.2",
     "@types/bn.js": "^4.11.5",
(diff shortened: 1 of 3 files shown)


sisco0:
Please take into account my comment at the source code. From my investigations it seems that the parsing ends when a `||` symbol is found.

fvictorio:
@sisco0 yes, this PR fixes that issue.

PR #1148, opened 2021-01-04
Fix typos in testing-contracts.md

peebeejay:
Change `assertion objet` -> `Assertion object`


alcuadrado:
Good catch, @peebeejay! Thanks for fixing this :)

PR #1150, opened 2021-01-04
Add getSigner to hardhat-ethers

scnale:
This pull request:
- Adds a `getSigner(address: string): Promise<SignerWithAddress>` function for getting a signer with an arbitrary address.
- Fixes `getContractAt` so it uses a read only provider as fallback if there are no available signers.


packages/hardhat-ethers/README.md
```diff
@@ -69,7 +69,8 @@ function getContractAt(nameOrAbi: string | any[], address: string, signer?: ethe
 function getSigners() => Promise<ethers.Signer[]>;
 ```
 
-The `Contract`s and `ContractFactory`s returned by these helpers are connected to the first signer returned by `getSigners` be default.
+The `Contract`s and `ContractFactory`s returned by these helpers are connected to the first signer returned by `getSigners` by default.
+If there is no signer available, the provider is used as is to provide read-only access.
```
alcuadrado, line 73:
```suggestion

If there is no signer available, `getContractAt` returns read-only contracts.
```

packages/hardhat-ethers/test/no-account.test.ts
```diff
@@ -0,0 +1,129 @@
...
+          ]);
+
+          const greeting = await contract.functions.greet();
+
+          assert.equal(greeting, "Hi");
+        });
+      });
+    });
+
+    describe("getSigner", function () {
```
alcuadrado, line 92: This test is out of place here.

packages/hardhat-ethers/test/no-account.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        const signers = await this.env.ethers.getSigners();
+        assert.isEmpty(signers);
+      });
+    });
+
+    describe("getContractAt", function () {
+      let signerAddress: string;
+
+      beforeEach(async function () {
+        await this.env.network.provider.request({
```
alcuadrado, line 26: I'd add a comment explaining this.

alcuadrado: It's pretty clever, but not obvious.

packages/hardhat-ethers/test/no-account.test.ts
alcuadrado: I mean in this file, named 'no-account'.

packages/hardhat-ethers/test/no-account.test.ts
```diff
@@ -0,0 +1,129 @@
+import { assert } from "chai";
```
alcuadrado, line 1: Do we have files with `.test.`?

packages/hardhat-ethers/test/no-account.test.ts
```diff
@@ -0,0 +1,129 @@
...
+        signerAddress = miner;
+
+        await this.env.run(TASK_COMPILE, { quiet: true });
+      });
+
+      describe("with the name and address", function () {
+        it("Should return an instance of a contract with a read-only provider", async function () {
+          const receipt = await deployGreeter(this.env, signerAddress);
+
+          assert.isTrue(this.env.ethers.provider._isProvider);
```
alcuadrado, line 41: I wouldn't test this. It's an implementation detail.

packages/hardhat-ethers/test/no-account.test.ts
```diff
@@ -0,0 +1,129 @@
...
+          const contract = await this.env.ethers.getContractAt(
+            "Greeter",
+            receipt.contractAddress
+          );
+
+          assert.isDefined(contract.provider);
+          assert.isNotNull(contract.provider);
+          assert.strictEqual(contract.provider, this.env.ethers.provider);
```
alcuadrado, line 49: I'd consider this strict equality an implementation detail. It could be a clone.

packages/hardhat-ethers/test/no-account.test.ts
```diff
@@ -0,0 +1,129 @@
...
+          assert.containsAllKeys(contract.functions, [
```
alcuadrado, line 50: IMO this is testing ethers, not the plugin.

packages/hardhat-ethers/test/no-account.test.ts
```diff
@@ -0,0 +1,129 @@
...
+          const greeterArtifact = await this.env.artifacts.readArtifact(
+            "Greeter"
+          );
+
+          const contract = await this.env.ethers.getContractAt(
+            greeterArtifact.abi,
+            receipt.contractAddress
+          );
+
+          assert.isDefined(contract.provider);
```
alcuadrado, line 77: Same comments apply here.

packages/hardhat-ethers/test/no-account.test.ts
scnale: I actually thought it could go here since it ensures that this account doesn't exist in the network.

packages/hardhat-ethers/test/no-account.test.ts
scnale: Not that I can see. It's currently unnecessary.

alcuadrado: I'd remove it, to keep things consistent.

packages/hardhat-ethers/test/no-account.test.ts
alcuadrado: I see your reasoning now, but that's not an actual precondition of the function, nor it's tested.

scnale: Hm, that's true. I can add an assert for the non existence of the account.

It's not a precondition, but it should be the most commonly used case. I.e. when impersonating an account, that account won't exist. I'm not sure if adding a test for an existing account is useful though, since as you say, it's not a precondition really.

fvictorio (approved):

PR #1151, opened 2021-01-04
Make console.sol work with solc 0.8.x

### LINKED ISSUE #1132: hardhat/console.sol v2.0.6 cannot work with Solidity 0.8.0
The Solidity version pragma statement in these files don't match any of the configured compilers in your config. Change the pragma or configure additional compiler versions in your hardhat config.
hardhat/console.sol (>=0.4.22 <0.8.0)
Please help. Thank you

fvictorio:
Closes #1132.


alcuadrado:
This is not documented, but `console.sol` is generated by https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/scripts/console-library-generator.js, so we should update that instead.

fvictorio:
Oops, fixed!

PR #1162, opened 2021-01-07
Fix name resolution for `hardhat/console.sol` when `hardhat` directory exists.

### LINKED ISSUE #998: bug: resolving `hardhat/console.log` fails if there's a directory named `hardhat` in scope
If e.g. you have a `hardhat` directory at the root of your project, you can't import `hardhat/console.log` anymore. It just stops looking for the file. Node module resolution would continue the path lookup scheme by further walking up the directory tree (also looking into all parent `node_modules/` directories along the way). I think hardhat import resolution should follow that same behavior.

scnale:
If the directory `hardhat` exists in the project root, then importing
`hardhat/console.sol` fails if it doesn't exist within such a directory.
This adds an exception for this case and ensures the path resolves to
the `console.sol` packaged with Hardhat.

Fixes #998.

### DIFF
--- a/packages/hardhat-core/src/utils/source-names.ts
+++ b/packages/hardhat-core/src/utils/source-names.ts
@@ -58,7 +58,14 @@ export async function isLocalSourceName(
   projectRoot: string,
   sourceName: string
 ): Promise<boolean> {
-  if (sourceName.includes(NODE_MODULES)) {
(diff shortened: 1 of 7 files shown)


alcuadrado:
There's no test showing that a contract can actually use `hardhat/console.sol` when there's a `hardhat/` dir in the root of the project. I believe this will work, but we should still test it.

PR #1170, opened 2021-01-11
docs: mention archive noe in mainnet-forking

paulrberg:
The first time I read mainnet forking I tried using my Infura account, but that didn't work. I think specifying that the node must be an archive node helps a bit.


fvictorio (approved):

fvictorio:
Thanks Paul!

PR #1172, opened 2021-01-11
Fix SSTORE gas cost in fork mode

alcuadrado:
This PR fixes the cost of SSTORE. Before this PR writing a value that used to be zero was incorrectly priced.


gitpusha:
I just upgraded to hardhat `2.0.8` which includes this fix and now my `codechecks` gas-reporter is shouting at me because gas increased. 

Awesome! 

I think this fixes part of this or maybe even whole: #990 ?

PR #1189, opened 2021-01-20
Add JSON-RPC `eth_signTypedData_v4` method to Hardhat Network.

scnale:
This pull request:
- Adds JSON-RPC method `eth_signTypedData_v4` and implements it with `signTypedData_v4` of `eth-sig-util`.
  - Its implementation mimics MetaMask's. It attempts to parse the typed message parameter as a JSON if it is a string.
- Modifies the JSON-RPC method `eth_signTypedData` to mark it as not supported.
- Adds JSON-RPC method `eth_signTypedData_v3` and marks it as not supported.
- Adds method `_signTypedData` for `SignerWithAddress` in `hardhat-ethers`.
- Adds test in `hardhat-ethers` for `SignerWithAddress._signTypedData`.

Context:  [EIP712](https://eips.ethereum.org/EIPS/eip-712) early implementation.

### DIFF
--- a/packages/hardhat-ethers/src/signers.ts
+++ b/packages/hardhat-ethers/src/signers.ts
@@ -4,13 +4,6 @@ import { ethers } from "ethers";
 // EIP-712 Typed Data
 // See: https://eips.ethereum.org/EIPS/eip-712
 
-export type SignTypedDataParams = Parameters<
-  ethers.providers.JsonRpcSigner["_signTypedData"]
->;
-
-export type TypedDataDomain = SignTypedDataParams[0];
-export type TypedDataFields = SignTypedDataParams[1];
-
 export class SignerWithAddress extends ethers.Signer {
   public static async create(signer: ethers.providers.JsonRpcSigner) {
     return new SignerWithAddress(await signer.getAddress(), signer);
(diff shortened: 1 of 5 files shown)


packages/hardhat-ethers/src/signers.ts
```diff
@@ -1,13 +1,30 @@
...
+  version?: string;
+  chainId?: ethers.BigNumberish;
+  verifyingContract?: string;
+  salt?: ethers.BytesLike;
+}
+
+interface TypedDataField {
+  name: string;
+  type: string;
+}
```
fvictorio, line 18: I would replace this with:

```
type SignTypedDataParams = Parameters<ethers.providers.JsonRpcSigner['_signTypedData']>
type TypedDataDomain = SignTypedDataParams[0]
type TypedDataFields = SignTypedDataParams[1]
```

fvictorio: In fact, we can use only the first line and then do:

```
  public _signTypedData(...args: SignTypedDataParams): Promise<string> {
    return this._signer._signTypedData(...args);
  }
```

below.

scnale: I actually attempted to do this at first but it wouldn't recognize the `_signTypedData` member for some reason. It might have been my IDE Typescript integration acting up though, I don't think I tried actually building it.

alcuadrado: I went ahead and implemented it, without creating a new type, but rather making it explicit that it receives the same params.

alcuadrado (approved):

PR #1190, opened 2021-01-20
Fixed balance default value in Wei.

sshmaxime:
The default value is actually 10000 ETH but writing 10000 in the balance field will not produce anything. You need to write 10000 ETH in Wei, i.e 10000000000000000000000.


ItsNickBarry:
I think it would be best to include the data type as well.

sshmaxime:
Done :) I think you have an issue with your Windows CI lol

ItsNickBarry:
I didn't read the existing text properly; it already says that the type is a String.  Maybe put it in quotes like this: `Default value: "10000000000000000000000" (i.e 10000 ETH).`

fvictorio:
Good catch @sshmaxime :tada:

PR #1211, opened 2021-01-30
Fix Solidity AST parser for at least one kind of function parameters in solc v0.8

scnale:
In the v0.8 compiler series, the type description AST node for struct parameters changed slightly. This triggered a latent bug in our parser logic since we were always attempting to calculate function selectors even for private or internal functions.

Since this is a modern compiler version, function selectors are present in function definition nodes for public and external functions. Restricting a bit more our parsing logic to avoid this calculation when unnecessary should suffice.


alcuadrado (approved):

flcoder:
When I compile a v0.8.4 contract where, instead of `address` I tend to use `IBEP20` directly as the parameter type, I get the warning from hardhat that "The Hardhat Network tracing engine could not be initialized." and with verbose it seems to choke here: hardhat:core:hardhat-network:node  TypeError: Cannot read property 'startsWith' of undefined
    at toCanonicalAbiType /node_modules/hardhat/src/internal/hardhat-network/stack-traces/compiler-to-model.ts:536:12

I'm not sure if this helps with anything but I figured I'd just leave it here FYI.

PR #1220, opened 2021-02-02
Update debugging-with-hardhat-network.md

nicolasguasca1:
For the require (Token.sol) first argument to evaluates to `false` is necessary to change the position of _expect_ and _await_ interchangeably:

Token.js (test)

    it("Should fail if sender doesn’t have enough tokens", async function () {
      const initialOwnerBalance = await hardhatToken.balanceOf(owner.address);

      // Try to send 1 token from addr1 (0 tokens) to owner (1000 tokens).
      // `require` will evaluate false and revert the transaction.
      **expect**(
        **await** hardhatToken.connect(addr1).transfer(owner.address, 1)
      ).to.be.revertedWith("Not enough tokens");

### DIFF
--- a/docs/tutorial/debugging-with-hardhat-network.md
+++ b/docs/tutorial/debugging-with-hardhat-network.md
@@ -23,7 +23,7 @@ function transfer(address to, uint256 amount) external {
     console.log("Sender balance is %s tokens", balances[msg.sender]);
     console.log("Trying to send %s tokens to %s", amount, to);
 
-    require(balances[msg.sender] >= amount, "Not enough tokens");
+    require(balances[msg.sender] <= amount, "Not enough tokens");
 
     balances[msg.sender] -= amount;
     balances[to] += amount;


alcuadrado:
I don't understand this PR. Can you explain it a little more?

nicolasguasca1:
Sure! Let me rephrase. The way is written in the contract, it requires the amount of msg.sender to be greater or equal to the value to be transfered in order to evaluate to false. Everything around that is correct.

However, one of the tests is set up to transfer value when the balance is shorter than the amount to transfer (actually 0). So that test should fail and it is indeed described as a "should fail" test. The curious thing is that after running the tests, the "should fail" one ends passing. In order for that test to fail, the require statement in the contract needs to be edited. ”Await" should be put in the place where "expect" is and viceversa. 

I'm assuming off course that the test are intended to do what they say. I just found it interesting while doing the tutorial. Hope this explanation helps and thank you so much for building such a great tool to develop btw.

alcuadrado:
Hey @nicolasguasca1, sorry we took so long to get back to this. The message "should fail..." indicates that the transfer should fail, not the test. The current version is actually correct, so I'm closing this issue. We appreciate you took the time to submit the change though :)

PR #1221, opened 2021-02-03
update plugins.js - Hardhat react, correct git rep

RobertoSnap:
Update hardhat react to https://github.com/symfoni/symfoni-monorepo/tree/master/packages/hardhat-react


alcuadrado:
Something is preventing netlify from building this branch. We'll investigate it soon.

PR #1223, opened 2021-02-03
feat: flatten licenses flag

nachomazzara:
Nowadays when trying to validate the code at Etherscan or other block explorers, and also trying to copy/paste the flattened code on Remix, you will get an error saying that you must have one license per file.

Screen Shot 2021-02-03 at 14 57 15
Screen Shot 2021-02-03 at 14 56 54

I propose two flags to extend `hardhat flatten`:

`unifyABIEncoderV2`: Remove all occurrences of `pragma experimental ABIEncoderV2` except the first one.
`removeLicences`: Remove all licenses of the files to be flattened.
`license`: Add a unique license at the beginning of the file flattened. `--license MIT` will add `// SPDX-License-Identifier: MIT`.

Usage:

`npx hardhat flatten  --unifyABIEncoderV2`
`npx hardhat flatten --removeLicenses`
`npx hardhat flatten --license MIT`
`npx hardhat flatten --license MIT --unifyABIEncoderV2`


nachomazzara:
@fvictorio @alcuadrado any thoughts on why CI is not passing?

packages/hardhat-core/test/builtin-tasks/flatten.ts
```diff
@@ -327,6 +327,7 @@ contract AWithLicense {}`
             license: "MIT",
           }
         );
+        console.log(allFlattened)
```
fvictorio: I guess this was temporary?

packages/hardhat-core/src/builtin-tasks/flatten.ts
```diff
@@ -115,6 +117,11 @@ subtask(
         flattened = `// SPDX-License-Identifier: ${license}\n\n${flattened}`;
       }
 
+      if (unifyABIEncoderV2) {
+        // Remove every line started with "pragma experimental ABIEncoderV2;" except the first one
+        flattened = flattened.replace(/pragma experimental ABIEncoderV2;\n/gm, (i => (m: string) => !i++ ? m : '')(0));
```
fvictorio: This line is very hard to understand, is there a way to make it more readable?

packages/hardhat-core/src/builtin-tasks/flatten.ts
```diff
@@ -70,11 +71,12 @@ subtask(
   "Returns all contracts and their dependencies flattened"
 )
   .addOptionalParam("files", undefined, undefined, types.any)
+  .addOptionalParam("unifyABIEncoderV2", undefined, false, types.boolean)
   .addOptionalParam("removeLicenses", undefined, false, types.boolean)
   .addOptionalParam("license", undefined, undefined, types.string)
   .setAction(
     async (
-      { files, removeLicenses = false, license }: FlattenInput,
+      { files, unifyABIEncoderV2 = false, removeLicenses = false, license }: FlattenInput,
```
fvictorio: This option seems very ad hoc. I understand that this is what was required in the issue, but it doesn't feel right.

Maybe it could be called `unifyPragmas`, even if it only does this for the ABI encoder v2 one. If it happens that there's another pragma that needs this, we can add it in the future.

@alcuadrado wdyt?

fvictorio: See also: https://github.com/nomiclabs/hardhat/issues/1050#issuecomment-838308098

nachomazzara: @fvictorio I have added it cause I forgot to add it when I made the PR. 

From Remix, you get the following errors when you flattened contracts with _ABIEncoder V2_. I would like to keep that flag.
Screen Shot 2021-05-11 at 08 32 32

About the naming, `unifyPragmas` can be confused because you can still have different _pragma solidity_ versions. Maybe `unifyExperimentalPragmas` works can't predict if we will have more experimental pragmas in the future.

packages/hardhat-core/src/builtin-tasks/flatten.ts
nachomazzara: Ok, I just want to keep it as functional as I can. It was a time ago the last time I've used an IIFE.

Do you think that with better variables names works? 

```ts
flattened = flattened.replace(/pragma experimental ABIEncoderV2;\n/gm, (index => (match: string) => !index++ ? match: '')(0));
```

If not, alternative:

```ts
let index = 0
flattened = flattened.replace(/pragma experimental ABIEncoderV2;\n/gm, (match: string) => !index++ ? match : '');
```

Other solutions will include maybe a complex regex or the usage of `split` & `replace`.

Any other idea?

packages/hardhat-core/test/builtin-tasks/flatten.ts
nachomazzara: Was removed on https://github.com/nomiclabs/hardhat/pull/1223/commits/2a291330e6c5d070241256c5059899fc7bc5b00f

fvictorio:
@nachomazzara are tests passing for you locally? I'm getting 4 failures with the latest changes.

packages/hardhat-core/src/builtin-tasks/flatten.ts
fvictorio: I just pushed a change for this to make it more like the code we tend to write. We usually avoid compact one-liners and err on the "verbose but clear" side.

nachomazzara:
We should be good to go!

jtakalai:
Any reason why this PR was closed? There is no "competing" PR open that also fixes this, only an open issue.

alcuadrado:
We decided to take a different direction which is described in #1499. We haven't executed on it yet, but mergint this PR would prevent us from doing it.

PR #1239, opened 2021-02-11
(WIP) Add node 14 and 15 to CI

fvictorio:


alcuadrado:
We don't need 13, it's been EOL'ed for a long time.

PR #1240, opened 2021-02-11
Fix tests that depend on a 0 nonce

fvictorio:
Several tests were assuming the first account had a 0 nonce, and therefore they broke after #979.

This PR mostly moves those tests to use the second account, that right now has a nonce of 0. This is very brittle, of course, but it should work for now.

There is only one non-trivial change, I will comment it inline.

### DIFF
--- a/packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
+++ b/packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
@@ -11,6 +11,10 @@ export const DEFAULT_NETWORK_ID = 234;
 export const DEFAULT_BLOCK_GAS_LIMIT = 6000000;
 export const DEFAULT_USE_JSON_RPC = false;
 export const DEFAULT_ALLOW_UNLIMITED_CONTRACT_SIZE = false;
+
+// Assumptions:
+// - First account has sent some transactions on mainnet
+// - Second and third accounts have a 0 nonce
 export const DEFAULT_ACCOUNTS = [
   {
     privateKey:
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts
```diff
@@ -614,13 +614,15 @@ describe("Evm module", function () {
...
+            const firstBlock = await mineEmptyBlock();
             await this.provider.send("evm_increaseTime", [100]);
             const snapshotBlock = await mineEmptyBlock();
+
             const snapshotId = await this.provider.send("evm_snapshot");
 
             assert.equal(
               quantityToNumber(snapshotBlock.timestamp),
-              getCurrentTimestamp() + 100
+              quantityToNumber(firstBlock.timestamp) + 100
```
fvictorio, line 625: I honestly don't know why this test was working before. Now it was failing for a difference of a few milliseconds. I guess this is related to the timestamp change in #979, but I don't see how.

Anyway, now I just mine a block at the beginning of the test to compare the result of `evm_increaseTime` against that.

packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
```diff
@@ -11,6 +11,10 @@ export const DEFAULT_NETWORK_ID = 234;
 export const DEFAULT_BLOCK_GAS_LIMIT = 6000000;
 export const DEFAULT_USE_JSON_RPC = false;
 export const DEFAULT_ALLOW_UNLIMITED_CONTRACT_SIZE = false;
+
+// Assumptions:
+// - First account has sent some transactions on mainnet
+// - Second and third accounts have a 0 nonce
```
alcuadrado, line 17: This will break 😂

packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts
alcuadrado: Milliseconds? I think the timestamp is in seconds.

PR #1244, opened 2021-02-12
Add warning for unsupported Solidity versions

fvictorio:


packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -215,3 +214,46 @@ function readPackageJson(packageName: string): PackageJson | undefined {
...
+      unsupportedVersions.push(solcVersion);
+    }
+  }
+
+  if (unsupportedVersions.length > 0) {
+    console.warn(
+      chalk.yellow(
+        `Solidity ${unsupportedVersions.join(", ")} ${
+          unsupportedVersions.length === 1 ? "is" : "are"
+        } not supported yet. You can still use Hardhat, but some features might not work correctly.
```
alcuadrado, line 252:
```suggestion
        } not fully supported yet. You can still use Hardhat, but some features might not work correctly.
```

alcuadrado, line 252:
```suggestion
        } not fully supported yet. You can still use Hardhat, but some features, like stack traces, might not work correctly.
```

PR #1249, opened 2021-02-13
Support projects with Solidity and Vyper files

transferAndCall:
Fixes mixed-language environments to support compiling Solidity and Vyper in the same project.

### DIFF
--- a/packages/hardhat-vyper/src/index.ts
+++ b/packages/hardhat-vyper/src/index.ts
@@ -1,6 +1,10 @@
-import { TASK_COMPILE } from "hardhat/builtin-tasks/task-names";
-import { extendConfig, task } from "hardhat/internal/core/config/config-env";
+import {
+  TASK_COMPILE_GET_COMPILATION_TASKS,
+  TASK_COMPILE_SOLIDITY,
+} from "hardhat/builtin-tasks/task-names";
+import { extendConfig, subtask } from "hardhat/internal/core/config/config-env";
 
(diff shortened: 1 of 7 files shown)


alcuadrado:
Ohh, you did it! thanks 😍

packages/hardhat-vyper/src/index.ts
```diff
@@ -1,14 +1,25 @@
...
 
 extendConfig((config) => {
   const defaultConfig = { version: "latest" };
   config.vyper = { ...defaultConfig, ...config.vyper };
 });
 
-task(TASK_COMPILE, async (_, { config, artifacts }) => {
+subtask(
+  TASK_COMPILE_GET_COMPILATION_TASKS,
+  async (): Promise<string[]> => {
```
alcuadrado, line 17: This is a little picky, but when you override a task you can call the original one, so here I'd do

```ts
  async (_, _, runSuper): Promise<string[]> => {
    const otherTasks = await runSuper();
    return [...otherTasks, TASK_COMPILE_VYPER];
  }
```

This way the plugin is more robust and can interact with others that also override this task.

alcuadrado (commented):
This looks good. Thanks for doing it, and adding a test.

I just left a minor comment.

packages/hardhat-vyper/src/index.ts
transferAndCall: Exactly what I was looking for.

alcuadrado:
We'll release this early next week. We try not to do releases during the weekend.

PR #1250, opened 2021-02-13
Increase timeout for hardhat-truffle5 tests

transferAndCall:
I don't have a Windows machine to verify, but it appears that this CI test is failing due to timeout for a long-running test.


alcuadrado:
This is a weird bug that only happens in the CI. We haven't identified why yet.

We'll probably get to this once #1087 gets merged.

transferAndCall:
Cool. I'll give it one shot of increasing the timeout further then if it still fails I'll just close it.

transferAndCall:
Windows gonna be Windows.

PR #1266, opened 2021-02-21
Implement debug_traceTransaction

LogvinovLeon:
This one is the same as https://github.com/nomiclabs/hardhat/pull/1265 but against `tx-pool-pr` branch, not `master`.


alcuadrado:
(moving the conversation here)

Thanks for resubmitting this PR, @LogvinovLeon. Once again, sorry for not making clear how to submit it in the first place.

I created a branch `mempool`, and merged the implementation of our mempool (which was on the EthWork's branch) into it, as it's close to being ready to land on `master.` I also changed the base branch of this PR to `mempool`.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -841,6 +845,37 @@ export class HardhatNode extends EventEmitter {
...
+
+    return this._runInBlockContext(
+      new BN(block.header.number).subn(1),
+      async () => {
+        const blockGasLimit = this.getBlockGasLimit();
+        const gasLeft = blockGasLimit.clone();
+        for (const tx of block.transactions) {
+          const txHash = tx.hash();
+          if (txHash.equals(hash)) {
+            this._vmDebugTracer.enableTracing();
```
alcuadrado, line 864: We discussed offline with @fvictorio that the enable/disable should probably be in a try/finally clause, in case the tracer throws in the future.

alcuadrado:
Hey @LogvinovLeon, thanks for implementing this. We've already reviewed it with @fvictorio and it looks good.

I left a minor comment about a `try/finally` that we'll fix, and we'll add more complex test cases. I'll get a few traces from mainnet txs, and use the forking functionality to rerun them locally and compare the results.

There's something missing in the current implementation though. That's the params `disableStorage`, `disableMemory` and
`disableStack`. I suspect that having those properly implemented (i.e. not just removing those fields from the output) can lead to a significant performance increase in some cases, as they make the tracing almost quadratic in space and complexity. Do you think you can implement those?

alcuadrado:
Hey @LogvinovLeon, can you give us push access to this branch?

LogvinovLeon:
Hi @alcuadrado. Sorry for taking so long on this. I've invited you and @fvictorio to my fork so that you can have push access. Also - added the implementation for the second optional argument that will enable non-quadratic tracing.

As for the try/finally - I'm not sure I get your point here. If enabling throws - I expect it to not be enabled and if disabling throws - we can't disable it either way even with try catch.

alcuadrado:
Hey @LogvinovLeon, I saw your changes a few days ago, but missed the response until now. Thanks for implementing the missing params.

What @fvictorio suggested is that if `_reRunTx` throws here:

```typescript
this._vmDebugTracer.enableTracing(config);
await this._reRunTx(tx, block, gasLeft);
this._vmDebugTracer.disableTracing();
```

the vmDebugTracer may end up in a weird state, initialized from the previous run. Not that it will happen right now, but doing it like this will be more future proof


```typescript
try {
  this._vmDebugTracer.enableTracing(config);
  await this._reRunTx(tx, block, gasLeft);
} finally {
  this._vmDebugTracer.disableTracing();
}
```

Franco will commit this change though. I'll push a few mainnet traces that he'll use to also create a new test.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
```diff
@@ -22,7 +22,23 @@ export class VMDebugTracer {
     this._afterTxHandler = this._afterTxHandler.bind(this);
   }
 
-  public enableTracing(config: RpcDebugTracingConfig) {
+  /**
+   * Run the `cb` callback and trace its execution
+   */
+  public async trace(
+    cb: () => Promise<void>,
```
LogvinovLeon, line 29: Maybe it's better to call this: `action` instead of `callback` for two reasons.
1. We already have a function `_runInBlockContext` that does similar stuff and there the parameter is called `action`.
2. `Callback` implies fire and forget callback syntax.

fvictorio: Good point, fixed!

alcuadrado:
> One thing worth noticing is that with Geth traces - some gas costs are different.
> Noticeable exceptions are memory expansion and calls. Geth report call gas costs as the cost of the whole call tree, not just the opcode. Also - adds a cost of memory expansion when it happens. ethereumjs-vm doesn't do that - so the results will be different.

Good catch, I was confused about those. This info should probably be available in ethereumjs-vm. I'll raise this concern on their discord.

alcuadrado:
Closing this in favor of #1328

PR #1269, opened 2021-02-23
Update mainnet-forking doc to mention which networks are supported

liamzebedee:
Some networks that I've read that are supported are: 

 - mainnet
 - kovan and xdai (https://github.com/nomiclabs/hardhat/pull/1207)


alcuadrado:
Hey @liamzebedee, thanks for submitting this. It's not correct, because, good news, we support those chains :)

PR #1273, opened 2021-02-25
fix: storage slot param validation

fubhy:
I discovered this while toying around with a fork of a fork (mainnet => hardhat fork => hardhat fork) to optimize our test run performance.

~~The storage slot parameter is currently incorrectly validated via the `rpcQuantity` type.~~


alcuadrado:
Can you describe how this failed? I think the error may be better fixed somewhere else, as according to the RPC """spec""" this param should be a `QUANTITY`: https://eth.wiki/json-rpc/API#eth_getstorageat

alcuadrado:
Thanks again, @fubhy! This will be released next week.

PR #1291, opened 2021-03-02
Add Proxy for Fetch in download.ts - Resolves issue 1280

cdesch:
Added functionality to allow `HTTP_PROXY` and `HTTPS_PROXY` environment variables for proxy settings.

This addresses the [issue#1280](https://github.com/nomiclabs/hardhat/issues/1280) to be able to download the compilers behind a corporate firewall via a proxy. 

This pull request also adds test cases `packages/hardhat-core/test/internal/util/download.ts` for the existing functionality without using a proxy and using a proxy.


cdesch:
I feel that the Windows CI Build is failing across the board as noted with PR https://github.com/nomiclabs/hardhat/pull/1268

cdesch:
It would be mighty cool if this PR was checked out.... might help some us behind proxies use HardHat instead of the TruffleSuite. 

:grin:
--

alcuadrado:
Hey @cdesch, thanks for submitting this PR. We'll review it in the next few days or early next week.

cdesch:
Thank you. Surely it will make HardHat a little more accessible :-p

cdesch:
@alcuadrado is there anything I can do to assist with getting this pull request merged? :bowtie:

fvictorio:
Hey @cdesch, sorry for not having reviewed this yet. I'll take a look this week.

cdesch:
@fvictorio Thanks. Let me know if there is anything I can do to help. I'm happy to revise or making another PR if you have feedback. There can be some fun issues that crop up with proxies and it is kind of an edge case for some folks, which is why I included tests. I worked a little bit on adding custom certificates for the proxy but it seems that isn't needed currently. 

I used ENV variables in my commit as it is a pretty standard process for most proxies and the same as when using a proxy for `npm`.  `hardhat.config.js` didn't feel like the right place, but you may have a different opinion about including proxy configs there.

fvictorio:
> I used ENV variables in my commit as it is a pretty standard process for most proxies and the same as when using a proxy for `npm`. `hardhat.config.js` didn't feel like the right place, but you may have a different opinion about including proxy configs there.

That seems fine to me, there seems to be a convention around those two environment variables and as far as I can tell the right place to handle them in this scenario is in Hardhat (there's an issue about these variables in `node-fetch` but they decided that it should be done at the application level).

fvictorio:
@cdesch Thanks a lot for this. I made some minor changes. I'll add some comments explaining them in case you are interested :slightly_smiling_face:

packages/hardhat-core/src/internal/util/download.ts
```diff
@@ -11,8 +13,23 @@ export async function download(
   const { pipeline } = await import("stream");
   const { default: fetch } = await import("node-fetch");
   const streamPipeline = util.promisify(pipeline);
+  const fetchOptions: FetchOptions = {
+    timeout: timeoutMillis,
+  };
 
-  const response = await fetch(url, { timeout: timeoutMillis });
+  if (process.env.HTTPS_PROXY !== undefined) {
+    fetchOptions.agent = new HttpsProxyAgent(process.env.HTTPS_PROXY);
```
fvictorio, line 21: There was some unnecessary casting here. The properties of `process.env` have type `string | undefined`, so if you check for that in the `if`, typescript will know that the type is `string` inside the block.

packages/hardhat-core/src/internal/util/download.ts
```diff
@@ -1,5 +1,7 @@
 import fs from "fs";
 import fsExtra from "fs-extra";
+import { HttpsProxyAgent } from "https-proxy-agent";
+import type { RequestInit as FetchOptions } from "node-fetch";
```
fvictorio, line 4: Manually defining `FetchOptions` wasn't necessary, it is better to import it from the module.

packages/hardhat-core/test/internal/util/download.ts
```diff
@@ -0,0 +1,117 @@
...
+    let oldHttpProxyEnv: string | undefined;
+    let proxy: any;
+    let proxyPort: number;
+    let connections = 0;
+
+    beforeEach(function (done) {
+      // Setup Proxy Server
+      proxy = new Proxy();
+      connections = 0;
+      proxy.on("connection", () => {
```
fvictorio, line 38: I added this so we can assert that the proxy is really being used.

packages/hardhat-core/test/internal/util/download.ts
```diff
@@ -0,0 +1,117 @@
...
+        // Assert that the file exists
+        assert.isTrue(await fsExtra.pathExists(downloadPath));
+      });
+
+      afterEach(function () {
+        if (oldHttpsProxyEnv !== undefined) {
+          process.env.HTTPS_PROXY = oldHttpsProxyEnv;
+        } else {
+          delete process.env.HTTPS_PROXY;
+        }
```
fvictorio, line 76: To save the current environment, you were doing something like this:

```
env = process.env
process.env.HTTP_PROXY = ...
```

But since `process.env` is an object, any modifications you were doing to it were also done to `env`, so nothing was restored afterwards.

I also have to explicitly use `delete` here because it turns out that

```
process.env.FOO = undefined
```

doesn't remove the `FOO` environment variable. Instead, it sets it to the string `"undefined"`. Terrible behavior from node, but it is what it is.

packages/hardhat-core/src/internal/util/download.ts
cdesch: Thanks! Yeah, I was struggling a bit on the TypeScript learning curve, but I'm slowly migrating a project to TypeScript currently.

packages/hardhat-core/src/internal/util/download.ts
cdesch: Ah. I think I was a bit confused on that because of the `const { default: fetch } = await import("node-fetch");` which seemed strange as to where to put the import. Thank you!

packages/hardhat-core/test/internal/util/download.ts
cdesch: Nice catch. I think adapted that portion from the `https-proxy-agent` package, and in fact I wasn't sure if the `env` really needed to be changed back at the end of the test because I wasn't sure if changes to the `process.env` were scoped to the test or not. It seems like they are. 

Thanks for catching this.

cdesch:
Thank you @fvictorio for the comments and the feedback! I learned a lot in the process to get this working.

cdesch:
@fvictorio Do you think this PR will make the next HH release? Possibly 2.2.x or minor like 2.1.3?

Would you like me to add docs for this?

alcuadrado:
Hey @cdesch, thanks for this PR. Unfortunately we have to focus on implementing Berlin right now, so we won't get merged this in the next couple of weeks. Once the new hardfork support is out, we'll review and release this.

cdesch:
Thanks @alcuadrado, Can you recommend how I can install the NPM package from github in the meantime? I understand that typically you can do a `npm install https://github.com/<username>/<repository>` but since the HardHat repo has multiple packages in it under `packages/* ` such as `packages/hardhat-core`, etc... How would I install just one or more of them manually until the next release? 

My apologies for being persistent, this issue is a show stopper for my team using hardhat on a new project, but we understand that you have priorities too.

alcuadrado:
I'm afraid it can't be installed like that. Check the contributing.md which explains how to build it locally.

cdesch:
@alcuadrado Okay. Thank you.

cdesch:
@alcuadrado and @fvictorio Happy Weekend! Any chance of looking at this PR this week? 😇

cdesch:
I updated this PR to the current Master branch and attempted to resolve the merge conflict with `yarn.lock`. 

The test suite fails with:
```
  1 failing

  1) Eth module
       JSON-RPC provider
         eth_sendTransaction
           return txHash
             Should return the hash of a reverted transaction:
     TypeError: Cannot read property 'txHash' of undefined
      at Context.<anonymous> (test/internal/hardhat-network/provider/modules/eth.ts:4005:39)
      at processTicksAndRejections (node:internal/process/task_queues:94:5)

```

but that seems to be from [0a575a38a](https://github.com/nomiclabs/hardhat/commit/0a575a38a74a0d84992836cd5cb21518969837bf) and [PR 1425](https://github.com/nomiclabs/hardhat/pull/1425), although I could be wrong. The error does not appear to be related to this PR.

cdesch:
Checking in to see if this can be merged in the next version. Is there anything I can do to help merge this in?

alcuadrado:
Hey @cdesch! I'm really sorry I took so long to review this. Thanks for submitting this PR, and for being patient.

As I read that you were struggling a bit with ts, I pushed [this commit](https://github.com/nomiclabs/hardhat/pull/1291/commits/1083492cf24561392eeaca2a67679e4755b20441) improving a few tiny bits, which you may find interesting.

I also added some docs about this. I wasn't sure where to place this info, so I added a warning box to the compilation guide for now. What do you think, @fvictorio?

alcuadrado (approved):

cdesch (commented):
Thank you

cdesch:
Thank you @alcuadrado! I think everything looks good. I will test as soon as the package is published.

alcuadrado:
@fvictorio if you think the docs change is ok we can merge this.

fvictorio:
Thanks @cdesch and sorry for the huge delay!

cdesch:
No worries, I completely understand! Thank you @fvictorio ! I'm ready to test it out!

hygkui:
`npx hardhat --show-stack-traces --verbose compile ` can show the download url, you can download and move it to the right dir such like '$HOME/Library/Caches/hardhat-nodejs/compilers/macosx-amd64/solc-macosx-amd64-v0.x.x+commit.fc410830', and it can be sloved the compiler download issue.

PR #1296, opened 2021-03-02
Enable caching in Ropsten forks

alcuadrado:
This PR enables caching in Hardhat Network's forking when using Ropsten.

The reason this wasn't enabled before was that there's no clear info about what was the max reorg on Ropsten. We assume now that 100 is a reasonable bound. This may not be the case, but we'd revisit it if it were to happen.

@fvictorio do you think we can add this to tomorrow's release?


fvictorio (approved):

fvictorio:
Sure.

PR #1309, opened 2021-03-08
fix: pad mixHash to 32 bytes

liamzebedee:
mixHash is a [32 byte value](https://sourcegraph.com/github.com/ethereum/go-ethereum/-/blob/core/types/block.go#L83:21). The call to`bufferToRpcData` only pads the `mixHash` to 32 hexadecimal characters, which is 16 bytes. 

Currently this is breaking my usage of a forked Hardhat network with running a graph-node - https://github.com/graphprotocol/graph-node/issues/2212.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/output.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/output.ts
@@ -113,7 +113,7 @@ export function getRpcBlock(
     // We pad this to 8 bytes because of a limitation in The Graph
     // See: https://github.com/nomiclabs/hardhat/issues/491
     nonce: pending ? null : bufferToRpcData(block.header.nonce, 16),
-    mixHash: pending ? null : bufferToRpcData(block.header.mixHash, 32),
+    mixHash: pending ? null : bufferToRpcData(block.header.mixHash, 64),
     sha3Uncles: bufferToRpcData(block.header.uncleHash),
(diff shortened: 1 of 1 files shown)


liamzebedee:
Rebased on `hardhat-core-v2.1.1`

alcuadrado:
Thanks for fixing this @liamzebedee!

packages/hardhat-core/src/internal/hardhat-network/provider/output.ts
```diff
@@ -112,8 +112,8 @@ export function getRpcBlock(
     parentHash: bufferToRpcData(block.header.parentHash),
     // We pad this to 8 bytes because of a limitation in The Graph
     // See: https://github.com/nomiclabs/hardhat/issues/491
-    nonce: pending ? null : bufferToRpcData(block.header.nonce, 16),
-    mixHash: pending ? null : bufferToRpcData(block.header.mixHash, 32),
+    nonce: pending ? null : bufferToRpcData(block.header.nonce, 16), // TODO: null when it's a pending block,
```
alcuadrado, line 115: I think these todos were accidentally re-added, but they have been solved

liamzebedee: Ah woops, rebased on an old tag! Just removed 'em.

alcuadrado:
This will be released in our next version. Thanks!

PR #1316, opened 2021-03-10
Instability features

fvictorio:
This PR adds three new features:

- A `hardhat_dropTransaction` RPC method that can be used to remove a transaction from the mempool
- A new `minGasPrice` configuration property and a new `evm_setMinGasPrice` RPC method to configure the minimum gas price accepted by the Hardhat Network
- Support for replacing transactions in the mempool with new transactions with a higher gas price

### `hardhat_dropTransaction`

You can drop a transaction like this:

```js
const txHash = "0xabc..."
const result = await network.provider.send("hardhat_dropTransaction", [txHash])
```

There are three possible scenarios:

- `txHash` corresponds to a tx in the mempool. In this case the tx is removed and `result` is `true`.
- `txHash` doesn't correspond to a transaction. In this case nothing happens and `result` is `false`.
- `txHash` corresponds to a mined transaction. In this case an `InvalidArgumentsError` is thrown.

### Minimum gas price

The `minGasPrice` property can be configured using either a string or a number, and is represented as a BN in the resolved config. **This diverges from the current behavior**, where the types of the user config and the resolved config are the same, except that they are optional in the user config. I think this is worth it, because we probably want to manipulate this value as a BN but I don't think we want to make users configure it that way. Accepting both numbers as a string is just a convenience, but we can probably drop the `number` option to simplify (and avoid potential precision issues).

The `evm_setMinGasPrice` receives a hex encoded number that represents the new gas price (also in wei). **This is also a divergence**, because other methods in the evm module accept numbers. I did it this way because gas price values can be larger than block gas limits or intervals. On the other hand, `Number.MAX_SAFE_INTEGER` is nine million gwei, so maybe we can use a `number` safely? (People from 2025 reading this: don't make fun of us.)

The default `minGasPrice` value is 0.

### Transaction replacement

Not much to say this. If you send a transaction with the same (address, nonce) pair than one that already exists in the mempool, that transaction is replaced **if** the new transaction has a gas price that is at least 10% higher. Otherwise an `InvalidInputError` is thrown. Notice that this is a different error than the one thrown by `hardhat_dropTransaction`. I think `InvalidArgumentsError` makes more sense, but I didn't use it for consistence with the errors currently thrown by `TxPool`


packages/hardhat-core/src/internal/core/config/config-resolution.ts
```diff
@@ -154,12 +155,18 @@ function resolveHardhatNetworkConfig(
...
+      clonedDefaultHardhatNetworkParams.minGasPrice
+  );
+
   const config = {
     ...clonedDefaultHardhatNetworkParams,
     ...hardhatNetworkConfig,
     accounts,
     forking,
     mining,
+    minGasPrice,
```
fvictorio, line 169: This is the kind of change produced by the divergence between the `HardhatNetworkUserConfig` types and the `HardhatNetworkConfig` types. If they had the same type, then passing `minGasPrice` like this wouldn't be necessary.

packages/hardhat-core/src/internal/core/providers/construction.ts
```diff
@@ -77,13 +77,14 @@ export function createProvider(
...
-      hardhatNetConfig.chainId!,
-      hardhatNetConfig.blockGasLimit!,
-      hardhatNetConfig.throwOnTransactionFailures!,
-      hardhatNetConfig.throwOnCallFailures!,
+      hardhatNetConfig.chainId,
+      hardhatNetConfig.chainId,
+      hardhatNetConfig.blockGasLimit,
+      hardhatNetConfig.minGasPrice,
+      hardhatNetConfig.throwOnTransactionFailures,
+      hardhatNetConfig.throwOnCallFailures,
```
fvictorio, line 87: The only important change here is that `minGasPrice` is a new argument. The other line changes are unrelated to this PR, it's just that all these non-null assertions weren't necessary. They were probably a leftover from the previous config types.

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
```diff
@@ -75,29 +75,79 @@ export class TxPool {
...
+      minGasPrice: bnToHex(minGasPrice),
     });
     this._deserializeTransaction = (tx) => deserializeTransaction(tx, common);
   }
 
   public async addTransaction(tx: Transaction) {
     const senderAddress = this._getSenderAddress(tx);
-    const senderNonce = await this.getExecutableNonce(senderAddress);
+    const accountNonce = await this._getAccountNonce(senderAddress);
+    const nextNonce = await this.getNextNonce(senderAddress);
```
fvictorio, line 91: I renamed `executableNonce` to `nextNonce`. I'm also using `accountNonce` as the nonce that only considers mined transactions. Please let me know if there's a better vocabulary to use for this.

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
```diff
@@ -75,29 +75,79 @@ export class TxPool {
...
 
-    await this._validateTransaction(tx, senderAddress, senderNonce);
+    await this._validateTransaction(tx, senderAddress, accountNonce);
 
     const txNonce = new BN(tx.nonce);
 
-    if (txNonce.eq(senderNonce)) {
+    if (txNonce.lte(nextNonce)) {
```
fvictorio, line 97: Any transaction that is `<= nextNonce` is considered a pending transaction. This assumes that `_validateTransaction` already checked that the tx nonce is not less than the account nonce.

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
```diff
@@ -200,18 +264,15 @@ export class TxPool {
         }
 
         const txNonce = new BN(deserializedTx.data.nonce);
-        const txGasLimit = new BN(deserializedTx.data.gasLimit);
 
         if (
-          txGasLimit.gt(this.getBlockGasLimit()) ||
-          txNonce.lt(senderNonce) ||
-          deserializedTx.data.getUpfrontCost().gt(senderBalance)
+          !this._isTxValid(deserializedTx, txNonce, senderNonce, senderBalance)
```
fvictorio, line 275: Moved this to the `_isTxValid` method and added the `minGasPrice` logic there.

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
```diff
@@ -200,18 +264,15 @@ export class TxPool {
...
         ) {
           newPending = this._removeTx(newPending, address, deserializedTx);
 
-          // if we are dropping a pending transaction, then we move
-          // all the following txs to the queued
-          if (txNonce.gt(senderNonce)) {
+          // if we are dropping a pending transaction with a valid nonce,
+          // then we move all the following txs to the queued list
+          if (txNonce.gte(senderNonce)) {
```
fvictorio, line 281: Off-by-one error.

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
```diff
@@ -473,10 +531,132 @@ export class TxPool {
...
+
+    if (newQueuedTxs !== undefined) {
+      this._setQueuedForAddress(accountAddress, newQueuedTxs);
+      return true;
+    }
+
+    return false;
+  }
+
+  private _replaceTx(
```
fvictorio, line 617: This method has logic that is common to both `_replacePendingTx` and `_replaceQueuedTx`

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
```diff
@@ -473,10 +531,132 @@ export class TxPool {
...
+  }
+
+  private _getMinNewGasPrice(currentGasPrice: BN): BN {
+    let minNewGasPrice = currentGasPrice.muln(110);
+
+    if (minNewGasPrice.modn(100) === 0) {
+      minNewGasPrice = minNewGasPrice.divn(100);
+    } else {
+      minNewGasPrice = minNewGasPrice.divn(100).addn(1);
+    }
```
fvictorio, line 664: This is probably overkill, but... oh well.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1667,4 +1697,9 @@ export class HardhatNode extends EventEmitter {
       filterId,
     });
   }
+
+  private async _isTransactionMined(hash: Buffer): Promise<boolean> {
+    const txReceipt = await this.getTransactionReceipt(hash);
+    return txReceipt !== undefined;
```
fvictorio, line 1703: **I'm not sure about this at all**. Is this enough? Or should we also consider a defined receipt with a falsy block number?

packages/hardhat-core/test/internal/core/config/config-resolution.ts
```diff
@@ -37,8 +38,8 @@ describe("Config resolution", () => {
         assert.isUndefined(config.solidity.compilers[0]?.settings?.evmVersion);
         assert.equal(config.defaultNetwork, "hardhat");
 
-        const hardhatNetworkConfig: HardhatNetworkUserConfig = config.networks
-          .hardhat as HardhatNetworkUserConfig;
+        const hardhatNetworkConfig: HardhatNetworkConfig = config.networks
+          .hardhat as HardhatNetworkConfig;
```
fvictorio, line 42: Another change related to the difference between the type of `minGasPrice` in the user type and the resolved type. This is the correct type anyway, it's just that it wasn't a problem before.

docs/config/README.md
```diff
@@ -84,6 +84,8 @@ You can set the following fields on the `hardhat` config:
 
 - `blockGasLimit`: The block gas limit to use in Hardhat Network's blockchain. Default value: `12450000`
 
+- `minGasPrice`: The minimum gas price accepted by the Hardhat Network. Default value: `0`
```
alcuadrado, line 87:
```suggestion
- `minGasPrice`: The minimum gas price accepted by the Hardhat Network. Transactions with a lower gas price are accepted but not mined. Default value: `0`
```

docs/hardhat-network/README.md
```diff
@@ -234,6 +234,21 @@ next block by using the "pending" block tag:
...
+method:
+
+```js
+const txHash = "0xabc..."
+await network.provider.send("hardhat_dropTransaction", [txHash])
+```
+
+You can also replace a transaction by sending a new one with the same nonce as
+the one that it's already in the mempool but with a higher gas price.
+Keep in mind that for this to work the new gas price has to be at least 10%
```
alcuadrado, line 249:
```suggestion
Keep in mind that, like in Geth, for this to work the new gas price has to be at least 10%
```

docs/hardhat-network/README.md
```diff
@@ -371,6 +387,7 @@ To customise it, take a look at [the configuration section](/config/README.md#ha
 - `evm_revert` – same as Ganache.
 - `evm_snapshot` – same as Ganache.
 - `evm_setNextBlockTimestamp` - this method works like `evm_increaseTime`, but takes the exact timestamp that you want in the next block, and increases the time accordingly.
+- `evm_setMinGasPrice` - change the minimum gas price accepted by the network (in wei)
```
alcuadrado, line 390: I think this should be `hardhat_`

packages/hardhat-core/src/internal/core/config/config-resolution.ts
alcuadrado: yeah, but i think this is a good trade off

packages/hardhat-core/src/internal/core/config/default-config.ts
```diff
@@ -42,6 +44,7 @@ export const defaultHardhatNetworkParams: Omit<
   accounts: defaultHardhatNetworkHdAccountsConfigParams,
   loggingEnabled: false,
   gasMultiplier: DEFAULT_GAS_MULTIPLIER,
+  minGasPrice: new BN(0),
```
alcuadrado, line 47: oh, this is interesting, I thought default config used user-level types.

packages/hardhat-core/src/internal/core/providers/construction.ts
alcuadrado: Thanks for the clarification. 

I wonder if there's a tslint (eslint?) rule to forbid unnecessary `!`s.

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
alcuadrado: TBH I don't understand what each means now.

What about something like `nextPendingNonce` and `nextConfirmedNonce`?

fvictorio: I like that idea, but let's wait until we merge as many PRs as we can, because this rename already caused a lot of conflicts with other PRs :sweat_smile:

docs/hardhat-network/README.md
fvictorio: Done in 8cd2d41d9cd8e0259e83ec8033b59e9548bf83a6.

PR #1318, opened 2021-03-10
Add BSC support to buidler-etherscan

MickdeGraaf:
Some of our repositories have not yet migrated to hardhat (I know shame on us) but we would still like to verify contracts on BSC.

This PR adds the relevant API url for bscscan.

### DIFF
--- a/packages/buidler-etherscan/src/network/prober.ts
+++ b/packages/buidler-etherscan/src/network/prober.ts
@@ -14,6 +14,7 @@ enum NetworkID {
   RINKEBY = 4,
   GOERLI = 5,
   KOVAN = 42,
+  BSC = 56,
 }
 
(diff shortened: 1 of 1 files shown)


fvictorio:
Hi @MickdeGraaf, thanks for this. Sadly, we aren't maintaining the buidler package or its plugins anymore. You should be able to use your fork though by publishing under your username. Here's how you do that in case you are interested:

1. Change the package name so that it's under your npm username, for example `"name": "@mickdegraaf/buidler-etherscan"
2. Commit it (npm doesn't let you publish otherwise)
3. Run `npm publish --access public`. The `--access public` part is necessary because the package is scoped under your user.

(This is from the top of my head, maybe I'm missing something here)

I'm closing this, but let me know if you need help with that.

PR #1327, opened 2021-03-15
Sort batched responses by id

fvictorio:
We are assuming that the array returned from a batch request has the same order as the array of requests. This is not always the case. ArchiveNode.io, for example, seems to return them (sometimes) in a different order.

This PR fixes this, by using the `id` field to sort the responses.

No tests for this yet. Maybe we could have a couple of fork tests that run on ArchiveNode? But not all of them because the CI is already too slow as it is, and it wouldn't be nice to add that load to them either.

### DIFF
--- a/packages/hardhat-core/src/internal/core/providers/http.ts
+++ b/packages/hardhat-core/src/internal/core/providers/http.ts
@@ -76,9 +76,14 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
     // the async call that follows would probably loose of the stack trace
     const error = new ProviderError("HttpProviderError", -1);
 
-    const requests = batch.map((r) =>
-      this._getJsonRpcRequest(r.method, r.params)
-    );
+    // we need this to sort the responses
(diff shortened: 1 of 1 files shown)


packages/hardhat-core/src/internal/core/providers/http.ts
```diff
@@ -76,9 +76,14 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
     // the async call that follows would probably loose of the stack trace
     const error = new ProviderError("HttpProviderError", -1);
 
-    const requests = batch.map((r) =>
-      this._getJsonRpcRequest(r.method, r.params)
-    );
+    // we need this to sort the responses
+    const idToIndexMap: Record<string, number> = {};
```
fvictorio, line 80: I think I'm being too cautious here. We can be quite confident that the ids of the requests are going to be in order (because we call `this._getJsonRpcRequest` serially), so maybe it's enough to just sort the responses by id.

That being said, this approach might be more future proof.

PR #1328, opened 2021-03-15
Add `debug_traceTransaction` RPC method

fvictorio:
This PR is the same as #1266 but rebased so that it depends on #1327. I can't push-force to that branch so I'm creating this new one.

cc @LogvinovLeon


WyseNynja:
I am so excited

iamdefinitelyahuman:
I'm tired of acting professional on Github.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
```diff
@@ -0,0 +1,582 @@
...
+  }
+
+  private _getStack(step: InterpreterStep): string[] {
+    const stack = step.stack
+      .slice()
+      .map((el: BN) => el.toString("hex").padStart(64, "0"));
+    return stack;
+  }
+
+  private async _stepToStructLog(step: InterpreterStep): Promise<StructLog> {
```
alcuadrado, line 311: Is this logic affected by Berlin?

fvictorio: Uhm, I think `_callConstantGas` changes, yeah. Damn. Does the VM exposes some way of knowing if some address was accessed or not?

fvictorio: There is an `isWarmedAddress` method in the state manager. I'll use that.

fvictorio: Fixed in 2f80919914f6afbf2b34527526fd296be29abb80

docs/hardhat-network/README.md
```diff
@@ -282,6 +282,45 @@ eth_call
...
+  {
+    disableMemory: true,
+    disableStack: true,
+    disableStorage: true,
+  }
+]);
+```
+
+If you are using [mainnet forking](https://hardhat.org/guides/mainnet-forking.html) with an archive node,
+you can get traces of transactions from the mainnet even if the node you are using
```
alcuadrado, line 314:
```suggestion
you can get traces of transactions from the remote network even if the node you are using
```

docs/hardhat-network/README.md
```diff
@@ -282,6 +282,45 @@ eth_call
...
+  "0x123...",
+  {
+    disableMemory: true,
+    disableStack: true,
+    disableStorage: true,
+  }
+]);
+```
+
+If you are using [mainnet forking](https://hardhat.org/guides/mainnet-forking.html) with an archive node,
```
alcuadrado, line 313:
```suggestion
If you are using [the forking functionality](https://hardhat.org/guides/mainnet-forking.html) with an archive node,
```

docs/hardhat-network/README.md
```diff
@@ -282,6 +282,45 @@ eth_call
...
+you can get traces of transactions from the mainnet even if the node you are using
+doesn't support
+`debug_traceTransaction`.
+
+There are some known limitations around this method:
```
alcuadrado, line 318:
```suggestion
### Known limitations
```

packages/hardhat-core/src/internal/hardhat-network/jsonrpc/client.ts
```diff
@@ -37,6 +37,15 @@ export class JsonRpcClient {
     return this._networkId;
   }
 
+  public async getDebugTraceTransaction(transactionHash: Buffer): Promise<any> {
```
alcuadrado, line 40: This is only used in a test, right? It surprised me when I first saw it, as we trace remote txs locally.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -926,6 +928,76 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+
+    return this._runInBlockContext(
+      new BN(block.header.number).subn(1),
+      async () => {
+        const blockNumber = block.header.number.toNumber();
+        const blockchain = this._blockchain;
+        let vm = this._vm;
+        if (
+          blockchain instanceof ForkBlockchain &&
+          blockNumber <= blockchain.getForkBlockNumber().toNumber()
```
alcuadrado, line 950:
```suggestion
          blockchain.getForkBlockNumber().gtn(blockNumber)
```

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
```diff
@@ -222,6 +226,10 @@ export class ForkBlockchain implements HardhatBlockchainInterface {
     throw new Error("Method not implemented.");
   }
 
+  public getNetworkId() {
```
alcuadrado, line 229: I don't think this belongs in here. I'd rather have this info in the node. 

Also, we should add a `hardhat_getForkedNetworkId` or something like that.

alcuadrado: Not in this PR, but I mentioned it because that will lead to the Node having this info somehow.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/debug.ts
```diff
@@ -0,0 +1,43 @@
...
+  private _traceTransactionParams(
+    params: any[]
+  ): [Buffer, RpcDebugTracingConfig] {
+    return validateParams(params, rpcHash, rpcDebugTracingConfig);
+  }
+
+  private async _traceTransactionAction(
+    hash: Buffer,
+    config: RpcDebugTracingConfig
+  ): Promise<object> {
```
alcuadrado, line 40: Why is this `object` instead of `RpcDebugTraceOutput`? I suspect this is here because `HardhatNode.traceTransaction` returns `Promise<object>`, but if its return type is removed ts correctly infers `Promise<RpcDebugTraceOutput>`.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1692,3 +1764,15 @@ Hardhat Network's forking functionality only works with blocks from at least spu
     });
   }
 }
+
+function getCommon(networkId: number, blockNumber: number): Common | undefined {
+  try {
+    const common = Common.forCustomChain(networkId, {});
```
alcuadrado, line 1770: I think we don't want a custom one here, but rather a normal/standard `Common`. You can get one with `new Common({chain: networkId})`.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1692,3 +1764,15 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+
+    common.setHardfork(common.activeHardfork(blockNumber));
+
+    return common;
+  } catch (e) {
+    return undefined;
```
alcuadrado, line 1776: IIRC we decided to throw an error in this case. As this feature will only work for the "standard" chains, the ones included in `@ethereumjs/common`.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -926,6 +928,76 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+
+          vm = new VM({
+            common,
+            activatePrecompiles: true,
+            stateManager: this._vm.stateManager,
+            blockchain: this._vm.blockchain,
+          });
+        }
+
+        const isPreSpuriousDragon = !vm._common.gteHardfork("spuriousDragon");
```
alcuadrado, line 962: I've already forgotten the reason for this. We should add a comment explaning it.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
```diff
@@ -0,0 +1,604 @@
...
+function isStructLog(
+  message: StructLog | DebugMessage | undefined
+): message is StructLog {
+  return message !== undefined && !("structLogs" in message);
+}
+
+const EMPTY_MEMORY_WORD = "0".repeat(64);
+
+export class VMDebugTracer {
+  private _enabled = false;
```
alcuadrado, line 51: I think this enabled/disabled logic isn't necessary for this tracer, as its listeners are injected right before tracing, and removed afterward.

alcuadrado: To be clear, by this i mean using this boolean

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
```diff
@@ -0,0 +1,604 @@
...
+
+    // geth does this for some reason
+    if (result.execResult.exceptionError?.error === "out of gas") {
+      rpcStructLogs[rpcStructLogs.length - 1].error = {};
+    }
+
+    this._lastTrace = {
+      gas: result.gasUsed.toNumber(),
+      failed: result.execResult.exceptionError !== undefined,
+      returnValue: result.execResult.returnValue.toString("hex"),
```
alcuadrado, line 206: Isn't this missing the `"0x"` prefix?

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
```diff
@@ -0,0 +1,604 @@
...
+        callCost
+      );
+
+      gasCost = constantGas + dynamicGas;
+    } else if (step.opcode.name === "CALLCODE") {
+      // finding an existing tx that uses CALLCODE or compiling a contract
+      // so that it uses tihs opcode is hard,
+      // so we just throw
+      throw new InvalidInputError(
+        "Transactions that use CALLCODE are not supported"
```
alcuadrado, line 426:
```suggestion
        "Transactions that use CALLCODE are not supported by Hardhat's debug_traceTransaction"
```

packages/hardhat-core/src/internal/hardhat-network/jsonrpc/client.ts
fvictorio: Yes, this is only used by the test script.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
fvictorio: Changed in 1d866916

packages/hardhat-core/src/internal/hardhat-network/provider/modules/debug.ts
fvictorio: I have no idea why I did this. Fixed in 6822de29e99e6ab617aa2b5b2bfc7a009beda637.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
fvictorio: I still don't understand the difference between them :sweat_smile:  But, anyway, fixed on a19fea87b32607120e01ba82959aad20c8120e51

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
fvictorio: Fixed on b326b89ee34484362a1e03ed34ce553fbbdadedb. I also renamed `getCommon` to `getCommonForTracing`. I'm not sure about the error message.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
fvictorio: Added a comment on 6fae2f732112c14b4830d3558f9027d8e967d8b3. I _think_ that was the reason.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
fvictorio: Fixed in 7d5a5b90165db3bf266d8891b9adcaebc75ecd36

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
fvictorio: Geth doesn't prefix it.

alcuadrado: 0.o

Issue #1293, opened 2021-03-02
Not possible to use nightly Solidity build for compiling


rmeissner:
Currently it is not possible to use nightly Solidity builds for the compile task

Version used: 2.0.10

alcuadrado:
Hey @rmeissner, 

I'm closing this issue in favor of #1426. 

Note that you can use a nightly version like this https://github.com/fvictorio/hardhat-examples/tree/master/custom-solc

PR #1329, opened 2021-03-15
Update downloader.ts

okwme:
as per https://docs.soliditylang.org/en/latest/installing-solidity.html#static-binaries the solc binary references should be updated to https://binaries.soliditylang.org from https://solc-bin.ethereum.org/


fvictorio:
Thanks a lot for this @okwme. I will fix the tests, merge and release this asap

fvictorio:
This should close #1293.

fvictorio:
Closing this since it turned out to be harder than we thought. There are still things to do about this though, they will be tracked in #1426.

PR #1330, opened 2021-03-16
use new smock default branch in docs

robert-e-davidson3:
The smock repo's new default branch is "nu-nu". NPM downloads the version on this branch by default. The usage changed substantially so it took some effort to find correct example code.


alcuadrado:
I'm a bit confused about this. Is this a permanent change?

robert-e-davidson3:
They changed their default branch again so I guess not. Closing since the link is unstable.

PR #1360, opened 2021-03-30
Update docs on supported flavor of signTypedData

spalladino:
Vanilla `eth_signTypedData` was marked as unsupported in https://github.com/nomiclabs/hardhat/pull/1189, in favor of the v4 version. This PR updates the docs to reflect that.


fvictorio:
Thanks, @spalladino. Can you point this to the `website` branch?

spalladino:
Done. Let me know if you want me to rebase as well!

alcuadrado:
Hey @spalladino, thanks for sending this PR. 

We were confused about how the docs were incorrect about this and realized that we accidentally introduced a breaking change. This method should work again soon, but we won't list it in the docs, as it's this EIP is not finalized, so we'd rather have most users using the _v4 variant.

PR #1366, opened 2021-04-02
Cleanup JSON-RPC related types

alcuadrado:
This is a cleanup PR that mostly move things around and fixes a few inconsistencies here and there. I did this in preparation for Berlin, as JSON-RPC related types and validation were becoming a mess.

It does multiple things, all of them related to JSON-RPC stuff:

* Kill `HardhatNetworkProviderError` in favor of using `ProviderError` everywhere for JSON-RPC related errors. 
* Make `ProviderError` a `CustomError`.
* Create common `io-ts` helpers that were previously repeated.
* Move the `io-ts` JSON-RPC validators to core, so that we can use them to validate data in the providers, and not just Hardhat Network.
* Use the validators in the local account providers. This makes it stricter, but that was always the intention with Hardhat. This meant adding `numberToQuantity` in many tests.
* Separate those types in different, more manageable files
* Properly type `validateParams` so that we don't need lots of overloads anymore
* Use the same functions to convert to/from numbers, BN, RPC QUANTITY and RPC DATA everywhere. There were many of those before.
* Align `CallParams` and `TransactionParams` with Geth's implementation of those concepts. This mostly meant that `eth_estimateGas` now uses `CallParams`.
* Clean up how Block Tags where managed. This is the next are the only significant changes. It was very messy before, and EIP-1898 was implemented properly. Sorry for including this here, I found it in the middle of the refactor and couldn't extract it.
* Fixed estimateGas, which was ignoring the block tag when choosing a default nonce.


packages/hardhat-core/src/internal/core/jsonrpc/types/input/blockTag.ts
```diff
@@ -0,0 +1,42 @@
...
+});
+
+export const rpcBlockTagName = t.keyof({
+  earliest: null,
+  latest: null,
+  pending: null,
+});
+
+// This is the new kind of block tag as defined by https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1898.md
+export const rpcNewBlockTag = t.union([
```
alcuadrado, line 22: I split the new, EIP-1989 block tag type and the old one. The reason for this is that Geth doesn't accept the new one everywhere. I checked when each is used and updated the Eth module accordingly.

packages/hardhat-core/src/internal/core/jsonrpc/types/input/filterRequest.ts
```diff
@@ -0,0 +1,27 @@
...
+import { optional } from "../../../../util/io-ts";
+import { rpcHash } from "../base-types";
+
+import { optionalRpcOldBlockTag } from "./blockTag";
+import { optionalRpcLogAddress } from "./logAddress";
+import { optionalRpcLogTopics } from "./logTopics";
+
+export const rpcFilterRequest = t.type(
+  {
+    fromBlock: optionalRpcOldBlockTag,
```
alcuadrado, line 12: This is an example where the old block tags are used.

packages/hardhat-core/src/internal/core/jsonrpc/types/input/hardhat-network.ts
```diff
@@ -0,0 +1,53 @@
+import * as t from "io-ts";
+
+import { optional } from "../../../../util/io-ts";
+import { rpcUnsignedInteger } from "../base-types";
+
+export const rpcForkConfig = optional(
```
alcuadrado, line 6: I put this types here, despite they being hardhat network specific. IDK what should be done instead 🤷‍♂️ We'll probably improve this once we plan how to extract Hardhat Network outside of core.

packages/hardhat-core/src/internal/core/jsonrpc/types/input/solc.ts
```diff
@@ -0,0 +1,22 @@
+import * as t from "io-ts";
+
+export const rpcCompilerInput = t.type(
```
alcuadrado, line 3: Same with Hardhat Network types. This is also specific to HH. Maybe move them to that other file? idk

packages/hardhat-core/src/internal/core/jsonrpc/types/input/subscribeRequest.ts
```diff
@@ -0,0 +1,21 @@
+import * as t from "io-ts";
+
+import { RpcFilterRequest } from "./filterRequest";
+
+// TODO: This types are incorrect. See: https://geth.ethereum.org/docs/rpc/pubsub
```
alcuadrado, line 5: This is very confusing. I even suspect that Geth's documentation is wrong, and our implementation is correct. I prefer to leave a TODO for now.

packages/hardhat-core/src/internal/core/jsonrpc/types/input/validation.ts
```diff
@@ -0,0 +1,75 @@
...
+// tslint:disable only-hardhat-error
+
+/**
+ * This function validates a list of params, throwing InvalidArgumentsError
+ * if the validation fails, and returning their already-parsed types if
+ * the validation succeeds.
+ *
+ * TODO: The type can probably be improved, removing the anys
+ */
+export function validateParams<
```
alcuadrado, line 15: This is some pretty dope TS magic

packages/hardhat-core/src/internal/core/jsonrpc/types/output/decodeJsonRpcResponse.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { InvalidResponseError } from "../../../providers/errors";
+
+/**
+ * This function decodes an RPC out type, throwing InvalidResponseError if it's not valid.
+ */
+export function decodeJsonRpcResponse<T>(value: unknown, codec: t.Type<T>) {
+  return codec.decode(value).fold(() => {
+    // tslint:disable-next-line
+    throw new InvalidResponseError(
```
alcuadrado, line 11: I invented this type of error. I think it makes sense to throw a `ProviderError` if the provider received an incorrect response.

packages/hardhat-core/src/internal/core/providers/accounts.ts
```diff
@@ -53,22 +65,24 @@ export class LocalAccountsProvider extends ProviderWrapperWithChainId {
     const params = this._getParams(args);
 
     if (args.method === "eth_sign") {
-      const [address, data] = params;
-
-      if (address !== undefined) {
-        if (data === undefined) {
-          throw new HardhatError(ERRORS.NETWORK.ETHSIGN_MISSING_DATA_PARAM);
+      if (params.length > 0) {
+        const [address, data] = validateParams(params, rpcAddress, rpcData);
```
alcuadrado, line 69: We validate things using the types here now. The main goal was to make sure that we keep this and Hardhat Network in sync, specially now that there are multiple transaction types.

This is where the API is made sightly stricter. But as I mentioned, Hardhat and Hardhat Network were always meant to be strict.

packages/hardhat-core/src/internal/core/providers/accounts.ts
```diff
@@ -146,57 +160,59 @@ export class LocalAccountsProvider extends ProviderWrapperWithChainId {
...
-    } else {
-      const common = Common.forCustomChain(
-        "mainnet",
-        {
-          chainId,
-          networkId: chainId,
-        },
-        "istanbul"
-      );
+    const common =
```
alcuadrado, line 202: This wasn't really necessary. Made this stylistic change to make printf-debugging it easier.

packages/hardhat-core/src/internal/core/providers/accounts.ts
```diff
@@ -235,6 +251,7 @@ abstract class SenderProvider extends ProviderWrapper {
       method === "eth_call" ||
       method === "eth_estimateGas"
     ) {
+      // TODO: Should we validate this type?
```
alcuadrado, line 254: I didn't validate some the wrappers that don't have local implementations of the methods. These just add some missing params and fwd the request.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -303,15 +314,15 @@ export class EthModule {
...
+    return validateParams(params, rpcCallRequest, optionalRpcNewBlockTag);
   }
 
   private async _callAction(
     rpcCall: RpcCallRequest,
-    blockTag: OptionalBlockTag
+    blockTag: OptionalRpcNewBlockTag
   ): Promise<string> {
-    const blockNumberOrPending = await this._resolveBlockTag(blockTag);
+    const blockNumberOrPending = await this._resolveNewBlockTag(blockTag);
```
alcuadrado, line 325: There's now different methods to resolve the block tags for the new and old kind of blocktags. The main difference is that the new one throws if the block is not found. 

The old one doesn't, but return `undefined`. This is like that because the old one is used in places where you explicitly ask for a value related to that tag, and `null` is returned if it's not found.

The new kind is used just as a context of another operation, like `eth_call`.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -612,18 +622,19 @@ export class EthModule {
...
-      filter.fromBlock = new BN(block.header.number);
-      filter.toBlock = new BN(block.header.number);
+      filter.fromBlock = block.header.number;
+      filter.toBlock = block.header.number;
     }
 
     const [fromBlock, toBlock] = await Promise.all([
-      this._extractBlock(filter.fromBlock),
-      this._extractBlock(filter.toBlock),
+      this._normalizeOldBlockTagForFilterRequest(filter.fromBlock),
```
alcuadrado, line 636: These blocktags aren't resolved, but rather normalized, because the filter-related code is a bit different.

packages/hardhat-core/src/internal/hardhat-network/provider/node-types.ts
```diff
@@ -44,7 +44,7 @@ export interface GenesisAccount {
 }
 
 export interface CallParams {
-  to: Buffer;
+  to?: Buffer;
```
alcuadrado, line 47: This was incorrect. We were using an empty Buffer to signal deployments, and that gave problems with v5.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -444,21 +445,29 @@ export class HardhatNode extends EventEmitter {
...
   ): Promise<EstimateGasResult> {
-    const tx = await this._getFakeTransaction({
-      ...txParams,
-      gasLimit: this.getBlockGasLimit(),
-    });
+    // We get the CallParams and transform it into a TransactionParams to be
+    // able to run it.
+    const txParams = {
+      ...callParams,
+      nonce: await this._getNonce(
```
alcuadrado, line 455: We were using the latest nonce when estimating gas, and ignoring the block tag. I fixed it.

packages/hardhat-core/src/internal/hardhat-network/provider/output.ts
```diff
@@ -78,27 +84,6 @@ export interface RpcLogOutput {
 
 // tslint:disable only-hardhat-error
 
-export function numberToRpcQuantity(n: number | BN): string {
```
alcuadrado, line 81: This is an example of some repeated helpers.

packages/hardhat-core/src/internal/hardhat-network/provider/output.ts
```diff
@@ -115,8 +100,8 @@ export function getRpcBlock(
     parentHash: bufferToRpcData(block.header.parentHash),
     // We pad this to 8 bytes because of a limitation in The Graph
     // See: https://github.com/nomiclabs/hardhat/issues/491
-    nonce: pending ? null : bufferToRpcData(block.header.nonce, 16),
-    mixHash: pending ? null : bufferToRpcData(block.header.mixHash, 64),
+    nonce: pending ? null : bufferToRpcData(block.header.nonce, 8),
```
alcuadrado, line 103: The new helper uses bytes, not chars

packages/hardhat-core/src/internal/util/io-ts.ts
```diff
@@ -0,0 +1,25 @@
+import * as t from "io-ts";
+
+export function optional<TypeT, OutputT, InputT>(
```
alcuadrado, line 3: io-ts repeated helpers. I moved them here.

packages/hardhat-core/test/internal/core/providers/accounts.ts
```diff
@@ -63,9 +64,8 @@ describe("Local accounts provider", () => {
       {
         from: privateKeyToAddress(accounts[0]),
         to: "0x2a97a65d5673a2c61e95ce33cecadf24f654f96d",
-        gasPrice: 0x3b9aca00,
-        nonce: 0x8,
-        chainId: 123,
+        gasPrice: numberToRpcQuantity(0x3b9aca00),
```
alcuadrado, line 67: This is were I had to take into a account that this provider's api is stricter.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -2102,119 +2101,6 @@ describe("Eth module", function () {
           assert.notEqual(logs2[0].address, "changed");
         });
 
-        it("Should accept block hashes as from", async function () {
```
alcuadrado, line 2105: Some tests were removed because this RPC methods don't accept the new kind of blocktag. I checked this with geth's code.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -4465,61 +4355,49 @@ describe("Eth module", function () {
...
-              await this.provider.send("eth_getBalance", [
-                zeroAddress(),
-                {
-                  blockNumber: block.number,
-                  requireCanonical: true,
-                },
-              ]),
-              0
-            );
+        it("should not accept both a blockNumber and requireCanonical", async function () {
```
alcuadrado, line 4364: This was also incorrectly implemented. Added a test.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -1189,11 +1202,15 @@ export class EthModule {
...
+    if ("blockNumber" in tag) {
+      return tag.blockNumber.toString();
+    }
+
+    return bufferToHex(tag.blockHash);
   }
 
   private _extractNormalizedLogTopics(
-    topics: LogTopics
+    topics: OptionalRpcLogTopics
```
alcuadrado, line 1213: This change from non optional to optional may look weird. It's because `logTopic` and `logAddress` already included `undefined`. I changed them to match what the rest of the types do.

alcuadrado:
I self-reviewed it and left a bunch of comments to make reviewing it easier.

packages/hardhat-core/src/internal/core/providers/accounts.ts
```diff
@@ -146,57 +160,59 @@ export class LocalAccountsProvider extends ProviderWrapperWithChainId {
...
+              chainId,
+              networkId: chainId,
+            },
+            "istanbul"
+          );
 
-      transaction = Transaction.fromTxData(txData, { common });
-    }
+    let transaction: TransactionT;
+    transaction = Transaction.fromTxData(txData, { common });
```
fvictorio, line 215:
```suggestion
    const transaction = Transaction.fromTxData(txData, { common });
```

PR #1372, opened 2021-04-05
Support older hard forks

alcuadrado:


packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -188,6 +196,40 @@ export class HardhatNode extends EventEmitter {
     return [common, node];
   }
 
+  private static _validateHardforks(
```
alcuadrado, line 199: We validate this here because:
1. The config validation can't make requests and fetch the remote chainId.
2. The `node` task can provide a different block number so the config validation won't be enough.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -188,6 +196,40 @@ export class HardhatNode extends EventEmitter {
...
+    localHardfork: string,
+    forkBlockNumber: number | undefined,
+    common: Common,
+    remoteChainId: number
+  ): void {
+    if (!common.gteHardfork("spuriousDragon")) {
+      throw new InternalError(
```
alcuadrado, line 206: Throwing `ProviderError`s because this are thrown from a request, as the `HardhatNode` instance is created on the first RPC request.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -4506,3 +4508,47 @@ describe("Eth module", function () {
     });
   });
 });
+
+describe("Eth module - special tests", function () {
```
alcuadrado, line 4512: Created this other tests as they depend on fixture projects, and not the `PROVIDERS` array.

packages/hardhat-core/test/internal/hardhat-network/provider/hardhat-network-options.ts
```diff
@@ -83,4 +85,58 @@ describe("Hardhat Network special options", function () {
       assert.isBelow(+secondBlock.timestamp, initialDate + 20);
     });
   });
+
+  describe("hardfork", function () {
+    describe("When not forking", function () {
+      useFixtureProject("hardhat-network-spurious-dragon");
+      useEnvironment();
+
+      it("should accept hard forks as late as chainnstart and use them", async function () {
```
fvictorio, line 94: `chainstart`?

packages/hardhat-core/test/internal/hardhat-network/provider/hardhat-network-options.ts
```diff
@@ -83,4 +85,58 @@ describe("Hardhat Network special options", function () {
...
+        const [sender] = await this.env.network.provider.send("eth_accounts");
+        await assert.isRejected(
+          this.env.network.provider.send("eth_sendTransaction", [
+            {
+              from: sender,
+              // This is a deployment with a constructor that just executes CHAIN_ID
+              // which as added in Istanbul
```
fvictorio, line 101:
```suggestion
              // which was added in Istanbul
```

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -188,6 +196,40 @@ export class HardhatNode extends EventEmitter {
...
+The hard fork must be at least spuriousDragon, but ${localHardfork} was given.`
+      );
+    }
+
+    if (forkBlockNumber !== undefined) {
+      let upstreamCommon: Common;
+      try {
+        upstreamCommon = new Common({ chain: remoteChainId });
+      } catch (error) {
+        return;
```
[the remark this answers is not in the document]
alcuadrado: Added a comment explaining it.

packages/hardhat-core/test/internal/hardhat-network/provider/hardhat-network-options.ts
alcuadrado: That's the first hardfork according to `@ethereumjs/common`.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -188,6 +196,40 @@ export class HardhatNode extends EventEmitter {
     return [common, node];
   }
 
+  private static _validateHardforks(
+    localHardfork: string,
```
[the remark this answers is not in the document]
alcuadrado: right

packages/hardhat-core/test/internal/hardhat-network/provider/hardhat-network-options.ts
alcuadrado: ohh there was a typo

PR #1373, opened 2021-04-06
Add ethernal plugin

antoinedc:
I think I somehow messed up with the rebase, so easier to just do another PR...


antoinedc:
Ping @alcuadrado :) is this ok now?

PR #1376, opened 2021-04-06
Add test block with EIP-2930 tx

fvictorio:
- Add a full-block test for a Ropsten block that includes an EIP-2930 tx.
- Add optional properties to the `RpcTransaction` type
- Implement `EIP2929StateManager` interface in `ForkStateManager`


packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
```diff
@@ -371,6 +374,69 @@ export class ForkStateManager implements StateManager {
     return value;
   }
 
+  // the following methods are copied verbatim from
+  // DefaultStateManager
+
+  public isWarmedAddress(address: Buffer): boolean {
```
alcuadrado, line 380: What about doing `return DefaultStateManager.prototype.isWarmedAddress.call(this, address)` instead? The same for the rest of the methods.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -47,7 +47,7 @@ interface ForkPoint {
    */
```
alcuadrado, line 47: I think this comment is outdated now. Maybe rename `blockNumber` to something more explicit like `blockToRun`?

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -650,15 +654,15 @@ describe("HardhatNode", () => {
...
-          afterBlockEvent
-        );
+        // Restore the receipt trie
+        (block as any).header.receiptTrie = modifiedBlock.header.receiptTrie;
 
-        const newBlock = await forkedNode.getBlockByNumber(
-          new BN(blockNumber + 1)
-        );
+        // TODO we should use modifiedBlock instead of block here,
+        // but we can't because of a bug in the vm
```
alcuadrado, line 661:
```suggestion
        // but we can't because of a bug in the vm
        // TODO: Change this once https://github.com/ethereumjs/ethereumjs-monorepo/pull/1185 is released.
```

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
alcuadrado: `ForkPoint` may also be outdated.

alcuadrado:
I left a few small comments, but LGTM

packages/hardhat-core/src/internal/hardhat-network/provider/fork/rpcToTxData.ts
```diff
@@ -14,4 +16,19 @@ export function rpcToTxData(rpcTransaction: RpcTransaction): TxData {
...
   };
+
+  if (rpcTransaction.type !== undefined) {
+    txData.type = rpcTransaction.type;
+  }
+  if (rpcTransaction.chainId !== undefined) {
+    txData.chainId = rpcTransaction.chainId;
+  }
+  if (rpcTransaction.accessList !== undefined) {
+    txData.accessList = rpcTransaction.accessList.map((item) => [
```
alcuadrado, line 27: Can't we use the other accessList format in `AccessListEIP2930TxData`? It's closer to the representation used everywhere else.

```ts
export type AccessList = AccessListItem[]
export type AccessListItem = {
  address: PrefixedHexString
  storageKeys: PrefixedHexString[]
}
```
```

File: packages/hardhat-core/src/internal/core/jsonrpc/types/output/transaction.ts [hunk 5]
```diff
@@ -3,6 +3,13 @@ import * as t from "io-ts";
 import { nullable, optional } from "../../../../util/io-ts";
 import { rpcAddress, rpcData, rpcHash, rpcQuantity } from "../base-types";
 
+const rpcAccessListItem = t.type({
```
alcuadrado, line 6: These are shared by input and out types. Maybe we should move them `base-types.ts`? Not 100% convinced either.

alcuadrado: Oh, this is also repeated between transaction and call requests. Maybe it's not a base type, but could go at the same level.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
fvictorio: Good idea, I'll give it a try.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/rpcToTxData.ts
fvictorio: I forgot to add a comment about this. The other format uses `string` instead of `Buffer`, so I used this one to avoid adding a new `io-ts` codec for these strings. But we can do that, or (maybe better) we can just transform them here.

packages/hardhat-core/src/internal/core/jsonrpc/types/output/transaction.ts
fvictorio: I thought about adding it to `base-types.ts`, but it felt very different to those types. I think a `access-list-types.ts` module at the same level as `base-types.ts` should work, wdyt?

alcuadrado: yeah, let's do that. I'll move them in my next pr

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
fvictorio: Done. Notice that this adds an implicit dependency between those method calls and certain instance variables that have to exist (`_accessedStorage` and `_accessedStorageReverted`). `tsc` doesn't seem to catch those errors.

alcuadrado: Maybe leave a comment? Either thing is dangerous in its own way

fvictorio: There is one:

```typescript
  // used by the DefaultStateManager calls
  private _accessedStorage: Array<Map<string, Set<string>>> = [new Map()];
  private _accessedStorageReverted: Array<Map<string, Set<string>>> = [
    new Map(),
  ];
```

PR #1377, opened 2021-04-06
Workaround transactions recreation issue

alcuadrado:


packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1077,6 +1080,15 @@ Hardhat Network's forking functionality only works with blocks from at least spu
 
       const block = await blockBuilder.build();
 
+      // We replace the block's transactions with the actual ones,
+      // as the block builder recreates them, turning fake transactions
+      // into real ones.
+      //
+      // IMPORTANT: this workaround only works because while BlockBuilder#addTransaction
+      // recreates the transactions you pass it, it actually runs yours.
+      block.transactions.splice(0, transactions.length);
```
fvictorio, line 1089: Just in case, this is equivalent to `.splice(0)`. Not sure which one is more idiomatic (I'd prefer to drop the second argument, since it makes it clearer in my mind that it's deleting everything, plus is similar to the `.slice(0)` idiom).

alcuadrado: Changed it. I have no preference.

fvictorio (approved):

PR #1381, opened 2021-04-07
Add ReadOnlyValidEIP2930Transaction

fvictorio:


packages/hardhat-core/src/internal/hardhat-network/provider/transactions/ReadOnlyValidEIP2930Transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+ * This class is like `ReadOnlyValidTransaction` but for
+ * EIP-2930 (access list) transactions.
+ */
+export class ReadOnlyValidEIP2930Transaction extends AccessListEIP2930Transaction {
+  public static fromTxData(
+    txData: AccessListEIP2930TxData,
+    opts?: TxOptions
+  ): never {
+    throw new InternalError(
+      "`fromTxData` is not implemented in ReadOnlyValidTransaction"
```
alcuadrado, line 24: These comments are outdated.

packages/hardhat-core/src/internal/hardhat-network/provider/transactions/ReadOnlyValidEIP2930Transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+    );
+  }
+}
+
+// Override private methods
+
+const ReadOnlyValidTransactionPrototype: any =
+  ReadOnlyValidEIP2930Transaction.prototype;
+
+ReadOnlyValidTransactionPrototype._validateTxV = function () {};
```
alcuadrado, line 154: This shouldn't make sense here. There's no v validation. I think there's no need for any of them.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
```diff
@@ -268,10 +269,18 @@ export class ForkBlockchain implements HardhatBlockchainInterface {
     });
 
     for (const transaction of rpcBlock.transactions) {
-      const tx = new ReadOnlyValidTransaction(
-        new Address(transaction.from),
-        rpcToTxData(transaction)
-      );
+      let tx;
+      if (transaction.accessList !== undefined) {
```
alcuadrado, line 273:
```suggestion
      // TODO: Check the type instead
      if (transaction.accessList !== undefined) {
```

alcuadrado: We can't do this in this branch yet.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -670,8 +660,8 @@ describe("HardhatNode", () => {
 
         // We do some manual comparisons here to understand why the root of the receipt tries differ.
         if (localReceiptRoot !== remoteReceiptRoot) {
-          for (let i = 0; i < block.transactions.length; i++) {
-            const tx = block.transactions[i];
+          for (let i = 0; i < newBlock.transactions.length; i++) {
+            const tx = newBlock.transactions[i];
```
fvictorio, line 664: I think the previous version was wrong, we should use `newBlock` here (it wasn't a big issue since this is only used when the receipts root don't match)

PR #1382, opened 2021-04-07
Update typescript readme to have better example

esemeniuc:
This example is closer to the sample code


fvictorio:
Hey @esemeniuc, sorry for not responding before. I think the current examples are fine, I don't think it's necessary to match the sample project.

Plus, these snippets are wrong: you shouldn't import plugins in your tests/scritps, they are supposed to be imported only from the config.

PR #1388, opened 2021-04-09
correct typo (buidling -> building)

manuelbarzi:


fvictorio:
I think this is intentional from the buidler days :sweat_smile:

alcuadrado:
Yes, that's the reason! Thanks for submitting the PR anyway :)

PR #1392, opened 2021-04-10
Smaller downloader fix

### LINKED ISSUE #1332: HH503 due to MacOS solc binaries having changed
```
Error HH503: Couldn't download compiler version 0.5.17. Checksum verification failed. Please check your connection.
```

Likely due to: https://github.com/ethereum/solc-bin/pull/76

Also see this PR https://github.com/crytic/echidna/pull/629

alcuadrado:
This PR fixes #1332 and replaces #1329.

The difference with this PR is that it doesn't add support for nightly versions, but just fixes #1332 and updates the compilers repository url.

The implementation in #1329 lead to some really slow startup times (>10sec) in some cases, so we should redesign it. I created it so that we can push this fix in the next release, and later add support for the nightlies.

### DIFF
--- a/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
+++ b/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
@@ -39,7 +39,7 @@ export interface CompilersList {
 
 const log = debug("hardhat:core:solidity:downloader");
 
-const COMPILER_FILES_DIR_URL_SOLC = "https://solc-bin.ethereum.org/";
+const COMPILER_FILES_DIR_URL_SOLC = "https://binaries.soliditylang.org/";
 
 async function downloadFile(
   url: string,
--- a/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
+++ b/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
@@ -92,6 +92,7 @@ export class CompilerDownloader {
 
     if (expectedKeccak256 !== compilerKeccak256) {
       await fsExtra.unlink(downloadedFilePath);
+      await fsExtra.unlink(this.getCompilersListPath(compilerBuild.platform));
 
       throw new HardhatError(ERRORS.SOLC.INVALID_DOWNLOAD, {
         remoteVersion: compilerBuild.version,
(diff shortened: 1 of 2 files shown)


fvictorio (approved):

cameel:
Hello. I saw there was some confusion regarding the solc-bin domain with people linking to this PR so I just wanted to explain the situation things for anyone who comes here in the future.

First, the domain change is not a part of the fix there. The fix is to redownload the file list if the binary does not match the checksum. That's on the assumption that the list might be corrupted but it also resolves the breakage caused recently by the faulty macOS binaries being rebuilt and replaced due to https://github.com/ethereum/solidity/issues/10183. The domain change is just a cosmetic thing because both https://solc-bin.ethereum.org and https://binaries.soliditylang.org point at the same exact host and both are supported long-term.

As for why there are two domains, here's the full story:
- Originally the Solidity binaries were hosted at https://ethereum.github.io/solc-bin/
- https://solc-bin.ethereum.org was a redirect pointing at it. We meant people to use the `ethereum.org` one but this was not communicated clearly and many projects were using `github.io` directly instead. Embarrassingly, even our own [solc-js](https://github.com/ethereum/solc-js) used it.
- The `github.io` hosting has a 10 GB size limit and we exceeded it so we had to switch to a different one. For people who were using the `ethereum.org` domain the transition should have been seamless but there was nothing we could do about `github.io` and we had to just deprecate it. We left it up but it hosts binaries only up to 0.7.2 and does not have the native binaries.
- Around the same time we got `soliditylang.org` domain and started moving all Solidity-related stuff under it. That's when the idea to add https://binaries.soliditylang.org came about. But this is just a cosmetic change - another domain pointing at the same thing. It was [documented](https://docs.soliditylang.org/en/latest/installing-solidity.html#static-binaries) but not widely announced because it's just fine to keep using https://solc-bin.ethereum.org (it still is and will be fully supported) so there was no point bothering people people to switch again. The new URL is recommended but no one has to change it just for the sake of it.

alcuadrado:
Hey, you are right, @cameel! the problem reported on telegram must have been because of the checksums. Thanks for chiming in!

PR #1402, opened 2021-04-14
link directly to the discord channel

Huge:
minor but nice a change


alcuadrado:
Hey @Huge, thanks for your PR. What's the benefit of linking like that? 

I only see a few disadvantages:

* That link is not really nice
* Invites are kind of messy to handle in discord
* Changing the is harder when you share it directly
* You can't memorize it nor pronounce it

I'm curious about your take, as I may as well be missing something.

fvictorio:
Closing this for now, please let us know if you still think there's a good reason to do this.

PR #1405, opened 2021-04-14
Fix last letter lib cut out in error msgs

Bidon15:


fvictorio:
Thank you for this @Bidon15! I just added some tests that trigger this bug, but otherwise LGTM.

fvictorio (approved):

PR #1413, opened 2021-04-20
prefix package name

kangarang:
prefix `hardhat-waffle` with `@nomiclabs` to be more explicit

[related](https://mobile.twitter.com/HardhatHQ/status/1362866790695641094)


fvictorio (approved):
Makes sense to me

PR #1416, opened 2021-04-21
Add output sources to artifact

conspyrosy:
Regarding the following issue: https://github.com/nomiclabs/hardhat/issues/1414


fvictorio:
Hey @conspyrosy, thank you for this. The original use case for this was adding the devDoc, and that turned out to be possible.

The reason we don't include the full output is that an artifact should only have information about a contract. If someone wants something beyond that (e.g., the ast, which is information about the *whole* source file), they should use the build info.

We are not 100% against this, but we want to understand better the driving use cases before doing this fix.

I'm going to close this for now, but thanks for contributing!

PR #1417, opened 2021-04-21
Add instructions for supporting typescript path mappings

spalladino:


fvictorio:
Thank you @spalladino!

aspiers:
Is this supposed to work for mapped `import` statements within `hardhat.config.ts`? Because no matter what I try, I cannot get it to work: https://discord.com/channels/750408878008827925/750408878008827928/906579332460331088

alcuadrado:
I don't think this works for imports within your config, as you are registering the module when the config is loaded.

You could use something like `node --require ts-node/register --require tsconfig-paths/register node_modules/.bin/hardhat`

aspiers:
Thanks for the reply!  I feared that might be the case.  What about imports within files imported from the config - I guess it's the same problem?  In which case this cannot be used for tasks, only scripts and tests?

alcuadrado:
if you add the import at the top of the file it will work

aspiers:
I added `import 'tsconfig-paths/register';` to the top of both `hardhat.config.ts` and the `tasks/mytask.ts` file.

I then added

```json
    "paths": {
      "magicpath": ["lib"]
    },
    "baseUrl": ".",
```

to the `compilerOptions` section of `tsconfig.json`, and also

    import { sesame } from 'magicpath';

to `tasks/mytask.ts`.

But even when running `node -r ts-node/register -r tsconfig-paths/register node_modules/.bin/hardhat`, I get:

    tasks/mytask.ts(7,24): error TS2307: Cannot find module 'magicpath' or its corresponding type declarations.

PR #1418, opened 2021-04-21
Fix link for Ropsten faucet

astronomotrous:


fvictorio:
Thank you @astronomotrous!

PR #1419, opened 2021-04-21
Fix issue with `selfdestruct`

### LINKED ISSUE #1415: Removal of contract code (selfdestruct) was not reflected in hardhat 2.2.0
I upgraded `hardhat` from  `2.0.6` to `2.2.0` and ran into a problem where transaction that triggers a contract to `selfdestruct` succeeded (`receipt.status == 1`) but querying the dead contract's code still shows that the contract contains code (`ethers.provider.getCode(CONTRACT_ADDRESS)`).
Contract balance can be correctly updated however, but not the contract code.

To reproduce:
- contract code (modified from `Greeter` example)
```solidity
//SPDX-License-Identifier: Unlicense
pragma solidity ^0.7.0;

contract Greeter {

  constructor() payable {}

  function kill() public {
    selfdestruct(msg.sender);
  }

}
```

- test
```javascript
const { expect } = require("chai");

describe("Greeter", function() {
  it("Should successfully update contract state after selfdestruct", async function() {
    const Greeter = await ethers.getContractFactory("Greeter");
    const greeter = await Greeter.deploy({value: 1000});
    await greeter.deployed();

    expect(await ethers.provider.getBalance(greeter.address)).to.equal(1000)
    expect(await ethers.provider.getCode(greeter.address)).to.not.equal("0x")
    await greeter.kill();
    expect(await ethers.provider.getBalance(greeter.address)).to.equal(0)
    expect(await ethers.provider.getCode(greeter.address)).to.equal("0x")
  });
});
```

Then install different versions of `hardhat` and run the test. So far I've tried `2.0.6`, `2.0.10`, `2.1.0`, `2.1.2` and they all work fine. Only `2.2.0` fail.

fvictorio:
Fixes #1415.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
@@ -18,7 +18,11 @@ import { JsonRpcClient } from "../../jsonrpc/client";
 import { GenesisAccount } from "../node-types";
 import { makeAccount } from "../utils/makeAccount";
 
-import { AccountState, makeAccountState } from "./AccountState";
+import {
+  AccountState,
+  makeAccountState,
+  makeEmptyAccountState,
+} from "./AccountState";
 import { randomHash } from "./random";
 
 const encodeStorageKey = (address: Buffer, position: Buffer): string => {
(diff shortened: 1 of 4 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
```diff
@@ -355,7 +359,10 @@ export class ForkStateManager implements EIP2929StateManager {
   }
 
   public async deleteAccount(address: Address): Promise<void> {
-    this._state.delete(address.toString());
+    // we set an empty account instead of deleting it to avoid
+    // re-fetching the state from the remote node
```
alcuadrado, line 363:
```suggestion
    // re-fetching the state from the remote node.
    // This is only valid post spurious dragon, but we don't support older hardforks when forking.
```

alcuadrado:
LGT. Just left a minor comment.

PR #1429, opened 2021-04-29
Suggestion: Update mocha to 8.x

mjlescano:
The new Mocha versions adds the ability to run tests in parallel and some other niceties, but, it includes some minor breaking changes, so it probably is going to need a major release.

## Braking Changes

Breaking changes of `mocha@8` can be reviewed here: https://github.com/mochajs/mocha/releases/tag/v8.0.0

### Things to keep in mind
* Mocha drops node 8 compatibility.
* People using this lib will update their used Mocha version when using the `npx hardhat test` task.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "@types/shelljs": "^0.8.6",
     "chai": "^4.2.0",
     "ganache-cli": "^6.12.0",
-    "mocha": "^7.1.2",
+    "mocha": "^8.3.2",
     "nyc": "^14.1.0",
     "prettier": "2.0.5",
     "rimraf": "^3.0.2",
--- a/packages/hardhat-core/package.json
+++ b/packages/hardhat-core/package.json
@@ -106,7 +106,7 @@
     "lodash": "^4.17.11",
     "merkle-patricia-tree": "^4.1.0",
     "mnemonist": "^0.38.0",
-    "mocha": "^7.1.2",
+    "mocha": "^8.3.2",
     "node-fetch": "^2.6.0",
     "qs": "^6.7.0",
     "raw-body": "^2.4.1",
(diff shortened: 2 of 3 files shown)


alcuadrado:
That's a good point @justinjmoses. I'm always a bit hesitant about upgrading mocha, as it may break some test suites. Maybe we should document how this is handled.

TBH mocha should be part of a plugin. It would make upgrading it easier, as users could upgrade Hardhat and stay in the same version of mocha.

mjlescano:
@alcuadrado, what about if mocha is defined as a `peerDependency` on the `package.json`, we could require it as `> 7.0.0`, but still let the user use their desired mocha version.

It would still be a breaking change for hardhat because Mocha is not going to be available anymore if the user didn't install it, but the migration would be really easy for them because they just have to run `npm install mocha@^7.1.2`.

alcuadrado:
@cgewecke would this break things on your plugins?

fvictorio:
Hey @mjlescano, thanks again for this PR. We talked about this and we feel it's risky to merge this since we aren't sure about which things it could potentially break. Instead, we think that this could be done via a "hardhat-mocha8" plugin.

If I'm not missing anything, the plugin would be quite simple: you just need to have mocha@8 as a dependency on it, and override the `TASK_TEST_RUN_MOCHA_TESTS` subtask so that it uses that dependency instead. I believe you would only need to copy the code of that subtask in our code, and check that the way mocha is used programmatically has not changed in the latest version.

I'm going to close this now, but if you are interested in doing a plugin like that and need assistance, please let me know!

mjlescano:
Hi @fvictorio!, yes, we actually did that [here](https://github.com/Synthetixio/synthetix/pull/1272/files#diff-704a6f98449e3c6d121baf9e07acb6684186b212595f7ee32191f271ce09742eR7-R19) :), sorry I didn't update about it here... but, for now, I think is best if we leave it inside our task for clarity. Thank you!

PR #1435, opened 2021-05-02
RPC methods for arbitrary account modifications

feuGeneA:
Written to conform to the spec at https://www.notion.so/RPC-methods-for-arbitrary-account-modifications-8694980326e34622a56bd76e4450c45c

Each commit is written to be complete and independent.  Consider reviewing each one separately in order to reduce the cognitive load.

-   [x] Implement high-level tests as described in [this comment](https://github.com/nomiclabs/hardhat/pull/1435#issuecomment-839146519).
    - implemented in https://github.com/nomiclabs/hardhat/pull/1435/commits/dcb48713eca59df00cd6e031ef8f9d8620718c34
-   [x] Fix failing test that @fvictorio added.
    - implemented in https://github.com/nomiclabs/hardhat/pull/1435/commits/d2dcf0c9203a45fa4917b253a8f73b02e0d7ae49
-   [x] Address https://github.com/nomiclabs/hardhat/pull/1435#discussion_r646886224
-   [x] Address https://github.com/nomiclabs/hardhat/pull/1435#discussion_r646887464


packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +369,374 @@ describe("Hardhat module", function () {
...
+          // Arrange 1: Send a transaction, in order to ensure a pre-existing
+          // state root.
+          await this.provider.send("eth_sendTransaction", [
+            {
+              from: DEFAULT_ACCOUNTS_ADDRESSES[0],
+              to: DEFAULT_ACCOUNTS_ADDRESSES[1],
+              value: "0x100",
+            },
+          ]);
+          await this.provider.send("evm_mine");
```
feuGeneA, line 441: Is this necessary here? Or is automine enabled?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +369,374 @@ describe("Hardhat module", function () {
...
```
feuGeneA, line 513: Is this necessary here? Or is automine enabled?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +369,374 @@ describe("Hardhat module", function () {
...
+        it("should throw an InvalidInputError if new nonce is smaller than the current nonce", async function () {
+          // Arrange: Send a transaction, in order to ensure a non-zero nonce.
+          await this.provider.send("eth_sendTransaction", [
+            {
+              from: DEFAULT_ACCOUNTS_ADDRESSES[0],
+              to: DEFAULT_ACCOUNTS_ADDRESSES[1],
+              value: "0x100",
+            },
+          ]);
+          await this.provider.send("evm_mine");
```
feuGeneA, line 569: Is this necessary here? Or is automine enabled?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +369,374 @@ describe("Hardhat module", function () {
...
+        it("should result in a modified nonce", async function () {
+          // Arrange: Send a transaction, in order to ensure a non-zero nonce.
+          await this.provider.send("eth_sendTransaction", [
+            {
+              from: DEFAULT_ACCOUNTS_ADDRESSES[0],
+              to: DEFAULT_ACCOUNTS_ADDRESSES[1],
+              value: "0x100",
+            },
+          ]);
+          await this.provider.send("evm_mine");
```
feuGeneA, line 589: Is this necessary here? Or is automine enabled?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +369,374 @@ describe("Hardhat module", function () {
...
+        it("should not result in a modified state root", async function () {
+          // Arrange 1: Send a transaction, in order to ensure a non-zero nonce.
+          await this.provider.send("eth_sendTransaction", [
+            {
+              from: DEFAULT_ACCOUNTS_ADDRESSES[0],
+              to: DEFAULT_ACCOUNTS_ADDRESSES[1],
+              value: "0x100",
+            },
+          ]);
+          await this.provider.send("evm_mine");
```
feuGeneA, line 615: Is this necessary here? Or is automine enabled?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +369,374 @@ describe("Hardhat module", function () {
...
+          // Arrange 1: Send a transaction, in order to ensure a pre-existing
+          // state root.
+          await this.provider.send("eth_sendTransaction", [
+            {
+              from: DEFAULT_ACCOUNTS_ADDRESSES[0],
+              to: DEFAULT_ACCOUNTS_ADDRESSES[1],
+              value: "0x100",
+            },
+          ]);
+          await this.provider.send("evm_mine");
```
feuGeneA, line 719: Is this necessary here? Or is automine enabled?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +369,374 @@ describe("Hardhat module", function () {
...
+                DEFAULT_ACCOUNTS_ADDRESSES[0],
+              ])
+            ).replace("0x", ""),
+            "hex"
+          );
+
+          // Act: Set the new balance.
+          const targetBalance = existingBalance.add(new BN(1)).mul(new BN(2));
+          // For sanity, ensure that we really are making a change:
+          assert.notEqual(targetBalance, existingBalance);
```
fvictorio, line 413: This is always true, because `targetBalance` and `existingBalance` are objects. (This might be valid when you use a chai plugin for BNs, but we don't have one yet.)

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +369,374 @@ describe("Hardhat module", function () {
...
+        it("should reject an invalid address", async function () {
+          await assertInvalidArgumentsError(
+            this.provider,
+            "hardhat_setCode",
+            ["0x1234", "0x0"],
+            'Errors encountered in param 0: Invalid value "0x1234" supplied to : ADDRESS'
+          );
+        });
+
+        it("should reject a non-numeric nonce", async function () {
```
fvictorio, line 472: This description seems wrong

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
alcuadrado: Yeah, automine is the default. No need to do this.

feuGeneA: Resolved via https://github.com/nomiclabs/hardhat/pull/1435/commits/5ef98460a0752336f3c134e48e958411f0144c62

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Resolved via https://github.com/nomiclabs/hardhat/pull/1435/commits/c07dabed71f8d2f57a8703652b43cf609ff0bc39

feuGeneA:
On Slack, @alcuadrado made the following change request:

> Something that we normally do is add a few higher level integration tests
>
> For example, we would add these:
>
> - Check that after using hardhat_setNonce you can send a tx with it.
> - Check that set balance let you actually fund an account. Probably by funding a non-existing account and sending a tx with it.
> - Set the code of an empty account and validate that it's actually executed when sending a tx to it.
> - Change a contract's code and check that the new one is executed.
> - Overwrite a contract's storage slot and check that executing an eth_call reads the new value.
>
> do you think you can add them?

packages/hardhat-core/src/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -226,6 +234,39 @@ export class HardhatModule {
...
+    const MAX_WORD_VALUE = new BN(2).pow(new BN(256));
+    if (validatedParams[1].gt(MAX_WORD_VALUE)) {
+      throw new InvalidInputError("Storage key must not be greater than 2^256");
+    }
+
+    if (validatedParams[2].length !== 32) {
+      throw new InvalidInputError(
+        "Storage value must be exactly 32 bytes long"
+      );
+    }
```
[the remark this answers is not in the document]
fvictorio: Hmm, maybe? But it seems like they are intentionally not checking that (plus it would be a breaking change), so I don't know if they'd be willing to do it.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +369,627 @@ describe("Hardhat module", function () {
...
+          // Assert: Check that the balance is the one we set
+          assert.equal(balancePreviousBlock, targetBalanceHex);
+        });
+
+        it("should fund an account and permit that account to send a transaction", async function () {
+          // Arrange: Fund a not-yet-existing account.
+          const notYetExistingAccount =
+            "0x1234567890123456789012345678901234567890";
+          const amountToBeSent = new BN(10);
+          const gasRequired = new BN("48000000000000000");
```
feuGeneA, line 492: Is there a constant that could be used here instead?

fvictorio: Is this the gas limit or the gas price? It seems a huge value to be the gas limit. If it's the gas price, you can probably omit it.

feuGeneA: Honestly I'm not sure.  I only stuck it in here because, without it, I get this error message: `InvalidInputError: sender doesn't have enough funds to send tx. The upfront cost is: 48000000000000010 and the sender's account only has: 10`.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +372,625 @@ describe("Hardhat module", function () {
...
+            "hex"
+          );
+
+          // Act: Set the new balance.
+          const targetBalance = existingBalance.add(new BN(1)).mul(new BN(2));
+          // For sanity, ensure that we really are making a change:
+          assert.notDeepEqual(targetBalance, existingBalance);
+          await this.provider.send("hardhat_setBalance", [
+            DEFAULT_ACCOUNTS_ADDRESSES[0],
+            `0x${targetBalance.toString(16)}`,
```
fvictorio, line 419: I think you can use the `numberToRpcQuantity` helper for things like this.

Similarly, there is a `rpcQuantityToBN` helper that you can use for the result of `eth_getBalance`.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
fvictorio: Ok, I understand this better now.

I think using an arbitrary big number like these is fine. But if you want to be more in control of what's going on, you can do this:

1) Use 21000 as the `gasRequired` here
2) When sending the transaction, set the `gas` value to 21000 and the `gasPrice` to 1
3) If you do this, then the resulting balance of `notYetExistingAccount` should be 0 again. You can assert that too if you want.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/9b5c3b3738538c16a6daf160b0d82dec27c78915 (and in the subsequent correction in https://github.com/nomiclabs/hardhat/pull/1435/commits/fe486a3deed2caa568c19dacc33596417dc94c18)

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +374,650 @@ describe("Hardhat module", function () {
...
+            ["0x1234", "0x0"],
+            'Errors encountered in param 0: Invalid value "0x1234" supplied to : ADDRESS'
+          );
+        });
+
+        it("should reject a non-numeric balance", async function () {
+          await assertInvalidArgumentsError(
+            this.provider,
+            "hardhat_setBalance",
+            [DEFAULT_ACCOUNTS_ADDRESSES[0].toString(), "xyz"],
```
fvictorio, line 392:
```suggestion
            [DEFAULT_ACCOUNTS_ADDRESSES[0], "xyz"],
```

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +374,650 @@ describe("Hardhat module", function () {
...
+            'Errors encountered in param 1: Invalid value "xyz" supplied to : QUANTITY'
+          );
+        });
+
+        it("should not reject valid argument types", async function () {
+          await this.provider.send("hardhat_setBalance", [
+            DEFAULT_ACCOUNTS_ADDRESSES[0].toString(),
```
fvictorio, line 399:
```suggestion
            DEFAULT_ACCOUNTS_ADDRESSES[0],
```

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +374,650 @@ describe("Hardhat module", function () {
...
+          const existingBalance = rpcQuantityToBN(
+            await this.provider.send("eth_getBalance", [
+              DEFAULT_ACCOUNTS_ADDRESSES[0],
+            ])
+          );
+
+          // Act: Set the new balance.
+          const targetBalance = existingBalance.add(new BN(1)).mul(new BN(2));
+          // For sanity, ensure that we really are making a change:
+          assert.notDeepEqual(targetBalance, existingBalance);
```
fvictorio, line 415: I would avoid making deep equal comparisons of BN objects. It's probably fine for now, but it seems somewhat brittle to me. Instead, I'd use `assert.isFalse(targetBalance.eq(existingBalance))`

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +374,650 @@ describe("Hardhat module", function () {
...
+
+          // Arrange 2: Capture the existing state root.
+          const oldStateRoot = (
+            await this.provider.send("eth_getBlockByNumber", ["latest", false])
+          ).stateRoot;
+
+          // Act: Set the new balance.
+          await this.provider.send("hardhat_setBalance", [
+            DEFAULT_ACCOUNTS_ADDRESSES[0],
+            intToHex(99),
```
fvictorio, line 449: For consistency, I would use `numberToRpcQuantity` here. (We don't use `intToHex` anywhere else in the code as far as I can tell)

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +374,650 @@ describe("Hardhat module", function () {
...
+        let abiEncoder: ethers.utils.Interface;
+        before(async function () {
+          [
+            ,
+            {
+              contracts: {
+                ["literal.sol"]: { Nine: contractNine },
+              },
+            },
+          ] = await compileLiteral(`
```
fvictorio, line 579: Personal preference: I would avoid destructuring this. I'd rather do:

```typescript
          const [, compilerOutput] = await compileLiteral(`
            contract Nine {
                function returnNine() public pure returns (int) { return 9; }
            }`);
          contractNine = compilerOutput.contracts["literal.sol"].Nine;
```

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +374,650 @@ describe("Hardhat module", function () {
...
+                to: DEFAULT_ACCOUNTS_ADDRESSES[0],
+                data: abiEncoder.encodeFunctionData("returnNine", []),
+              },
+              "latest",
+            ]),
+            abiEncoder.encodeFunctionResult("returnNine", [9])
+          );
+        });
+
+        it("should, when setting code on an existing contract account, result in code that can actually be executed", async function () {
```
fvictorio, line 668: I like this test a lot :smile:

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -931,6 +944,49 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+    newCode: Buffer
+  ): Promise<void> {
+    await this._stateManager.putContractCode(address, newCode);
+  }
+
+  public async setAccountNonce(address: Address, newNonce: BN): Promise<void> {
+    const account = await this._stateManager.getAccount(address);
+    if (newNonce.lt(account.nonce)) {
+      throw new InvalidInputError(
+        "New nonce must not be smaller than the existing nonce"
```
fvictorio, line 971: Let's add the current nonce and the used nonce as part of the message, that's usually super helpful. Something like: "New nonce (3) must not be smaller than the existing nonce (5)"

packages/hardhat-core/src/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -180,6 +201,72 @@ export class HardhatModule {
...
+    const validatedParams = validateParams(
+      params,
+      rpcAddress,
+      rpcQuantity,
+      rpcData
+    );
+
+    const MAX_WORD_VALUE = new BN(2).pow(new BN(256));
+    if (validatedParams[1].gt(MAX_WORD_VALUE)) {
+      throw new InvalidInputError("Storage key must not be greater than 2^256");
```
fvictorio, line 249: We should show the value used in the error message.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -180,6 +201,72 @@ export class HardhatModule {
...
+    }
+
+    if (validatedParams[2].length !== 32) {
+      throw new InvalidInputError(
+        "Storage value must be exactly 32 bytes long"
```
fvictorio, line 254: Here too, we should include the value used.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +374,650 @@ describe("Hardhat module", function () {
...
+              },
+            },
+          ] = await compileLiteral(
+            `contract Storage {
+              function getValue(uint256 slot) public view returns (uint256 result) {
+                assembly { result := sload(slot) }
+              }
+              function setValue(uint256 slot, uint256 val) public {
+                assembly { sstore(slot, val) }
+              }
```
fvictorio, line 960: The `setValue` function doesn't seem to be used, can we remove it?

packages/hardhat-core/src/internal/hardhat-network/provider/node-types.ts
```diff
@@ -88,6 +88,7 @@ export interface Snapshot {
   txPoolSnapshotId: number;
   blockTimeOffsetSeconds: BN;
   nextBlockTimestamp: BN;
+  irregularStatesByBlockNum: Map<string, Buffer>;
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/85e74ad8b729a2b416f7c4f95ea57aa29a7ec22d

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -711,8 +713,11 @@ Hardhat Network's forking functionality only works with blocks from at least spu
       txPoolSnapshotId: this._txPool.snapshot(),
       blockTimeOffsetSeconds: this.getTimeIncrement(),
       nextBlockTimestamp: this.getNextBlockTimestamp(),
+      irregularStatesByBlockNum: this._irregularStatesByBlockNum,
     };
 
+    this._irregularStatesByBlockNum = new Map(this._irregularStatesByBlockNum);
```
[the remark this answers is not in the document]
feuGeneA: It clones the map so that the post-snapshot environment will have its own copy.  This way, any irregular states that come in while the snapshot is in effect will be applied only to the map that's specific to that snapshot.  When the snapshot is reverted, the map referred to by the snapshot object will again be used, and the clone will be garbage collected since the variable that was referring to it will then be referring to the original map that was in use before the snapshot was taken.

I found that, without this line, when a snapshot was in effect, any changes to the map would remain in the map even after the snapshot was reverted, because it was all operating on the same map.  At first I tried to use a simple assignment rather than `new Map()`, but it still didn't work, and I concluded that that was because a simple assignment was only creating an additional reference to the same object instance, and changing to `new Map()` made everything work nicely, so I assumed that this validated that conclusion.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -742,7 +747,15 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
     this._blockchain.deleteLaterBlocks(snapshot.latestBlock);
-    await this._stateManager.setStateRoot(snapshot.stateRoot);
+    this._irregularStatesByBlockNum = snapshot.irregularStatesByBlockNum;
+    const irregularStateOrUndefined = this._irregularStatesByBlockNum.get(
+      (await this.getLatestBlock()).header.number.toString()
+    );
+    await this._stateManager.setStateRoot(
+      irregularStateOrUndefined !== undefined
+        ? irregularStateOrUndefined
+        : snapshot.stateRoot
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/6bfb405cb94e2753799d604ee4bd405a899780f4

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1525,12 +1581,25 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
     }
-    return this._stateManager.setStateRoot(block.header.stateRoot);
+
+    const irregularStateOrUndefined = this._irregularStatesByBlockNum.get(
+      block.header.number.toString()
+    );
+    return this._stateManager.setStateRoot(
+      irregularStateOrUndefined !== undefined
+        ? irregularStateOrUndefined
+        : block.header.stateRoot
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/6bfb405cb94e2753799d604ee4bd405a899780f4

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -931,6 +944,49 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
   }
 
+  public async setAccountBalance(
+    address: Address,
+    newBalance: BN
+  ): Promise<void> {
+    const account = await this._stateManager.getAccount(address);
+    account.balance = newBalance;
+    await this._stateManager.putAccount(address, account);
+    this._irregularStatesByBlockNum.set(
```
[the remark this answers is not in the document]
feuGeneA: I considered this but concluded that the answer was no, because I thought that the World State Trie only includes balances, but now that you're asking the question I'm realizing that I was not seeing past the oversimplification of the diagram [here](https://ethereum.stackexchange.com/a/6413/6802).  I will set in the next three functions too, and also add tests to verify this behavior.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1525,12 +1581,25 @@ Hardhat Network's forking functionality only works with blocks from at least spu
 
   private async _setBlockContext(block: Block): Promise<void> {
     if (this._stateManager instanceof ForkStateManager) {
+      if (this._irregularStatesByBlockNum.has(block.header.number.toString())) {
+        throw new InternalError(
+          "Irregular states are not supported on a forked chain"
```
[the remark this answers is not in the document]
feuGeneA: This was a temporary measure! Thank you!

packages/hardhat-core/src/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -180,6 +201,72 @@ export class HardhatModule {
...
+
+  private async _setNonceAction(address: Buffer, newNonce: BN) {
+    await this._node.setAccountNonce(new Address(address), newNonce);
+    return true;
+  }
+
+  // hardhat_setStorageSlot
+
+  private _setStorageSlotParams(params: any[]): [Buffer, BN, Buffer] {
+    const validatedParams = validateParams(
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/d743619f7cc7a340de29f66f9d61c0e05ba23a54

packages/hardhat-core/test/internal/hardhat-network/stack-traces/compilation.ts
```diff
@@ -43,6 +51,21 @@ function getSolcInput(
...
+function getSolcInputForFiles(
+  sources: string[],
+  compilerOptions: CompilerOptions
+): CompilerInput {
+  return getSolcInput(getSolcSourceFileMapping(sources), compilerOptions);
+}
+
+function getSolcInputForLiteral(
+  source: string,
+  filename: string = "literal.sol",
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/c0b5ea682260ac50b1e41112efff9f2de0cdffb1

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/eb567038446c36e4a17fb6331123f7f70806ee90

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/eb567038446c36e4a17fb6331123f7f70806ee90

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/2013b5abe2c2cf6ac80f144ccdf24db27e755551

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/f73a29c9a7979e6f7460ab9284284d1510117b66

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/5a0631dac3f3f4ab0a1ab7c8c1b89aee6e2c3207

packages/hardhat-core/src/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/6726019c4d478e7835219a7f402580f16138d319

packages/hardhat-core/src/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/bd8b99f4803777485dee5b18d05673f663cfd204

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/8554a3e4a4b1a2d7ef0449e8cca93e029e270f4f

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
feuGeneA: Validated in https://github.com/nomiclabs/hardhat/pull/1435/commits/4e84f59dcdf3f06faa294f84be93d90c926929eb (see [CI result](https://github.com/nomiclabs/hardhat/runs/2775774777#step:9:6879)) and addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/bed37937ee1009830cc3400acee2d7ece1359e57

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
feuGeneA: I was just being overly defensive I guess.  Reverted in https://github.com/nomiclabs/hardhat/pull/1435/commits/82c501835fb4b9510f59baedf4d0d1516da9e1ca

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
alcuadrado: TIL: you can clone a map with `new Map(map)`. I used to do `new Map(map.entries())`

Agreed that cloning is necessary. Some time ago this method cloned lots of things like that.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -931,6 +945,49 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+  public async setAccountCode(
+    address: Address,
+    newCode: Buffer
+  ): Promise<void> {
+    await this._stateManager.putContractCode(address, newCode);
+    await this._persistIrregularWorldState();
+  }
+
+  public async setAccountNonce(address: Address, newNonce: BN): Promise<void> {
+    const account = await this._stateManager.getAccount(address);
```
alcuadrado, line 967: I haven't think about this until now, but what should happen if there are pending txs?

If automining is disabled there's both a nonce that comes from the state manager, but there may also be pending txs that each of them have nonces.

Not sure what should we do here. Drop the pending txs if their nonces are too low? Forbid this method if there are pending txs?

Wdyt, @feuGeneA @fvictorio ?

alcuadrado:
I just reviewed the entire PR again, great work, @feuGeneA! Nice tests :)

I just left a comment regarding something we hadn't considered yet.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: I thought more about this, and I think we should forbid using this method if you have pending transactions. I think it's highly improbable that someone will be limited by this in practice, and it significantly simplifies the implementation.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -365,6 +374,737 @@ describe("Hardhat module", function () {
...
+        it("should not reject valid argument types", async function () {
+          await this.provider.send("hardhat_setStorageSlot", [
+            DEFAULT_ACCOUNTS_ADDRESSES[0],
+            numberToRpcQuantity(0),
+            `0x${"ff".repeat(32)}`,
+          ]);
+        });
+
+        it("should result in modified storage", async function () {
+          const targetStorageSlot = 1;
```
fvictorio, line 989: This variable doesn't seem to be used (the other variables called `targetStorageSlot` seem unused too).

packages/hardhat-core/src/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -180,6 +202,76 @@ export class HardhatModule {
...
+
+  private async _setNonceAction(address: Buffer, newNonce: BN) {
+    await this._node.setAccountNonce(new Address(address), newNonce);
+    return true;
+  }
+
+  // hardhat_setStorageSlot
+
+  private _setStorageSlotParams(params: any[]): [Buffer, BN, Buffer] {
+    const [b, slotIndex, value] = validateParams(
```
fvictorio, line 241: Let's use `address` instead of `b`. (Sorry, I should've suggested this in the initial comment, but I didn't realize that the first variable had to be extracted too for the return value).

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
fvictorio: That seems fine to me, yes :+1:

feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/74ecff3a60db7c360aebc8ed9853ce932af6bd1b

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Thanks for noticing this. I wish there was a linter rule for this.  Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/4d5b673e39e93d1f14e7ef40da599280bc106d95

packages/hardhat-core/src/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/79467c56d0ba37208027aaff8ee92e1f2d88623c

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1435/commits/8e9a113376462500a2ad6ade6908a6680fd0a59b

I took a very quick look at actually checking by address, and it didn't seem trivial so I skipped it for now, but I'm happy to look into that if you guys think it's important.

alcuadrado: I'm fine with this. We can always relax that restriction by checking by address.

alcuadrado (approved):
lgtm

PR #1437, opened 2021-05-03
Update mainnet-forking.md

Leon-Africa:
Steps for forking Binance Smart Chain


fvictorio:
Hey @Leon-Africa, thanks for this, but since we don't officially support BSC, adding this to our official docs might give the wrong impression. Maybe you can suggest they add it to their docs somewhere?

Leon-Africa (commented):
(2022) hardhat not supporting BSC

PR #1445, opened 2021-05-06
Add support for top-level functions

fvictorio:
These are two very big commits.

The first one (f8207b4) is the refactor. It's probably better to just ignore it, and to just check the `SolidityTracer` and `ErrorInferrer` files.

The other one adds some tests for top-level functions, updates the model so that the `contract` field in `ContractFunction` is optional (maybe we should rename that class?), and changes the `SolidityCallSite` to print a function name instead of a method name when there is no contract.


packages/hardhat-core/src/internal/hardhat-network/stack-traces/model.ts
```diff
@@ -23,6 +23,7 @@ export enum ContractFunctionType {
   RECEIVE,
   GETTER,
   MODIFIER,
+  FREE_FUNCTION,
 }
 
 export enum ContractFunctionVisibility {
```
alcuadrado, line 29: Maybe rename this to `FunctionVisibility` now?

packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
```diff
@@ -673,9 +679,17 @@ const solidity06Compilers: CompilerOptions[] = [
...
 
 const solidity07Compilers: CompilerOptions[] = [
+  // {
+  //   solidityVersion: "0.7.0",
+  //   compilerPath: "soljson-v0.7.0+commit.9e61f92b.js",
+  // },
+  // {
+  //   solidityVersion: "0.7.1",
+  //   compilerPath: "soljson-v0.7.1+commit.f4a555be.js",
+  // },
```
alcuadrado, line 689: I'd also add 0.7.2 and 0.7.3, commented out, to make running regression tests easier.

packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
```diff
@@ -247,6 +247,12 @@ function compareStackTraces(
   description: StackFrameDescription[],
   runs: number | undefined
 ) {
+  assert.equal(
+    trace.length,
+    description.length,
+    `Expected a trace of length ${description.length} but got one with length ${trace.length}`
```
alcuadrado, line 253: I think I removed this a long time ago because it was annoying when debugging. Things changed a lot, so if you think that's not the case anymore, fine for me.

alcuadrado:
@fvictorio this looks good to me, but I'm not sure how to review the `ErrorInferrer` apart from the general structure. Anything you think I should review in detail?

I did review the free-functions-related code, and looks good.

packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
fvictorio: It was useful at some point in the refactor. I guess if it becomes annoying we can remove it again.

alcuadrado:
As discussed, I reviewed `ErrorInferrer#inferAfterTracing` and `ErrorInferrer#_checkLastSubmessage`, and they look good. I'm going to resolve the conflicts and merge this now.

alcuadrado:
The CI failed because of the usual timeouts. I'm merging this.

PR #1452, opened 2021-05-11
Add support for solidity 0.8.0

fvictorio:
Depends on #1445.

This PR adds support for the panic codes introduced in [Solidity 0.8.0](https://blog.soliditylang.org/2020/12/16/solidity-v0.8.0-release-announcement/). It also improves how return data is handled, so that revert reasons are never ignored if present.

### Reviewing this PR

Commit dc8c2698f6c20120a0e48fd85967b59836e2c6c3 and a7d5baf7ca7ea7ac135e36ea5230b40ff56c6176 are probably best reviewed separately.

The first one adds support for panic errors. Most of its changes are tests. Then there's a `panic-codes.ts` module that converts a code to a description, and a new heuristic in the error inferrer.

The second one adds the `ReturnData` class, adapts all the necessary parts of the code, and changes how the node builds the returned error. This last part is the most important one of the commit.


packages/hardhat-core/src/internal/hardhat-network/stack-traces/error-inferrer.ts
```diff
@@ -1442,6 +1514,11 @@ export class ErrorInferrer {
 
     return this._failsRightAfterCall(trace, callStepIndex);
   }
+
+  private _isPanicReturnData(returnData: Buffer): boolean {
```
alcuadrado, line 1518: This feels a bit duplicated. Can't you do `new ReturnData(returnData).is...`?

packages/hardhat-core/src/internal/hardhat-network/stack-traces/error-inferrer.ts
```diff
@@ -368,6 +381,50 @@ export class ErrorInferrer {
...
+  private _checkPanic(
+    trace: DecodedEvmMessageTrace,
+    stacktrace: SolidityStackTrace,
+    lastInstruction: Instruction
+  ): SolidityStackTrace | undefined {
+    if (!this._isPanicReturnData(trace.returnData)) {
+      return;
+    }
+
+    // If the last frame is an internal function, it means that the trace
```
alcuadrado, line 396: This is hard to follow. Maybe some ascii art or an example?

alcuadrado (approved):
This PR looks good. Just left two minor comments.

alcuadrado:
PS: I just noticed the comments in a few commits. I haven't reviewed that.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/error-inferrer.ts
fvictorio: Yes, good catch.

PR #1453, opened 2021-05-12
Improve and simplify hardhat-etherscan error message

fvictorio:
Users always get confused about this message. I also simplified the logic a little because I don't think it's worth that complexity.


packages/hardhat-etherscan/src/network/prober.ts
```diff
@@ -29,7 +29,7 @@ enum NetworkID {
   HECO = 128,
   HECO_TESTNET = 256,
   // Fantom mainnet
-  OPERA = 250
+  OPERA = 250,
```
fvictorio, line 32: Unrelated to this PR, I don't know why the CI didn't catch this.

alcuadrado: This happened in a few other PRs. Not sure what's wrong.

PR #1454, opened 2021-05-12
Printing hardhat config path in verbose mode

Bidon15:


packages/hardhat-core/src/internal/core/project-structure.ts
```diff
@@ -9,11 +10,17 @@ import { ERRORS } from "./errors-list";
...
 
+const log = debug("hardhat:core:prj-structure");
+
 export function isCwdInsideProject() {
-  return (
-    findUp.sync(TS_CONFIG_FILENAME) !== null ||
-    findUp.sync(JS_CONFIG_FILENAME) !== null
-  );
+  const jsConfigPath = findUp.sync(JS_CONFIG_FILENAME);
+  const tsConfigPath = findUp.sync(TS_CONFIG_FILENAME);
```
Bidon15, line 17: Maybe a good idea to relocate *ConfigPath vars out of functions as getUserConfigPath() contains the same vars.

packages/hardhat-core/src/internal/core/project-structure.ts
```diff
@@ -9,11 +10,17 @@ import { ERRORS } from "./errors-list";
...
+  const path = jsConfigPath !== null ? jsConfigPath : tsConfigPath;
+  if (path !== null) {
```
Bidon15, line 19: I didn't want to display the null path during project creation

alcuadrado:
Hey, thanks for submitting this issue, @Bidon15! 

I wonder if this isn't a better place to log it: https://github.com/nomiclabs/hardhat/blob/d49afce4178f846241381168f2404b21e148ced3/packages/hardhat-core/src/internal/core/config/config-loading.ts#L44

Bidon15:
> Hey, thanks for submitting this issue, @Bidon15!
> 
> I wonder if this isn't a better place to log it:
> 
> https://github.com/nomiclabs/hardhat/blob/d49afce4178f846241381168f2404b21e148ced3/packages/hardhat-core/src/internal/core/config/config-loading.ts#L44

Thanks for the guide @alcuadrado !
Yes. This look much better then my initial one.

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -41,7 +44,7 @@ export function loadConfigAndTasks(
     hardhatArguments !== undefined ? hardhatArguments.config : undefined;
 
   configPath = resolveConfigPath(configPath);
-
+  log(`Hardhat config path located at ${configPath}`);
```
alcuadrado, line 47:
```suggestion
  log(`Loading Hardhat config from ${configPath}`);
```

What about this message instead? It makes it clear what was happening at the time the log was printed.

Bidon15: Yeap. Clearer

alcuadrado:
Oh, the linter is failing

PR #1455, opened 2021-05-13
[WIP] Don't run the forking tests on every node version

alcuadrado:


alcuadrado:
This approach doesn't fix the problem with alchemy, so I'm closing this PR.

PR #1458, opened 2021-05-17
Fix awful error for balance

Bidon15:
Fix Awful error when account balance:

1. Contains hex string
2. Have invalid decimal string

### DIFF
--- a/packages/hardhat-core/src/internal/core/config/config-validation.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-validation.ts
@@ -64,6 +64,7 @@ export const DotPathReporter: Reporter<string[]> = {
 };
 
 const HEX_STRING_REGEX = /^(0x)?([0-9a-f]{2})+$/gi;
+const DEC_STRING_REGEX = /^[1-9]+[0-9]+([0-9])?$/g;
 
 function isHexString(v: unknown): v is string {
   if (typeof v !== "string") {
@@ -73,6 +74,14 @@ function isHexString(v: unknown): v is string {
   return v.trim().match(HEX_STRING_REGEX) !== null;
 }
 
+function isDecString(v: unknown): v is string {
+  if (typeof v !== "string") {
+    return false;
+  }
+
+  return v.match(DEC_STRING_REGEX) !== null;
+}
+
 export const hexString = new t.Type<string>(
   "hex string",
   isHexString,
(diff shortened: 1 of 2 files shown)


packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -64,6 +64,7 @@ export const DotPathReporter: Reporter<string[]> = {
 };
 
 const HEX_STRING_REGEX = /^(0x)?([0-9a-f]{2})+$/gi;
+const DEC_STRING_REGEX = /^[1-9]+[0-9]+([0-9])?$/g;
```
fvictorio, line 67: I think this regex is wrong. it should be:

```js
const DEC_STRING_REGEX = /^[1-9]+[0-9]*$/g;
```

More importantly, there should be a test that catched this.

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -73,19 +74,33 @@ function isHexString(v: unknown): v is string {
...
+}
+
 export const hexString = new t.Type<string>(
   "hex string",
   isHexString,
   (u, c) => (isHexString(u) ? t.success(u) : t.failure(u, c)),
   t.identity
 );
 
+export const decString = new t.Type<string>(
```
fvictorio, line 92: I would replace all `dec` for `decimal`, to make it more clear. I know we use `hex` instead of `hexadecimal`, but `hex` is easier to understand that `dec` IMO.

packages/hardhat-core/test/internal/core/config/config-validation.ts
```diff
@@ -449,6 +449,66 @@ describe("Config validation", function () {
...
+            () =>
+              validateConfig({
+                networks: {
+                  [HARDHAT_NETWORK_NAME]: {
+                    accounts: [{ privateKey: "0xxxxx", balance: "-123" }],
+                  },
+                },
+              }),
+            ERRORS.GENERAL.INVALID_CONFIG
+          );
```
fvictorio, line 510: There should be a couple of tests for _valid_ configs, that would've watched the regex error. For example:

- 0
- 1
- 10
- 1234

fvictorio (changes requested):

fvictorio:
Thanks for working on this @Bidon15! There are a couple of issues, but they shouldn't be too hard to fix.

packages/hardhat-core/src/internal/core/config/config-validation.ts
Bidon15: I totally forgot that we can have 0 balance. I'll apply this:
`^(0|[1-9][0-9]*)$`

fvictorio: Oops, that's right, my regex didn't accept 0 :sweat_smile:

fvictorio (approved):

packages/hardhat-core/src/internal/core/config/config-validation.ts
Bidon15: Same as my initial 😂

alcuadrado:
Nice! Thanks, @Bidon15 !

PR #1483, opened 2021-05-25
Allow forking cache on network without known reorg

phated:
I was trying to fork the xdai network and noticed that nothing was being cached (which is a massive problem because their archive RPC drops connections all the time); however, I found that it wouldn't be cached because it is not in the list of known reorgs since `actualMaxReorg` was being checked instead of `maxReorg` which has the fallback.

This change allows any forked network to be cached.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
@@ -81,7 +81,7 @@ Please use block number ${lastSafeBlock} or wait for the block to get ${
   const cacheToDiskEnabled =
     forkConfig.blockNumber !== undefined &&
     forkCachePath !== undefined &&
-    actualMaxReorg !== undefined;
+    maxReorg !== undefined;
 
   const forkClient = new JsonRpcClient(
     provider,


alcuadrado:
I never know what we should do in this case. Maybe print a warning and cache anyway? What do you think? I'm inclined to play safe here, but hardhat is used a lot with other chains now, so maybe our "no cache unless we are 100% sure is safe" is a disservice to lots of users.

@fvictorio, wdyt?

fvictorio:
I would like to try this out and see what happens right now and experiment with some error messages. @phated what URL are you using? `https://xdai-archive.blockscout.com`?

phated:
> what URL are you using? https://xdai-archive.blockscout.com?

Yep! It's a little spotty right now because the Dark Forest round is happening 😅

phated:
Any news about this? I had to publish a fork so I could have a cache for xdai.

alcuadrado:
I have the feeling that we should at least print a warning, but that may also be extremely annoying, right?

Another alternative is adding a config that overrides our default max reorg numbers, but this would be annoying in a different sense.

We can also print the warning just once. Maybe this is the best alternative. We already do that in some cases.

phated:
I think including it in the warning message that is already printed makes sense. Why would you want 5 block reorg instead of the 30 that is the current default?

jacobrosenthal:
Would a separate PR that simply adds xdai network to `hardhat/packages/hardhat-core/src/internal/hardhat-network/provider/utils/reorgs-protection.ts` with a 5 block reorg be more readily mergable?

alcuadrado:
> Would a separate PR that simply adds xdai network to `hardhat/packages/hardhat-core/src/internal/hardhat-network/provider/utils/reorgs-protection.ts` with a 5 block reorg be more readily mergable?

Yes, sorry we took so long with this. If you send a smaller PR, which only fixes this for xdai by returning a reasonable value there it'd be easier to merge. We can then fix it in the general case.

I don't know what's a reasonable value for xdai, but something slightly above its longest historic reorg will work.

fvictorio:
Hey @phated, thanks for working on this. We think it would be better to properly discuss the underlying problem in a new issue and then see which solution would make sense for us. I think the problem here is that the forking feature doesn't use a disk cache for unknown networks. One alternative solution we discussed is to let users configure the "safe" number of blocks for the forked network.

Please let me know if you can't or don't want to open that issue and I'll do it!

jacobrosenthal:
@igorbarinov can you comment on https://github.com/nomiclabs/hardhat/pull/1483#issuecomment-857014074 Were looking to whitelist xdai network https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/hardhat-network/provider/utils/reorgs-protection.ts#L10


We need number of blocks "that should be safe to consider as the * largest possible reorg in a network"

igorbarinov:
from what we know from the security audit https://github.com/poanetwork/aura-review-commentary/blob/master/aura-audit-report.md#finality-delay it's 2*n_v where n_v is 17 at the moment https://blockscout.com/xdai/mainnet/validators with the maximum number of 19 per a staking epoch. 

I'd say 38 (2*19) is a safe threshold

alcuadrado:
Thanks everyone for helping with this. This change has been released in Hardhat 2.4.0 🥳

PR #1484, opened 2021-05-26
Add ArgumentsExtenders

passabilities:
Adds an `extender` function for modifying the Hardhat arguments before executing anything. This allows for performing verification on the `hre.network` before connecting to / forking a network. 

The `HardhatConfig.networks.hardhat` configuration is not dynamically loaded and is unable to change after the `hardhat` script executes. This means that you can only have 1 configuration set for being able to fork a network. If you want to fork multiple networks, you will need to manually update the `hardhat` network config before executing the script. 

Instead, it would be nice to specify that I want to fork in the command I want to run and the network I want to fork:
`FORKING_ENABLED=true yarn hh test --network {mainnet|polygon}`

```typescript
extendArguments((args: Partial<HardhatArguments>) => {
  if (process.env.FORKING_ENABLED === 'true' && args.network !== HARDHAT_NETWORK_NAME) {
    process.env.FORKING_NETWORK = args.network
    args.network = 'hardhat'
  }
})

extendConfig((config: HardhadConfig) => {
  if (process.env.FORKING_NETWORK != null) {
    const networkConfig = config.networks[process.env.FORKING_NETWORK]
    if (!('url' in networkConfig)) {
      throw new Error(`Invalid forking network config`)
    }

    const { url: forkingUrl, ...rest } = networkConfig
    const forkingConfig: HardhatNetworkUserConfig = {
      ...rest,
      chainId: 31337,
      accounts,
      forking: {
        url: forkingUrl,
        blockNumber: networkConfig.forkBlockNumber,
        enabled: true
      },
      saveDeployments: true,
      live: false
    }

    merge(
      config.networks.hardhat,
      forkingConfig
    )
  }
})
```

```typescript
{
  ...,
  networks: {
    kovan: networkConfig({
      url: process.env.ALCHEMY_KOVAN_KEY!,
      chainId: 42,
      live: true
    }),
    mainnet: networkConfig({
      url: process.env.ALCHEMY_MAINNET_KEY!,
      chainId: 1,
      live: true,
      forkBlockNumber: 12480795,
    }),
    polygon: networkConfig({
      url: `https://rpc-mainnet.maticvigil.com`,
      chainId: 137,
      live: true,
      // forkBlockNumber: 14891625,
    }),
    localhost: networkConfig({
      url: 'http://127.0.0.1:8545',
      timeout: 100000,
    }),
  }
}
```

### DIFF
--- a/packages/hardhat-core/package.json
+++ b/packages/hardhat-core/package.json
@@ -26,6 +26,7 @@
     "node": ">=8.2.0"
   },
   "scripts": {
+    "prepack": "yarn build",
     "lint:fix": "prettier --write \"src/**/*.{js,ts}\" \"test/**/*.{js,ts}\" && yarn lint-src --fix && yarn lint-tests --fix",
     "lint": "yarn lint-src && yarn lint-tests",
     "lint-tests": "NODE_OPTIONS='--require ts-node/register/transpile-only' tslint --config tslint.json --project ./tsconfig.json",
--- a/packages/hardhat-core/src/internal/cli/cli.ts
+++ b/packages/hardhat-core/src/internal/cli/cli.ts
@@ -16,7 +16,6 @@ import { getEnvHardhatArguments } from "../core/params/env-variables";
 import { HARDHAT_PARAM_DEFINITIONS } from "../core/params/hardhat-params";
 import { isCwdInsideProject } from "../core/project-structure";
 import { Environment } from "../core/runtime-environment";
-import { loadTsNode, willRunWithTypescript } from "../core/typescript-support";
 import { Reporter } from "../sentry/reporter";
 import { isRunningOnCiServer } from "../util/ci-detection";
 import {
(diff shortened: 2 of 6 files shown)


fvictorio:
Hey @passabilities, thanks for this. If I understand correctly, this would let you do things like:

```
hh test --network rinkeby
```

and for that to actually fork rinkeby instead of using rinkeby as a JSON-RPC network. Is that correct?

I understand how this could be useful, but I feel this can be accomplished just with environment variables + doing dynamic things in the config. For example:

```js

const { FORKING_NETWORK } = process.env

const networksUrls = {
  mainnet: "<MAINNET_URL>"
  rinkeby: "<RINKEBY_URL>"
}

module.exports = {
  networks: {
    hardhat: {
      forking: FORKING_NETWORK === undefined ? {} : {
        url: networksUrls[FORKING_NETWORK]
      }
    },
    mainnet: { url: networksUrls.mainnet },
    rinkeby: { url: networksUrls.rinkeby }
  }
}
```

and then you do

```
FORKING_NETWORK=rinkeby hh test
```

(give or take, I haven't run this but I think it illustrates the idea).

I'm not saying that I'm against adding this, but we are careful when adding features like this one, so I'd like to understand better what things would this enable that are not possible right now.

passabilities:
the main idea was to be able to run some setup code rather than have to type the configuration out in terminal

fvictorio:
> the main idea was to be able to run some setup code rather than have to type the configuration out in terminal

I'm not sure I understood this. Can you elaborate on that?

fvictorio:
Hey @passabilities, I'm going to close this now because we aren't quite sure about this feature. If you still think this would be a valuable addition, please open an issue first to discuss what problem this would solve and alternative solutions (if any).

PR #1497, opened 2021-05-30
Wait for tx to be mined in sample project

fvictorio:
The reason to do this is that otherwise the test doesn't work out-of-the-box when run against a testnet.

### DIFF
--- a/packages/hardhat-core/sample-project/test/sample-test.js
+++ b/packages/hardhat-core/sample-project/test/sample-test.js
@@ -4,11 +4,13 @@ describe("Greeter", function() {
   it("Should return the new greeting once it's changed", async function() {
     const Greeter = await ethers.getContractFactory("Greeter");
     const greeter = await Greeter.deploy("Hello, world!");
-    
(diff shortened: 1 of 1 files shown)


packages/hardhat-core/sample-project/test/sample-test.js
```diff
@@ -4,11 +4,13 @@ describe("Greeter", function() {
...
     const Greeter = await ethers.getContractFactory("Greeter");
     const greeter = await Greeter.deploy("Hello, world!");
-    
     await greeter.deployed();
+
     expect(await greeter.greet()).to.equal("Hello, world!");
 
-    await greeter.setGreeting("Hola, mundo!");
+    const setGreetingTx = await greeter.setGreeting("Hola, mundo!");
+    await setGreetingTx.wait();
```
alcuadrado, line 12: Maybe dd a comment here? Most people don't do this.

fvictorio: Makes sense, added.

PR #1501, opened 2021-06-01
Split eth module tests in multiple files

fvictorio:
The new structure for these tests is this:

```
eth
├── methods
│   ├── blockNumber.ts
│   ├── call.ts
│   ├── estimateGas.ts
│   ├── ...
│   ├── index.ts
│   ├── ...
│   └── unsubscribe.ts
├── blockFilters.ts
├── blockTags.ts
├── gasUsage.ts
├── hardforks.ts
└── receiptsRoot.ts
```

The `eth/methods` directory has tests for all the different RPC methods under the eth module. There is one file for each method, except for some of them (mostly the unsupported ones) that are all included together in the `eth/methods/index.ts` file.

The tests inside `eth` but outside `eth/methods` are other tests that don't exactly belong to a single method, like the hardfork tests, or the gas usage tests.

A downside here is that now you can't add a `.only` and run all the tests in the module, but I think it's worth it. Plus, you can always do:

```
yarn mocha --recursive "test/internal/hardhat-network/provider/modules/eth/**/*.ts"
```

if you need that.


alcuadrado (approved):
LGTM. I like this idea. The `eth.ts` file was too large.

PR #1509, opened 2021-06-05
don't allow own package name in solidity file import

### LINKED ISSUE #1500: Importing a local solidity file from its package name can fail in unexpected ways when using a monorepo
This is a bug report of an edge case that can only happen if you have a hardhat project inside a monorepo, and the hardhat project is also an npm project `p`.

If a solidity file within `p` imports another one using `import "p/contracts/..."`, the compilation will fail. The reason for this is that the file ends up with two source names `p/contract/...` and `contracts/...`.

This only happens within a monorepo, because otherwise `require.resolve("p/...")` would fail. In a monorepo it resolves because the top `node_modules` will have a symlink to `p`.

We should forbid importing a file using its own npm project name. We should tell the user which import to use instead.

This bug can be reproduced like this:
1. clone https://github.com/alcuadrado/balancer-core-v2 -- it has a faulty import in `pkg/pool-weighted/contracts/smart/LiquidityBootstrappingPool.sol`
2. cd into the project
3. `git checkout compilation-bug-reproduction`
4. `yarn`
5. `cd pkg/pool-weighted`
6. `yarn hardhat compile --show-stack-traces --verbose`

KaanKC:
Verifies that a solidity file import doesn't contain the own package's name. See the linked issue for further details.

The PR is still missing tests. I wanted some feedback before I spend more time on it.

close #1500 

Closes HH-763

### DIFF
--- a/packages/hardhat-core/src/utils/source-names.ts
+++ b/packages/hardhat-core/src/utils/source-names.ts
@@ -3,6 +3,7 @@ import path from "path";
 import { HardhatError } from "../internal/core/errors";
 import { ERRORS } from "../internal/core/errors-list";
 import { FileNotFoundError, getFileTrueCase } from "../internal/util/fs-utils";
+import { getPackageName } from "../internal/util/packageInfo";
 
 const NODE_MODULES = "node_modules";
 
@@ -227,4 +228,18 @@ async function getSourceNameTrueCase(
     // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
     throw error;
   }
+}
+
+/**
+ * This function returns true if the sourceName contains the current package's name
+ * as a substring
+ */
+export async function includesOwnPackageName(
+  sourceName: string
+): Promise<boolean> {
+  const packageName = await getPackageName(sourceName);
+  if (packageName !== "") {
+    return sourceName.startsWith(`${packageName}/`);
+  }
+  return false;
 }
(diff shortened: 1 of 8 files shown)


packages/hardhat-core/src/utils/source-names.ts
```diff
@@ -169,6 +170,19 @@ export function isAbsolutePathSourceName(sourceName: string): boolean {
...
 }
 
+/**
+ * This function returns true if the sourceName contains the current package's name
+ * as a substring
+ */
+export async function includesOwnPackageName(sourceName: string): Promise<boolean> {
+  const packageName = await getPackageName(sourceName);
+  if (packageName) {
+    return sourceName.includes(packageName);
```
alcuadrado, line 180: Shouldn't this be includes `${packageName}/`?

alcuadrado (commented):
Hey, sorry we took so long to review this. It looks really good, but can you add some tests?

Also, it'd be easier to merge if you rebase it against master.

Thanks!

packages/hardhat-core/src/utils/source-names.ts
fvictorio: Maybe ```startsWith(`${packageName}/`)``` would be a bit safer.

KaanKC: good point! I will update the PR over the weekend

KaanKC:
I've fixed linter issues, added tests, and rebased.

fvictorio:
Thank you @KaanKC and sorry for taking so long to review this!

Issue #1040, opened 2020-11-19
Allow plugins to provide more artifacts


wighawag:
Currently hardhat offer a function `hre.artifacts.readArtifact`  (as well as other method in `hre.artifacts`)
Unfortunately such `artifacts` field is hardwired at runtime creation with a specific folder and as a read-only field it is not supposed to be modified.

`hardhat-deploy` (and even `builder-deploy` before) offer a mechanism for external import of artifacts, it support different format like truffle too. It does so with its own custom function so user can deploy these artifacts.

But since hardhat, as mentioned do not allow plugin to override the canonical way to get artifacts, other plugin that offer new features to artifacts are not aware of hardhat-deploy artifacts, even if they are compatible.

There are different approches to solve this. One approach is to allow plugin to hook into the `getArtifact` process so plugins can import artifact from other sources, even transforming it to fit hardhat expected format.

This way any plugin that add features to artifacts can access them via a standard mechanism

fvictorio:
If I'm understanding correctly, you would like to add extra fields to the JSON artifacts, and then be able to consume them, is that correct?

One thing to keep in mind is that if you add fields, the current type of `Artifact` won't have those extra fields, but that's an easy problem to circumvent.

wighawag:
It is not about adding extra field to artifacts, but being able for a plugin to add more artifacts, from different sources. And having that reachable to other plugin

hardhat-deploy already offer that feature but these artifacts are not seen by other plugins

wighawag:
Any plan to support that somehow? 

It has resurfaced with my update of hardhat-deploy-ethers (wip) that is now an extension of hardhat-ethers 

hardhat-deploy-ethers used to re-write the function of hardhat-ethers but now this is not easily possible, nor recommended (would need to guess any future addition of function)

One of the reason for the re-write was to ensure `ethers.getContractFactoryByName` would still consider the extra artifacts that hardhat-deploy allow user to provide.

It would be a lot better if somehow plugin could override hre.artifacts directly, this wwould make any plugin using artifacts.readArtifact support the extra artifacts provided.

rhlsthrm:
I would very much like this feature to be able to use external libraries in my Hardhat project!

wighawag:
I started to look at implementating it but I am not sure the best forward

[...]

PR #1514, opened 2021-06-07
Allow overriding `artifacts.readArtifact` in `getContractFactory`

heri16:
Allows third-party plugins to override `artifacts.readArtifact` getter in `ethers.getContractFactory()`.

E.g. hardhat-deploy plugin has config for `external.contract.artifacts`.

### DIFF
--- a/packages/hardhat-ethers/src/internal/helpers.ts
+++ b/packages/hardhat-ethers/src/internal/helpers.ts
@@ -2,6 +2,7 @@ import type { ethers } from "ethers";
 import { NomicLabsHardhatPluginError } from "hardhat/plugins";
 import {
   Artifact,
+  Artifacts,
   HardhatRuntimeEnvironment,
   NetworkConfig,
 } from "hardhat/types";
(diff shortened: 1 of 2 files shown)


heri16:
Issue #1515

heri16:
Issue #1040

fvictorio:
Fair warning: thinking about these things and making decisions on them takes us a lot of time, so sorry in advance if this discussion gets a bit stalled.

alcuadrado:
Something that we discussed with Ronnan was having a function like `extendArtifactsResolution` that takes a callback which would be used if a user tries to read an artifact that doesn't exist locally.

One issue I see with this is that there's no way to distinguish between an unexisting artifact and not having compiled yet.

heri16:
Since some artifacts might not be compiled yet, or are precompiled by an external plugin, or an artifact compiled by hardhat might need to be morphed by an external plugin... maybe an outright overrideArtifactsResolution is cleaner?

fvictorio:
Hey @heri16, thanks for working on this, but we think we need to properly discuss this (in #1040, I guess) before deciding what to do. Please feel free to comment on that discussion with this proposed solution and we can continue over there.

PR #1516, opened 2021-06-07
Accept signer address in `ethers.getContractAt()` and `ethers.getContractFactory()`

heri16:
Accept signer address as string param to `hre.ethers.getContractAt()` and `hre.ethers.getContractFactory()`


fvictorio:
Hey @heri16, thanks for sending this. Unfortunately, we try not to merge new features that we haven't discussed first. Feel free to open an issue suggesting this change so we can see if it makes sense for us to include it.

I know this is a very small change, but we are very picky about new features because each one adds complexity (i.e., potential bugs and things to document), so we err on the side of not adding new things.

PR #1517, opened 2021-06-07
[WIP] Emit warning for empty config object

Endlessline:
Newbie here, hopefully this is getting to the right area.

### DIFF
--- a/packages/hardhat-waffle/src/waffle-chai.ts
+++ b/packages/hardhat-waffle/src/waffle-chai.ts
@@ -43,7 +43,8 @@ export function waffleChai(chai: Chai.ChaiStatic, utils: Chai.ChaiUtils) {
   const {
     supportRevertedWith,
   } = require(`${waffleChaiPath}/matchers/revertedWith`);
-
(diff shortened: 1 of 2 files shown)


alcuadrado:
Hey @Endlessline,

thanks for submitting this PR, but I'm a bit confused about it, as an empty config is perfectly valid. Can you explain what was your intention here, please? Thanks!

packages/hardhat-core/src/internal/context.ts
```diff
@@ -78,16 +78,25 @@ export class HardhatContext {
   }
 
   public getFilesLoadedDuringConfig(): string[] {
```
alcuadrado, line 80: This is an unrelated concept. What we want to check is the `userConfig` here: https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/core/config/config-loading.ts#L66

fvictorio:
@Endlessline the commits in this PR seem completely off. Maybe you did a wrong merge/rebase or something like that?

Endlessline:
Good looking out @fvictorio, I reset the head in my fork and closed this out. I'll start anew so as not to muddy up the history.

PR #1524, opened 2021-06-08
core: Permit nulls in geth's optional JSON-RPC input fields

### LINKED ISSUE #985: HardhatNetworkProvider._send() rejects null `from` value
HardhatNetworkProvider._send(method,params) rejects a null value for params[0].from. It accepts an undefined value. I noticed this while writing an adapter for the dydx Solo package, which seems to make calls with from=null.

### LINKED ISSUE #1441: Accept `null` in the "to" field of a transaction
To deploy a contract in hardhat, you need to send a tx without the `to` field. But geth also accepts a `"to": null` field for the same thing, and its client uses that, so we probably should too.

feuGeneA:
Fixes #1441 
Fixes #985


fvictorio:
We should add some tests for this. Not for everything that accepts null, but at least for a couple of them.

packages/hardhat-core/src/internal/util/io-ts.ts
```diff
@@ -1,5 +1,28 @@
 import * as t from "io-ts";
 
+export function optionalOrNullable<TypeT, OutputT, InputT>(
+  codec: t.Type<TypeT, OutputT, InputT>,
+  name: string = `${codec.name} | undefined`
+): t.Type<TypeT | undefined, OutputT | undefined, InputT | undefined | null> {
+  return new t.Type(
+    name,
+    (u: unknown): u is TypeT | undefined =>
+      u === undefined || u === null || codec.is(u),
```
fvictorio, line 10: I'm still confused about the difference between the type guard and the decoding function, but I think this shouldn't have the `|| u === null` part. Also notice that it's not compatible with the signature, which is saying that `u is TypeT | undefined`

packages/hardhat-core/src/internal/util/io-ts.ts
```diff
@@ -1,5 +1,28 @@
...
+    (i, c) => {
+      if (i === undefined || i === null) {
+        return t.success(undefined);
+      }
+      return codec.validate(i, c);
+    },
+    (a) => {
+      if (a === undefined || a === null) {
```
fvictorio, line 18: This `a === null` doesn't seem necessary either. Notice that the type of `a` is `TypeT | undefined`.

packages/hardhat-core/src/internal/util/io-ts.ts
alcuadrado: Yes, `=== null` shouldn't be here.

What makes the type guard vs decoding function so weird is that TypeT can be more general than OutputT. What for? What's the use case for this? I have no idea.

feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1524/commits/6aad332ce28da8583cf291cab7a7e32ee12f9425

fvictorio:
This looks good to me. I would only add one more test: deploying a contract by setting the `to` field to `null` since that's the most likely scenario where someone explicitly uses a `null`.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/sendTransaction.ts
```diff
@@ -135,43 +135,47 @@ describe("Eth module", function () {
...
+                  blockNumber: "0x6",
+                  contractAddress: "0x6ea84fcbef576d66896dc2c32e139b60e641170c",
+                  cumulativeGasUsed: "0xcf0c",
+                  from: "0xda4585f6e68ed1cdfdad44a08dbe3979ec74ad8f",
+                  gasUsed: "0xcf0c",
+                  logs: [],
+                  logsBloom:
+                    "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+                  status: "0x1",
+                  to: null,
```
fvictorio, line 166: Shouldn't you use `toValue` here?

feuGeneA: :facepalm: :pray: 

Note, though, that it shouldn't be used here, because this is the receipt; rather, it should be used further up, in the RPC call.  Regardless, :facepalm: for not even using `toValue`!

Addressed in https://github.com/nomiclabs/hardhat/pull/1524/commits/f11a9f6d705b0463df3afeeda1491e324fd69f56

fvictorio: We really need a `no-unused-vars` linter warning, but we won't have one until we migrate to eslint :slightly_frowning_face:

fvictorio (approved):

PR #1525, opened 2021-06-08
Add support for custom errors

fvictorio:
Depends on #1452.


packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-stack-trace.ts
```diff
@@ -45,6 +46,7 @@ export interface SourceReference {
   contract?: string;
   function?: string;
   line: number;
+  range: [number, number];
```
alcuadrado, line 49: I don't get this. Why was this added? Pretty sure I'm missing something.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-stack-trace.ts
```diff
@@ -84,6 +86,13 @@ export interface PanicErrorStackTraceEntry {
   isInvalidOpcodeError: boolean;
 }
 
+export interface CustomErrorStackTraceEntry {
+  type: StackTraceEntryType.CUSTOM_ERROR;
+  message: string;
+  sourceReference: SourceReference;
+  isInvalidOpcodeError: boolean;
```
alcuadrado, line 93: I don't think this field is necessary. Aren't they always reverts?

packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-stack-trace.ts
```diff
@@ -84,6 +86,13 @@ export interface PanicErrorStackTraceEntry {
   isInvalidOpcodeError: boolean;
 }
 
+export interface CustomErrorStackTraceEntry {
+  type: StackTraceEntryType.CUSTOM_ERROR;
+  message: string;
```
alcuadrado, line 91: What's the message here? The return data?

packages/hardhat-core/src/internal/util/abi-helpers.ts
```diff
@@ -0,0 +1,46 @@
+import * as abi from "@ethersproject/abi";
+
+export class AbiHelpers {
```
alcuadrado, line 3: I like this

alcuadrado (commented):
This PR looks good. It's even simpler than what I had in mind :) 

I'm not sure how to review the tests though. Can you point me to the ones you think are important?

packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-stack-trace.ts
fvictorio: This is used in `ErrorInferrer.filterRedundantFrames`

packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-stack-trace.ts
fvictorio: Yes, both for panic and custom errors. I removed it in aced929927af780b1413b2b73a437f956db3a9b4.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-stack-trace.ts
fvictorio: This has the already parsed error message. I'll add a comment explaining that.

fvictorio:
> I'm not sure how to review the tests though. Can you point me to the ones you think are important?

The two new directories are [custom-errors](https://github.com/nomiclabs/hardhat/tree/custom-errors/packages/hardhat-core/test/internal/hardhat-network/stack-traces/test-files/0_8/custom-errors) and [panic-codes](https://github.com/nomiclabs/hardhat/tree/custom-errors/packages/hardhat-core/test/internal/hardhat-network/stack-traces/test-files/0_8/panic-codes). There's also the (skipped) [try-catch-uncaught](https://github.com/nomiclabs/hardhat/tree/custom-errors/packages/hardhat-core/test/internal/hardhat-network/stack-traces/test-files/0_8/special-cases/try-catch-uncaught) which is a known bug.

fvictorio:
I tested this with the recent 0.8.5 release of solidity and there is only one test failing. It's a very specific test, and it fails because a redundant frame is not properly removed. I don't think it's a blocker, though, so I'll open an issue to track that and expand on the failing test and why it fails.

alcuadrado (approved):
LGTM

PR #1526, opened 2021-06-09
Fix wrong error message when evm_setNextBlockTimestamp receives the c…

### LINKED ISSUE #1520: Wrong error message when `evm_setNextBlockTimestamp` receives the current timestamp
Using the timestamp of the latest block in `evm_setNextBlockTimestamp` returns this error:

```
InvalidInputError: Timestamp 1623172727 is lower than previous block's timestamp 1623172727
```

The message should say `is lower than or equal to`

RhnSharma:
Fixes #1520
Hi @fvictorio, how does this look?
Let me know if this needs any changes.
Thanks


fvictorio:
Hey @RhnSharma, thank you for this. I think we should be validating this error message. Do you feel like giving that a try?

This is the relevant test:

https://github.com/nomiclabs/hardhat/blob/2da1508b6ec3aa32ce063629bf55dc791ab10f03/packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts#L227-L244

Instead of using `.catch` like that, we should use the `assertInvalidInputError` helper. It is already imported in that file, and you can see how it's used in other tests.

If you don't want or don't have the time to do this, please let me know!

RhnSharma:
@fvictorio I tried to validate the error message. Take a look at it and let me know if this needs any changes. 
Could you please guide me a little on the changes to be made?
Thanks

fvictorio:
I think this part of what you've added is (mostly) correct:

```js
      await assertInvalidInputError(
        this.provider,
        "evm_setNextBlockTimestamp",
        [timestamp - 1],
        "Timestamp should be less than previous block's timestamp"
      );
```

but that should replace the whole block (i.e. the whole `this.provider.send(...).catch(...) statement). Also, the error message that it's asserting is wrong.

RhnSharma:
@fvictorio I made the changes. Are the changes correct now?
Please take a look at it and let me know if this needs any changes.
Thanks

fvictorio:
@RhnSharma The last argument to `assertInvalidInputError` is the error message you expect (not the assertion error message), so these tests are failing. Were you able to run them locally?

RhnSharma:
@fvictorio I have made the changes. The tests are no longer failing locally. 
Please take a look at it.
Thanks

fvictorio (approved):

fvictorio:
Thank you @RhnSharma!

PR #1531, opened 2021-06-09
Add tip for yarn fan

sniper365:


fvictorio:
Hey @sniper365, what do you mean? A lot of people use hardhat with yarn without major issues. Can you give us an example where it fails? Also, are you referring to yarn v1 or v2?

fvictorio:
I'm closing this for now since I don't think we'll add a warning like that without a good reason.

PR #1532, opened 2021-06-09
Add polygonscan.com to Etherscan network list

BrknRobot:


marcellobardus:
+1

fvictorio:
Ok, I was finally able to test this and it seems to work fine, but I couldn't verify a contract anyway. But I _think_ it's an etherscan issue, not related to this particular change, so I'll go ahead and merge it. Thank you @BrknRobot!

jsidorenko:
Could this be pushed to npm? Many thanks in advance!

fvictorio:
Hi @jsidorenko. Last time I tried, polygonscan verification API wasn't working. We are going to wait until it does before publishing this, since it doesn't make sense to release something that we know it won't work (even if the problem isn't on our side).

fvictorio:
We just released `@nomiclabs/hardhat-etherscan@2.1.3`, which includes this change. Thank you again @BrknRobot!

PR #1536, opened 2021-06-10
Rename hardhat_setStorageSlot to hardhat_setStorageAt

feuGeneA:


fvictorio:
`docs/hardhat-network/README.md` needs to be updated too

fvictorio:
Oh and another thing I noticed:

```
    if (positionIndex.gt(MAX_WORD_VALUE)) {
      throw new InvalidInputError(
        `Storage key must not be greater than 2^256. Received ${positionIndex.toString()}.`
      );
    }
```

I think that should be `gte`, because there are 2^256 slots and they start at 0? (I'm not 100% about this, but... 95% sure)

feuGeneA:
> Oh and another thing I noticed:
> 
> ```
>     if (positionIndex.gt(MAX_WORD_VALUE)) {
>       throw new InvalidInputError(
>         `Storage key must not be greater than 2^256. Received ${positionIndex.toString()}.`
>       );
>     }
> ```
> 
> I think that should be `gte`, because there are 2^256 slots and they start at 0? (I'm not 100% about this, but... 95% sure)

Great catch.  I'm with you on the 95%.  Addressed in https://github.com/nomiclabs/hardhat/pull/1536/commits/a5b434342cec110667fb31b7ccb1fd2f3b306215

fvictorio (approved):

PR #1540, opened 2021-06-12
Split the forking tests into their own job

alcuadrado:
🕯🙏


alcuadrado:
For the casual viewer: our CI fails because some requests to Alchemy timeout, but those are not reproducible outside of github actions.

PR #1542, opened 2021-06-13
Fork network name

wighawag:
This PR allow user to specify a network to fork from, instead of an url.
It will use the url set in the network config for thar network name

This is a very useful feature for plugin like `hardhat-deploy` because that way, the plugins can know which network it was forked from and bring the usual context for that network (named accounts, deployments, etc...)

This fixes https://github.com/nomiclabs/hardhat/issues/1164


alcuadrado:
We've been discussing this further with @fvictorio.

We had these alternative ideas that may be useful to fix the problem you described to me on tg.

1. Make the `HardhatArguments`, including the network name, and the `taskName` available in the config.
2. Make the `HardhatArguments` extensible, so that you could define, for example, `--fork-network`.

Note that they contradict each other and only one of them can be implemented, and that in both cases you would require an API to hook into the provider initialization process.

paulrberg:
This is not strictly related to the use case mentioned by @wighawag, but:

> Make the HardhatArguments, including the network name, and the taskName available in the config.

This will implicitly solve https://github.com/nomiclabs/hardhat/issues/1795, since I could check at the top-level scope for the network chosen by the user, choose either the `ETHERSCAN_API_KEY` or the `POLYGONSCAN_API_KEY` env var and pass it to the `apiKey` field.

sebastianst:
Any update on this?

alcuadrado:
Closing to clean up the repo. We still like this idea, and may implement it.

PR #1551, opened 2021-06-15
Split forking tests

alcuadrado:
This PR disables the Hardhat Network's forking tests in all of the CI jobs, and creates a new one that runs them with a single configuration (i.e. node version and os).

This isolates the CI's instability in a single job, so that any problem with Alchemy doesn't cancel the entire CI run. Furthermore, somehow making less requests in parallel makes the test run more reliable.  We should investigate why this happens, but in the meantime let's go for this approach so that we have a reliable CI once again.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -83,6 +83,7 @@ jobs:
         env:
           INFURA_URL: ${{ secrets.INFURA_URL }}
           ALCHEMY_URL: ${{ secrets.ALCHEMY_URL }}
+          DO_NOT_SET_THIS_ENV_VAR____IS_HARDHAT_CI: true
         run: yarn test
 
   test_macos:
@@ -114,6 +115,7 @@ jobs:
         env:
           INFURA_URL: ${{ secrets.INFURA_URL }}
           ALCHEMY_URL: ${{ secrets.ALCHEMY_URL }}
+          DO_NOT_SET_THIS_ENV_VAR____IS_HARDHAT_CI: true
         run: yarn test
 
   test_linux:
@@ -150,4 +152,5 @@ jobs:
         env:
           INFURA_URL: ${{ secrets.INFURA_URL }}
           ALCHEMY_URL: ${{ secrets.ALCHEMY_URL }}
+          DO_NOT_SET_THIS_ENV_VAR____IS_HARDHAT_CI: true
         run: yarn test
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -57,7 +57,7 @@ jobs:
             packages/*/node_modules
             packages/hardhat-core/test/internal/hardhat-network/stack-traces/compilers
             packages/hardhat-core/test/internal/hardhat-network/provider/.hardhat_node_test_cache
-          key: ${{ runner.os }}-node-${{ matrix.node }}-${{ hashFiles('yarn.lock') }}-${{ hashFiles('packages/hardhat-core/test/internal/hardhat-network/provider/node.ts') }}
+          key: ${{ runner.os }}-node-10-${{ hashFiles('yarn.lock') }}-${{ hashFiles('packages/hardhat-core/test/internal/hardhat-network/provider/node.ts') }}
       - name: Install node-gyp-cache
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/CI.yml
```diff
@@ -57,7 +57,7 @@ jobs:
             packages/*/node_modules
             packages/hardhat-core/test/internal/hardhat-network/stack-traces/compilers
             packages/hardhat-core/test/internal/hardhat-network/provider/.hardhat_node_test_cache
-          key: ${{ runner.os }}-node-${{ matrix.node }}-${{ hashFiles('yarn.lock') }}-${{ hashFiles('packages/hardhat-core/test/internal/hardhat-network/provider/node.ts') }}
+          key: ${{ runner.os }}-node-10-${{ hashFiles('yarn.lock') }}-${{ hashFiles('packages/hardhat-core/test/internal/hardhat-network/provider/node.ts') }}
```
alcuadrado, line 60: Fixing an error here. This isn't run with a matrix.

.github/workflows/CI.yml
```diff
@@ -80,9 +80,6 @@ jobs:
       - name: Install
         run: yarn --frozen-lockfile
       - name: Run tests
-        env:
```
alcuadrado, line 83: Deleting these env variables disables the forking tests.

.github/workflows/CI.yml
```diff
@@ -147,7 +141,48 @@ jobs:
...
+      - name: Install
+        run: yarn --frozen-lockfile
+      - name: Clean
+        run: yarn clean
+      - name: Run core tests
         env:
           INFURA_URL: ${{ secrets.INFURA_URL }}
           ALCHEMY_URL: ${{ secrets.ALCHEMY_URL }}
-        run: yarn test
+          DO_NOT_SET_THIS_ENV_VAR____IS_HARDHAT_CI: true
```
alcuadrado, line 183: This disables the HTTP timeouts

.github/workflows/CI.yml
```diff
@@ -147,7 +141,48 @@ jobs:
...
+          FORCE_COLOR: 3
```
alcuadrado, line 184: We need to force chalk to output colors for Hardhat Network's logging tests to work.

.github/workflows/CI.yml
```diff
@@ -147,7 +141,48 @@ jobs:
...
+        run: |
+          cd packages/hardhat-core
+          yarn build
+          npx mocha --recursive "test/internal/hardhat-network/{helpers,jsonrpc,provider}/*.ts"
```
alcuadrado, line 188: The reason to run them this way is that we only want to run the forking-related tests.

packages/e2e/.mocharc.json
```diff
@@ -1,5 +1,5 @@
 {
   "require": "ts-node/register/files",
   "ignore": ["test/fixture-projects/**/*"],
-  "timeout": 60000
+  "timeout": 180000
```
alcuadrado, line 4: I increased this timeout as it fails very often when running under macos in the ci

.github/workflows/CI.yml
```diff
@@ -147,7 +141,48 @@ jobs:
       - name: Clean
         run: yarn clean
       - name: Run tests
+        run: yarn test
+
+  # We should run this tests with multiple configurations
+  # but somehow the requests to Alchemy take much longer when doing that.
+  # As a temporal workaround, we run them with a single config.
```
feuGeneA, line 148:
```suggestion
  # As a temporary workaround, we run them with a single config.
```

.github/workflows/CI.yml
```diff
@@ -147,7 +141,48 @@ jobs:
...
+  # 
+  # Once we properly refactor Hardhat Network's tests we should
+  # add more configurations (different OS and node versions).
+  #
+  # We should also understand what's going on before blindly refactoring them.
+  test_fork:
+    name: Test Hardhat Network's fork
```
feuGeneA, line 155:
```suggestion
    name: Test Hardhat Network's forking functionality
```

alcuadrado:
Thanks for the suggestions, @feuGeneA :)

PR #1554, opened 2021-06-16
Easy fixes discovered while doing a deep dive on documentation

feuGeneA:
-   [x] Remove markdown from all _messages_ in `error-list.ts`.


docs/guides/compile-contracts.md
```diff
@@ -144,7 +146,7 @@ contracts
 └── Qux.sol
 ```
 
-the structure of your artifact directory then could look like this:
+then the structure of your artifact directory then would look like this:
```
fvictorio, line 149:
```suggestion
then the structure of your artifact directory would look like this:
```

fvictorio:
The changes in the `errors-list.ts` module are a bit risky. I for one think we shouldn't set a precedent of making those messages "public API", but it's still something to consider.

feuGeneA:
> The changes in the `errors-list.ts` module are a bit risky. I for one think we shouldn't set a precedent of making those messages "public API", but it's still something to consider.

In a call we concurred that because the public API consists of the error codes rather than the error messages, this is not a concern.

docs/getting-started/README.md
```diff
@@ -204,7 +204,7 @@ npx hardhat run scripts/sample-script.js --network localhost
 
 ---
 
-Congrats! You have created a project, ran a Hardhat task, compiled a smart contract, installed a Waffle integration plugin, wrote and ran a test using the Waffle and ethers.js plugins, and deployed a contract.
+Congrats! You have created a project, run a Hardhat task, compiled a smart contract, installed a Waffle integration plugin, written and run a test using the Waffle and ethers.js plugins, and deployed a contract.
```
alcuadrado, line 207: This is a question to learn from this: why is this in infinitive despite "created" being in simple past?

docs/guides/compile-contracts.md
```diff
@@ -132,7 +134,7 @@ You shouldn't interact with these files directly.
 
 The [HRE] has an `artifacts` object with helper methods. For example, you can get a list with the paths to all artifacts by calling `hre.artifacts.getArtifactPaths()`.
 
-You can also read an artifact using the name of the contract by calling `hre.artifacts.readArtifact("Bar")` and that will give us the content of the artifact for the `Bar` contract. This would only work if there was just one contract `Bar` in the whole project, but calling `hre.artifacts.readArtifact("Foo")`, would throw an error if there were two `Foo` contracts. To disambiguate this case, you would have to use the **Fully Qualified Name** of the contract: `hre.artifacts.readArtifact("contracts/Foo.sol:Foo")`.
+You can also read an artifact using the name of the contract by calling `hre.artifacts.readArtifact("Bar")` and that will yield the content of the artifact for the `Bar` contract. This would only work if there was just one contract named `Bar` in the whole project; it would throw an error if there were two. To disambiguate this case, you would have to use the **Fully Qualified Name** of the contract: `hre.artifacts.readArtifact("contracts/Bar.sol:Bar")`.
```
alcuadrado, line 137: I think I'd use return instead of yield, as `yield` has a very specific meaning within js.

packages/hardhat-core/sample-project/hardhat.config.js
```diff
@@ -3,7 +3,7 @@ require("@nomiclabs/hardhat-waffle");
 // This is a sample Hardhat task. To learn how to create your own go to
 // https://hardhat.org/guides/create-task.html
 task("accounts", "Prints the list of accounts", async () => {
-  const accounts = await ethers.getSigners();
+  const accounts = await hre.ethers.getSigners();
```
alcuadrado, line 6: What's the reason for this change? Normally people who prefer to use the global `hre` use its fields.

More context: We didn't define `global.hre` initially. We added it because it made giving support easier, as we know that `hre` will always be available for people who use the global stuff, and clear for people importing it explicitly.

docs/guides/create-task.md
```diff
@@ -223,7 +223,7 @@ Manually creating a `Promise` can look challenging, but you don't have to do tha
 
 ### Defining parameters
 
-Hardhat tasks can receive `--named` parameters with a value, `--flags`, positional and variadic parameters. Variadic parameters act like JavaScript's rest parameters. The Config DSL `task` function returns an object with methods to define all of them. Once defined, Hardhat takes control of parsing parameters, validating them, and printing help messages.
+Hardhat tasks can receive named parameters with a value (eg `--parameterName parameterValue`), flags with no value (eg `--flagName`), positional parameters, or variadic parameters. Variadic parameters act like JavaScript's rest parameters. The Config DSL `task` function returns an object with methods to define all of them. Once defined, Hardhat takes control of parsing parameters, validating them, and printing help messages.
```
alcuadrado, line 226:
```suggestion
Hardhat tasks can receive named parameters with a value (eg `--parameter-name parameterValue`), flags with no value (eg `--flag-name`), positional parameters, or variadic parameters. Variadic parameters act like JavaScript's rest parameters. The Config DSL `task` function returns an object with methods to define all of them. Once defined, Hardhat takes control of parsing parameters, validating them, and printing help messages.
```

Hardhat requires camelCased param names when defining a task, but expected them to be --like-this in the CLI.

alcuadrado: For example, `"paramName"` should be used to define a task, but when running it through the cli, you must use `--param-name`.

docs/guides/ganache-tests.md
```diff
@@ -1,16 +1,16 @@
...
+Rather than using Ganache, we recommend using the built-in [Hardhat Network](../hardhat-network/README.md) network to test your
 smart contracts, as it generates [combined JavaScript and Solidity stack traces](../hardhat-network/README.md#solidity-stack-traces),
 making debugging easier.
 
 If you still want to run your tests using Ganache, you can do it in two ways.
 
 ## Manually running Ganache
 
-You don't need to do anything especial to use Ganache if you don't want to.
+You don't need to do anything special to use Ganache if you don't want to.
```
alcuadrado, line 11: Any tick to remember the difference between special and especial?

docs/guides/migrate-from-buidler.md
```diff
@@ -6,7 +6,7 @@ This guide will teach you how to migrate your project from Buidler into Hardhat.
 
 ## Installing the Hardhat packages
 
-The first thing you need to do to migrate your project, is installing the new npm packages.
+The first thing you need to do to migrate your project is to install the new npm packages.
```
alcuadrado, line 9: When should we remove this section? @fvictorio @fzeoli

docs/guides/typescript.md
```diff
@@ -66,7 +66,7 @@ require("@nomiclabs/hardhat-waffle");
 // This is a sample Hardhat task. To learn how to create your own go to
 // https://hardhat.org/guides/create-task.html
 task("accounts", "Prints the list of accounts", async () => {
-  const accounts = await ethers.getSigners();
+  const accounts = await hre.ethers.getSigners();
```
alcuadrado, line 69: I think it's valuable to show that if you were using a global hre field, you have to turn it into an explicit access. But now that I re-read this, it wasn't explicit either.

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -963,9 +963,9 @@ If you aren't overriding compilation-related tasks, please report this as a bug.
     INVALID_SOURCE_NAME_BACKSLASHES: {
       number: 1002,
       message:
-        "Invalid source %name%. The source name uses backslashes (\\) instead of slashes (/).",
+        "Invalid source %name%. The source name uses backslashes (\\\\) instead of slashes (/).",
```
alcuadrado, line 966: I tried this in a node console

```
console.log("Invalid source %name%. The source name uses backslashes (\\\\) instead of slashes (/).")
```

and it printed: Invalid source %name%. The source name uses backslashes (\\) instead of slashes (/).

alcuadrado:
Great work, @feuGeneA !

docs/guides/migrate-from-buidler.md
fzeoli: I think we could delist it already, and if someone asks through support channels we can share it

docs/getting-started/README.md
feuGeneA: In this case "run" is not an infinitive, but rather a past participle. Technically the form here is "have run."

docs/guides/ganache-tests.md
feuGeneA: Honestly I would have said that "especial" is not even a word, so that says something, but I just searched and apparently it is. Of course there's "especially", but there's no (commonly used) adjective form of that adverb.  Interestingly, "specially" seems to exist, but is rarely (never?) used; "specifically" seems to be the commonly used equivalent of "specially."

docs/getting-started/README.md
alcuadrado: Ohhh! Makes sense! Thanks

packages/hardhat-core/src/internal/core/errors-list.ts
feuGeneA: At first I was confused by your comment, but then I reproduced what you're saying, and I saw the problem, and then I edited your comment in order to surround your output with backticks (`), and now the problem is visible in your comment as well.

I wrote this code before you explained to me the difference between `message` and `description`, but now that you've explained it to me, I understand the reason for your comment here.

Addressed in https://github.com/nomiclabs/hardhat/pull/1554/commits/5be5b9d9e422488f3b7dc246b29888265b6e8e1f

docs/guides/compile-contracts.md
feuGeneA: Good call, thank you. Addressed in https://github.com/nomiclabs/hardhat/pull/1554/commits/71e5f8202f01aabe58a0aedc5bc5230f13de46e6

packages/hardhat-core/sample-project/hardhat.config.js
feuGeneA: I brought this up on a call with @fvictorio . I complained that it felt weird to me that we'd use `ethers` unqualified like that (with no import), and if I remember correctly he said that it was an oversight and that we should be using `hre.`.

docs/guides/typescript.md
feuGeneA: I changed this for the same reasons described in https://github.com/nomiclabs/hardhat/pull/1554#discussion_r660831402.

packages/hardhat-core/sample-project/hardhat.config.js
fvictorio: @alcuadrado notice that this is within a task in the hardhat config. It seems very weird to me that the global variables from the `hre` are available there (although I understand why that happens).

That being said, I just now noticed that we should also add the hre to parameters:

```
task("accounts", "Prints the list of accounts", async (taskArgs, hre) => {
```

docs/guides/ganache-tests.md
fvictorio: So the rule of thumb would be to always use `special` and `especially` and it will be fine 99% of the time?

feuGeneA: Yes, and I'd even add another 9 to that! (99.9%)

Also, though, don't forget about "specifically," as it's very common.

packages/hardhat-core/sample-project/hardhat.config.js
feuGeneA: I added the parameters in https://github.com/nomiclabs/hardhat/pull/1554/commits/5f11182769677497a57031b790baea1939fc7bf9

docs/guides/typescript.md
feuGeneA: I updated this in https://github.com/nomiclabs/hardhat/pull/1554/commits/5f11182769677497a57031b790baea1939fc7bf9 in response to https://github.com/nomiclabs/hardhat/pull/1554#discussion_r661469038

packages/hardhat-core/sample-project/hardhat.config.js
alcuadrado: Looks much better with the params!

What I found it confusing is that it was using the `hre` global variable, instead of directly going for `ethers`.

docs/guides/migrate-from-buidler.md
feuGeneA: De-listed in https://github.com/nomiclabs/hardhat/pull/1554/commits/967e92181fd3bdcecc90b840b0193886bae829af

alcuadrado (approved):

PR #1558, opened 2021-06-17
Bump the supported solidity version range

fvictorio:


packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
```diff
@@ -814,6 +823,18 @@ describe("Stack traces", function () {
...
 });
 
+describe("Solidity support", function () {
+  it("check that the latest tested version is within the supported version range", async function () {
+    const latestSupportedVersion = solidityCompilers
+      .map((sc) => sc.solidityVersion)
+      .sort(semver.compare)[solidityCompilers.length - 1];
+
+    assert.isTrue(
+      semver.satisfies(latestSupportedVersion, SUPPORTED_SOLIDITY_VERSION_RANGE)
```
fvictorio, line 833: This doesn't check that the range is tight. For example, if we increment it to `<=0.8.5` now, this test will continue to pass. AFAIK there's no way to do this with the `semver` module, but something we can do is to add some tests that bump the patch, minor and major version of `lastSupportedVersion` and assert that they are **not** within the range.

alcuadrado: We can add it by bumping a minor and the next major's `x.0.0`. That should cover most cases.

fvictorio: Why wouldn't we do it for the minor too? It's way more likely than a next major (I don't know when there will be a solidity 1.0.0) and it's just an extra line. Am I missing something here?

alcuadrado: Sorry, I meant patch. Also, i always forget that solc versions are always `0.x.y`.

We can do this. If  `x.y.z` is the latest one, we can test with `x+1.0.0`, `x.y+1.0`, and `x.y.z+1`

fvictorio: That's my original idea, I must explained it really badly :sweat_smile: I'll do that then!

fvictorio: 44c1baee50e5d5a0251ffdef75c930601379c9d2

packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
```diff
@@ -814,6 +823,39 @@ describe("Stack traces", function () {
...
+
+    assert.isTrue(
+      semver.satisfies(
+        latestSupportedVersion,
+        SUPPORTED_SOLIDITY_VERSION_RANGE
+      ),
+      `Expected ${latestSupportedVersion} to be within the ${SUPPORTED_SOLIDITY_VERSION_RANGE} range`
+    );
+
+    const nextPatchVersion = semver.inc(latestSupportedVersion, "patch")!;
```
alcuadrado, line 840: Oh, this was easier than expected! Nice!

alcuadrado (approved):

PR #1559, opened 2021-06-17
Add FIFO mode to the transaction pool

ngotchac:
This adds a new parameter to the Hardhat config: `networks.hardhat.txpool` which can be an object:
```ts
{
  fifo: boolean;
}
```

Setting `fifo` to `true` will overwrite the default txpool ordering, so that transactions are mined in a FIFO queue.

This is especially useful when using the fork feature of Hardhat (great feature BTW), if you want to replay a block up to a certain point. Since some blocks on ETH mainnet are not ordered by decreasing gasprice anymore (cf. Flashbots), we have to find a way for the user to mine transactions in a specific, non-traditional, order.

I'm not sure about the naming of the config though...

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node-types.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node-types.ts
@@ -3,6 +3,7 @@ import { RunBlockResult } from "@ethereumjs/vm/dist/runBlock";
 import { BN } from "ethereumjs-util";
 
 import { BuildInfo } from "../../../types";
+import { HARDHAT_MEMPOOL_SUPPORTED_ORDERS } from "../../constants";
 import { MessageTrace } from "../stack-traces/message-trace";
 
 import type { ReturnData } from "./return-data";
@@ -28,6 +29,7 @@ interface CommonConfig {
   initialDate?: Date;
   tracingConfig?: TracingConfig;
   initialBaseFeePerGas?: number;
+  mempoolOrder: MempoolOrder;
 }
 
 export type LocalNodeConfig = CommonConfig;
@@ -47,6 +49,8 @@ export interface TracingConfig {
 }
 
 export type IntervalMiningConfig = number | [number, number];
+
+export type MempoolOrder = typeof HARDHAT_MEMPOOL_SUPPORTED_ORDERS[number];
 
 export interface GenesisAccount {
   privateKey: string;
--- a/packages/hardhat-core/src/types/config.ts
+++ b/packages/hardhat-core/src/types/config.ts
@@ -13,6 +13,7 @@
 // trying to augment the config types.
 
 import type { BN } from "ethereumjs-util";
+import { HARDHAT_MEMPOOL_SUPPORTED_ORDERS } from "../internal/constants";
 
 // Networks config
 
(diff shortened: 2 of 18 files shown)


alcuadrado:
Hey @ngotchac,

This is a great idea! We normally prefer to discuss new features in an issue first, but I guess we can do it here.

Maybe we should have a setting that let's you choose between different tx sorting algos instead. The only two that come to my mind are by gas price and fifo though. Can something like this be extended in the future without a breaking change?

ngotchac:
Hey, yeah sorry I just saw this new template for PRs, so I should have created an issue first :/ I can close this PR in the meanwhile if needed.
Otherwise, yeah I think we could have some kind of `string` in the `txpool` object (with a key like `sort` or `sortAlgo`) instead of the current `boolean`. Then it can be easily passed down to the node. I was also wondering whether it would make sense to expose it via an RPC call? So one can switch on the fly. I'm open to suggestions :)

fvictorio:
I haven't checked the code yet, but with respect to the config, I feel it should be inside the current `mining` field? That is:

```
hardhat: {
  mining: {
    auto: false,
    interval: 5000,
    mempool: {
      order: 'fifo' | 'priority'
    }
  }
}
```

ngotchac:
Hey @fvictorio, this really a better config name than what I had :p I'll update to match those.

alcuadrado:
> ```
>     mempool: {
>       order: 'fifo' | 'priority'
>     }
> ```

My only concern with this is that adding a new order option would be a breaking change.

fvictorio:
> My only concern with this is that adding a new order option would be a breaking change.

Because of the type or for some other reason?

ngotchac:
Yeah I'm not sure I get it either. It could be a breaking change if one of the order algo is removed in a future version I guess ; otherwise all the config's fields are optional.

alcuadrado:
Sorry I wasn't clear. The breaking change would be because of the types.

fvictorio:
> The breaking change would be because of the types.

Would that still be an issue if we use `string` as the type instead of `"fifo" | "priority"`?

ngotchac:
Hey, just to let you know that I rebased this PR, so it should now be merge-able.

alcuadrado:
Completely random comment that came to my minad while reviewing this: this is the PR 1559 🥳

packages/hardhat-core/src/internal/hardhat-network/provider/node-types.ts
```diff
@@ -48,6 +50,8 @@ export interface TracingConfig {
 
 export type IntervalMiningConfig = number | [number, number];
 
+export type MempoolOrder = typeof HARDHAT_MEMPOOL_SUPPORTED_ORDERS[number];
```
alcuadrado, line 53: Sweet!

packages/hardhat-core/src/types/config.ts
```diff
@@ -170,11 +171,22 @@ export interface HttpNetworkHDAccountsConfig {
...
 
 export interface HardhatNetworkMiningUserConfig {
   auto?: boolean;
   interval?: number | [number, number];
+  mempool?: HardhatNetworkMempoolUserConfig;
+}
+
+export interface HardhatNetworkMempoolConfig {
+  // order should be on of `HARDHAT_MEMPOOL_SUPPORTED_ORDERS` entries
+  order: typeof HARDHAT_MEMPOOL_SUPPORTED_ORDERS[number];
```
alcuadrado, line 185: I'm going to change this to `string`. We recently [had a similar conversation](https://github.com/nomiclabs/hardhat/pull/1928/files/199053ebb11514426e8a4d152bf33c680ca1cd99#r746783321) where we decided to use a slightly unsafe to avoid exposing something from `internal`.

Also, i think having `string` here makes me feel more comfortable with the possibility of adding a new option in the future. The reason is that this string would be guaranteed at runtime to have a valid value, so expanding the set of values won't be a compilation error.

alcuadrado:
I'll add docs for this in a new PR

PR #1561, opened 2021-06-17
Improve e2e tests

fvictorio:
This makes the e2e tests run in two "modes" of execution: installing hardhat with `npm` or `yarn`. Both are used when you run `yarn test`.

The code is not great but hopefully the comments help.

Something I realized while doing this is that... maybe everything should be inside the `run-tests.js` script? Because now the call sequence is something like this:

- Run the script
- The script uses shelljs to do a bunch of stuff to setup everything
- The script uses shelljs to execute mocha
  - The mocha tests uses shelljs to execute hardhat
- The script uses shelljs to cleanup the fixture projects

While it would be simpler to remove mocha and to just do inside the script itself what the test does. But I didn't want to invest a lot of time on this, so I'll leave it as it is.


.github/workflows/e2e-ci.yml
```diff
@@ -4,13 +4,15 @@ on:
   push:
     branches: [$default-branch]
     paths:
+      - "packages/e2e/**/*"
```
alcuadrado, line 7: Good catch!

packages/e2e/run-tests.js
```diff
@@ -0,0 +1,101 @@
...
+const shell = require("shelljs");
+
+shell.set("-e");
+
+const rootDir = path.join(__dirname, "..", "..");
+const hardhatCoreDir = path.join(rootDir, "packages", "hardhat-core");
+const fixtureProjectsDir = path.join(__dirname, "test", "fixture-projects");
+
+if (process.argv[2] !== "npm" && process.argv[2] !== "yarn") {
+  console.error("Usage: node prepare-tests.js <npm|yarn>");
```
alcuadrado, line 12:
```suggestion
  console.error("Usage: node run-tests.js <npm|yarn>");
```

packages/e2e/run-tests.js
```diff
@@ -0,0 +1,101 @@
...
+      "package.json.template"
+    );
+    const packageJsonPath = path.join(fixtureDir, "package.json");
+    shell.cp(packageJsonTemplatePath, packageJsonPath);
+
+    if (isYarn) {
+      shell.exec(`yarn add ${hardhatPackagePath} --no-save`);
+    } else {
+      shell.exec(`npm install ${hardhatPackagePath} --no-save`);
+    }
```
alcuadrado, line 76: Can we remove the hardhat package here?

alcuadrado:
It would be great to have an explanation of how these tests work. Maybe in the readme? Otherwise you have to read the code to understand what's in this package.

fvictorio:
> It would be great to have an explanation of how these tests work. Maybe in the readme? Otherwise you have to read the code to understand what's in this package.

Yes, totally. I actually already forgot most of it. I'll write something :+1:

packages/e2e/run-tests.js
fvictorio: Done.

feuGeneA (dismissed):
Please add `packages/e2e/test/fixture-projects/*/{yarn.lock,package.json,package.json.lock}` to `.gitignore

fvictorio:
Done. It seems like `.gitignore` doesn't accept the `{a,b}` syntax though :slightly_frowning_face:

feuGeneA:
> Done. It seems like `.gitignore` doesn't accept the `{a,b}` syntax though

Ah yeah I was hopeful but not expectant haha. I think it's a bash glob thing?

alcuadrado (approved):

PR #1562, opened 2021-06-18
Test increaseTime behaviour

gpetraglia:
Regarding the discussion on https://github.com/nomiclabs/hardhat/issues/1079:

Make explicit some increaseTime behaviour that were not so clear with
the previous tests.
These behaviour include:
 - If increaseTime is called N times, all of the calls are taken into
account to calculate the new block timestamp
 - If real time elapses, that's also taken into account

The decision to keep this is partially due to the fact that we don't
want to break previously working workflow.

------

There will be another PR that updates the documentation, keeping in mind the needs for documentation PRs

### DIFF
--- a/packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
@@ -411,6 +411,18 @@ describe("HardhatNode", () => {
     describe("timestamp tests", () => {
       let clock: sinon.SinonFakeTimers;
 
+      const assertIncreaseTime = async (expectedTime: number) => {
+        const block = await node.getLatestBlock();
+        const blockTimestamp = block.header.timestamp.toNumber();
+
+        // We check that the time increased at least what we had expected
+        // but allow a little bit of POSITIVE difference(i.e. that the
+        // actual timestamp is a little bit bigger) because time may have ellapsed
+        // We assume that the test CAN NOT have taken more than a second
+        assert.isAtLeast(blockTimestamp, expectedTime);
+        assert.isAtMost(blockTimestamp, expectedTime + 1);
+      };
+
       beforeEach(() => {
         clock = sinon.useFakeTimers(Date.now());
       });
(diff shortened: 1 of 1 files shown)


packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -5,8 +5,10 @@ import VM from "@ethereumjs/vm";
 import { AfterBlockEvent, RunBlockOpts } from "@ethereumjs/vm/dist/runBlock";
 import { assert } from "chai";
 import { Address, BN, bufferToHex } from "ethereumjs-util";
+import { pass } from "fp-ts/lib/Writer";
```
fvictorio, line 8: This doesn't seem to be used.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -5,8 +5,10 @@ import VM from "@ethereumjs/vm";
...
 import sinon from "sinon";
 
+import { getClosestCallerPackage } from "../../../../internal/util/caller-package";
```
fvictorio, line 11: Neither does this.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -495,12 +511,29 @@ describe("HardhatNode", () => {
...
+        await assertIncreaseTime(now + delta);
+      });
+
+      it("mining a block having increaseTime called twice counts both calls", async () => {
+        const now = getCurrentTimestamp();
+        const delta = 30;
+        node.increaseTime(new BN(delta));
+        node.increaseTime(new BN(delta));
         await node.mineBlock();
+        await assertIncreaseTime(now + delta * 2);
```
fvictorio, line 525: `assertIncreaseTime` calls `mineBlock`, so I guess the previous line should be removed?

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -495,12 +511,29 @@ describe("HardhatNode", () => {
...
-        const block = await node.getLatestBlock();
-        const blockTimestamp = block.header.timestamp.toNumber();
-        assert.equal(blockTimestamp, now + 30);
+      it("mining a block having called increaseTime takes into account 'real' passing time", async () => {
+        const now = getCurrentTimestamp();
+        const delta = 30;
+        const elapsedTimeInSeconds = 3;
+        node.increaseTime(new BN(delta));
+        clock.tick(elapsedTimeInSeconds * 1_000);
+        await node.mineBlock();
```
fvictorio, line 534: Ditto here.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
fvictorio: Actually, I have a slight preference for removing the `mineBlock` call from the `assertIncreaseTime` and having to call it explicitly in the test.

gonzalpetraglia: Ok, will do that

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
gpetraglia: Vscode might pulled a little sneaky on me. Fixed

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
gpetraglia: Fixed

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
gpetraglia: Fixed

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
gpetraglia: Fixed

fvictorio (approved):
Thanks @gonzalpetraglia, I like these tests :rocket:

fvictorio:
Closing this in favor of #1682. I think I can't update this PR because it's created from the master branch of the fork.

PR #1563, opened 2021-06-18
Document some 'edge' cases of evm_increaseTime

gpetraglia:
IncreaseTime had some unclear behaviour which was documented in this PR and tested in another PR(https://github.com/nomiclabs/hardhat/pull/1562).

The original discussion came out of this issue https://github.com/nomiclabs/hardhat/issues/1079

### DIFF
--- a/docs/hardhat-network/README.md
+++ b/docs/hardhat-network/README.md
@@ -426,7 +426,7 @@ To customise it, take a look at [the configuration section](/config/README.md#ha
 
 #### Special testing/debugging methods
 
-- `evm_increaseTime` – same as Ganache.
+- `evm_increaseTime` – same as Ganache(Note that this method takes into account 'real' time ellapsing and if it is called multiple times, all of the calls are taken into account).
 - `evm_mine` – same as Ganache
 - `evm_revert` – same as Ganache.
 - `evm_snapshot` – same as Ganache.


fvictorio:
What do you mean by "takes into account 'real' time ellapsing"?

fvictorio:
Ugh, I see what you mean now. I always thought that it was just `last timestamp + increased value`, not that it was forwarding time like that.

We should probably recommend people to always use `evm_setNextBlockTimestamp`, `evm_increaseTime` is a mess.

gonzalpetraglia:
Yes, that's what I thought. It is quite a mess. That's why I wanted to create a new method to have one that actually behaves like that, `last timestamp + increased value`

gonzalpetraglia:
Do you think I should go for a more throughout explanation here, with examples? 

Or just recommend that they don't use it?

The thing with evm_setNextBlockTimestamp is that it does require more things from the dev, like knowing the last timestamp and  doing some extra calculations. Maybe that could be handled in a plugin tho

fvictorio:
I replied in the other issue too, but: I'd rather not add more time-related RPC methods. I understand that `setNextBlockTimestamp` can be a bit harder to use though. To me, the proper solution is to have a library (that might be in core or not, I'm not sure) for time-related stuff, so you can use a higher-level API instead of raw RPC calls.

fvictorio:
Btw, we are working on having a proper reference section with more detailed explanations of each RPC method, so maybe we should just include this on that (cc @feuGeneA). I'm going to close this PR since I think we'll have that new section soon enough.

PR #1587, opened 2021-06-24
Migrate from tslint to eslint

### LINKED ISSUE #843: Migrate to eslint
tslint has been deprecated for some time, and the typescript ecosystem is moving to eslint. We should do it too.

We use a few non-standard rules that may make the migration harder. They are:
* [`only-buidler-error`](https://github.com/nomiclabs/buidler/blob/1974956585b4d6ae240bdb3ec50e31d7f478a1dc/packages/buidler-core/tslint/onlyBuidlerErrorRule.ts)
* [`naming-convention`](https://github.com/ajafff/tslint-consistent-codestyle/blob/master/docs/naming-convention.md)
* [`no-unnecessary-else`](https://github.com/ajafff/tslint-consistent-codestyle/blob/master/docs/no-unnecessary-else.md)

fvictorio:
Closes #843.

This PR removes tslint, adds eslint with a config as similar as possible, and enables the `@typescript-eslint/no-unused-vars` rule. The initial work was done with [tslint-to-eslint-config](https://github.com/typescript-eslint/tslint-to-eslint-config).

This only lints typescript files right now. Linting both javascript and typescript files is, of course, supported by eslint, but it was more involved (you need to setup your config to only lint js and then override the config for ts files) and since tslint was only linting typescript files anyway I decided to do the same.

### `only-hardhat-error`

We had a custom tslint rule: [`only-hardhat-error`](https://github.com/nomiclabs/hardhat/blob/8047ce1e14581e6fba13cfeef099d4a38bcd0f0f/packages/hardhat-core/tslint/onlyHardhatErrorRule.ts). Tslint supports a custom directory with rules. Eslint, on the other hand, requires that you use a plugin to add custom rules. So I created a new package: `@nomiclabs/eslint-plugin`.

Right now it only has the `only-hardhat-error` rule. Also, tslint supports typescript out of the box, but eslint doesn't. To make things simpler, I implemented the rule in plain js, but we should probably migrate it to typescript when/if we add more custom eslint rules.

### Decisions

I had to make a bunch of decisions during this migration:

- [`@typescript-eslint/no-var-requires`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-var-requires.md) seems to be different than tslint's [`no-var-requires`](https://palantir.github.io/tslint/rules/no-var-requires/). It seems like tslint allows dynamic requires inside functions but eslint doesn't, and it can't be configured. We can enable it anyway and disable the rule on a per-line basis.
- I disabled the [`arrow-body-style`](https://eslint.org/docs/rules/arrow-body-style) because it has a minor (but annoying) difference with what tslint was doing: it disallows arrow functions with a single return statement. Since we have a bunch of those, I preferred to just disable it. Later I found out that `eslint-plugin-prettier` [also disables it](https://github.com/prettier/eslint-plugin-prettier#arrow-body-style-and-prefer-arrow-callback-issue), so it seems like it was the right call.
- I removed the [`eslint-plugin-jsdoc`](https://github.com/gajus/eslint-plugin-jsdoc) plugin. We had a [`jsdoc-format`](https://palantir.github.io/tslint/rules/jsdoc-format/) rule in tslint, but the plugin seems to have a lot of differences. I started to disable each rule that was giving errors that tslint didn't... until I found out that I had disabled all of them. So I just removed the whole plugin.
- In tslint we had the [`member-access`](https://palantir.github.io/tslint/rules/member-access/) rule, while in eslint we use [`explicit-member-accessibility`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-member-accessibility.md). The latter seems more powerful, allowing things like "explicit/implicit" accessibility for constructors or getters. I decided to have explicit accessibility for getters (that is, you always have to do `public get ...`) and implicit for constructors (that is, you can't use `public constructor` because it's redundant). This is very easy to change and I think this rule supports autofix, so we can change this.
- Some nested `try/catch` were triggering the `no-shadow` rule. I did the lazy thing and just renamed the inner ones to `err` or `e` (I don't remember right now).

### `no-unused-vars`

I added the `@typescript-eslint/no-unused-vars` rule. This meant more changes, but this is a really useful rule to have, so I decided to do it.

There were a lot of easy to fix warnings (mainly unused imports or destructured properties). There were also some cases where some variables (especially parameters) should've been deleted, but I didn't feel so comfortable doing it. So I configured the rule so that ignores variables that start with `_` and I fixed those cases by adding a leading underscore to the name.

### `tslint-to-eslint-config` log

<details>
  <summary>This is the log from <code>tslint-to-eslint-config</code></summary>
  
  ```
7 ESLint rules behave differently from their TSLint counterparts:
  * arrow-body-style:
    - ESLint will throw an error if the function body is multiline yet has a one-line return on it.
  * no-invalid-this:
    - Functions in methods will no longer be ignored.
  * @typescript-eslint/no-unused-expressions:
    - The TSLint optional config "allow-new" is the default ESLint behavior and will no longer be ignored.
  * no-underscore-dangle:
    - Leading and trailing underscores (_) on identifiers will now be ignored.
  * no-redeclare:
    - ESLint does not support check-parameters.
  * import/no-extraneous-dependencies:
    - ESLint does not support whitelisting modules
  * padded-blocks:
    - ESLint's padded-blocks rule also bans a blank line before a closing brace.

6 rules are not known by tslint-to-eslint-config to have ESLint equivalents:
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "naming-convention".
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "no-inferred-empty-object-type".
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "no-unnecessary-else".
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "only-hardhat-error".
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "prettier".
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "strict-comparisons".  
  ```
</details>


packages/hardhat-waffle/src/waffle-chai.ts
```diff
@@ -64,17 +64,19 @@ export function waffleChai(chai: Chai.ChaiStatic, utils: Chai.ChaiUtils) {
 }
 
 function supportCalledOnContract(Assertion: Chai.AssertionStatic) {
+  // eslint-disable-next-line import/no-extraneous-dependencies
```
fvictorio, line 67: I'm not sure about this. Shouldn't `chai` be a peer dependency of `hardhat-waffle`?

packages/hardhat-core/test/internal/hardhat-network/helpers/retrieveForkBlockNumber.ts
```diff
@@ -1,10 +1,9 @@
-/* tslint:disable:no-string-literal */
 import { ForkBlockchain } from "../../../../src/internal/hardhat-network/provider/fork/ForkBlockchain";
 import { HardhatNetworkProvider } from "../../../../src/internal/hardhat-network/provider/provider";
 
 export async function retrieveForkBlockNumber(
   provider: HardhatNetworkProvider
-) {
+): Promise<number> {
```
fvictorio, line 6: I had to do this because otherwise `@typescript-eslint/restrict-plus-operands` was giving me false positives in a lot of tests. I have no idea why, maybe it was a bug. Anyway, this fixed it :man_shrugging:

config/eslint/eslintrc.js
```diff
@@ -0,0 +1,205 @@
+module.exports = {
+  env: {
+    browser: true,
```
alcuadrado, line 3:
```suggestion
    browser: false,
```

I tried this locally and the linter passed. Not sure why it was `true`.

config/eslint/eslintrc.js
```diff
@@ -0,0 +1,205 @@
...
+      },
+    ],
+    "@typescript-eslint/prefer-for-of": "error",
+    "@typescript-eslint/prefer-function-type": "error",
+    "@typescript-eslint/prefer-namespace-keyword": "error",
+    "@typescript-eslint/restrict-plus-operands": "error",
+    "@typescript-eslint/strict-boolean-expressions": [
+      "error",
+      {
+        allowAny: true,
```
alcuadrado, line 138: I think this should be set to `false`, but it'd require a few changes. Should we open an issue about this?

config/eslint/eslintrc.js
```diff
@@ -0,0 +1,205 @@
...
+      "error",
+      {
+        path: "always",
+        types: "prefer-import",
+        lib: "always",
+      },
+    ],
+    "@typescript-eslint/unified-signatures": "error",
+    "constructor-super": "error",
+    eqeqeq: ["error", "always"],
```
alcuadrado, line 151: this name 😂

config/eslint/eslintrc.js
```diff
@@ -0,0 +1,205 @@
...
+    "no-caller": "error",
+    "no-cond-assign": "error",
+    "no-debugger": "error",
+    "no-duplicate-case": "error",
+    "no-duplicate-imports": "error",
+    "no-eval": "error",
+    "no-extra-bind": "error",
+    "no-new-func": "error",
+    "no-new-wrappers": "error",
+    "no-return-await": "error",
```
alcuadrado, line 180: I'm ok with this rule for now, but TBH I'd rather enforce awaits everywhere, as that produces better async stack traces.

https://twitter.com/a0viedo/status/1305911417355333633

packages/e2e/prepare-tests.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
alcuadrado, line 1: weird. Shouldn't `fs-extra` be a dependency here?

packages/hardhat-core/.eslintignore
```diff
@@ -0,0 +1 @@
+test/fixture-projects/**/*
```
alcuadrado, line 1: I think we can use the same as in `.prettierignore` instead:

```suggestion
/test/fixture-projects/**/artifacts
/test/fixture-projects/**/cache
```

packages/hardhat-core/.gitignore
```diff
@@ -107,3 +107,5 @@ typings/
 .dynamodb/
 
 test/internal/hardhat-network/provider/.hardhat_node_test_cache/
+
+!.eslintrc.js
```
alcuadrado, line 111: Is this really needed?

alcuadrado: Also, our custom ignore rules should be above the "Below is Github's node gitignore template" comment

packages/hardhat-ganache/.eslintignore
```diff
@@ -0,0 +1 @@
+test/fixture-projects/**/*
```
alcuadrado, line 1: Same here. Should we ignore cache and artifacts instead?

packages/hardhat-waffle/.eslintignore
```diff
@@ -0,0 +1 @@
+test/fixture-projects/**/*
```
alcuadrado, line 1: same here

packages/hardhat-waffle/src/waffle-chai.ts
alcuadrado: I think this is meant to import the version installed by `ethereum-waffle`

alcuadrado:
Great work! I love how many unnecessary imports went away.

config/eslint/eslintrc.js
fvictorio: Agree :+1: https://github.com/nomiclabs/hardhat/issues/1595

config/eslint/eslintrc.js
fvictorio: Makes sense to me: https://github.com/nomiclabs/hardhat/issues/1596

packages/e2e/prepare-tests.js
fvictorio: Yeah, makes sense. I just moved all the `devDependencies` to `dependencies` since this is a private package.

packages/hardhat-core/.eslintignore
fvictorio: Fixed in 106b9c443b454483fe8bdac40f2263f85af8bc60

packages/hardhat-core/.gitignore
fvictorio: Yeah, it's needed because of the `/*.js` ignore before.

Agree with having our rules above the comment, fixed.

packages/hardhat-core/.eslintignore
fvictorio: I might need to revert this. If we do that, we get this error:

```
Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: test/fixture-projects/broken-typescript-config-project/hardhat.config.ts.
The file must be included in at least one of the projects provided
```

If I then use an override to disable `parserOptions.project` for the `**/fixture-projects/**/*.ts` files, eslint complains because:

```
Error while loading rule '@typescript-eslint/await-thenable': You have used a rule which requires parserServices to be generated. You must therefore provide a value for the "parserOptions.project" property for @typescript-eslint/parser.
```

There surely are ways to fix this. For example, we could split the main config into two: one with the rules that require type information and the other with the rest, and then we'd only use the latter for linting the fixture projects. But I don't think the value added would be worth it.

fvictorio: Tentatively reverted, but let me know if you'd rather lint those files anyway.

alcuadrado: tbh, now I wonder if we were ever linting those ts files.

Reverting it looks like the best trade-off.

packages/hardhat-core/.gitignore
alcuadrado: ohh! `/*.js`, I see.

alcuadrado (approved):

PR #1591, opened 2021-06-25
Fix LATEST_DEPENDENCY_VERSIONS workflow

alcuadrado:
This PR fixes a Github Actions workflow that's run with the latest version of every dependency.  This workflow is scheduled to run every 8 hours.

The changes are:
- Upgrade node to version 12, as solcjs has a bug with node 10.
- Install the correct docker image of vyper
- Stop installing `node-gyp-cache`
- Use a better workflow name. No idea why I used a THAT_NAME in the first place.
- Removed the forking-related tests, as those are too unstable to run in a cron job, and those don't use anything new wrt dependencies.

Note that I tested the workflow by making it run on push, and then rebased this PR.

### DIFF
--- a/.github/workflows/LATEST_DEPENDENCY_VERSIONS.yml
+++ b/.github/workflows/LATEST_DEPENDENCY_VERSIONS.yml
@@ -1,4 +1,4 @@
-name: LATEST_DEPENDENCY_VERSIONS
+name: Run the entire test suite with the latest dependency versions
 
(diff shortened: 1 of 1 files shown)


alcuadrado:
OT: I think this is the first PR I make that doesn't unnecessarily run the test suite 🥳

fvictorio (approved):

PR #1592, opened 2021-06-26
add hardhat_getForkedChainId JSON-RPC method

aftermathdigital:
- [ ] (If this PR includes a **documentation change**) This PR's branch was created from the `website` branch, and this PR uses the `website` branch as its base branch.
- [ ] (If this PR includes a **bug fix**) Relevant tests have been included.
- [X] (If this PR includes a **new feature**) The change was previously discussed on an Issue or with someone from the team.

---

Add the `hardhat_getForkedChainId` JSON-RPC method - which:

- In the case this chain is forked, returns the chainId of the chain from which it was forked.
- In the case this chain is not forked, returns null

Includes a documentation update describing the functionality.


packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
```diff
@@ -110,9 +113,18 @@ async function getNetworkId(provider: HttpProvider) {
   const networkIdString = (await provider.request({
     method: "net_version",
   })) as string;
+
   return parseInt(networkIdString, 10);
 }
 
+async function getChainId(provider: HttpProvider) {
+  const chainIdString = (await provider.request({
+    method: "eth_chainId",
```
alcuadrado, line 122: I forgot to mention that this method may not be available. Take a look at how we handle this here: https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/core/providers/chainId.ts#L11

packages/hardhat-core/test/internal/hardhat-network/jsonrpc/client.ts
```diff
@@ -74,6 +74,7 @@ describe("JsonRpcClient", () => {
           fakeProvider as any,
           1,
           123,
+          1,
```
aftermathdigital, line 77: I don't understand why this chainId couldn't be 'undefined', but the test returned a slightly incorrect storage value when the chainId wasn't `1`

aftermathdigital: nevermind, I was being a clown :D

fvictorio:
@alcuadrado this is ready (assuming tests will pass). The only thing I'm not sure about is the returned value of the new method. Right now it returns a number, while `eth_chainId` returns a hex string. IMO we should return that format too. What do you think?

fvictorio:
I think this is no longer necessary since we implemented https://github.com/NomicFoundation/hardhat/pull/3382?

Sorry @aftermathdigital for not merging it :disappointed:

aftermathdigital:
> I think this is no longer necessary since we implemented #3382?
> 
> Sorry @aftermathdigital for not merging it 😞

No worries, was fun to work on it all the same.

Congratulations on the football!

PR #1624, opened 2021-06-30
Set strict boolean expressions allow any false in eslintrc

### LINKED ISSUE #1595: Set `allowAny: false` in `@typescript-eslint/strict-boolean-expressions`
See https://github.com/nomiclabs/hardhat/pull/1587#discussion_r658935665

istareatscreens:
- [x] (If this PR includes a **new feature**) The change was previously discussed on an Issue or with someone from the team.

---

This PR sets "strict boolean expressions" allow-any to false in eslintrc and fixes all linting errors that occur as a result as requested in #1595


packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -65,7 +65,7 @@ export class Artifacts implements IArtifacts {
     } catch (error) {
       if (
         typeof error.message === "string" &&
-        error.message.includes("no matching file exists")
+        (error.message as string).includes("no matching file exists")
```
fvictorio, line 68: There are several of these, so if we end up doing what I mentioned before, we should create a helper, something like `assertErrorIsObject` or whatever.

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -469,18 +469,21 @@ export function getArtifactFromContractOutput(
   contractName: string,
   contractOutput: any
 ): Artifact {
-  const evmBytecode = contractOutput.evm && contractOutput.evm.bytecode;
+  const evmBytecode =
+    Boolean(contractOutput.evm) && contractOutput.evm.bytecode;
```
fvictorio, line 473:
```suggestion
  const evmBytecode = contractOutput?.evm?.bytecode;
```

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -469,18 +469,21 @@ export function getArtifactFromContractOutput(
...
   let bytecode: string =
-    evmBytecode && evmBytecode.object ? evmBytecode.object : "";
+    Boolean(evmBytecode) && Boolean(evmBytecode.object)
```
fvictorio, line 475: Same comment here and in the rest of the changes that use `Boolean` (we should use `?.` and `??` instead).

packages/hardhat-core/src/internal/core/providers/backwards-compatibility.ts
```diff
@@ -54,7 +54,7 @@ export class BackwardsCompatibilityProviderAdapter extends EventEmitterWrapper
       }
 
       response.error = {
-        code: error.code ? +error.code : -1,
+        code: Boolean(error.code) ? +error.code : -1,
```
fvictorio, line 57:
```suggestion
        code: error.code !== undefined ? +error.code : -1,
```

packages/hardhat-core/src/internal/hardhat-network/jsonrpc/server.ts
```diff
@@ -72,7 +72,7 @@ export class JsonRpcServer implements IJsonRpcServer {
 
   public close = async () => {
     await Promise.all([
-      new Promise((resolve, reject) => {
+      new Promise<void>((resolve, reject) => {
```
fvictorio, line 75: This doesn't seem necessary?

fvictorio (changes requested):
I didn't finish reviewing the PR, but in general my concern is that we shouldn't fix these errors with type casting. Instead, we should have better (runtime) checks. Adding a new assertion helper for checking that some undefined value is an object should probably help in more than half of these cases.

packages/hardhat-core/src/internal/hardhat-network/jsonrpc/server.ts
istareatscreens: If not changed eslint throws an error

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -1452,7 +1452,8 @@ function needsCompilation(
 
 function hasCompilationErrors(output: any): boolean {
   return (
-    output.errors && output.errors.some((x: any) => x.severity === "error")
+    (output.errors ?? []).length !== 0 &&
+    output.errors.some((x: any) => x.severity === "error")
```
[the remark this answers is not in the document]
istareatscreens: Yeah I agree that is definitely cleaner

packages/hardhat-core/src/internal/hardhat-network/jsonrpc/server.ts
fvictorio: Really? That's weird, I changed it and it didn't complain :/ Maybe it's a setup thing?

istareatscreens: Highly likely, should I discard this change? I feel like it does no harm.

packages/hardhat-core/src/internal/artifacts.ts
istareatscreens: I may be wrong but line 472 seems like a bug since it should be returning an object but it is returning a boolean.

istareatscreens:
Removed Boolean wraps with undefined checks as suggested using ?? and ?. 

Did some refactoring on the code I thought was redundant or verbose

Waiting for decision on how to handle string casts

packages/hardhat-core/src/builtin-tasks/flatten.ts
```diff
@@ -42,7 +42,11 @@ function getSortedFiles(dependenciesGraph: DependencyGraph) {
     const sortedNames = [...new Set(withEntries)];
     return sortedNames.map((n) => filesMap[n]);
   } catch (error) {
-    if (error.toString().includes("Error: There is a cycle in the graph.")) {
+    if (
```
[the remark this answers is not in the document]
feuGeneA: I vote yes.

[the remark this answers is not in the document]
alcuadrado: Sure, let's upgrade it! 🚀

packages/hardhat-core/src/internal/hardhat-network/jsonrpc/server.ts
fvictorio: Yeah, I'd rather not have unnecessary type annotations.

fvictorio:
Thanks a lot for working on this @istareatscreens! I'll take another look as soon as I can.

packages/hardhat-core/src/internal/hardhat-network/jsonrpc/server.ts
istareatscreens: Ok removed

packages/hardhat-core/src/internal/artifacts.ts
istareatscreens: When doing the assertion I get the error "Property 'message' does not exist on type 'object'", can I just use any instead of unknown as the error type?

```
    } catch (error: unknown) {
      assertHardhatInvariant(
        typeof error === "object" && error !== null,
        "Error is not an object"
      );
      if (
        (error?.message?.includes("no matching file exists") as boolean) ?? //Error on error?.message
        false
      ) {
```

packages/hardhat-core/src/builtin-tasks/flatten.ts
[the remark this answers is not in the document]
istareatscreens: When doing the assertion I get the error "Property 'message' does not exist on type 'object'", can I just use any instead of unknown? This is after updating to prettier 2.1.2 not that that should matter

```
    } catch (error: unknown) {
      assertHardhatInvariant(
        typeof error === "object" && error !== null,
        "Error is not an object"
      );
      if (typeof error.message === "string" &&
        error.message.includes("no matching file exists") //error occurs here
      ) 
```

packages/hardhat-core/src/internal/artifacts.ts
istareatscreens: When doing the assertion I get the error "Property 'message' does not exist on type 'object'".  Removing unknown declaration removes this issue (leaving it as any)

```
    } catch (error: unknown) {
      assertHardhatInvariant(
        typeof error === "object" && error !== null,
        "Error is not an object"
      );
      if (
        (error?.message?.includes?.("no matching file exists") as boolean) ?? //error here on error?.message
        false
      ) {
```

fvictorio:
@alcuadrado @feuGeneA my commits are a bit of a mess, sorry. I didn't realize that the prettier script wasn't running on typescript files, so the formatting part is split into two commits. I'll explain each one of my commits here:

- 33ebdb72f14028e8efc556668184eba4fe1d9577 just removes the options for that rule, since `allowAny: false` is the default.
- 813f6e300d7f7fd5707d4a5bfbe9ae05e03f097d upgrades and runs prettier (but not in ts files, because I'm an idiot). I ended up using version 2.3.2 because the number of files changed wasn't that different, and I guess that if we upgrade, it's better to just use the latest version.
- a0978facde6409c1bc9e1176e758747e9e64edd6 is the most important commit, since it's the one that has non-cosmetic changes. I typed some errors as `unknown` and added an `assertIsError` helper.
- c7ac57370d2f5069d9b03ae1717a11d17d3b3e50 runs prettier in all ts files. You can probably ignore that one. Btw, I thought about adding the `ts` extension to the prettier scripts, but I guess that one was left out intentionally since that is checked by eslint, right? That makes sense to me, since otherwise we would be running prettier twice in each typescript file when we run the lint script.

istareatscreens:
Please let me know if you require any more assistance on this. I just wanted to thank you for being patient with me as this was my first contribution to an opensource project.

fvictorio:
@istareatscreens I think we can take it from here, but thanks a lot for working on this!

fvictorio:
Closing this in favor of #3220.

Sorry for not merging this one @istareatscreens, but it has a lot of merge conflicts so I just did it from scratch. Thanks a lot for the initial work on this!

PR #1628, opened 2021-07-02
Fix linter error for clean CI run

feuGeneA:
Because I noticed that CI was failing on the `master` branch.


alcuadrado (approved):

fvictorio:
The problem is that the linter doesn't run for external PRs, and I always forgot to click "Approve and run". I wonder if there's some way to have a smarter configuration for external contributions?

PR #1629, opened 2021-07-02
Replace TxPriotityHeap with TransactionQueue, which supports EIP-1559

### LINKED ISSUE #1609: Update the mempool sorting algorithm to support EIP-1559
- Transactions are ordered by `gasPrice` (if pre-1559) or `maxPriorityFeePerGas`
- Transactions with a maxFeePerGas below the base fee are not mined. If automining is enabled, these transactions are immediately rejected.

alcuadrado:
This PR fixes #1609.


packages/hardhat-core/test/internal/hardhat-network/provider/TransactionQueue.ts
```diff
@@ -0,0 +1,303 @@
...
+        const tx9 = createTestTransaction({
+          gasPrice: 92,
+        });
+
+        const tx10 = createTestTransaction({
+          gasPrice: 91,
+        });
+
+        const queue = new TransactionQueue(
+          makeOrderedTxMap([tx1, tx2, tx3, tx4, tx5, tx6, tx7, tx8, tx9, tx10])
```
fvictorio, line 137: Something I always think about these kinds of tests... should this array be shuffled?

packages/hardhat-core/test/internal/hardhat-network/provider/TransactionQueue.ts
```diff
@@ -0,0 +1,303 @@
...
+        assert.equal(queue.getNextTransaction(), tx1.data);
+        assert.equal(queue.getNextTransaction(), tx2.data);
+        assert.equal(queue.getNextTransaction(), tx3.data);
+        assert.equal(queue.getNextTransaction(), tx4.data);
+        queue.removeLastSenderTransactions();
+
+        assert.equal(queue.getNextTransaction(), tx6.data);
+        assert.equal(queue.getNextTransaction(), tx7.data);
+        assert.equal(queue.getNextTransaction(), tx8.data);
+        queue.removeLastSenderTransactions();
```
fvictorio, line 297: This doesn't match the description in the comments, I think you meant to call this after tx7?

packages/hardhat-core/test/internal/hardhat-network/provider/TransactionQueue.ts
alcuadrado: Yes! I'm glad you raised this. I had the exact same thought, and wasn't sure if we should do that.

packages/hardhat-core/test/internal/hardhat-network/provider/TransactionQueue.ts
alcuadrado: Right. Good catch!

packages/hardhat-core/src/internal/hardhat-network/provider/TransactionQueue.ts
```diff
@@ -0,0 +1,121 @@
...
+  left: OrderedTransaction,
+  right: OrderedTransaction,
+  baseFee?: BN
+) {
+  const leftEffectiveMinerFee = getEffectiveMinerFee(left, baseFee);
+  const rightEffectiveMinerFee = getEffectiveMinerFee(right, baseFee);
+
+  const cmp = leftEffectiveMinerFee.cmp(rightEffectiveMinerFee);
+  return cmp === 0 ? right.orderId - left.orderId : cmp;
+}
```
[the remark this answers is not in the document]
alcuadrado: Good point. I replaced `MaxHeap` by just `Heap`, which uses an standard comparator, and made our comparator more explicit.

alcuadrado:
Note: The forking functionality hasn't been adapted to London yet, and its tests are not expected to pass for now.

fvictorio (approved):

PR #1637, opened 2021-07-07
Add more links to ethers documentation

fvictorio:
Some users still don't quite understand that ethers.js is a separate project. I added more links to ethers docs in the plugin docs as a first step towards making this more clear.


alcuadrado (approved):
Good idea

PR #1647, opened 2021-07-10
Multiple adjustments for London

### LINKED ISSUE #1598: Add an initialBaseFeePerGas config to Hardhat Network
This option doesn't make sense if the HF is not at least London. Should we throw in that case?

Alternative names to consider `baseFeePerGas`, `initialBaseFee`, `baseFee`.

### LINKED ISSUE #1614: Handle base fee of the first block after the genesis correctly
- The base fee of first block after genesis is tricky
    - The base fee is reduced because the genesis block has 0 gas used. This is probably not what users want when they set an initial base fee.

alcuadrado:
_(disclaimer: not proud about this PR, it has a few unrelated things, but I was struggling to make progress)_

This PR:

* Normalize `baseFee` and `baseFeePerGas` to `baseFeePerGas` everywhere.
* Implements an `initialBaseFeePerGas` config in Hardhat Network.
* Partially fixes the ForkedBlockchain so that it can work with EIP-1559 networks. The `Common` is important when parsing remote blocks now.
* Adds a workaround for Erigon's https://github.com/ledgerwatch/erigon/issues/2288
* Removes a `console.log` that was accidentally left in the codebase.
* Improves and adds tests to the new-ish `hardforks.ts` utils.
* Updates the config validation so that you can't use EIP-1559-specific fields if you are not using London, and the other way round too.
* Implements logic in `HardhatNode` to arbitrarily set the next block's base fee per gas.
* Tests that the base fee increases according to EIP-1559 with and without forking a remote network.
* Validates that EIP-1559 txs can't be send if not running in London.
* Updates the default gasPrice/maxFee logic to match what Geth does.
* Fixes how HardhatNode decides if a txs should be mineable.
* Updates the default initial base fee per gas.
* Fixes a large amount of tests that had hardcoded gasPrices.
* Fix a breaking change introduced in Immutable.js
* Fixes the TxPool class so that it also works with fake eip1559 txs.
* Adds info about the base fee per gas to hardhat network's logging.
* Fix debug_traceTransaction so that it also works with EIP-1559 txs.
* Validates that automined EIP-1559 txs pay enough fees, rejecting them otherwise.
* Workarounds a limitation of @ethereumjs/vm that arises from #1666 
* Adds a `baseFeePerGas` field to the JSON-RPC representation of blocks.

## Explanation of the initial base fee logic

This deserves an explanation, as it's more complex that I thought it would be.

* If you are not forking a network, the `initialBaseFeePerGas` sets the `baseFeePerGas` of the genesis block. If none is provided, we use `1gwei`, as that's what the EIP actually uses.
* If you are forking a remote network things get more tricky, as always.
  * This setting now controls the first local block.
  * If you provider a value, it's used for that block.
  * If you don't provider one, two things can happen
     * If the remote network hasn't activated EIP-1559 we default to `1gwei`.
     * If the remote network has activated EIP-1559 we use the `baseFeePerGas` that EIP-1559 dictates, based on the block you are forking from.

Fixes #1614 
Fixes #1598


packages/hardhat-core/src/internal/hardhat-network/provider/output.ts
```diff
@@ -178,8 +178,13 @@ export function getRpcBlock(
     timestamp: numberToRpcQuantity(new BN(block.header.timestamp)),
     transactions,
     uncles: block.uncleHeaders.map((uh: any) => bufferToRpcData(uh.hash())),
-    baseFeePerGas: numberToRpcQuantity(block.header.baseFeePerGas ?? 0),
   };
+
+  if (block.header.baseFeePerGas) {
```
alcuadrado, line 183: I think we always need to return it if london is active. We should check what geth does.

alcuadrado: i.e. what should the `else` be?

packages/hardhat-core/src/internal/util/hardforks.ts
```diff
@@ -1,10 +1,58 @@
...
  * that is, if it includes all its changes.
  */
-export function hardforkGte(hardforkA: string, hardforkB: string): boolean {
-  const common = new Common({ chain: "mainnet" });
-  return common.hardforkGteHardfork(hardforkA, hardforkB);
+export function hardforkGte(
+  hardforkA: HardforkName,
+  hardforkB: HardforkName
+): boolean {
+  const indexA = HARDFORKS_ORDER.lastIndexOf(hardforkA);
```
alcuadrado, line 54: Add todo explaining that we don't want to load ethereumjs here

packages/hardhat-core/src/internal/hardhat-network/provider/output.ts
alcuadrado: My bad, just checked with Geth and it doesn't show this field for non-eip-1559 blocks.

packages/hardhat-core/src/internal/util/hardforks.ts
alcuadrado: Done

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -162,6 +175,19 @@ export class HardhatNode extends EventEmitter {
       initialBlockTimeOffset = new BN(
         getDifferenceInSeconds(new Date(forkBlockTimestamp), new Date())
       );
+
+      if (hardforkGte(hardfork, HardforkName.LONDON)) {
```
fvictorio, line 179: I think this block merits a comment. Something like:

> In some scenarios we set the `baseFeePerGas` of the first block that will be mined after the fork

(Which scenarios are these is clear enough after reading the code)

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -417,6 +458,7 @@ Hardhat Network's forking functionality only works with blocks from at least spu
     }
 
     await this._resetNextBlockTimestamp();
+    this._resetUserProvidedNextBlockBaseFeePerGas();
```
fvictorio, line 461: Unrelated to this PR, but why is `_resetNextBlockTimestamp` async? Worth fixing it here (assuming it's a mistake) just because the inconsistency is more obvious after this change.

fvictorio (approved):
All comments are very minor, so I'm pre-approving this.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: Done

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: Done

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -1211,16 +1211,31 @@ You can use them by running Hardhat Network with 'hardfork' ${ACCESS_LIST_MIN_HA
...
           ? this._rpcAccessListToNodeAccessList(rpcTx.accessList)
           : [];
 
+      if (rpcTx.maxPriorityFeePerGas === undefined) {
+        rpcTx.maxPriorityFeePerGas = await this._node.getMaxPriorityFeePerGas();
+      }
+
+      if (rpcTx.maxFeePerGas === undefined) {
+        // If we are here EIP-1559 is active and this can't be undefined
+        const baseFeePerGas = (await this._node.getNextBlockBaseFeePerGas())!;
```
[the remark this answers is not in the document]
alcuadrado: I created `assertHardhatNetworkInvariant`, as we shouldn't be throwing `HardhatError`s from Hardhat Network.

fvictorio (approved):

PR #1676, opened 2021-07-15
Implement ReadOnlyValidEIP1559Transaction

alcuadrado:
This PR implements the EIP-1559 version of the transaction class that we use to read remote txs when forking.

This completes most of the forking functionality changes required for London.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/fork/rpcToTxData.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/fork/rpcToTxData.ts
@@ -6,14 +6,13 @@ import { RpcTransaction } from "../../../core/jsonrpc/types/output/transaction";
 export function rpcToTxData(
   rpcTransaction: RpcTransaction
 ): TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData {
+  const isEip1559 = rpcTransaction.type?.eqn(2) ?? false;
+
   return {
     gasLimit: rpcTransaction.gas,
     // NOTE: RPC EIP-1559 txs still have this field for backwards compatibility,
     //  but FeeMarketEIP1559TxData doesn't.
-    gasPrice:
-      rpcTransaction.type?.eqn(2) ?? false
-        ? undefined
-        : rpcTransaction.gasPrice,
+    gasPrice: isEip1559 ? undefined : rpcTransaction.gasPrice,
     to: rpcTransaction.to ?? undefined,
     nonce: rpcTransaction.nonce,
     data: rpcTransaction.input,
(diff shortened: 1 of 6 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/fork/rpcToTxData.ts
```diff
@@ -1,34 +1,33 @@
...
-): TxData | AccessListEIP2930TxData {
-  const txData: AccessListEIP2930TxData = {
+): TxData | AccessListEIP2930TxData | FeeMarketEIP1559TxData {
+  return {
     gasLimit: rpcTransaction.gas,
-    gasPrice: rpcTransaction.gasPrice,
+    // NOTE: RPC EIP-1559 txs still have this field for backwards compatibility,
+    //  but FeeMarketEIP1559TxData doesn't.
+    gasPrice:
+      rpcTransaction.type?.eqn(2) ?? false
```
fvictorio, line 14: I can't create a suggestion for this, but I would extract this to a variable:

```ts
  const isEIP1559 = rpcTransaction.type?.eqn(2) ?? false;

  return {
    gasLimit: rpcTransaction.gas,
    // NOTE: RPC EIP-1559 txs still have this field for backwards compatibility,
    //  but FeeMarketEIP1559TxData doesn't.
    gasPrice: isEIP1559 ? undefined : rpcTransaction.gasPrice,
```

packages/hardhat-core/src/internal/hardhat-network/provider/transactions/ReadOnlyValidEIP1559Transaction.ts
```diff
@@ -0,0 +1,159 @@
...
+  FeeMarketEIP1559TxData,
+  FeeMarketEIP1559ValuesArray,
+} from "@ethereumjs/tx/src/types";
+import { InternalError } from "../../../core/providers/errors";
+
+/* eslint-disable @nomiclabs/only-hardhat-error */
+
+/**
+ * This class is like `ReadOnlyValidTransaction` but for
+ * EIP-2930 (access list) transactions.
```
fvictorio, line 15: This comment seems to be wrong. There are also other comments that refer to 2930 instead of 1559.

alcuadrado: Done

packages/hardhat-core/src/internal/hardhat-network/provider/fork/rpcToTxData.ts
alcuadrado: Done

fvictorio (approved):

PR #1681, opened 2021-07-16
Added run function to hello-world example task

naddison36:
Added the `run` function to the example `hello-world` task so it will compile.


fvictorio:
Thanks! I modified it a bit to make it more explicit.

naddison36:
Thanks. The docs are now much clearer.

PR #1632, opened 2021-07-05
Add Advanced sample project

feuGeneA:
-  [x] This PR includes a **new feature**, and the change was previously discussed on an Issue or with someone from the team.
-  [ ] Verify and demonstrate usage of hardhat-etherscan.
-  [ ] Resolve (fix or ignore) outstanding lint errors.
-  [ ] Add automated tests of project creation, including the exercising of clean runs of tests and linting in the newly created project.

PR #1686, opened 2021-07-16
Command-line-interface-level test of the sample project's suggested cmds

feuGeneA:
-  [x] Move test out of `hardhat-core` and into `e2e`.


feuGeneA:
I'd appreciate some feedback on possible sources of the problem exhibited in that Windows CI failure.

alcuadrado:
My suspicion is that `npx` is installing `hardhat` instead of using the local one. Try using `yarn` instead.

packages/hardhat-core/src/internal/cli/project-creation.ts
```diff
@@ -239,26 +242,37 @@ export async function createProject() {
...
-      return;
-    }
+  const useDefaultPromptResponses =
+    process.env.HARDHAT_CREATE_SAMPLE_PROJECT_WITH_DEFAULTS !== undefined;
+
+  if (useDefaultPromptResponses) {
+    responses = {
+      projectRoot: process.cwd(),
+      shouldAddGitIgnore: true,
+      shouldAddGitAttributes: true,
```
fvictorio, line 252: We should remove this variable, it's not used anymore now that GitHub has support for Solidity.

fvictorio:
Some thoughts about this:

- I think that, when possible, we should try to avoid adding code for test-related reasons, like the `HARDHAT_CREATE_SAMPLE_PROJECT_WITH_DEFAULTS` environment variable in this case. But I can't think of an alternative here, so maybe it's fine in this case.
- Wouldn't it be better if this test lived in the `e2e` package? I don't see any advantage in having it as part of core, and the `e2e` package already has the packaging logic ready, and it works both with npm and yarn. This would mean waiting until #1561 is merged, though.

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -85,11 +85,7 @@ async function main() {
 
     showStackTraces = hardhatArguments.showStackTraces;
 
-    if (
-      hardhatArguments.config === undefined &&
-      !isCwdInsideProject() &&
-      process.stdout.isTTY === true
-    ) {
+    if (hardhatArguments.config === undefined && !isCwdInsideProject()) {
```
alcuadrado, line 88: Why is the `isTTY` check removed? The reason for it was to avoid things like piping hardhat to hang forever.

feuGeneA: Right, I meant to call this out before review, thanks for noticing.

The reason it's removed is that when hardhat is run to generate the sample project, during the scope of this test, it's being run via a node.js child process, and there is no TTY in that case, so it wouldn't fall into the project generation logic without this change.

alcuadrado: Maybe `process.stdout.isTTY || process.env.HARDHAT_CREATE_SAMPLE_PROJECT_WITH_DEFAULTS !== undefined`?

packages/hardhat-core/src/internal/cli/project-creation.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1686/commits/afdbe683bf2f10cda8d770728c94bf56fa9b7c5e

feuGeneA:
> Some thoughts about this:
> 
> * I think that, when possible, we should try to avoid adding code for test-related reasons, like the `HARDHAT_CREATE_SAMPLE_PROJECT_WITH_DEFAULTS` environment variable in this case. But I can't think of an alternative here, so maybe it's fine in this case.

I discussed this with @alcuadrado and we decided this was the most pragmatic approach, after ruling out an additional but undocumented command-line parameter like `--create-project-with-defaults`.  Thinking more, though, perhaps it would be nice to have an additional, _documented_ command-line param, like `--non-interactive`, akin to yarn's `--yes`.  What do you think?

> * Wouldn't it be better if this test lived in the `e2e` package? I don't see any advantage in having it as part of core, and the `e2e` package already has the packaging logic ready, and it works both with npm and yarn. This would mean waiting until [Improve e2e tests #1561](https://github.com/nomiclabs/hardhat/pull/1561) is merged, though.

The issue I have with what's happening in `e2e` right now (and what's happening in that PR) is that it's using its own "fixture" project rather than exercising the sample project that actually gets produced by the CLI.  Since I'm planning on making some pretty significant changes to that generation process, over in #1632 , I want a test that exercises the generated projects.  And doing it within `hardhat-core` made sense to me because I wanted the tests as close to the changes as possible.  `e2e` seems like the right place to test custom/non-vanilla projects (including a minimal/basic one, as you've got in the current fixture), but as a test of an unmodified, freshly-generated-by-the-CLI project, `hardhat-core` feels right to me.  I'm open to having my mind changed, but that's my line of thinking.

fvictorio:
> Thinking more, though, perhaps it would be nice to have an additional, documented command-line param, like --non-interactive, akin to yarn's --yes. What do you think?

That's an interesting idea, and it could even make sense for users. I'm not a representative user, but I create hardhat projects all the time so it would be useful to me at least :stuck_out_tongue: But seriously, I know that some power users tend to create new hardhat projects frequently too.

The thing to decide would be if we have a single flag (like `--yes`) that installs some pre-defined project, or if we have a `--project` parameter that receives some project identifier. Of course, we could have both. We should also think about backwards compatibility (does that apply to `hh` being executed in a non-project directory?) and about the `hh` shorthand's autocomplete (can we extend it to include this params?)

I wouldn't do this in this PR though, but if you agree this is a valuable feature, I can create an issue for it.

> The issue I have with what's happening in e2e right now (and what's happening in that PR) is that it's using its own "fixture" project rather than exercising the sample project that actually gets produced by the CLI. Since I'm planning on making some pretty significant changes to that generation process, over in #1632 , I want a test that exercises the generated projects. 

That's just the current situation, but I don't see why there couldn't be other tests that use a temporary directory, install the built package (like it's done right now in the fixture projects) and run some checks. The main reason I think this is a good idea is that the "build the tgz and install" approach is a bit closer to the real thing (that is, it's more end-to-end, the next best thing would be to actually use a registry). By moving them to the e2e we would also have a single approach to testing the whole thing (instead of doing the `add file://` thing in one place and the `yarn build` thing in another place).

feuGeneA:
> My suspicion is that `npx` is installing `hardhat` instead of using the local one. Try using `yarn` instead.

I tried that but it didn't fix the problem.  I dug deeper and found the problem, which I transformed from a `TODO` to an Issue: #1698 .

To get a clean CI run, I simply skipped this test on Windows, in https://github.com/nomiclabs/hardhat/pull/1686/commits/4a22f2eede20345ae16016565cc81206940bd6d9

packages/hardhat-core/src/internal/cli/cli.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1686/commits/2a8901ff84eb2c2894df3e60bf63b4752d167487

feuGeneA:
> I wouldn't do this in this PR though, but if you agree this is a valuable feature, I can create an issue for it.

I personally agree it would be valuable.

> That's just the current situation, but I don't see why there couldn't be other tests that use a temporary directory, install the built package (like it's done right now in the fixture projects) and run some checks. The main reason I think this is a good idea is that the "build the tgz and install" approach is a bit closer to the real thing (that is, it's more end-to-end, the next best thing would be to actually use a registry). By moving them to the e2e we would also have a single approach to testing the whole thing (instead of doing the `add file://` thing in one place and the `yarn build` thing in another place).

Thank you for elaborating. You've convinced me. I'll move the test over there.

packages/e2e/test/fixture-projects/sample-project/package.json.template
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "basic-project",
+  "version": "1.0.0",
+  "license": "MIT",
+  "dependencies": {}
+}
```
fvictorio, line 6: Why is this file needed?

packages/e2e/test/index.ts
```diff
@@ -34,4 +36,44 @@ describe("e2e tests", function () {
...
+        shell.exec(`${hardhatBinary}`, {
+          env: {
+            ...process.env,
+            HARDHAT_CREATE_SAMPLE_PROJECT_WITH_DEFAULTS: "true",
+          },
+        });
+        shell.exec(`${hardhatBinary} accounts`);
+        shell.exec(`${hardhatBinary} compile`);
+        shell.exec(`${hardhatBinary} test`);
+        shell.exec("node scripts/sample-script.js");
```
fvictorio, line 71: I think these commands are actually not working. The reason the tests pass anyway is that `shelljs`, by default, doesn't throw when a command run with `shell.exec` exits with a non-zero code. I don't blame you for not realizing this, it's bitten me more times than I can remember :sweat: 

The quick fix here is to use `shell.set("-e");` at the beginning. I don't really like this approach because it feels like a global thing (my guess is that changes the whole module behavior and I don't understand the implications of that). The alternative is to do what I do in the previous tests and assert the value of the exit code.

feuGeneA: They actually were working, but I changed one to fail, and you're right, failures were not being handled correctly.  Addressed in https://github.com/nomiclabs/hardhat/pull/1686/commits/df3ed5454a55dd21bdfc1d44564caf790da36c68

packages/e2e/test/fixture-projects/sample-project/package.json.template
feuGeneA: I created it because I was receiving this error:
```
Error: cp: no such file or directory: /home/gene/dev/nomiclabs/hardhat/packages/e2e/test/fixture-projects/sample-project/package.json.template
    at Object.error (/home/gene/dev/nomiclabs/hardhat/node_modules/shelljs/src/common.js:110:27)
    at /home/gene/dev/nomiclabs/hardhat/node_modules/shelljs/src/cp.js:246:14
    at Array.forEach (<anonymous>)
    at Object._cp (/home/gene/dev/nomiclabs/hardhat/node_modules/shelljs/src/cp.js:243:11)
    at Object.cp (/home/gene/dev/nomiclabs/hardhat/node_modules/shelljs/src/common.js:384:25)
    at setup (/home/gene/dev/nomiclabs/hardhat/packages/e2e/run-tests.js:70:11)
    at main (/home/gene/dev/nomiclabs/hardhat/packages/e2e/run-tests.js:25:3)
    at Object.<anonymous> (/home/gene/dev/nomiclabs/hardhat/packages/e2e/run-tests.js:96:1)
    at Module._compile (internal/modules/cjs/loader.js:936:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:947:10)
```

fvictorio: Ugh, right, I had forgotten about that :sweat_smile:

packages/e2e/test/index.ts
```diff
@@ -34,4 +40,38 @@ describe("e2e tests", function () {
...
+    // of the individual suggested commands, but because the project creation
+    // takes so long (really it's the dependency installation that's so time
+    // consuming) even a single test already has a crazy long run time, so for
+    // expediency this is one big complex test.
+
+    useFixture("sample-project");
+
+    it("should permit successful execution all of the suggested commands", async function () {
+      if (os.type() === "Windows_NT") {
+        // cf. https://github.com/nomiclabs/hardhat/issues/1698
```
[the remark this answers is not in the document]
alcuadrado: btw, I love what you did here: creating an issue about it to add a comment here

alcuadrado (approved):
This new version is so simple :)

packages/e2e/test/index.ts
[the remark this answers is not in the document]
feuGeneA: https://en.wikipedia.org/wiki/Cf.

Thanks to your question, I realized I was using this abbreviation wrong. :sweat_smile: I've changed it: https://github.com/nomiclabs/hardhat/pull/1686/commits/b510b904e26ce4ea86c45b219d73b915c3858324

PR #1687, opened 2021-07-17
London updates to eth_sendTransaction and eth_sendRawTransaction

### LINKED ISSUE #1599: Update `eth_sendTransaction` to support EIP-1559
- It should work with legacy (`gasPrice`) transactions
- It should work with 1559 transactions
- Invalid combinations should be rejected (e.g., both `gasPrice` and `maxFeePerGas`)
- Missing one of the 1559 fields should be rejected or handled [like geth](https://github.com/ethereum/pm/issues/328#issuecomment-853234014) (we need to decide this)
- Having no field should be handled somehow (we need to decide this)

### LINKED ISSUE #1600: Update `eth_sendRawTransaction` to support EIP-1559
See #1599

alcuadrado:
This PR updates the validations and default values of these methods, and tests them.

Fixes #1599 
Fixes #1600

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
@@ -84,6 +84,7 @@ import {
 import { assertHardhatNetworkInvariant } from "../utils/assertions";
 import { ModulesLogger } from "./logger";
 
+const EIP1559_MIN_HARDFORK = "london";
 const ACCESS_LIST_MIN_HARDFORK = "berlin";
 const EIP155_MIN_HARDFORK = "spuriousDragon";
 
@@ -332,7 +333,7 @@ export class EthModule {
     rpcCall: RpcCallRequest,
     blockTag: OptionalRpcNewBlockTag
   ): Promise<string> {
-    this._validateAccessListHardforkRequirement(rpcCall);
+    this._validateTransactionAndCallRequest(rpcCall);
 
     const blockNumberOrPending = await this._resolveNewBlockTag(blockTag);
 
@@ -403,7 +404,7 @@ export class EthModule {
     callRequest: RpcCallRequest,
     blockTag: OptionalRpcNewBlockTag
   ): Promise<string> {
-    this._validateAccessListHardforkRequirement(callRequest);
+    this._validateTransactionAndCallRequest(callRequest);
 
     // estimateGas behaves differently when there's no blockTag
     // it uses "pending" as default instead of "latest"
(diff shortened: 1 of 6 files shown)


packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/sendRawTransaction.ts
```diff
@@ -98,27 +99,56 @@ describe("Eth module", function () {
...
+                ],
+                "too low for the next block, which has a baseFeePerGas of"
+              );
+            });
+          });
 
-            try {
-              // sends a tx with 21000 gas to the 0x1 precompile
+          describe("With automining disabled", function () {
+            it("Should enqueue txs that can't be mined in the next block", async function () {
```
fvictorio, line 146: Shouldn't we assert something here? At least that the tx exists, to be sure that it's not dropped.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/hardforks.ts
```diff
@@ -177,42 +177,43 @@ describe("Eth module - hardfork dependant tests", function () {
...
           );
         });
 
-        it("Should reject an eth_sendRawTransaction if the tx uses an access list", async function () {
-          const berlinCommon = this.common.copy();
-          berlinCommon.setHardfork("berlin");
-
-          const signedTx = getSampleSignedAccessListTx(berlinCommon);
+        it("Should validate the chain id in eth_sendRawTransaction using an eip-1559 tx", async function () {
+          const signedTx = getSampleSignedEIP1559Tx();
```
fvictorio, line 202: I'm confused about how this works. It's because `getSampleSignedEIP1559Tx` is not receiving a `common` and therefore the chainId in the raw tx is invalid? I wonder if we can be more explicit here.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -1653,8 +1634,30 @@ You can use them by running Hardhat Network with 'hardfork' ${ACCESS_LIST_MIN_HA
...
       throw new InvalidArgumentsError(`Trying to send an EIP-155 transaction, but they are not supported by the current hardfork.  
-      
+
 You can use them by running Hardhat Network with 'hardfork' ${EIP155_MIN_HARDFORK} or later.`);
     }
   }
+
+  private _validateRawTransactionHardforkRequirements(rawTx: Buffer) {
+    if (rawTx[0] <= 0x7f && rawTx[0] !== 1 && rawTx[0] !== 2) {
+      throw new InvalidArgumentsError(`Invalid transaction`);
```
fvictorio, line 1644: I know this was just moved around, but can we use a more explicit error message here? Otherwise if a user runs into this, the cause is going to be harder to pin down.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/sendRawTransaction.ts
alcuadrado: yeah, good point.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/hardforks.ts
alcuadrado: Makes sense. I made the common required in every helper.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
alcuadrado: Done

fvictorio (approved):

Issue #1688, opened 2021-07-17
BASEFEE always returns 0 when running an `eth_call` with a London block


alcuadrado:
I'm opening this issue to document a limitation in our `eth_call` implementation.

Transcript of an internal conversation in the ethereumjs discord:

> @alcuadrado:
> I found another limitation of the vm wrt to geth's and other's
> eth_call should run an EVM message in the context of a block and return its return data. Nothing new here.
> 
> This can only be done with runTx in the vm, as runCall has some limitations we've discussed a lot of times.
> 
> runTx requires a tx with a gasPrice/maxFeePerGas&maxPriorityFeePerGas. This value should be 0, and this is how other clients behave.
> 
> If the block where the eth_call is running has a baseFeePerGas, this tx will be rejected, as its fee params would be too low (that 0). We should have a way to disable this check.
> 
> As a workaround, we are creating a new block with baseFeePerGas = 0, but this will also change the value returned by BASEFEE.

penandlim:
Note that this is currently blocking estimating costs on forked network for Arbitrum Bridge as some of their inner working contracts reference basefee to calculate the cost for bridging.

Looks like it returns zero on view / callstatic but uses expected value on write calls.

penandlim:
Relevant issue: https://github.com/NomicFoundation/hardhat/issues/3028

fvictorio:
Thanks for the extra data @penandlim, I will update the priority of this issue.

PR #1689, opened 2021-07-17
Adapt `eth_estimateGas` and `eth_call` to London

### LINKED ISSUE #1436: Balance of msg.sender shouldn't be modified whan a view function is called
Given this contract:

```solidity
contract Foo {
  function printBalance() public view returns(uint256) {
    return msg.sender.balance;
  }
}
```

Deploying it and doing this in Hardhat network:

```js
console.log((await ethers.provider.getBalance(signer.address)).toString());
console.log((await foo.printBalance()).toString());
```

prints two different numbers, while doing the same thing in Rinkeby prints the same number.

### LINKED ISSUE #1601: Update `eth_call` to support EIP-1559
This issue requires more research. We should probably do [what geth does](https://github.com/ethereum/go-ethereum/pull/23027).

### LINKED ISSUE #1602: Update `eth_estimateGas` to support EIP-1559
It's unclear how we should change it yet, but implementing `eth_call` first should give us a better insight into this.

alcuadrado:
This PR turned out to be more challenging than expected.

The reason for this are that:

(1) `eth_call` should run an EVM Message that doesn't pay gas unless the user specifies it. EIP-1559 makes this more complex, as you can't just set the gas price to 0, because of the baseFeePerGas. To workaround this we create a new block with base fee 0 when we run an `eth_call`, but that will break BASEFEE (see: #1688).

(2) The default `maxFeePerGas` and `maxPriorityFeePerGas` in `eth_estimateGas` can be tricky to compute, specially when it's in an older block. I'm not sure how to test this yet.

On the plus side, this PR also fixes #1436.

fixes #1601
fixes #1602

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -2141,11 +2141,8 @@ Hardhat Network's forking functionality only works with blocks from at least spu
     if (maxPriorityFeePerGas === undefined) {
       maxPriorityFeePerGas = await this.getMaxPriorityFeePerGas();
 
-      if (
-        callParams.maxFeePerGas !== undefined &&
-        callParams.maxFeePerGas.lt(maxPriorityFeePerGas)
-      ) {
-        maxPriorityFeePerGas = callParams.maxFeePerGas;
+      if (maxFeePerGas !== undefined && maxFeePerGas.lt(maxPriorityFeePerGas)) {
+        maxPriorityFeePerGas = maxFeePerGas;
       }
     }
 
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -472,14 +472,38 @@ Hardhat Network's forking functionality only works with blocks from at least spu
     call: CallParams,
     blockNumberOrPending: BN | "pending"
   ): Promise<RunCallResult> {
-    const tx = await this._getFakeTransaction({
-      ...call,
-      nonce: await this._getNonce(new Address(call.from), blockNumberOrPending),
-    });
+    let txParams: TransactionParams;
+
+    const nonce = await this._getNonce(
+      new Address(call.from),
+      blockNumberOrPending
+    );
+
+    if (call.gasPrice !== undefined || !this.isEip1559Active()) {
+      txParams = {
+        gasPrice: new BN(0),
+        nonce,
+        ...call,
+      };
+    } else {
+      const maxFeePerGas =
+        call.maxFeePerGas ?? call.maxPriorityFeePerGas ?? new BN(0);
+      const maxPriorityFeePerGas = call.maxPriorityFeePerGas ?? new BN(0);
+
+      txParams = {
+        ...call,
+        nonce,
+        maxFeePerGas,
+        maxPriorityFeePerGas,
+        accessList: call.accessList ?? [],
+      };
+    }
+
+    const tx = await this._getFakeTransaction(txParams);
 
     const result = await this._runInBlockContext(
       blockNumberOrPending,
-      async () => this._runTxAndRevertMutations(tx, blockNumberOrPending)
+      async () => this._runTxAndRevertMutations(tx, blockNumberOrPending, true)
     );
 
     const traces = await this._gatherTraces(result.execResult);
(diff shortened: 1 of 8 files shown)


packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/call.ts
```diff
@@ -488,6 +494,203 @@ describe("Eth module", function () {
...
+                  from: CALLER,
+                  to: contractAddress,
+                  data: balanceSelector,
+                  gasLimit: numberToRpcQuantity(gasLimit),
+                  gasPrice: numberToRpcQuantity(gasPrice),
+                },
+              ]);
+
+              assert.isTrue(
+                rpcDataToBN(balanceResult).lt(
```
fvictorio, line 568: Why is `lt` used here? In my mind, it should be either equal to that number, or maybe be between that number and the eth balance. I don't understand why it's less than that :confused:

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -568,13 +592,77 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+
+        if (
+          callParams.maxFeePerGas !== undefined &&
+          callParams.maxFeePerGas.lt(maxPriorityFeePerGas)
+        ) {
+          maxPriorityFeePerGas = callParams.maxFeePerGas;
+        }
+      }
+
+      if (callParams.maxPriorityFeePerGas === undefined) {
```
fvictorio, line 623: Does this `if` makes sense? It seems to me that, if if entered the previous one, then it will enter in this one, so it can just be removed and the two code blocks merged.

fvictorio: Maybe this was meant to be `if (maxFeePerGas === undefined)`? If that's the case, it would mean that we need to add a test.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -568,13 +592,77 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+        if (blockNumberOrPending !== "pending") {
+          const block = await this.getBlockByNumber(blockNumberOrPending);
+
+          maxFeePerGas = maxPriorityFeePerGas.add(
+            block?.header.baseFeePerGas ?? new BN(0)
+          );
+        }
+
+        const baseFeePerGas = await this.getNextBlockBaseFeePerGas();
+        maxFeePerGas = baseFeePerGas!.muln(2).add(maxPriorityFeePerGas);
```
fvictorio, line 633: I think these two statements should be inside an `else` block, otherwise the previous code is useless, isn't it? And if that's the case, I would switch the blocks (that is, I would start with the `=== "pending"` case, to avoid a conceptual double-negative)

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -568,13 +592,77 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+          accessList: callParams.accessList ?? [],
+        };
+      }
+    } else {
+      // EIP-1559 tx
+      txParams = {
+        ...callParams,
+        nonce,
+        maxFeePerGas: maxFeePerGas!,
+        maxPriorityFeePerGas: maxPriorityFeePerGas!,
```
fvictorio, line 662: I know I'm getting annoying with this but... I would avoid the non-null assertion and use an `assert*Invariant` :sweat_smile: Especially because the previous code is somewhat complex, so it's not obvious that these values will be defined here.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/call.ts
```diff
@@ -488,6 +494,203 @@ describe("Eth module", function () {
...
+            const contract = compilerOutput.contracts["literal.sol"].C;
+            deploymentBytecode = `0x${contract.evm.bytecode.object}`;
+            balanceSelector = `0x${contract.evm.methodIdentifiers["balance()"]}`;
+          });
+
+          const CALLER = DEFAULT_ACCOUNTS_ADDRESSES[2];
+          let contractAddress: string;
+          let ethBalance: BN;
+
+          function deployContractAndGetEthBalance() {
```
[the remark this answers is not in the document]
alcuadrado: I'd be ok with that, but it's not possible, as `this.provider` is `undefined` by the time this is called.

I think this is the only reason to set things (i.e the provider) in the mocha context.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/call.ts
alcuadrado: Fixed!

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: it should be `maxFeePerGas`.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/call.ts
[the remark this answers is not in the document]
fvictorio: Oh, right, that makes sense.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/estimateGas.ts
```diff
@@ -206,6 +215,163 @@ describe("Eth module", function () {
                 "maxPriorityFeePerGas (2) is bigger than maxFeePerGas (1)"
               );
             });
+
+            describe.only("Default values", function () {
```
fvictorio, line 219: We really need to add the `--forbid-only` thing.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2056,6 +2124,46 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+
+    if (maxPriorityFeePerGas === undefined) {
+      maxPriorityFeePerGas = await this.getMaxPriorityFeePerGas();
+
+      if (
+        callParams.maxFeePerGas !== undefined &&
+        callParams.maxFeePerGas.lt(maxPriorityFeePerGas)
+      ) {
+        maxPriorityFeePerGas = callParams.maxFeePerGas;
+      }
```
fvictorio, line 2149: I would move this after `maxFeePerGas`, where it will be defined:

```ts
if (maxFeePerGas.lt(maxPriorityFeePerGas)) {
  maxPriorityFeePerGas = maxFeePerGas;
}
```

Not just because there's one check less, but because it makes more conceptual sense to me. Plus, right now the `maxFeePerGas` set in the following block is always greater or equal than `maxPriorityFeePerGas`, but that might be just a circumstantial thing.

That being said, maybe there is a good reason why this _needs_ to be done here?

alcuadrado: The default `maxFeePerGas` is defined based on `maxPriorityFeePerGas`, so it's not as simple as moving it.

We could do an `else` for `if (maxFeePerGas === undefined) {`, but that would look even worse imo.

fvictorio: Oh, that makes sense, disregard this then!

fvictorio (approved):

PR #1726, opened 2021-07-24
Give Waffle runtime augment its own interface

flux627:
- [ ] (If this PR includes a **documentation change**) This PR's branch was created from the `website` branch, and this PR uses the `website` branch as its base branch.
- [ ] (If this PR includes a **bug fix**) Relevant tests have been included.
- [ ] (If this PR includes a **new feature**) The change was previously discussed on an Issue or with someone from the team.

---

Re: NomicFoundation/hardhat-waffle#3 

This change allows projects like TypeChain to augment Hardhat's Waffle integration (specifically, overloading `deployContract` for type support). Should have no other consequence.


packages/hardhat-waffle/src/type-extensions.ts
```diff
@@ -10,21 +10,23 @@ import type { ContractJSON } from "ethereum-waffle/dist/esm/ContractJSON";
 import type { Contract, providers, Signer } from "ethers";
 import "hardhat/types/runtime";
 
+interface HardhatWaffle {
```
alcuadrado, line 13: I think this should be placed in another file. Probably it's own file.

alcuadrado:
Thanks for sending this PR!

I only left a minor comment.

packages/hardhat-waffle/src/type-extensions.ts
flux627: Done- let me know if this is what you had in mind

fvictorio: For consistency with how this is done in the rest of our packages, I would move it to a `src/types.ts` file instead of `src/HardhatWaffle.ts`

fvictorio: Oh, and I'm pretty sure you are going to get a linter error about import groups. You can run `yarn lint` inside the package directory to check that.

flux627: Renamed to `types.ts`. Also ran linting check, and got no errors.

fvictorio:
Hey @flux627, sorry for closing this, but we are moving `hardhat-waffle` to [another repository](https://github.com/NomicFoundation/hardhat-waffle/). If you want, you could rebase this change there, although I would understand if you no longer care about this change after all this time :sweat_smile:.

Issue #1732, opened 2021-07-28
Testing methods reject hex-encoded parameters


wolflo:
Testing methods like `evm_mine` and `evm_increaseTime` expect integer parameters. Unlike other methods, they reject these parameters if they are not decimal-encoded.

### To reproduce

`npx hardhat --version` -> `2.5.0`

Run `npx hardhat node` and make the following rpc calls:
```
> curl -H "Content-Type: application/json" -X POST --data \
        '{"id":1337,"jsonrpc":"2.0","method":"evm_increaseTime","params":[32]}' \
        http://localhost:8545
{"jsonrpc":"2.0","id":1337,"result":"41"}

> curl -H "Content-Type: application/json" -X POST --data \
        '{"id":1337,"jsonrpc":"2.0","method":"evm_increaseTime","params":["0x20"]}' \
        http://localhost:8545
{"jsonrpc":"2.0","id":1337,"error":{"code":-32602,"message":"Errors encountered in param 0: Invalid value \"0x20\" supplied to : number"}}
```

wolflo:
This appears to be enforced in the test suite for `evm_increaseTime`. 

https://github.com/nomiclabs/hardhat/blob/c9e8a609be8b7751ce375a22159d4d2af162dad2/packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts#L127-L131

I would disagree that this should be the case, though. 

I don't know the reasoning for returning the total adjusted time from this method as a decimal-encoded value rather than hex-encoded, but that does seem to be intentional. However, ganache does still accept a hex-encoded argument to the method. Changing this behavior requires a special case for `evm_increaseTime` in any library that automatically encodes arguments to rpc methods as hex strings.

fvictorio:
I think it makes sense to be more permissive and accept hex-encoded arguments in these methods. But we can't change the output of the method since that would be a breaking change.

@alcuadrado thoughts?

wolflo:
Seems that the opposite is true, too. Ganache accepts decimal-encoded values for non-test methods like `eth_getBlockByNumber`, while hardhat (and geth) reject them.

`ganache-cli`
```
~ ❯ curl -H "Content-Type: application/json" -X POST --data \
        '{"id":1337,"jsonrpc":"2.0","method":"eth_getBlockByNumber","params":[1, true]}' \
        http://localhost:8545
{"id":1337,"jsonrpc":"2.0","result":null}

[...]

PR #1735, opened 2021-07-29
Accept hex strings for rpc test methods

wolflo:
Addresses #1732.

Adds a new type, private to the `evm` module, that includes both `number` and `rpcQuantity` types.

Modifies `evm_increaseTime`, `evm_mine`, and `evm_setNextBlockTimestamp` to accept this new type.


alcuadrado:
Thanks for this PR, @wolflo !

PR #1751, opened 2021-08-03
Split PR template into multiple templates

feuGeneA:
[Creating a pull request template for your repository](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) suggests that this should be possible:

> You can create a _PULL_REQUEST_TEMPLATE/_ subdirectory in any of the supported folders to contain multiple pull request templates

I modeled these changes after https://github.com/stackblitz/webcontainer-core/issues/new/choose .  I'm not sure how to test them before they're merged!


alcuadrado (approved):

feuGeneA:
Upon verification, I found that while the templates are usable, the usability is too weird to actually be useful.

There is no nice GUI displaying the possible PR templates anywhere that I can find.  It seems that the only way to use them is by adding a query parameter to the URL where you would create a PR.  For example, recall that if you navigate to `https://github.com/nomiclabs/hardhat/compare/BRANCHNAME`, you will be presented with an interface to create a PR with `BRANCHNAME` as the source branch.  From there, you can append `?template=new_feature.md`, or `?template=bug_fix.md`, and the template will then appear in the PR description form.

Because there is no GUI for this, and because there is no obvious way to easily direct people to use the template query parameter, I'm going to revert this change.

PR #1753, opened 2021-08-03
Require ethers from hardhat explicitly

panoplied:

Require ethers from hardhat explicitly. There are configurations of common tools (VSCode for example) where editor automatically finds and adds require from "ethers" and not from "hardhat" (ethers gets installed into the project earlier in the tutorial), which often leads beginners to unexpected errors.


alcuadrado:
Hey @panoplied,

Thanks for submitting this PR! Unfortunately, I don't think it should be merged, as it leads to inconsistencies within the docs, and that can be confusing.

I created #1756 to keep track of this and improve it.

PR #1754, opened 2021-08-04
Fix confusing comment

panoplied:

---

Minor fix of comment with the numbers of tokens. The Token.sol actually deposits 1000000 tokens to owner's account, not 1000.


alcuadrado:
Good catch! Thanks

PR #1760, opened 2021-08-04
Drop support for Node.js 10

alcuadrado:
This PR drops support for Node.js 10. This means that we stop running that version on our CI, and that we print a warning if you use it.

It also adds documentation about our stability guarantees, and links it in the warning mentioned above.

### DIFF
--- a/docs/_redirects
+++ b/docs/_redirects
@@ -24,6 +24,7 @@ https://www.usehardhat.com/* https://hardhat.org/:splat 301!
 /console-log /hardhat-network/#console-log 302
 /discord https://discord.gg/TETZs2KK4k 302
 /hre /advanced/hardhat-runtime-environment.html 302
+/nodejs-versions /reference/stability-guarantees.html 302
 
 ## Hardhat migration
 
--- a/docs/reference/stability-guarantees.md
+++ b/docs/reference/stability-guarantees.md
@@ -1,0 +1,26 @@
+# Stability guarantees
+
+Hardhat doesn't follow semver strictly, but it still aims to be a stable and easy-to-use platform, and we won't introduce breaking changes without notice. Instead, we may introduce breaking changes in some minor versions.
+
+Those breaking changes are part of these categories:
+
+- Dropping support for unmaintained Node.js versions
+- Changing the default config of Hardhat Network
+
+## Hardhat Network default config
+
+Hardhat Network should closely resemble Ethereum Mainnet by default. Given that Ethereum does introduce breaking changes in the form of hardforks or network upgrades, we need to do the same with its default config. For example, we would eventually change the default `hardfork` config field in the network `hardhat`.
+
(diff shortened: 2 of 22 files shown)


docs/reference/stability-guarantees.md
```diff
@@ -0,0 +1,26 @@
...
+
+Hardhat supports every currently maintained Node.js version, up to two months after its end-of-life. After that period of time, we will stop testing against it, and print a warning when trying to use it. These actions will lead to a new minor version being released.
+
+We recommend running Hardhat using the current LTS Node.js version. You can learn about it [here](https://nodejs.org/en/about/releases/).
+
+## How to avoid the breaking changes introduced by Hardhat
+
+In general, there's no need to avoid them. Using Hardhat with any actively maintained Node.js version means that you'd be running with defaults that closely resemble Ethereum Mainnet.
+
+If you want to play extra-safe, and make sure your project will continue running years down the road using the same Node.js version, please, install Hardhat using a [tilde range](https://docs.npmjs.com/cli/v7/using-npm/semver#tilde-ranges-123-12-1).
```
fvictorio, line 26:
```suggestion
If you want to play extra-safe, and make sure your project will continue running years down the road using the same Node.js version, please install Hardhat using a [tilde range](https://docs.npmjs.com/cli/v7/using-npm/semver#tilde-ranges-123-12-1).
```

docs/_redirects
```diff
@@ -24,6 +24,7 @@ https://www.usehardhat.com/* https://hardhat.org/:splat 301!
 /console-log /hardhat-network/#console-log 302
 /discord https://discord.gg/TETZs2KK4k 302
 /hre /advanced/hardhat-runtime-environment.html 302
+/nodejs-versions /reference/stability-guarantees.html 302
```
fvictorio, line 27: Should we show this in the navbar?

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -38,12 +38,18 @@ async function printVersionMessage(packageJson: PackageJson) {
...
-function ensureValidNodeVersion(packageJson: PackageJson) {
+function printWarningAboutNodeJsVersionIfNeceesary(packageJson: PackageJson) {
   const requirement = packageJson.engines.node;
   if (!semver.satisfies(process.version, requirement)) {
-    throw new HardhatError(ERRORS.GENERAL.INVALID_NODE_VERSION, {
-      requirement,
-    });
+    console.warn(
+      chalk.yellow(
+        `You are using an version of Node.js that is not supported by Hardhat, and it may work incorrectly, or not work at all.
```
fvictorio, line 46:
```suggestion
        `You are using a version of Node.js that is not supported by Hardhat, and it may work incorrectly, or not work at all.
```

docs/_redirects
alcuadrado: If you meant the sidebar, yes. I'll add it.

fvictorio (approved):

PR #1768, opened 2021-08-06
Fixes after london hardfork

fvictorio:
- [x] Fix existing tests
- [x] Make `eth_gasPrice` return valid values when forking a network with London enabled
- [x] Use this better `eth_gasPrice` instead of using 1000 gwei of gas price.


packages/hardhat-core/test/internal/hardhat-network/provider/baseFeePerGas.ts
```diff
@@ -83,7 +88,7 @@ describe("Block's baseFeePerGas", function () {
             });
           });
         } else {
-          describe("When forking from a remote network the next block must have the right value", function () {
+          describe("When forking from a remote network the forked block must have the right value", function () {
```
alcuadrado, line 91:
```suggestion
          describe("When forking from a remote network the next block must have the right value", function () {
```

I think this is outdated?

PR #1778, opened 2021-08-10
docs: mention "london" among accepted values for "evmVersion"

paulrberg:

---

Since [v2.6.0](https://github.com/nomiclabs/hardhat/releases/tag/hardhat-core-v2.6.0), Hardhat accepts "london" as one of the allowed values for `evmVersion`.

### DIFF
--- a/docs/guides/compile-contracts.md
+++ b/docs/guides/compile-contracts.md
@@ -59,7 +59,7 @@ module.exports = {
 
 - `optimizer`: an object with `enabled` and `runs` keys. Default value: `{ enabled: false, runs: 200 }`.
 
-- `evmVersion`: a string controlling the target evm version. One of `homestead`, `tangerineWhistle`, `spuriousDragon`, `byzantium`, `constantinople`, `petersburg`, `istanbul`, and `berlin`. Default value: managed by `solc`.
+- `evmVersion`: a string controlling the target evm version. One of `homestead`, `tangerineWhistle`, `spuriousDragon`, `byzantium`, `constantinople`, `petersburg`, `istanbul`, `berlin` or `london`. Default value: managed by `solc`.
 
 If any of your contracts have a version pragma that is not satisfied by the compiler version you configured, then Hardhat will throw an error.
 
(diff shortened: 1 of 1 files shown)


fvictorio:
I wonder if listing the hardforks make sense here. The actual list depends on the solc version you are using, right? Maybe we shouldn't list any hardfork at all and just mention that is a solc configuration.

@alcuadrado @feuGeneA what do you think?

feuGeneA:
> I wonder if listing the hardforks make sense here. The actual list depends on the solc version you are using, right? Maybe we shouldn't list any hardfork at all and just mention that is a solc configuration.
> 
> @alcuadrado @feuGeneA what do you think?

I think it makes sense to list them here, but I suggest we (a) prefix the sentence with "For example, ..." rather than just "One of...", and (b) shorten the list to just a few, perhaps just "istanbul, berlin or london" (also note (c) the shift from "and" to "or"), all in order to avoid sounding too authoritative, reinforcing the deference to the solc reference that we're already linking to previously in this doc.

Also, @paulrberg note that you introduced a line break into the text; I believe that's the reason for the failed linter run in the PR.

alcuadrado:
I agree with @feuGeneA here!

docs/guides/compile-contracts.md
```diff
@@ -59,7 +59,7 @@ module.exports = {
 
 - `optimizer`: an object with `enabled` and `runs` keys. Default value: `{ enabled: false, runs: 200 }`.
 
-- `evmVersion`: a string controlling the target evm version. One of `homestead`, `tangerineWhistle`, `spuriousDragon`, `byzantium`, `constantinople`, `petersburg`, `istanbul`, and `berlin`. Default value: managed by `solc`.
+- `evmVersion`: a string controlling the target evm version. One of `homestead`, `tangerineWhistle`, `spuriousDragon`, `byzantium`, `constantinople`, `petersburg`, `istanbul`, `berlin` or `london`. Default value: managed by `solc`.
```
feuGeneA, line 62:
```suggestion
- `evmVersion`: a string controlling the target evm version. For example: `istanbul`, `berlin` or `london`. Default value: managed by `solc`.
```

fvictorio (approved):

PR #1779, opened 2021-08-10
Update and pin devDependencies

fvictorio:
This PR upgrades some devDependencies and pins all of them, except for when `hardhat` or one of our plugins is a dev dependency.

This also updates the `check-dependencies` script. Now it checks that devDependencies are pinned (except for the mentioned ones) and it's also more permissive about having two different semver ranges for the same dependency: if one of those ranges is a pinned version, and if it satisfies the other one, then it's considered fine. This is necessary to have permissive `peerDependencies` but a pinned `devDependency` for the same package.

One non-trivial devDependency that I upgraded is the `@types/node` one. Only one change was needed after doing this, check inline comments.


packages/hardhat-core/src/internal/context.ts
```diff
@@ -10,9 +10,10 @@ import { ERRORS } from "./core/errors-list";
 import { TasksDSL } from "./core/tasks/dsl";
 import { getRequireCachedFiles } from "./util/platform";
 
-export type GlobalWithHardhatContext = NodeJS.Global & {
-  __hardhatContext: HardhatContext;
-};
+export type GlobalWithHardhatContext = NodeJS.Global &
+  typeof globalThis & {
+    __hardhatContext: HardhatContext;
+  };
```
fvictorio, line 16: This was the only necessary change after upgrading `@types/node`.

packages/hardhat-core/package.json
```diff
@@ -55,40 +55,40 @@
...
+    "@types/chai": "4.2.21",
+    "@types/chai-as-promised": "7.1.4",
+    "@types/ci-info": "2.0.0",
+    "@types/debug": "4.1.7",
+    "@types/find-up": "2.1.1",
+    "@types/fs-extra": "5.1.0",
+    "@types/glob": "7.1.4",
+    "@types/lodash": "4.14.172",
+    "@types/mocha": "5.2.7",
+    "@types/node": "12.20.19",
```
fvictorio, line 67: @alcuadrado can you double-check that this is the version we want to use here?

alcuadrado: double-checked ✅

alcuadrado:
I'm a bit hesitant about this PR. It feels a bit too much, and it may lead us to never ever update a dev dependency, not even patches.

I know this may sound contradicting, as I was the one that mentioned that we should pin our tools, but I didn't mean every dev dependency. To me the ones that should be pinned are the ones that compile / validate / modify the code, as they tend to encode subjective preferences of style that change with versions. For example, a set of "recommended" linting rules, or the default prettier options, can change with time.

fvictorio:
Oh, that makes sense to me. Which ones should we pin then? My guess is:

- All the eslint ones
  - `@typescript-eslint/eslint-plugin`
  - `@typescript-eslint/parser`
  - `eslint`
  - `eslint-config-prettier`
  - `eslint-plugin-import`
  - `eslint-plugin-prettier`
- `prettier`

What do you think? (I'll probably make a new PR with the proper pinned list, plus the `@types/node` bump, so I'm closing this one)

fvictorio:
Prettier is actually already pinned (as it should), so the `eslint` ones, and maybe `typescript`?

PR #1723, opened 2021-07-23
Fix eth_subscribe for web3 clients

feuGeneA:
Fixes #588 

-  [ ] Reproduce the problem in a test case.
-  [ ] Fix the problem.

Issue #588, opened 2020-05-20
web3 client not receiving eth_subscribe data from hardhat network


Detoo:
Hi, I am having a weird issue where my web3 client connecting to a local buidler-evm wasn't able to receive any event. I have a runnable example along with a control group here (https://github.com/Detoo/buidler-eth-subscribe-issue) but here's a quick summary of what I saw:

## The code snippet

```javascript
greeter.events.Greet()
        .on("data", (event) => console.log('event:', event))
        .on("error", (error) => console.log('error:', error))

await greeter.methods.greet().send()
```

## Expected Results

For sanity check, I tested the same codes with a ganache-cli evm instead and it did work:

```
$ npm run main

> buidler-eth-subscribe-issue@1.0.0 main /Users/detoo/Dropbox/playground/buidler-eth-subscribe-issue
> buidler run src/main.ts

All contracts have already been compiled, skipping compilation.
event: {
  logIndex: 0,
  transactionIndex: 0,
  transactionHash: '0x161b5b7ea58d5c634bcdef15780f5382d65d3ad257e970ef23bb827ea94fd58e',
  blockHash: '0x21da0dab9277f0a4cd7ff44db9c5ad85d2106e851f5d532b38bf5c1f88b85e34',
  blockNumber: 2,
  address: '0xbFbd287DDF9dE89ba9781c92aa21654Dac99c283',
  type: 'mined',
  id: 'log_2406fd8c',
  returnValues: Result {
    '0': '0xbFbd287DDF9dE89ba9781c92aa21654Dac99c283',
    '1': '123',
    from: '0xbFbd287DDF9dE89ba9781c92aa21654Dac99c283',
    value: '123'
  },
  event: 'Greet',
  signature: '0xc3b6fa780771bb4c77c875ed1937bd523871551d0234535b464710f4b4a10788',
  raw: {
    data: '0x000000000000000000000000000000000000000000000000000000000000007b',
    topics: [
      '0xc3b6fa780771bb4c77c875ed1937bd523871551d0234535b464710f4b4a10788',
      '0x000000000000000000000000bfbd287ddf9de89ba9781c92aa21654dac99c283'
    ]
  }
}
```

## Actual Results

However, when it run against a buidler-evm, I got nothing in response:

```
$ npm run main

> buidler-eth-subscribe-issue@1.0.0 main /Users/detoo/Dropbox/playground/buidler-eth-subscribe-issue
> buidler run src/main.ts

[...]

PR #1780, opened 2021-08-10
Fix WebSocket `eth_subscribe` subscriptions

ngotchac:
- [ ] This PR includes a **documentation change**, so its branch was created from the `website` branch, and this PR uses the `website` branch as its base branch.
- [ ] This PR includes a **bug fix**, so relevant tests have been included.
- [ ] This PR includes a **new feature**, so the change was previously discussed on an Issue or with someone from the team.

---

It seems that subscriptions weren't working with WebSocket connections, but I think it was actually only these small mistakes that prevented it.

I tried to subscribe to `newHeads` on this branch and everything seems to be working now. I haven't written tests for it since I'm not 100% sure how they would be (I guess creating a full Hardhat node and connecting to it via a WS client?)

This might fix this issue: https://github.com/nomiclabs/hardhat/issues/588


feuGeneA:
I'm going to take a look at this today.  I already had on my agenda to get some tests in place around this functionality (whether the tests would pass or fail was an open question in my mind).  I'll try out your changes here along with that testing effort, and report back here.

feuGeneA:
I started writing tests to exercise this stuff, but I have not finished.  I tried to run them with your changes in place, but they are not passing... yet! :slightly_smiling_face:

Feel free to take a stab at integrating with them -- they're over in #1723 -- but please realize that I'm very much still hacking at this point in time. I'll be digging in further later this week, and I will report back here again with any findings.

And thank you very much for the contribution!

feuGeneA:
An update:

I made a test that uses hardhat-ethers to receive contract events via WebSocket from ganache, and that test is passing just fine.  You can see the test in [this commit](https://github.com/nomiclabs/hardhat/pull/1792/commits/ec681b0be4468e9882fa668411e9593340b01122), over in #1792.  I then tried to add an analogous test that uses Hardhat Network instead of ganache, with the intention of then trying out your changes to see if it would help the test pass, but I realized that the existing test infrastructure was incorrect for such a test because Hardhat Network was being interfaced with in-process rather than via HTTP/WS.  So, overall, not very helpful for this PR, but that's my update. :)

I will be continuing to investigate #588, and along with that I will be continuing to work with your PR here.

I discussed this PR with the team, and we said that we would be willing to accept this PR even though it lacks automated tests, *if* we could manually verify the changes.  If you could provide a manual test plan for us to follow in such a verification, that would be super helpful.

Thanks again for your efforts here.

fvictorio:
I can confirm that this indeed fixes the websocket issues and it also makes ethers.js event work fine with ws.

fvictorio:
@alcuadrado this is now ready for review.

Something I noticed while working on this, though, is that there's a weird "deadline" thing for events. I mean this:

https://github.com/ngotchac/hardhat/blob/0be58ab515ba2e17e520ad8b2e745ddcbe7c524a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts#L972

If I understand this correctly, this means that all subscriptions will expire after 5 minutes. I guess this is a decision that was taken at some point, but to be honest, it seems like a problematic thing to have. Is there a good reason to keep this?

ngotchac:
Hey guys, sorry for not being super responsive. Great news that it indeed fixes everything, and thanks a ton @fvictorio for the tests, they look perfect ! :ok_hand: 
Glad this can be merged and fixed :)

Regarding the _deadline_, I vaguely remember something similar in Geth or Parity-Ethereum ; I'll try to find out

alcuadrado (approved):

fvictorio:
That makes sense, thanks a lot @ngotchac!

PR #1802, opened 2021-08-17
Fix traceTransaction for FakeSender transactions

### LINKED ISSUE #1799: `traceTransaction` doesn't work for FakeSender transactions
The [traceTransaction(...)](https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts#L1235-L1256) function takes an incoming `tx`, and converts it into either @ethereumjs' Transaction (legacy), EIP1559Transaction or EIP2930Transaction object without considering the type of `tx`. This `txWithCommon` value is forwarded to the vm (see relevant code below).

```js
public async traceTransaction(hash: Buffer, config: RpcDebugTracingConfig) {
  ...
      for (const tx of block.transactions) {
        let txWithCommon: TypedTransaction;
        if (tx.type === 0) {
          txWithCommon = new Transaction(tx, {
            common: vm._common,
          });
        } else if (tx.type === 1) {
          txWithCommon = new AccessListEIP2930Transaction(tx, {
            common: vm._common,
          });
        } else if (tx.type === 2) {
          txWithCommon = new FeeMarketEIP1559Transaction(
            { ...tx, gasPrice: undefined },
            {
              common: vm._common,
            }
          );
        } else {
          throw new InternalError(
            "Only legacy, EIP2930, and EIP1559 txs are supported"
          );
        }
  ...
}
```

If `tx` is of type FakeSenderTransaction, then it gets converted to type Transaction and we lose the sender information. This results in weird errors downstream since `getSenderAddress()` of the new Transaction now returns a garbage address. I noticed this when using hardhat with [brownie](https://github.com/eth-brownie/brownie) where a call to `debug_traceTransaction` results in `Error: the tx doesn't have the correct nonce. account has nonce of: 0 tx has nonce of: xxx`. This is because of the garbage sender address.

We should use the same transaction type when copying `tx` to `txWithCommon`. I tested this by tweaking the code a bit, and it seemed to work. I'd be happy to create a PR for this.

```js
if (tx.type === 0) {
  if (tx instanceof FakeSenderTransaction) {
    txWithCommon = new FakeSenderTransaction(
      tx.getSenderAddress(),
      tx,
      {
        common: vm._common,
      }
    );
  } else {
    txWithCommon = new Transaction(tx, {
      common: vm._common,
    });
  }
}
```

## Steps to reproduce
1. Launch [brownie](https://github.com/eth-brownie/brownie) (>= 1.16.0) with hardhat-fork 
`brownie console --network hardhat-fork` 

2. Transact using a random account
```python
deployer = "0x9C33eaCc2F50E39940D3AfaF2c7B8246B681A374"
router = Contract.from_explorer("0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D")

tx = router.WETH.transact({'from': deployer})
```

3. Do a `debug_traceTransaction` call
```
tx.return_value
# See error
RPCRequestError: Error: the tx doesn't have the correct nonce. account has nonce of: 0 tx has nonce of: 5
```

shuklaayush:

---

This PR fixes `debug_traceTransaction` calls for FakeSender transactions. While tracing transactions of type FakeSender***Transaction, I made sure that they are not converted to normal Transaction types and we end up losing sender information.

Fixes #1799

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -1235,20 +1235,50 @@ Hardhat Network's forking functionality only works with blocks from at least spu
         for (const tx of block.transactions) {
           let txWithCommon: TypedTransaction;
           if (tx.type === 0) {
-            txWithCommon = new Transaction(tx, {
-              common: vm._common,
-            });
-          } else if (tx.type === 1) {
-            txWithCommon = new AccessListEIP2930Transaction(tx, {
-              common: vm._common,
-            });
-          } else if (tx.type === 2) {
-            txWithCommon = new FeeMarketEIP1559Transaction(
-              { ...tx, gasPrice: undefined },
-              {
+            if (tx instanceof FakeSenderTransaction) {
+              txWithCommon = new FakeSenderTransaction(
+                tx.getSenderAddress(),
+                tx,
+                {
+                  common: vm._common,
+                }
+              );
+            } else {
+              txWithCommon = new Transaction(tx, {
                 common: vm._common,
-              }
-            );
+              });
+            }
+          } else if (tx.type === 1) {
+            if (tx instanceof FakeSenderAccessListEIP2930Transaction) {
+              txWithCommon = new FakeSenderAccessListEIP2930Transaction(
+                tx.getSenderAddress(),
+                tx,
+                {
+                  common: vm._common,
+                }
+              );
+            } else {
+              txWithCommon = new AccessListEIP2930Transaction(tx, {
+                common: vm._common,
+              });
+            }
+          } else if (tx.type === 2) {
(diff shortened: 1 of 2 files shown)


alcuadrado:
Hey @shuklaayush! Thanks a lot for sending this PR. The fix and test are correct.

We already have a PR for this, which was lacking tests, so I added your tests into that one and merged it. It was really helpful, thanks.

shuklaayush:
@alcuadrado No worries, happy to see this fixed. I missed that PR since I couldn't find an issue related to this. 

Btw, I think you tagged the wrong 'ayush' in your comment https://github.com/nomiclabs/hardhat/pull/1586#issuecomment-902082030

alcuadrado:
Sorry about that! I meant to tag you

PR #1812, opened 2021-08-19
Bump hardhat-core to 2.6.1

fvictorio:


fvictorio:
Forking failures are all timeouts. e2e failure is a known issue on PRs that bump the version (I will look into that at some point).

PR #1813, opened 2021-08-19
Install Changesets

alcuadrado:
This PR installs [Changesets](https://github.com/atlassian/changesets), a tool that makes keeping track of changes and publishing new versions in a monorepo easier.

From now on, every PR changing something user-facing should come with its own changeset file. To generate one, you can run `yarn changeset` in the root of the project. A bot will post a warning in every PR that doesn't include one.

When new changesets are merged into master, a Github action opens/updates a PR with the list of changes that are pending release, their changelogs and the necessary modifications to the `package.json` files. Merging it doesn't automatically publish the new versions. Once merged, `yarn changeset publish` should be run in the root of the repo.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,0 +1,13 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@1.6.0/schema.json",
+  "changelog": "@changesets/cli/changelog",
+  "commit": false,
+  "linked": [],
+  "access": "public",
+  "baseBranch": "master",
+  "updateInternalDependencies": "minor",
+  "ignore": [],
+  "___experimentalUnsafeOptions_WILL_CHANGE_IN_PATCH": {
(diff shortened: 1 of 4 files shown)


alcuadrado:
> * [x]  Understand why sometimes "Updated dependencies [undefined]" has `undefined` in it, as shown [here](https://github.com/atlassian/changesets/pull/400).

The text between parens is the commit that contained the changeset that lead to the version updates. It's `undefined` if the changeset hasn't been committed, which shouldn't happen. `yarn changeset version` shouldn't be run with uncommitted changes.

alcuadrado:
> * [ ]  Create changesets for all the unreleased changes that we have
> *

I changed my mind about this. We should merge it now, otherwise the bot will be extremely annoying. I'll create an issue about the older changesets.

PR #1831, opened 2021-08-26
Add support for user-defined types as mapping keys

k06a:
Fixes: https://github.com/nomiclabs/hardhat/issues/1829 and https://github.com/nomiclabs/hardhat/issues/1564

Here is example of possibility to have this in Solidity:

https://github.com/solidity-parser/parser/blob/1de119f5e56f6bc924103148abac03f14d2e9a7c/test/ast.ts#L1873-L1887

Not tested, but it could be a good start to fix this issue :)


packages/hardhat-core/src/internal/hardhat-network/stack-traces/compiler-to-model.ts
```diff
@@ -565,6 +597,14 @@ function toCanonicalAbiType(type: string): string {
     return "ufixed128x128";
   }
 
+  if (isEnumType(type)) {
```
k06a, line 600: Should we revert these change?

Issue #1596, opened 2021-06-28
Replace `no-return-await` with `@typescript-eslint/return-await`


fvictorio:
We are using `"no-return-wait": "error"`, but it would've be better to do the opposite and to actually enforce `return await` everywhere, so that we have better async stack traces. See [this comment](https://github.com/nomiclabs/hardhat/pull/1587#discussion_r658937143).

Links:

- [`no-return-await`](https://eslint.org/docs/rules/no-return-await)
- [`@typescript-eslint/return-await`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/return-await.md)

alcuadrado:
Closed by #1835

PR #1835, opened 2021-08-29
Replace no-return-await with @typescript-eslint/return-await

istareatscreens:

---

This PR sets no-return-await to "off" and sets @typescript-eslint/return-await to "error" in eslintrc config to enforce return await as requested in #1596 . It also fixes all linting errors that occur as a result of the change.


alcuadrado:
Thanks for submitting this PR, @istareatscreens!

PR #1847, opened 2021-09-01
Fix newHeads subscription

ngotchac:

---

So it seems that the `newHeads` subscription actually receives some events when there is a call to a `pending` block, for example in `getTransactionsCount`. This is because it mines a new pending block to then run an action on this block, before reverting it. But when the pending block is mined, it emits a new-block event to the subscribers, which doesn't really make sense.

~So this PR fixes that, by adding a `bool` argument to the `mineBlock` method so the caller can choose if the method should notify subscribers or not.~

So here there's a bit of code duplication between `mineBlock` and `_runInPendingBlockContext`, regarding the next block timestamp computation ; but I couldn't find a better way to do this.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -2018,8 +2018,14 @@ Hardhat Network's forking functionality only works with blocks from at least spu
 
   private async _runInPendingBlockContext<T>(action: () => Promise<T>) {
     const snapshotId = await this.takeSnapshot();
+    const [blockTimestamp] = this._calculateTimestampAndOffset();
+    const needsTimestampIncrease =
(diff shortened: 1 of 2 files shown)


alcuadrado:
> So this PR fixes that, by adding a `bool` argument to the `mineBlock` method so the caller can choose if the method should notify subscribers or not.

I'm a bit lost. The changes don't reflect this description. Is this PR complete? Maybe you forgot to push some commits?

ngotchac:
> I'm a bit lost. The changes don't reflect this description. Is this PR complete? Maybe you forgot to push some commits?

Hey, actually that was the first way I found to fix the issue, but it's wasn't great so I updated my fix, but forgot to update the PR description :/ I'll do that now.

ngotchac:
Hey @alcuadrado ; I just rebased my PR. Would you be willing to give it another look? :)

PR #1857, opened 2021-09-05
fix: hasParts3 check in uncompressSourcemaps function

paulrberg:

---

I couldn't think of a test but there was definitely a bug on that line.


alcuadrado:
Good catch @paulrberg! Thanks!

I'll read the issue you linked later this week. It looks really interesting.

alcuadrado:
(merging while the CI was running as it passed and the changeset re-triggered it)

PR #1863, opened 2021-09-07
Request metadata json string by default.

chriseth:

---

It would be very nice if by default, the metadata artefact generated by the Solidity compiler could be stored at least somewhere. This PR (at least I hope) stores this information in the "build-info".

The metadata contains all required information to reproduce a Solidity build. If you have access to it and to the source files, you can get a byte-by-byte match on the bytecode regardless of which compilation tool you use. This means it is independent of the version of hardhat or all its dependencies and will _just work_ if you can get access to the Solidity compiler version mention in the metadata json file.

The most important piece of information stored inside the metadata file is the full path names of the source files. These are required to properly reproduce the build (imports would not work well otherwise).

The [sourcify](https://sourcify.dev) project will even be able to pick up the metadata json string inside the build-info file. This means in order to source-verify a hardhat project, it is enough to upload the sources and the files inside the `artifacts/build-info` directory, once this PR is accepted.


alcuadrado:
Hey @chriseth, thanks for sending this PR.

Does having the medatada enabled by default adds anything over storing the entire solc inpunt json as we do?

Could a tool just modify that solc input json to enable the metadata, recompile, and get the exact same output? Or would it be modified by this change?

chriseth:
If you directly have the metadata as a file, you can publish it on ipfs without having to recompile, but I can check if we can at least modify sourcify to also directly work with the input json - I was actually not aware that this was exactly the compiler input :)

alcuadrado:
Let me think a bit more about any possible implication of enabling the metadata by default, but this PR looks good.

> I can check if we can at least modify sourcify to also directly work with the input json

That would be great! I'd be much more comfortable recommending people to verify their sources on sourcify if it also accepted the solc input, as some users can still be affected by [this issue](https://github.com/ethereum/sourcify/issues/246#issuecomment-683852833). 

We have seen people affected by it when verifying on etherscan,  as we first using a minimal input first, equivalent to the metadata.

> I was actually not aware that this was exactly the compiler input :)

Yeah, we also store the compiler short and log version strings, so that we can always reproduce the builds, and even debug any possible error.

PR #1876, opened 2021-09-13
chore: update alchemy linkIm not sure if thats an active API key but it may be work dropping in favor of `<API_KEY>`

nick-staked:


fvictorio:
> not sure if thats an active API key but it may be work dropping in favor of <API_KEY>

It's not an active API key. I generally prefer to use "obviously" made-up placeholders instead of things like `<API_KEY>` because they give you an idea of the format and length of the thing you need to insert. But maybe it isn't obvious enough? I don't have a strong preference about this anyway.

fvictorio:
Merging this because the linting error is related to the dependencies issue that is fixed as part of #1884.

PR #1881, opened 2021-09-14
(hardhat-etherscan): add support verify for fuse network

developerfred:
This PR adds support for verifying smart contracts against [fuse explorer ](https://explorer.fuse.io/api-docs)

### DIFF
--- a/packages/hardhat-etherscan/src/network/prober.ts
+++ b/packages/hardhat-etherscan/src/network/prober.ts
@@ -38,6 +38,8 @@ enum NetworkID {
   POLYGON_MUMBAI = 80001,
   // Arbitrum
   ARBITRUM_ONE = 42161,
+  // Fuse
+  FUSE = 122,
 }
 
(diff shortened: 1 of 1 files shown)


alcuadrado:
Hey @developerfred, thanks for submitting this PR.

I'm not sure if it's correct though. Is explorer.fuse.io a custom deployment of etherscan? This plugin only works with etherscan and its variations for other networks.

developerfred:
> Hey @developerfred, thanks for submitting this PR.
> 
> I'm not sure if it's correct though. Is explorer.fuse.io a custom deployment of etherscan? This plugin only works with etherscan and its variations for other networks.

yes it is a customized version like Matic, Xdai and other evm explorers...

alcuadrado:
Thanks for the quick reply! I read more of its documentation, and figured that it's a deployment of blockscout, and this plugin doesn't work with it.

I think a hardhat-blockscout plugin would be useful, but we can't prioritize it at the moment.

PR #1884, opened 2021-09-15
Make e2e tests work when version is bumped

fvictorio:
This fixes the issue with the e2e tests not working when we bumped the version. It also simplifies the setup significantly, I believe.

Before this, the way tests were executing worked like this:

- Build the hardhat tgz
- Copy the `package.json.template` files to `package.json` files
- Install hardhat in each fixture project
- Then for each fixture project:
  - Copy it to a temporary directory
  - Run the tests in that temp dir
- Remove the hardhat tgz
- Clean up the fixture projects

The problem here is that there's a mix of things: some temporary files were created inside the fixture projects and then these directories were moved to a temporary directory, where they were further modified (for example, installing more stuff).

The new setup seems much better to me:

- Build the hardhat tgz
- For each fixture project:
  - Copy it to a temporary directory
  - Install hardhat using the tgz file
  - Run the tests in that temp dir
- Remove the hardhat tgz

The extra complexity added by this is that now `run-tests.js` needs to "pass" information to the tests, and this is done using a couple of environment variables. But I think the benefits of a much simpler setup (and that now the tests work for a bumped version) compensate for it.

---

This also fixes the `check-dependencies.js` issue introduced in the last release.


packages/e2e/run-tests.js
```diff
@@ -16,26 +15,25 @@ if (process.argv[2] !== "npm" && process.argv[2] !== "yarn") {
...
+  });
   shell.set("-e");
 
-  cleanup(fixtures);
-  // we remove the tgz file because some tests might reinstall the fixture
-  // project dependencies, and that could fail if the tgz file isn't there
-  // anymore
+  // we remove the tgz file after the tests are run, because some of them might
+  // reinstall the fixture project dependencies, and that would fail if the tgz
+  // file isn't there anymore
```
feuGeneA, line 36: This comment is confusing to me.

"after the tests are run" ... as opposed to what?

And, it's saying, "we remove the tgz file...because...[some things] would fail if the tgz file isn't there anymore" ... seems counterintuitive.

feuGeneA (approved):

packages/e2e/run-tests.js
fvictorio: Thanks, this was indeed not very useful. I updated the comment. Does it look better now?

feuGeneA: It does.  But, it makes me wonder if that's the right thing to do.  When the tests fail and need investigation, it's super nice to have that .tgz around so that I can easily replicate the failure.  And re-acquainting myself with the code enough to find that line and comment it out feels pretty inefficient.  Would it be possible/practical to do any of:
* leave the file behind in the event of failure,
* store the file in /tmp/whatever instead of the package directory so that we can leave it behind all the time without causing too much noise, or
* do both: put it in /tmp, leave it in the case of failure, delete it in the case of success.

(To be clear, my Approval isn't contingent on answering these questions, much less implementing the decision.)

fvictorio: That makes sense, but for now I guess we can just comment out that line if we need to keep the tgz file. This approach wouldn't be enough if the tests fail non-deterministically (so we can't just re-run them and comment out that line) **and** the tgz is important to understand why it failed, which seems very unlikely to me.

(Merging the PR, but we can continue discussing this anyway!)

PR #1913, opened 2021-09-29
Add smock to plugin list

0xGorilla:


alcuadrado:
Hey @0xGorilla, i think the link in url is wrong and that's why the build is failing. Can you fix that and also remove the older smock entry? Thanks!

0xGorilla:
Thanks for the quick reply! I believe the error happens because in smock we do use `README.rst` instead of `README.md` so we can import it in readthedocs.

Any idea of how to solve this? By the way, I just pushed the previous smock plugin deletion.

image

PR #1915, opened 2021-09-30
Force empty Uncles for forked-nodes

ngotchac:

---

There is an issue with the forked-node, when handling blocks with uncles. It seems that the list of uncles is omitted since the methods to find the uncles are not implemented (IIUC), however the hash passed to the RPC responses is the actual real header uncles hash. If the real block had uncles, some libraries would expect the uncles list to be not-empty (cf. [geth](https://github.com/ethereum/go-ethereum/blob/f2491c5ed77b9554bd979a426db741a99acf6fa4/ethclient/ethclient.go#L121-L123))

This PR fixes this by ensuring we always respond with the uncles hash set as the hash of an empty list.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/fork/rpcToBlockData.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/fork/rpcToBlockData.ts
@@ -1,4 +1,5 @@
 import { BlockData } from "@ethereumjs/block";
+import { KECCAK256_RLP_ARRAY } from "ethereumjs-util";
 
 import { RpcBlockWithTransactions } from "../../../core/jsonrpc/types/output/block";
 
@@ -8,7 +9,8 @@ export function rpcToBlockData(rpcBlock: RpcBlockWithTransactions): BlockData {
   return {
     header: {
       parentHash: rpcBlock.parentHash,
-      uncleHash: rpcBlock.sha3Uncles,
+      // We don't implement uncles - set the hash as the hash of an empty array
+      uncleHash: KECCAK256_RLP_ARRAY,
       coinbase: rpcBlock.miner,
       stateRoot: rpcBlock.stateRoot,
       transactionsTrie: rpcBlock.transactionsRoot,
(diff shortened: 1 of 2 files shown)


alcuadrado:
Hey @ngotchac, sorry we had you waiting for so long! I'm going through your PRs right now.

I'm a bit hesitant about this one, as changing the `uncleHash` property of the header should lead to a different block hash. I'm worried that some users may get the JSON-RPC serialization of the block and compute the hash based on that. For instance, ethers.js does something similar for some transaction fields (computes the sender based on the signature iirc).

What do you think? Maybe we should leave this as is and implement the RPC methods related to uncles?

ngotchac:
Hey @alcuadrado ! Thanks for reviewing my last PRs :)
So regarding your question, I actually stumbled upon the issue using Geth's `ethclient` [here](https://github.com/ethereum/go-ethereum/blob/e761255ba7e19ec7c380f008f9035da3caa77994/ethclient/ethclient.go#L120-L126) ; so I get an error since the uncle hash doesn't match.
I don't know if it would break other projects, but the current implementation is wrong so I'd suggest finding a solution, maybe in a new breaking release?

alcuadrado:
Are you getting this error? https://github.com/ethereum/go-ethereum/blob/e761255ba7e19ec7c380f008f9035da3caa77994/ethclient/ethclient.go#L125

ngotchac:
> Are you getting this error? https://github.com/ethereum/go-ethereum/blob/e761255ba7e19ec7c380f008f9035da3caa77994/ethclient/ethclient.go#L125 -- any way to reproduce it?
> 
> I'll try to figure out why the actual hashes were missing.

Yes exactly. The issue is when forking and fetching a block with an uncle, the uncle list sent back from Hardhat will be empty, whereas the uncle hash will not be the right one (since it's forwarding the right one from mainnet). Another solution would be to implement uncles fetching

vince91:
> Are you getting this error? https://github.com/ethereum/go-ethereum/blob/e761255ba7e19ec7c380f008f9035da3caa77994/ethclient/ethclient.go#L125 -- any way to reproduce it?
> 
> I'll try to figure out why the actual hashes were missing.

https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts#L282

At line 282, the property `uncles` is lost after the call to `rpcToBlockData`, the returned type `BlockData` from @ethereumjs does not have a `uncles` property. 

I created an issue with step to reproduce.

ngotchac:
Hey @alcuadrado ; I just rebased my PR. Would you be willing to give it another look? :)

PR #1917, opened 2021-10-01
Add Troubleshooting documentation for mainnet forking

thatguyintech:

Multiple users have been confused after seeing this issue:
https://github.com/nomiclabs/hardhat/issues/1507

Adding this note to the troubleshooting section will help clarify
ways to solve this problem.

### DIFF
--- a/docs/hardhat-network/guides/mainnet-forking.md
+++ b/docs/hardhat-network/guides/mainnet-forking.md
@@ -104,4 +104,8 @@ This will reset Hardhat Network, starting a new instance in the state described 
 
 When using Infura without the archival add-on, you will only have access to the state of the blockchain from recent blocks. To avoid this problem, you can use either a local archive node or a service that provides archival data, like [Alchemy].
 
+### "Error HH109: Network connection timed-out. Please check your internet connection and networks config"
+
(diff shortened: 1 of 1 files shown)


fvictorio:
Thanks a lot @thatguyintech, this seems really useful.

> One way to fix this is to try connecting to a different internet network.

Can you clarify what you mean by that? Using a different VPN, a different IPS, or something else?

thatguyintech:
> Using a different VPN, a different ISP, or something else?

Yeah I think all of these may help, it's not clear to me which one seems to be the more common culprit. I'll add that into the docs to be more specific about things people can try!

thatguyintech:
@fvictorio updated! lmk how that looks

thatguyintech:
@fvictorio is this copy ok?

thatguyintech:
@feuGeneA any thoughts?

alcuadrado:
Re-reading this PR I got the idea/impression that we are throwing a misleading error when HH109 is thrown from the forking functionality's JSON-RPC client. I think what we should do instead is catching it [here](https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/hardhat-network/jsonrpc/client.ts#L319) and throwing something with more context.

fvictorio:
Hey @thatguyintech, sorry for the auto-close and for not merging this PR.

We recently changed the structure of our site significantly, so if you want to open a new PR with this change, we'll try to get it merged as soon as we can.

thatguyintech:
Awesome, I'll take a look

PR #1922, opened 2021-10-03
Set the default NVM version to 16

wesfloyd:
NVM versions < 16 cause a memory fault issue on Macs with the M1 processor at this step: 
https://hardhat.org/tutorial/testing-contracts.html 

because of this error: 
https://stackoverflow.com/questions/65856300/wasm-code-commit-allocation-failed-process-out-of-memory

Hopefully this is helpful?

---

### DIFF
--- a/docs/tutorial/setting-up-the-environment.md
+++ b/docs/tutorial/setting-up-the-environment.md
@@ -30,10 +30,10 @@ Make sure you have `git` installed. Otherwise, follow [these instructions](https
 There are multiple ways of installing Node.js on MacOS. We will be using [Node Version Manager (nvm)](http://github.com/creationix/nvm). Copy and paste these commands in a terminal:
 
 ```
-curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.35.2/install.sh | bash
-nvm install 12
-nvm use 12
-nvm alias default 12
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash
+nvm install 16
+nvm use 16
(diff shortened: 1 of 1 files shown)


fvictorio:
We usually use the lowest supported node version in the docs. I don't think changing it is a big deal, but I don't know if the fact that it doesn't work with some machines is enough reason to do it (especially because I guess M1 users know they will have compatibility issues).

@alcuadrado wdyt?

alcuadrado:
Can you give more info on how to reproduce this, @wesfloyd? I have an M1 mac.

wesfloyd:
Yes @alcuadrado, steps to reproduce the bug:

- (Testing on a 2021 M1 Macbook air with 8GB memory)
- Followed the [tutorial steps 2-5](https://hardhat.org/tutorial/testing-contracts.html)
- At Step 5 when I attempt to run "npx hardhat test" to see validation of the first test case
- Expected result: "Token contract ✓ Deployment should assign the total supply of tokens to the owner"
- Actual result:  I get the "Allocation failed - process out of memory" error.

Is this helpful?

alcuadrado:
Hey @wesfloyd, I haven't had time to take a look at this yet, but thanks for the instructions. They should be enough.

alcuadrado:
I couldn't reproduce this, but I suspect it has to do with my environment being tuned to work with node.js on an M1. I tried to start from a clean slate, but I probably failed to do it.

Should we merge this PR anyway, @fvictorio? node@16 is the active LTS version.

fvictorio:
@alcuadrado sure

Issue #1666, opened 2021-07-14
When forking a remote network, `eth_call` and `eth_estimateGas` can be run with an incorrect hardfork


alcuadrado:
If you fork a remote chain, and call `eth_call`, `eth_estimateGas`, or any other rpc method that requires (re)executing a tx, with a block tag pointing to a remote block, we may use the wrong hardfork, and hence get different results.

The reason for this is that we always use the currently active hardfork, while Geth derives the hardfork from the block tag.

This can only affect you if you are forking a remote chain, as otherwise all the blocks will be mined using the same hardfork.

Fixing this can be trickier than expected, as the forking functionality should support any chain, not just the ones natively supported by ethereumjs. I think we have two options for non-native chains:

(1) Keep doing what we do now, risking to give incorrect results.
(2) Fail when a remote block tag is used. Indicating how to ask for support for it.

feuGeneA:
Backlink: https://github.com/nomiclabs/hardhat/blob/854a149835ed0763bb7396189f299fd094409f61/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts#L2135

feuGeneA:
Released in 2.8.0.

PR #1928, opened 2021-10-04
Permit configuration of hardfork activation blocks

feuGeneA:
For #1666, allow `HardhatNode` user to configure the block numbers at which specific hardforks are to be applied.

- [x] Consider additional changes for estimateGas, or determine that the existing changes already cover that case.
- [x] Add an entry to the docs in the Mainnet Forking section that provides more explanation of this feature than the current reference entry. And add a link from the Reference entry to the explanation in Mainnet Forking. And update the URL in the `Could not find a hardfork` error message in `HardhatNode._selectHardforkFromActivations()`. \[Update: completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/f90781a33b5443cd1fbd458f8679c7cb79e6d054 ]
- [x] Also consider `debug_traceTransaction` and whether it needs any additional changes.  It creates its own VM/common/whatever, and that custom thing will need to have the proper hardfork set on it. [Update: completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/d37850b0c152f28427a427a96054890381084f9d ]
- [x] Store a record of common networks and their hardfork activation history, and apply those records during the `HardhatNode` instantiation process, based on the `forkNetworkId` in use.
- [x] Move user config from within `forking` to be under `networks.hardhat.chains`. See https://github.com/nomiclabs/hardhat/pull/1928#discussion_r737804346 .  Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6095b4a4348804b2680603f4108d65b0792788bb
- [x] Remove the optionality from the hardfork history property in the non-`User` config object. Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6095b4a4348804b2680603f4108d65b0792788bb
- [x] Split each of `HardforkActivationHistory` and `HardforkActivationsByChain` into separate types for `*UserConfig` and non-`UserConfig` types, and, for the non-`User` config objects, use a `Map` rather than an index-signatured object. (Keep the index-signatured object form for the `UserConfig` objects.) Source: https://github.com/nomiclabs/hardhat/pull/1928#discussion_r737743303 . Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6095b4a4348804b2680603f4108d65b0792788bb
- [x] Merge default configs with user-specified configs, per https://github.com/nomiclabs/hardhat/pull/1928#discussion_r737745979 . Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6095b4a4348804b2680603f4108d65b0792788bb
- [x] Test the merging of default configs with user-specified configs, per https://github.com/nomiclabs/hardhat/pull/1928#discussion_r737746220 . Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/e6e6b8cf4f9de8a597ca51011c095693ea7f8b39
- [x] Check _supported_ hardforks only at runtime, and at config validation time simply validate that the specified hardforks are _valid_.  Source: # 1 in https://github.com/nomiclabs/hardhat/pull/1928#pullrequestreview-791005179 .  See also https://github.com/nomiclabs/hardhat/pull/1928#discussion_r737803902 and https://github.com/nomiclabs/hardhat/pull/1928#discussion_r737820939 . Completed for validation time in https://github.com/nomiclabs/hardhat/pull/1928/commits/e6e6b8cf4f9de8a597ca51011c095693ea7f8b39 , and for runtime in https://github.com/nomiclabs/hardhat/pull/1928/commits/57b07bcb4d8dc0c0de49a43ab9669d919a7da735 .
- [x] Fix bug in tracing changes: https://github.com/nomiclabs/hardhat/pull/1928#discussion_r733086837 . Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/e8ac4fd1ce097d549e31d74264a48bc0f2b87153
- [x] Add an optional block number parameter to `HardhatNode.isEip1559Active()` and use that block number to examine the hardfork history in order to make the determination of EIP 1559 or not, and ensure that all of `runCall()`, `_getEstimateGasFeePriceFields()` and `_runTxAndRevertMutations()` pass in the block number to `isEip1559Active()`. Source: #'s 2 & 3 in https://github.com/nomiclabs/hardhat/pull/1928#pullrequestreview-791005179 . Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6a0581dc609caab12c7ec6f12b2107c7ade945c8
- [x] Change `getCommonForTracing()` so that, in the case of a historical block, it sets the hardfork based on the hardfork history, rather than just `common.activeHardfork(blockNumber)`.  Source: # 4 in https://github.com/nomiclabs/hardhat/pull/1928#pullrequestreview-791005179 . Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/68d7abf65bfc865d1ff6cf796c6fa952f98f1577
- [x] Move the "is this a historical block?" checks into `_selectHardforkFromActivations` and return `this._vm._common.harfork()` if they are not true. Source: https://github.com/nomiclabs/hardhat/pull/1928#discussion_r737817008 . Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/8c299e903c9d7f8ae4ae379b8a6a45f7a8ef0ebc
- [x] Raise an Issue about our use of `forkNetworkId` in `HardhatNode`. We should be using `forkChainId` instead.
- [x] Validate hardfork names just before using them. Source: https://github.com/nomiclabs/hardhat/pull/1928/#discussion_r737820939 . Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/194528604185e0c80f591ae4fe5a2cd14fa41ee4
- [x] Address comment threads about documentation.
- [x] Move the "is historical" condition into the `_selectHardfork()` method. ([Source](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746868949))  Outcome: Verified that `_selectHardfork()` is already using `this._vm._common.hardfork()` in the non-historical case.)
- [x] Stop validating the default hardfork name at execution time (since it's already been validated). ([Source](https://github.com/nomiclabs/hardhat/pull/1928/files/199053ebb11514426e8a4d152bf33c680ca1cd99#r746874575)) Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/e05a3bc697d76b21f7a476828bbccc5d335b69dd
- [x] Eliminate the now-redundant "is historical?" check from `isEip1559Active()` (now that the previous checklist item's change to `_selectHardfork()`, which is used by `isEip1559Active()`). ([Source](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746870927)). Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/f4b28f5c0ddd6304171f3267b64982e8b00c8f97 .
- [x] Change `_getCommonForTracing()` again, so that rather than just calling `setHardfork()`, pass chain params into the local `Common` instantation. ([Source](https://github.com/nomiclabs/hardhat/pull/1928/files/199053ebb11514426e8a4d152bf33c680ca1cd99#r746876184)) Completed in https://github.com/nomiclabs/hardhat/pull/1928/commits/36cbc626c470d10c4e2f6a795ad32b3ee89d9018 .
- [x] Stop resolving config to defaults at node instantiation time (since that already happened at config resolution time). ([Source](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746863010))
- [x] Stop validating hardfork names at node instantiation time (since we already do it at execution time). ([Source](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746872110))
- [x] Instead of mutating the common each time, create a new one each time, based on `this._vm._common` but with some different params (i.e. chain id, network id, hardfork). ([Source](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746935880).)
- [x] Move and improve `No HF history config'd` error message, from node instantiation to `_selectHardfork()`. (It may already be there already, in which case just confirm & consolidate.) (Sources: [move](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746863967), [improve](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746866545))
- [x] Improve `Unsupported hardfork name` error message. ([Source](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746866731)) Consider `Trying to run a call or transaction using a block whose hardfork is ${x}, but it's not supported by Hardhat Network`. ([Source](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746872962))
- [x] Make error message more descriptive when rejecting hardfork names in config validation. ([Source](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746805357))
- [x] Respect the `UserConfig` suffix. (Sources: [1](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746779381), [2](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746781910))
- [x] Raise an Issue: come up with a list of all the chains that we want to support by default, and add them to `defaultHardhatNetworkParams.chains`. ([Source](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746784950))
- [x] Await @fvictorio 's response [here](https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746783575), then likely change config types to just use `string` instead of exposing the internal type `HardforkName`.
- [x] Fix failing test scenario. ([Source](https://github.com/nomiclabs/hardhat/pull/1928#issuecomment-973164285))
- [x] Rebase on `development` branch.
- [x] Add a changeset.


packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -318,7 +341,8 @@ Hardhat Network's forking functionality only works with blocks from at least spu
     genesisAccounts: GenesisAccount[],
     tracingConfig?: TracingConfig,
     private _forkNetworkId?: number,
-    nextBlockBaseFee?: BN
+    nextBlockBaseFee?: BN,
+    private readonly _hardforkActivationBlocks: HardforkActivationBlocks = mainnetHardforkActivations
```
feuGeneA, line 345: This doesn't make sense to default to mainnet because users will use other networks too (testnets, etc).

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -245,7 +264,11 @@ export class HardhatNode extends EventEmitter {
       genesisAccounts,
       tracingConfig,
       forkNetworkId,
-      nextBlockBaseFeePerGas
+      nextBlockBaseFeePerGas,
+      hardforkActivationBlocks !== undefined &&
+      hardforkActivationBlocks[chainId] !== undefined
```
feuGeneA, line 269: Use `forkNetworkId`, not `chainId`, because it's only used for the remote one.

feuGeneA: Can possibly use `hardforkActivationBlocks?[chainId] ?? undefined`.  Or will `?[` work? Isn't that just `?.`? but isn't `[x]` the same as `.x`?

packages/hardhat-core/src/internal/hardhat-network/provider/node-types.ts
```diff
@@ -15,6 +16,11 @@ export function isForkedNodeConfig(
   return "forkConfig" in config && config.forkConfig !== undefined;
 }
 
+export type HardforkActivationBlocks = Record<
```
feuGeneA, line 19: Bad naming here.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2297,6 +2325,37 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
     return { maxFeePerGas, maxPriorityFeePerGas };
   }
+
+  /** search this._hardforkActivationBlocks for the highest block number that
+   * isn't higher than blockNum, and then return that found block number's
+   * associated hardfork name. */
+  private _selectHardforkFromActivations(
+    blockNum: BN
+  ): typeof HARDHAT_NETWORK_SUPPORTED_HARDFORKS[number] {
+    let highestFound: {
```
feuGeneA, line 2335: If we're not running a forked node, we should immediately return the local hardfork, the one that was used in the constructor parameter for the node.

feuGeneA: Also need to just use the constructor-given hardfork in the case where the block number to execute against is higher than the block number that was forked from.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2297,6 +2325,37 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+      hardfork: typeof HARDHAT_NETWORK_SUPPORTED_HARDFORKS[number];
+      block: number;
+    } = { hardfork: "", block: 0 };
+    Object.entries(this._hardforkActivationBlocks).forEach((entry) => {
+      const [hardfork, block] = entry;
+      if (!HARDHAT_NETWORK_SUPPORTED_HARDFORKS.includes(hardfork)) {
```
feuGeneA, line 2341: Can move this to Hardhat Runner's config validation logic. At this level of abstraction we assume everything has already been validated.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2297,6 +2325,37 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+        throw new InvalidInputError(
+          `hardforkActivationBlocks contained unsupported hardfork "${hardfork}"`
+        );
+      }
+      if (block > highestFound.block && new BN(block).lte(blockNum)) {
+        highestFound = { hardfork, block };
+      }
+    });
+    if (highestFound.hardfork === "" && highestFound.block === 0) {
+      throw new InvalidJsonInputError(
```
feuGeneA, line 2351: Throw `InternalError` instead.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2297,6 +2325,37 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+        `No known hardfork for block ${blockNum}. HardhatNode config's hardforkActivationBlocks: ${JSON.stringify(
```
feuGeneA, line 2352: This error message needs to be a lot better.  Explain the situation, and give a link to the documentation.

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -86,7 +87,10 @@ export class HardhatNetworkProvider
     private readonly _initialDate?: Date,
     private readonly _experimentalHardhatNetworkMessageTraceHooks: BoundExperimentalHardhatNetworkMessageTraceHook[] = [],
     private _forkConfig?: ForkConfig,
-    private readonly _forkCachePath?: string
+    private readonly _forkCachePath?: string,
+    private readonly _hardforkActivationBlocks?: {
```
feuGeneA, line 91: Consider creating a named type for this chainId mapping, like perhaps `HardforkActivationBlocksPerChain`. That will help resolve the bad naming of this field vs the type.

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkCommon.ts
```diff
@@ -3,12 +3,13 @@ import Common from "@ethereumjs/common";
 import { ForkedNodeConfig } from "../node-types";
 
 export async function makeForkCommon(config: ForkedNodeConfig) {
+  const { chainId, networkId, networkName } = config;
```
feuGeneA, line 6: Changes to this file are unnecessary.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2154,6 +2178,10 @@ Hardhat Network's forking functionality only works with blocks from at least spu
         (blockContext.header as any).baseFeePerGas = new BN(0);
       }
 
+      this._vm._common.setHardfork(
```
feuGeneA, line 2181: Isn't this affecting the next thing that runs against `this._vm._common`? We should:
1. surround subsequent `runTx()` with a `try`/`finally` with a `finally` that re-sets the hardfork to what it was before this call execution.
2. Add a test for this.

feuGeneA (commented):
All of these comments were generated on a live review call with @alcuadrado .

Create an issue about the JSON-RPC layer always validating things using the local hardfork, which doesn't make a lot of sense when forking from a remote network and running eth_call/eth_estimateGas/debug_traceTransaction on older blocks. 

Add in config support at the hardhat network level. Take these things into consideration:
user config is first "validated" and then "resolved" (defaults assigned).
`HardhatNetworkUserConfig` type has lots of optionality, but the "resolved" version is just `HardhatNetworkConfig` and has little to no optinality.
the code that transforms user config to resolved config is in core/src/internal/core/config/config-resolution.ts
cf. config-validation.ts
2 things to note:
    - the types in config.ts should not import anything from other modules (hardhat-network, etc) but rather repeat the values as needed, because plugins can extend those types.
    - we try not to load ethereumjs in config-resolution.ts, because loading it is very slow, and hardhat always loads this config-resolution at every startup and we don't want to slow down the startup.

packages/hardhat-core/src/internal/core/config/config-resolution.ts
```diff
@@ -148,9 +149,15 @@ function resolveHardhatNetworkConfig(
...
-  const blockNumber = hardhatNetworkConfig?.forking?.blockNumber;
-  if (blockNumber !== undefined && forking !== undefined) {
-    forking.blockNumber = hardhatNetworkConfig?.forking?.blockNumber;
+  if (forking !== undefined) {
+    const blockNumber = hardhatNetworkConfig?.forking?.blockNumber;
+    if (blockNumber !== undefined) {
+      forking.blockNumber = hardhatNetworkConfig?.forking?.blockNumber;
+    }
+
+    forking.hardforkActivationsByChain =
```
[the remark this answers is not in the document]
feuGeneA: Addressed the code change, but not the test, in Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6095b4a4348804b2680603f4108d65b0792788bb

packages/hardhat-core/src/types/config.ts
```diff
@@ -69,10 +69,20 @@ export interface HDAccountsUserConfig {
...
 
+export type HardforkActivationHistory = Record<
+  string /* hardfork name */,
+  number /* block number */
+>;
+
+export interface HardforkActivationsByChain {
+  [chainId: number]: HardforkActivationHistory;
+}
+
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6095b4a4348804b2680603f4108d65b0792788bb

packages/hardhat-core/src/internal/core/config/config-resolution.ts
```diff
@@ -152,9 +153,11 @@ function resolveHardhatNetworkConfig(
...
     forking.blockNumber = hardhatNetworkConfig?.forking?.blockNumber;
   }
-  const activations = hardhatNetworkConfig?.forking?.hardforkActivationsByChain;
-  if (activations !== undefined && forking !== undefined) {
-    forking.hardforkActivationsByChain = activations;
+
+  if (forking !== undefined) {
+    forking.hardforkActivationsByChain =
+      hardhatNetworkConfig?.forking?.hardforkActivationsByChain ??
+      defaultHardforkActivationsByChain;
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6095b4a4348804b2680603f4108d65b0792788bb

packages/hardhat-core/src/internal/core/config/config-resolution.ts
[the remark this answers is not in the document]
feuGeneA: Test is in https://github.com/nomiclabs/hardhat/pull/1928/commits/e6e6b8cf4f9de8a597ca51011c095693ea7f8b39

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1255,6 +1275,15 @@ Hardhat Network's forking functionality only works with blocks from at least spu
             );
           }
 
+          if (
+            this._forkBlockNumber !== undefined &&
+            blockNumber < this._forkBlockNumber
+          ) {
```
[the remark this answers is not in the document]
feuGeneA: Addressed this somewhat in https://github.com/nomiclabs/hardhat/pull/1928/commits/e1a246261008fbf582601e9e1a0db68f62022f15 .  I like what it did to the function, and how it approved _some_ of the cases you're talking about.  However, in some areas of the code, we only do `setHardfork()` under certain conditions, so the need to keep those local "is historical?" checks in place anyways, so the impact of this idea turned out to be less than it seemed it would be.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: https://github.com/nomiclabs/hardhat/issues/2015

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -319,6 +348,31 @@ export function getValidationErrors(config: any): string[] {
...
+        hardhatNetwork.forking.hardforkActivationsByChain !== undefined
+      ) {
+        Object.entries(
+          hardhatNetwork.forking.hardforkActivationsByChain
+        ).forEach((chainEntry) => {
+          const [, hardforkHistory] = chainEntry;
+          Object.entries(hardforkHistory as HardforkActivationHistory).forEach(
+            (hardforkEntry) => {
+              const [hardforkName] = hardforkEntry;
+              if (!HARDHAT_NETWORK_SUPPORTED_HARDFORKS.includes(hardforkName)) {
```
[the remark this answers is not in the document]
feuGeneA: Addressed:
* Validation-time check: https://github.com/nomiclabs/hardhat/pull/1928/commits/9a92bcc006e855c137dfb42285486156a8b72d3b#diff-cd6765512f8f5e1431a675a3ede5d86714e7665970a45b1f75c97245fdc7cab8R143
* Runtime check: https://github.com/nomiclabs/hardhat/pull/1928/commits/57b07bcb4d8dc0c0de49a43ab9669d919a7da735#diff-f931c37110b4492bfc050d8112ad82e5be4ad8f3ca6bc615ac31cdb19e80453dR213

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2297,6 +2340,44 @@ Hardhat Network's forking functionality only works with blocks from at least spu
 
     return { maxFeePerGas, maxPriorityFeePerGas };
   }
+
+  private _selectHardforkFromActivations(
+    blockNumber: BN
+  ): typeof HARDHAT_NETWORK_SUPPORTED_HARDFORKS[number] {
+    /** search this._hardforkActivations for the highest block number that
```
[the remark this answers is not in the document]
feuGeneA: I did this somewhat in [this commit](https://github.com/nomiclabs/hardhat/pull/1928/commits/e1a246261008fbf582601e9e1a0db68f62022f15), but see my comment [here](https://github.com/nomiclabs/hardhat/pull/1928/#discussion_r740665517).

packages/hardhat-core/src/types/config.ts
```diff
@@ -141,6 +151,7 @@ export interface HardhatNetworkForkingConfig {
   enabled: boolean;
   url: string;
   blockNumber?: number;
+  hardforkActivationsByChain?: HardforkActivationsByChain;
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/9a92bcc006e855c137dfb42285486156a8b72d3b#diff-bddef7387c11db950d8bb99684ba2e222d54844e4930e3b4adf83f0edd3b3f20

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2154,6 +2184,16 @@ Hardhat Network's forking functionality only works with blocks from at least spu
         (blockContext.header as any).baseFeePerGas = new BN(0);
       }
 
+      if (
+        this._forkBlockNumber !== undefined &&
+        blockContext!.header.number.lt(new BN(this._forkBlockNumber))
+      ) {
+        originalHardfork = this._vm._common.hardfork();
+        this._vm._common.setHardfork(
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/ab41e1d6c715547d507676e28400f247071f9b40

packages/hardhat-core/src/internal/core/config/config-validation.ts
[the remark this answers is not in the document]
feuGeneA: I inadvertently ended up doing the runtime check at a different place than you requested. I did the runtime check during instantiation of a `HardhatNode` object, as already cited above.  Then I came back and re-read the comments and saw "right before using it," so I made an additional change for that, [here](https://github.com/nomiclabs/hardhat/pull/1928/commits/194528604185e0c80f591ae4fe5a2cd14fa41ee4).

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
feuGeneA: I tried to address this, but ran into test failures.

The change I made: https://github.com/nomiclabs/hardhat/pull/2016

The test failure:
```
   1) Forked provider
       Using Alchemy
         eth_call
           when used in the context of a past block
             when the block number is less or equal to the fork block number
               does not affect previously added storage data:
     Error: BlockHeader.cliqueSigner() call only supported for clique PoA networks
      at BlockHeader._requireClique (/home/runner/work/hardhat/hardhat/node_modules/@ethereumjs/block/src/header.ts:692:13)
      at BlockHeader.cliqueSigner (/home/runner/work/hardhat/hardhat/node_modules/@ethereumjs/block/src/header.ts:797:10)
      at VM._runTx (/home/runner/work/hardhat/hardhat/node_modules/@ethereumjs/vm/src/runTx.ts:422:28)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
      at VM.runTx (/home/runner/work/hardhat/hardhat/node_modules/@ethereumjs/vm/src/runTx.ts:182:20)
      at HardhatNode._runTxAndRevertMutations (src/internal/hardhat-network/provider/node.ts:2228:14)
      at HardhatNode._runInBlockContext (src/internal/hardhat-network/provider/node.ts:2002:14)
      at HardhatNode.runCall (src/internal/hardhat-network/provider/node.ts:529:20)
      at EthModule._callAction (src/internal/hardhat-network/provider/modules/eth.ts:353:9)
      at HardhatNetworkProvider._sendWithLogging (src/internal/hardhat-network/provider/provider.ts:131:22)
      at HardhatNetworkProvider.request (src/internal/hardhat-network/provider/provider.ts:108:18)
      at Context.<anonymous> (test/internal/hardhat-network/provider/forked-provider.ts:162:15)
```
Full failure result is at https://github.com/nomiclabs/hardhat/runs/4075291823?check_suite_focus=true

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2297,6 +2340,44 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+      }
+    });
+    if (highestFound.hardfork === "" && highestFound.block === 0) {
+      throw new InternalError(
+        `Could not find a hardfork to run for block ${blockNumber}, after having looked for one in the HardhatNode's hardfork activation history, which was: ${JSON.stringify(
+          activations
+        )}. For more information, see https://hardhat.org/hardhat-network/reference/#config`
+      );
+    }
+    return highestFound.hardfork;
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/194528604185e0c80f591ae4fe5a2cd14fa41ee4

docs/hardhat-network/reference/README.md
```diff
@@ -13,52 +13,90 @@
...
+- An array of the initial accounts that the Hardhat Network will create. Each of them must be an object with `privateKey` and `balance` fields.
+
+#### `blockGasLimit`
+
+The block gas limit to use in Hardhat Network's blockchain. Default value: `30_000_000`
+
+#### `minGasPrice`
+
+The minimum gas price accepted by the Hardhat Network. Transactions with a lower gas price are accepted but not mined. Default value: `0`
+
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/bd54c6df09d8c16fc37ab71c8f531a0ed8a0c977

docs/hardhat-network/reference/README.md
```diff
@@ -13,52 +13,90 @@
 
 Default behavior: Hardhat will automatically run in-process when `defaultNetwork` is empty or set to `hardhat`.
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/8729b1e467526afe77f0999477c84423b1e8f9d4

docs/hardhat-network/reference/README.md
```diff
@@ -13,52 +13,90 @@
...
 
+### Supported Fields
+
 You can set the following fields on the `hardhat` config:
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6ee7ac51724412b09a2acb739298d349ce566470

docs/hardhat-network/guides/mainnet-forking.md
```diff
@@ -98,10 +98,42 @@ await network.provider.request({
...
+          berlin: 10000000,
+          london: 20000000,
+        },
+      }
+    }
+  }
+}
+```
+
+In this context, a "historical block" is one whose number is prior to `forking.blockNumber`. If you try to run code in the context of a historical block, _without_ having a hardfork history, then an error will be thrown. The known hardfork histories of most public networks are assumed as defaults.
```
alcuadrado, line 123:
```suggestion
In this context, a "historical block" is one whose number is prior to the block you forked from. If you try to run code in the context of a historical block, _without_ having a hardfork history, then an error will be thrown. The known hardfork histories of most public networks are assumed as defaults.
```

forking.blockNumber is not always present

packages/hardhat-core/src/types/config.ts
```diff
@@ -25,6 +26,18 @@ export type NetworkUserConfig =
...
 
+export interface HardforkHistoryUserConfig {
+  [hardforkName: string]: number /* block number */;
+}
+
+export interface HardhatNetworkChainUserConfig {
+  hardforkHistory?: HardforkHistoryUserConfig;
+}
+
+export interface HardhatNetworkChainUserConfigByChainId {
```
alcuadrado, line 37: I'd try to respect the `UserConfig` suffix. What about `HardhatNetworkChainByChainIdUserConfig`?

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -191,6 +197,24 @@ export class HardhatNode extends EventEmitter {
           }
         }
       }
+
+      if (config.chains.has(forkNetworkId)) {
+        hardforkActivations = config.chains.get(forkNetworkId)!.hardforkHistory;
+      } else if (defaultHardhatNetworkParams.chains.has(forkNetworkId)) {
```
alcuadrado, line 203: This `else if` shouldn't be here, as the default config is handled by hardhat, not hardhat network, and already resolved here.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -191,6 +197,24 @@ export class HardhatNode extends EventEmitter {
...
+        hardforkActivations =
+          defaultHardhatNetworkParams.chains.get(
+            forkNetworkId
+          )!.hardforkHistory;
+      } else {
+        throw new InternalError(
+          `No hardfork history configured for chain ID ${forkNetworkId}`
+        );
+      }
+      hardforkActivations.forEach((block, hardforkName) => {
```
alcuadrado, line 213: I'd also validate this when needed, to make this less disruptive. This is not necessary, but my reasoning is as follows:

- The node is initialized in the first request to the provider.
- Making this validation in the very first request will lead to an HFs related error in a seemingly unrelated request, like `eth_chainId`.
- Making it when we need to use a the unsupported hardfork will show the error in a situation where the user has more context.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -191,6 +197,24 @@ export class HardhatNode extends EventEmitter {
...
+        if (!HARDHAT_NETWORK_SUPPORTED_HARDFORKS.includes(hardforkName)) {
+          throw new InternalError(`Unsupported hardfork ${hardforkName}`);
```
alcuadrado, line 215: Same for this error message. It should give a better explanation.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
alcuadrado: I think that we are doing these checks to avoid setting the hardfork to the same value that it already has. If we move the "is historical" condition into the `_selectHardfork` method, we will be setting the HF more often that we strictly need to, but i believe that's innocuos.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2273,7 +2328,19 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+      block !== undefined &&
+      this._forkBlockNumber !== undefined &&
+      block.lt(new BN(this._forkBlockNumber))
+    ) {
+      return this._vm._common.hardforkGteHardfork(
+        this._selectHardfork(block),
+        "london"
+      );
+    }
     return this._vm._common.gteHardfork("london");
```
alcuadrado, line 2344: I think this can also be simplified if _selectHardfork also had the "is historic" condition inside

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: update: This is already validated in `_selectHardfork` so it can be removed

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2313,18 +2383,74 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
-function getCommonForTracing(networkId: number, blockNumber: number): Common {
-  try {
-    const common = new Common({ chain: networkId });
+  private _selectHardfork(
+    blockNumber: BN,
+    defaultName = this._vm._common.hardfork() as HardforkName
+  ): HardforkName {
+    function assertSupported(n: HardforkName) {
+      if (!HARDHAT_NETWORK_SUPPORTED_HARDFORKS.includes(n)) {
+        throw new InternalError(`Unsupported hardfork ${n}`);
```
alcuadrado, line 2393: This error should give more context about what's going on, like:

`Trying to run a call or transaction using a block whose hardfork is ${x}, but it's not supported by Hardhat Network`

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2313,18 +2383,74 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+      }
+    }
+    if (
+      this._forkBlockNumber === undefined ||
+      blockNumber.gte(new BN(this._forkBlockNumber))
+    ) {
+      assertSupported(defaultName);
```
alcuadrado, line 2400: I don't think this assert is necessary. That's already validated.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2313,18 +2383,74 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+      throw new InternalError(
+        `No known hardfork for execution on historical block ${blockNumber.toString()} (relative to fork block number ${
+          this._forkBlockNumber
+        }). The node was not configured with a hardforkActivationHistory.  See http://hardhat.org/hardhat-network/guides/mainnet-forking.html#using-a-custom-hardfork-history`
+      );
+    }
+  }
 
-    common.setHardfork(common.activeHardfork(blockNumber));
+  private _getCommonForTracing(networkId: number, blockNumber: number): Common {
```
alcuadrado, line 2438: The change here doesn't fully address what we discussed, as it still fails if you want to trace a chain that's not supported by ethereumjs

packages/hardhat-core/src/types/config.ts
alcuadrado: `HardhatNetworkChainsUserConfig`

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -191,6 +197,24 @@ export class HardhatNode extends EventEmitter {
...
       }
+
+      if (config.chains.has(forkNetworkId)) {
+        hardforkActivations = config.chains.get(forkNetworkId)!.hardforkHistory;
+      } else if (defaultHardhatNetworkParams.chains.has(forkNetworkId)) {
+        hardforkActivations =
+          defaultHardhatNetworkParams.chains.get(
+            forkNetworkId
+          )!.hardforkHistory;
+      } else {
```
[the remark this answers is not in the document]
alcuadrado: I think that this validation is already present in `_selectHardfork`, but i'm not completely sure.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: This value here is config.networks.hardhat.hardfork, which is already validated to be a supported one.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: Update: The new approach for creating commons does address this by default, as we are going to be imitating the logic of `Common.forCustomChain`. The same method would be used for standard and non-standard chains.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/e05a3bc697d76b21f7a476828bbccc5d335b69dd

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: I confirmed that _selectHardfork has sufficient "is historic" checks, and then removed the redundant "is historic" checks in https://github.com/nomiclabs/hardhat/pull/1928/commits/f4b28f5c0ddd6304171f3267b64982e8b00c8f97

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/36cbc626c470d10c4e2f6a795ad32b3ee89d9018

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
feuGeneA: Here's a different failure to look at.  I tried to make the change to create a new VM/Common each time, but it caused some test failures that aren't immediately obvious to me.  See the change and the failure in https://github.com/nomiclabs/hardhat/pull/2059

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/852ae99c486919a431acbc11fc1ce60b17507313

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6889614eecae24597dae138bc80312a5d9cd46e2

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/ae26cf55748f9ce3700e1936b8785295d890e94e

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -117,12 +118,55 @@ const HardhatNetworkHDAccountsConfig = t.type({
...
   url: t.string,
   blockNumber: optional(t.number),
 });
 
+function isValidHardforkName(name: string) {
+  return Object.values(HardforkName).includes(name as HardforkName);
+}
+
+const HardforkNameType = new t.Type<HardforkName>(
+  "HardforkName",
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/489abcbd0339d20740b5932c8762a466b71ae5ab

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
feuGeneA: I confirmed that this validation is indeed already present in `_selectHardfork()`, and I removed the redundant validation from this area, in [this commit](https://github.com/nomiclabs/hardhat/pull/1928/commits/b031c4ed03d5eae992244e268e2eea3cb62a29b3).

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -191,6 +197,24 @@ export class HardhatNode extends EventEmitter {
...
+        throw new InternalError(
+          `No hardfork history configured for chain ID ${forkNetworkId}`
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/f6a5dbe1fe1aa1007192aa4e2cf31e1f1c7c5426

packages/hardhat-core/src/types/config.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/f59333aebda4a5a77fd995da2ca7b60125b2b7bf

packages/hardhat-core/src/types/config.ts
```diff
@@ -100,6 +114,17 @@ export interface NetworksConfig {
 
 export type NetworkConfig = HardhatNetworkConfig | HttpNetworkConfig;
 
+export type HardforkHistoryConfig = Map<HardforkName, /* blockNumber */ number>;
+
+export interface HardhatNetworkChainConfig {
+  hardforkHistory: HardforkHistoryConfig;
+}
+
+export type HardhatNetworkChainConfigByChainId = Map<
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/f59333aebda4a5a77fd995da2ca7b60125b2b7bf

packages/hardhat-core/src/types/config.ts
```diff
@@ -100,6 +114,17 @@ export interface NetworksConfig {
 
 export type NetworkConfig = HardhatNetworkConfig | HttpNetworkConfig;
 
+export type HardforkHistoryConfig = Map<HardforkName, /* blockNumber */ number>;
```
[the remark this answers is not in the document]
feuGeneA: I drafted this up here: https://github.com/nomiclabs/hardhat/pull/2060

packages/hardhat-core/src/internal/core/config/default-config.ts
```diff
@@ -47,6 +48,60 @@ export const defaultHardhatNetworkParams: Omit<
   loggingEnabled: false,
   gasMultiplier: DEFAULT_GAS_MULTIPLIER,
   minGasPrice: new BN(0),
+  chains: new Map([
+    [
+      1, // mainnet
```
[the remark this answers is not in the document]
feuGeneA: I raised https://github.com/nomiclabs/hardhat/issues/2061

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: This particular `throw` is gone now, but you can see the improved error messaging in [this commit](https://github.com/nomiclabs/hardhat/pull/1928/commits/f6a5dbe1fe1aa1007192aa4e2cf31e1f1c7c5426).

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/f6a5dbe1fe1aa1007192aa4e2cf31e1f1c7c5426

packages/hardhat-core/src/types/config.ts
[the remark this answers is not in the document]
fvictorio: I'm not sure. I agree that something from `internal` shouldn't be used, but what if we make that enum public? That doesn't sound like a terrible idea.

> Also, maybe we should use string, as it would make this type more future-proof.

Is that so? If we use an enum instead of a string, and then we add support for a new hardfork, the user will have to upgrade hardhat anyway, and so the updated type will be available.

That being said, I don't mind using a string here, provided that:

- When a new hardfork value is added, we get compilation errors in all the right places
- The supported values are validated on runtime.

But I think that will be very close to also having these values typed in the config, and that's a slightly better experience for typescript users, so... maybe just doing that it's better.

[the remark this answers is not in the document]
fvictorio: Oh, I realize now that this is the resolved config, not the user config. I think my comment still applies though, but it would mean also adding it to the user config.

(Anyway, just to be clear: I don't mind using a string, provided we have internal type-safety around the supported hardforks and proper runtime validation, if that lets us merge this sooner)

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2313,18 +2367,79 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+    defaultName = this._vm._common.hardfork() as HardforkName
+  ): HardforkName {
+    if (
+      this._forkBlockNumber === undefined ||
+      blockNumber.gte(new BN(this._forkBlockNumber))
+    ) {
+      return defaultName;
+    }
+    if (this._hardforkActivations === undefined) {
+      throw new InternalError(
```
fvictorio, line 2382: I'm pretty sure this error will never be thrown. The reason is that `this._hardforkActivations` can never be undefined because it's initialized like this:

```typescript
let hardforkActivations: HardforkHistoryConfig = new Map();
```

The fix is just changing that line to:

```typescript
    let hardforkActivations: HardforkHistoryConfig | undefined;
```

(I verified this locally.)

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -730,6 +737,144 @@ describe("HardhatNode", () => {
...
+            constant: true,
+            inputs: [],
+            name: "Hello",
+            outputs: [{ name: "", type: "string" }],
+            payable: false,
+            stateMutability: "pure",
+            type: "function",
+          },
+        ])
+      );
```
fvictorio, line 815:
```suggestion
      const contractInterface = new ethers.utils.Interface([
        "function Hello() public pure returns (string)",
      ]);
```

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -730,6 +737,144 @@ describe("HardhatNode", () => {
...
+    };
+
+    const [, regularNode] = await HardhatNode.create(forkedNodeConfig);
+
+    const nodeCfgWithEarlyLondon = forkedNodeConfig;
+    let earlyLondonMainnetConfig = forkedNodeConfig.chains.get(1);
+    if (earlyLondonMainnetConfig === undefined) {
+      earlyLondonMainnetConfig = { hardforkHistory: new Map() };
+    }
+    earlyLondonMainnetConfig.hardforkHistory.set(
```
fvictorio, line 781: Isn't this modifying (by shared references) the `defaultHardhatNetworkParams.chains` map?

Even if it's not, it isn't obvious to me that it isn't. Maybe this could start with something like:

```js
const nodeCfgWithEarlyLondon = {
  ...forkedNodeConfig,
  chains: new Map(forkedNodeConfig.chains)
}
```

just to be sure.

fvictorio: Wait, no, that will do a shallow copy. Oh well.

docs/hardhat-network/guides/mainnet-forking.md
```diff
@@ -98,10 +98,42 @@ await network.provider.request({
...
+If you're forking an unusual network, and if you want to execute EVM code in the context of a historical block retrieved from that network, then you will need to configure Hardhat Network to know which hardforks to apply to which blocks. (If you're forking a well-known network, Hardhat Network will automatically choose the right hardfork for the execution of your EVM code, based on known histories of public networks, so you can safely ignore this section.)
+
+For example, to configure a hardfork activation history for a network with `chainId` 99:
+
+```js
+networks: {
+  hardhat: {
+    forking: {
+      url: "https://your.node/",
+      hardforkActivationsByChain: {
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/ba781a599d08ee43e71786f117f1b047d4097164

docs/hardhat-network/reference/README.md
```diff
@@ -11,54 +11,86 @@
...
+An optional boolean that disables the contract size limit imposed by the [EIP 170](https://eips.ethereum.org/EIPS/eip-170). Default value: `false`
+
+#### `forking`
+
+An object that describes the [forking](../guides/mainnet-forking.md) configuration that can have the following fields:
+
+- `url`: a URL that points to a JSON-RPC node with state that you want to fork off. There's no default value for this field. It must be provided for the fork to work.
+- `blockNumber`: an optional number to pin which block to fork from. If no value is provided, the latest block is used.
+- `enabled`: an optional boolean to switch on or off the fork functionality. Default value: `true` if `url` is set, `false` otherwise.
+- `hardforkActivationsByChain`: an optional object whose properties are chain ID's and whose values are objects with properties representing hardfork names and values representing the block numbers at which those named hardforks were activated on their respective chains. For example, the recent Ethereum mainnet hardfork history might be represented as `{ 1: { berlin: 12244000, london: 12965000 } }`.
```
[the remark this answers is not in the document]
feuGeneA: Addressed all of these things in https://github.com/nomiclabs/hardhat/pull/1928/commits/ba781a599d08ee43e71786f117f1b047d4097164

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -117,12 +118,57 @@ const HardhatNetworkHDAccountsConfig = t.type({
...
+  HardforkNameType,
+  t.number,
+  "HardhatNetworkHardforkHistory"
+);
+
+const HardhatNetworkChainConfig = t.type({
+  hardforkHistory: HardhatNetworkHardforkHistory,
+});
+
+const HardhatNetworkChainsConfig = t.record(Integer, HardhatNetworkChainConfig);
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6f641f1c08ede380c462a78cd5e894259a2d469e

packages/hardhat-core/src/internal/hardhat-network/provider/node-types.ts
```diff
@@ -28,8 +28,14 @@ interface CommonConfig {
...
   tracingConfig?: TracingConfig;
   initialBaseFeePerGas?: number;
+  chains: HardhatNetworkChainsConfig;
 }
 
+export {
+  HardhatNetworkChainsConfig,
+  HardhatNetworkChainConfig,
+} from "../../../types";
+
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6f641f1c08ede380c462a78cd5e894259a2d469e

packages/hardhat-core/src/internal/core/config/default-config.ts
```diff
@@ -47,6 +48,60 @@ export const defaultHardhatNetworkParams: Omit<
...
+      },
+    ],
+    [
+      42, // kovan
+      {
+        hardforkHistory: new Map([
+          [HardforkName.BYZANTIUM, 5067000],
+          [HardforkName.CONSTANTINOPLE, 9200000],
+          [HardforkName.PETERSBURG, 10255201],
+          [HardforkName.ISTANBUL, 14111141],
```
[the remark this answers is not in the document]
feuGeneA: I guess ethereumjs just hasn't updated this file? https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/common/src/chains/kovan.json#L65

packages/hardhat-core/src/internal/core/config/default-config.ts
```diff
@@ -47,6 +48,60 @@ export const defaultHardhatNetworkParams: Omit<
   loggingEnabled: false,
   gasMultiplier: DEFAULT_GAS_MULTIPLIER,
   minGasPrice: new BN(0),
+  chains: new Map([
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/5bd58779e95d9712acc0f3956b89c32d611f7203

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: I took a slightly different approach. I decided to change the check rather than the initialization, in [this commit](https://github.com/nomiclabs/hardhat/pull/1928/commits/6acba23bf668ecc283032382cfac936f354c64b5), because Pato has previously expressed that there should definitely always be a hardfork history given in node construction, because the optionality only exists at the `UserConfig` layer.  What do you think?

packages/hardhat-core/src/internal/core/config/config-validation.ts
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/0d50f837bfee36afdfcf6bc7497cc00a3352eeed

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2313,18 +2367,79 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+        `No known hardfork for execution on historical block ${blockNumber.toString()} (relative to fork block number ${
+          this._forkBlockNumber
+        }). The node was not configured with a hardforkActivationHistory.  See http://hardhat.org/hardhat-network/guides/mainnet-forking.html#using-a-custom-hardfork-history`
+      );
+    }
+    /** search this._hardforkActivations for the highest block number that
+     * isn't higher than blockNumber, and then return that found block number's
+     * associated hardfork name. */
+    const activations = this._hardforkActivations!; // yes !, just asserted it.
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/6acba23bf668ecc283032382cfac936f354c64b5

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2170,6 +2198,17 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
 
+      originalCommon = (this._vm as any)._common;
+      (this._vm as any)._common = new Common({
+        chain: {
+          // eslint-disable-next-line @typescript-eslint/dot-notation
+          ...this._vm._common["_chainParams"],
+          chainId: this._configChainId,
+          networkId: this._configNetworkId,
+        },
+        hardfork: this._selectHardfork(blockContext!.header.number),
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/82f205f03ed11fc5a2e8ef05808415291434e680

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2313,18 +2367,79 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
-}
 
-function getCommonForTracing(networkId: number, blockNumber: number): Common {
-  try {
-    const common = new Common({ chain: networkId });
+  private _selectHardfork(
+    blockNumber: BN,
+    defaultName = this._vm._common.hardfork() as HardforkName
+  ): HardforkName {
+    if (
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/3cf332dd39e27c6c8f728cf40687000694a8962b

packages/hardhat-core/test/internal/core/config/config-resolution.ts
```diff
@@ -645,6 +647,76 @@ describe("Config resolution", () => {
...
+      });
+
+      describe("chains", function () {
+        it("should default to default", function () {
+          const resolvedConfig = resolveConfig(__filename, {});
+          assert.deepEqual(
+            Array.from(resolvedConfig.networks.hardhat.chains.entries()),
+            Array.from(defaultHardhatNetworkParams.chains.entries())
+          );
+          for (const chain of Array.from(
```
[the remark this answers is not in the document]
feuGeneA: Addressed the redundancy in https://github.com/nomiclabs/hardhat/pull/1928/commits/c0b71672f2a7a914916adc1f7ffff9f9bc1953ad

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2170,6 +2198,17 @@ Hardhat Network's forking functionality only works with blocks from at least spu
         (blockContext.header as any).baseFeePerGas = new BN(0);
       }
 
+      originalCommon = (this._vm as any)._common;
```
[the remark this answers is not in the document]
feuGeneA: @alcuadrado wdyt?

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
fvictorio: Yes, that should work too :+1:

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/d9848e5ddb39815d39b45f443b92a5c2a80e4fac .  Let me know what you think of that function I made, and specifically whether it should be exported from that location, and even more specifically whether we should be exporting it at all.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -730,6 +737,144 @@ describe("HardhatNode", () => {
       });
     }
   });
+
+  it("should run calls in the right hardfork context", async function () {
+    // fork mainnet at the block when EIP-1559 activated (12965000), and try to
+    // run a call that specifies gas limits in EIP-1559 terms, but run that
+    // call one block earlier, and expect that call to fail because it should
+    // have specified gas limits in PRE-EIP-1559 terms.
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/dbbd20df2f2932a654cee706782a957d780037da

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
feuGeneA: We're now running with a fresh `Common` each time, addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/ae26cf55748f9ce3700e1936b8785295d890e94e

packages/hardhat-core/src/types/config.ts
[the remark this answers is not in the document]
feuGeneA: The type-safety around valid (not supported, but valid) hardfork names happens via this: https://github.com/nomiclabs/hardhat/pull/1928/files#diff-cd6765512f8f5e1431a675a3ede5d86714e7665970a45b1f75c97245fdc7cab8R146

The runtime validation happens here: https://github.com/nomiclabs/hardhat/pull/1928/files#diff-f931c37110b4492bfc050d8112ad82e5be4ad8f3ca6bc615ac31cdb19e80453dR2409

packages/hardhat-core/src/internal/core/config/config-validation.ts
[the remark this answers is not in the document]
alcuadrado: Note: it's actually because keys of an object ARE string

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1255,6 +1280,8 @@ Hardhat Network's forking functionality only works with blocks from at least spu
             );
           }
 
+          vm._common.setHardfork(this._selectHardfork(new BN(blockNumber)));
```
alcuadrado, line 1283: This line surprises me a lot. Shouldn't we directly construct the new common (i.e. in `this._getCommonForTracing`) with the right hf?

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
alcuadrado: It's a public variable, that we already used. If they remove it in a pr their ci, which builds HH, will fail.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2170,6 +2198,17 @@ Hardhat Network's forking functionality only works with blocks from at least spu
         (blockContext.header as any).baseFeePerGas = new BN(0);
       }
 
+      originalCommon = (this._vm as any)._common;
+      (this._vm as any)._common = new Common({
+        chain: {
+          // eslint-disable-next-line @typescript-eslint/dot-notation
+          ...this._vm._common["_chainParams"],
+          chainId: this._configChainId,
+          networkId: this._configNetworkId,
```
alcuadrado, line 2207: I think these are incorrect. If the block is a historic one we should be using the remote network's values.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: upadte: it already does

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2313,18 +2367,81 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+        },
+        hardfork: this._selectHardfork(new BN(blockNumber)),
+      });
+
+      common.setHardfork(
+        this._selectHardfork(
+          new BN(blockNumber),
+          common.activeHardfork(blockNumber) as HardforkName
+        )
+      );
```
alcuadrado, line 2438: This is duplicated, as the hardfork was already set in the constructor. Also, i'll feel more comfortable if we don't mutate custom commons, as that has lead to weird results before.

packages/hardhat-core/src/types/config.ts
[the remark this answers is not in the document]
alcuadrado: I'd be ok with a test that leads to the same guarantees that franco is asking for.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
fvictorio: I think I would just put it in the same test file (or at least in some `test/**/utils.ts` file or something, if for some reason you don't want to add it to that file). `src/types` is a public module, so doing unnecessary exports there is not a good idea.

packages/hardhat-core/src/types/config.ts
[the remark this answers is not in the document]
feuGeneA: Addressed with the merging of #2060

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
fvictorio: I still find this whole test very hard to understand. We should have at least one top-comment explaining the approach. Another way to make it clearer is to add error messages to the different assertions (something that, admittedly, we don't usually do, but it might be worth in this case to make the test more self-explanatory).

fvictorio:
A scenario that is not being correctly handled is this:

```
      chains: {
        99: {
          hardforkHistory: {
            london: 100
          }
        }
      }
```

In that case, if a call is made with a block below 100, the hardfork cannot be obtained and it should fail. That doesn't happen right now.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/bdcd35e3428a7ab79256eb9661be0a5bfcc0c83c

packages/hardhat-core/src/internal/core/config/default-config.ts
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/c94da84e58ad579f461c0056ac4127dc35d7f785

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/d57f8f1962ca22e6747861dab8456a0fb00e2da8

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: I think this is leftover code (:grimacing:) from when we weren't instantiating a new common. Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/3f50f0d620206c19642a9c82eecd8847073a6822

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/1928/commits/b1c59f1d82c0c0157d8e90a9416e07c1323f1e3c

docs/hardhat-network/reference/README.md
```diff
@@ -11,54 +11,95 @@
...
+
+#### `chains`
+
+An object that configures chain-specific options. Each key is a number representing a chain ID, and each value is an object configuring the chain with that ID. In the inner object, the following fields are supported:
+
+- `hardforkHistory`: an object whose keys are strings representing hardfork names (eg `"london"`, `"berlin"`) and whose values are numbers specifying the block at which that hardfork was activated.
+
+The default value includes configurations for several well known chains (eg mainnet, chain ID `1`); using this field is only useful when forking unusual networks. The user may override the defaults for some chain ID's while leaving the defaults in place for other chain ID's. Overriding the default for a chain ID will replace the entire configuration for that chain.
+
+For more details, see [Using a custom hardfork history](../guides/mainnet-forking.md#using-a-custom-hardfork-history].
```
fvictorio, line 94:
```suggestion
For more details, see [Using a custom hardfork history](../guides/mainnet-forking.md#using-a-custom-hardfork-history).
```

packages/hardhat-core/test/internal/core/config/config-resolution.ts
```diff
@@ -645,6 +647,60 @@ describe("Config resolution", () => {
...
+            networks: {
+              hardhat: { chains: { 1: { hardforkHistory: { london: 999 } } } },
+            },
+          };
+          const resolvedConfig = resolveConfig(__filename, userConfig);
+          it("If the user provides values for a chain that's included in the default, should use the users' values, and ignore the defaults for that chain.", function () {
+            assert.deepEqual(resolvedConfig.networks.hardhat.chains.get(1), {
+              hardforkHistory: new Map([[HardforkName.LONDON, 999]]),
+            });
+            assert.deepEqual(resolvedConfig.networks.hardhat.chains.get(1), {
```
fvictorio, line 673: This second assert is not necessary, right?

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -730,6 +757,188 @@ describe("HardhatNode", () => {
...
+
+        nodeConfig.chains.set(1, chainConfig);
+
+        [, nodeWithEarlyLondon] = await HardhatNode.create(nodeConfig);
+      });
+
+      it("should accept post-EIP-1559 gas semantics when running in the context of the block of the EIP-1559 activation", async function () {
+        assert.equal(
+          "Hello World",
+          await runCall(post1559GasOpts, blockBefore1559, nodeWithEarlyLondon)
```
fvictorio, line 888: These tests are much, much clearer now, thanks.

This one is a bit surprising but I think this behavior is fine (and even if it isn't, I don't think we should put a lot of thought into what happens if someone uses the wrong hardfork block numbers).

feuGeneA:
> A scenario that is not being correctly handled is this:
> 
> ```
>       chains: {
>         99: {
>           hardforkHistory: {
>             london: 100
>           }
>         }
>       }
> ```
> 
> In that case, if a call is made with a block below 100, the hardfork cannot be obtained and it should fail. That doesn't happen right now.

Demonstrated in https://github.com/nomiclabs/hardhat/pull/1928/commits/1df6bca45aa734fea312774da98933c0214f867e and fixed in https://github.com/nomiclabs/hardhat/pull/1928/commits/2c614dc5931b0ed781de7575e99d1e76f0e962fb

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2313,18 +2365,74 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
-  } catch (e) {
-    throw new InternalError(
-      `Network id ${networkId} does not correspond to a network that Hardhat can trace`
+    /** search this._hardforkActivations for the highest block number that
+     * isn't higher than blockNumber, and then return that found block number's
+     * associated hardfork name. */
+    const hardforkHistory: Array<[name: string, block: number]> = Array.from(
+      this._hardforkActivations.entries()
+    );
+    const [hardfork] = hardforkHistory.reduce(
```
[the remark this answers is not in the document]
feuGeneA: Fixed in https://github.com/nomiclabs/hardhat/commit/2c614dc5931b0ed781de7575e99d1e76f0e962fb

packages/hardhat-core/test/internal/core/config/config-resolution.ts
feuGeneA: Removed in https://github.com/nomiclabs/hardhat/pull/1928/commits/540c97ffeb3f16ca75a53c559bad66608f6a4402

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2313,18 +2365,74 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
     return { maxFeePerGas, maxPriorityFeePerGas };
   }
-}
 
-function getCommonForTracing(networkId: number, blockNumber: number): Common {
-  try {
-    const common = new Common({ chain: networkId });
+  private _selectHardfork(
+    blockNumber: BN,
+    defaultName = this._vm._common.hardfork() as HardforkName
```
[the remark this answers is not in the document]
feuGeneA: Removed that param in https://github.com/nomiclabs/hardhat/pull/1928/commits/04804be39615fcb2a65dfd2b642c2f69445d0e03

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
feuGeneA: Resolving this in light of https://github.com/nomiclabs/hardhat/pull/1928#discussion_r752680478

fvictorio (approved):

alcuadrado (approved):
Sorry for the delay

PR #1945, opened 2021-10-12
Workaround antlr prototype pollution issue

alcuadrado:
This PR works around an issue reported by @ajsantander.

### DIFF
--- a/.changeset/stale-rockets-leave.md
+++ b/.changeset/stale-rockets-leave.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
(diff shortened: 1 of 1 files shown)


alcuadrado:
I think a better approach would be to have a function/factory to initialize the hardhat context and hre. There's a lot of unnecessary repetition there. I'll open an issue about this.

Should we merge and release this as is now, and tackle the repetition in the future?

fvictorio:
> Should we merge and release this as is now, and tackle the repetition in the future?

Sure!

PR #1954, opened 2021-10-15
gas-providers.ts - fix klaytn rpc

klaytnmapdev:
https://github.com/nomiclabs/hardhat/issues/1953

There is no "web3_clientVersion" method in Klaytn RPC. Added exception handling code.

### DIFF
--- a/packages/hardhat-core/src/internal/core/providers/gas-providers.ts
+++ b/packages/hardhat-core/src/internal/core/providers/gas-providers.ts
@@ -312,11 +312,16 @@ export class GanacheGasMultiplierProvider extends MultipliedGasEstimationProvide
 
   private async _isGanache(): Promise<boolean> {
     if (this._cachedIsGanache === undefined) {
-      const clientVersion = (await this._wrappedProvider.request({
-        method: "web3_clientVersion",
-      })) as string;
+      try {
+        const clientVersion = (await this._wrappedProvider.request({
+          method: "web3_clientVersion",
+        })) as string;
 
-      this._cachedIsGanache = clientVersion.includes("TestRPC");
+        this._cachedIsGanache = clientVersion.includes("TestRPC");
+      } catch (error) {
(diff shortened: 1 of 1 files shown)


fvictorio:
Hi @klaytnmapdev. As far as I know, `web3_clientVersion` is a very common method, so maybe the right fix here is to get it implemented in whichever client Klaytn uses (I'm not familiar with the project). We try to avoid this kind of chain-specific code if we can.

wighawag:
I had a similar issue with some goerli node
is `web3_clientVersion` part of any spec? if not I think catching the case it it is not implemented like this PR do is a good idea

fvictorio:
Hmm, that's weird. In that case, maybe it makes sense to add this change.

@alcuadrado wdyt?

blockjoe:
While `web3_clientVersion` is pretty common, it's not defined as a method that Ethereum clients must support.

The specification of what methods all Ethereum clients must support can be found here: https://github.com/ethereum/execution-apis.

danielattilasimon:
This is also affecting the Kiln testnet. Apparently, `web3_clientVersion` isn't implemented there either:
```javascript
fetch("https://rpc.kiln.themerge.dev", {
  method: "POST",
  headers: { "Content-Type": "application/json" },
  body: JSON.stringify({
    jsonrpc: "2.0",
    method: "web3_clientVersion",
    params: [],
    id: 1337,
  }),
})
  .then(response => response.json())
  .then(console.log);
```

Which currently yields:
```
{
  "jsonrpc": "2.0",
  "id": 1337,
  "error": {
    "code": -32601,
    "message": "the method web3_clientVersion does not exist/is not available"
  }
}
```

fvictorio:
Re-opening this since it seems that several nodes don't have this method.

blockjoe:
Want to just bring attention to this since, again, this is not part of the Ethereum JSON-RPC spec.

fvictorio:
You are right @blockjoe, we will revisit this.

PR #1994, opened 2021-10-26
[WIP] add esm support 

1dcbh:
WIP! I just created the PR to pull locally, will run tests are report back soon.


phated:
The way that hardhat imports tasks and plugins also won't work with ESM because all ESM modules are executed async and hardhat plugins often rely on the ordering of those imports.

alcuadrado:
That's an excellent point, @phated! thanks

fvictorio:
Everything suggests that adding support for ESM is a major endeavor that we won't be able to do until our next major (v3, which is not even in our roadmap). So I think we should close this, unless you see some way forward @1dcbh.

PR #2008, opened 2021-10-29
extra tests for privkey validation

### LINKED ISSUE #1635: Improve error message for invalid private key
This is a valid config:

```js
module.exports = {
  solidity: "0.7.3",
  networks: {
    rinkeby: {
      url: "https://rinkeby.infura.io/v3/",
      accounts: ["0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"]
    }
  }
};
```

But adding or removing a single character to the private key (or changing an `a` to a `g`) results in this error message:

```
Error HH8: There's one or more errors in your config file:

  * Invalid value {"url":"https://rinkeby.infura.io/v3/","accounts":["0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"]} for HardhatConfig.networks.rinkeby - Expected a value of type HttpNetworkConfig.
```

There is a bigger issue here about our config validation being awful in general, but I think we should handle this particular case right now, since it shows up **a lot** in the discord server, and it's a really bad experience for users.

morgansliman:
Closes #1635

### DIFF
--- a/.changeset/rude-lemons-turn.md
+++ b/.changeset/rude-lemons-turn.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+extra tests for privkey validation
--- a/.changeset/rude-lemons-turn.md
+++ b/.changeset/rude-lemons-turn.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-extra tests for privkey validation
+Improve validation of private keys in the Hardhat config
--- a/packages/hardhat-core/src/internal/core/config/config-validation.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-validation.ts
@@ -3,11 +3,6 @@ import { Context, getFunctionName, ValidationError } from "io-ts/lib";
 import { Reporter } from "io-ts/lib/Reporter";
 
 import {
-  HardhatConfig,
-  HardhatNetworkAccountsUserConfig,
-  HardhatNetworkAccountUserConfig,
-} from "../../../types";
-import {
   HARDHAT_NETWORK_NAME,
   HARDHAT_NETWORK_SUPPORTED_HARDFORKS,
 } from "../../constants";
--- a/packages/hardhat-core/src/internal/core/config/config-validation.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-validation.ts
@@ -305,7 +305,7 @@ export function getValidationErrors(config: any): string[] {
 
       // manual validation of accounts
       if (Array.isArray(accounts)) {
-        for (const [index, account] of Object.entries(accounts)) {
+        for (const [index, account] of accounts.entries()) {
           if (typeof account !== "object") {
             errors.push(
               getPrivateKeyError(
@@ -319,7 +319,7 @@ export function getValidationErrors(config: any): string[] {
 
           const { privateKey, balance } = account;
 
-          validatePrivateKey(privateKey, +index, HARDHAT_NETWORK_NAME, errors);
+          validatePrivateKey(privateKey, index, HARDHAT_NETWORK_NAME, errors);
 
           if (typeof balance !== "string") {
             errors.push(
@@ -430,11 +430,7 @@ export function getValidationErrors(config: any): string[] {
       } else if (typeof accounts === "string") {
         if (accounts !== "remote") {
           errors.push(
-            getPrivateKeyError(
-              0,
-              networkName,
-              `Expected "remote", received ${accounts}`
-            )
+            `Invalid 'accounts' entry for network '${networkName}': expected an array of accounts or the string 'remote', but got the string '${accounts}'`
           );
         }
       } else if (accounts !== undefined) {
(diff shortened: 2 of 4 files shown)


packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -201,6 +201,19 @@ Please run: npm install --save-dev typescript`,
 Please run this and try again: \`npm install --save-dev typescript\``,
       shouldBeReported: false,
     },
+    INVALID_PRIVATE_KEY: {
```
alcuadrado, line 204: I think that instead of adding an error type i'd still use invalid config, but making sure that the error message is clear. @fvictorio, is this what you had in mind?

fvictorio: Yes. Something similar to the other "manual" checks we do for the config that are not part of `io-ts` validation.

morgansliman: Got it :+1:

morgansliman:
Oh, forgot there's other tests that will fail now..

.changeset/rude-lemons-turn.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+extra tests for privkey validation
```
fvictorio, line 5:
```suggestion
Improve validation of private keys in the Hardhat config
```

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -237,8 +289,7 @@ export function getValidationErrors(config: any): string[] {
       }
 
       // Validating the accounts with io-ts leads to very confusing errors messages
-      const configExceptAccounts = { ...hardhatNetwork };
-      delete configExceptAccounts.accounts;
+      const { accounts, ...configExceptAccounts } = hardhatNetwork;
```
fvictorio, line 292: Nice

fvictorio:
I need to take a deeper look at the code, but before doing that, I think the error messages could be improved a bit.

### Current messages

image

image

### Possible improvements

image

image

fvictorio:
There seems to be a test case missing for this part of the code:

image

I added a commit installing `nyc` because for some reason we had it configured but not installed. You can run `yarn nyc mocha test/internal/core/config/config-validation.ts` and check the `coverage/index.html` to see this output.

(For the record: I'm not a fan of measuring test coverage, but it's a useful tool for this particular kind of code, to check if some obvious test scenario is missing.)

fvictorio:
Ugh, installing `nyc` breaks the build for some reason, and I don't want to go down that rabbit hole. So I dropped my commit. You can still use it with `npx nyc --reporter html mocha path/to/test`

morgansliman:
> ### Possible improvements
> image

I was also debating about whether to use indices rather than the value. I'll make that change.

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -318,7 +319,7 @@ export function getValidationErrors(config: any): string[] {
 
           const { privateKey, balance } = account;
 
-          validatePrivateKey(privateKey, HARDHAT_NETWORK_NAME, errors);
+          validatePrivateKey(privateKey, +index, HARDHAT_NETWORK_NAME, errors);
```
fvictorio, line 322: If you use `accounts.entries` instead of `Object.entries(accounts)` in the for, `index` will be a number and it won't be necessary to make this cast.

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -430,7 +431,7 @@ export function getValidationErrors(config: any): string[] {
         if (accounts !== "remote") {
           errors.push(
             getPrivateKeyError(
-              accounts,
+              0,
```
fvictorio, line 434: I think I would use a custom error message here instead of using `getPrivateKeyError` with a 0 index.

packages/hardhat-core/src/internal/core/config/config-validation.ts
morgansliman: Oh nice

fvictorio (commented):
Very minor comments, other than that looks good!

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -53,6 +53,59 @@ function getErrorMessage(path: string, value: any, expectedType: string) {
   )} for ${path} - Expected a value of type ${expectedType}.`;
 }
 
+function getPrivateKeyError(index: any, network: string, message: string) {
```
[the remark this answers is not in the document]
morgansliman: umm great question.. oversight on my part, will fix

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -53,6 +53,59 @@ function getErrorMessage(path: string, value: any, expectedType: string) {
...
+  return `Invalid account: #${index} for network: ${network} - ${message}`;
+}
+
+function validatePrivateKey(
+  privateKey: any,
```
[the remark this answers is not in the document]
morgansliman: agreed, fixing

fvictorio (approved):

Issue #1981, opened 2021-10-22
Implement personal_sign in Hardhat Network


alcuadrado:
As mentioned in #1972, ethers now uses this method, so we should implement it now.

fvictorio:
Added in `hardhat@2.6.8`.

mds1:
It seems this is only supported for signers included in the default hardhat config, which makes sense. I have some fuzz tests that use ethers' `Wallet.createRandom()` to generate random wallets to sign with `personal_sign`, and impersonating them is of course not sufficient since the provider doesn't have the private key. 

I could increase the number of initial accounts or randomize the mnemonic/path used, but this feels a little clunky. One solution is to update `hardhat_impersonateAccount` to take a private key instead of an address, but curious if you had any other thoughts

mds1:
Worth noting that these tests were previously using `wallet.signMessage`, which it seems did not require a request to the provider, so there might be a workaround within ethers that I should look into also

fvictorio:
@mds1 My guess is that you are doing something like this, right?

```js
wallet = ethers.Wallet.createRandom()
await hre.network.provider.send("hardhat_impersonateAccount", [wallet.address])
signer = await ethers.getSigner(wallet.address)
signedMessage = await signer.signMessage("foo")
```

In that case, no, this won't work. But if you use `wallet.signMessage` instead, that will work because the wallet has access to the private key.

mds1:
Yep, that's what I'm doing and totally understand why it doesn't work!

Mainly asked about impersonating via private key or other ideas because I thought `signMessage` didn't add the message prefix, whereas `personal_sign` does, so they'd give different signatures. Though maybe the behavior of `signMessage` has changed since I last looked

fvictorio:
Hmm, I get the same result if I use the default signer (which uses `personal_sign`) and if I create a wallet with the private key of that signer and sign the same message.

mds1:
Ah ok, so I looked at the code and (1) the Hardhat implementation of `eth_sign`, (2) the Hardhat implementation of `personal_sign`, and (3) ethers' `signMessage`, all add the message prefix. (Not all `eth_sign` implementations add the prefix).

As a result, these three signing approaches are all equivalent with Hardhat + ethers

```typescript
// hardhat account 0 with default mnemonic and path path: "m/44'/60'/1'/0"
[...]

PR #2009, opened 2021-10-29
Add personal_sign RPC method

fvictorio:
Related to #1981.

### DIFF
--- a/.changeset/chilled-cooks-listen.md
+++ b/.changeset/chilled-cooks-listen.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Add `personal_sign` RPC method
+Add support for the `personal_sign RPC method to Hardhat Network.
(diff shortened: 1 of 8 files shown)


.changeset/chilled-cooks-listen.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Add `personal_sign` RPC method
```
alcuadrado, line 5:
```suggestion
Add support for the `personal_sign RPC method to Hardhat Network.
```

I think we should write them as we would for the release. wdyt?

alcuadrado (approved):
Just left a suggestion, but LGTM

.changeset/chilled-cooks-listen.md
fvictorio: Yes, makes sense!

PR #2012, opened 2021-11-01
Support Solidity 0.8.6

feuGeneA:


alcuadrado (approved):
I love it when these are so straightforward :)

PR #2016, opened 2021-11-02
PR to a PR: Ensure we always set a chainId along w/ a hardfork

feuGeneA:


alcuadrado (approved):

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2220,6 +2221,8 @@ Hardhat Network's forking functionality only works with blocks from at least spu
         this._vm._common.setHardfork(
           this._selectHardfork(blockContext!.header.number)
         );
+        originalChainId = this._vm._common.chainId();
+        this._vm._common.setChain(blockContext!._common.chainId());
```
alcuadrado, line 2225: I know I already approved this, but I think this is not correct.

The block context can be a forked block, and those have a a common that may have an incorrect chainId. I think we should use the chainId that we already fetched from the network when initializing the Node. I think networkId is used because of a bug.

PR #2017, opened 2021-11-02
Support Solidity 0.8.8

feuGeneA:
- [x] Ensure existing tests pass.
- [x] Add tests of new Language Feature, "User Defined Value Type: allows creating a zero cost abstraction over a value type with stricter type requirements."  Ensure that UDVT's can be used as params and return values.

### DIFF
--- a/.changeset/smooth-turkeys-grab.md
+++ b/.changeset/smooth-turkeys-grab.md
@@ -1,6 +1,5 @@
 ---
 "hardhat": patch
-"docs": patch
 ---
 
 Added support for Solidity 0.8.8
--- a/.changeset/smooth-turkeys-grab.md
+++ b/.changeset/smooth-turkeys-grab.md
@@ -1,0 +1,6 @@
+---
+"hardhat": patch
+"docs": patch
+---
+
+Added support for Solidity 0.8.8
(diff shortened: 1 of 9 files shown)


fvictorio (approved):

.changeset/smooth-turkeys-grab.md
```diff
@@ -0,0 +1,6 @@
+---
+"hardhat": patch
+"docs": patch
```
alcuadrado, line 3: I hate that the link adds docs autmatically. It breaks changesets.
```suggestion
```

fvictorio: Oops, didn't notice that.

packages/hardhat-core/test/internal/hardhat-network/stack-traces/test-files/0_8/user-defined-value-types/c.sol
```diff
@@ -0,0 +1,12 @@
+pragma solidity ^0.8.8;
+
+import "./../../../../../../../console.sol";
```
fvictorio, line 3: You can remove this now (and also the import in the `test.json`).

feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2017/commits/ba4bfc18efd3c0fba58a98b9821fd42db3ddd9f7

Issue #1549, opened 2021-06-15
Increase timeout for localhost network


fvictorio:
A lot of users are using the `localhost` network with a forked hardhat node. This frequently leads to timeouts for the first run. We should consider just increasing it to a larger value.

alcuadrado:
We should try to reproduce it, so that we have a better idea about which timeout to use. Now it's 20sec.

fvictorio:
We are going to increase the localhost and mocha timeouts to 40s.

zendevil:
Is there a way to manually increase the timeout? I'm getting timeouts running my tests.

fvictorio:
> Is there a way to manually increase the timeout? I'm getting timeouts running my tests.

```js
networks: {
  localhost: {
    url: "http://localhost:8545",
    timeout: 40000
  }
},
mocha: {
  timeout: 40000
}
```

feuGeneA:
Released in [hardhat v2.8.3](https://github.com/nomiclabs/hardhat/releases/tag/hardhat%402.8.3)

PR #2026, opened 2021-11-04
changed timeout from 20s to 40s in docs/config

Rishabhraghwendra18:
- [x] Because this PR includes a **documentation change**, it uses the `master` branch as its base branch.
----

changed timeout from 20s to 40s in docs/config
issue: #1549

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -37,7 +37,7 @@ module.exports = {
     artifacts: "./artifacts"
   },
   mocha: {
-    timeout: 20000
+    timeout: 40000
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


alcuadrado:
Hey @Rishabhraghwendra18, thanks for sending this PR. I think it only updates this value in the docs, and not the actual code. Doesn't it?

Rishabhraghwendra18:
Ya you are correct @alcuadrado . Actually I didn't find the appropriate file where I have to make this change in code . I thought , the code may get update and since docs refrence to 2s only so I thought of updating it

PR #2023, opened 2021-11-03
Avalanche verify support

marcelomorgado:
- [X] Because this PR includes a **new feature**, it uses the `development` branch as its base branch.
---

Support contracts verification for the snowtrace.io (Etherscan-like Avalanche explorer)

PR #2028, opened 2021-11-04
Avalanche & Avalanche Fuji

CyberSensei1:
Avalanche & Avalanche Fuji  snowtrace.io support added


fvictorio:
Hey @CyberSensei1, thanks for sending this, but there's already an open PR that adds support for snowtrace (#2023)

PR #2032, opened 2021-11-04
Support the mining of multiple blocks in a single call

### LINKED ISSUE #1112: Mine several blocks in a single call

feuGeneA:
Related to #1112 .

- [x] Draft initial interface stubs and internal data storage.
- [x] Take into account the number of blocks of pending transactions that were mined when deciding how many *more* blocks to mine. [Source 1](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r743917512), [source 2](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r743918057).
- [x] Stop saving empty block ranges in snapshot data. [Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r743918709).
- [x] Implement the mining of the final, valid block, the one whose parent is the final *invalid* block, in new method `HardhatNode.mineBlocks()`. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/1b2f05326f3d796c5c459cb05f4bbc125df718aa)
- [x] Implement the usage of the empty block ranges in `BlockchainData.getBlockByNumber()`. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/17d4897681b5ca3a484bcb1cf10d6e13dd87ce73)
- [x] Ensure that `eth_blockNumber` increments appropriately after `hardhat_mine` has run.
- [x] ~Ensure that the coinbase address reflects payments to the miner after `hardhat_mine` has run.~ [Edit: we decided against this.]
- [x] ~Ensure that the state root is correct after `hardhat_mine` has run.~ [Edit: we decided against this.]
- [x] Encode "Test 4" from [the spec](https://www.notion.so/nomiclabs/Mining-multiple-blocks-edad145bbce94d958d778b134588f9d6#90f895a9cb9a4eef92d64e47ac917a12).
- [x] Ensure that `hardhat_reset` continues to work after `hardhat_mine` has run. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/fe86d358f12809b03d2b7014bf84379536dc6878)
- [x] Add testing to ensure that snapshots still work. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/9850a492363a5bab9dcd3306bf6187ce5cd3ecb3)
- [x] Add testing of timestamps on lazily created blocks. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/0ecf846fb58d1f2e1227f6cd591283ea740ab88c)
- [x] Fix bug that Franco found, described in [this comment](https://github.com/nomiclabs/hardhat/pull/2032#issuecomment-995264386). (Bug demonstrated in https://github.com/nomiclabs/hardhat/pull/2032/commits/d5c44462f5abe025e56832ebf3be6cbc44f33a1d and fixed in https://github.com/nomiclabs/hardhat/pull/2032/commits/9dfbf083e78fc2124c41dec34aab27d70f74e45d)
- [x] Add documentation. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/3f0dee65d04a90584530f6d7e1ef47f44a0c23ba)
- [x] Fix (and test for) off-by-1 error in `BlockchainData.reserveBlocks()`. ([Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r779563093))
- [x] Consider whether `BlockchainData.unreserveBlocksAfter()` should check that `b !== oldReservation.first`, like we're doing in `fulfillBlockReservation` ([Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r779558688))
- [x] Clarify usage of placeholder block in reservation in `BlockchainData`, and ensure there's no off-by-1 error. ([Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r779240680))
- [x] Iron out bugs in `HardhatBlockchain.deleteLaterBlocks()`. ([Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r779233206))
- [x] Have `HardhatNode` pass its `this._vm._common` into `HardhatBlockchain` (and `ForkBlockchain`), so that it can pass it into `BlockchainData.fulfillBlockReservation()`, so that `HardhatNode` can remain ignorant of the whole "reservation" concept. ([Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r780447890)) Then, remove the reservation-related methods from `HardhatBlockchainInterface`. ([Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r779196385))
- [x] Update `deleteLaterBlocks()` on `ForkBlockchain` as was done for `HardhatBlockchain`. ([Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r779570591))
- [x] Refactor `HardhatBlockchain` and `ForkBlockchain` to extend a common base class implementation. (Done in https://github.com/nomiclabs/hardhat/pull/2032/commits/e27cf777fd160db2a4323217d495d07e26e874cd)
- [x] Consider the "capstone block" as a fulfilled reservation, not an unfulfilled one; and, ensure that the mined sequence of blocks is terminated by a fully valid block. ([Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r787055570))
- [x] Ensure that `setNextBlockTimestamp` continues to work. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/5559b8ecfb8b929f00166674d07fd31b92fa7c7f)
- [x] Ensure that historical calls work with blocks mined via `hardhat_mine`. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/e040073ff544cac0465e68b9459dbbb54604f409)
- [x] Ensure that, when using `hardhat_mine`, the requested timestamp interval is also applied to blocks mined normally from transactions in the mempool. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/09d6cf7a58e7e573c0d3d23fb5a92385e8ae3485)
- [x] In `BlockchainData._calculateTimestampForReservedBlock()`, assert block's existence rather than falling back to `0`. ([Source](https://github.com/nomiclabs/hardhat/pull/2032#discussion_r779556176)).
- [x] Document, in what makes these reserved blocks invalid, alongside the invalid state root and lack of credit to coinbase, that the `baseFeePerGas` will be incorrect. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/e7dc03f9aa4d6314f3765263cf27adb36df78a0a)
- [x] Ensure that the doc is clear on the fact that the interval only applies to blocks mined within a `hardhat_mine` call (not to blocks mined subsequent to that call). (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/4ed64df918a6178252a45bfa32059a585bb8ba69)
- [x] Explain `newBlockFilter` limitation in docs. (Completed in https://github.com/nomiclabs/hardhat/pull/2032/commits/758a7c37806535f99795322075d9133cea89d65e)


feuGeneA:
@alcuadrado Can you do a pre-review here and let me know if I'm going in the right direction?

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -462,6 +462,23 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
   }
 
+  public async mineEmptyBlocks(
+    blockCount: BN = new BN(1),
+    intervalInSeconds: BN = new BN(1)
+  ) {
+    // first mine any pending transactions:
+    while (this._txPool.hasPendingTransactions()) {
+      await this.mineBlock();
+    }
```
alcuadrado, line 472: I think this while should also has to take `blockCount` into account. Maybe something like:

```suggestion
    let mined = 0;
    while (blockCount.gtn(mined) && this._txPool.hasPendingTransactions()) {
      await this.mineBlock();
      mined += 1;
    }
```

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -462,6 +462,23 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+
+    const latestBlockNumber = await this.getLatestBlockNumber();
+    this._blockchain.addEmptyBlockRange({
+      first: latestBlockNumber,
+      last: latestBlockNumber.add(blockCount),
```
alcuadrado, line 477:
```suggestion
      last: latestBlockNumber.add(blockCount).subn(mined),
```

also should take the amount of mined blocks so far.

Also, if mined === blockCount this part should be skipped.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -870,6 +887,7 @@ Hardhat Network's forking functionality only works with blocks from at least spu
       irregularStatesByBlockNumber: this._irregularStatesByBlockNumber,
       userProvidedNextBlockBaseFeePerGas:
         this.getUserProvidedNextBlockBaseFeePerGas(),
+      emptyBlockRanges: this._blockchain.getEmptyBlockRanges(),
```
alcuadrado, line 890: I'm not really sure if this is necessary. Wouldn't the latestBlock be enough?

alcuadrado (commented):
The general approach looks good @feuGeneA. I just left a few minor comments

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/8a96e90c9c37326b3ed35ecb1433da4a13458128

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: > also should take the amount of mined blocks so far.

Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/4c6ad94fd0b602827f10565a120bbb58af4117e4

> Also, if mined === blockCount this part should be skipped.

Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/4ceab4a641349ad7581e2caff86b970b9ca89080

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/4c6ad94fd0b602827f10565a120bbb58af4117e4

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
feuGeneA: Missed some cleanup in that last commit.  Done here: https://github.com/nomiclabs/hardhat/pull/2032/commits/8134fb818354c42bda395b1b2d90d96c2d1c6976

fvictorio:
I just started testing it (I will test it more and review it tomorrow), but I found a pretty significant bug:

```
$ hh console
> (await network.provider.send("eth_getBlockByNumber", ["0x0", false])).timestamp
'0x61ba6bf1'
> await network.provider.send("hardhat_mine", ["0xa", "0x100"])
true
> await network.provider.send("hardhat_mine", ["0xa", "0x100"])
true
> await network.provider.send("eth_blockNumber")
'0x14'
> (await network.provider.send("eth_getBlockByNumber", ["0x13", false])).timestamp
'0x900'
```

It's not easy to reproduce: I think you need to call `hardhat_mine` _twice_, with an interval, and then fetch one of the blocks in the second batch. Curiously, the blocks in the first batch have the correct timestamp.

feuGeneA:
> I just started testing it (I will test it more and review it tomorrow), but I found a pretty significant bug:
> 
> ```
> $ hh console
> > (await network.provider.send("eth_getBlockByNumber", ["0x0", false])).timestamp
> '0x61ba6bf1'
> > await network.provider.send("hardhat_mine", ["0xa", "0x100"])
> true
> > await network.provider.send("hardhat_mine", ["0xa", "0x100"])
> true
> > await network.provider.send("eth_blockNumber")
> '0x14'
> > (await network.provider.send("eth_getBlockByNumber", ["0x13", false])).timestamp
> '0x900'
> ```
> 
> It's not easy to reproduce: I think you need to call `hardhat_mine` _twice_, with an interval, and then fetch one of the blocks in the second batch. Curiously, the blocks in the first batch have the correct timestamp.

@fvictorio Thanks for pointing this out. I demonstrated in https://github.com/nomiclabs/hardhat/pull/2032/commits/0ecf846fb58d1f2e1227f6cd591283ea740ab88c that this problem doesn't exist when only calling `hardhat_mine` once, and demonstrated in https://github.com/nomiclabs/hardhat/pull/2032/commits/d5c44462f5abe025e56832ebf3be6cbc44f33a1d that, as you described, there's a bug when calling it twice, and in https://github.com/nomiclabs/hardhat/pull/2032/commits/9dfbf083e78fc2124c41dec34aab27d70f74e45d I fixed the bug.

packages/hardhat-core/src/internal/hardhat-network/provider/HardhatBlockchain.ts
```diff
@@ -62,9 +95,15 @@ export class HardhatBlockchain implements HardhatBlockchainInterface {
     if (actual === undefined) {
       throw new Error("Invalid block");
     }
-    const nextBlock = this._data.getBlockByNumber(
-      new BN(actual.header.number).addn(1)
-    );
+
+    const nextBlockNumber = new BN(actual.header.number).addn(1);
```
fvictorio, line 99: `deleteLaterBlocks` is updated in `HardhatBlockchain` but not in `ForkBlockchain`. Is that intentional?

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -795,6 +795,69 @@ describe("HardhatNode", () => {
...
+      await node.mineBlocks(new BN(10));
+      const latestBlockNumber = node.getLatestBlockNumber();
+      assert.equal(
+        latestBlockNumber.toString(),
+        previousLatestBlockNumber.addn(10).toString()
+      );
+    });
+
+    describe("shouldn't break snapshotting", async function () {
+      it("when doing mineBlocks() before a snapshot", async function () {});
```
fvictorio, line 820: Forgot to remove/fill this one?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -189,6 +189,315 @@ describe("Hardhat module", function () {
...
+        });
+
+        it("should permit the mining of a regular block afterwards", async () => {
+          await this.ctx.provider.send("hardhat_mine", [
+            numberToRpcQuantity(3),
+          ]);
+          await this.ctx.provider.send("evm_mine");
+        });
+
+        describe("should premit the retrieval of a reserved block", function () {
```
fvictorio, line 216:
```suggestion
        describe("should permit the retrieval of a reserved block", function () {
```

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -189,6 +189,315 @@ describe("Hardhat module", function () {
...
+              await getLatestBlockNumber(),
+              latestBlockNumber + 1000
+            );
+          });
+        });
+
+        describe("should reflect timestamps properly", function () {
+          it("with only one hardhat_mine invocation", async () => {
+            const originalLatestBlockNumber = await getLatestBlockNumber();
+            const timestampBefore = rpcQuantityToNumber(
```
fvictorio, line 292: Maybe it's worth creating a `getBlockTimestamp` helper to make this test more readable.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -585,11 +607,15 @@ Hardhat Network's forking functionality only works with blocks from at least spu
   }
 
   public async getLatestBlock(): Promise<Block> {
+    if (this._blockchain.isReservedBlock("latest")) {
```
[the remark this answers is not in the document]
feuGeneA: The reason I wrote it this way is because the `Common` needs to be passed into the creation of the block, via ethereumjs' `Block.fromBlockData()`, and the node is the only level of abstraction that already has a common handy.

I can't remember exactly *why* we need to pass the common into `fromBlockData()`, but I have a vague memory of it relating to having the proper hardfork set on the block, and if you comment out that argument from the `fromBlockData()` call, there will be some test failures with errors like `calcNextBaseFee() can only be called with EIP1559 being activated`.

The only way I can see to take your suggestion would be to store a copy of the common within the `HardhatBlockchain` class, but that seems counter-intuitive to me, but I'm open to suggestions!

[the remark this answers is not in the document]
feuGeneA: I talked to Pato about this. Keeping the Common as member data within the Blockchain classes feels weirder but also safer and less error prone, so we're going to opt for that approach.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/b557db173f5a4ef56205fbfa7a5562ecf52222fe

packages/hardhat-core/src/internal/hardhat-network/provider/BlockchainData.ts
```diff
@@ -14,6 +19,27 @@ export class BlockchainData {
   private _transactions: Map<string, TypedTransaction> = new Map();
   private _transactionReceipts: Map<string, RpcReceiptOutput> = new Map();
   private _totalDifficulty: Map<string, BN> = new Map();
+  public blockReservations: Array<{
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/dfda4342cadd440327a981d10e3f0cedb0bbc4ca

packages/hardhat-core/src/internal/hardhat-network/provider/HardhatBlockchain.ts
```diff
@@ -24,11 +29,12 @@ export class HardhatBlockchain implements HardhatBlockchainInterface {
...
-      return this._data.getBlockByNumber(new BN(blockHashOrNumber)) ?? null;
-    }
-    if (BN.isBN(blockHashOrNumber)) {
-      return this._data.getBlockByNumber(blockHashOrNumber) ?? null;
+    if (typeof blockHashOrNumber === "number" || BN.isBN(blockHashOrNumber)) {
+      const blockNumber =
+        typeof blockHashOrNumber === "number"
+          ? new BN(blockHashOrNumber)
+          : blockHashOrNumber;
+      return this._data.getBlockByNumber(blockNumber) ?? null;
```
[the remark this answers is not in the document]
feuGeneA: Haha, thank you for catching this.  In a previous draft I had made functional changes to this method, and the change you're seeing here seemed necessary alongside the additional functional changes, but now that I've removed those additional functional changes this stuff should go too.  I've removed it in https://github.com/nomiclabs/hardhat/pull/2032/commits/feae806cbc9f27c09533d0a08657f6cf15ed66a3

packages/hardhat-core/src/internal/hardhat-network/provider/BlockchainData.ts
```diff
@@ -14,6 +19,27 @@ export class BlockchainData {
...
+    first: BN;
+    last: BN;
+    interval: BN;
+  }> = new Array();
+
+  public reserveBlocks(first: BN, count: BN, interval: BN, common: Common) {
+    const last = first.add(count);
```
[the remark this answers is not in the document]
feuGeneA: Bug demonstrated (test added) in https://github.com/nomiclabs/hardhat/pull/2032/commits/26aec19836b7b972a3fee06fca1e414f883b0c8a

packages/hardhat-core/src/internal/hardhat-network/provider/BlockchainData.ts
```diff
@@ -110,4 +136,119 @@ export class BlockchainData {
...
+    ) {
+      // delete the existing reservation, but re-add its first half:
+
+      const reservationIndex = this._findBlockReservation(b);
+
+      const oldReservation = this.blockReservations[reservationIndex];
+
+      this.blockReservations.splice(reservationIndex, 1);
+
+      this.blockReservations.push({
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/c0e7b14fc9701a396b9a206fabd9d3f2a6552d0e

packages/hardhat-core/src/internal/hardhat-network/provider/BlockchainData.ts
```diff
@@ -110,4 +136,119 @@ export class BlockchainData {
...
+
+  public fulfillBlockReservation(blockNumber: BN, common: Common): Block {
+    // number should lie within one of the reservations listed in
+    // this.blockReservations. in addition to adding the given block, that
+    // reservation needs to be split in two in order to accomodate access to
+    // the given block.
+
+    const reservationIndex = this._findBlockReservation(blockNumber);
+    if (reservationIndex === -1) {
+      throw new HardhatError(ERRORS.GENERAL.ASSERTION_ERROR, {
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/1f3c210cea9f517c75d50a7e0d06a83e667ac473

packages/hardhat-core/src/internal/hardhat-network/provider/HardhatBlockchain.ts
```diff
@@ -62,9 +95,15 @@ export class HardhatBlockchain implements HardhatBlockchainInterface {
...
       throw new Error("Invalid block");
     }
-    const nextBlock = this._data.getBlockByNumber(
-      new BN(actual.header.number).addn(1)
-    );
+
+    const nextBlockNumber = new BN(actual.header.number).addn(1);
+
+    if (this._data.isReservedBlock(nextBlockNumber)) {
+      this._data.unreserveBlocksAfter(nextBlockNumber);
```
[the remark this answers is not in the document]
feuGeneA: I found that the problem you describe also exists when doing `hardhat_mine` followed by a regular tx (not just when doing `hardhat_mine` twice). I wrote a test of that scenario in https://github.com/nomiclabs/hardhat/pull/2032/commits/721df4b54d25b1c1511ce4cf56ef3c4dee8dbfe3 (which you can see fails as expected in the CI run for that commit), and then got that test to pass (and did some significant cleanup of the "delete blocks" code) in https://github.com/nomiclabs/hardhat/pull/2032/commits/5c07c74448b68e9d03c9d5159f48d536a39652bf.  Then, just to confirm, I also added a test of your specific reproduction scenario in https://github.com/nomiclabs/hardhat/pull/2032/commits/52f16735c75750e008181ec5523934a0b287a7c2

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/51e2d3eccb262e2e56a6d6842f610dd606a17c15

packages/hardhat-core/src/internal/hardhat-network/provider/types/HardhatBlockchainInterface.ts
```diff
@@ -1,15 +1,21 @@
...
 
 import { FilterParams } from "../node-types";
 import { RpcLogOutput, RpcReceiptOutput } from "../output";
 
 export interface HardhatBlockchainInterface extends BlockchainInterface {
+  addBlock(block: Block): Promise<Block>;
   addTransactionReceipts(receipts: RpcReceiptOutput[]): void;
+  reserveBlocks(count: BN, interval: BN, common: Common): void;
+  isReservedBlock(numberOrLatest: BN | "latest"): boolean;
+  fulfillBlockReservation(numberOrLatest: BN | "latest", common: Common): Block;
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/51e2d3eccb262e2e56a6d6842f610dd606a17c15

packages/hardhat-core/src/internal/hardhat-network/provider/HardhatBlockchain.ts
feuGeneA: No, not intentional, thank you for pointing this out.  I updated `ForkBlockchain` in https://github.com/nomiclabs/hardhat/pull/2032/commits/29992fce6ff842d6b51d71ed3013471fdade222b, but then also significantly re-worked both in https://github.com/nomiclabs/hardhat/pull/2032/commits/e27cf777fd160db2a4323217d495d07e26e874cd

packages/hardhat-core/src/internal/hardhat-network/provider/BlockchainData.ts
```diff
@@ -14,6 +19,27 @@ export class BlockchainData {
...
+  public blockReservations: Array<{
+    first: BN;
+    last: BN;
+    interval: BN;
+  }> = new Array();
+
+  public reserveBlocks(first: BN, count: BN, interval: BN, common: Common) {
+    const last = first.add(count);
+    this.blockReservations.push({ first, last, interval });
+    this.addBlock(
```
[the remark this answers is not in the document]
feuGeneA: This is a confusing topic. Thank you for raising these questions.

First off, you were right about block 4 in your example; I've addressed that off-by-1 error in https://github.com/nomiclabs/hardhat/pull/2032/commits/37346fc9f8016d3dcd6ca5d6fe06d59077c18c7f.

You're also right that it's nonsensical to have that "capstone block" be replaced upon "fulfillment."  I'm going to change it so that block is not considered "reserved," so that it will not be replaced upon access.

Furthermore, after discussing with Pato, we decided that the sequence of blocks mined via `hardhat_mine` will consist of:
* regular block(s) comprised of any pending transactions
* one or more "reserved" blocks
* a "fulfilled reservation" block
* a regular (fully valid) block

The change from the previous model is the addition of the regular block at the end.  Under this new model, the final block in the mined sequence will be fully valid.  And, whereas before we were saying that you needed to request at least 2 blocks in order to induce a "reservation" (assuming that there are no pending transactions), now we're going to be saying that you need to request at least 3 blocks in order to induce a reservation.

I'll add this to the checklist: Consider the "capstone block" as a fulfilled reservation, not an unfulfilled one; and, ensure that the mined sequence of blocks is terminated by a fully valid block.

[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/88507135d25937d7d8a96488e7eff951cb5427ad

packages/hardhat-core/src/internal/hardhat-network/provider/BlockchainData.ts
```diff
@@ -110,4 +136,119 @@ export class BlockchainData {
...
+    }
+
+    const reservation = this.blockReservations[reservationIndex];
+
+    const blockNumberBeforeReservation = reservation.first.subn(1);
+
+    const previousTimestamp = this.isReservedBlock(blockNumberBeforeReservation)
+      ? this._calculateTimestampForReservedBlock(blockNumberBeforeReservation)
+      : this.getBlockByNumber(blockNumberBeforeReservation)?.header.timestamp ??
+        new BN(0);
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/nomiclabs/hardhat/pull/2032/commits/fb1342dbafab3fb53294359e776446ad7c799d01

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -817,7 +817,31 @@ describe("HardhatNode", () => {
...
+        await node.sendTransaction(
+          createTestTransaction({
+            from: DEFAULT_ACCOUNTS_ADDRESSES[1],
+            to: Address.fromString(
+              "0x1111111111111111111111111111111111111111"
+            ),
+            gasLimit: 21000,
+          })
+        );
+        await node.mineBlocks(new BN(1));
```
feuGeneA, line 835: Is this line necessary? Shouldn't it be auto-mining? Even if it is necessary, it feels weird to have it here, since you could just call `mineBlock()` instead, and since `mineBlocks(1)` literally just calls `mineBlock()` for you, and since seeing `mineBlocks()` within the test kinda suggests that it's essential to the test (it's not...right?), since `mineBlocks()` is the thing under test.

fvictorio: I think this `node` instance is created with `automine: false` at the top-level `describe`.

feuGeneA:
@fvictorio all your commits to date LGTM! Thank you for them!

.changeset/green-turkeys-fry.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": minor
+---
+
+Support the mining of multiple blocks in a single call (#1112)
```
alcuadrado, line 5:
```suggestion
Add a new RPC method, `hardhat_mine`, which supports the mining of multiple blocks in a single call (#1112)
```

docs/hardhat-network/reference/README.md
```diff
@@ -390,6 +390,14 @@ Call [`hardhat_stopImpersonatingAccount`](#hardhat-stopimpersonatingaccount) to
 
 Returns `true` if automatic mining is enabled, and `false` otherwise. See [Mining Modes](../explanation/mining-modes.md) to learn more.
 
+#### `hardhat_mine`
+
+Sometimes you may want to advance the latest block number of the Hardhat Network by a large number of blocks. One way to do this would be to call the `evm_mine` RPC method multiple times, but this is too slow if you want to mine thousands of blocks. The `hardhat_mine` method can mine any number of blocks at once, in constant time. (It exhibits the same performance no matter how many blocks are mined.)
+
+`hardhat_mine` accepts two parameters, both of which are optional. The first parameter is the number of blocks to mine, and defaults to 1. The second parameter is the interval between the timestamps of each block, _in seconds_, and it also defaults to 1. (The interval is applied only to blocks mined in the given method invocation, not to blocks mined afterwards.)
+
+Note that blocks mined in this way are technically not valid blocks. Specifically, they have an invalid parent hash and state root hash, the coinbase account will not have been credited with block rewards, and the `baseFeePerGas` will be incorrect.
```
alcuadrado, line 399: I think we should make it explicit that the latest block is valid.

docs/hardhat-network/reference/README.md
```diff
@@ -390,6 +390,24 @@ Call [`hardhat_stopImpersonatingAccount`](#hardhat-stopimpersonatingaccount) to
...
+// mine 256 blocks
+await hre.network.provider.send("hardhat_mine", ["0x100"]);
+
+// mine 1000 blocks with an interval of 1 minute
+await hre.network.provider.send("hardhat_mine", ["0x3e8", "0x3c"]);
+```
+
+Note that blocks mined via this method (except for the final one) are technically not valid blocks. Specifically, they have an invalid parent hash and state root hash, the coinbase account will not have been credited with block rewards, and the `baseFeePerGas` will be incorrect. (The final block in a sequence produced by `hardhat_mine` will always be fully valid.)
+
+Also note that blocks created via `hardhat_mine` will not trigger any new-block events. This includes filters created via `eth_newBlockFilter`, as well as WebSocket subscriptions to new-block events.
```
alcuadrado, line 409: I don't think this is correct. Some of them will trigger them, right @fvictorio ?

docs/hardhat-network/reference/README.md
```diff
@@ -390,6 +390,24 @@ Call [`hardhat_stopImpersonatingAccount`](#hardhat-stopimpersonatingaccount) to
...
+
+```js
+// mine 256 blocks
+await hre.network.provider.send("hardhat_mine", ["0x100"]);
+
+// mine 1000 blocks with an interval of 1 minute
+await hre.network.provider.send("hardhat_mine", ["0x3e8", "0x3c"]);
+```
+
+Note that blocks mined via this method (except for the final one) are technically not valid blocks. Specifically, they have an invalid parent hash and state root hash, the coinbase account will not have been credited with block rewards, and the `baseFeePerGas` will be incorrect. (The final block in a sequence produced by `hardhat_mine` will always be fully valid.)
```
alcuadrado, line 407:
```suggestion
Note that some empty blocks mined via this method (except for the final one) are technically not valid blocks. Specifically, they have an invalid parent hash, and the coinbase account will not have been credited with block rewards, and the `baseFeePerGas` will be incorrect. (The final block in a sequence produced by `hardhat_mine` will always be fully valid.)
```

docs/hardhat-network/reference/README.md
alcuadrado, line 409:
```suggestion
Also note that empty blocks created via `hardhat_mine` may not trigger any new-block events. This includes filters created via `eth_newBlockFilter`, as well as WebSocket subscriptions to new-block events.
```

alcuadrado: Some of them trigger the events, right @fvictorio ?

alcuadrado (approved):
I left two documentation comments, but pre-approving this anyway.

🚀

PR #2038, opened 2021-11-05
Print warning when user tries to use solc remapping

### LINKED ISSUE #1877: Print a warning when a user tries to use a solc remapping
Hardhat doesn't support solc remappings. 
We should print a warning if we detect that a user is trying to use them via their config.

morgansliman:


packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -232,6 +237,31 @@ Learn more about compiler configuration at https://hardhat.org/config"
...
 
+function checkUnsupportedRemappings(userConfig: any) {
+  if (typeof userConfig.solidity === "object") {
+    let remappings: boolean;
+    if ("compilers" in userConfig.solidity) {
+      remappings =
+        userConfig.solidity.compilers.filter(
+          ({ settings }: SolcUserConfig) => settings?.remappings !== undefined
+        ).length > 0;
+    } else {
```
alcuadrado, line 248: I think there's another format here, which is `solidity: "x.x.x"`

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -232,6 +237,31 @@ Learn more about compiler configuration at https://hardhat.org/config"
   }
 }
 
+function checkUnsupportedRemappings(userConfig: any) {
```
alcuadrado, line 240: Maybe we can use HardhatUserConfig here, as it was already validated

packages/hardhat-core/src/internal/core/config/config-loading.ts
morgansliman: there is, but it's covered by the initial `if (typeof userConfig.solidity === 'object') {...}` check

packages/hardhat-core/src/internal/core/config/config-loading.ts
morgansliman: I decided not to simply so that this function remains in parity with the other function it's being called with (`checkMissingSolidityConfig`), but I'm happy to change that

alcuadrado: Makes sense

packages/hardhat-core/src/internal/core/config/config-loading.ts
alcuadrado: Oops, i didn't noticed that. you are right

packages/hardhat-core/test/fixture-projects/solidity-config-warnings/remapping-in-list.js
```diff
@@ -0,0 +1,15 @@
...
+  solidity: {
+    compilers: [
+      {
+        version: "0.7.0",
+        settings: {
+          remappings: [":g=/dir"],
+        },
+      },
+    ],
+    overrides: {
```
alcuadrado, line 11: oh, i almost forgot this, but the overrides can also have remappings

morgansliman: ahh good catch

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -77,6 +81,7 @@ export function loadConfigAndTasks(
 
   if (showSolidityConfigWarnings) {
     checkMissingSolidityConfig(userConfig);
+    checkUnsupportedRemappings(userConfig);
```
fvictorio, line 84: @alcuadrado Would it be a bit better if this was done in the next `if (showSolidityConfigWarnings` (when the config is already resolved)? It might be a stretch, but maybe some plugin adds it or something. Or do you prefer to just check what users write?

fvictorio: Plus, that should simplify the logic of the check function a lot.

fvictorio (approved):

PR #2041, opened 2021-11-05
Adding solution to https://hardhat.org/tutorial/debugging-with-hardhat-network.html

architec:
- [x] Because this PR includes a **documentation change**, it uses the `master` branch as its base branch.

---

Great tutorial, please close this PR if you think it will not be helpful. 

This PR is for adding a solution to https://hardhat.org/tutorial/debugging-with-hardhat-network.html.

When I was doing this tutorial, I feel having a solution at the end of each step would be beneficial because I would like to compare my progress with the solution.

### DIFF
--- a/docs/tutorial/debugging-with-hardhat-network.md
+++ b/docs/tutorial/debugging-with-hardhat-network.md
@@ -59,3 +59,6 @@ Trying to send 100 tokens to 0xe5904695748fe4a84b40b3fc79de2277660bd1d3
 ```
 
 Check out the [documentation](/hardhat-network/README.md#console-log) to learn more about this feature.
+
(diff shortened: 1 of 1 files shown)


architec:
Tutorial Step 7: https://github.com/architec/hardhat-tutorial/tree/step-7

fvictorio:
@architec what do you mean by "solution" here? Having a repo with the result of doing each step?

It could make sense to have a repo with the full result of doing the tutorial, and maybe with branches/tags for each step. But if we do that, we'll have to do it under our organization (mainly for security reasons), so it's unlikely that we can merge this PR :slightly_frowning_face:

architec:
no problem. I will close this PR.

PR #2057, opened 2021-11-10
Use 127.0.0.1 as default hostname

alcuadrado:
Banteg originally reported that with the new version of Node.js he couldn't connect to `hardhat node`. The reason is that now node resolves `"localhost"` to an ipv6, so we only listen on ipv6 by default.

This PR fixes that by making it explicit that we want the ipv4 version of `localhost` (i.e. `127.0.0.1`)

### DIFF
--- a/.changeset/selfish-falcons-decide.md
+++ b/.changeset/selfish-falcons-decide.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix an issue with new versions of Node.js that prevented clients from connecting to Hardhat's node using 127.0.0.1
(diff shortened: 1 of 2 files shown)


fvictorio:
Should we update the default `localhost` network too?

alcuadrado:
> Should we update the default `localhost` network too?

it's already `127.0.0.1`. See: 
https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/core/config/default-config.ts#L17

alcuadrado:
@fvictorio can you review/approve this?

fvictorio (approved):

PR #2059, opened 2021-11-11
In `_runTxAndRevertMutations`, run in brand new VM

feuGeneA:
Attempts to address review comment https://github.com/nomiclabs/hardhat/pull/1928#discussion_r746935880

But why isn't it working? Is it because I'm not passing enough stuff into the Common constructor's first argument? makeCommon puts all kinds of extra stuff in there. Though, makeForkCommon doesn't put much in there, and the fork tests are failing too.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -135,6 +135,8 @@ export class HardhatNode extends EventEmitter {
       allowUnlimitedContractSize,
       tracingConfig,
       minGasPrice,
+      networkId,
+      chainId,
     } = config;
 
     let common: Common;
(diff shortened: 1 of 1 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2207,12 +2212,22 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
 
-      originalHardfork = this._vm._common.hardfork();
-      this._vm._common.setHardfork(
-        this._selectHardfork(blockContext!.header.number)
-      );
+      const vm = new VM({
+        common: new Common({
+          chain: {
+            // eslint-disable-next-line @typescript-eslint/dot-notation
+            ...Common["_getChainParams"]("mainnet"),
```
alcuadrado, line 2219:
```suggestion
            this._vm._common["_chainParams"]
```

PR #2083, opened 2021-11-19
update typescript and prettier versions

morgansliman:
- updated typescript `"~4.0.3"` -> `"~4.5.2"`
- updated prettier `"2.3.2"` -> `"~2.4.1"`


.changeset/dull-apples-film.md
```diff
@@ -0,0 +1,20 @@
...
+"@nomiclabs/hardhat-solpp": patch
+"@nomiclabs/hardhat-truffle4": patch
+"@nomiclabs/hardhat-truffle5": patch
+"@nomiclabs/hardhat-vyper": patch
+"@nomiclabs/hardhat-waffle": patch
+"@nomiclabs/hardhat-web3-legacy": patch
+"@nomiclabs/hardhat-web3": patch
+---
+
+update typescript and prettier versions
```
alcuadrado, line 20: I think it's better if we don't use a changeset here, as it's not a "releasable" change

alcuadrado:
Note: the reason there are many compilation errors now is that TS introduces breaking changes in minor releases. That's why we use `~`

morgansliman:
Yeah, I've fixed most of them so far (the vast majority are just derived
from `error` in catch clauses defaulting to `unknown` now as opposed to
`any`.

There is one breaking change I'm not sure about the best way to fix,
though. The change is typescript removed `Reflect.enumerate` since it's
deprecated now in Javascript everywhere. Admittedly, I don't have much
experience with using `Reflect` or `Proxy`, so I'm really not sure what an
appropriate fix would be.

alcuadrado:
> There is one breaking change I'm not sure about the best way to fix, though. The change is typescript removed `Reflect.enumerate` since it's deprecated now in Javascript everywhere. Admittedly, I don't have much experience with using `Reflect` or `Proxy`, so I'm really not sure what an appropriate fix would be.

Ouch, this sounds problematic. Let's discuss it next week.

packages/e2e/package.json
```diff
@@ -22,7 +22,7 @@
     "@types/chai": "^4.2.0",
     "@types/fs-extra": "^5.1.0",
     "@types/mocha": "^5.2.6",
-    "@types/node": "^10.17.24",
+    "@types/node": "^16.11.9",
```
alcuadrado, line 25: We normally keep this aligned with the oldest version of node that we support, which is 12.x

packages/hardhat-core/src/internal/core/params/argumentTypes.ts
```diff
@@ -185,14 +185,19 @@ export const inputFile: CLIArgumentType<string> = {
...
+          {
+            name: argName,
+            value: strValue,
+          },
+          error
+        );
+      }
+
+      // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
+      throw error;
```
alcuadrado, line 200: Maybe we can create a helper for this pattern.

Something like

```ts
handleError(error, (e:Error) => {
   // do things here if error is an instance of Error
});
```

alcuadrado (commented):
I left some minor comments, but this looks good

docs/package.json
```diff
@@ -15,11 +15,11 @@
     "dotenv": "^6.2.0",
     "exports-loader": "^1.1.1",
     "imports-loader": "^1.2.0",
-    "prettier": "2.3.2",
+    "prettier": "~2.4.1",
```
fvictorio, line 18: The official recommendation is to use a pinned version of prettier. This is probably fine, but I'd rather err on the side of being cautious because when this causes a problem, it's a weird and hard to identify problem (e.g., the CI saying something that cannot be reproduced locally).

packages/hardhat-core/src/internal/core/params/argumentTypes.ts
morgansliman: I know @fvictorio has mentioned to me that our error system could use a refactor in the future; maybe we wait and add the helper into that refactor? 

Just a thought. I'm fine either way.

morgansliman: Just adding my thoughts here for posterity's sake:

The original reason I handled the errors this way in this PR was because I don't know most of the areas of the code base the errors touched, so leaving it as close to current functionality was what I aimed for. In the future, either in new code or an error system refactor, we can aim to properly narrow the error type so we don't have to have eslint disables everywhere.

alcuadrado (approved):

packages/hardhat-etherscan/test/unit/solc/bytecode.ts
```diff
@@ -79,7 +79,12 @@ describe("Compiler bytecode and deployed bytecode matching", () => {
         moreData: 42,
         mysteriousString: "mystery",
       };
-      const mockMetadata = Buffer.from(encode(mockSolcMetadataMapping));
+      const encoded = encode(mockSolcMetadataMapping);
+      const mockMetadata = Buffer.from(
+        encoded.buffer,
+        encoded.byteOffset,
+        encoded.byteLength
+      );
```
fvictorio, line 87: I'm curious about this change, what does it mean?

fvictorio (approved):
Pre-approving because comments are really minor

packages/hardhat-core/src/internal/cli/bootstrap.ts
```diff
@@ -34,5 +34,5 @@ const childProcess = fork(`${__dirname}/cli`, process.argv.slice(2), {
 });
 
 childProcess.once("close", (status) => {
-  process.exit(status);
+  process.exit(status as number);
```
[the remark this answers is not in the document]
morgansliman: Ah, as of Node 15, `status` is typed as `number | null`, so when I updated `@types/node` to 16, I added this change. I just didn't realize it was a node 15 thing, so I didn't think to revert this change when I downgraded `@types/node` back to 12

packages/hardhat-etherscan/test/unit/solc/bytecode.ts
morgansliman: Functionally the new code is equivalent to the old code, but I believe this was due to a subtle change in how the `Buffer.from` function types its arguments that caused the object returned from `encode` to not be directly passable to `Buffer.from` anymore. This workaround just utilizes buffer's other two optional arguments that were already provided by `encode` luckily.

PR #2089, opened 2021-11-20
Document the FIFO mempool mode

alcuadrado:
This PR documents the changes introduce by #1559

### DIFF
--- a/docs/hardhat-network/explanation/mining-modes.md
+++ b/docs/hardhat-network/explanation/mining-modes.md
@@ -4,7 +4,7 @@ Hardhat Network can be configured to **automine** blocks, immediately upon recei
 
 You can use one of these modes, both or neither. By default, only the automine mode is enabled.
 
-When automine is disabled, every sent transaction is added to the mempool, which contains all the transactions that could be mined in the future. By default, Hardhat Network's mempool follows the same rules as Geth. This means, among other things, that transactions are prioritized by fees paid to the miner (and then by arrival time), and that invalid transactions are dropped. An alternative FIFO mempool is also available.
+When automine is disabled, every sent transaction is added to the mempool, which contains all the transactions that could be mined in the future. By default, Hardhat Network's mempool follows the same rules as Geth. This means, among other things, that transactions are prioritized by fees paid to the miner (and then by arrival time), and that invalid transactions are dropped. In addition to the default mempool behavior, an alternative FIFO behavior is also available.
 
 When automine is disabled, pending transactions can be queried via the `eth_getBlockByNumber` RPC method (with `"pending"` as the block number argument), they can be removed using the `hardhat_dropTransaction` RPC method, and they can be replaced by submitting a new transaction with the same nonce but with a 10+% increase in fees paid to the miner.
 
--- a/docs/hardhat-network/explanation/mining-modes.md
+++ b/docs/hardhat-network/explanation/mining-modes.md
@@ -4,7 +4,7 @@ Hardhat Network can be configured to **automine** blocks, immediately upon recei
 
 You can use one of these modes, both or neither. By default, only the automine mode is enabled.
 
-When automine is disabled, every sent transaction is added to the mempool, which contains all the transactions that could be mined in the future. By default, Hardhat Network's mempool follows the same rules as Geth. This means, among other things, that transactions are prioritized by fees paid to the miner (and then by arrival time), and that invalid transactions are dropped. An alternative FIFO mempool is also available. 
+When automine is disabled, every sent transaction is added to the mempool, which contains all the transactions that could be mined in the future. By default, Hardhat Network's mempool follows the same rules as Geth. This means, among other things, that transactions are prioritized by fees paid to the miner (and then by arrival time), and that invalid transactions are dropped. An alternative FIFO mempool is also available.
 
 When automine is disabled, pending transactions can be queried via the `eth_getBlockByNumber` RPC method (with `"pending"` as the block number argument), they can be removed using the `hardhat_dropTransaction` RPC method, and they can be replaced by submitting a new transaction with the same nonce but with a 10+% increase in fees paid to the miner.
 
@@ -38,7 +38,7 @@ networks: {
   hardhat: {
     mining: {
       mempool: {
-        order: "fifo"
+        order: "fifo";
       }
     }
   }
--- a/docs/hardhat-network/explanation/mining-modes.md
+++ b/docs/hardhat-network/explanation/mining-modes.md
@@ -29,16 +29,16 @@ const pendingBlock = await network.provider.send("eth_getBlockByNumber", [
 
 ### Mining transactions in FIFO order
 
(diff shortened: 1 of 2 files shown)


docs/hardhat-network/explanation/mining-modes.md
```diff
@@ -4,13 +4,15 @@ Hardhat Network can be configured to **automine** blocks, immediately upon recei
...
 
-When automine is disabled, every sent transaction is added to the mempool, which contains all the transactions that could be mined in the future. Hardhat Network's mempool follows the same rules as geth. This means, among other things, that transactions are prioritized by fees paid to the miner (and then by arrival time), and that invalid transactions are dropped. Pending transactions can be queried via the `eth_getBlockByNumber` RPC method (with `"pending"` as the block number argument), they can be removed using the `hardhat_dropTransaction` RPC method, and they can be replaced by submitting a new transaction with the same nonce but with a 10+% increase in fees paid to the miner.
+When automine is disabled, every sent transaction is added to the mempool, which contains all the transactions that could be mined in the future. By default, Hardhat Network's mempool follows the same rules as Geth. This means, among other things, that transactions are prioritized by fees paid to the miner (and then by arrival time), and that invalid transactions are dropped. An alternative FIFO mempool is also available.
```
alcuadrado, line 7: I split this paragraph into two because I felt that it was too long after my change.

docs/hardhat-network/reference/README.md
```diff
@@ -109,6 +109,28 @@ networks: {
 
 This means that no new blocks will be mined by the Hardhat Network, but you can manually mine new blocks using the `evm_mine` RPC method. This will generate a new block that will include as many pending transactions as possible.
 
+#### Transaction ordering
+
+Hardhat Network can sort mempool transactions in two different ways. How they are sorted will alter which transactions from the mempool get included in the next block, and in which order.
```
alcuadrado, line 114: I don't like that here i'm using "sorting" all the time, and then the setting is "order". 

Any rewording addressing this will be welcome :)

docs/hardhat-network/reference/README.md
```diff
@@ -124,7 +124,7 @@ networks: {
   hardhat: {
     mining: {
       mempool: {
-        order: "fifo"
+        order: "fifo";
```
alcuadrado, line 127: This `;` is actually incorrect. `prettier` added it.

feuGeneA: This has bitten me before too! I had an external contributor step in and remove the `;` in some docs I had written. Slightly embarrassing haha.

One option would be to add `<!-- prettier-ignore -->` to the line before ` ```js`, which i found mentioned [here](https://prettier.io/blog/2017/11/07/1.8.0.html).

A seemingly better solution would be to use ` ```json` rather than ` ```js`, which I tried out locally and seems to work as expected.

docs/hardhat-network/explanation/mining-modes.md
```diff
@@ -4,13 +4,15 @@ Hardhat Network can be configured to **automine** blocks, immediately upon recei
...
```
feuGeneA, line 7:
```suggestion
When automine is disabled, every sent transaction is added to the mempool, which contains all the transactions that could be mined in the future. By default, Hardhat Network's mempool follows the same rules as Geth. This means, among other things, that transactions are prioritized by fees paid to the miner (and then by arrival time), and that invalid transactions are dropped. In addition to the default mempool behavior, an alternative FIFO behavior is also available.
```

docs/hardhat-network/explanation/mining-modes.md
```diff
@@ -25,6 +27,24 @@ const pendingBlock = await network.provider.send("eth_getBlockByNumber", [
 ]);
 ```
 
+### Mining transactions in FIFO order
+
+The way Hardhat Network's mempool orders transactions is customizable. By default, they are prioritized following Geth's rules, but you can enable a FIFO mode instead. This mode ensures that transactions are added to blocks in the same order they are sent, and is useful to recreate blocks from other networks.
```
feuGeneA, line 32: Maybe I'm picking too small of a nit here, but I'd prefer to leave the word "mode" out of this, since we're already formalizing that word with the "Mining Modes" article title.  I'd prefer to keep using "behavior" here, for the sakes of consistency and differentiation.

```suggestion
The way Hardhat Network's mempool orders transactions is customizable. By default, they are prioritized following Geth's rules, but you can enable a FIFO behavior instead, which ensures that transactions are added to blocks in the same order they are sent, and which is useful to recreate blocks from other networks.
```

docs/hardhat-network/explanation/mining-modes.md
```diff
@@ -25,6 +27,24 @@ const pendingBlock = await network.provider.send("eth_getBlockByNumber", [
...
+
+You can enable the FIFO mode in your config with:
```
feuGeneA, line 34:
```suggestion
You can enable the FIFO behavior in your config with:
```

docs/hardhat-network/explanation/mining-modes.md
```diff
@@ -25,6 +27,24 @@ const pendingBlock = await network.provider.send("eth_getBlockByNumber", [
...
+
+```js
+networks: {
+  hardhat: {
+    mining: {
+      mempool: {
+        order: "fifo";
```
feuGeneA, line 41:
```suggestion
```json
networks: {
  hardhat: {
    mining: {
      mempool: {
        order: "fifo"
```

docs/hardhat-network/reference/README.md
feuGeneA: I think this paragraph in particular is just fine, but I hear what you're saying and I'm offering a few tweaks to the subsequent paragraphs.

docs/hardhat-network/reference/README.md
```diff
@@ -109,6 +109,28 @@ networks: {
...
 
+#### Transaction ordering
+
+Hardhat Network can sort mempool transactions in two different ways. How they are sorted will alter which transactions from the mempool get included in the next block, and in which order.
+
+The first sorting mode, called `"priority"`, mimics Geth's behavior. This means that it prioritizes transactions based on the fees paid to the miner. This is the default sorting mode.
+
+The second sorted mode, called `"fifo"`, keeps the mempool transactions sorted in the order they arrive.
+
+You can change the sorting mode with:
```
feuGeneA, line 120:
```suggestion
The first ordering mode, called `"priority"`, mimics Geth's behavior. This means that it prioritizes transactions based on the fees paid to the miner. This is the default.

The second ordering mode, called `"fifo"`, keeps the mempool transactions sorted in the order they arrive.

You can change the ordering mode with:
```

feuGeneA (changes requested):

docs/hardhat-network/explanation/mining-modes.md
fvictorio: This should have a link to the proper section in the reference.

alcuadrado:
I've already addressed every comment, @feuGeneA @fvictorio

feuGeneA (approved):

fvictorio (approved):

Issue #1918, opened 2021-10-01
Make coinbase address configurable


JosephInGithub:
Hello,

It would be nice if you could add the possibility to change the miner address like in geth with miner.setEtherbase.

Best regards,

Joseph

fvictorio:
Released in `hardhat@2.7.0`. Check the [release notes](https://github.com/nomiclabs/hardhat/releases/tag/hardhat%402.7.0) to learn more.

PR #2090, opened 2021-11-20
Configurable coinbase

alcuadrado:
This PR addresses #1918 by adding a new `coinbase` field to the Hardhat Network config and introducing a new `hardhat_setCoinbase` RPC method.

### DIFF
--- a/packages/hardhat-core/src/types/config.ts
+++ b/packages/hardhat-core/src/types/config.ts
@@ -44,6 +44,7 @@ export interface HardhatNetworkUserConfig {
   initialDate?: string;
   loggingEnabled?: boolean;
   forking?: HardhatNetworkForkingUserConfig;
+  coinbase?: string;
 }
 
 export type HardhatNetworkAccountsUserConfig =
@@ -120,6 +121,7 @@ export interface HardhatNetworkConfig {
   initialDate: string;
   loggingEnabled: boolean;
   forking?: HardhatNetworkForkingConfig;
+  coinbase?: string;
 }
 
 export type HardhatNetworkAccountsConfig =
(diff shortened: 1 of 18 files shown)


packages/hardhat-core/src/types/config.ts
```diff
@@ -120,6 +121,7 @@ export interface HardhatNetworkConfig {
   initialDate: string;
   loggingEnabled: boolean;
   forking?: HardhatNetworkForkingConfig;
+  coinbase?: string;
```
feuGeneA, line 124: Why optional here? I guess I'm still trying to wrap my head around why some fields here should be optional any some shouldn't.  I thought after the #1666 work we wanted to minimize the optionality on this type.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/index.ts
```diff
@@ -43,10 +43,10 @@ describe("Eth module", function () {
       });
 
       describe("eth_coinbase", async function () {
-        it("should return the the hardcoded coinbase address", async function () {
+        it("should return the the default coinbase address", async function () {
```
feuGeneA, line 46:
```suggestion
        it("should return the default coinbase address", async function () {
```

feuGeneA (approved):

packages/hardhat-core/src/types/config.ts
alcuadrado: This is something I struggled a bit with too. I think that to keep backward compatibility new fields to existing types need to be optional. I kind of hate it, but otherwise existing code may stop working.

PR #2099, opened 2021-11-26
change: remove 0x from private key in tutorial

dagumak:
During the deployment stage, Hardhat will throw an error saying `Invalid account: #0 for network: ropsten - private key too long, expected 32 bytes`.  I removed the `0x` string concatenation to make it easier for other people.


fvictorio:
I think that happened because the env variable already had a `0x`. Anyway, since we accept PKs both with and without the 0x prefix, I think this change makes sense.

dagumak:
> I think that happened because the env variable already had a `0x`. Anyway, since we accept PKs both with and without the 0x prefix, I think this change makes sense.

Thank you for the merge!

Issue #1466, opened 2021-05-19
Should have recover logic dealing with cached `list.json` on compiler download interruption


AmagiDDmxh:
### Step to reproduce
1. run `npx hardhat run scripts/deploy.js --network localhost` which would install new compiler
2. In the middle, interrupt it by press `control` + `c`
3. run `npx hardhat run scripts/deploy.js --network localhost` which would install new compiler again
4. Error prompts
```bash
An unexpected error occurred:

SyntaxError: /Users/.../Library/Caches/hardhat-nodejs/compilers/macosx-amd64/list.json: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at /Users/.../hardhat-hackathon-boilerplate/node_modules/jsonfile/index.js:33:18
    at /Users/.../hardhat-hackathon-boilerplate/node_modules/graceful-fs/graceful-fs.js:123:16
```

During the interruption, write stream on the `list.json` simply turn off instead of recover the previous file, I guess.

alcuadrado:
Thanks for reporting this issue, @AmagiDDmxh! 

I think the problem is that the interruption came in the middle of the `list.json` download. We should download it to a temporal file instead, and then move it, as it's a shorter (if not atomic) operation.

In the meantime, run `npx hardhat clean --global`.

kanej:
I have been able to reproduce the error based on the assumption that the `list.json` file is only partially downloaded:

1. create a new hardhat project based on the basic template
2. Start `npx hardhat node`
3. Run the sample deploy script `npx hardhat run ./scripts/sample-script.js --network localhost` triggering a compile (and compiler download)
4. Edit the `/Users/<user-name>/Library/Caches/hardhat-nodejs/compilers/macosx-amd64/list.json` file to be a partial and invalid json file, so deleting half of it for instance
5. Edit the `contracts/Greeter.sol` to force recompile on next deploy
6. Rerun the sample deploy script `npx hardhat run ./scripts/sample-script.js --network localhost`

That gives the expected error:

```shell
An unexpected error occurred:

SyntaxError: /Users/<user-name>/Library/Caches/hardhat-nodejs/compilers/macosx-amd64/list.json: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at /Users/<user-name>/projects/sandbox/another/node_modules/jsonfile/index.js:33:18
    at /Users/<user-name>/projects/sandbox/another/node_modules/graceful-fs/graceful-fs.js:123:16
[...]

PR #2102, opened 2021-11-26
Fix recovering from malformed compiler list

kanej:
This PR is an attempt to deal with #1466 

A halt during compiler download can lead the compiler list file to be partially written (and so invalid json). On a subsequent run the parsing of the compiler list fails and leaves the hardhat install in a broken state (requiring a `npx hardhat clean --global`).

This PR changes the compiler downloader behaviour to parse the file and on `SyntaxError`, redownload and retry the parse once more. Other errors that are not syntax based are rethrown.

## Manual testing

Additional tests have been added to the test suite, but if you are attempting to reproduce the bug manually:

1. create a new hardhat project based on the basic template
2. Start `npx hardhat node`
3. Run the sample deploy script `npx hardhat run ./scripts/sample-script.js --network localhost` triggering a compile (and compiler download)
4. Edit the `/Users/<user-name>/Library/Caches/hardhat-nodejs/compilers/macosx-amd64/list.json` file to be a partial and invalid json file, so deleting half of it for instance
5. Edit the `contracts/Greeter.sol` to force recompile on next deploy
6. Rerun the sample deploy script `npx hardhat run ./scripts/sample-script.js --network localhost`

That gives the expected error:

```shell
An unexpected error occurred:

SyntaxError: /Users/<user-name>/Library/Caches/hardhat-nodejs/compilers/macosx-amd64/list.json: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at /Users/<user-name>/projects/sandbox/another/node_modules/jsonfile/index.js:33:18
    at /Users/<user-name>/projects/sandbox/another/node_modules/graceful-fs/graceful-fs.js:123:16
    at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:71:3)
```

With the fixes in this branch the second run of the deploy script will succeed, as it overwrites the compilers list json file on detecting it if malformed.

### DIFF
--- a/.changeset/strange-rocks-poke.md
+++ b/.changeset/strange-rocks-poke.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix a bug that prevented Hardhat from working if the compilers list was partially downloaded (issue #1466)
--- a/.changeset/wise-nails-lick.md
+++ b/.changeset/wise-nails-lick.md
@@ -2,4 +2,4 @@
 "@nomiclabs/hardhat-ethers": patch
 ---
 
-Adds a custom formatter to better display BigNumber's in Hardhat console (issue #2109).
+Adds a custom formatter to better display BigNumber's in Hardhat console (issue #2109)
--- a/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
+++ b/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
@@ -155,13 +155,26 @@ export class CompilerDownloader {
   }
 
   public async getCompilersList(
-    platform: CompilerPlatform
+    platform: CompilerPlatform,
+    pendingRetries: number = 3
   ): Promise<CompilersList> {
     if (!(await this.compilersListExists(platform))) {
       await this.downloadCompilersList(platform);
     }
 
-    return fsExtra.readJson(this.getCompilersListPath(platform));
+    try {
+      return await fsExtra.readJSON(this.getCompilersListPath(platform));
+    } catch (error) {
+      // if parsing throws a syntax error, redownload and parse once more
+      if (!(error instanceof SyntaxError) || pendingRetries === 0) {
+        // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
+        throw error;
+      }
+
+      // remove the malformed list and retry
+      await fsExtra.remove(this.getCompilersListPath(platform));
+      return this.getCompilersList(platform, pendingRetries - 1);
+    }
   }
 
   public async getCompilerBuild(version: string): Promise<CompilerBuild> {
--- a/packages/hardhat-core/src/internal/util/download.ts
+++ b/packages/hardhat-core/src/internal/util/download.ts
@@ -9,6 +9,19 @@ interface FetchOptions {
   agent?: undefined | HttpsProxyAgent.HttpsProxyAgent;
 }
 
+const TEMP_FILE_PREFIX = "tmp-";
+
+function resolveTempFileName(filePath: string): string {
+  const { root, dir, ext, name } = path.parse(filePath);
+
+  return path.format({
+    root,
+    dir,
+    ext,
+    name: `${TEMP_FILE_PREFIX}${name}`,
+  });
+}
+
 export async function download(
   url: string,
   filePath: string,
@@ -43,8 +56,12 @@ export async function download(
   const response = await fetch(url, fetchOptions);
 
   if (response.ok && response.body !== null) {
+    const tmpFilePath = resolveTempFileName(filePath);
+
     await fsExtra.ensureDir(path.dirname(filePath));
-    return streamPipeline(response.body, fs.createWriteStream(filePath));
+    await streamPipeline(response.body, fs.createWriteStream(tmpFilePath));
+
+    return fsExtra.move(tmpFilePath, filePath);
   }
 
   // Consume the response stream and discard its result
--- a/packages/hardhat-core/src/internal/util/download.ts
+++ b/packages/hardhat-core/src/internal/util/download.ts
@@ -12,10 +12,9 @@ interface FetchOptions {
 const TEMP_FILE_PREFIX = "tmp-";
 
 function resolveTempFileName(filePath: string): string {
-  const { root, dir, ext, name } = path.parse(filePath);
+  const { dir, ext, name } = path.parse(filePath);
 
   return path.format({
-    root,
     dir,
     ext,
     name: `${TEMP_FILE_PREFIX}${name}`,
--- a/packages/hardhat-core/test/internal/solidity/compiler/downloader.ts
+++ b/packages/hardhat-core/test/internal/solidity/compiler/downloader.ts
@@ -196,22 +196,36 @@ describe("Compiler downloader", function () {
 
   describe("Get compilers lists and CompilerBuild", function () {
     let compilersDir: string;
-    let downloadCalled: boolean;
+    let downloadCallCount: number;
+    let successDownloadTry: number;
     let mockDownloader: CompilerDownloader;
 
     beforeEach(async function () {
       compilersDir = this.tmpDir;
 
-      downloadCalled = false;
+      downloadCallCount = 0;
+      successDownloadTry = 0;
 
       mockDownloader = new CompilerDownloader(compilersDir, {
         download: async () => {
-          downloadCalled = true;
-          await saveMockCompilersList();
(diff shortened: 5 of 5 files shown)


.changeset/strange-rocks-poke.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
```
kanej, line 2: Should this be `@hardhat-core`? @fvictorio

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
```diff
@@ -159,7 +159,19 @@ export class CompilerDownloader {
       await this.downloadCompilersList(platform);
     }
 
-    return fsExtra.readJson(this.getCompilersListPath(platform));
+    try {
+      return await fsExtra.readJSON(this.getCompilersListPath(platform));
+    } catch (error) {
+      // if parsing throws a syntax error, redownload and parse once more
+      if (!(error instanceof SyntaxError)) {
+        // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
```
kanej, line 167: My intention here was to touch nothing unless it is explicitly the parsing error, so rethrowing ... but that does mean running against the eslint rule.

.changeset/strange-rocks-poke.md
alcuadrado: No, `hardhat` is fine. `packages/hardhat-core` is named like that to make talking about `hardhat` (the npm package) simpler

.changeset/strange-rocks-poke.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Fix - allow recovering from partially downloaded compiler list
```
alcuadrado, line 5: These messages end up in the changelong and other user-facing places, so we don't use any kind of structured message, but rather a plain English explanation. Something like 

> Fix a bug that prevented Hardhat from working if the compilers list was partially downloaded

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
alcuadrado: I like this change!

I also think that we should implement the log to download to a tmp file and move once ready, as a general helper, because we download other things that are harder to validate. For example, the compiler's binaries.

kanej: I have modified `packages/hardhat-core/src/internal/util/download.ts` to download to a temp file in the same directory are the intended final filepath so `compilers/list.json` will initially be downloaded to `compilers/tmp-list.json`; on successful download the file is then moved into the intended final filepath.

I can only find two usages of download - download the compiler list and download an individual compiler. So I have made the tmp file change directly in `download.ts` but I can change that to be a separate function `download-via-tmp` or whatever and apply it on compilers list initially. What is your preference?

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
```diff
@@ -159,7 +159,19 @@ export class CompilerDownloader {
...
+        throw error;
+      }
+
+      await this.downloadCompilersList(platform);
+
+      return await fsExtra.readJSON(this.getCompilersListPath(platform));
```
fvictorio, line 173: One alternative here is to do just `return getCompilersList(platform)`. That feels better to me, but if we do that we should also include a `retryLimit` variable.

fvictorio: Something like this:

```typescript
  public async getCompilersList(
    platform: CompilerPlatform,
    pendingRetries = 3
  ): Promise<CompilersList> {
    if (!(await this.compilersListExists(platform))) {
      await this.downloadCompilersList(platform);
    }

    try {
      return await fsExtra.readJSON(this.getCompilersListPath(platform));
    } catch (error) {
      // if parsing throws a syntax error, redownload and parse once more
      if (!(error instanceof SyntaxError) || pendingRetries === 0) {
        // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
        throw error;
      }

      // remove the malformed list and retry
      await fsExtra.remove(this.getCompilersListPath(platform));
      return this.getCompilersList(platform, pendingRetries - 1);
    }
  }
```

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
alcuadrado: I think this makes sense

packages/hardhat-core/test/internal/solidity/compiler/downloader.ts
```diff
@@ -316,6 +316,98 @@ describe("Compiler downloader", function () {
...
+              downloadCount++;
+            },
+            forceSolcJs: true,
+          });
+
+          try {
+            await failingDownloader.getCompilersList(CompilerPlatform.WASM);
+          } catch (error) {
+            assert.isTrue(error instanceof SyntaxError);
+          }
```
fvictorio, line 359: This doesn't quite work because this doesn't check that there is actually an error. You can use `isRejected` for this, see for example:

https://github.com/nomiclabs/hardhat/blob/f911aa3b76a61835a1a83740065370fc6ab79183/packages/hardhat-core/test/internal/core/providers/accounts.ts#L412-L418

packages/hardhat-core/test/internal/solidity/compiler/downloader.ts
```diff
@@ -316,6 +316,98 @@ describe("Compiler downloader", function () {
         });
       });
     });
+
+    describe("When downloading fails", function () {
+      describe("getCompilersList", function () {
+        it("retries on a failed download", async function () {
+          let downloadCount = 0;
+          const failingDownloader = (mockDownloader = new CompilerDownloader(
```
fvictorio, line 324: This feels a bit weird. Can't you just assign to `mockDownloader` and use that?

packages/hardhat-core/test/internal/solidity/compiler/downloader.ts
kanej: Thanks for that. I couldn't find the async throws assertion on `chai`, I hadn't grasped we had `chai-as-promised`; and `isRejected` actually works rather than my homebrew.

packages/hardhat-core/test/internal/solidity/compiler/downloader.ts
kanej: I have moved the more complicated download function up, and used it across the tests in this block.

.changeset/wise-nails-lick.md
```diff
@@ -2,4 +2,4 @@
 "@nomiclabs/hardhat-ethers": patch
 ---
 
-Adds a custom formatter to better display BigNumber's in Hardhat console (issue #2109).
+Adds a custom formatter to better display BigNumber's in Hardhat console (issue #2109)
```
fvictorio, line 5: This doesn't belong to this PR, does it?

(If this just means removing a period from the changeset for consistency or something, I don't mind. Just pointing it out in case it means something strange happened during the rebase.)

fvictorio (approved):
Pre-approving, just one very minor comment.

.changeset/wise-nails-lick.md
kanej: You are correct, this shouldn't be here. I added in this changeset, wondered whether the convention was adding a fullstop, checked the other changesets and found one with a comma and one without and make both of mine consistent with the third.

kanej: The changeset has already been released as it turns out.

packages/hardhat-core/src/internal/util/download.ts
```diff
@@ -9,6 +9,19 @@ interface FetchOptions {
   agent?: undefined | HttpsProxyAgent.HttpsProxyAgent;
 }
 
+const TEMP_FILE_PREFIX = "tmp-";
+
+function resolveTempFileName(filePath: string): string {
+  const { root, dir, ext, name } = path.parse(filePath);
+
+  return path.format({
+    root,
```
fvictorio, line 18: Small comment I missed the previous time I reviewed this:

According to [the docs](https://nodejs.org/api/path.html#pathformatpathobject), «pathObject.root is ignored if pathObject.dir is provided», so I think we can remove `root` from here and from the result of `path.parse`.

alcuadrado (approved):

packages/hardhat-core/src/internal/util/download.ts
kanej: Fix applied

Issue #2109, opened 2021-11-28
Add functionality to prettify BigNumber


mrthankyou:
## What's The Problem
One of the most frustrating issues with BigNumber is that I can't prettify a BigNumber into a readable format. For example, if I was using Hardhat console to print out a BigNumber, I would get `999999999999999999292` as string format. This is difficult to read.

## Proposed Solution
If Hardhat extended BN.js with an additional function to format a BigNumber into a readable format, it would make it easier for developers to read these BigNumbers.

For example, the following code shows (I'm sure there are many ways to do this) how to convert this BigNumber to a readable format:

```javascript
bigNumber.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",")

// bigNumber as a string:                  999999999999999999292 
// bigNumber using the format above:       999,999,999,999,999,999,292

```

fvictorio:
This is a good idea. The way BigNumbers are shown is certainly a big point of friction.

I think the right approach would be to leverage the inspect symbol. This [has been suggested](https://github.com/ethers-io/ethers.js/issues/1955) for ethers.js but rejected. Maybe we can do it ourselves in `hardhat-ethers`? @alcuadrado wdyt?

fvictorio:
Fixed in [`@nomiclabs/hardhat-ethers@2.0.3`](https://github.com/nomiclabs/hardhat/releases/tag/%40nomiclabs%2Fhardhat-ethers%402.0.3).

mrthankyou:
I totally forgot about this issue. I'm so happy I see you guys fixed this. Thank you so much @fvictorio and team!

0xthk:
> Fixed in [`@nomiclabs/hardhat-ethers@2.0.3`](https://github.com/nomiclabs/hardhat/releases/tag/%40nomiclabs%2Fhardhat-ethers%402.0.3).

Hey, looks like this issue comes back after upgrade hardhat to `v2.10.0`. If I print the big number from contract call, it won't apply the formatter.

<img width="381" alt="image" src="https://user-images.githubusercontent.com/1924968/188864735-e0914fc1-803c-41d9-ac83-a48991ae9a6f.png">

fvictorio:
@0xthk which exact versions of hardhat, @nomiclabs/hardhat-ethers and ethers are you using?

0xthk:
> @0xthk which exact versions of hardhat, @nomiclabs/hardhat-ethers and ethers are you using?

Here is the version in our package.json

[...]

PR #2114, opened 2021-11-30
Add custom formatter for BigNumber's in console

kanej:
Extends the BigNumber prototype provided by `hardhat-ethers` to include a formatter for node inspection tools. Large BigNumbers that are inspected in hardhat console will now display with commas to help readability.

The formatter uses `,` for readability in large BigNumbers:

```
1
1,000
999,999,999
```

Relates to #2109.

## Preview

image

## Manual Testing

To test manually, build the `hardhat-ethers` plugin locally and link (see `CONTRIBUTION.md`) the plugin to a local hardhat project. From that project run hardhat console (i.e. `hh console`), you should then be able to see the newly formatted BigNumbers:

```shell
> ethers.BigNumber.from(999999999999789)
999,999,999,999,789
```

### DIFF
--- a/.changeset/wise-nails-lick.md
+++ b/.changeset/wise-nails-lick.md
@@ -1,0 +1,5 @@
+---
+"@nomiclabs/hardhat-ethers": patch
+---
+
+Adds a custom formatter to better display BigNumber's in Hardhat console (issue #2109).
--- a/packages/hardhat-ethers/test/index.ts
+++ b/packages/hardhat-ethers/test/index.ts
@@ -2,6 +2,7 @@ import { assert } from "chai";
 import { ethers } from "ethers";
 import { NomicLabsHardhatPluginError } from "hardhat/plugins";
 import { Artifact } from "hardhat/types";
+import util from "util";
 
 import { EthersProviderWrapper } from "../src/internal/ethers-provider-wrapper";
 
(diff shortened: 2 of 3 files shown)


kanej:
Are we happy with this format (i.e. `999,999`). In the ethers playground an explicit reference is put to the value being a `BigNumber`, so:

```
BigNumber.from("-0x2a")
// { BigNumber: "-42" }
```

I kept the format the same as the original issue, but it can be changed.

fvictorio:
Yeah, I think we should be more explicit to avoid confusion, but I don't like the `{ BigNumber: "-42" }` thing because it suggests it's an object with a `BigNumber` property.

What about:

```
> ethers.BigNumber.from(999999999999789)
BigNumber: 999,999,999,999,789
```

cc @alcuadrado @morgansliman @feuGeneA wdyt?

.changeset/wise-nails-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomiclabs/hardhat-ethers": patch
+---
+
+Adds a custom formatter to better display BigNumber's in HardHat console (issue #2109).
```
fvictorio, line 5: :stuck_out_tongue: 

```suggestion
Adds a custom formatter to better display BigNumber's in Hardhat console (issue #2109).
```

feuGeneA:
> Yeah, I think we should be more explicit to avoid confusion, but I don't like the `{ BigNumber: "-42" }` thing because it suggests it's an object with a `BigNumber` property.
> 
> What about:
> 
> ```
> > ethers.BigNumber.from(999999999999789)
> BigNumber: 999,999,999,999,789
> ```
> 
> cc @alcuadrado @morgansliman @feuGeneA wdyt?

No strong opinions for me. I don't have much context here. (I had to look up both `Symbol` and `nodejs.util.inspect.custom`. :see_no_evil: ) But I wanted to point out that my view of the ethers playground doesn't exhibit that weirdness with `BigNumber` looking like a property. It came out like this:

```node
> new BN(11)
BigNumber { value: "11" }
> BigNumber.from(11)
BigNumber { value: "11" }
> BigNumber.from("-0x2a")
BigNumber { value: "-42" }
```

morgansliman:
> Yeah, I think we should be more explicit to avoid confusion, but I don't like the `{ BigNumber: "-42" }` thing because it suggests it's an object with a `BigNumber` property.
> 
> What about:
> 
> ```
> > ethers.BigNumber.from(999999999999789)
> BigNumber: 999,999,999,999,789
> ```
> 
> cc @alcuadrado @morgansliman @feuGeneA wdyt?

No strong opinions from me on this either. I do agree that the end result shouldn't display like BigNumber is a property, but it seems like that's also not a consistent issue (RE: Gene)? 

My initial reaction is that this comma formatter should only be implemented into an explicit `toString` method, rather than the default internal behavior, but I'm fine either way really. @fvictorio

alcuadrado:
> > Yeah, I think we should be more explicit to avoid confusion, but I don't like the `{ BigNumber: "-42" }` thing because it suggests it's an object with a `BigNumber` property.
> > What about:
> > ```
> > > ethers.BigNumber.from(999999999999789)
> > BigNumber: 999,999,999,999,789
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > cc @alcuadrado @morgansliman @feuGeneA wdyt?
> 
> No strong opinions for me. I don't have much context here. (I had to look up both `Symbol` and `nodejs.util.inspect.custom`. 🙈 ) But I wanted to point out that my view of the ethers playground doesn't exhibit that weirdness with `BigNumber` looking like a property. It came out like this:
> 
> ```js
> > new BN(11)
> BigNumber { value: "11" }
> > BigNumber.from(11)
> BigNumber { value: "11" }
> > BigNumber.from("-0x2a")
> BigNumber { value: "-42" }
> ```

I like that this alternative doesn't look like js syntax, so it would be less confusing, and that it's what ethers does.

packages/hardhat-ethers/test/index.ts
```diff
@@ -21,6 +21,54 @@ describe("Ethers plugin", function () {
...
+
+      describe("Custom formatters", function () {
+        const customSymbol = Symbol.for("nodejs.util.inspect.custom");
+
+        const assertBigNumberFormat = function (
+          BigNumber: any,
+          value: string | number,
+          expected: string
+        ) {
+          assert.equal(BigNumber.from(value)[customSymbol](), expected);
```
alcuadrado, line 33: I think I'd do something like `util.format("%o", value)`, as that's what the console does.

alcuadrado:
> My initial reaction is that this comma formatter should only be implemented into an explicit `toString` method, rather than the default internal behavior, but I'm fine either way really. @fvictorio

This is an interesting comment. I kind of like the commas because they are not valid js syntax. The reason I like that property is that avoiding unexperienced users from copying these strings into a js file/script is a plus to me, as they probably won't notice that they are losing precision because of turning the `BigNumber`s into `number`

fvictorio:
Ok, let's do the same that the ethers playground does. It seems good enough: it tells you that it's a BigNumber, it hints at the fact that there's some object involved, and the value shown is a string so (hopefully) people won't copy and paste it and use it as a number.

We don't need backward compatibility here, so we can always improve this later if it turns out to be necessary.

packages/hardhat-ethers/test/index.ts
kanej: Nice, that is far better.

kanej:
> Ok, let's do the same thing that the ethers playground does. It seems good enough: it tells you that it's a BigNumber, it hints at the fact that there's some object involved, and the value shown is a string so (hopefully) people won't copy and paste it and use it as a number.
> 
> We don't need backward compatibility here, so we can always improve this later if it turns out to be necessary.

I have pushed the format change to match ethers playground, so:

```shell
Welcome to Node.js v14.18.1.
Type ".help" for more information.
> ethers.BigNumber.from(1)
BigNumber { value: "1" }
> ethers.BigNumber.from(1000)
BigNumber { value: "1000" }
> ethers.BigNumber.from(9999999999342)
BigNumber { value: "9999999999342" }
> ethers.BigNumber.from(-1000)
BigNumber { value: "-1000" }
```

I have removed the comma formatting (as that is what playground does), it is of course easy to add it back if we would prefer:

```shell
> ethers.BigNumber.from(9999999999342)
BigNumber { value: "9,999,999,999,342" }
```

alcuadrado (approved):
🚀

Issue #1716, opened 2021-07-22
[hardhat-ethers] `getContractFactory` and `getContractAt` should have an overload that take an `Artifact`


alcuadrado:
_This was reported by @nventuro_

If you have your own artifacts-loading logic, and have a contract that needs to be linked, our plugin doesn't help you. You'd have to link the libraries manually.

We should have overloads that take an `Artifact`, and take care of the library linking.

`getContractAt` doesn't really need this overload, but it would be great to be consistent.

Rishabhraghwendra18:
is this issue still open ?

alcuadrado:
Yes, do you want to implement it?

Rishabhraghwendra18:
Ya... I want to give it a try since it's a good first issue . So I think I can do this

alcuadrado:
This is a good first issue for you, @Rishabhraghwendra18

kanej:
This will get closed when the release goes out.

fvictorio:
Done in [`@nomiclabs/hardhat-ethers@2.0.4`](https://github.com/nomiclabs/hardhat/releases/tag/%40nomiclabs%2Fhardhat-ethers%402.0.4).

PR #2116, opened 2021-11-30
Add artifact functions in ethers `getContractFactoryFromArtifact` and `getContractAtFromArtifact`

kanej:
In `hardhat-ethers`, allow passing of artifacts (resolved via custom code) into an equivalent of the `getContractFactory` call, this is instead of resolving by contract name.

The intention here is to support custom contract artifact resolution but still allow the leveraging of `hardhat-ethers`.

Two new functions have been added to support this without altering the existing signature:

* getContractFactoryFromArtifact
* getContractAtFromArtifact

Relates to #1716

## Example Usage

```typescript
import { ethers } from "hardhat";
import { Artifacts } from "hardhat/internal/artifacts";
import { Artifact } from "hardhat/types";

export async function getArtifact(contract: string): Promise<Artifact> {
  const artifactsPath: string = "./artifacts";

  const artifacts = new Artifacts(artifactsPath);
  return artifacts.readArtifact(contract);
}

async function main() {
  const artifact = await getArtifact("Greeter");

  const Factory = await ethers.getContractFactoryFromArtifact(artifact);

  const greeter = await Factory.deploy("Hello, Hardhat!");

  await greeter.deployed();

  console.log("Greeter deployed to:", greeter.address);

  const response = await greeter.greet();

  console.log("First Response", response);

  const alternateGreeter = await ethers.getContractAtFromArtifact(
    artifact,
    greeter.address
  );

  const secondResponse = await alternateGreeter.greet();

  console.log("Second Response", secondResponse);
}

// We recommend this pattern to be able to use async/await everywhere
// and properly handle errors.
main().catch((error) => {
  console.error(error);
  process.exitCode = 1;
});
```

## TODO

* [x] support artifact in signature of `getContractFactoryFromArtifact`
* [x] support artifact in signature of `getContractAtFromArtifact`
* [x] resolve type issue in consuming lib
* [x] determine and update any docs affected

### DIFF
--- a/.changeset/smooth-worms-end.md
+++ b/.changeset/smooth-worms-end.md
@@ -1,0 +1,5 @@
+---
+"@nomiclabs/hardhat-ethers": patch
+---
+
+Add equivalents in hardhat-ethers for `getContractFactory` and `getContractAt` that support passing `Artifact`, specifically `getContractFactoryFromArtifact` and `getContractAtFromArtifact` (issue #1716)
--- a/packages/hardhat-ethers/README.md
+++ b/packages/hardhat-ethers/README.md
@@ -60,8 +60,11 @@ function getContractFactory(name: string, signer: ethers.Signer): Promise<ethers
 
 function getContractFactory(name: string, factoryOptions: FactoryOptions): Promise<ethers.ContractFactory>;
 
+function getContractFactoryFromArtifact(artifact: Artifact, signer?: ethers.Signer): Promise<ethers.ContractFactory>;
 
 function getContractAt(nameOrAbi: string | any[], address: string, signer?: ethers.Signer): Promise<ethers.Contract>;
+
+function getContractAtFromArtifact(artifact: Artifact, address: string, signer?: ethers.Signer): Promise<ethers.Contract>;
 
 function getSigners() => Promise<ethers.Signer[]>;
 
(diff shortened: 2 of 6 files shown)


.changeset/smooth-worms-end.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomiclabs/hardhat-ethers": patch
```
kanej, line 2: Given that this is changing the type signatures (and so can break existing code) should this be `minor`?

kanej: We have opted to add functions rather alter existing signatures. Hence this can be reduced to minor update.

kanej:
I have test locally with newly created `advanced typescript project`, and we no longer get the type clash with typechain overriding the method signatures - getContractFactoryFromArtifact, getContractAtFromArtifact

packages/hardhat-ethers/README.md
```diff
@@ -60,9 +60,12 @@ function getContractFactory(name: string, signer: ethers.Signer): Promise<ethers
 
 function getContractFactory(name: string, factoryOptions: FactoryOptions): Promise<ethers.ContractFactory>;
 
+function getContractFactoryFromArtifact(artifact: Artifact, signer?: ethers.Signer): Promise<ethers.ContractFactory>;
```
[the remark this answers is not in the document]
kanej: Good catch, it should be both. I have added in all the combinations of signatures for the current ethers interface (we had left out some functions/signature combinations) - so it now consistent with how the readme show the `getContractFactory` function.

fvictorio (approved):

PR #2118, opened 2021-12-01
Update CONTRIBUTING.md

fvictorio:
The current explanation about the `master` and `development` branches is misleading; it's not clear that most of the time you just want to use the `master` branch (or, in other words, if you are reading the contributing doc you probably need to use that branch).


alcuadrado (approved):
lgtm

alcuadrado (approved):

PR #2122, opened 2021-12-02
Add xDai and Sokol networks

nikitosing:

---

Add support for xDai and Sokol networks.


kanej (changes requested):
Given we control the base url and the extension I think the url joiner function is fine as an approach.

But I would be tempted to put a test to stop someone coming along and going - oh, you should just use new URL() rather than some bespoke thing ...

kanej (approved):

Issue #1448, opened 2021-05-09
Extending hardhat-etherscan to support multiple API keys


wschwab:
## Extending `hardhat-etherscan` to support multiple API keys

**Summary:**

The [`hardhat-etherscan` plugin](https://hardhat.org/plugins/nomiclabs-hardhat-etherscan.html) currently provides an interface to provide one API key for auto-verification. In the current ecosystem there is a need for being able to support multiple API keys. This feature request looks at how `hardhat-etherscan` works now, and outlines potential solutions.

### Current structure of `hardhat-etherscan`

The `hardhat-etherscan` enables auto-verification via an `etherscan` object in `hardhat.config`:
```javascript
module.exports = {
  networks: {
    mainnet: { ... }
  },
  etherscan: {
    // Your API key for Etherscan
    // Obtain one at https://etherscan.io/
    apiKey: "YOUR_ETHERSCAN_API_KEY"
  }
};
```
(code taken from [docs](https://hardhat.org/plugins/nomiclabs-hardhat-etherscan.html) for `hardhat-etherscan`)

### Problems

In today's ecosystem, Etherscan has spun off block explorers for popular sidechains such as BSC ([BSCScan](https://bscscan.com/)) and Fantom ([FTMScan](https://ftmscan.com/)), making the naming convention of the plugin itself slightly misleading. This recommends renaming in future versions to `hardhat-verification`. If Blockscout explorers become available, this will be doubly true.

In addition, calling the object in the config `etherscan` has the same concern.

Neither of these are meant to be the primary issue, though. The larger issue is that the `etherscan` object can only hold one API key at a time (via `etherscan.apiKey`). It would be valuable for projects deploying in multiple ecosystems to support multiple API keys.

#### 1. Long-term proposal

In the long term we would like to recommend renaming the plugin to `hardhat-verification`, and replacing the `etherscan` object in the config with `explorerApiKeys`, with the key being the name of a supported block explorer, something akin to:
```javascript
module.exports = {
  networks: {
    mainnet: { ... }
  },
  explorerApiKeys: {
    etherscan: "YOUR_ETHERSCAN_API_KEY",
    bscScan: "YOUR_BSCSCAN_API_KEY",
    ftmScan: "YOUR_FTMSCAN_API_KEY",
  }
};
```

Renaming the plugin and the object has the added benefit of backwards compatability - the `hardhat-etherscan` and corresponding `etherscan` object could remain supported without conflict.

[...]

PR #2130, opened 2021-12-06
Support multiple api keys

kanej:
Etherscan has expanded to provided block explorers and verifiers for popular sidechains. The config in the `hardhat-etherscan` plugin assumes there is only one `apiKey` for use in verification, even though the project may be targeting multiple network/sidechains.

This PR adds support in the `hardhat-etherscan` configuration for setting an api key for etherscan per network (i.e. a project could support project verification on both mainnet and polygon at once without hacks).

## Approach

Extend the current allowed structure of the `hardhat-etherscan` from:

```javascript
{
   etherscan: {
      apiKey: string, 
   }
}
```

to:

```javascript
{
   etherscan: {
      apiKey: string | EtherscanApiKeys, 
   }
}
```

Where `EtherscanApiKeys` is an object with one property for each supported network mapping to the apiKey to used when verifying on that network.

The supported networks are derived from the `networkConfig` object that combines the chainId and etherscan urls, and is used to define the typesafety of the etherscan config object (only networks present on the config object are allowed in a valid etherscan config).

Relates to #1448

## TODO

We'll release a new major version.
* [x] Add support for `EtherscanApiKeys` under etherscan config
* [x] Refactor supported networks so that each one has a name (used in the EtherscanApiKeys type)
* [x] The relationship between supported networks and EtherscanApiKeys should be typesafe
* [x] The key to be used when verifying a contract should be the one that maps to the currently selected network if the using EtherscanApiKeys, or the apiKey string otherwise
* [x] Update the README
* [x] Update the contribution guide
* [x] Manual testing against a sample of alternate networks

### DIFF
--- a/packages/hardhat-etherscan/README.md
+++ b/packages/hardhat-etherscan/README.md
@@ -124,11 +124,9 @@ module.exports = {
 
 ### Multiple API keys and alternative block explorers
 
-If your project targets multiple evm sidechains, you many need to verify using the block explorers particular to those chains, and so need to set multiple API keys.
+If your project targets multiple EVM-compatible networks that have different explorers, you'll need to set multiple API keys.
 
-A block explorer API key can be set per chain, hardhat will use the `chainId` of the specified network to determine which block explorer and API key to use.
-
-To configure the API keys for the chains you are using, provide an object with under `etherscan/apiKey` with properties matching the chain:
+To configure the API keys for the chains you are using, provide an object under `etherscan/apiKey` with the identifier of each chain as the key. **This is not necessarily the same name that you are using to define the network**. The snippet below shows the values you should use for each chain:
 
 ```js
 module.exports = {
--- a/packages/hardhat-etherscan/README.md
+++ b/packages/hardhat-etherscan/README.md
@@ -145,28 +145,28 @@ module.exports = {
         kovan: "YOUR_ETHERSCAN_API_KEY",
         // binance smart chain
         bsc = "YOUR_BSCSCAN_API_KEY",
-        bsc_testnet = "YOUR_BSCSCAN_API_KEY",
+        bscTestnet = "YOUR_BSCSCAN_API_KEY",
         // huobi eco chain
         heco = "YOUR_HECOINFO_API_KEY",
-        heco_testnet = "YOUR_HECOINFO_API_KEY",
+        hecoTestnet = "YOUR_HECOINFO_API_KEY",
         // fantom mainnet
         opera = "YOUR_FTMSCAN_API_KEY",
-        ftm_testnet = "YOUR_FTMSCAN_API_KEY",
+        ftmTestnet = "YOUR_FTMSCAN_API_KEY",
         // optimistim
-        optimistic_ethereum = "YOUR_OPTIMISTIC_ETHERSCAN_API_KEY",
-        optimistic_kovan = "YOUR_OPTIMISTIC_ETHERSCAN_API_KEY",
+        optimisticEthereum = "YOUR_OPTIMISTIC_ETHERSCAN_API_KEY",
+        optimisticKovan = "YOUR_OPTIMISTIC_ETHERSCAN_API_KEY",
         // polygon
         polygon = "YOUR_POLYGONSCAN_API_KEY",
-        polygon_mumbai = "YOUR_POLYGONSCAN_API_KEY",
+        polygonMumbai = "YOUR_POLYGONSCAN_API_KEY",
         // arbitrum
-        arbitrum_one = "YOUR_ARBISCAN_API_KEY",
-        arbitrum_testnet = "YOUR_ARBISCAN_API_KEY",
+        arbitrumOne = "YOUR_ARBISCAN_API_KEY",
+        arbitrumTestnet = "YOUR_ARBISCAN_API_KEY",
         // avalanche
         avalanche = "YOUR_SNOWTRACE_API_KEY",
-        avalanche_fuji_testnet = "YOUR_SNOWTRACE_API_KEY",
+        avalancheFujiTestnet = "YOUR_SNOWTRACE_API_KEY",
         // moonriver
         moonriver = "YOUR_MOONRIVER_MOONSCAN_API_KEY",
-        moonbase_alpha = "YOUR_MOONRIVER_MOONSCAN_API_KEY",
+        moonbaseAlpha = "YOUR_MOONRIVER_MOONSCAN_API_KEY",
     }
   }
 };
--- a/packages/hardhat-etherscan/README.md
+++ b/packages/hardhat-etherscan/README.md
@@ -58,6 +58,8 @@ module.exports = {
 };
 ```
 
+Alternatively you can specify more than one block explorer API key, by passing an object under the `apiKey` property, see [`Multiple API keys and alternative block explorers`](#multiple-api-keys-and-alternative-block-explorers).
+
 Lastly, run the `verify` task, passing the address of the contract, the network where it's deployed, and the constructor arguments that were used to deploy it (if any):
 
 ```bash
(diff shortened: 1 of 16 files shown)


packages/hardhat-etherscan/README.md
```diff
@@ -120,6 +122,56 @@ module.exports = {
...
+  etherscan: {
+    apiKey: {
+        mainnet: "YOUR_ETHERSCAN_API_KEY",
+        ropsten: "YOUR_ETHERSCAN_API_KEY",
+        rinkeby: "YOUR_ETHERSCAN_API_KEY",
+        goerli: "YOUR_ETHERSCAN_API_KEY",
+        kovan: "YOUR_ETHERSCAN_API_KEY",
+        // binance smart chain
+        bsc = "YOUR_BSCSCAN_API_KEY",
+        bsc_testnet = "YOUR_BSCSCAN_API_KEY",
```
kanej, line 148: @fvictorio did we decide this should become camel case?

alcuadrado: I think it should be camel case, as we are installing `eslint` with the recommended settings to our new users, and it enforces camel case.

kanej: I switched the enums to camelcase so `bscTestnet` rather than `bsc_testnest`.

packages/hardhat-etherscan/src/network/prober.ts
```diff
@@ -20,30 +20,31 @@ export async function getEtherscanEndpoints(
   }
 
   const chainIdsToNames = new Map(
-    Object.entries(chainConfig).map(([mapName, config]) => [
+    entries(chainConfig).map(([chainName, config]) => [
```
fvictorio, line 23: Created an `entries` function that has the proper types. I think we also have this function somewhere in `hardhat-core`, and probably in other places too. At some point, it might be useful to have a `hardhat-common` published package, with utils and common config.

packages/hardhat-etherscan/src/types.ts
```diff
@@ -1,43 +1,40 @@
-/* eslint-disable @typescript-eslint/naming-convention */
-
-export enum Chains {
-  mainnet = "mainnet",
-  ropsten = "ropsten",
-  rinkeby = "rinkeby",
-  goerli = "goerli",
-  kovan = "kovan",
+type Chain =
```
fvictorio, line 1: Converted this to a union type instead of an enum. I think this simplifies things a bit, but there might be an issue with this approach that I'm not seeing. Please let me know if that's the case and I'll revert this change.

packages/hardhat-etherscan/src/types.ts
```diff
@@ -1,43 +1,40 @@
...
+  | "moonbaseAlpha";
 
 export type ChainConfig = {
-  [Network in Chains]: EtherscanChainConfig;
+  [Network in Chain]: EtherscanChainConfig;
 };
 
-export type EtherscanApiKeys = {
-  [Network in keyof ChainConfig]?: string;
+type EtherscanApiKeys = {
```
fvictorio, line 36: Removed some unused exports.

packages/hardhat-etherscan/test/unit/ChainConfig.ts
```diff
@@ -3,7 +3,7 @@ import { chainConfig } from "../../src/ChainConfig";
 
 describe("Chain Config", () => {
   it("should have no duplicate chain ids", () => {
-    const chainIds = Object.values(chainConfig);
+    const chainIds = Object.values(chainConfig).map((config) => config.chainId);
```
fvictorio, line 6: I think this was an error. The way it was before, this test would always pass.

packages/hardhat-etherscan/src/index.ts
```diff
@@ -50,7 +50,7 @@ import {
   getEtherscanEndpoints,
   retrieveContractBytecode,
 } from "./network/prober";
-import resolveEtherscanApiKey from "./resolveEtherscanApiKey";
+import { resolveEtherscanApiKey } from "./resolveEtherscanApiKey";
```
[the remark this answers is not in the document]
kanej: Ah, I see.

What do you think of pulling in the `eslint-plugin-import` module and banning them via eslint with this rule:

https://github.com/import-js/eslint-plugin-import/blob/HEAD/docs/rules/no-default-export.md

packages/hardhat-etherscan/src/types.ts
kanej: I don't know any reason, other than a hazy preference for enums. Do we have a general preference for string unions over enums?

packages/hardhat-etherscan/test/unit/ChainConfig.ts
kanej: Well spotted. I think I refactored the structure of chainConfig, and the test worked before and after.

I have added a type guard on chainIds (number[]) to pickup changes in the structure of chainConfig.

packages/hardhat-etherscan/src/index.ts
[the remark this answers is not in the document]
fvictorio: That would be great, yeah. Feel free to send a PR with that change if you want. If not, please open an issue about doing that change so we don't forget.

packages/hardhat-etherscan/src/types.ts
fvictorio: It just feels slightly better in this case. It's a bit less code and IMO it's closer to the intention of what we want here (a list of keys). In other words, the way the enum was used was equivalent to just doing this, and without any benefit from using an enum.

The other reason that comes to mind is that enums exist at runtime, which makes them a more hybrid/complex thing. Union types are just types, so thinking about them is easier (again, IMO).

fvictorio:
I modified some error messages. I think they look a bit better now:

image

image


There is only one extra thing I'd like to have here related to error handling: to show a proper error when the user configures an API key under an invalid network name. This is only relevant for javascript projects, since in typescript that will simply not compile.

I think this is important because it's likely that users won't read the docs, and they'll try to use the network name as the key.

packages/hardhat-etherscan/src/index.ts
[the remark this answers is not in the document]
kanej: I have opened a PR with the rule here: https://github.com/nomiclabs/hardhat/pull/2163

kanej:
I have added a check on the etherscan api keys during `extendConfig`, so it throws if a js project has included an unsupported network as a key:

image

I have repeated the manual test of a verify against ropsten, but that was using the same linking approach as before @fvictorio.

packages/hardhat-etherscan/src/config.ts
```diff
@@ -8,6 +27,16 @@ export const etherscanConfigExtender: ConfigExtender = (
 
   if (config.etherscan !== undefined) {
     const customConfig = config.etherscan;
+
+    const unallowedChains = verifyAllowedChains(customConfig);
+
+    if (unallowedChains.length > 0) {
+      throw new NomicLabsHardhatPluginError(
+        pluginName,
+        `Etherscan API token "${unallowedChains[0]}" is for an unsupported network`
```
fvictorio, line 36:
```suggestion
        `Etherscan API token "${unallowedChains[0]}" is for an unsupported network

Learn more at https://hardhat.org/plugins/nomiclabs-hardhat-etherscan.html#multiple-api-keys-and-alternative-block-explorers`
```

packages/hardhat-etherscan/test/unit/etherscanConfigExtender.ts
```diff
@@ -0,0 +1,67 @@
...
+        etherscan: {
+          apiKey: {
+            newhotness: "example_token",
+            newhotness2: "example_token",
+          },
+        },
+      } as any;
+
+      etherscanConfigExtender(resolvedConfig, invalidEtherscanConfig);
+    }, 'Etherscan API token "newhotness" is for an unsupported network');
```
fvictorio, line 65: I think this test will break with the other suggestion:

```suggestion
    }, /Etherscan API token "newhotness" is for an unsupported network/);
```

fvictorio (approved):
Small comments, pre-approving

packages/hardhat-etherscan/test/unit/etherscanConfigExtender.ts
kanej: So did I, but after applying and testing locally it worked ... so lets move on and never think about it again.

Shelvak:
I pulled master and just use the multi-api-key, amazing feature :+1:  Thank you so much :rocket: :rocket:

fvictorio:
@Shelvak oh, thank you for testing it! We'll probably release this feature this week.

PR #2131, opened 2021-12-06
Update pragma version from ^0.7.3 to >=0.7.3

ogzhanolguncu:
Update pragma version from ^0.7.3 to >=0.7.3 to prevent error of 'Source file requires different compiler version' in VSCode using solidity extension.

---


alcuadrado:
Thanks for sending this PR, @ogzhanolguncu! Unfortunately, we consider `>=` pragmas to be a bad practice, so we won't be merging it.

I understand how annoying that error is. We'll have [good news](https://medium.com/nomic-labs-blog/hardhat-vscode-9de29467fc26) for VSCode users soon.

PR #2134, opened 2021-12-07
Increase timeout in localhost from 25s to 40s

### LINKED ISSUE #1549: Increase timeout for localhost network

Amit0617:
fixes #1549 One of the issues raised there.
---

There was a commit already but that was for docs rather than actual code. So i made the required one.

### DIFF
--- a/packages/hardhat-core/.mocharc.json
+++ b/packages/hardhat-core/.mocharc.json
@@ -6,5 +6,5 @@
     "test/fixture-projects/**/*.js",
     "test/helpers/**/*.ts"
   ],
-  "timeout": 25000
+  "timeout": 40000
 }


fvictorio:
Hey @Amit0617, what you modified is the timeout of Hardhat's test suite. The issue is about increasing the default value of the timeout property for the `localhost` network.

Amit0617:
> Hey @Amit0617, what you modified is the timeout of Hardhat's test suite. The issue is about increasing the default value of the timeout property for the `localhost` network.

I made that following label `package:hardhat-core` in that issue.

Amit0617:
> Hey @Amit0617, what you modified is the timeout of Hardhat's test suite. The issue is about increasing the default value of the timeout property for the `localhost` network.

Made a new commit #2161 
I believe this was the required one.

fvictorio:
Closing in favor of #2161.

Issue #1490, opened 2021-05-27
Show a warning when the config is empty


fvictorio:
An user could make something like this in the config:

```ts
let config: HardhatUserConfig = {
  ...
}
```

and forget to do `export default config`. We should show a warning when that happens.

alcuadrado:
I don't think this is feasible, as a file that doesn't export anything in node is equivalent to one exporting an empty object, and that's a valid config.

fvictorio:
Oh, I know, but I think it's a good trade-off. We can just check `Object.keys(config).length` and I think it's way more likely that this helps someone than this annoys a user that explicitly wants an empty config (I don't see why anyone would do that).

Endlessline:
I took a stab at this, is it something you all think would be useful? It sounds like it still may be up for debate.

fvictorio:
Fixed in [`hardhat@2.8.1`](https://github.com/nomiclabs/hardhat/releases/tag/hardhat%402.8.1).

PR #2138, opened 2021-12-08
Change behaviour to error on empty HardhatConfig

kanej:
If a user fails to export from the HardhatConfig.{js,ts} file or exports an empty object, we now throw an error.

This means updating some of our tests cases which have been using the `{}` object as the base case.

Relates to #1490


alcuadrado:
This is a great PR, @kanej! It's great that you also updated the docs examples!

IIRC we were going to print a warning, and not error in this case. @fvictorio any idea?

kanej:
> This is a great PR, @kanej! It's great that you also updated the docs examples!
> 
> IIRC we were going to print a warning, and not error in this case. @fvictorio any idea?

That is me, I have read the issue and somehow completely interpreted "warning" with "error". I will update the PR @alcuadrado

kanej:
Switched this PR to just warn on empty config rather than error.

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -38,7 +38,16 @@ export function resolveConfigPath(configPath: string | undefined) {
 
 export function loadConfigAndTasks(
   hardhatArguments?: Partial<HardhatArguments>,
-  { showSolidityConfigWarnings } = { showSolidityConfigWarnings: false }
+  {
+    showEmptyWarnings = false,
```
fvictorio, line 42: Personal preference: I'd rename this to `showEmptyConfigWarning`

packages/hardhat-core/test/fixture-projects/solidity-config-warnings/config-without-solidity.js
```diff
@@ -1 +1 @@
-module.exports = {};
+module.exports = { solidity: undefined };
```
fvictorio, line 1: This feels a bit ugly. Maybe it would be better to add some other field to prevent the config from being empty, like:

```
module.exports = {
  networks: {}
}
```

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -220,6 +233,25 @@ function readPackageJson(packageName: string): PackageJson | undefined {
   }
 }
 
+function checkEmptyConfig(
+  userConfig: any,
+  {
+    showSolidityConfigWarnings = false,
```
fvictorio, line 239: I don't think it's necessary to have default values here, especially because it's a private function that's only called once.

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -220,6 +233,25 @@ function readPackageJson(packageName: string): PackageJson | undefined {
...
+  }: { showSolidityConfigWarnings: boolean } = {
+    showSolidityConfigWarnings: false,
+  }
+) {
+  if (userConfig === undefined || Object.keys(userConfig).length === 0) {
+    let warning = `Hardhat config is returning an empty config object, check the export from the config file if this is unexpected.\n`;
+
+    if (!showSolidityConfigWarnings) {
+      warning += `Learn more about compiler configuration at https://hardhat.org/config"\n`;
```
fvictorio, line 248: This one doesn't apply only to compilation.

```suggestion
      warning += `\nLearn more about configuring Hardhat at https://hardhat.org/config"\n`;
```

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -220,6 +233,25 @@ function readPackageJson(packageName: string): PackageJson | undefined {
...
+  {
+    showSolidityConfigWarnings = false,
+  }: { showSolidityConfigWarnings: boolean } = {
+    showSolidityConfigWarnings: false,
+  }
+) {
+  if (userConfig === undefined || Object.keys(userConfig).length === 0) {
+    let warning = `Hardhat config is returning an empty config object, check the export from the config file if this is unexpected.\n`;
+
+    if (!showSolidityConfigWarnings) {
```
fvictorio, line 247: Let's add a comment here explaining that this is meant to avoid showing the "learn more..." thing twice.

I don't really like this workaround, but tbh I would've probably done the same thing. Next time we modify something around config and warnings, we should refactor this and pay this technical debt.

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -220,6 +233,23 @@ function readPackageJson(packageName: string): PackageJson | undefined {
...
+  userConfig: any,
+  { showSolidityConfigWarnings }: { showSolidityConfigWarnings: boolean }
+) {
+  if (userConfig === undefined || Object.keys(userConfig).length === 0) {
+    let warning = `Hardhat config is returning an empty config object, check the export from the config file if this is unexpected.\n`;
+
+    // This 'learn more' section is also printed by the solidity config warning,
+    // so we need to check to avoid printing it twice
+    if (!showSolidityConfigWarnings) {
+      warning += `\nLearn more about configuring Hardhat at https://hardhat.org/config"\n`;
```
fvictorio, line 246:
```suggestion
      warning += `\nLearn more about configuring Hardhat at https://hardhat.org/config\n`;
```

fvictorio (approved):
Minor comment, but pre-approving

Issue #1720, opened 2021-07-23
Tests can't be run programmatically more than once


fvictorio:
Reproduction steps:

1. Create the sample project
2. Create a script like this:

    ```js
    const hre = require("hardhat");

    async function main() {
      await hre.run("test")
      await hre.run("test")
    }

    main()
      .then(() => process.exit(0))
      .catch(error => {
        console.error(error);
        process.exit(1);
      });
    ```

3. Run `hh run script.js`

**Expected result**: Tests should be run twice.

**Actual result**: the second run doesn't find any tests:

```

  Greeter
Deploying a Greeter with greeting: Hello, world!
Changing greeting from 'Hello, world!' to 'Hola, mundo!'
    ✓ Should return the new greeting once it's changed (782ms)


  1 passing (785ms)



  0 passing (0ms)
```

This seems like a "mocha caching something" thing.

alcuadrado:
> This seems like a "mocha caching something" thing.

It's mocha loading the test files with `require`, so they only get interpreted once.

0xww:
I found same issue closed on mocha repo [#2783](https://github.com/mochajs/mocha/issues/2783). can be helpful.

fvictorio:
Reopening so we can close it when this is released.

fvictorio:
Fixed in [`hardhat@2.8.1`](https://github.com/nomiclabs/hardhat/releases/tag/hardhat%402.8.1).

PR #2140, opened 2021-12-08
Run tests twice

kanej:
This is a mocha caching and cleanup issue addressed here: mochajs/mocha#2783.

Relates to #1720

## Testing

An integration test has been added to e2e to cover the run twice scenario and guard against regression of the bug.

### Manual

To test manually:

1. create a new advanced typescript hardhat project
2. link to `hardhat-core
3. create a new scripts file `./scripts/multi-run.js`:

```javascript
// eslint-disable-next-line import/no-extraneous-dependencies
const hre = require("hardhat");

async function main() {
  const code = await hre.run("test");

  if (code > 0) {
    console.error("Failed first test run");
    process.exit(1);
  }

  const secondCode = await hre.run("test");

  if (secondCode > 0) {
    console.error("Failed second test run");
    process.exit(1);
  }

  console.log("\nFinished and complete\n");
}

main()
  .then(() => process.exit(0))
  .catch((error) => {
    console.error(error);
    process.exit(1);
  });

```

4. Run the script `npx hardhat run ./scripts/multi-run.js`
5. Confirm the script runs without error and that tests are present in both runs:

image

As a test against a more in-depth test suite, I did the equivalent of the above on the open-zepplin test suite (updating to the latest hardhat libs first) and was able to run the suite twice programmatically with the script above.

## TODO

* [x] Apply fix from mocha suggestions
* [x] Look at options on unit tests
* [x] Run against existing project with large test suite


alcuadrado:
This is really interesting! This new API of mocha is something I wasn't expecting tbh.

I think the decision we need to make is if we are ok with bumping the mocha dependency. If we do, I think we should:

1. Document that we may do it in our stability guarantees docs.
2. Bump the minor version.
3. Make sure that changes between 5 and 9 aren't too aggressive.
4. Make sure mocha 9 supports all the node.js versions than we do.

kanej:
> This is really interesting! This new API of mocha is something I wasn't expecting tbh.
> 
> I think the decision we need to make is if we are ok with bumping the mocha dependency. If we do, I think we should:
> 
>     1. Document that we may do it in our stability guarantees docs.
> 
>     2. Bump the minor version.
> 
>     3. Make sure that changes between 5 and 9 aren't too aggressive.
> 
>     4. Make sure mocha 9 supports all the node.js versions than we do.

This fix is adding `mocha.dispose` after run in the `test` task. This resolves the problem in local testing for a simple project - more testing is needed on larger test suites.

`dispose` was introduced in mocha `7.2.0` [changelog](https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#720--2020-05-22) as part of Mocha's multi-run fix (https://github.com/mochajs/mocha/pull/4234).

Hardhat-core has a dependency on `^7.1.2` which get resolved to `7.2.0` - so `dispose` is present in a local build. `dispose` is not present in our current `@types/mocha` however, and you have to bump the version to 9 to get it.

So in this PR I propose:

* bump mocha as a dep in `hardhat-core` to `7.2.0` to be explicit
* bump mocha as a dev-dep everywhere else to `7.2.0` for consistency in our local dev environment
* bump `@types/mocha` to 9 in dev-deps in all packages to allow use of dispose and consistency across packages - no code changes are required to support this. Note that the sample typescript project already provides a default `@types/mocha` on version 9.0.0
 
I think moving the dep in `hardhat-core` from `^7.1.2` to `7.2.0` is safe and explicit and doesn't require a change to our stability promises. From the package.json mocha 7.2 supports node 8.10 and above:

image

@alcuadrado @fvictorio

alcuadrado:
That's a good find! I think that makes sense

kanej:
I have pulled the `open-zepplin` repo and done a multi-test script using this branch, and the their suite ran twice successfully with all tests.

If we are happy with the version bumps this PR is ready for review.

packages/e2e/test/fixture-projects/basic-project/test/simple.js
```diff
@@ -0,0 +1,5 @@
+describe("simple", () => {
+  it("should pass", () => {
+    // throw new Error("not equal");
```
fvictorio, line 3: Can we remove this comment?

fvictorio (approved):
Just one very minor comment, otherwise LGTM

PR #2144, opened 2021-12-09
Add @primitivefi/hardhat-dodoc to the plugins

clemlak:
I just released a new plugin to generate documentation based on the NatSpec comments of the contracts.
Can we add it to the community plugins list?


fvictorio:
This is really cool @clemlak, thank you!

PR #2163, opened 2021-12-16
Add eslint rule to ensure no default exports

kanej:
As part of hardhats coding conventions we prefer named over default exports from modules. This PR enables an eslint rule to support this.

This is using the eslint import plugin (already installed) and updates the global config.

## Preview

image


alcuadrado (approved):
Nice!

PR #2183, opened 2021-12-22
preserve user's readme on project creation

morgansliman:


feuGeneA (commented):
I wonder if it would be better to take a more general approach.  Like, just overwrite everything (but give the user a warning that that will happen, shown when presenting the project type selection menu), or, use the `errorOnExist` option of `fxExtra.copy`, and simply fail if anything would be overwritten.

morgansliman:
While I don't think using `errorOnExist` would be a good user experience (if the user wanted to keep their own README they'd be forced to do what this PR is doing anyways), I do think there could be room to ask them during the project menu like you suggest.

Something like "hey, we noticed you already have a README; would you like to keep yours or overwrite it with ours?", but that may be overkill too. Not sure.

feuGeneA:
> While I don't think using `errorOnExist` would be a good user experience (if the user wanted to keep their own README they'd be forced to do what this PR is doing anyways), I do think there could be room to ask them during the project menu like you suggest.
> 
> Something like "hey, we noticed you already have a README; would you like to keep yours or overwrite it with ours?", but that may be overkill too. Not sure.

That probably is overkill, in my opinion. I was aiming to reduce the complexity here, not add to it.

Forcing the user to do what this PR is doing is not too crazy, in my opinion, since it's probably a relatively rare occurrence, and that manual step would act as a sort of medium-weight confirmation, sort of how AWS S3 makes you physically type out `permanently delete` in order to delete a file.

To be clear, I was thinking we could _catch_ what gets thrown via `errorOnExist`, rather than letting it bubble up to the user, so that we could control the experience and provide some nice messaging.

In any case, I'm just sharing my thoughts, brainstorming, and I'm happy to defer to others' opinions.

fvictorio:
I agree with @feuGeneA that we should take a more general approach, even if the original issue only mentions the `README` file.

With respect to the `errorOnExist` option, I think that won't work. When you copy directories, it copies as many files as it can and it fails when it finds a conflict. This means that if we do that, we'll pollute the directory even if the project creation then fails.

Two options come to mind:

1. Use `{ overwrite: false }` in `fsExtra.copy`. That will just ignore existing files. We could also show a warning indicating which files were left intact.
2. Do what `create-react-app` does:

    image

Both seem good enough to me. I personally prefer 2, since I think project creation should be all-or-nothing. I also think that that approach will result in simpler code, even if it's a bit harder to implement.

If we do option 2, notice that `create-react-app` seems to complain about almost _any_ existing file (except things like `.git`, apparently). I wouldn't do that; I would just check for potential conflicts.

morgansliman:
image

packages/hardhat-core/src/internal/cli/project-creation.ts
```diff
@@ -142,6 +142,44 @@ async function printWelcomeMessage() {
...
+      srcFiles = union(
+        srcFiles,
+        fsExtra.readdirSync(path.join(srcPath, "advanced"))
+      );
+      break;
+    case Action.CREATE_ADVANCED_TYPESCRIPT_SAMPLE_PROJECT_ACTION:
+      srcFiles = union(
+        srcFiles,
+        fsExtra.readdirSync(path.join(srcPath, "advanced")),
+        fsExtra.readdirSync(path.join(srcPath, "advanced-ts"))
```
alcuadrado, line 168: Won't this warn about unnecessary things? For example, `hardhat.config.js` in a ts project.

I don't remember how the `advanced-ts` project creation works. Is it applied on top of the `advanced` one?

alcuadrado (commented):
I like this approach. It's general enough and safe.

I just left two minor comments.

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -216,6 +216,19 @@ If you were trying to create a new project, please try again using Windows Subsy
 You can learn how to use Hardhat by reading the [Getting Started guide](../getting-started).`,
       shouldBeReported: false,
     },
+    CONFLICTING_FILES: {
+      number: 16,
+      message: `The directory %dest% contains files that could conflict:
+
+%conflicts%
+
+Either try using a new directory name, or remove the files listed above.`,
```
fvictorio, line 225:
```suggestion
Either try using a new directory, or remove the files listed above.`,
```

packages/hardhat-core/src/internal/cli/project-creation.ts
```diff
@@ -142,6 +142,44 @@ async function printWelcomeMessage() {
...
+  const packageRoot = getPackageRoot();
+
+  const srcPath = path.join(packageRoot, "sample-projects");
+  const destFiles = fsExtra.readdirSync(dest);
+  let srcFiles: string[] = fsExtra.readdirSync(path.join(srcPath, "basic"));
+
+  switch (projectType) {
+    case Action.CREATE_ADVANCED_SAMPLE_PROJECT_ACTION:
+      srcFiles = union(
+        srcFiles,
```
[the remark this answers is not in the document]
fvictorio: Yes.

packages/hardhat-core/src/internal/cli/project-creation.ts
fvictorio: I'm ok with this trade-off, my guess is that it won't matter that much in practice.

packages/hardhat-core/src/internal/cli/project-creation.ts
```diff
@@ -142,6 +142,44 @@ async function printWelcomeMessage() {
   );
 }
 
+async function checkForDuplicates(
+  dest: string,
```
fvictorio, line 146: I would rename this to `projectRoot`.

.changeset/hip-owls-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+preserve user's readme on project creation (#1942)
```
alcuadrado, line 5:
```suggestion
Preserve any existing user's README when initializing a project (#1942)
```

What about this wording? I think something like this will look better in our release notes.

alcuadrado (approved):

PR #2204, opened 2022-01-01
add `hardhat-4byte-uploader` and `hardhat-test-short-circuit` to plugin list

ItsNickBarry:


fvictorio:
Thanks Nick! I'm curious about the `hardhat-test-short-circuit` plugin. Does it make the tests behave as if you've passed the `--bail` flag to mocha, or does it do something else?

ItsNickBarry:
I wasn't familiar with `--bail`.  This is a bit different.  The user can run a task to abort pending tests, and any stack traces from failures are then printed immediately.  There are better ways to achieve this (I'm told that Webstorm IDE prints each stack trace in real time), but none that work directly with Hardhat as far as I know.

PR #2206, opened 2022-01-03
ethers.parseUnit return undefined

DedicatedDev:
**bug fix**
in Typescript, ethers.parseUnit return  undefined. this issue is fixed in ether 5.0.17. so I've updated package json file in hardhat/packages/ether


fvictorio:
@DedicatedDev `hardhat-ethers` doesn't come with a specific version of ethers, that's why it's a peer dependency. If you are using version X and there's a fix in a later version, you can just upgrade the `ethers` package.

Merging this would mean forcing people to use a version after `5.0.17`. Admittedly, not a lot of people must be using those old versions, but I think we'd rather keep the peer dependency loose if there isn't a good reason to change it.

PR #2212, opened 2022-01-05
Remove custom state manager

### LINKED ISSUE #1590: Remove custom state manager and update the one used when forking
We added a custom (non-fork) state manager to work around a bug in EthereumJS. That bug has been allegedly fixed, so we should remove this workaround and see if it works now.

feuGeneA:
To fix issue #1590 
- [x] Remove custom state manager.
- [ ] Update the state manager used when forking. (Unnecessary? See [this comment](https://github.com/nomiclabs/hardhat/issues/1590#issuecomment-1005850041).)


fvictorio (approved):
Tested this with the sample project (with and without forking) and OZ's tests.

PR #2219, opened 2022-01-06
Show a special error when outside of a project on windows

### LINKED ISSUE #1408: Hardhat cannot be initialized in `git bash`
Reproduction steps:

1. Use windows (sorry)
2. Open git bash
3. Install hardhat in an empty directory `npm i --save-dev hardhat`
4. `npx hardhat`

The project initialization project should show up, but instead the "not inside a Hardhat project" message is shown.

alcuadrado:
This PR fixes #1408

@aftermathdigital, @fvictorio WDYT?

### DIFF
--- a/packages/hardhat-core/src/internal/cli/cli.ts
+++ b/packages/hardhat-core/src/internal/cli/cli.ts
@@ -114,7 +114,7 @@ async function main() {
       // situation we throw a special error instructing the user to use WSL or
       // powershell to initialize the project.
       if (process.platform === "win32") {
-        throw new HardhatError(ERRORS.GENERAL.NON_LOCAL_INSTALLATION);
+        throw new HardhatError(ERRORS.GENERAL.NOT_INSIDE_PROJECT_ON_WINDOWS);
       }
     }
 
--- a/packages/hardhat-core/src/internal/cli/cli.ts
+++ b/packages/hardhat-core/src/internal/cli/cli.ts
@@ -94,20 +94,28 @@ async function main() {
 
     showStackTraces = hardhatArguments.showStackTraces;
 
-    if (
-      hardhatArguments.config === undefined &&
-      !isCwdInsideProject() &&
-      (process.stdout.isTTY === true ||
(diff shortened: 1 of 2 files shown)


packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -94,20 +94,28 @@ async function main() {
...
+        await createProject();
+        return;
+      }
+
+      // Many terminal emulators in windows fail to run the createProject()
+      // workflow, and don't present themselves as TTYs. If we are in this
+      // situation we throw a special error instructing the user to use WSL or
+      // powershell to initialize the project.
+      if (process.platform === "win32") {
+        throw new HardhatError(ERRORS.GENERAL.NON_LOCAL_INSTALLATION);
```
fvictorio, line 117: Shouldn't this be...?

```suggestion
        throw new HardhatError(ERRORS.GENERAL.NOT_INSIDE_PROJECT_ON_WINDOWS);
```

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -201,6 +201,21 @@ Please run: npm install --save-dev typescript`,
...
+
+If you were trying to create a new project, please try again using Windows Subsystem for Linux (WSL) or PowerShell.
+`,
+      title:
+        "You are not inside a Hardhat project and Hardhat failed to initialize a new one",
+      description: `You are trying to run Hardhat outside of a Hardhat project, and we couldn't initialize one.
+
+If you were trying to create a new project, please try again using Windows Subsystem for Linux (WSL) or PowerShell.
+
+You can learn how to use Hardhat by reading the [Getting Started guide](../getting-started).`,
```
fvictorio, line 216: Does it make sense to include this line?

alcuadrado: Well, I guess that if it makes sense in HH1 it makes sense here, as it's the same error, just for windows.

fvictorio (approved):

PR #2222, opened 2022-01-07
Align `@types/mocha` with `mocha`'s version

yhuard:

---

In https://github.com/nomiclabs/hardhat/pull/2140, `@types/mocha` has been updated to `^9.0.0`, yet hardhat still uses `mocha` v7.

This PR aligns `@types/mocha` version with `mocha`'s.

Note that the type definition provided by [`@types/mocha@7.0.2`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3dce8b667f77c414b123be249d1ea6e98804ea4f/types/mocha/index.d.ts) is not 100% accurate, e.g. it's missing the [`dispose()` function](https://github.com/mochajs/mocha/blob/v7.2.0/lib/mocha.js#L575), but I think it's still better than using `@types/mocha@9.0.0`. 

Of course, the ideal solution would be to upgrade `mocha` to v9 ;)

### DIFF
--- a/packages/hardhat-core/src/builtin-tasks/test.ts
+++ b/packages/hardhat-core/src/builtin-tasks/test.ts
@@ -57,6 +57,10 @@ subtask(TASK_TEST_RUN_MOCHA_TESTS)
       mocha.run(resolve);
     });
 
+    // [@types/mocha@7.0.2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3dce8b667f77c414b123be249d1ea6e98804ea4f/types/mocha/index.d.ts)
+    // is missing the definition of `dispose`, which is actually implemented
+    // https://github.com/mochajs/mocha/blob/v7.2.0/lib/mocha.js#L575
+    // @ts-ignore
     mocha.dispose();
 
     return testFailures;
--- a/packages/hardhat-core/src/builtin-tasks/test.ts
+++ b/packages/hardhat-core/src/builtin-tasks/test.ts
@@ -60,7 +60,7 @@ subtask(TASK_TEST_RUN_MOCHA_TESTS)
     // [@types/mocha@7.0.2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3dce8b667f77c414b123be249d1ea6e98804ea4f/types/mocha/index.d.ts)
     // is missing the definition of `dispose`, which is actually implemented
     // https://github.com/mochajs/mocha/blob/v7.2.0/lib/mocha.js#L575
-    // @ts-ignore
+    // @ts-expect-error
     mocha.dispose();
 
     return testFailures;
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -21,7 +21,7 @@
   "dependencies": {
     "@types/chai": "^4.2.0",
     "@types/fs-extra": "^5.1.0",
-    "@types/mocha": "^9.0.0",
+    "@types/mocha": "^7.0.2",
     "@types/node": "^12.0.0",
     "@types/shelljs": "^0.8.6",
     "@typescript-eslint/eslint-plugin": "4.29.2",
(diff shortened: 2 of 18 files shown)


fvictorio:
Hmm, I think this makes sense if it's compatible with the changes done in #2140 (which it seems like it is, otherwise the CI would have some compilation errors).

@kanej wdyt?

fvictorio:
> it's missing the [`dispose()` function](https://github.com/mochajs/mocha/blob/v7.2.0/lib/mocha.js#L575)

Oh, missed this in the description, sorry. Yeah, I'm not sure what the best approach is here.

yhuard:
Unfortunately, the definition of `dispose` was only [added to `@types/mocha@8.0.4`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49382), yet the function was already [available in `mocha@7.2.0`](https://github.com/mochajs/mocha/blob/v7.2.0/lib/mocha.js#L575). The types definitions are not perfect. 
I find it weird though to rely on [type definitions targeting mocha 9.0](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/mocha/index.d.ts#L1) while still using mocha 7.
Just my 2¢ :)

packages/hardhat-core/src/builtin-tasks/test.ts
```diff
@@ -57,6 +57,10 @@ subtask(TASK_TEST_RUN_MOCHA_TESTS)
       mocha.run(resolve);
     });
 
+    // [@types/mocha@7.0.2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3dce8b667f77c414b123be249d1ea6e98804ea4f/types/mocha/index.d.ts)
+    // is missing the definition of `dispose`, which is actually implemented
+    // https://github.com/mochajs/mocha/blob/v7.2.0/lib/mocha.js#L575
+    // @ts-ignore
```
kanej, line 63: Can we change this to `@ts-expected-error` so that the compiler will shout at us when it is safe to remove the comment?

kanej:
Yeah. These are rubbish tradeoffs. My instincts are with you @yhuard, it feels right that the lib and types should have the same version number; the problem is this is just not the guarantee that `@types` makes.

We can artificially force the same version number and paper over the type errors, but I think that would just be telling ourselves what we want to hear. If we opened up an issue with v7.2.0 of `@types/mocha` due to the lack of a dispose method, they can justifiably say that there is not a hard connection between version numbers in mocha and types - it is not a bug.

Given that `@types/mocha` don't give the guarantee of version compatibility, I don't think we can add that constraint back in down stream.

@yhuard your right that an upgrade to `9.0.0` would be ideal - it probably isn't that painful an upgrade.

At the same time, this is not a ditch I am willing to die in, I don't think it makes a big difference now, and I think when we upgrade to 9.0.0 this discussion goes away for a few years until the next time we use a function that is in mocha but not types for the same version.

packages/hardhat-core/src/builtin-tasks/test.ts
yhuard: Good point, thanks! I've replaced `@ts-ignore` with `@ts-expect-error`.

yhuard:
> it feels right that the lib and types should have the same version number; the problem is this is just not the guarantee that @types makes

They try to keep the major versions aligned, at least. Regarding `@types/mocha`, I know they do, I checked it. But there's no guarantee that the type definitions are 100% correct, as can be seen in this case.

I also don't think DefinitelyTyped would accept a PR to update an old `@types/mocha` version.

Regarding an upgrade to mocha v9, I can give it a look if you want. I haven't seen many huge breaking changes in the logs. The main change is that mocha v9 only supports Node.js ≥ 12. This makes sense since all the previous versions are not maintained anymore.

tannakartikey:
With b97808ff2fae55271a77adc759b1d08719ec7ec4 Mocha is upgraded so I think this is outdated now.

yhuard:
Exact, thanks @tannakartikey

PR #2228, opened 2022-01-08
add unit for gasPrice

aldilaff:
Make it more clear what the unit for gasPrice should be

---

Hello! I am fairly new to the web3 world but it took me a long time to figure out what the unit for gasPrice should be when trying to deploy to mainnet. I hope this helps other developers looking at the page.

### DIFF
--- a/docs/hardhat-network/reference/README.md
+++ b/docs/hardhat-network/reference/README.md
@@ -30,7 +30,7 @@ Its value should be `"auto"` or a number. If a number is used, it will be the ga
 
 #### `gasPrice`
 
-Its value should be `"auto"` or a number. This parameter behaves like `gas`. Default value: `"auto"`.
+Its value should be `"auto"` or a number. This parameter behaves like `gas`. Default value: `"auto"`. The unit for this field is in Gwei.
 
 #### `gasMultiplier`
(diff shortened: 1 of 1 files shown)


fvictorio:
Hey @aldilaff, thanks for pointing out that the description wasn't mentioning the unit, but the unit used is wei, not gwei. I updated the file to be more explicit about that.

aldilaff:
@fvictorio thank you for updating the unit. I don't think I can sign the CLA right now so please feel free to apply the change in a different PR. Is possible to also add a unit for the `gas` field? Thank you!

fvictorio:
> Is possible to also add a unit for the gas field?

What do you mean? There aren't units for gas AFAIK.

aldilaff:
I see. I'm still new to the space so I apologize about that. Reading that `gasPrice`"behaves like `gas`.", I thought that `gas` would also have a unit.

fvictorio:
Oh I see. No, gas is just gas: one transaction spends an X amount of it.

Then `gasPrice` is how much ETH (in wei) you pay for each unit of gas used. And since ETH has units (wei, gwei, eth), then gasPrice has those too.

PR #2268, opened 2022-01-18
docs: add @georacle/hardhat-georacle to community plugins

tacotaha:
This PR adds a new community plugin: https://github.com/georacleapi/hardhat-georacle.
The plugin allows users to deploy contracts with Georacle and access location data on chain.


fvictorio:
Hey @tacotaha, it seems like this plugin is already included in our [list](https://hardhat.org/plugins/)!

Issue #2260, opened 2022-01-14
Typescript errors in 2.0.4 eg Not all code paths return a value


payprmoney:
When I bump from version 2.0.3 to 2.0.4, I get new typescript compilation errors I didn't see before:

```
node_modules/hardhat/src/internal/hardhat-network/stack-traces/model.ts:278:55 - error TS7030: Not all code paths return a value.

278   public static fromABI(name: string, inputs: any[]): CustomError | undefined {
                                                          ~~~~~~~~~~~~~~~~~~~~~~~

node_modules/hardhat/src/types/config.ts:273:11 - error TS2503: Cannot find namespace 'Mocha'.

273   mocha?: Mocha.MochaOptions;
              ~~~~~

node_modules/hardhat/src/types/config.ts:281:10 - error TS2503: Cannot find namespace 'Mocha'.

281   mocha: Mocha.MochaOptions;
             ~~~~~


Found 3 errors.
```

I have confirmed that everything works great when moving back to 2.0.3

fvictorio:
Hi @payprmoney, thanks for reporting this. You are upgrading to version `2.0.4` of what? Hardhat? Or some plugin?

Also, is your repo public, or there's some way I can reproduce this?

payprmoney:
The specific library we are updating is `@nomiclabs/hardhat-ethers` from `2.0.3` to `2.0.4`.

The repo is [ethereum-contracts](https://github.com/paypr/ethereum-contracts/) and [this is a PR](https://github.com/paypr/ethereum-contracts/pull/44) that is [failing](https://app.circleci.com/pipelines/github/paypr/ethereum-contracts/213/workflows/caa182ff-ac07-4772-a734-e50e8d03ee66/jobs/191).

fvictorio:
Thank you! I'll take a look.

fvictorio:
Hey @payprmoney, thanks a lot for reporting this. Definitely a bug on our side. We'll fix this and release a new version soon.

The underlying problem is that, when you run `tsc`, the Harhdat source is being compiled, and it's not compatible with your config. For the time being, there is a (lame) workaround you can do:

- Install `@types/mocha`
- Change your `tsconfig.json` to have `noImplicitReturns: false`

This assumes that you need the new methods added in hardhat-ethers v2.0.4. If not, then downgrading is perfectly fine.

fvictorio:
Yes, we fixed this but haven't released it yet. We'll do it early next week.

[...]

PR #2272, opened 2022-01-19
Fix imports with hardhat/src

fvictorio:
Related to #2260.

### DIFF
--- a/config/eslint/eslintrc.js
+++ b/config/eslint/eslintrc.js
@@ -207,7 +207,7 @@ module.exports = {
     ],
     "use-isnan": "error",
     "no-restricted-imports": ["error", {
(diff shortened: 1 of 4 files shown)


config/eslint/eslintrc.js
```diff
@@ -206,5 +206,8 @@ module.exports = {
       },
     ],
     "use-isnan": "error",
+    "no-restricted-imports": ["error", {
+      patterns: ["hardhat/src"]
```
alcuadrado, line 210:
```suggestion
      patterns: ["hardhat/src", "@nomiclabs/*/src"]
```

alcuadrado (approved):
Left a small suggestion, but LGTM. We should ship this quickly.

PR #2275, opened 2022-01-19
Update building-plugins.md

mpaler:

---

Took me a while to figure out I had to import extendEnvironment into hardhat.config.ts to be able to extend the environment there. This might save some folks some headaches.

### DIFF
--- a/docs/advanced/building-plugins.md
+++ b/docs/advanced/building-plugins.md
@@ -19,6 +19,8 @@ The Hardhat Runtime Environment (HRE) is configured through a queue of extension
 For example, adding the following to `hardhat.config.js`:
 
 ```js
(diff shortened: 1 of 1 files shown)


fvictorio:
Hey @mpaler, sorry for not reviewing this before. Importing the `extendEnvironment` function is not necessary in js projects, and the example mentions that this is a `hardhat.config.js` file. And in fact, since it's a js file, the `import` will be a syntax error (well, unless it's an ES module, but we don't support those yet).

I'm not sure how we can improve this. Maybe adding a `require` instead, even if it's not necessary? The problem is that that is not consistent with the rest of the docs.

alcuadrado:
I agree with @fvictorio.

Maybe we can add a notice explaining what Franco wrote.

PR #2279, opened 2022-01-20
warn on node version greater than current LTS

### LINKED ISSUE #2120: Print a warning if an unsupported Node.js version is used
We are currently printing a warning if a version of Node.js that's too old and we don't support any more is used. We should do the equivalent for versions that are too new.

morgansliman:
closes #2120

### DIFF
--- a/.changeset/rude-suns-speak.md
+++ b/.changeset/rude-suns-speak.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+warn on node version greater than current LTS
--- a/.changeset/rude-suns-speak.md
+++ b/.changeset/rude-suns-speak.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-warn on node version greater than current LTS
+A warning is now shown when a node version greater than the current LTS is used.
--- a/packages/hardhat-core/src/internal/cli/cli.ts
+++ b/packages/hardhat-core/src/internal/cli/cli.ts
@@ -41,14 +41,14 @@ async function printVersionMessage(packageJson: PackageJson) {
   console.log(packageJson.version);
 }
 
-function printWarningAboutNodeJsVersionIfNeceesary(packageJson: PackageJson) {
+function printWarningAboutNodeJsVersionIfNecessary(packageJson: PackageJson) {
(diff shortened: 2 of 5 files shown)


alcuadrado:
LGTM except that we should reword [this](https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/cli/cli.ts#L51) to something like "Please, make sure you are using a supported version of Node.js."

alcuadrado (approved):

.changeset/rude-suns-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+warn on node version greater than current LTS
```
fvictorio, line 5:
```suggestion
A warning is now shown when a node version greater than the current LTS is used.
```

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -41,14 +41,14 @@ async function printVersionMessage(packageJson: PackageJson) {
...
   const requirement = packageJson.engines.node;
   if (!semver.satisfies(process.version, requirement)) {
     console.warn(
       chalk.yellow(
         `You are using a version of Node.js that is not supported by Hardhat, and it may work incorrectly, or not work at all.
 
-Please, upgrade your Node.js version.
+Please, make sure you are using a supported version of Node.js.
 
 To learn more about which versions of Node.js are supported go to https://hardhat.org/nodejs-versions`
```
feuGeneA, line 53: This isn't a comment on this PR per se, but I thought this was worth raising: I noticed that that link redirects to the higher-level heading "Stability Guarantees", but I wonder if it should instead go to the lower-level heading "Node.js versions support" (https://hardhat.org/reference/stability-guarantees.html#node-js-versions-support). I have no idea where those redirects are encoded... @alcuadrado ?

fvictorio: I agree with this. Added a commit with a better redirect.

feuGeneA:
This ready to merge?

morgansliman:
> This ready to merge?

I believe so. Rerunning that one (commonly) failed test and then I'll go ahead and merge.

PR #2281, opened 2022-01-21
Increase test timeout in truffle plugins

fvictorio:
Checking if this makes the `hardhat-truffle` tests fail less often in the CI when macos is used.


fvictorio:
There's definitely something going on with github's macos environment. See https://github.com/actions/virtual-environments/issues/4896

I guess we can just ignore those errors for now, as we've been doing lately.

PR #2280, opened 2022-01-21
add empty .prettierc in sample projects to match local solhint config

dievardump:
- [x] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.

---

Without .prettierrc, IDE's Prettier plugin will use the user default configuration.

User's default configuration sometimes conflicts with the sample projects solhint configuration.

Adding an empty .prettierrc fixes this conflict and ensures that prettier will follow local rules

PR #2286, opened 2022-01-21
Rm .prettierrc from advanced-ts sample project

feuGeneA:
It was added in #2280 to both the advanced project and the advanced
TypeScript project, but it doesn't need to also be in the TS project
because the TS project is overlaid on top of the files in the advanced
project; see
https://github.com/nomiclabs/hardhat/blob/217b6e45c54496735f1fbf111fc65ecaef0c51e8/packages/hardhat-core/src/internal/cli/project-creation.ts#L151-L154


fvictorio (approved):

fvictorio:
Oops, I missed that when I reviewed that PR, sorry.

Issue #2299, opened 2022-01-25
Add ReactJS project template


JaisPiyush:
# Limitations
Creating a hardhat project with frontend framework requires to setup workspaces and write all `babel` and bundler configurations from scratch. 

# Proposal
Introduce new `ReactJS` project template with complete typescript integration, within in the same package.json allowing easy management of project.

PR #2310, opened 2022-01-28
Added ReactJS project setup template with complete typescript support

JaisPiyush:
- [x] Because this PR includes a **new feature**, the change was previously discussed on an Issue.

---

**Issue #2299** 

# Limitations
Creating a hardhat project with frontend framework requires to setup workspaces and write all `babel` and `bundler` configurations from scratch.

# Changes
- Added complete `webpack`, `babelrc`, `eslint`, `react-refresh` and `typescript` configurations.
- `Dev` or `Prod` server can be spun using commands `{npm|yarn} run dev` or `{npm|yarn| run prod`
- `HRE ethers` of hardhat is not browser compatible, thus the frontend is entirely using `ethersjs`

## Filte Structure
```
root
  --- contracts
  --- tests
  --- scripts
  --- public
  --- web # contains all front end files
```


packages/hardhat-core/src/internal/cli/project-creation.ts
```diff
@@ -76,6 +82,37 @@ const ADVANCED_TYPESCRIPT_SAMPLE_PROJECT_DEPENDENCIES: Dependencies = {
...
+  "@types/react-dom": "^17.0.11",
+  "css-loader": "^6.5.1",
+  "ethers": "^5.5.3",
+  "html-webpack-plugin": "^5.5.0",
+  "react-refresh": "^0.11.0",
+  "webpack": "^5.66.0",
+  "webpack-cli": "^4.9.1",
+  "webpack-dev-server": "^4.7.3",
+  "webpack-merge": "^5.8.0",
+  "style-loader": "^3.3.1",
```
feuGeneA, line 102: Thank you for the contribution offer!

I haven't thoroughly reviewed the code, but I have an initial question: Does this tool chain have some precedence "in the wild"?

I'm the one who created the advanced projects, and a driving factor in that creation was the selection of the tools, and that selection was driven by how we saw Hardhat being used in real, active, successful products.  From the real world usage we observed, we derived what felt like the community's "best practices."  These bundles of tools are decidedly opinionated, but it's important to us that it's the _community's_ opinion, not ours.

Can you cite any known codebases that incorporate both Hardhat and a front-end?  I'll share with you below the projects that I surveyed as the basis for the current advanced project templates, in case you want to look there.  From (vague) memory, I believe that MyCrypto has a front-end in their codebase alongside Hardhat, but I don't remember any others.

* https://github.com/paulrberg/solidity-template/blob/main/package.json
* https://github.com/rhlsthrm/typescript-solidity-dev-starter-kit/blob/master/package.json
* https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/package.json
* https://github.com/MyCryptoHQ/MyCrypto/blob/master/package.json
* Aragon
* https://github.com/Synthetixio/synthetix/blob/develop/package.json
* https://github.com/celo-org/optics-monorepo/blob/b110e6e145df262c7f25fe896d25833426c5ae6f/solidity/optics-core/package.json
* https://github.com/Instadapp/dsa-connectors
* https://github.com/MolochVentures/moloch/blob/4e786db8a4aa3158287e0935dcbc7b1e43416e38/package.json
* kleros
* https://github.com/celer-network/rollup-contracts/blob/8a1d735cb4af3aa557d106701a73e65ff7a22f00/package.json
* https://github.com/aave/protocol-v2/blob/53b27235ad2e1501854adcde72bf3b835bbe4def/package.json
* https://github.com/UMAprotocol/protocol
* https://github.com/mstable/mStable-contracts/blob/master-v2/package.json

JaisPiyush: Thank you for looking after this ambitious project!

I understand your point and I am delighted  that you are concerned about "proof-of-community-interest".  Considering the HRE, the vision for hardhat's inception is understandable. Likewise truffle, intermediate or beginner programmers like me are trying to setup one monolithic project containing backend and frontend. 

Several projects can be found on https://github.com/search?p=3&q=hardhat+%2B+react&type=Commits, a simple github search. Although the projects on the link are not big or established as mentioned by you, but community is trying something.

feuGeneA: Thanks for the response.

To clarify, it's not about proof of community _interest_. Rather, it's about proof of established community _best practices_.

I think that the template menu should only offer "tried and true" solutions.  I'm thrilled to hear that the community is trying, and actually doing, integrations of both front-end and back-end within Hardhat projects, but unless/until those attempts converge at some set of "best practices", my opinion is that we should leave it out of that menu.  If we were to add in such an experimental thing, it would create a maintenance liability for us, and we already have limited bandwidth as it is :)

To clarify again, I am open to changes like this, but only using established best practices. Does such a set of best practices already exist? I have no idea! If you can demonstrate (via citation) that such best practices do exist, then we may proceed with changes like this*, but if they don't exist yet, then I think we should wait for them.

\* (I mean that _you_ may proceed with this PR, and we will (as our time permits) respond and review. We do not have the bandwidth to take on work like this this for ourselves at this time.)

alcuadrado:
Hey @JaisPiyush! Thanks for your interest in contributing to Hardhat. Unfortunately, this is not something we are looking to add any time soon.

I think it would be better if you create a template repository instead.

These are two that come to my mind:

https://github.com/paulrberg/solidity-template
https://github.com/pcaversaccio/hardhat-project-template-ts

PR #2318, opened 2022-01-31
Add artifacts extensibility support

Pzixel:

---

I hit this today when I was trying to make it work well with `TypeChain`. The problem is that current declaration can't be extended. Imagine you would like to augment `artifacts` with knowledge that it has some `IERC20` artifact which corresponds to an interface. It won't work today:

```ts 
declare module "hardhat/types/artifacts" {
    interface Artifacts {
        require(name: "IERC20"): IERC20Contract; // Duplicate identifier 'require'.ts(2300)
        // or
        require: (name: "IERC20") => IERC20Contract; // Subsequent property declarations must have the same type.  Property 'require' must be of type '(name: string) => any', but here has type '(name: "IERC20") => IERC20Contract'.ts(2717)
    }
}
```

But if we change it the way I propose it will work, and users will have a nice help from compiler and type inference:

### DIFF
--- a/packages/hardhat-truffle4/src/type-extensions.ts
+++ b/packages/hardhat-truffle4/src/type-extensions.ts
@@ -4,7 +4,7 @@ import "hardhat/types/runtime";
 
 declare module "hardhat/types/artifacts" {
   export interface Artifacts {
-    require: (name: string) => any;
+    require(name: string): any;
   }
 }
 declare module "hardhat/types/runtime" {
--- a/packages/hardhat-truffle4/test/types.ts
+++ b/packages/hardhat-truffle4/test/types.ts
@@ -1,0 +1,25 @@
+import {artifacts} from "hardhat";
+
+type Dummy = unknown;
+
+type IERC20Contract = Dummy;
+
+type Equals<X, Y> =
+    (<T>() => T extends X ? 1 : 2) extends
+    (<T>() => T extends Y ? 1 : 2) ? true : false;
+
+type Assert<_X extends true> = void;
+
+// assert Artifacts can be extended
+
+declare module "hardhat/types/artifacts" {
+  export interface Artifacts {
+    require(name: 'IERC20'): IERC20Contract;
+  }
+}
+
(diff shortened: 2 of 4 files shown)


Pzixel:
Ping?

alcuadrado:
Hey @Pzixel,

Sorry for the delay!

Can you explain the intention of this PR? I'm not sure if I get it.

Pzixel:
See tests `packages/hardhat-truffle4/test/types.ts` they are kinda self-explanatory. This is an expected use-case.

In its current form `artifacts` is a property and it cannot be augmented in any way. So we're stuck witn `string -> any`.
But if it's a function on interface then we can add some overloads in user code, e.g.: 

```ts
declare global {
  namespace Truffle {
    interface Artifacts {
      require(name: "Ownable"): OwnableContract;
      require(name: "IERC20"): IERC20Contract;
      require(name: "IUniswapV3SwapCallback"): IUniswapV3SwapCallbackContract;
      require(name: "IUniswapV3Pool"): IUniswapV3PoolContract;
      ....
  }
}
```

so when we write:
```ts
const IERC20 = artifacts.require('IERC20');
```
TS will infer `IERC20Contract` type here. But it requires `Artifacts` in `module "hardhat/types/artifacts"` to be exensible.

alcuadrado:
I see! Yes, this is a great idea! We do something similar in the ethers plugin. Thanks!

Do you mind taking a look at the CI as it's failing?

Pzixel:
Sure, I'l see how to fix it. IT was passing lints when I posted first time so I though I'm ok.

I'm glad you see this as a worthy initiative.

alcuadrado:
Yeah, it's great actually. Take look at this: https://github.com/dethcrypto/TypeChain/issues/370 — You may want to do the same for truffle

Pzixel:
I was going to enhance typechain generation but this problem was a blocker. I'm not quite sure if it's what you do mean

Pzixel:
Ping?

alcuadrado:
Thanks @Pzixel!

PR #2319, opened 2022-02-01
Bump solpp to 0.11.5 in hardhat-solpp

davidtperk:
This PR updates solpp in hardhat-solpp to 0.11.5, which fixes https://github.com/nomiclabs/hardhat/issues/1412.


fvictorio:
Hey @davidtperk, thanks for sending this and sorry for not reviewing it before! For security reasons, we don't commit `yarn.lock` changes from external contributors, so I submitted #2412 to fix this problem.

PR #2333, opened 2022-02-04
Disable macos tests in the CI

fvictorio:
We'll re-enable them when https://github.com/actions/virtual-environments/issues/4896 is fixed.


fvictorio:
Leaving the `trigger ci` commit to show how this works, will drop it before merging.

alcuadrado (approved):
Nice approach

PR #2334, opened 2022-02-04
instead of privateKey, it should be public key

heeropunjabi:

---


alcuadrado:
Thanks for submitting this PR, but this change is incorrect. Those are effectively private keys.

PR #2359, opened 2022-02-09
Add `--parallel`, `--bail` and `--grep` params to the test task

fvictorio:
Add support for the `--parallel`, `--bail` and `--grep` flags in the test task.

The first commit is just the mocha upgrades, the second one adds the `--parallel` flag. Notice that this flag also adds `hardhat/register` to Mocha's required files. This makes parallel tests mostly work out-of-the-box, but there might be a risk here that I'm not seeing.

Pending stuff:

- [x] Check if `hardhat-gas-reporter` and `solidity-coverage` work on parallel mode. Document how to make them work if not, or just document the limitation if making them work is too much effort.
- [x] Run tests in parallel mode in the e2e tests
- [x] Add `--bail` and `--grep` flags
- [x] Document this mode, its limitations, and tips.
- [x] Add changeset.


fvictorio:
`solidity-coverage` works fine because it has its own `coverage` task.

`hardhat-gas-reporter` doesn't work with parallel tests. I created [an issue](https://github.com/cgewecke/hardhat-gas-reporter/issues/98) about it, but we should also document that and also be explicit about this in the release notes.

docs/guides/waffle-testing.md
```diff
@@ -125,6 +125,23 @@ expect(await greeter.greet()).to.equal("Hola, mundo!");
...
+
+You can run your tests in parallel by using the `--parallel` flag:
+
+```
+$ npx hardhat test --parallel
+```
+
+Most of the time, running your tests serially or in parallel should produce the same results, but there are some scenarios where tests run in parallel will behave differently:
+
+- In serial mode, all the test files share the same instance of the [Hardhat Runtime Environment](/advanced/hardhat-runtime-environment.html), but in parallel mode each test file will have its own fresh instance. For example, if one test file deploys a contract, then that contract won't exist in the other test files.
```
feuGeneA, line 138:
```suggestion
- In serial mode, all the test files share the same instance of the [Hardhat Runtime Environment](/advanced/hardhat-runtime-environment.html), but in parallel mode each test file will have its own fresh instance. For example, if one test file deploys a contract, then that deployment won't exist when running the other test files.
```

docs/guides/waffle-testing.md
```diff
@@ -125,6 +125,23 @@ expect(await greeter.greet()).to.equal("Hola, mundo!");
...
+- The `.only` modifier doesn't work in parallel mode. You can use [`--grep`](https://mochajs.org/#-grep-regexp-g-regexp) to run specific tests.
```
feuGeneA, line 139:
```suggestion
- The `.only` modifier doesn't work in parallel mode. As an alternative, you can use [`--grep`](https://mochajs.org/#-grep-regexp-g-regexp) to run specific tests.
```

docs/guides/waffle-testing.md
```diff
@@ -125,6 +125,23 @@ expect(await greeter.greet()).to.equal("Hola, mundo!");
...
+- The duration of individual tests might be different, because parallel mode uses more system resources. There's a chance that some tests start timing out for that reason. You can increase the tests timeout if you run into this problem.
```
feuGeneA, line 140:
```suggestion
- Because parallel mode uses more system resources, the duration of individual tests might be longer, so there's a chance that some tests start timing out for that reason. You can increase the tests timeout if you run into this problem.
```

packages/hardhat-core/src/builtin-tasks/test.ts
```diff
@@ -1,4 +1,5 @@
 import chalk from "chalk";
+import type { MochaOptions } from "mocha";
```
morgansliman, line 2: Personally I always like type-exclusive imports to be first, but that's subjective and certainly not a blocker for merging since I don't think it's a standard we adhere to across the codebase.

packages/hardhat-core/src/builtin-tasks/test.ts
```diff
@@ -43,24 +44,59 @@ subtask(TASK_TEST_GET_TEST_FILES)
...
+    ) => {
+      const { default: Mocha } = await import("mocha");
 
-    const testFailures = await new Promise<number>((resolve) => {
-      mocha.run(resolve);
-    });
+      const mochaConfig: MochaOptions = {
+        ...config.mocha,
+        bail,
+        grep,
```
morgansliman, line 75: Should/do we allow these to be configured in `config.mocha` as well? If so, this would overwrite them (i.e. if they were set in `config.mocha` but not passed as params to the subtask).

fvictorio: Oh, I hadn't noticed that, thanks for pointing it out. I actually think that would be the correct behavior though? Normally CLI flags override what's in the config (for example, that's what Prettier does by default). But it's good to know that this is the case. I will add a comment explaining that this is intentional.

fvictorio: Wait I see what you mean. For the `grep` flag this works correctly, but `bail` has a default value, and so it will always override whatever is in the config. I'll fix this, thank you.

docs/guides/waffle-testing.md
```diff
@@ -125,6 +125,23 @@ expect(await greeter.greet()).to.equal("Hola, mundo!");
 
 We can modify the state of a contract in the same way we read from it. Calling `setGreeting` will set a new greeting message. After the `Promise` is resolved, we perform another assertion to verify that the greeting change took effect.
 
+### Running tests in parallel
```
alcuadrado, line 128: I'd move this to its own guide, to make it more visible.

Also, this is not a waffle thing.

packages/hardhat-core/src/builtin-tasks/test.ts
alcuadrado: That's a nice idea. Can we add a linter rule for it?

packages/hardhat-core/src/builtin-tasks/test.ts
alcuadrado: Great catch! I was also tricked by this

packages/hardhat-core/src/builtin-tasks/test.ts
```diff
@@ -80,14 +116,23 @@ task(TASK_TEST, "Runs mocha tests")
     []
   )
   .addFlag("noCompile", "Don't compile before running this task")
+  .addFlag("parallel", "Run tests in parallel")
+  .addFlag("bail", "Stop running tests after the first test failure")
+  .addOptionalParam("grep", "Only run tests matching the given regexp")
```
alcuadrado, line 121:
```suggestion
  .addOptionalParam("grep", "Only run tests matching the given string or regexp")
```

packages/hardhat-core/src/builtin-tasks/test.ts
```diff
@@ -43,24 +44,59 @@ subtask(TASK_TEST_GET_TEST_FILES)
 subtask(TASK_TEST_SETUP_TEST_ENVIRONMENT, async () => {});
 
 subtask(TASK_TEST_RUN_MOCHA_TESTS)
+  .addFlag("parallel", "Run tests in parallel")
+  .addFlag("bail", "Stop running tests after the first test failure")
+  .addOptionalParam("grep", "Only run tests matching the given regexp")
```
alcuadrado, line 49:
```suggestion
  .addOptionalParam("grep", "Only run tests matching the given string or regexp")
```

alcuadrado: That's closer to `mocha --help`'s wording

alcuadrado:
As discussed offline, this PR looks great, but we should test that the params actually enable those features. @fvictorio suggested using a fixture project for that.

docs/guides/waffle-testing.md
fvictorio: Applied the suggestion and added some links about increasing timeouts. Check fe479e063ef253c8180b2eab7ab9be86177d427d.

docs/guides/waffle-testing.md
fvictorio: Done.

packages/hardhat-core/src/builtin-tasks/test.ts
fvictorio: Fixed.

fvictorio:
@alcuadrado this is ready for review.

fvictorio:
We could do something like [this](https://github.com/serverless/test/blob/d27d35e94a5814a57587435a520d7d53f13e056e/setup/patch.js#L42-L54).

alcuadrado (approved):
LFG!

PR #2366, opened 2022-02-10
Better support for running full block tests

fvictorio:
This PR:

- Fixes the `test-run-forked-block.ts`. Now both that script and the full block test in `provider/node` use the same underlying helper function.
- Adds a `test-recent-mainnet-block.ts` script, that tests a recent block (20 blocks ago) and assumes the `ALCHEMY_URL` variable is set.

The recent mainnet block script could be something we run periodically in a GH action.


packages/hardhat-core/scripts/test-run-forked-block.ts
```diff
@@ -68,12 +9,20 @@ if (rpcUrlArg === undefined) {
...
   console.warn("No blockNumber given");
   usage();
 }
 
-main(rpcUrlArg, +blockNumberArg)
+async function main(rpcUrl: string, blockNumber: number, chainId: number) {
+  const remoteCommon = new Common({ chain: chainId });
+  const hardfork = remoteCommon.getHardforkByBlockNumber(blockNumber);
+
+  await runFullBlock(rpcUrlArg, +blockNumberArg, chainId, hardfork)
```
alcuadrado, line 22:
```suggestion
  await runFullBlock(rpcUrlArg, blockNumber, chainId, hardfork)
```

alcuadrado (approved):
LGTM except for a minor comment

fvictorio:
@alcuadrado This PR includes the changes from https://github.com/nomiclabs/hardhat/pull/2367. I merged that branch to make sure that this works with the latest versions of ethereumjs. But I can undo that merge if you prefer to merge those PRs separately.

PR #2370, opened 2022-02-11
Remove `node-fetch` everywhere, in favor of `undici`

feuGeneA:
Most things "just worked" when switching libraries, with the exception of the compiler downloader, `download.ts`. Ultimately I found that by requesting the compiler downloads via an `http://` url rather than an `https://` one, the tests work, so I made that change.

- [x] Migrate all uses of `node-fetch` to `undici`.
- [x] Fix the problem with `download.ts`.
- [x] Remove `node-fetch` from `package.json`.


packages/hardhat-core/src/internal/util/download.ts
```diff
@@ -1,14 +1,8 @@
 import fs from "fs";
 import fsExtra from "fs-extra";
-import HttpsProxyAgent from "https-proxy-agent";
```
alcuadrado, line 3: Shouldn't we remove this dependency from the package.json?

packages/hardhat-core/src/internal/util/download.ts
```diff
@@ -27,47 +21,48 @@ export async function download(
...
-    agent: undefined,
-  };
 
-  // Check if Proxy is set https
-  if (process.env.HTTPS_PROXY !== undefined) {
-    // Create the proxy from the environment variables
-    const proxy: string = process.env.HTTPS_PROXY;
-    fetchOptions.agent = new HttpsProxyAgent.HttpsProxyAgent(proxy);
-  }
+  const dispatcher = (() => {
```
alcuadrado, line 27: I don't understand this pattern

packages/hardhat-core/src/internal/util/download.ts
```diff
@@ -27,47 +21,48 @@ export async function download(
...
 
   // Fetch the url
-  const response = await fetch(url, fetchOptions);
+  const response = await request(url, {
+    dispatcher,
+    headersTimeout: timeoutMillis,
+    maxRedirections: 10,
+    method: "GET",
+    headers: {
+      "User-Agent": "hardhat",
```
alcuadrado, line 46: Nice!

What if we add the version number here? It can be useful for debugging.

packages/hardhat-core/test/internal/util/download.ts
```diff
@@ -51,7 +51,7 @@ describe("Compiler List download with proxy", function () {
     it("Should call download with the right params", async function () {
       const compilersDir = this.tmpDir;
       const downloadPath = path.join(compilersDir, "downloadedCompilerProxy");
-      const expectedUrl = `https://solc-bin.ethereum.org/wasm/list.json`;
+      const expectedUrl = `http://solc-bin.ethereum.org/wasm/list.json`;
```
alcuadrado, line 54: I suspect that this is hiding a bug. Note that this is testing an https proxy.

alcuadrado: @fvictorio you mentioned something about this.

alcuadrado:
I left a few minor comments, but this looks good.

The only thing I don't really like is that there's a repeated check for the status code being a 2xx.

packages/hardhat-core/src/internal/util/download.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2370/commits/9ae76438d6f97ebcd43e0576d1df4b05673ab755

packages/hardhat-docker/src/hardhat-docker.ts
```diff
@@ -76,14 +76,18 @@ export class HardhatDocker {
...
-      return res.ok;
+      const { request } = await import("undici");
+      const res = await request(imageEndpoint, { method: "GET" });
+
+      // "The Fetch Standard allows users to skip consuming the response body
+      // by relying on garbage collection to release connection resources.
+      // Undici does not do the same. Therefore, it is important to always
+      // either consume or cancel the response body."
+      // https://undici.nodejs.org/#/?id=garbage-collection
+      // It's not clear how to "destroy", so we'll just consume:
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2414/commits/8c93508d795190f8652eb906cacada9c1c5b12a1

packages/hardhat-core/src/internal/util/download.ts
alcuadrado: Let's ignore this for now. I thought this was the HTTP provider, which would be much more useful.

We can add that in a future PR

packages/hardhat-core/src/internal/util/download.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2370/commits/fd1528fdcf62a14950baf21206c21f3845133d8d

packages/hardhat-core/src/internal/util/download.ts
feuGeneA: Had already drafted it, so pushed it up. Addressed in https://github.com/NomicFoundation/hardhat/pull/2370/commits/fd875048a9437dea109e654b44331679d7d3e45e

packages/hardhat-core/test/internal/util/download.ts
feuGeneA: If I change it back to `https`, I get the error message `Error: Failed to download https://solc-bin.ethereum.org/wasm/list.json - 400 received. Only "http:" protocol prefix is supported`. That message appears to come from the npm package `proxy` that we're using for the test, which has: 
```
		if ('http:' != parsed.protocol) {
			// only "http://" is supported, "https://" should use CONNECT method
			res.writeHead(400);
			res.end('Only "http:" protocol prefix is supported\n');
			return;
		}
```
(https://github.com/TooTallNate/proxy/blob/d677ef31fd4ca9f7e868b34c18b9cb22b0ff69da/proxy.js#L213-L218)

Who is it that should be using `CONNECT`? I would guess that it should be `undici`, specifically its `ProxyAgent` that we're using.

alcuadrado: Let's defer this to a future PR. I created this issue: https://github.com/NomicFoundation/hardhat/issues/2418

alcuadrado (approved):

packages/hardhat-core/src/internal/util/download.ts
```diff
@@ -27,47 +25,48 @@ export async function download(
...
-  // Consume the response stream and discard its result
-  // See: https://github.com/node-fetch/node-fetch/issues/83
-  const _discarded = await response.arrayBuffer();
-
   // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
   throw new Error(
-    `Failed to download ${url} - ${response.statusText} received`
+    `Failed to download ${url} - ${
+      response.statusCode
+    } received. ${await response.body.text()}`
```
fvictorio, line 70: If the `if` was not entered because `response.body` was null, this will throw while building the error message.

PR #2372, opened 2022-02-11
Parallel compilation for solidity

morgansliman:
- parallel pre-downloads for compilers
- parallelizes compilation jobs scaled to number of cpu cores
- aggregates and logs all encountered errors, as opposed to only the first one thrown


packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -1257,8 +1272,12 @@ subtask(TASK_COMPILE_SOLIDITY_LOG_COMPILATION_RESULT)
   .addParam("quiet", undefined, undefined, types.boolean)
   .setAction(
     async ({ compilationJobs }: { compilationJobs: CompilationJob[] }) => {
-      if (compilationJobs.length > 0) {
-        console.log("Solidity compilation finished successfully");
+      const count = compilationJobs.length;
+
+      if (count > 0) {
+        console.log(
+          `Compiled ${count} Solidity ${pluralize(count, "file")} successfully`
```
fvictorio, line 1279: This is printing the number of compilation jobs, not the number of compiled files.

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -379,38 +393,87 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE_JOBS)
...
-        artifactsEmittedPerJob.push({
-          compilationJob,
-          artifactsEmittedPerFile,
-        });
+      const { compilationSuccesses, compilationFailures } =
+        filterCompilationResults(results);
+
+      if (compilationFailures.length > 0) {
+        for (const { errors } of compilationFailures) {
+          console.log(errors);
```
fvictorio, line 463: This is logging the raw JSONs.

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -379,38 +393,87 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE_JOBS)
...
+      const downloader = new CompilerDownloader(compilersCache);
+      await downloader.downloadCompilersList();
+
+      /**
+       * This is a pretty hacky solution to avoid a major refactor and/or breaking changes.
+       * The hack is the fact that we're just pre-downloading all of the necessary compilers
+       * before actually running the job, so all the later code assumes the compiler is already
+       * downloaded because if it wasn't then parallel compilation will break.
+       */
+      await chunkedPromiseAll(
```
fvictorio, line 435: This should be done serially. Otherwise, there are some scenarios where the same version is downloaded at the same time and it results in an error.

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -379,38 +393,106 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE_JOBS)
...
-            compilationJob,
-            compilationJobs: sortedCompilationJobs,
-            compilationJobIndex: i,
-            quiet,
+      const results = await chunkedPromiseAll<
+        CompilationSuccess | CompilationFailure
+      >(
+        compilationJobs.map((compilationJob, compilationJobIndex) => {
+          return () =>
+            run(TASK_COMPILE_SOLIDITY_COMPILE_JOB, {
```
fvictorio, line 457: There is a subtle problem here related to `chunkedPromiseAll` swallowing the errors. If `TASK_COMPILE_SOLIDITY_COMPILE_JOB` throws for some reason (say, the compiler process crashes or whatever), then the code will try to iterate over the error and throw an error like this:

```
An unexpected error occurred:

TypeError: errors is not iterable
```

You can check this by creating a test project and modifying the subtask to just throw some error.

Maybe we don't have to swallow the errors in `chunkedPromiseAll`? For the first usage (the downloads), that would let us remove the for-loop that checks if some result is an error. And for the second usage (the compilations), we don't throw expected errors, so we don't really need to swallow them.

This would also help simplify the code of `chunkedPromiseAll`.

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -379,38 +393,106 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE_JOBS)
...
-            job2.getSolcConfig().version
+      log(`Compiling ${compilationJobs.length} jobs`);
+
+      const versionList: string[] = [];
+      for (const job of compilationJobs) {
+        const solcVersion = job.getSolcConfig().version;
+
+        // versions older than 0.4.11 don't work with hardhat
+        // see issue https://github.com/nomiclabs/hardhat/issues/2004
+        if (semver.lt(solcVersion, COMPILE_TASK_FIRST_SOLC_VERSION_SUPPORTED)) {
```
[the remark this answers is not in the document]
morgansliman: ok, I was on the fence about which would be faster since moving it inside would mean an extra array search for invalid versions.

packages/hardhat-core/src/builtin-tasks/compile.ts
morgansliman: The entire point of "swallowing" the errors in `chunkedPromiseAll` is so that we can log an aggregate of the errors at the end. If I remove that logic, then it'll just throw the first error encountered like it currently does.

packages/hardhat-core/test/internal/util/chunked-promise-all.ts
```diff
@@ -3,7 +3,7 @@ import sinon from "sinon";
 
 import { chunkedPromiseAll } from "../../../src/internal/util/chunked-promise-all";
 
-describe("chunked promise all", () => {
+describe.only("chunked promise all", () => {
```
fvictorio, line 6:
```suggestion
describe("chunked promise all", () => {
```

packages/hardhat-core/test/internal/util/chunked-promise-all.ts
```diff
@@ -0,0 +1,132 @@
+import { assert } from "chai";
+import sinon from "sinon";
+
+import { chunkedPromiseAll } from "../../../src/internal/util/chunked-promise-all";
+
+describe.only("chunked promise all", () => {
```
fvictorio, line 6:
```suggestion
describe("chunked promise all", () => {
```

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -426,36 +430,41 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE_JOBS)
...
-          return () =>
-            run(TASK_COMPILE_SOLIDITY_COMPILE_JOB, {
-              compilationJob,
-              compilationJobs,
-              compilationJobIndex,
-              quiet,
-            });
-        }),
-        os.cpus().length
+      if (downloadResults instanceof Error) {
```
fvictorio, line 451: I don't think this is necessary, since `p-map` will reject with the aggregate error (not return it).

.changeset/healthy-actors-hunt.md
```diff
@@ -0,0 +1,6 @@
+---
+"@nomiclabs/hardhat-e2e-tests": patch
```
alcuadrado, line 2:
```suggestion
```

We need to remove this or changesets will fail

packages/hardhat-core/src/builtin-tasks/compile.ts
alcuadrado: Yes, according to the docs it rejects with that error, hence the `await` throws

.changeset/healthy-actors-hunt.md
morgansliman: that's right, I always forget that

packages/hardhat-core/src/builtin-tasks/compile.ts
morgansliman: Oh I see, I misread the docs

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -379,39 +381,105 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE_JOBS)
...
-        const compilationJob = sortedCompilationJobs[i];
+      /**
+       * I'm not super proud of this, but we have to download the list before the compilers,
+       * otherwise the next downloads fail with a race condition.
+       *
+       * :)
+       */
+      const compilersCache = await getCompilersDir();
+      const downloader = new CompilerDownloader(compilersCache);
+      await downloader.downloadCompilersList();
```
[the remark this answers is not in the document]
alcuadrado: This always downloads the compiler list. This makes the compilation slower and prevents it from working offline.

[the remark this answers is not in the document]
alcuadrado: What if we remove this and don't download the compilers in parallel?

alcuadrado (changes requested):
I left a few minor comments, but I think this is almost ready.

packages/hardhat-core/package.json
```diff
@@ -126,6 +127,7 @@
     "mnemonist": "^0.38.0",
     "mocha": "^7.2.0",
     "node-fetch": "^2.6.0",
+    "p-map": "^4.0.0",
```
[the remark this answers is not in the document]
morgansliman: https://github.com/sindresorhus/p-map/issues/44#issuecomment-892730191

packages/hardhat-core/src/builtin-tasks/compile.ts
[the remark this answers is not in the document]
morgansliman: I made this point to Franco as well, but I think if we're making changes to speed up compilation anyways, then ensuring it's as fast as currently possible (i.e. parallel downloads) should be the focus.

That being said, you're right that this would prohibit offline compilation. I would suggest instead that we just catch this one error and let the code continue. If they already have the necessary compilers downloaded and would otherwise be able to compile offline, then it'll just work, and if not then it'll fail in the next download attempt as usual.

packages/hardhat-core/package.json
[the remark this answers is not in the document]
alcuadrado: Ah, yes. Sindre is doing that annoying thing.

packages/hardhat-core/src/builtin-tasks/compile.ts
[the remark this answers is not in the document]
alcuadrado: But I think downloading this here will, in 99% of the cases, make the compilation slower vs just trying to download the compilers serially and removing this logic. The reason is that most of the time the task that tries to download the compiler won't do anything, as they'll already be downloaded.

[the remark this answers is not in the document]
morgansliman: What if we add a check here to download only if it doesn't exist already? Like the one that already exists in the Downloader class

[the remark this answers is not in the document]
fvictorio: @alcuadrado found a couple of extra issue related to downloading the compilers in parallel.

The first one has to do with wasm compilers: if the native compilers are not working for some reason (for example, permissions issues), then the compilation jobs will try to download the wasm compilers. If the same version is downloaded in parallel, we get the original race condition again.

This can be fixed by also downloading the wasm list beforehand. And we need to check in both cases that they don't exist before downloading them for compilation to work offline, and to not add unnecessary download times to the compilation. At this point, and considering that downloading the compilers in parallel is better only in very specific cases (when you are compiling for the first time a project that uses multiple compilers), we think it makes more sense to do the simple thing and download them sequentially.

Btw, even if we would do all of that, there's still another problem: if someone uses a couple of new versions of the compiler, but the list already exists locally, then when the `GET_SOLC_BUILD` subtask is executed in parallel, the list will be downloaded in parallel, and then we get the same race condition again. This is a very edge case, but it's not impossible.

alcuadrado (approved):

packages/hardhat-core/src/builtin-tasks/compile.ts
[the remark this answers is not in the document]
alcuadrado: > for example, permissions issues
The most common example would be running under a different architecture on linux, which got quiet common because of Docker and M1s.

PR #2374, opened 2022-02-11
feat(hardhat-ganache): update ganache-core to ganache

LuisUrrutia:
The package hardhat-ganache uses an old version of ganache and I did the update from ganache-core to ganache 7, but I'm having problem with some tests, so I make this PR in order to look for help too

### DIFF
--- a/packages/hardhat-ganache/src/ganache-service.ts
+++ b/packages/hardhat-ganache/src/ganache-service.ts
@@ -1,48 +1,106 @@
 import debug from "debug";
 import { NomicLabsHardhatPluginError } from "hardhat/internal/core/errors";
 import { URL } from "url";
+import Ganache from "ganache";
 
 const log = debug("hardhat:plugin:ganache-service");
 
 declare interface GanacheOptions {
-  url: string;
-  keepAliveTimeout?: number;
-  accountKeysPath?: string; // Translates to: account_keys_path
-  accounts?: object[];
-  allowUnlimitedContractSize?: boolean;
-  blockTime?: number;
-  dbPath?: string; // Translates to: db_path
-  debug?: boolean;
-  defaultBalanceEther?: number; // Translates to: default_balance_ether
-  fork?: string | object;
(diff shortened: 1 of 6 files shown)


fvictorio:
We'll have to look at this in more depth, but as a first comment: we use `yarn`, so the `package-lock.json` here shouldn't be included.

packages/hardhat-ganache/src/ganache-service.ts
```diff
@@ -221,12 +279,12 @@ export class GanacheService {
     }
 
     // Test for unsupported commands
-    if (options.accounts !== undefined) {
-      throw new NomicLabsHardhatPluginError(
-        "@nomiclabs/hardhat-ganache",
-        "Config: ganache.accounts unsupported for this network"
-      );
-    }
+    // if (options.wallet && options.wallet.accounts !== undefined) {
```
alcuadrado, line 282: Why is this commented out?

packages/hardhat-ganache/test/fixture-projects/hardhat-project-with-configs/hardhat.config.ts
```diff
@@ -4,14 +4,19 @@ export default {
...
+      wallet: {
+        defaultBalance: 9,
+        totalAccounts: 3,
+        mnemonic:
+          "polar velvet stereo oval echo senior cause cruel tube hobby exact angry",
+      },
     },
   },
-  solidity: "0.5.15",
+  solidity: "0.8.11",
```
alcuadrado, line 21: Keeping it to 0.5.15 makes the ci go faster. Can we revert this change?

alcuadrado (commented):
Wow! Nice PR, thanks. I just left a few minor comments.

Can you also remove the pacakge-lock? We use a yarn monorepo, not npm.

alcuadrado:
Also, can you allow us to modify your branch/pr so that we can add a changeset?

PR #2394, opened 2022-02-16
removed missed docker references

morgansliman:
- just removing a few places where docker was referenced in the context of being needed for vyper, since it no longer is needed
- also updated the github actions workflow for vyper to no longer pull the docker compiler unnecessarily


fvictorio (approved):

alcuadrado:
Nice!

PR #2398, opened 2022-02-17
TYPO console.log -> console.sol

wardu:
Change import instruction, from console.log to console.sol.

### DIFF
--- a/docs/tutorial/debugging-with-hardhat-network.md
+++ b/docs/tutorial/debugging-with-hardhat-network.md
@@ -4,7 +4,7 @@
 
 ## Solidity `console.log`
 
-When running your contracts and tests on **Hardhat Network** you can print logging messages and contract variables calling `console.log()` from your Solidity code. To use it you have to import **Hardhat**'s`console.log` from your contract code.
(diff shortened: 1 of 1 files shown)


docs/tutorial/debugging-with-hardhat-network.md
```diff
@@ -4,7 +4,7 @@
 
 ## Solidity `console.log`
 
-When running your contracts and tests on **Hardhat Network** you can print logging messages and contract variables calling `console.log()` from your Solidity code. To use it you have to import **Hardhat**'s`console.log` from your contract code.
+When running your contracts and tests on **Hardhat Network** you can print logging messages and contract variables calling `console.log()` from your Solidity code. To use it you have to import **Hardhat**'s`console.sol` from your contract code.
```
alcuadrado, line 7:
```suggestion
When running your contracts and tests on **Hardhat Network** you can print logging messages and contract variables calling `console.log()` from your Solidity code. To use it you have to import `hardhat/console.sol` from your contract code.
```

alcuadrado (commented):
Thanks for this PR. It was confusing, you are right.

What do you think about the suggestion I left?

Also, no need to create an issue for a PR this size :)

docs/tutorial/debugging-with-hardhat-network.md
wardu: Yes, looks good.
Yes, maybe overkill :) I'll remember for next time.

Issue #2115, opened 2021-11-30
Revert due to overflow misdetected as `CONTRACT_TOO_LARGE_ERROR` on a contract compiled with `viaIR: true`


cameel:
I'm trying to run OpenZeppelin tests using the new, experimental IR code generator of the Solidity compiler (`viaIR: true`) and I'm running into failures that do not happen when using the default code generator. I tracked it down to Hardhat's stack trace module misdetecting a revert caused by an integer overflow as `CONTRACT_TOO_LARGE_ERROR`.

I think that the most likely cause is that some the current heuristics make too strong assumptions about the output from the default code generator and need to be updated to handle the new one as well. I cannot also exclude the possibility that it's actually some kind of problem with compiler's output. Please check it out and let me know if there's actually something that needs to be fixed in the compiler.

### Repro
Just OpenZeppelin tests with version pragmas stripped from contracts and custom compiler settings:

```bash
git clone https://github.com/OpenZeppelin/openzeppelin-contracts.git
cd openzeppelin-contracts

cat <<EOF >> hardhat.config.js
module.exports = {
    solidity: {
        version: '0.8.10',
        settings: {
            viaIR: true,
            optimizer: {
                enabled: true,
                runs: 200,
            },
        },
    },
    networks: {
        hardhat: {
            blockGasLimit: 10000000,
            allowUnlimitedContractSize: false,
        },
    },
};
EOF

npm install
find . test -name '*.sol' -type f -print0 | xargs -0 sed -i -E -e 's/pragma solidity [^;]+;/pragma solidity *;/'
npm test test/token/ERC20/extensions/ERC20FlashMint.test.js
```
Note: `ERC20FlashMint.test.js` is not the only test that fails - 57 out of 751 tests fail if you run the full test suite - but it's the one I investigated so far.

#### Output
```
Compiling 219 files with 0.8.10
Compilation finished successfully


  Contract: ERC20FlashMint
    maxFlashLoan
      ✓ token match
      ✓ token mismatch
    flashFee
      ✓ token match
      ✓ token mismatch (44ms)
    flashLoan
      ✓ success (63ms)
      ✓ missing return value (51ms)
      ✓ missing approval (39ms)
      ✓ unavailable funds (51ms)
[...]

PR #2399, opened 2022-02-17
Minimally patch the viaIR issue for running unit tests

jflatow:
This minimally patches the viaIR issue for a particular case where we are encountering it.

I believe this is also addresses the cause of #2115.

PR #2400, opened 2022-02-17
Logically patch the viaIR issue for running unit tests

jflatow:
This attempts to patch the viaIR issue (as encountered in a particular case) in a logical way (as opposed to minimally).

I believe this is also addresses the cause of #2115.


alcuadrado:
Hey, thanks for submitting this patch.

Does this PR replace #2399?

Also, can you provide a plain-english explanation of the changes here?

jflatow:
This is an alternative to #2399, which I don't really know enough about the internals of hardhat to evaluate if its a good idea or not. But the issue in #2115 seems to be that the `sourceReference` is sometimes undefined in calls to `sourceReferenceToSolidityCallsite`, so this patch attempts to more rigorously ensure that it has a value rather than patching the case I was encountering. In this one I attempt to always use the same fallback logic of 'if source reference, convert to call site, otherwise build an unknown source reference', as opposed to having it be handled in a seemingly inconsistent way between stack entry types.

fvictorio:
Same comment as in [here](https://github.com/NomicFoundation/hardhat/pull/2399#issuecomment-1260723806). Closing in favor of #3205.

PR #2410, opened 2022-02-22
Re-enable macos jobs in the CI

fvictorio:


fvictorio:
The macos issue seems to have been fixed. I ran the full suite three times and it seems to work much better now. There still are some false negatives, but I don't think they are related to the macos thing.

alcuadrado (approved):

fvictorio:
I'm going to wait a bit before merging this until there's more information on this issue: https://github.com/actions/virtual-environments/issues/4896

fvictorio:
I'm going to close this, and set a reminder to look into this one month from now.

Issue #2305, opened 2022-01-26
ability to specify `--chain-id` to `node` command to allow


dbeal-eth:
It is currently not possible to specify the chain ID for a hardhat network invoked with `npx hardhat node` through any means other than the config file. This means it is impossible to specify the `--chain-id` argument in the command line, or execute a chain host pragmatically with `hre.run('node', { chainId: 100 })`.

Please add the `chainId` parameter to the `node` command so it is possible to override this value either pragmatically or case-by-case through the command line.

webdevabhijit:
@kanej can I work in this issue?

fvictorio:
@webdevabhijit sure! [This](https://github.com/NomicFoundation/hardhat/blob/5a7d04d9b75dc631b731ddb03f5e64b7556f90a9/packages/hardhat-core/src/builtin-tasks/node.ts#L247) is the entry point where that optional param should be added.

We don't have tests for the `node` task, and they would be tricky to do, so if the change is simple enough, I think we might be ok with not having tests for that change (not sure though).

webdevabhijit:
@fvictorio I have added the feature. Should I make a pull request?

fvictorio:
Sure.

immense055:
ac904898b4da93e40b83d7789582dcd2c1ba3a32

dbeal-eth:
this is still relevant.

brunoguerios:
Hi everyone! Is there an ETA on this feature?

goums:
would love to have this as well

fvictorio:
I looked a little into this and it's not as straightforward as I thought, because the underlying network is created before the task is executed.

The easiest way to do this is to waint until either https://github.com/NomicFoundation/hardhat/issues/1697 or https://github.com/NomicFoundation/hardhat/issues/3074 are available.

In the meantime, a workaround is to do this in your config:

```js
  networks: {
    hardhat: {
      chainId: Number(process.env.HARDHAT_CHAIN_ID ?? 31337)
    },
  },
```

and then start the node with `HARDHAT_CHAIN_ID=123 npx hardhat node`. Not ideal, but it should get the job done until we can add a proper param.

Samboy76:
Hi,

I´m facing a similar issue.

I am trying to fork 5 environments and config 5 mainnet networks within the same hardhat.config.js shown below:

    require('dotenv').config();
    require("@nomicfoundation/hardhat-toolbox");
    
    const FORK_MAINNET = true;
[...]

Issue #1697, opened 2021-07-20
Make `hardhat_reset` more powerful


fvictorio:
Ideally `hardhat_reset` would let you change anything that you can configure in the config of the Hardhat network. Right now only the `forking` part of it can be changed.

I'm not sure if this can be done in a backward-compatible way. At the very least, I made the huge mistake of using different names in the config and in the reset method. But maybe we can add the names in the config to the RPC method and accept both.

Update: any work on this should also include adding a proper network helper in the `hardhat-network-helpers` package

fvictorio:
Closing this in favor of having better support for multiple networks and changing networks on runtime.

PR #2411, opened 2022-02-22
Add support for `--chainid` argument in `node` command 

prajnastra:
- [ ] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team. (Issues #2305)

---

This PR adds support for `--chainid` argument in `npx hardhat node` command.

### DIFF
--- a/packages/hardhat-core/src/builtin-tasks/node.ts
+++ b/packages/hardhat-core/src/builtin-tasks/node.ts
@@ -94,14 +94,17 @@ Private Key: ${privateKey}`;
 subtask(TASK_NODE_GET_PROVIDER)
   .addOptionalParam("forkUrl", undefined, undefined, types.string)
   .addOptionalParam("forkBlockNumber", undefined, undefined, types.int)
+  .addOptionalParam("chainid", undefined, 31337, types.int)
   .setAction(
     async (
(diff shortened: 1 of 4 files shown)


alcuadrado:
Thanks, @webdevabhijit!

@fvictorio will review this after we get our next release out.

webdevabhijit:
@alcuadrado Sure.

luncht1me:
Still relevant.

fvictorio:
Hey @ethabhijit, thanks for the PR and sorry for taking so long to respond.

I don't think this implementation works, unless I'm missing something. The reason is that it relies on using `hardhat_reset` with the new `chainId`, but I'm pretty sure that `hardhat_reset` doesn't support that. So I would say that this feature is blocked by https://github.com/NomicFoundation/hardhat/issues/1697

In the meantime, for people landing here, a workaround for this missing feature is to use an environment variable. You should set something like this in your hardhat config:

```js
networks: {
  hardhat: {
    chainId: Number(process.env.HARDHAT_CHAIN_ID) ?? 31337
  },
},
```

And then `HARDHAT_CHAIN_ID=123 npx hardhat node` should use 123 as a chain id.

PR #2414, opened 2022-02-23
Test `HttpProvider`

feuGeneA:


packages/hardhat-core/test/internal/core/providers/http.ts
```diff
@@ -1,7 +1,37 @@
...
+        result: "whatever",
+      };
+      mockPool.intercept({ method: "POST", path: "/" }).reply(200, response);
+      const provider = new HttpProvider(url, networkName, {}, 20000, mockPool);
+      const result = await provider.request({ method: "net_version" });
+      assert.equal(result, response.result);
+    });
+  });
+
   describe("429 Too many requests - retries", function () {
```
alcuadrado, line 35: This test doesn't really work. We should replace it with a mocked one instead.

packages/hardhat-core/src/internal/core/providers/http.ts
```diff
@@ -37,7 +37,8 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
     private readonly _url: string,
     private readonly _networkName: string,
     private readonly _extraHeaders: { [name: string]: string } = {},
-    private readonly _timeout = 20000
+    private readonly _timeout = 20000,
+    client: Pool | undefined = undefined
```
alcuadrado, line 41: What about 

```suggestion
    dispatcher: Dispatcher | undefined = undefined
```

and change the field

```
  private _client: Pool;
```

to

```
  private _dispatcher: Dispatcher;
```

packages/hardhat-core/test/internal/core/providers/http.ts
```diff
@@ -1,7 +1,37 @@
...
+import { MockAgent, MockPool } from "undici";
+
 import { HttpProvider } from "../../../../src/internal/core/providers/http";
+import { SuccessfulJsonRpcResponse } from "../../../../src/internal/util/jsonrpc";
 import { ALCHEMY_URL } from "../../../setup";
 
 describe("HttpProvider", function () {
+  const url = "http://some.node";
+  const networkName = "NetworkName";
+  const mockPool = (() => {
```
alcuadrado, line 11: I find this pattern confusing tbh

alcuadrado:
pre-review: I like this approach

packages/hardhat-core/src/internal/core/providers/http.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2414/commits/c2893b9a6a701be53a294d726c52259a1887f761

packages/hardhat-core/test/internal/core/providers/http.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2414/commits/8e2ad09dcc67c590b9b836a283ff58ba6cdd7f33

packages/hardhat-core/test/internal/core/providers/http.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2414/commits/5c50144dc2b807a25c6d9b4229145817c2b1022e

alcuadrado:
I felt a bit uneasy about the tests because a single pool is shared, and because it's unclear in which order undici intercepts are evaluated, and I felt that maybe we weren't returning a 429 at all, as we aren't testing that.

I created a script that validates that.

I suggest we create a new pool in each test, and also that we assert that a 429 is returned. We can do it like this

```js
const { MockPool, MockAgent } = require("undici");

const TOO_MANY_REQUEST_STATUS = 429;

const successResponse = {
  jsonrpc: "2.0",
  id: 1,
  result: "whatever",
};

async function main() {
  const agent = new MockAgent({
    // as advised by https://undici.nodejs.org/#/docs/best-practices/writing-tests
    keepAliveTimeout: 10, // milliseconds
    keepAliveMaxTimeout: 10, // milliseconds
  });
  agent.disableNetConnect();

  const pool = new MockPool("http://url.com", { agent });

  let tooManyRequestsReturned = false;

  pool.intercept({ method: "POST", path: "/" }).reply(() => {
    tooManyRequestsReturned = true;

    return {
      statusCode: TOO_MANY_REQUEST_STATUS,
      data: "",
      responseOptions: { headers: { "retry-after": "1" } },
    };
  });

  pool.intercept({ method: "POST", path: "/" }).reply(200, successResponse);

  console.log(await pool.request({ method: "POST", path: "/" }));

  console.log(await pool.request({ method: "POST", path: "/" }));
}

main().catch(console.error);
````

File: packages/hardhat-core/test/internal/core/providers/http.ts [hunk 4]
```diff
@@ -1,28 +1,63 @@
...
+
+const TOO_MANY_REQUEST_STATUS = 429;
+
+function makeMockPool(url: string): MockPool {
+  const agent = new MockAgent({
+    // as advised by https://undici.nodejs.org/#/docs/best-practices/writing-tests
+    keepAliveTimeout: 10, // milliseconds
+    keepAliveMaxTimeout: 10, // milliseconds
+  });
+  agent.disableNetConnect();
```
fvictorio, line 17:
```suggestion
  // throw when requests are not matched in a MockAgent intercept
  agent.disableNetConnect();
```

alcuadrado (approved):

PR #2422, opened 2022-02-25
Use a low keepAlive in the tests http provider

fvictorio:
If we use the default timeouts, the tests that use the http provider have to wait 4 seconds for the server to close, because it attempts a graceful shutdown and it still has open connections.


feuGeneA:
Any reason to use a `Client` instead of a `Pool`?

fvictorio:
> Any reason to use a `Client` instead of a `Pool`?

I don't think so, I'm just using the simplest object. I first implemented it with a `Pool` but I didn't see a reason to have a connection pool in this case.

Jibanandamht:
let me

PR #2428, opened 2022-02-26
Add chai matchers for various BigNumbers

feuGeneA:
Closes HH-432.

As usual, consider reviewing commit-by-commit, or consider referring back to individual commits via `blame` when the context is unclear during a review of all of the changes at once.


packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,217 @@
...
+  name: string,
+  negativeName: string,
+  utils: Chai.ChaiUtils
+) {
+  return (_super: (...args: any[]) => any) =>
+    overwriteBigNumberFunction(method, name, negativeName, _super, utils);
+}
+
+function normalize(source: SupportedNumber | string): bigint {
+  if (source instanceof BigNumberEthers) {
```
fvictorio, line 72: This doesn't work if the user has a different version of ethers (or of the other libraries). I think this is another good reason to use another approach (again, maybe based in optional dependencies, or in `any`ing everything).

packages/hardhat-test-helpers/src/chai/types.ts
```diff
@@ -0,0 +1,16 @@
+// eslint-disable-next-line @typescript-eslint/no-namespace, @typescript-eslint/no-unused-vars
+declare namespace Chai {
```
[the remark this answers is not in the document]
feuGeneA: The existing Chai types specify that these comparison operators should only accept `number` (or `number|Date` for some of them). Without these changes, we can't override the methods to support additional types. Here's the actual error output that occurs: https://gist.github.com/feuGeneA/4e09115228b205887a41d981c2409de7

Since you've got me thinking about it, though, I'm wondering if we should specify here which types we're overloading it for, rather than just saying `any`. What do you think about that?

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,217 @@
...
+//   isNumber
+//   isNotNumber
+//   closeTo
+//   approximately
+
+import { BigNumber as BigNumberEthers } from "ethers";
+import { BigNumber as BigNumberJs } from "bignumber.js";
+import BN from "bn.js";
+
+export type SupportedNumber =
```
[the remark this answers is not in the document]
feuGeneA: It's used in the tests. I'm guessing you didn't expand the diff that was hidden because the changes to that file are so large.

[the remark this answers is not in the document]
fvictorio: Oh yes, that makes sense, thank you!

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/b317f4d52a2c7be29f913d3292e1bcf0e7c862c3

packages/hardhat-test-helpers/package.json
```diff
@@ -55,5 +55,11 @@
   },
   "peerDependencies": {
     "hardhat": "^2.0.0"
+  },
+  "dependencies": {
+    "bignumber.js": "^9.0.2",
+    "bn.js": "^5.1.0",
+    "chai": "^4.2.0",
+    "ethers": "^5.0.0"
```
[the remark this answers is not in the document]
feuGeneA: Discussed with Pato, and he recommended a specific approach, which I've implemented in https://github.com/NomicFoundation/hardhat/pull/2428/commits/b317f4d52a2c7be29f913d3292e1bcf0e7c862c3

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,264 @@
...
+  chaiUtils: Chai.ChaiUtils
+) {
+  return function (this: Chai.AssertionStatic, ...args: any[]) {
+    const [actualArg] = args;
+    const expectedFlag = chaiUtils.flag(this, "object");
+    if (chaiUtils.flag(this, "doLength") && isEthersBigNumber(actualArg)) {
+      // TODO: consider whether we really need this case (support for eg
+      // `expect('foo').to.have.length.of.at.least(BigInt(2));`, and if so then
+      // get it tested and generalize it to work with all the different
+      // BigNumbers.
```
fvictorio, line 154: I think we do need this. It could happen because of a test like:

```js
const tokens = await contract.getTokens()
const numberOfTokens = await contract.getTokensCount()
expect(tokens).to.have.length(numberOfTokens)
```

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,264 @@
...
+        return lhs <= rhs;
+      } else {
+        throw new Error(`Unknown comparison operation ${method}`);
+      }
+    }
+    if (isBigNumber(expectedFlag) || isBigNumber(actualArg)) {
+      const expected = normalize(expectedFlag);
+      const actual = normalize(actualArg);
+      throwIfUnsafe(expectedFlag);
+      throwIfUnsafe(actualArg);
```
fvictorio, line 177:
```suggestion
      throwIfUnsafe(expectedFlag);
      throwIfUnsafe(actualArg);
      const expected = normalize(expectedFlag);
      const actual = normalize(actualArg);
```

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,264 @@
...
+      this.assert(
+        compare(functionName, expected, actual),
+        `expected ${expected} to ${readableName} ${actual}`,
+        `expected ${expected} to ${readableNegativeName} ${actual}`,
+        expected,
+        actual
```
fvictorio, line 183: Using `expected` and `actual` here means that the error looks like this:

image

I wonder if the 3n and 4n will be confusing for users.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,264 @@
...
+        return lhs >= rhs;
+      } else if (method === "lte") {
+        return lhs <= rhs;
+      } else {
+        throw new Error(`Unknown comparison operation ${method}`);
+      }
+    }
+    if (isBigNumber(expectedFlag) || isBigNumber(actualArg)) {
+      const expected = normalize(expectedFlag);
+      const actual = normalize(actualArg);
```
fvictorio, line 175: Isn't this inverted? If you do `expect(x).to.equal(y)`, `x` is the actual value and `y` is the expected one, but here they are swapped. This results in the wrong error message.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,264 @@
...
+      isBigNumber(finishArg)
+    ) {
+      const expected = normalize(expectedFlag);
+      const start = normalize(startArg);
+      const finish = normalize(finishArg);
+      throwIfUnsafe(expectedFlag);
+      throwIfUnsafe(startArg);
+      throwIfUnsafe(finishArg);
+      this.assert(
+        start <= expected && finish >= expected,
```
fvictorio, line 215:
```suggestion
        start <= expected && expected <= finish,
```

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,264 @@
...
+      throwIfUnsafe(actualArg);
+      throwIfUnsafe(deltaArg);
+      function abs(i: bigint): bigint {
+        return i < 0 ? BigInt(-1) * i : i;
+      }
+      this.assert(
+        abs(expected - actual) <= delta,
+        `expected ${expected} to be close to ${actual}`,
+        `expected ${expected} not to be close to ${actual}`,
+        `A number between ${actual - delta} and ${actual + delta}`,
```
fvictorio, line 257: This message is not shown when a test with this matcher fails, I'm not sure why. Maybe it depends on the used reporter.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,264 @@
...
+      const delta = normalize(deltaArg);
+      throwIfUnsafe(expectedFlag);
+      throwIfUnsafe(actualArg);
+      throwIfUnsafe(deltaArg);
+      function abs(i: bigint): bigint {
+        return i < 0 ? BigInt(-1) * i : i;
+      }
+      this.assert(
+        abs(expected - actual) <= delta,
+        `expected ${expected} to be close to ${actual}`,
```
fvictorio, line 255: The default error message from mocha is a bit nicer:

```
     AssertionError: expected 3 to be close to 5 +/- 1
```

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
feuGeneA: I committed this change, but now I realize I don't like it. :)

The code seems more intuitive with your suggestion, BUT, with the suggestion in place the error message given when trying to use a fractional argument (eg `1.1`) changes from `The number 1.1 cannot be converted to a BigInt because it is not an integer` to `Cannot compare to unsafe integer 1.1. Consider using BigInt(1.1) instead. For more details, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger`.  The message for the "unsafe" case suggests that you wrap the argument in `BigInt(...)`, but that only makes sense if you're already using an integer, and calling `normalize()` first ensures that a non-integer argument raises an error before we attempt to check for integer safety.  Perhaps we could rename the function to `throwIfUnsafeInt()`? Just to be more clear? Just brainstorming...

I'm open to further changes here, but for now I've reverted this suggestion.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,264 @@
...
+//   isNumber
+//   isNotNumber
+//   closeTo
+//   approximately
+
+import { BigNumber as EthersBigNumberType } from "ethers";
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { BigNumber as BigNumberJsType } from "bignumber.js";
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { default as BNType } from "bn.js";
```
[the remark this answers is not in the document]
feuGeneA: My understanding is that since I'm only ever using eg `BigNumberJsType` as a type, and never actually instantiating it, `tsc` doesn't actually translate this `import` statement into a `require()` expression, so no such error can happen.

Indeed, I checked the `bigNumber.js` built from this source file, and there are no `require()` statements corresponding to these `import` statements.

Further, I tried adding in a simple `const b = new BNType(3);` in the file scope, and that *did* induce `tsc` to translate the `BNType` `import` to a `require("bn.js")`, suggesting that my presumption (non-instantiation avoids inducing a `require`) seems to be correct.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
fvictorio: What if the `normalize` function also checks that the integer is safe? Or is there some scenario where we would call that function with an unsafe integer and that would be ok?

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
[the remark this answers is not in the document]
fvictorio: Oh, ok. I guess it feels weird to me to use a dev dependency in this way, but maybe it's fine. We should use `import type` though to make sure no one uses the value accidentally.

@alcuadrado do you have thoughts about this?

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
feuGeneA: I think it's not inverted. Note that the tests also exercise comparisons of `number` vs `number`, which induces a fallback to `_super.apply()`, and the tests are structured such that the error message is expected to be consistent across both `number`-vs-`number` and `BigNumber` comparisons.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
feuGeneA: Great catch, thank you. Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/77524a583f8538c11e7ee80904dc247d92db72d2

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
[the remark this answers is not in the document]
feuGeneA: Switched to using `import type` in https://github.com/NomicFoundation/hardhat/pull/2428/commits/7d80addfc666eccc02d94ec3e307acf7e86d3420

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
feuGeneA: I like that, thanks. Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/d875eb1e09dfcf7e66ff3a31a086882e5ae815ad

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
```diff
@@ -0,0 +1,264 @@
...
+
+function isBN(n: any) {
+  try {
+    // eslint-disable-next-line import/no-extraneous-dependencies
+    const BN: typeof BNType = require("bn.js");
+    return n instanceof BN && BN.isBN(n);
```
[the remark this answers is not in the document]
feuGeneA: I did some experimentation in an external client package, and learned first-hand what you mean here. I dropped the `instanceof` checks in https://github.com/NomicFoundation/hardhat/pull/2428/commits/bd84dd0f930960f6d588141a3f058f508bb87e31. Also in that commit I simplified the conversion of a `bignumber.js` `BigNumber` to `BigInt` because I realized that that `BigNumber`'s `.toString()` argument *defaults to* `10`, so we don't actually need to specify it, so we can collapse the conversion to just one code path, so it actually doesn't matter if `ethers.BigNumber.isBigNumber` returns true for instances of `bignumber.js`.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
fvictorio: Oh, I'm ok with leaving it, that's fine and probably a bit more defensive.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
fvictorio: Hmm, I'm not convinced. If I do `expect(2).to.equal(3)` with chai without any extensions, this is the error:

image

But if I add `chai.use(bnChai)` and do `expect(2).to.equal(new BN(3))` I get:

image

Notice that the actual and expected are inverted now.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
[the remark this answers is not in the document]
alcuadrado: `import type` doesn't get translated to a runtime `require`, so it doesn't fail if not present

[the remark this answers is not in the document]
alcuadrado: afaik that's actual the goal of that TS construct, being explicit that you don't want a runtime import.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
feuGeneA: Do something like https://github.com/NomicFoundation/hardhat/blob/bd1d6b971fb213d47ad62f556ef30ec58480155d/packages/hardhat-ethers/src/internal/index.ts#L20 to do a sort of "repr" (eg from Python), to have the error message be as depicted here but to have the "diff" (constructed from the `expected`/`actual` arguments on line 182,183) reflect the type of BigNumber object that was given in the input to the matcher.

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/bb29b72ea39b990792a3c23acc3d2f95ae6b2030

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/5031927991113e6726e6a7be83b3263df1eeaf5b

packages/hardhat-test-helpers/src/chai/matchers/bigNumber.ts
feuGeneA: @alcuadrado and I discussed offline and decided that we should do this instead: https://github.com/NomicFoundation/hardhat/pull/2428/commits/c28827d1a2c404f15e813e5bb656932fcbe12c0a

packages/hardhat-chai-matchers/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Nomic Labs LLC
```
fvictorio, line 3: Not sure about this but

```suggestion
Copyright (c) 2022 Nomic Labs LLC
```

packages/hardhat-chai-matchers/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "rimraf": "^3.0.2",
+    "ts-node": "^8.1.0",
+    "typescript": "~4.5.2"
+  },
+  "peerDependencies": {
+    "ethers": "^5.0.0",
+    "hardhat": "^2.0.0"
+  },
+  "dependencies": {
+    "chai": "^4.2.0"
```
fvictorio, line 64: Shouldn't this be a peer dependency too?

packages/hardhat-chai-matchers/src/index.ts
```diff
@@ -0,0 +1 @@
+export { bnChai } from "./bnChai";
```
fvictorio, line 1: I think we should add `chai.use(bnChai)` here

packages/hardhat-chai-matchers/src/bigNumber.ts
```diff
@@ -0,0 +1,329 @@
...
+    typeof source === "string" ||
+    typeof source === "number" ||
+    typeof source === "bigint"
+  ) {
+    // first construct a BigInt, to allow it to throw if there's a fractional value:
+    const toReturn = BigInt(source);
+    // then check the source for integer safety:
+    if (typeof source === "number" && !Number.isSafeInteger(source)) {
+      throw new RangeError(
+        `Cannot compare to unsafe integer ${source}. Consider using BigInt(${source}) instead. For more details, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger`
```
fvictorio, line 158: I'm pretty sure the suggestion to use `BigInt(${source})` is wrong. We should recommend using `${source}n` instead.

Anyway, this is the function that we'll move to `hardhat/common` and we'll work on a proper explanation and suggestion there.

packages/hardhat-chai-matchers/src/bigNumber.ts
```diff
@@ -0,0 +1,329 @@
...
+// very least get them tested.
+//   isAbove
+//   isAtLeast
+//   isBelow
+//   isAtMost
+//   isAt
+//   isNumber
+//   isNotNumber
+//   closeTo
+//   approximately
```
fvictorio, line 12: I think we can drop this comment. Most of these methods work. `isNumber` doesn't (that is, `expect(1n).to.be.a('number')` throws), but I think that's fine.

packages/hardhat-chai-matchers/src/bigNumber.ts
```diff
@@ -0,0 +1,329 @@
...
+    }
+  };
+}
+
+function overrideCloseTo(utils: Chai.ChaiUtils) {
+  return (_super: (...args: any[]) => any) =>
+    overwriteBigNumberCloseTo(_super, utils);
+}
+
+function overwriteBigNumberCloseTo(
```
fvictorio, line 285: Doing `expect(1).to.be.closeTo(2)` throws this error:

```
AssertionError: the arguments to closeTo or approximately must be numbers, and a delta is required
```

But doing `expect(1n).to.be.closeTo(2)` throws this:

```
Error: cannot convert undefined to BigInt
```

Not a huge deal, really, but I would fix it if it's not a lot of work.

packages/hardhat-chai-matchers/src/bigNumber.ts
```diff
@@ -0,0 +1,329 @@
...
+  } catch (e) {
+    return false;
+  }
+}
+
+function isBigNumberJsBigNumber(n: any) {
+  try {
+    // eslint-disable-next-line import/no-extraneous-dependencies
+    const BigNumber: typeof BigNumberJsType = require("bignumber.js").BigNumber;
+    return n instanceof BigNumber && BigNumber.isBigNumber(n);
```
fvictorio, line 45: I think we discussed this but I don't remember right now. Why do we use `instanceof` here but not in the other checks?

packages/hardhat-chai-matchers/src/bigNumber.ts
fvictorio: Also: we should tell apart the scenarios where `isSafeInteger` returns false because the number is too large, and the scenarios where it's a non-integer (e.g., `3.14`), and show different error messages in both cases.

fvictorio: Oh, I just realized that that's meant to be handled by the `BigInt(source)` call. I think we should be the ones doing the check though, so that we can show a better error.

feuGeneA: Using the `1n` syntax (rather than the `BigInt(1)` syntax) results in a `tsc` error:
```
test/bigNumber.ts:19:12 - error TS2737: BigInt literals are not available when targeting lower than ES2020.

19     expect(1n).to.equal(new BN(1));
              ~~
```

packages/hardhat-chai-matchers/package.json
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/e27c53b52117e9ae2974ce966e1954adbd00c69f

packages/hardhat-chai-matchers/src/index.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/eb16ef2afdd2f0a1ab21e56aa1e4c62d115226da

packages/hardhat-chai-matchers/src/bigNumber.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/0e93f3537044051aa362c9974c768ca78aa50c6c

packages/hardhat-chai-matchers/src/bigNumber.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/1dd96ce8a13a5676f42ca17111989c6ba3e02be8. I didn't completely drop it, because some of them actually don't work. But I revised it to remove the ones that do work, and to remove `isNumber` and `isNotNumber`.

packages/hardhat-chai-matchers/src/bigNumber.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/f701a2ba9ff1db93039535bf51b2188903f938a4. I had removed the others previously but I missed this one. Good catch!

packages/hardhat-chai-matchers/package.json
feuGeneA: In that https://github.com/NomicFoundation/hardhat/commit/e27c53b52117e9ae2974ce966e1954adbd00c69f I accidentally removed it as a devDependency, but I restored it in https://github.com/NomicFoundation/hardhat/pull/2428/commits/e6038e832f2111b0985a605084cf074545141faf

packages/hardhat-chai-matchers/src/bigNumber.ts
fvictorio: Sorry, I meant in the error message. I think we should do `Consider using ${source}n instead.` instead of `Consider using BigInt(${source}) instead.` Then I mentioned the thing about `3.14` because a message suggesting to use `3.14n` would be really misleading.

packages/hardhat-chai-matchers/src/bigNumber.ts
fvictorio: What, really? I'm pretty sure I tried `isAbove`, `isAtLeast`, `isBelow` and `isAtMost`. Maybe I did something wrong, but I'm fairly confident that they were working. I missed `isAt` though.

packages/hardhat-chai-matchers/src/bigNumber.ts
fvictorio: By the way, I think this is something that we can do more properly when we move this logic to a `hardhat/common` helper, so feel free to disregard this. We can merge this PR and then discuss the error message when we make that change (assuming that you are not going to do that in this PR).

packages/hardhat-core/src/common/bigInt.ts
```diff
@@ -0,0 +1,97 @@
...
+    return BigInt(source.toString());
+  } else if (
+    typeof source === "string" ||
+    typeof source === "number" ||
+    typeof source === "bigint"
+  ) {
+    if (typeof source === "number") {
+      if (!Number.isInteger(source)) {
+        // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
+        throw new RangeError(
```
fvictorio, line 26: We have to use a `HardhatError` here.

packages/hardhat-core/src/common/bigInt.ts
```diff
@@ -0,0 +1,97 @@
...
+        );
+      }
+      if (!Number.isSafeInteger(source)) {
+        // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
+        throw new RangeError(
+          `Cannot convert unsafe integer ${source} to BigInt. Consider using BigInt(${source}) instead. For more details, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger`
+        );
+      }
+    }
+    return BigInt(source);
```
fvictorio, line 37: We should do [exhaustiveness checking](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#exhaustiveness-checking) here.

packages/hardhat-chai-matchers/src/bigNumber.ts
feuGeneA: Just remove this comment. It's for assertion types that are not `expect`, and everyone out there uses `expect`. So just leave it undone until someone complains about it.

feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/2b0fc7464ee7fbd03dee901630b82aa9ce791963

packages/hardhat-chai-matchers/src/bigNumber.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/c9dafded0e0bebd5fd01395f681c1a186af30cdf

packages/hardhat-core/src/common/bigInt.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/af03a9a678c475c62e4e6f0a93db87c2ce901ba5

packages/hardhat-core/src/common/bigInt.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2428/commits/44e5d96974af7755511f929f34008d4fd7a4978c

fvictorio (approved):

packages/hardhat-chai-matchers/package.json
```diff
@@ -0,0 +1,66 @@
...
```
frangio, line 64: Shouldn't this be a peer dependency?

PR #2440, opened 2022-03-01
Update plugins.js to add Hardhat Metadata Plugin

krruzic:
add  plugin hardhat-network-metadata. This plugin allows you to add an extra field `metadata` within a network config like so:
```js
{ ...,
  metadata: { 
    anything: "goes here", 
    seriously: 1224233, 
    literallyWhateverYouWant: { a: 1 }
   }
}
```


alcuadrado:
Nice! Thanks for sending this pr

PR #2454, opened 2022-03-04
Fix OOM during parallel compilation

fvictorio:
Projects with a huge number of contracts like OZ's [contract-wizard](https://github.com/OpenZeppelin/contracts-wizard) can get an OOM error during compilation. This PR fixes that problem, by reducing the potential size of the array that accumulates the compilation results to what it was in v2.8.4.

Even with that change, it's still possible to get an OOM if the concurrency level is high and the available memory is not enough. This happens because multiple compilation results are present at the same time, and they may not fit in the available memory. In that case, the workaround is to increase the memory or to reduce the concurrency level. To make the latter option possible, this PR adds a `--concurrency` param to the compile task.

### DIFF
--- a/.changeset/slow-pianos-double.md
+++ b/.changeset/slow-pianos-double.md
@@ -1,0 +1,6 @@
+---
+"hardhat": patch
+---
+
+- Fixed a potential OOM error during parallel compilation
+- Added a `--concurrency` param to the compile task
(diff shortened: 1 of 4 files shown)


fvictorio:
@alcuadrado do you want to review this or can I merge it?

Issue #2380, opened 2022-02-13
Forked hardhat node becomes unresponsive when connecting with Metamask


aliveli186:
After upgrading to v2.8.4 I started to receive this error.
How can I increase the memory size?

```
<--- Last few GCs --->

[2096:000002B2E4087E60]  1317204 ms: Scavenge (reduce) 2046.2 (2082.6) -> 2045.5 (2082.6) MB, 4.7 / 0.0 ms  (average mu
= 0.115, current mu = 0.004) allocation failure
[2096:000002B2E4087E60]  1319918 ms: Mark-sweep (reduce) 2046.9 (2083.2) -> 2045.9 (2083.2) MB, 1411.0 / 0.1 ms  (+ 1043
.8 ms in 1247 steps since start of marking, biggest step 4.4 ms, walltime since start of marking 2482 ms) (average mu =
0.112, current

<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 00007FF6D9A630AF v8::internal::CodeObjectRegistry::~CodeObjectRegistry+112511
 2: 00007FF6D99F2216 DSA_meth_get_flags+65542
 3: 00007FF6D99F30CD node::OnFatalError+301
 4: 00007FF6DA324B6E v8::Isolate::ReportExternalAllocationLimitReached+94
 5: 00007FF6DA30F09D v8::SharedArrayBuffer::Externalize+781
 6: 00007FF6DA1B268C v8::internal::Heap::EphemeronKeyWriteBarrierFromCode+1468
 7: 00007FF6DA1BF339 v8::internal::Heap::PublishPendingAllocations+1129
 8: 00007FF6DA1BC30A v8::internal::Heap::PageFlagsAreConsistent+2842
 9: 00007FF6DA1AEF89 v8::internal::Heap::CollectGarbage+2137
10: 00007FF6DA1AD140 v8::internal::Heap::AllocateExternalBackingStore+2000
11: 00007FF6DA1D1A76 v8::internal::Factory::NewFillerObject+214
12: 00007FF6D9F04CB5 v8::internal::DateCache::Weekday+1797
13: 00007FF6DA3B2541 v8::internal::SetupIsolateDelegate::SetupHeap+494417
14: 00007FF6DA407404 v8::internal::SetupIsolateDelegate::SetupHeap+842260
[...]

PR #2456, opened 2022-03-04
Temporary fix to #2380

alcuadrado:
This is a quick fix to #2380. This is not the ideal fix because the fee history is cached per-request instead of per-block, which leads to lots of cache misses. Also, if the remote network can't answer an `eth_feeHistory` we fail, but we could do something better in that case.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -72,6 +72,7 @@ import { VMTracer } from "../stack-traces/vm-tracer";
 
 import "./ethereumjs-workarounds";
 import { rpcQuantityToBN } from "../../core/jsonrpc/types/base-types";
+import { JsonRpcClient } from "../jsonrpc/client";
 import { bloomFilter, Filter, filterLogs, LATEST_BLOCK, Type } from "./filter";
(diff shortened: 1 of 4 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1403,21 +1411,61 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+    // costly to do.
     let pendingBlock: Block | undefined;
     if (lastBlock.gte(pendingBlockNumber)) {
       pendingBlock = await this.getBlockByNumber("pending");
     }
 
+    // This is part of a temporary fix to https://github.com/NomicFoundation/hardhat/issues/2380
+    const firstLocalBlock = !rangeIncludesRemoteBlocks
+      ? oldestBlock
+      : BN.min(new BN(this._forkBlockNumber!), lastBlock).addn(1);
```
fvictorio, line 1465:
```suggestion
      : BN.min(new BN(this._forkBlockNumber!).addn(1), lastBlock);
```

fvictorio: Disregard this.

fvictorio (approved):

PR #2458, opened 2022-03-05
Add pre-release workflow

fvictorio:


packages/hardhat-etherscan/test/fixture-projects/hardhat-project/contracts/TestReentrancyGuardLocal.sol
```diff
@@ -3,8 +3,9 @@ pragma solidity 0.5.15;
 import "./ReentrancyGuard.sol";
 
 contract TestReentrancyGuardLocal is ReentrancyGuard {
+  string message = "placeholder";
```
alcuadrado, line 6: Can you explain this?

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,36 @@
+name: Lint
```
alcuadrado, line 1:
```suggestion
name: Pre-release tests
```

packages/hardhat-etherscan/test/fixture-projects/hardhat-project/contracts/TestReentrancyGuardLocal.sol
fvictorio: The way the hardhat-etherscan e2e tests is that they replace this placeholder string for some random string. That way the contract deployed is completely new. Otherwise you get an "Already verified" error.

This contract didn't have the placeholder for some reason, so its test was failing.

fvictorio:
Not today.

alcuadrado (approved):

PR #2464, opened 2022-03-08
Update deployment guide: wait for deployment confirmation

rellfy:
Makes use of the `deployed` method in the deployment guide so that users are aware of the method used to wait for contract deployment confirmation.


fvictorio:
Thanks a lot for this @rellfy. Maybe for consistency we can use the same code as in the sample project's example script?

```js
  const greeter = await Greeter.deploy("Hello, Hardhat!");

  await greeter.deployed();

  console.log("Greeter deployed to:", greeter.address);
```

rellfy:
@fvictorio I've updated the code to match the example script

fvictorio:
Thank you @rellfy!

PR #2471, opened 2022-03-09
Add a note about mocha.parallel option

marcelomorgado:

---

Tiny doc improvement.

### DIFF
--- a/docs/guides/parallel-tests.md
+++ b/docs/guides/parallel-tests.md
@@ -6,6 +6,8 @@ You can run your tests in parallel by using the `--parallel` flag:
 $ npx hardhat test --parallel
 ```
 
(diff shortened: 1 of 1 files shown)


docs/guides/parallel-tests.md
```diff
@@ -6,6 +6,8 @@ You can run your tests in parallel by using the `--parallel` flag:
 $ npx hardhat test --parallel
 ```
 
+or set `parallel: true` to the `mocha` section of the `hardhat` config file, that'll make parallelism to work for plugins like `solidity-coverage` that don't have such flag.
```
feuGeneA, line 9:
```suggestion
Alternatively, use `parallel: true` in the `mocha` section of `hardhat.config`, which will enable parallelism for plugins that don't have a `--parallel` flag, like `solidity-coverage`.
```

feuGeneA (changes requested):

fvictorio:
Thank you @marcelomorgado!

PR #2476, opened 2022-03-10
fix: don't override grep options in test task

### LINKED ISSUE #2459: Do not override config.mocha.grep when running test without --grep option
The current behavior 

```typescript
      const mochaConfig: MochaOptions = {
        ...config.mocha,
        grep: taskArgs.grep,
      };
```
prevents plugins from making use of `config.mocha.grep` as it will be unset if the option is not coming from the cli.

I would suggest this type of logic instead

```typescript
      if (taskArgs.grep !== undefined) {
        mochaConfig.grep = taskArgs.grep;
      }
```

smartcontracts:

---

Fixes a bug in the `test` task that would override the `grep` options in the config file. 

Fixes #2459

### DIFF
--- a/.changeset/quick-moons-report.md
+++ b/.changeset/quick-moons-report.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+- Fix a bug that would override mocha grep options within the test task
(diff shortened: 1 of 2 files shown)


tynes:
Bump on this :)

fvictorio:
Thank you, Kelvin! I always forget that ` { a: 1, ...{a: undefined} }` results in `{ a: undefined }` :disappointed:

Issue #2352, opened 2022-02-08
"LoggingEnabled : false",  in an 'out of process' hardhat node doesn't work.


iangregsondev:
Hi,

I am running hardhat node as 'out of process', so I am physically running

```
npx hardhat node
```

I have the following in my hardhat config

```
networks: {
    hardhat: {
      loggingEnabled: false,
```

The output of the methods being called are still there even though the logging has been disabled. You can see them when anything is talking to the node.

This wasn't what I was expecting..

So, if I then call the following, it does disable the logging

```
 await hre.network.provider.send("hardhat_setLoggingEnabled", [
    false,
  ]);
```

So that worked, after successfully calling the "hardhat_setLoggingEnabled", no more logs can be seen.

Is this correct ?

I assumed that "logginEnabled" set to false should start the node with no logging but it does not.

Can you confirm if this is correct, maybe I am misunderstanding the "logginEnabled" setting?

Thanks.

morgansliman:
Hi @iangregsondev , thanks for opening the issue!

This does appear to be a bug, and seems to be a quick fix so I've opened a PR to allow this to be configurable as you've described. This should be fixed in the next release after the PR gets merged, so feel free to keep an eye on that for updates! We'll also close this issue when we make that release as well.

Thanks!

iangregsondev:
Thanks @morgansliman, much appreciated for looking at it.

PR #2477, opened 2022-03-10
allow node task logging to be configurable

morgansliman:
In reference to #2352

### DIFF
--- a/.changeset/twelve-coats-rhyme.md
+++ b/.changeset/twelve-coats-rhyme.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+allow node task logging to be configurable
--- a/.changeset/twelve-coats-rhyme.md
+++ b/.changeset/twelve-coats-rhyme.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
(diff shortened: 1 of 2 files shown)


.changeset/twelve-coats-rhyme.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+allow node task logging to be configurable
```
feuGeneA, line 5: Should be ready for release notes. In particular, should use past tense, and should mention the bug being fixed.

```suggestion
Fixed bug (#2352) preventing the disabling of node task logging via hardhat.config.
```

feuGeneA (approved):

.changeset/twelve-coats-rhyme.md
morgansliman: Thanks! I couldn't remember the exact format we agreed on.

alcuadrado:
Sorry I'm late about this, but I don't think this fixes it. The reason is that `hardhatNetworkConfig.loggingEnabled` is never `undefined`.

Unfortunately testing this is really complicated.

@fvictorio wdyt?

PR #2480, opened 2022-03-11
Fix problem with CI not running all core tests

feuGeneA:
Note that CI runs for this PR show that, for the forking test Action, 1856+35+1 (passing+pending+failing) tests were run, whereas, for [a recent run without this change](https://github.com/NomicFoundation/hardhat/runs/5517403119?check_suite_focus=true), only 404+2 tests were run.

The problem seems to be that our usage of `mocha --recursive "test/internal/hardhat-network/{helpers,jsonrpc,provider}/*.ts"` was not picking up the tests in the subdirectories of `provider` because the glob pattern was only resolving to `.ts` files and not to any folder names, so there was nothing for `--recursive` to act upon. (Thinking further, I suspect that using `--recursive` has no effect when used alongside `.../*.ts`, since no folders will be targeted.)

### DIFF
--- a/packages/hardhat-core/package.json
+++ b/packages/hardhat-core/package.json
@@ -32,8 +32,8 @@
     "prettier": "prettier \"**/*.{js,md,json}\"",
     "test": "mocha --recursive \"test/**/*.ts\"",
     "test:except-tracing": "mocha --recursive \"test/**/*.ts\" --invert --grep \"Stack traces\"",
-    "test:tracing": "mocha --recursive \"test/internal/hardhat-network/{helpers,stack-traces}/*.ts\"",
-    "test:forking": "mocha --recursive \"test/internal/hardhat-network/{helpers,jsonrpc,provider}/*.ts\"",
+    "test:tracing": "mocha --recursive \"test/internal/hardhat-network/{helpers,stack-traces}/**/*.ts\"",
+    "test:forking": "mocha --recursive \"test/internal/hardhat-network/{helpers,jsonrpc,provider}/**/*.ts\"",
     "build": "tsc --build .",
     "clean": "rimraf builtin-tasks internal types utils *.d.ts *.map *.js build-test tsconfig.tsbuildinfo test/internal/hardhat-network/provider/.hardhat_node_test_cache"
   },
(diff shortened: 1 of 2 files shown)


feuGeneA:
Currently there is (thankfully only) a single failure upon the re-enabling of these tests:
```
  1) Evm module
       Alchemy Forked provider
         evm_increaseTime
           should accept a hex string param:

      AssertionError: expected 57 to equal 123
      + expected - actual

      -57
      +123
      
      at Context.<anonymous> (test/internal/hardhat-network/provider/modules/evm.ts:145:18)
      at processTicksAndRejections (internal/process/task_queues.js:97:5)
```
It's not clear to me yet why this is failing...

feuGeneA:
For some reason, in the test
```
     Evm module
       Alchemy Forked provider
         evm_increaseTime
           should accept a hex string param
```
the initial offset was not `0`, despite the test's assumption that it should be.  In fact, the initial offset was negative!  Thus the expected offset was significantly lower than what was expected.  I'm not sure why it was negative, nor whether that in and of itself should be considered problematic, but I went ahead and made the test capture and use the *actual* initial offset (instead of assuming it's `0`), and now the test is passing.

fvictorio:
> I'm not sure why it was negative, nor whether that in and of itself should be considered problematic

This is the reason:

https://github.com/NomicFoundation/hardhat/blob/bd1d6b971fb213d47ad62f556ef30ec58480155d/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts#L233-L237

The offset is the difference between the current timestamp (as returned by `new Date()`) and the actual timestamp we want the next block to have. This is relevant, for example, when you set an `initialDate` in your hardhat config. If the `initialDate` is one hour in the past, then each block will have a timestamp of `currentTimestamp - 3600`.

For forked networks, the same argument applies, but instead of the `initialDate`, the timestamp of the forked block is used.

Fun fact I just discovered: even if you don't set an `initialDate` and the current timestamp is used by default, the offset might be negative, because there could be a difference in time between the moment that the configuration is resolved (and therefore the `initialDate` is assigned to `new Date()`) and the moment the hardhat network is created, because this happens after the first request is received.

fvictorio (approved):
Approving this, but I would wait for @alcuadrado's approval too.

alcuadrado (approved):
@fvictorio's analysis is right. We also discussed this with @feuGeneA on a call and arrived to the same conclusions.

PR #2495, opened 2022-03-15
Update Pinned Block Number in Mainnet Forking guide

### LINKED ISSUE #2330: Update Block Number in Block Pinning Documentation
The `blockNumber` in the [documentation](https://hardhat.org/hardhat-network/guides/mainnet-forking.html) is `11095000` which is 471 days ago. I think it should be updated to `14135803` which is the block number at the moment of my typing this issue! 

Why?

There's a bunch of newly deployed contracts and the such since the last year. I was testing with pinning and the contract I was interacting with didn't exist!

Will it need updated again? 

Yes. But like an out-of-date copyright on a website this looks bad, and if I can save 5 minutes of someone else's time, I say we go for it.

Namaskar-1F64F:

---

This PR changes the value of the pinned block in the mainnet forking guide from 11095000 to 14393928. I chose 14393928 because it was the block when I made the change.

closes #2330

### DIFF
--- a/docs/hardhat-network/guides/mainnet-forking.md
+++ b/docs/hardhat-network/guides/mainnet-forking.md
@@ -46,7 +46,7 @@ networks: {
   hardhat: {
     forking: {
       url: "https://eth-mainnet.alchemyapi.io/v2/<key>",
-      blockNumber: 14393928
+      blockNumber: 14390000
     }
   }
 }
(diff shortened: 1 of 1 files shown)


fvictorio:
Thank you @Namaskar-1F64F. I just changed the number to make it look less random.

Issue #1585, opened 2021-06-23
Accept leading zeros in QUANTITY


fvictorio:
This throws an error:

```js
await ethers.provider.send("hardhat_setBalance", ["0x7D6bF26aA0B80A44b2ba41edAE1AA7663B6B20ac", ethers.utils.parseEther("1").toHexString()])
```

The reason it doesn't work is that 1 eth in (padded) hex is `0x0de0b6b3a7640000`. Since users rely on `ethers.utils.parseEther` a lot, we should accept this (plus, it's a valid hex IMO).

The only thing we should decide is if we accept any amount of leading zeros or only hex strings padded to an even length. I'll check what geth does and update this issue.

alcuadrado:
Unfortunately, this goes against the "definition" of the JSON-RPC types, as documented [here](https://ethereum.org/en/developers/docs/apis/json-rpc/#hex-value-encoding).

If we relax this type's validation, users can end up with test suites that only run with Hardhat Network despite only using standard json-rpc methods, which is something we want to avoid.

fvictorio:
Yeah, that makes sense for the main quantity type. I wonder though if we can have a more permissive type for our methods, like `hardhat_setBalance`, just so that the `send("hardhat_setBalance", parseEther(...).toHexString())` snippet works.

5hanth:
+1

```
    await network.provider.request({
      method: 'hardhat_setBalance',
      params: ['0x7147C79F2B8d744DDeCD70318cf4a3471cc4fACa', '0xde0b6b3a7640000']
    })
```

```     Error: invalid object key - 0 (argument="transaction:0", value={"0":"e","1":"t","2":"h","3":"_","4":"g","5":"e","6":"t","7":"B","8":"a","9":"l","10":"a","11":"n","12":"c","13":"e","gasLimit":{"type":"BigNumber","hex":"0xcf09"},"from":"0x7147C79F2B8d744DDeCD70318cf4a3471cc4fACa"}, code=INVALID_ARGUMENT, version=properties/5.4.0)```

alcuadrado:
This isn't related to this issue, @5hanth. Please double check your code.

5hanth:
my bad. it's not related to this. @alcuadrado there was another part that's trying to get balance.  ```  await network.provider.request({
      method: 'hardhat_setBalance',
      params: [myWallet, ethers.utils.parseEther('10.0').toHexString()]
    })``` works fine.

[...]

PR #2500, opened 2022-03-16
Add loadFixture and JSON-RPC test helpers

fvictorio:
Fixes HH-428.

### `setMiningMode`

We have two JSON-RPC methods to change the mining mode: `evm_setAutomine` and `evm_setIntervalMining`. They can be used independently, so any of the four combinations of "automine enabled/disabled" and "interval mining enabled/disabled" are possible.

In practice, though, I think most people only care about enabling and disabling automining. Interval mining seems to me to not be used as much. And combining both automining and interval mining is something that I doubt many people do.

Since these helper methods can be more opinionated, I think they should better reflect this fact. We should still allow the four combinations to be used, but the more common scenarios should be easier to do.

One idea:

```js
await setMiningMode(false) // disable automining, disable interval mining
await setMiningMode(true) // enable automining, disable interval mining
await setMiningMode(500) // disable automining, set interval mining to 500
await setMiningMode({ auto: true, interval: 500 }) // enable automining, set interval mining to 500
```

The last method could be, of course, used to set the mining modes in any way. The first three would then be shortcuts for the more common scenarios.

Another alternative is to have multiple methods:

```js
await enableAutomine()
await disableAutomine()
await setIntervalMining(500)
await setMiningMode(...)
```

Whose behavior would match the statements of the previous snippet. My problem with this is that I don't quite like that a method called `disableAutomine` would also disable interval mining. With a single method, it might be clearer that it's a shared behavior.

### Reset

The `hardhat_reset` is a bit of a mess. People expect it to reset the network to the initial configured state, but if you call it without any parameters, it will reset to a non-fork, blank state. That is: if your hardhat config has a forking configuration, and you call `send("hardhat_reset")`, it will be reset to an empty, local (non-forked) hardhat network.

So one idea is to do that in the `reset` helper. The implementation would then be something like:

```js
function reset(explicitConfig?) {
  if (explicitConfig === undefined) {
    // read the hre.config.networks.hardhat value
    // call hardhat_reset with that
  } else {
    // call hardhat_reset with explicitConfig
  }
```

I'm not really sure if this is possible, nor if it's a good idea, but I think we should at least think about this.


packages/hardhat-test-helpers/src/helpers/dropTransaction.ts
```diff
@@ -0,0 +1,12 @@
+import { getHardhatProvider, assertHexString } from "../utils";
+
+export async function dropTransaction(txHash: string): Promise<void> {
```
fvictorio, line 3: This should return a `boolean`, as the JSON-RPC method does.

packages/hardhat-test-helpers/src/helpers/getStorageAt.ts
```diff
@@ -0,0 +1,24 @@
...
+import {
+  getHardhatProvider,
+  assertValidAddress,
+  toRpcQuantity,
+} from "../utils";
+
+export async function getStorageAt(
+  hexAddress: string,
+  index: NumberLike,
+  block = "latest"
```
fvictorio, line 11: I think this should be `block: NumberLike | "latest" = "latest"` instead, otherwise only strings can be passed as the block.

packages/hardhat-test-helpers/src/helpers/getStorageAt.ts
```diff
@@ -0,0 +1,24 @@
+import type { NumberLike } from "../types";
+import {
+  getHardhatProvider,
+  assertValidAddress,
+  toRpcQuantity,
+} from "../utils";
+
+export async function getStorageAt(
+  hexAddress: string,
```
fvictorio, line 9: Here, and in the other methods that receive an address, we should use `address` instead of `hexAddress` as the parameter name. The reason is that it will be a bit clearer/nicer when the user hovers over the function and gets information about the function.

packages/hardhat-test-helpers/src/helpers/setStorageAt.ts
```diff
@@ -0,0 +1,31 @@
...
+  getHardhatProvider,
+  assertValidAddress,
+  assertHexString,
+  toRpcQuantity,
+} from "../utils";
+
+export async function setStorageAt(
+  hexAddress: string,
+  index: NumberLike,
+  code: string
```
fvictorio, line 12: We should accept more types of values here. Requiring a string of 32 bytes is not very ergonomic.

I think accepting a `NumberLike` would be enough. Then after converting it to a hex-encoded version, we should pad it so that it has 32 bytes, as it's required by the spec for this method.

packages/hardhat-test-helpers/test/helpers/dropTransaction.ts
```diff
@@ -0,0 +1,36 @@
...
+describe("dropTransaction", function () {
+  useEnvironment("simple");
+  const account = "0x000000000000000000000000000000000000beef";
+  const recipient = "0x000000000000000000000000000000000000beee";
+
+  it("should drop a given transaction from the mempool", async function () {
+    await this.hre.network.provider.send("evm_setAutomine", [false]);
+    await hh.setBalance(account, "0xaaaaaaaaaaaaaaaaaaaaaa");
+    await hh.impersonateAccount(account);
+    await hh.mine();
```
fvictorio, line 15: I think it's not necessary to mine a new block here.

packages/hardhat-test-helpers/src/utils.ts
```diff
@@ -67,3 +67,22 @@ export function toRpcQuantity(x: NumberLike): string {
 
   return hex.startsWith("0x") ? hex.replace("0x0", "0x") : `0x${hex}`;
 }
+
+export function assertValidAddress(hexAddress: string): void {
```
fvictorio, line 71: This function also has to verify the checksum of the address, unless the address is all lowercase.

packages/hardhat-test-helpers/test/helpers/dropTransaction.ts
```diff
@@ -0,0 +1,36 @@
...
+    );
+
+    // ensure tx is in mempool
+    assert.equal(pendingTxs[0].hash, txHash);
+    await hh.dropTransaction(txHash);
+    pendingTxs = await this.hre.network.provider.send(
+      "eth_pendingTransactions"
+    );
+    assert.equal(pendingTxs.length, 0);
+  });
```
fvictorio, line 35: We should add a test more to check that the tx arg is validated.

packages/hardhat-test-helpers/test/helpers/getStorageAt.ts
```diff
@@ -0,0 +1,41 @@
...
+
+describe("getStorageAt", function () {
+  useEnvironment("simple");
+  const account = "0x000000000000000000000000000000000000beef";
+  const code =
+    "0x000000000000000000000000000000000000000000000000000000000000beef";
+
+  it("should get the storage of a given address", async function () {
+    await hh.setStorageAt(account, "0x1", code);
+    await hh.mine();
```
fvictorio, line 17: This `mine` is also unnecessary, unless I'm missing something. Same comment for the rest of the `hh.mine`s in the other tests.

packages/hardhat-test-helpers/src/helpers/impersonateAccount.ts
```diff
@@ -0,0 +1,12 @@
+import { getHardhatProvider, assertValidAddress } from "../utils";
+
+export async function impersonateAccount(hexAddress: string): Promise<void> {
```
fvictorio, line 3: Adding this comment just as a reminder that we need to add jsdocs to all the public functions.

packages/hardhat-test-helpers/src/helpers/takeSnapshot.ts
```diff
@@ -0,0 +1,35 @@
+import { getHardhatProvider } from "../utils";
+
+export interface SnapshotRestorer {
+  restore(): Promise<void>;
```
fvictorio, line 4: We should add a jsdoc here too.

packages/hardhat-test-helpers/src/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+    hex = x.toHexString();
+  } else if ("toString" in x) {
+    hex = x.toString(16);
+  } else {
+    throw new Error(
+      `[hardhat-test-helpers] ${x} cannot be converted to an RPC quantity`
+    );
+  }
+
+  return hex.startsWith("0x") ? hex.replace("0x0", "0x") : `0x${hex}`;
```
fvictorio, line 68: Something I forgot to mention: this doesn't work with `0` because it results in sending `0x` instead of `0x0`, which is invalid. Totally my fault of course, I was the one that wrote this method.

packages/hardhat-test-helpers/src/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+      method: "hardhat_getAutomine",
+    });
+
+    return true;
+  } catch (e) {
+    return false;
+  }
+}
+
+export async function getHardhatProvider(): Promise<EIP1193Provider> {
```
fvictorio, line 20: We should cache this method, otherwise each call to a helper will make one extra JSON-RPC call to check if the network is a Hardhat network.

packages/hardhat-test-helpers/src/utils.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import type { NumberLike } from "./types";
+
+async function checkIfHardhatNetwork(
+  provider: EIP1193Provider
+): Promise<boolean> {
+  try {
+    // we call a method that only exists in the hardhat network
+    await provider.request({
+      method: "hardhat_getAutomine",
```
fvictorio, line 11: I used this method because I was being lazy, but we should use `web3_clientVersion` instead and parse the result.

We should also handle the case where this method doesn't exist, just in case. If it doesn't, then it's not a Hardhat Network and we just return false too.

packages/hardhat-test-helpers/src/utils.ts
morgansliman: Are you sure? I just tested it and it doesn't seem like our RPC methods nor the official EVM methods reject a non-checksum (or lowercase) address parameter. I just tested with an all capitalized address as an example.

packages/hardhat-test-helpers/test/helpers/getStorageAt.ts
morgansliman: Whoops, you're right. That was just me mindlessly copy/paste/editing tests and not giving it as much thought as I should've 😅

packages/hardhat-test-helpers/src/utils.ts
morgansliman: What's the reason for this line at all? I'm trying to understand what the purpose is so I can rewrite the logic, but I'm not sure what exactly it's trying to achieve.

packages/hardhat-test-helpers/src/utils.ts
fvictorio: Yes, but the test helpers don't have to have the same level of strictness that the RPC layer. In some cases we are being stricter (e.g., throwing when an invalid snapshot is used) and in others we are being more relaxed (accepting more kinds of number values). In this case I think it makes sense to add the extra check, since it's a common expectation.

packages/hardhat-test-helpers/src/utils.ts
fvictorio: The context is [this issue](https://github.com/NomicFoundation/hardhat/issues/1585).

When you use ethers's `.toHexString`, the resulting string can be left-padded with an extra zero to make the number of characters even. This means that some values can be something like `0x01` instead of `0x1`. The spec of the JSON-RPC layer [doesn't accept this](https://github.com/ethereum/execution-apis/blob/cd7f22ee84be0b18eab91cf0f00f40af75bd55b1/src/schemas/base-types.json#L42-L46), so Hardhat throws an error to be compliant.

morgansliman (approved):

Issue #1117, opened 2020-12-20
hardhat-etherscan breaks on hardhat/console.sol


ldub:
when running 

```
npx hardhat verify --network mainnet
```

i was getting the error

```
[Error: ENOENT: no such file or directory, open '/Users/lev/dev/ldub/token-2/artifacts/build-info/4e7389b25e253b27f945b011dc8f896b.json'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/lev/dev/ldub/token-2/artifacts/build-info/4e7389b25e253b27f945b011dc8f896b.json'
}
```

this was being thrown by `const buildInfo = await artifacts.getBuildInfo(fqName);` on line https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-etherscan/src/solc/bytecode.ts#L97 when `fqName` was `hardhat/console.sol:console`. 

i added

```
if (fqName === "hardhat/console.sol:console") continue
```

in `token-2/node_modules/@nomiclabs/hardhat-etherscan/dist/src/solc/bytecode.js` and then the script verified just fine 

I am new to hardhat and dont know the root cause of why its even looking at the console but i hope this is sufficient information for @ItsNickBarry or someone else to fix :)

Thanks for the awesome tools all!

ldub:
i should note that i used the `hardhat-dependency-compiler`  plugin and it could be that plugin's bug

alcuadrado:
Hey @ldub, can you create a repo to reproduce this? Thanks!

SirLancelot-OG:
For anyone running into the same issue, `npx hardhat clean` resolve the problem for me.

fvictorio:
Closing this for lack of more info. Please re-open if you can give us instructions on how to reproduce it.

mindqaugas:
I had the smae issue "npx hardhat clean" resolved my problem

ru19wu06:
@SirLancelot-OG  help me a lot!!

MZubairLohar:
> I had the smae issue "npx hardhat clean" resolved my problem

it worked for me as well ! thanks dude

alcuadrado:
I'm reopening this issue and it seems to be pretty common.

Decentra-Ecosystem:
> For anyone running into the same issue, `npx hardhat clean` resolve the problem for me.

Thank you thank you thank you!!!!

chefpickle:
[...]

PR #2507, opened 2022-03-18
removed recompilation from verify task

morgansliman:
This PR is in reference to issue #1117 

I don't believe we should be recompiling just for contract verification, and should instead be using the existing artifacts, however I'm a bit worried changing this now would be a breaking change.

Regardless, I'm opening the PR with this change, and so we have a good place to discuss alternative solutions.

-----

The core problem being fixed (refer to linked issue for reproduction steps from @fvictorio) is that when we change the contract name (or presumably make any change to the contract), the verification pipeline recompiles **only that contract** and not `console.sol`.
This results in the existing artifact debug file for `console.sol` pointing to the old hash of the `contracts/Greeter.sol:Greeter` build-info and not the newly recompiled `contracts/Greeter.sol:Foo` hash and causing the error from the original issue.

### DIFF
--- a/packages/hardhat-etherscan/src/index.ts
+++ b/packages/hardhat-etherscan/src/index.ts
@@ -230,9 +230,6 @@ Possible causes are:
     throw new NomicLabsHardhatPluginError(pluginName, message);
   }
 
-  // Make sure that contract artifacts are up-to-date.
-  await run(TASK_COMPILE);
-
   const contractInformation: ExtendedContractInformation = await run(
     TASK_VERIFY_GET_CONTRACT_INFORMATION,
     {


fvictorio:
> This results in the existing artifact debug file for console.sol pointing to the old hash of the contracts/Greeter.sol:Greeter build-info and not the newly recompiled contracts/Greeter.sol:Foo hash

I think this is a significant bug and we should fix it.

alcuadrado:
> I don't believe we should be recompiling just for contract verification, and should instead be using the existing artifacts

What if the sources were changed after the latest compilation?

fvictorio:
Closing this for now, we'll revisit this issue in a couple of weeks.

Issue #2165, opened 2021-12-16
Several large smart contracts cause `RangeError: Invalid string length` in JSON.stringify


citizen-stig:
# Intro
I have a private Ethereum setup, which has large limit for contract size and large limit for gas per block. I would like to deploy smart contracts there using hardhat.

# Steps to Reproduce

1. Clone repo https://github.com/citizen-stig/hardhat-large-contracts-compilation
2. Run `npm install`
3. Run `npx hardhat compile`

# Expected Results

All contracts are compiled

# Actual Result

Compilation fails with

```
RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at stringify (/Users/ngolub/workspace/hardhat-large-compile/node_modules/jsonfile/index.js:86:18)
    at Object.writeFile (/Users/ngolub/workspace/hardhat-large-compile/node_modules/jsonfile/index.js:101:11)
    at /Users/ngolub/workspace/hardhat-large-compile/node_modules/universalify/index.js:13:12
    at new Promise (<anonymous>)
    at Object.writeFile (/Users/ngolub/workspace/hardhat-large-compile/node_modules/universalify/index.js:7:14)
    at Artifacts.saveBuildInfo (/Users/ngolub/workspace/hardhat-large-compile/node_modules/hardhat/src/internal/artifacts.ts:166:19)
    at SimpleTaskDefinition.action (/Users/ngolub/workspace/hardhat-large-compile/node_modules/hardhat/src/builtin-tasks/compile.ts:793:31)
    at Environment._runTaskDefinition (/Users/ngolub/workspace/hardhat-large-compile/node_modules/hardhat/src/internal/core/runtime-environment.ts:217:14)
    at Environment.run (/Users/ngolub/workspace/hardhat-large-compile/node_modules/hardhat/src/internal/core/runtime-environment.ts:129:14)
 ```
 
 # Environment
 
```
npx hardhat --version
2.8.0
node --version
v16.13.1
```

MacOS, but same result is on Linux

fvictorio:
This seems important. Thanks @citizen-stig for the reproduction example.

There seems to be a limit in V8 that causes `JSON.stringify` to not work with large inputs. We might need to move from plain `JSON.parse`/`JSON.stringify` to some module for doing these operations with streams.

citizen-stig:
> This seems important. Thanks @citizen-stig for the reproduction example.
> 
[...]

PR #2512, opened 2022-03-21
Handle large JSON payloads

Dzejkop:

---


alcuadrado:
Hey @Dzejkop, this PR is interesting, but can you give more info on the motivation behind it? When did you encounter a problem with the existing code?

Thanks!

Dzejkop:
@alcuadrado sure, it's the same case as https://github.com/NomicFoundation/hardhat/issues/2165.

We're working on a private depoloyment with large limits for contract size. We encountered an error with the `JSON.stringify` method which would crash if the contract was too large.

fvictorio:
Not stale

citizen-stig:
@fvictorio Could you please re-open it?

fvictorio:
Hey @Dzejkop, I'm (finally) looking into this PR, and something I would like to do is to actually test it in a scenario that hits the `JSON.stringify`/`JSON.parse` limits.

I tried to create some synthetic examples with a lot of data, but I made solc crash while compiling those. Do you have an example of a scenario where this happens?

citizen-stig:
Hello @fvictorio , there's a repo with crafted example: https://github.com/citizen-stig/hardhat-large-contracts-compilation

I've just checked it with commands from READM.md and it reproduces the issue.

Please let me know if you have been able to reproduce it.

citizen-stig:
Also details are available in the corresponding issue: https://github.com/NomicFoundation/hardhat/issues/2165

fvictorio:
Thanks a lot @citizen-stig, I thought I had seen a reproduction repo in some issue but couldn't find it :man_facepalming:

fvictorio:
Oh, I see. I couldn't reproduce this because it's actually fixed in the latest versions, which save the build info files in a more piece-wise way. We did that for performance reasons, but as a side-effect, it fixed #2165. That's also why there are a lot of merge conflicts in this PR.

I'm going to close this PR now. Sorry for throwing away your work @Dzejkop :disappointed: But the alternative here is to do this from scratch, and that also means adding a new dependency, which we normally try not to do.

@gitpoap-bot @Dzejkop deserves a gitpoap

Dzejkop:
Thanks for the fix @fvictorio!

And thanks for keeping track of this @citizen-stig

PR #2530, opened 2022-03-24
Fix #2528

### LINKED ISSUE #2528: Impossible to run forked node on latest HH version
**Steps to reproduce:**

1. `npm init -y`
2. `npx hardhat` and select to create a config file
3. `npm i hardhat`
4. `npx hardhat node --fork https://eth-mainnet.alchemyapi.io/v2/<key>` as per https://hardhat.org/hardhat-network/guides/mainnet-forking.html#forking-from-mainnet

Output: `Error HH604: Error running JSON-RPC server: Expected 'isActivatedEIP' to only be called with 2930 or 1559`

Never had this error before. I can only assume this is some sort of problem with either the latest Hardhat or the latest Ethereumjs.

alcuadrado:
This PR fixes #2528

### DIFF
--- a/packages/hardhat-core/src/internal/util/hardforks.ts
+++ b/packages/hardhat-core/src/internal/util/hardforks.ts
@@ -1,5 +1,34 @@
 import { assertHardhatInvariant } from "../core/errors";
 
+export const BERLIN_EIPS = new Set([
+  // Homestead
+  2, 7, 8,
+  // Tangerine Whistle
+  150, 158,
+  // Spurious Dragon
+  155, 160, 161, 170,
+  // Byzantium
+  100, 140, 196, 197, 198, 211, 214, 649, 658,
+  // Constantinople
+  101, 105, 123, 145,
+  // Istanbul
+  1108, 1344, 152, 1884, 2028, 2200,
+  // Muir Glacier
+  2384,
+  // Berlin
+  2565, 2718, 2929, 2930,
+]);
+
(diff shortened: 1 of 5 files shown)


packages/hardhat-core/test/internal/hardhat-network/provider/modules/hardhat.ts
```diff
@@ -140,6 +140,34 @@ describe("Hardhat module", function () {
             impersonatedAddress
           );
         });
+
+        it.only("lets you impresonate a contract", async function () {
```
asafyish, line 144: Probably want to remove this .only when pushing.

fvictorio: Good catch @asafyish. Fixed.

packages/hardhat-core/src/internal/util/hardforks.ts
```diff
@@ -1,5 +1,34 @@
...
+  // Homestead
+  2, 7, 8,
+  // Tangerine Whistle
+  150, 158,
+  // Spurious Dragon
+  155, 160, 161, 170,
+  // Byzantium
+  100, 140, 196, 197, 198, 211, 214, 649, 658,
+  // Constantinople
+  101, 105, 123, 145,
```
fvictorio, line 13: Is this correct? According to ethereum.org, it should be:

```suggestion
  1014, 1052, 1234, 145,
```

Issue #2510, opened 2022-03-20
Very large eth_debugTrace responses get dropped


WyseNynja:
If I run a forked node and then try to query debug_traceTransaction for [0x2cb2393747247606d0d534950fcbfe2f787161c909c693189c0aed1dacf1f02b](https://etherscan.io/tx/0x2cb2393747247606d0d534950fcbfe2f787161c909c693189c0aed1dacf1f02b), it thinks for a while and then I get back nothing.

```
$ curl --verbose -H "Content-Type: application/json" --data '{ "jsonrpc": "2.0", "method": "debug_traceTransaction", "params": [ "0x2cb2393747247606d0d534950fcbfe2f787161c909c693189c0aed1dacf1f02b", {  "disableStorage": true } ], "id": 256}' http://localhost:8545
*   Trying 127.0.0.1:8545...
* Connected to localhost (127.0.0.1) port 8545 (#0)
> POST / HTTP/1.1
> Host: localhost:8545
> User-Agent: curl/7.74.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 179
> 
* upload completely sent off: 179 out of 179 bytes
* Empty reply from server
* Connection #0 to host localhost left intact
curl: (52) Empty reply from server
```

If I send this query to my erigon node, it returns a whopping 987 MB of data.

PR #2545, opened 2022-03-29
Improve debug_traceTransaction and add opcodes filter

zemse:
When it comes to memory consumption, the debug_traceTransaction goes brrrrr. For complex enough DeFi transactions, the RPC response is huge as well. This is particularly a show stopper for [hardhat-tracer](https://github.com/zemse/hardhat-tracer).

This PR does the following:
- Keeps data in StructLogs in Buffer format while in internal use, and only converts them into hex strings when exporting to RpcStructLog. So if user has disabled any of the memory, stack, or storage, that Buffer data is not exported to hex string. This is just about 50% memory savings, which is not as significant as the opcodes filter in the next point.
- Adds an opcode filter that can be optionally passed by the user, which prevents exporting struct logs that are not needed. This considerably reduces memory consumption, i.e. at first, memory is allocated when hardhat generates rpc response and then again it is allocated at the receiving end in the same machine. As an example use case of this feature, hardhat-tracer plugin only needs certain opcodes (related to calls, storage, and events) in the struct logs in order to function and a related issue (#2510) may also benefit from this feature.

Sorry for not discussing this first, I was just trying to find cause of the problem and did not know what to do, and ended up with almost the changes in this PR, so I finally thought to just proceed and create a PR. Please let me know any changes that are needed. Finally thanks for creating Hardhat, it's an extremely amazing tool! (ik a thanks is not at all enough but still).

### DIFF
--- a/.changeset/smooth-dryers-clean.md
+++ b/.changeset/smooth-dryers-clean.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Improve debug_traceTransaction memory usage and add opcodes filter to reduce rpc response size for debugging complex transactions.
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/input/debugTraceTransaction.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/input/debugTraceTransaction.ts
@@ -8,6 +8,7 @@ export const rpcDebugTracingConfig = optionalOrNullable(
       disableStorage: optionalOrNullable(t.boolean),
       disableMemory: optionalOrNullable(t.boolean),
       disableStack: optionalOrNullable(t.boolean),
+      opcodes: optionalOrNullable(t.array(t.string)),
     },
     "RpcDebugTracingConfig"
   )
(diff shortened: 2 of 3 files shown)


joshua0702k:
This PR will be very helpful as we are using hardhat-tracer as well. We allow our integrators to deploy a set of mock contracts without worrying about compilation/deployment details. Hardhat-tracer allows them to debug without having to console.log our contracts/recompile. However, with how heavy debug-traces are, even 16gb is not enough to finish the process.

joshpwrk:
still very much interested in getting this merged

VasilyGerrans:
What happened to this PR? `hardhat-tracer` is an incredibly valuable tool for Hardhat debugging. Without this PR, it can't work properly. Hardhat can't be competitive against frameworks with native debug tracing unless it has its own equivalent.

fvictorio:
I believe it was accidentally closed. I'm working on getting as many PRs handled as I can, and this one is certainly very important.

zemse:
I've been working on a workaround to skip using debug_tt. You can try `hardhat-tracer@beta` and checkout [v2-beta branch](https://github.com/zemse/hardhat-tracer/tree/v2-beta) for some docs. 

This implementation hijacks the ethereumjs vm in hardhat and directly gets required debug data from there and hence it's quite memory efficient than usual debug_tt.

PR #2553, opened 2022-03-31
Add return data to JSON-RPC response

fvictorio:
The `data` object returned by the `eth_sendTransaction`, `eth_sendRawTransaction`, `eth_call` and `eth_estimateGas` methods now include a `message` string (which is equal to the top-level message) and a `data` string, with the hex-encoded return data.

Some non-obvious changes here; I will explain them in the comments.

Closes HH-500.


packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -1526,10 +1539,13 @@ export class EthModule {
...
   }
 
+  /**
+   * Returns the trace of the sent tx
+   */
   private async _handleMineBlockResults(
     results: MineBlockResult[],
     sentTx: TypedTransaction
-  ) {
+  ): Promise<GatherTracesResult> {
```
fvictorio, line 1548: I added this to handle the error-throwing logic in the callee. That way, adding the extra fields to the error is done in a single place. Otherwise the error would've been modified both here and in a `catch` in the callee.

alcuadrado (approved):

Issue #2541, opened 2022-03-28
Hardhat node not mining any blocks. How can I get the deployed contract address? 


0xHyoga:
I can't seem to produce blocks like in previous versions of hardhat. But I still see console.log messages when deploying. 

![image](https://user-images.githubusercontent.com/97303883/160467181-00707513-b45b-4e67-994c-76f2f15df71f.png)

I read about some features but as it mentions them the default is to mine blocks. My hardhat config file has not changed since before, so it shouldn't be affected. Is this feature removed?

0xHyoga:
Hello, I have fixed the issue on my end by going back to hardhat 2.9.1 and updating it on my config file.

My other projects I can see had the 2.9.1 version and the console worked perfectly, but all my newer projects were using 2.9.2. I suspect the problem might be there.

Thanks

alcuadrado:
Hi @0xHyoga,

Can you help us reproduce this by posting your config and steps to reproduce it?

Thanks

hussamsh:
I can also confirm this behavior. I had the same problem running `hardhat node` with v2.9.2,  node transaction logs were not appearing even though the transactions themselves were successful.

for instance, deploying a new contract will not print the `eth_sendRawTransaction` log or changing automine with ` network.provider.send("evm_setAutomine", [false])` will not print `evm_setAutomine` to the console.

Reverting to 2.9.1 solved the issue.

Here is how I tested it:
1) On a new VM I installed node/npm, then I created a new hardhat project and installed hardhat v2.9.2 globally with `sudo npm install hardhat@2.9.2 -g`
2) I ran `hardhat node` and tried changing the mining behavior and deploying a dummy contract - no logs were printed.
3) Did the same steps above but with `sudo npm install hardhat@2.9.1 -g` - logs appeared.

0xHyoga:
Hi @alcuadrado after noticing that my old projects were mining the blocks when using `npx hardhat node` I did another test with a brand new hardhat sample project. My configuration file below just in case: 

```
require("dotenv").config();
require("@nomiclabs/hardhat-etherscan");
require("@nomiclabs/hardhat-waffle");
require("hardhat-gas-reporter");
require("solidity-coverage");

// This is a sample Hardhat task. To learn how to create your own go to
// https://hardhat.org/guides/create-task.html
task("accounts", "Prints the list of accounts", async (taskArgs, hre) => {
  const accounts = await hre.ethers.getSigners();

  for (const account of accounts) {
[...]

PR #2556, opened 2022-03-31
expose frozen userConfig in hardhatConfig

morgansliman:
This should fix the bug I introduced in #2477

### DIFF
--- a/packages/hardhat-core/src/internal/core/config/config-loading.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-loading.ts
@@ -109,7 +109,7 @@ export function loadConfigAndTasks(
     checkUnsupportedRemappings(resolved);
   }
 
-  return resolved;
+  return { ...resolved, userConfig: frozenUserConfig };
 }
 
 function deepFreezeUserConfig(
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -109,7 +109,7 @@ export function loadConfigAndTasks(
     checkUnsupportedRemappings(resolved);
   }
 
-  return resolved;
+  return { ...resolved, userConfig: frozenUserConfig };
```
fvictorio, line 112: This seems redundant, since `userConfig` is already added by `resolveConfig`. That is: if I undo this line, then it still compiles fine.

If we _want_ `config.userConfig` to be frozen, then I would pass it as a third parameter to `resolveConfig` and use it there.

alcuadrado:
I think this is generally good, but I would move the `userConfig` field to the hre. Sorry for the confusion, @morgansliman, that's what I had in mind originally. The reason is that the `userConfig` being part of the `config` makes the config system slightly more complex.

alcuadrado:
Fix #2541

alcuadrado (approved):

Issue #2004, opened 2021-10-28
Error: write EPIPE while building `0.4.8` code


sshmaxime:
Error while building `0.4.8` code.

```
Compiling 1 file with 0.4.8
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:94:16)
```

Reproduce bug:
- Go to the CryptoPunksMarket contract and copy it [link](https://etherscan.io/address/0xb47e3cd837ddf8e4c57f05d70ab865de6e193bbb#code). 
- Paste it in a .sol file in your hardhat project.
- Add in your `hardhat.config.ts` this solidity compiler:
```
 {
                version: '0.4.8',
                settings: {
                    optimizer: {
                        enabled: true,
                        runs: 200
                    }
                }
            }
```
- `yarn hardhat compile`

dependencies:
```
        "@nomiclabs/hardhat-ethers": "^2.0.2",
        "@nomiclabs/hardhat-waffle": "^2.0.1",
        "@openzeppelin/contracts": "^4.3.2",
        "@typechain/ethers-v5": "^8.0.0",
        "@typechain/hardhat": "^3.0.0",
        "@types/chai": "^4.2.22",
        "@types/mocha": "^9.0.0",
        "@types/node": "^16.11.6",
        "@typescript-eslint/eslint-plugin": "^5.2.0",
        "@typescript-eslint/parser": "^5.2.0",
        "chai": "^4.3.4",
        "eslint": "^8.1.0",
        "ethereum-waffle": "^3.4.0",
        "ethers": "^5.5.1",
        "hardhat": "^2.6.7",
        "prettier": "^2.4.1",
        "prettier-package-json": "^2.6.0",
        "prettier-plugin-solidity": "^1.0.0-beta.18",
        "ts-node": "^10.4.0",
        "typechain": "^6.0.0",
        "typescript": "^4.4.4"
```

fvictorio:
Closing this one for now. Adding support for versions this old is hard, so we are not planning to do it for now. If someone really needs this, please comment here!

JPiao:
Hi! Im running into the same issue with 0.5.17?

```
$ npx hardhat compile && typechain --target=ethers-v5 "abi/*.json"
Compiling 38 files with 0.5.17
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:98:16)
error Command failed with exit code 1.
```

[...]

PR #2559, opened 2022-04-01
Fix smart contract compilation when the downloaded solc is for WASM platform.

### LINKED ISSUE #2329: Compile fails  for 0.5.17
Hey team,

I recently upgraded to the M1 and my compile seems to be breaking. 

Initially it's because theres no M1 solc build for the M1 architecture but I believe this is the work around https://github.com/nomiclabs/hardhat/issues/2125

Here is my config:

```
subtask<{ solcVersion: string }>(TASK_COMPILE_SOLIDITY_GET_SOLC_BUILD, async (args, hre, runSuper) => {
  if (args.solcVersion === "0.5.17") {
    const compilerPath = path.join(__dirname, "0.5.17+commit.d19bba13.Darwin.appleclang");

    return {
      compilerPath,
      isSolcJs: false, // if you are using a native compiler, set this to false
      version: args.solcVersion,
      // this is used as extra information in the build-info files, but other than
      // that is not important
      longVersion: "0.5.17+commit.d19bba13.Darwin.appleclang"
    }
  }

  // we just use the default subtask if the version is not 0.8.5
  return runSuper();
})

compilers: [
      {
        version: "0.5.17",
        settings: {
          optimizer: {
            enabled: true,
            runs: 200,
          },
        },
      },
      {
        version: "0.8.4",
        settings: {
          optimizer: {
            enabled: true,
            runs: 200,
          },
        },
      },
    ]
```

However, when I try to npx hardhat comile, I run into

```
Compiling 38 files with 0.5.17
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:98:16)
```

I found this issue and was wondering if it was related https://github.com/nomiclabs/hardhat/issues/2004? I believe 0.5.0 should be supported by hardhat?

Thanks!

### LINKED ISSUE #2554: `solcjs` is not invoked correctly on platforms where there isn't an available `solc` binary
I'm trying to run `hardhat` on FreeBSD, for which there isn't an official `solc` binary.
When I run `npx hardhat compile` it correctly download the solcjs file into `$HOME/.cache/hardhat-nodejs/compilers/wasm/soljson-v0.8.4+commit.c7e474f2.js`, but it tries to run it as a native binary instead of using the javascript wrapper, failing miserably with the following error:
```
Error: write EPIPE
    at afterWriteDispatched (internal/stream_base_commons.js:156:25)
    at writeGeneric (internal/stream_base_commons.js:147:3)
    at Socket._writeGeneric (net.js:798:11)
    at Socket._write (net.js:810:8)
    at writeOrBuffer (internal/streams/writable.js:358:12)
    at Socket.Writable.write (internal/streams/writable.js:303:10)
    at output (node_modules/hardhat/src/internal/solidity/compiler/index.ts:73:22)
    at new Promise (<anonymous>)
    at NativeCompiler.compile (node_modules/hardhat/src/internal/solidity/compiler/index.ts:59:34)
    at SimpleTaskDefinition.action (node_modules/hardhat/src/builtin-tasks/compile.ts:665:37)
```

alexdupre:
When the downloaded solidity compiler is the WASM version (because there is no available native compiler for the selected platform) it is still run as a native compiler, because of the missing else statement that set the correct platform.

Closes #2554 and fixes #2329 and #2004 as well.

### DIFF
--- a/.changeset/hungry-yaks-enjoy.md
+++ b/.changeset/hungry-yaks-enjoy.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix a bug preventing to run the solcjs compiler.
--- a/packages/hardhat-core/src/builtin-tasks/compile.ts
+++ b/packages/hardhat-core/src/builtin-tasks/compile.ts
@@ -590,6 +590,8 @@ subtask(TASK_COMPILE_SOLIDITY_GET_SOLC_BUILD)
             log("Native solc binary doesn't work, using solcjs instead");
             nativeBinaryFailed = true;
           }
+        } else {
+          platform = CompilerPlatform.WASM;
         }
       }
 


alcuadrado:
Thanks @alexdupre!

I'm pretty sure this is correct. Can you take a quick look, @fvictorio ?

fvictorio:
Good catch @alexdupre, thank you!

PR #2561, opened 2022-04-02
Allow to stop time flow and mine blocks with the same timestamp.

### LINKED ISSUE #2560: It's not possible to mine multiple blocks with the same timestamp
There are EVM blockchains with sub-second block production frequency. It'd be very useful if hardhat could be configured to mine blocks with the same timestamp instead of always requiring increasing timestamps.

alexdupre:
This allows to emulate EVM blockchains with sub-second block creation
time, and have accurate time based tests where the timestamp is only
increased manually.

Closes #2560


alexdupre:
Both seem spurious CI failures.

alexdupre:
I'm still interested in getting it merged.

alexdupre:
Toc! Toc!

alexdupre:
...

fvictorio:
Hey @alexdupre, sorry for the extremely late response. I think https://github.com/NomicFoundation/hardhat/pull/3228 covers the same need that this PR, doesn't it?

alexdupre:
Well, it's not exactly the same, but better than nothing. If I'm reading it correctly, enabling the `allowBlocksWithSameTimestamp` flag has the effect that all close transactions will end up in different blocks with the same timestamp, breaking the previous assumption that timestamps were different. In other words, it looks like a global on/off, while with my PR it was possible to enable/disable it only for specific scenarios.

fvictorio:
> In other words, it looks like a global on/off, while with my PR it was possible to enable/disable it only for specific scenarios.

I think that's accurate, yes. I guess we can also add some RPC methods to change the behavior on runtime.

alexdupre:
> I guess we can also add some RPC methods to change the behavior on runtime.

That would be appreciated and I think will fulfill my needs.

alexdupre:
> I guess we can also add some RPC methods to change the behavior on runtime.

@fvictorio I've seen you merged PR #3228, are you planning to add a RPC method, too?

fvictorio:
Hey @alexdupre, thanks for the reminder, I forgot about this :man_facepalming:

I opened https://github.com/NomicFoundation/hardhat/issues/3608 and I'll try to implement it as soon as I can. Hopefully we can include the option and the helper in the same release.

I'm going to close this PR now, thanks again!

PR #2563, opened 2022-04-02
Chain ID does not match the chain ID of Common

biddls:
Had this error but couldn't find anything on the web for it so I thought that this might help someone else later on

---

added a bug fix I didn't find anything googling for, thought it might help


fvictorio:
@biddls do you have reproduction steps for the underlying problem? Maybe it's something that can be fixed instead of working around it with the chain id setting.

biddls:
Hi, yea sorry I can't remember the reproduction steps. Instead of being an error to fix I feel like it's a user error that needs a better error message to tell the user what the error may be. I posted about it in the discord server [here](https://discord.com/channels/750408878008827925/750408878008827928/959833558791385158) 

Hope it helps

PR #2565, opened 2022-04-05
Add hardhat-forta community plugin

haseebrabbani:
Adding hardhat-forta community plugin


alcuadrado:
Thanks @haseebrabbani !

Issue #2230, opened 2022-01-08
The slot parameter of `eth_getStorageAt` should accept hex encoded unsigned integers


fvictorio:
We are expecting the storage slot parameter of `getStorageAt` to be a `QUANTITY`, but the spec says that the parameter should be:

```
title  hex encoded unsigned integer
type  string
pattern  ^0x[0-9a-f]{64}$
```

Notice that the value **should** have a length of 64, but we'll reject many of those (because values with leading zeros are not a valid `QUANTITY`)

I don't know if we can start rejecting shorter values though. On one hand, we try to be as spec-compliant as possible, but in the other hand that would be a breaking change.

meetmangukiya:
Maybe we should just support `pattern  ^0x[0-9a-f]{0,64}$` until next major version update so we cover all the cases while also fixing the current bug?

meetmangukiya:
@fvictorio is there any easy workaround known for this for now until this gets fixed? Turn off the validation somehow?

fvictorio:
@meetmangukiya no workaround AFAIK, but I'll try to get this prioritized.

RomiRand:
Bump, I'd love to see this fixed :slightly_smiling_face:

brockelmore:
Did https://github.com/NomicFoundation/hardhat/pull/2581 close this?

fvictorio:
Yes! Thanks for the reminder.

PR #2581, opened 2022-04-08
Make eth_getStorageAt spec-compliant

fvictorio:
According to [the spec](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/execution-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false), the storage slot in `eth_getStorageAt` **should** have a length of 66 (0x + 32 bytes).

While this is technically a bug fix, it will probably break things for many people. There are two solutions:

- To use `helpers.getStorageAt`, since the network helpers will be released along with this change (or soon enough).
- To use `ethers.utils.hexZeroPad(storageSlot, 32)` around the value to make it valid

### DIFF
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
@@ -29,6 +29,41 @@ export const rpcHash = new t.Type<Buffer>(
   t.identity
 );
 
+export const rpcStorageSlot = new t.Type<BN>(
+  "Storage slot",
+  BN.isBN,
+  validateStorageSlot,
+  t.identity
+);
+
+function validateStorageSlot(u: unknown, c: t.Context): t.Validation<BN> {
+  if (typeof u !== "string") {
+    return t.failure(
+      u,
+      c,
+      `Storage slot argument must be a string, got '${u}'`
+    );
+  }
+
+  if (u.match(/^0x(?:[0-9a-fA-F]*)*$/) === null) {
+    return t.failure(
+      u,
+      c,
+      `Storage slot argument must be a valid hexadecimal, got '${u}'`
+    );
+  }
+
+  if (u.length !== 66) {
+    return t.failure(
+      u,
+      c,
+      `Storage slot argument must have a length of 66 (32 bytes), but '${u}' has a length of ${u.length}`
+    );
+  }
+
+  return t.success(new BN(toBuffer(u)));
+}
+
 export const rpcAddress = new t.Type<Buffer>(
   "ADDRESS",
   (v): v is Buffer => Buffer.isBuffer(v) && v.length === ADDRESS_LENGTH_BYTES,
@@ -86,6 +121,15 @@ export function numberToRpcQuantity(n: number | BN): string {
   );
 
   return `0x${n.toString(16)}`;
+}
+
+export function numberToRpcStorageSlot(n: number | BN): string {
+  assertHardhatInvariant(
+    typeof n === "number" || BN.isBN(n),
+    "Expected number"
+  );
+
+  return `0x${n.toString(16).padStart(64, "0")}`;
 }
 
 /**
--- a/packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/getStorageAt.ts
+++ b/packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/getStorageAt.ts
@@ -1,7 +1,11 @@
 import { assert } from "chai";
 
-import { numberToRpcQuantity } from "../../../../../../../src/internal/core/jsonrpc/types/base-types";
+import {
+  numberToRpcQuantity,
+  numberToRpcStorageSlot,
+} from "../../../../../../../src/internal/core/jsonrpc/types/base-types";
 import { workaroundWindowsCiFailures } from "../../../../../../utils/workaround-windows-ci-failures";
+import { assertInvalidArgumentsError } from "../../../../helpers/assertions";
 import { EXAMPLE_CONTRACT } from "../../../../helpers/contracts";
 import { setCWD } from "../../../../helpers/cwd";
 import {
@@ -39,7 +43,7 @@ describe("Eth module", function () {
               assert.strictEqual(
                 await this.provider.send("eth_getStorageAt", [
                   exampleContract,
-                  numberToRpcQuantity(3),
+                  numberToRpcStorageSlot(3),
                 ]),
                 "0x0000000000000000000000000000000000000000000000000000000000000000"
               );
@@ -47,7 +51,7 @@ describe("Eth module", function () {
               assert.strictEqual(
                 await this.provider.send("eth_getStorageAt", [
                   exampleContract,
-                  numberToRpcQuantity(4),
+                  numberToRpcStorageSlot(4),
                 ]),
                 "0x0000000000000000000000000000000000000000000000000000000000000000"
               );
@@ -55,7 +59,7 @@ describe("Eth module", function () {
               assert.strictEqual(
                 await this.provider.send("eth_getStorageAt", [
                   DEFAULT_ACCOUNTS_ADDRESSES[0],
-                  numberToRpcQuantity(0),
+                  numberToRpcStorageSlot(0),
                 ]),
                 "0x0000000000000000000000000000000000000000000000000000000000000000"
               );
@@ -74,7 +78,7 @@ describe("Eth module", function () {
                 assert.strictEqual(
                   await this.provider.send("eth_getStorageAt", [
                     exampleContract,
-                    numberToRpcQuantity(2),
+                    numberToRpcStorageSlot(2),
                     numberToRpcQuantity(firstBlock),
                   ]),
                   "0x0000000000000000000000000000000000000000000000000000000000000000"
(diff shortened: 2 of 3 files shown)


packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
```diff
@@ -29,6 +29,41 @@ export const rpcHash = new t.Type<Buffer>(
...
+      c,
+      `Storage slot argument must be a valid hexadecimal, got '${u}'`
+    );
+  }
+
+  if (u.length !== 66) {
+    return t.failure(
+      u,
+      c,
+      `Storage slot argument must have a length of 66 (32 bytes), but '${u}' has a length of ${u.length}`
```
fvictorio, line 60: I added more useful and specific error messages here to help people that are using this method with shorter values, which we were (wrongly) accepting before.

aathan:
This fix does not resolve the issue I'm having in Foundry, which is related to failed eth_getStorageAt calls in hardhat vs direct full-node RPC.  I installed this branch and ran the problem test and I get the following. I also reported this to Foundry at https://github.com/gakonst/foundry/issues/1297

```
eth_getStorageAt

  Errors encountered in param 1: Invalid value "0x0000000000000000000000000000000000000000000000000000000000000003" supplied to : QUANTITY
```

aathan:
I think maybe the problem is from the calls to numberToRpcQuantity() in the eth_getStorageAt implementation itself, in getStorageAt.ts?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/getStorageAt.ts
```diff
@@ -229,14 +233,73 @@ describe("Eth module", function () {
...
+                "0x0101010101010101010101010101010101010101",
+                "0x0000000000000000000000000000000000000000000000000000000000000000",
+              ]),
+              "0x0000000000000000000000000000000000000000000000000000000000000000"
+            );
+
+            assert.strictEqual(
+              await this.provider.send("eth_getStorageAt", [
+                "0x0101010101010101010101010101010101010101",
+                "0xcd39aa866fd639607c7241f617cf83f33c646551f3d205f2905c5abacca2db85",
```
feuGeneA, line 258: Do these bytestrings have any specific relevance? How does the expected output relate to these inputs?

packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
feuGeneA: I think you could also expand this a little more, just to be a little more educational:
```suggestion
      `Storage slot argument must have a length of 66 (32 bytes (2 characters per byte) plus the "0x" prefix), but '${u}' has a length of ${u.length}`
```

feuGeneA (approved):
Left a suggestion and a question, this LGTM without necessarily resolving those.

packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
```diff
@@ -29,6 +29,41 @@ export const rpcHash = new t.Type<Buffer>(
...
+      c,
+      `Storage slot argument must be a string, got '${u}'`
+    );
+  }
+
+  if (u.match(/^0x(?:[0-9a-fA-F]*)*$/) === null) {
+    return t.failure(
+      u,
+      c,
+      `Storage slot argument must be a valid hexadecimal, got '${u}'`
```
alcuadrado, line 52:
```suggestion
      `Storage slot argument must be a valid hexadecimal prefixed with "0x", got '${u}'`
```

packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
alcuadrado:
```suggestion
      `Storage slot argument must have a length of 66 ("0x" + 32 bytes), but '${u}' has a length of ${u.length}`
```

alcuadrado (approved):
Just left a few comments, but LGTM

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/getStorageAt.ts
fvictorio: Just a random storage slot, and an address that is (almost surely) empty. Added some comments explaining this.

fvictorio:
(Just a reminder for myself: we are not merging this until the network helpers are released)

fvictorio:
This is now released in `hardhat@2.9.5`. This change means that the `eth_getStorageAt` method is harder to use because the storage slot has to be fully padded. As a better alternative to this, you can use our [network helpers](https://nomiclabs.notion.site/Public-Hardhat-Network-Helpers-216cf3c665124895bf1f6e2f9343f86f) package (beta), that makes this kind of interaction easier.

mattsse:
> According to [the spec](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/execution-apis/assembled-spec/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false), the storage slot in eth_getStorageAt should have a length of 66 (0x + 32 bytes).

this is a bit confusing to me because this spec https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getstorageat says it's a [QUANTITY](https://ethereum.org/en/developers/docs/apis/json-rpc/#quantities-encoding)...

what am I missing?

alcuadrado:
Hey @mattsse!

We are trying to follow the [OpenRPC-based spec](https://github.com/ethereum/execution-apis), but it can have mistakes. When in doubt we try to copy whatever Geth does tbh. We want to avoid users thinking that what they are doing should work, just to realize it doesn't when going into production and switching to Geth. 

Note that in `eth_getStorageAt` [the spec uses an `uint256`](https://github.com/ethereum/execution-apis/blob/00ccfc66b3e1260ae9de2a4d4138156754bebff5/src/eth/state.yaml#L26), and not an `uint` (the equivalent of `QUANTITY`). ([example with `uint`](https://github.com/ethereum/execution-apis/blob/00ccfc66b3e1260ae9de2a4d4138156754bebff5/src/eth/state.yaml#L49) instead)

This was also surprising to me, as before that spec we also used that description of the JSON-RPC as our first-line source of truth, so I just re-checked it.

In Geth [it uses a `hash` type](https://github.com/ethereum/go-ethereum/blob/b628d7276624c2d8ea7dd97d2259a2c2fce7d3cc/interfaces.go#L93), which is 32 bytes. I guess that when [they were writing this spec of this method](https://github.com/ethereum/execution-apis/commit/2a8f5c9c5d8f3e46d00efd0bc89533b888a253a5) `uint256` was introduced to differentiate it from an actual hash.

mattsse:
I see, this is clear now, thank you!

aathan:
Hi guys, I've been following this thread since back in April when I reported a related issue to both the foundry and hardhat projects (see my comments on this bug in April). As a result I've dug into all of the history and my conclusion is that I believe there was a misunderstanding of the OpenRPC spec here on hardhat's side.  Specifically:

https://github.com/NomicFoundation/hardhat/issues/2230

contains a reference to a spec that states:

```
title  hex encoded unsigned integer
type  string
pattern  ^0x[0-9a-f]{64}$
```

However, the spec referenced at the opening of #2581 (scroll to top) says:

schema

title
hex encoded unsigned integer

```
pattern
^0x[0-9a-f]{0,64}$
```

... and from https://perldoc.perl.org/perlre we have `{n,m}       Match at least n but not more than m times`

The issue of the hash vs uint vs uint256 type in the implementation seems distinct from what SHOULD be accepted by the RPC encoding specification.  There is no source provided for the spec quoted in #2230, so it's not clear which spec is the correct one. That being said `{0,64}` also seems wrong, as this would allow `0x` as valid. `{1,64}` seems to be what was intended there.

Has anyone found what geth actually accepts as valid in its ingress handler for RPC requests?

aathan:
I looked at geth and it does *not* appear to enforce a length on hash types:

```
func (s *BlockChainAPI) GetStorageAt(ctx context.Context, address common.Address, key string, blockNrOrHash rpc.BlockNumberOrHash) (hexutil.Bytes, error) {
	state, _, err := s.b.StateAndHeaderByNumberOrHash(ctx, blockNrOrHash)
	if state == nil || err != nil {
		return nil, err
	}
	res := state.GetState(address, common.HexToHash(key))
	return res[:], state.Error()
}
```

which eventually calls:
```
func BytesToHash(b []byte) Hash {
	var h Hash
	h.SetBytes(b)
	return h
}
```

which is implemented like this:
```
// SetBytes sets the hash to the value of b.
// If b is larger than len(h), b will be cropped from the left.
func (h *Hash) SetBytes(b []byte) {
	if len(b) > len(h) {
		b = b[len(b)-HashLength:]
	}

	copy(h[HashLength-len(b):], b)
}
```

alcuadrado:
wow, thanks for digging deeper into this! This situation has been a whole mess and I unfortunately collaborated to it. You are right, and geth's implementation doesn't validate anything. Just look at this, wtf

image

PR #2607, opened 2022-04-13
Add time helpers and tests

zoeyTM:
Fixes HH-462

### DIFF
--- a/packages/hardhat-test-helpers/src/helpers/time/duration/seconds.ts
+++ b/packages/hardhat-test-helpers/src/helpers/time/duration/seconds.ts
@@ -1,6 +1,6 @@
 /**
  * Converts seconds into seconds (noop)
  */
-export function seconds(n: number = 1): number {
+export function seconds(n: number): number {
   return n;
 }


packages/hardhat-test-helpers/test/helpers/time/setNextBlockTimestamp.ts
```diff
@@ -0,0 +1,55 @@
...
+    await hh.mine();
+
+    const endHeight = await hh.time.latestBlock();
+    const endTimestamp = await hh.time.latest();
+
+    assert.equal(initialHeight + 1, endHeight);
+    assert.equal(newTimestamp, endTimestamp);
+  });
+
+  it("should set the next block to the given timestamp [Date]", async function () {
```
morgansliman, line 33: One thing I noticed about this test and the one above it is that, when running only this specific suite via `describe.only("time#setNextBlockTimestamp"...`, the tests pass as expected.

However, when running the full suite of tests for the test helpers, these two tests fail, specifically on the second assert in each test where the `endTimestamp` is 1 millisecond larger than `newTimestamp` in both cases.

My assumption is that this is due to system lag, and that the solution would simply be to assert that the timestamps are within 1-2 ms of each other, rather than exactly equal. Before doing that, though, I wanted another set of eyes on it to ensure that there wasn't actually a bug somewhere that I missed. 

Also worth noting is that these tests take significantly longer to complete in the full suite than they do when running exclusively their own suite.

fvictorio: The problem was that the helper was using `evm_increaseTime` instead of `evm_setNextBlockTimestamp`. I changed that.

`evm_increaseTime(5)` doesn't mean "set the next timestamp to the previous timestamp, plus 5". Instead, it means "simulate that 5 seconds have passed". That is: they are added to whatever time has really passed so far.

This is easier to understand with an example: if the current block timestamp is X, you call `increaseTime(5)`, wait 10 seconds, and mine a block, the new timestamp will be X + 15, not X + 5.

morgansliman: I see, that makes sense.

PR #2608, opened 2022-04-13
Fix passing options and depth when inspecting a lazyObject

### LINKED ISSUE #1473: Hardhat console hangs when using lazyObject with a circular dependency that involves the HRE
To trigger the bug, use this hardhat config:

```js
class Foo {
  constructor(hre) {
    this.hre = hre;
  }
}

extendEnvironment(hre => {
  hre.foo = lazyObject(() => new Foo(hre))
});
```

Then start `hh console` and evaluate the `hre`. Note that if you are using node >= 12, then just typing "hre" in the console will trigger this bug.

---

I'm pretty sure the problem is here:

https://github.com/nomiclabs/hardhat/blob/d59ef54e108f3ab2891dd963f2f9913c073ab0f5/packages/hardhat-core/src/internal/util/lazy.ts#L39-L42

Replacing that function with `return ""` makes the problem go away. There's something in the intersection of `lazyObject`, the HRE and `util.inspect` that causes this.

Notice that the snippet above is the minimal reproduction as far as I can tell. A simple cicrularity with `lazyObject` doesn't cause this:

```js
extendEnvironment(hre => {
  hre.foo = lazyObject(() => {
    let foo = {}
    foo.foo = foo;

    return foo
  })
}
```

And a circularity with the hre but without `lazyObject` doesn't cause it either:

```js
extendEnvironment(hre => {
  hre.foo = new Foo(hre)
}
```

pcarranzav:
...or lazyFunction

---

Hi! This PR changes the way we call `util.inspect` in lazy objects and functions:
- The current state is that `util.inspect` is called without any additional arguments.
- From [node docs](https://nodejs.org/api/util.html#custom-inspection-functions-on-objects), the custom inspect function is called with three arguments `(depth, options, inspect)` so we can use these when calling inspect on the real object/function.
- In this PR we pass those options, including the depth, to the inspect call. This avoids losing any options but most importantly avoids losing the depth, which would otherwise make us recurse infinitely if we had a lazyObject within a circular object.
- Because of this, I believe this fixes #1473.

### DIFF
--- a/.changeset/chilled-cycles-smile.md
+++ b/.changeset/chilled-cycles-smile.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix passing options and depth when inspecting a lazyObject or lazyFunction
--- a/packages/hardhat-core/src/internal/util/lazy.ts
+++ b/packages/hardhat-core/src/internal/util/lazy.ts
@@ -1,4 +1,4 @@
-import { InspectOptions } from "util";
+import util, { InspectOptions } from "util";
 
 import { HardhatError } from "../core/errors";
 import { ERRORS } from "../core/errors-list";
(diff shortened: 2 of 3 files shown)


alcuadrado:
Thanks, @pcarranzav! I fixed a bug triggered by older versions of Node. Will merge when the CI passes.

pcarranzav:
Thanks @alcuadrado! (And sorry I missed that bug!)

PR #2609, opened 2022-04-14
DOC: Update required nodejs version to v16 in setup document

gaga5lala:
Update nodejs version to Active LTS (v16) to prevent complaint `You are using a version of Node.js that is not supported by Hardhat, and it may work incorrectly, or not work at all.`.

https://nodejs.org/en/about/releases/

### DIFF
--- a/docs/tutorial/setting-up-the-environment.md
+++ b/docs/tutorial/setting-up-the-environment.md
@@ -8,7 +8,7 @@ Most Ethereum libraries and tools are written in JavaScript, and so is **Hardhat
 
 ## Installing Node.js
 
-You can [skip](./creating-a-new-hardhat-project.md) this section if you already have a working Node.js `>=16.0` installation. If not, here's how to install it on Ubuntu, MacOS and Windows.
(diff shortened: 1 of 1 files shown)


docs/tutorial/setting-up-the-environment.md
```diff
@@ -8,7 +8,7 @@ Most Ethereum libraries and tools are written in JavaScript, and so is **Hardhat
 
 ## Installing Node.js
 
-You can [skip](./creating-a-new-hardhat-project.md) this section if you already have a working Node.js `>=12.0` installation. If not, here's how to install it on Ubuntu, MacOS and Windows.
+You can [skip](./creating-a-new-hardhat-project.md) this section if you already have a working Node.js `>=16.0` installation. If not, here's how to install it on Ubuntu, MacOS and Windows.
```
alcuadrado, line 11:
```suggestion
You can [skip](./creating-a-new-hardhat-project.md) this section if you already have a working Node.js `>=12.0` installation. If not, here's how to install it on Ubuntu, MacOS and Windows.
```

This is the only thing that should be kept in 12

PR #2627, opened 2022-04-20
properPrivateKey, properAddress, and hexEquals chai matchers

morgansliman:
Fixes HH-499

### DIFF
--- a/packages/hardhat-chai-matchers/src/properAddress.ts
+++ b/packages/hardhat-chai-matchers/src/properAddress.ts
@@ -1,0 +1,12 @@
+export function supportProperAddress(Assertion: Chai.AssertionStatic) {
+  Assertion.addProperty("properAddress", function (this: any) {
+    const subject = this._obj;
+    this.assert(
+      /^0x[0-9-a-fA-F]{40}$/.test(subject),
+      `Expected "${subject}" to be a proper address`,
+      `Expected "${subject}" not to be a proper address`,
+      "proper address (eg.: 0x1234567890123456789012345678901234567890)",
+      subject
+    );
+  });
+}
--- a/packages/hardhat-chai-matchers/src/properPrivateKey.ts
+++ b/packages/hardhat-chai-matchers/src/properPrivateKey.ts
@@ -1,0 +1,12 @@
+export function supportProperPrivateKey(Assertion: Chai.AssertionStatic) {
+  Assertion.addProperty("properPrivateKey", function (this: any) {
+    const subject = this._obj;
+    this.assert(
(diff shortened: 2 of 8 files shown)


fvictorio (approved):

packages/hardhat-chai-matchers/test/properPrivateKey.ts
```diff
@@ -0,0 +1,45 @@
+import { expect, AssertionError } from "chai";
+
+import "../src";
+
+/* eslint-disable @typescript-eslint/no-unused-expressions */
+
+describe("Proper private", () => {
```
feuGeneA, line 7: Can you please update all these test descriptions (and `it` blocks) to say "private key" instead of just "private"?

packages/hardhat-chai-matchers/src/properAddress.ts
```diff
@@ -0,0 +1,12 @@
+export function supportProperAddress(Assertion: Chai.AssertionStatic) {
+  Assertion.addProperty("properAddress", function (this: any) {
+    const subject = this._obj;
+    this.assert(
```
feuGeneA, line 4: I wonder if this shouldn't be split into two separate assertions, one asserting that it's the proper length, and the other asserting that it has the proper characters, so that we can tell the user "it's not the right length" or "it has non-hex characters" etc.

packages/hardhat-chai-matchers/src/properPrivateKey.ts
```diff
@@ -0,0 +1,12 @@
+export function supportProperPrivateKey(Assertion: Chai.AssertionStatic) {
+  Assertion.addProperty("properPrivateKey", function (this: any) {
+    const subject = this._obj;
+    this.assert(
```
feuGeneA, line 4: I wonder if this shouldn't be split into two separate assertions, one asserting that it's the proper length, and the other asserting that it has the proper characters, so that we can tell the user "it's not the right length" or "it has non-hex characters" etc.

packages/hardhat-chai-matchers/src/properAddress.ts
fvictorio: Agree in principle, but tbh we are mainly adding these matchers for compatibility reasons. I don't think they are used a lot, so I wouldn't invest much more time on them.

packages/hardhat-chai-matchers/test/properPrivateKey.ts
fvictorio: Fixed in [b07ff26](https://github.com/NomicFoundation/hardhat/pull/2627/commits/b07ff26264bbdec399261bb28f754ada0875bb46).

Issue #2591, opened 2022-04-10
The wrong chain id is used when running `eth_call` and forking a remote network


alcuadrado:
The chain ID is always the same as the remote network.

See this minimal reproduction https://github.com/alcuadrado/wrong-chain-id-on-calls

egillh210:
Came here to report this same issue. Was implementing ERC20 permits, which surfaced this issue. Metamask doing a eth_estimateGas static call, which would fail due to the DOMAIN_SEPERATOR not matching the cache due to the block.chainid mismatch. Very painful to debug 😭

Would be very greatful if you guys could get this fix merged soon 🤝

Could you confirm that `eth_estimateGas` calls follow the same code-path as `eth_call`, that is `_runTxAndRevertMutations`?

feuGeneA:
@egillh210 
> Could you confirm that `eth_estimateGas` calls follow the same code-path as `eth_call`, that is `_runTxAndRevertMutations`?

Yes, it does

rodrigoherrerai:
Hello, I'm using the latest version of Hardhat and a pretty  weird error is happening regarding this topic. If I am forking mainnet and call a "getChainId()" function from the test suits, I get the correct answer. But, from the inside the contract, I always get 31337, even if I am forking mainnet.

PR #2631, opened 2022-04-21
Fix chain ID when running in a newly-created block on a forked chain

feuGeneA:
For #2591, corresponding to https://linear.app/nomic-foundation/issue/HH-532/the-wrong-chain-id-is-used-when-running-eth-call-and-forking-a-remote

- [x] Find and propose fix.
- [x] Add a test.


alcuadrado (approved):
LGTM

PR #2636, opened 2022-04-22
Catch nested usages of `loadFixture`

fvictorio:
Also, use only custom errors in the package. This was in part a pre-requisite to catch nested `loadFixture`s, but it also makes the code more readable and produces nicer error messages.


morgansliman (approved):
Love it! LGTM

Issue #2266, opened 2022-01-16
The node/recommended plugin in the advanced typescript template breaks eslint


ArshanKhanifar:
# What's Wrong
The `node/recommended` plugin only does `js` import files, it complains with `node/no-missing-import` messages if the module is a typescript module. 

# Steps to Reproduce
1. make a new hardhat typescript project
2. make a new `hello.ts` file under `./scripts/hello.ts`
```
export const hello = () => console.log("hello");
```
3. import that file in the default `./scripts/deploy.ts` file:
```
import { ethers } from "hardhat";
import { sayshi } from "./sayshi";

async function main() {
// etc. etc.
```
4. At this point, my ide (Webstorm) complains: 

![image](https://user-images.githubusercontent.com/10492324/149679811-c02c197b-f666-44a3-baf2-f84749ee3e6b.png)
If you try running eslint, it will complain too... 

```
yarn eslint scripts/*.ts
# yada yada
/Users/arshankhanifar/hardhat-debug/scripts/deploy.ts
  7:23  error  "./hello" is not found  node/no-missing-import

✖ 1 problem (1 error, 0 warnings)
```

# Repo That Shows This
Here's a [repo that shows this](https://github.com/ArshanKhanifar/hardhat-bug-demo)

# Proposed Fix
Maybe use `plugin:@typescript-eslint/recommended` instead? That's what [nounsdao ](https://github.com/nounsDAO/nouns-monorepo/blob/master/.eslintrc.js#L9) is doing.

ArshanKhanifar:
Sorry couldn't find the typescript template repo, so thought I'd put it here...

vcheeney:
@ArshanKhanifar Thank you! I was experiencing a similar problem while importing the type of my token. Removing the _plugin:node/recommended_ plugin from the ESLint extends array resolved the linting issue.

alcuadrado:
Thanks for opening this issue, @ArshanKhanifar.

the-vampiire:
is this the intended behavior for a typescript project? can the codegen be updated to include the correct plugin?

PR #2638, opened 2022-04-22
advanced-ts: fix eslint plugin breaking ts imports

the-vampiire:

---

fixes https://github.com/NomicFoundation/hardhat/issues/2266

in the typescript starter it is using `plugin:node/recommended` for eslint which breaks every import done across files (see linked issue). this is strange considering typescript projects are already installing the ts-eslint rules.

some in the thread suggested removing the rules entirely but i have found that replacing them with the `@typescript-eslint/recommended` is a drop-in replacement.

### DIFF
--- a/packages/hardhat-core/sample-projects/advanced-ts/.eslintrc.js
+++ b/packages/hardhat-core/sample-projects/advanced-ts/.eslintrc.js
@@ -9,7 +9,7 @@ module.exports = {
   extends: [
     "standard",
     "plugin:prettier/recommended",
-    "plugin:node/recommended",
+    "plugin:@typescript-eslint/recommended",
   ],
   parser: "@typescript-eslint/parser",
   parserOptions: {


the-vampiire:
in addition i recommend adding https://www.npmjs.com/package/eslint-plugin-chai-friendly

this will prevent the "no unused expressions" for things like `.to.be.true`

the-vampiire:
i still think this is worth considering. easy PR to merge - only changes a dev dep. improves devx

jakehockey10:
Is there a workaround while this is still be considered?

fvictorio:
Hey @the-vampiire, since you sent this PR we've changed our sample projects. They no longer include an `eslint` config, that's why there are merge conflicts here.

Since no having eslint anymore makes the problem pointed by you no longer relevant, I'm going to close this PR. Thanks a lot for sending it and sorry for not responding sooner!

PR #2644, opened 2022-04-26
Add `.reverted` matcher

fvictorio:
I tested this with Uniswap V3's repo (after commenting out the `.to.emit` and `.revertedWith` assertions) and it worked fine. Coverage of the `src/matchers.ts` file is 100%. I also did some "manual mutation testing" to be sure that the tests are correct.

Reviewing the diff probably doesn't make sense. I suggest just reading the `src/reverted.ts` and `test/reverted.ts` files from scratch.


fvictorio:
Closed in favor of #2652.

PR #2656, opened 2022-04-29
Support `.deep.equal` numerical comparisons

fvictorio:
Only `.deep.equal` is supported. There are other matchers that use the `.deep` modifier (like `.include` for example), but I think supporting just `.equal` is fine for now.

The implementation is similar to what chai itself does, but simplified and with support for bigints/big numbers.

This also fixes an issue with the current implementation: custom messages were not being respected. So for example if you did:

```
expect(1).to.equal(2n, "a message")
```

then `"a message"` would be ignored.


packages/hardhat-chai-matchers/src/bigNumber.ts
```diff
@@ -134,6 +136,33 @@ function overwriteBigNumberFunction(
...
+          const normalizedB = normalizeToBigInt(b);
+          return normalizedA === normalizedB;
+        } catch (e) {
+          // use default comparator
+          return null;
+        }
+      };
+
+      const prevLockSsfi = chaiUtils.flag(this, "lockSsfi");
+      chaiUtils.flag(this, "lockSsfi", true);
```
feuGeneA, line 159: wtf is `lockSsfi`?

fvictorio: This is part of the stuff I just copied from chai. From what I understand, they do this to have better stack traces:

> ssfi: start stack function - used to prevent callback stacks from being shown in errors.

I will add a comment clarifying this.

fvictorio: Done. Let me know if the comment makes sense.

feuGeneA (commented):
FYI, I checked out `hardhat-chai-matchers` and then merged this PR's `chai-bn-deep-equal` into it, and then ran `yarn install`, and was left with the following changes to `yarn.lock`:

```diff
diff --git a/yarn.lock b/yarn.lock
index d5a50d37..21557800 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -3475,16 +3475,11 @@ ansi-styles@~1.0.0:
   resolved "https://registry.yarnpkg.com/ansi-styles/-/ansi-styles-1.0.0.tgz#cb102df1c56f5123eab8b67cd7b98027a0279178"
   integrity sha1-yxAt8cVvUSPquLZ817mAJ6AnkXg=
 
-antlr4@4.7.1:
+antlr4@4.7.1, antlr4@~4.8.0:
   version "4.7.1"
   resolved "https://registry.yarnpkg.com/antlr4/-/antlr4-4.7.1.tgz#69984014f096e9e775f53dd9744bf994d8959773"
   integrity sha512-haHyTW7Y9joE5MVs37P2lNYfU2RWBLfcRDD8OWldcdZm5TiCE91B5Xl1oWSwiDUSd4rlExpt2pu1fksYQjRBYQ==
 
-antlr4@~4.8.0:
-  version "4.8.0"
-  resolved "https://registry.yarnpkg.com/antlr4/-/antlr4-4.8.0.tgz#f938ec171be7fc2855cd3a533e87647185b32b6a"
-  integrity sha512-en/MxQ4OkPgGJQ3wD/muzj1uDnFSzdFIhc2+c6bHZokWkuBb6RRvFjpWhPxWLbgQvaEzldJZ0GSQpfSAaE3hqg==
-
 antlr4ts@^0.5.0-alpha.4:
   version "0.5.0-alpha.4"
   resolved "https://registry.yarnpkg.com/antlr4ts/-/antlr4ts-0.5.0-alpha.4.tgz#71702865a87478ed0b40c0709f422cf14d51652a"
```

packages/hardhat-chai-matchers/src/bigNumber.ts
```diff
@@ -134,6 +136,33 @@ function overwriteBigNumberFunction(
         expected,
         actual
       );
+    } else if (functionName === "eq" && chaiUtils.flag(this, "deep")) {
```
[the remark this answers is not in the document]
fvictorio: The problem is that the inputs to `.to.deep.equal` will be arrays/objects, so they will never be big numbers. I guess we could traverse all objects and check if any member is a big number, but that doesn't seem like a good idea to me.

So yes, we are changing the behavior of _all_ `.to.deep.equal` calls. It would be nice if deep equality would just work when you extend the basic `.equal` matchers, but sadly that's not how chai works. That's why I re-implemented the deep equality comparison, using the same package chai uses, but using a custom comparator function. I'm not super happy with this solution, but I can't think of a better approach.

packages/hardhat-chai-matchers/test/bigNumber.ts
```diff
@@ -864,6 +864,64 @@ describe("BigNumber matchers", function () {
...
+          // successful assertions
+          expect([a]).to.deep.equal([b]);
+          expect([[a], [a]]).to.deep.equal([[b], [b]]);
+          expect({ x: a }).to.deep.equal({ x: b });
+          expect({ x: { y: a } }).to.deep.equal({ x: { y: b } });
+          expect({ x: [a] }).to.deep.equal({ x: [b] });
+
+          // failed assertions
+          expect(() => expect([a]).to.not.deep.equal([b])).to.throw(
+            AssertionError
```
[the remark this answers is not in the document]
fvictorio: Agree that we should have some error message checking here. I'm not sure what's the best way to do it because the way the different numbers are serialized might differ, but I'll give it a try.

packages/hardhat-chai-matchers/src/bigNumber.ts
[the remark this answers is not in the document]
feuGeneA: I see, that makes sense to me, thank you for the explanation.

packages/hardhat-chai-matchers/test/bigNumber.ts
```diff
@@ -877,20 +877,32 @@ describe("BigNumber matchers", function () {
 
           // failed assertions
           expect(() => expect([a]).to.not.deep.equal([b])).to.throw(
-            AssertionError
+            AssertionError,
+            // [\s\S]* is used here instead of .* so that newlines are included
```
feuGeneA, line 881: Is there a reason not to use [the `s` flag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#advanced_searching_with_flags)?

packages/hardhat-chai-matchers/test/bigNumber.ts
```diff
@@ -877,20 +877,32 @@ describe("BigNumber matchers", function () {
...
+            /expected \[[\s\S]*\] to not deeply equal \[[\s\S]*\]/
```
feuGeneA, line 882: Is it not possible (or practical) to actually put the value in here? eg `${a}`? `/expected \[.*${b}.*\] to not deeply equal \[.*${a}.*\]/s`?

feuGeneA (approved):

packages/hardhat-chai-matchers/test/bigNumber.ts
fvictorio: Oh, great catch, thank you. Fixed in 1e5bc1c53a09ceb16171cee195c8ea41ae6bb062.

packages/hardhat-chai-matchers/test/bigNumber.ts
fvictorio: This is a bit hard, because the value in the error message is sometimes `1`, sometimes `1n`, and sometimes `BigNumber { value: "1" }` (or something like that). So I opted for the lazy approach of checking that the error looks _somewhat_ correct.

PR #2661, opened 2022-05-03
Add support for verifying on custom explorers

fvictorio:
Closes HH-591.


no2chem:
@fvictorio Thanks for the credit!

I don't think this works, however - It seems that the type system is overconstrained.

consider the snippet you posted: 
```

etherscan: {
  apiKey: {
    mychain: "your-api-key"
  },
  customChains: [{
    network: "mychain",
    chainId: 123456,
    urls: {
        apiURL: "https://mychainscan.com/api",
        browserURL: "https://mychainscan.com",
    },
  }]
}

```

This fails with

```
hardhat.config.ts:89:3 - error TS2322: Type '{ apiKey: { "mychain": string; }; customChains: { network: string; chainId: number; urls: { apiUrl: string; browserUrl: string; }; }[]; }' is not assignable to type 'EtherscanUserConfig'.
  Types of property 'apiKey' are incompatible.
    Type '{ "mychain": string; }' is not assignable to type 'string | EtherscanApiKeys | undefined'.
```

The issue appears to be in `types.ts`, that type `EtherscanApiKeys` requires `Network in Chain` as an index.

```
type EtherscanApiKeys = {
  [Network in Chain]?: string;
};
```

If you replace it with
```
type EtherscanApiKeys = {
  [network : `${string}`]: string;
};
```

It wil succeed. 

Sure, one could use a `string` for the apiKey, but you might actually want different api keys for different networks...

no2chem:
It seems that it also doesn't work when there is more than one chain defined:

for instance, something like:

```
{
  "apiKey": "no-key",
  "customChains": [
    {
      "network": "chain_31337",
      "chainId": 31337,
      "urls": {
        "apiURL": "http://localhost:6100/api",
        "browserURL": "http://localhost:6100/"
      }
    },
    {
      "network": "chain_31338",
      "chainId": 31338,
      "urls": {
        "apiURL": "http://localhost:6101/api",
        "browserURL": "http://localhost:6101/"
      }
    }
  ]
}
```

results in 

```
HardhatError: HH10: Error while loading Hardhat's configuration.
You or one of your plugins is trying to modify the userConfig.etherscan.customChains.0 value from a config extender
```

fvictorio:
Oh, thanks a lot for the feedback @no2chem, I'll look into this.

ihorbond:
good call @no2chem ! I'm still struggling with getting the contract to verify for Cronos though even after following all of the above I get the generic message:
Error in plugin @nomiclabs/hardhat-etherscan: The contract verification failed.
Reason: Fail - Unable to verify

calvinaco:
@fvictorio I suspect the issue reported above is due to the order of execution of `etherscanConfigExtender` and the  `extendConfig` defined in custom plugin. With inline Ethermint config it is working, but with different combination `hardhat.config.ts` there are some collision of config and report different errors.

Providing a brief summary of test cases first. I can provide a more detail summary later if needed. I didn't have time to write it in details now. 

Note: inline means the config is provided in `hardhat.config.ts`

1. inline Ethermint plugin config with custom chain and their API keys: Working
2. Using `extendConfig` to define custom network + inline API key of custom chain
```
Error in plugin @nomiclabs/hardhat-etherscan: Please provide an Etherscan API token via hardhat config. For example:

{
  ...
  etherscan: {
    apiKey: {
      cronos: 'your API key'
    }
  }
}

See https://etherscan.io/apis
```
Reason: The `etherscanConfigExtender` is executed before the custom `extendConfig`. So the provided keys fail the `etherscanConfigExtender` API key validness check.

3. inline custom network B,C + Using `extendConfig` to define custom network A
```
Error HH10: Error while loading Hardhat's configuration.
You or one of your plugins is trying to modify the userConfig.etherscan.customChains.2 value from a config extender
```
Reasons: the `customChains` is frozen when resolved in `etherscanConfigExtender`, the cloning in that function seems did not loosen the fronzen restriction. on the array. Is this an intended behaviour?

---

I can spend some time to try to resolve them in a fork after learning the expected behaviour. @fvictorio feel free to let me know if there's anythings I can offer help.

fvictorio:
@calvinaco which issue are you referring to? The one reported by @ihorbond? Because 99% of the time that is either the etherscan servers having a problem (this tends to happen when they add a new explorer for some chain, understandably) or some problem like using the wrong constructor arguments.

I'm going to try to get a new rc for this today, and I'll double-check that it indeed works for custom chains, but the "Unable to verify" is unlikely to be caused by the custom chains feature itself (although still possible).

fvictorio:
@no2chem thanks again for the great feedback. I published a new RC version (`3.1.0-rc.1`) under the `rc` tag. Running `npm i @nomiclabs/hardhat-etherscan@rc` should upgrade it.

calvinaco:
@fvictorio sorry I mixed up some of the comments I read across different thread. Let me explain the issues I encounter in more details:

I couldn't reproduce the issue reported by @ihorbond.yet, But I am able to make custom explorer verification work by defining the configs inside `hardhat.config.ts` so that should mean this PR is working in general.

Perhaps @ihorbond could share more on your setup.

---

The next thing I try is to use the [plugin approach](https://github.com/fvictorio/hardhat-custom-chain-plugin-example/blob/master/src/index.ts) you shared on another thread against this branch and found some issues (some are reported and some maybe new)

Issue 1: `etherscan.apiKey` type definitions issue

```
hardhat.config.ts:89:3 - error TS2322: Type '{ apiKey: { "mychain": string; }; customChains: { network: string; chainId: number; urls: { apiUrl: string; browserUrl: string; }; }[]; }' is not assignable to type 'EtherscanUserConfig'.
  Types of property 'apiKey' are incompatible.
    Type '{ "mychain": string; }' is not assignable to type 'string | EtherscanApiKeys | undefined'.
```
This issue is reported earlier in this thread and I just apply a trivial fix first. I think we can define some better types
```
type EtherscanApiKeys = {
  [network: string]: string;
};
```

---

Issue 2: Cannot push new custom chain in `extendConfig` when there are already custom chains defined in `hardhat.config.ts`

How to reproduce:

hardhat.config.ts
```
import "plugin.ts"
...
  etherscan: {
    apiKey: {
      cronos_testnet: <string>process.env["CRONOSCAN_API"],
    },
    customChains: [
      {
        network: "cronos_testnet",
        chainId: 25,
        urls: {
          apiURL: "https://testnet.cronoscan.com/api",
          browserURL: "https://testnet.cronoscan.com/",
        }
      }
    ]
  },
...
```

plugin.ts
```
...
extendConfig(
  (config: HardhatConfig, userConfig: Readonly<HardhatUserConfig>) => {
    config.etherscan.customChains.push({
        network: "cronos",
        chainId: 25,
        urls: {
          apiURL: "https://api.cronoscan.com/api",
          browserURL: "https://cronoscan.com/",
        }
      });
  }
);
...
```

This will give error

```
Error HH10: Error while loading Hardhat's configuration.
You or one of your plugins is trying to modify the userConfig.etherscan.customChains.1 value from a config extender
```

After some debug trace, I saw it's because of this line
https://github.com/NomicFoundation/hardhat/blob/62ceb3d40298e433e458f0c0624a1e068a64ab93/packages/hardhat-etherscan/src/config.ts#L47

When transpiled to JS, it is converted to
```
resolvedConfig.etherscan = Object.assign(Object.assign({}, defaultConfig), customConfig);
```
and `etherscan.customChains` array is still frozen as this is not a deep copy, hence the error. Not sure if this is intended or not.

---

Issue 3: Cannot specify `apiKey` in hardhat config when custom chain is defined in `extendConfig`

plugin.ts
```
import "@nomiclabs/hardhat-etherscan";
...
extendConfig(
  (config: HardhatConfig, userConfig: Readonly<HardhatUserConfig>) => {
    config.etherscan.customChains.push({
        network: "cronos",
        chainId: 25,
        urls: {
          apiURL: "https://api.cronoscan.com/api",
          browserURL: "https://cronoscan.com/",
        }
      });
  }
);
```

hardhat.config.ts
```
import "plugin.ts"
...
etherscan: {
    apiKey: {
      cronos: <string>process.env["CRONOSCAN_API"],
    }
  },
...
```

will result in
```
Error in plugin @nomiclabs/hardhat-etherscan: Etherscan API token "cronos" is for an unsupported network
```

The cause I found is because the line `extendConfig(etherscanConfigExtender);` is run when `hardhat-etherscan` being imported (defined in  https://github.com/NomicFoundation/hardhat/blob/62ceb3d40298e433e458f0c0624a1e068a64ab93/packages/hardhat-etherscan/src/index.ts#L124). This line is always before the custom chain plugin `extendConfig` because `hardhat-etherscan` are always imported first. So `verifyAllowedChains` inside  `etherscanConfigExtender` kicks in too early and reject the API key for custom network that is to be denied later.

fvictorio:
@calvinaco this is extremely useful, thanks a lot. I'll fix those problems as soon as I can.

fvictorio:
@calvinaco I published a new `@rc` version that I think fixed the issues you mentioned. Sorry for so many inconveniences. Can you try again and let me know if it works? Thanks a lot.

calvinaco:
> @calvinaco I published a new `@rc` version that I think fixed the issues you mentioned. Sorry for so many inconveniences. Can you try again and let me know if it works? Thanks a lot.

No worries! I just test with the latest commit and the issues I reported are resolved! Thanks for the quick updates.

morgansliman (approved):
LGTM

PR #2662, opened 2022-05-03
Ether balance matchers

morgansliman:
fixes HH-497


packages/hardhat-chai-matchers/src/changeEtherBalance.ts
```diff
@@ -0,0 +1,93 @@
...
+    `0x${txBlockNumber.toString(16)}`,
+  ]);
+
+  const balanceBefore = await provider.send("eth_getBalance", [
+    address,
+    `0x${(txBlockNumber - 1).toString(16)}`,
+  ]);
+
+  if (options?.includeFee !== true && address === txResponse.from) {
+    const gasPrice = txResponse.gasPrice ?? txReceipt.effectiveGasPrice;
```
fvictorio, line 85: This is technically a bug, but writing a test for it is quite hard, so I'm fine with just fixing it.

The reason why it's a bug is that, when automining is disabled, `txResponse.gasPrice` will be equal to `maxFeePerGas`, which can be much higher than the effective gas price. But writing a test for that is hard because, even when automining is disabled, if the block is mined "too soon" then `txResponse.gasPrice` will have the correct value (because of a combination of how ethers does things, and event loop shenanigans).

```suggestion
    const gasPrice = txReceipt.effectiveGasPrice ?? txResponse.gasPrice;
```

packages/hardhat-chai-matchers/src/changeEtherBalances.ts
```diff
@@ -0,0 +1,125 @@
...
+  options: BalanceChangeOptions
+) {
+  return Promise.all(
+    accounts.map(async (account) => {
+      if (
+        options?.includeFee !== true &&
+        (await getAddressOf(account)) === txResponse.from
+      ) {
+        const txReceipt = await txResponse.wait();
+        const gasPrice = txResponse.gasPrice ?? txReceipt.effectiveGasPrice;
```
fvictorio, line 115: See the previous comment for an explanation.

```suggestion
        const gasPrice = txReceipt.effectiveGasPrice ?? txResponse.gasPrice;
```

packages/hardhat-chai-matchers/src/changeEtherBalance.ts
```diff
@@ -0,0 +1,93 @@
...
+import { BalanceChangeOptions } from "./misc/balance";
+
+export function supportChangeEtherBalance(Assertion: Chai.AssertionStatic) {
+  Assertion.addMethod(
+    "changeEtherBalance",
+    function (
+      this: any,
+      account: Account | string,
+      balanceChange: BigNumberish,
+      options: BalanceChangeOptions
```
fvictorio, line 13: Shouldn't this be

```suggestion
      options?: BalanceChangeOptions
```

?

packages/hardhat-chai-matchers/src/changeEtherBalance.ts
```diff
@@ -0,0 +1,93 @@
...
+      const subject = this._obj;
+      const derivedPromise = Promise.all([
+        getBalanceChange(subject, account, options),
+        getAddressOf(account),
+      ]).then(([actualChange, address]) => {
+        this.assert(
+          actualChange.eq(BigNumber.from(balanceChange)),
+          `Expected "${address}" to change balance by ${balanceChange} wei, ` +
+            `but it has changed by ${actualChange} wei`,
+          `Expected "${address}" to not change balance by ${balanceChange} wei,`,
```
fvictorio, line 24:
```suggestion
          `Expected "${address}" to not change balance by ${balanceChange} wei, but it did`,
```

packages/hardhat-chai-matchers/src/changeEtherBalances.ts
```diff
@@ -0,0 +1,125 @@
...
+} from "./misc/balance";
+
+export function supportChangeEtherBalances(Assertion: Chai.AssertionStatic) {
+  Assertion.addMethod(
+    "changeEtherBalances",
+    function (
+      this: any,
+      accounts: Array<Account | string>,
+      balanceChanges: BigNumberish[],
+      options: BalanceChangeOptions
```
fvictorio, line 16: Same comment as before here, shouldn't this be optional?

packages/hardhat-chai-matchers/test/changeEtherBalance.ts
```diff
@@ -0,0 +1,550 @@
...
+      });
+    });
+
+    describe("Transaction Callback (1559 tx)", () => {
+      describe("Change balance, one account", () => {
+        it("Should pass when expected balance change is passed as string and is equal to an actual", async () => {
+          await expect(() =>
+            sender.sendTransaction({
+              to: receiver.address,
+              maxFeePerGas: 1,
```
fvictorio, line 232: Using the same `maxFeePerGas` and `maxPriorityFeePerGas` is conceptually equivalent to using `gasPrice`. I would update all the `maxFeePerGas` to another value so that these tests are a bit more robust.

Issue #2170, opened 2021-12-17
`ethers.getContractAt` should validate that the address is actually a contract


alcuadrado:
`const myContract = await ethers.getContractAt("MyContract", addr);` should fail if `addr` does not belong to a contract.

kanej:
@alcuadrado is the idea here to use something like ethers [getCode](https://docs.ethers.io/v5/api/providers/provider/#Provider-getCode) to verify there is contract code and warn otherwise?

webdevabhijit:
@alcuadrado can I work in this issue if it is not assigned yet.

alcuadrado:
> @alcuadrado is the idea here to use something like ethers [getCode](https://docs.ethers.io/v5/api/providers/provider/#Provider-getCode) to verify there is contract code and warn otherwise?

The former.

> @alcuadrado can I work in this issue if it is not assigned yet.

Sure! Thanks

SfyMantissa:
@alcuadrado Hello!
Can I also join the current assignees?
I would really like to contribute. I am sure I can do this.

kanej:
> @alcuadrado Hello! Can I also join the current assignees? I would really like to contribute. I am sure I can do this.

Unless @ethabhijit wants to shout, I think you can work on it. Thanks.

feuGeneA:
Fixed via #2916

fvictorio:
Fixed in [hardhat-ethers v2.1.1](https://github.com/NomicFoundation/hardhat/releases/tag/%40nomiclabs%2Fhardhat-ethers%402.1.1)

PR #2665, opened 2022-05-04
Add ethers.getContractAt validation that the given address is actually a contract 

derawi:
Working on issue: https://github.com/NomicFoundation/hardhat/issues/2170

Included a test


alcuadrado:
This bot is way too agressive!

fvictorio:
Hey @derawi, sorry for not responding before.

We had this validation at some point and then we [removed it](https://github.com/NomicFoundation/hardhat/releases/tag/%40nomiclabs%2Fhardhat-ethers%402.2.1) because it broke smock, so I don't think we'll include it again for the time being.

Thanks and sorry again for not responding sooner!

PR #2667, opened 2022-05-04
Upgrade ethereumjs dependencies and remove unused code

LogvinovLeon:

---


alcuadrado:
Nice! Thanks 😍

Can you give me permission to push to that branch or add a changeset like this?

```md
---
"hardhat": patch
"@nomiclabs/hardhat-truffle4": patch
"@nomiclabs/hardhat-truffle5": patch
---

Upgrade ethereumjs dependencies and use their new functionality.
```

LogvinovLeon:
Added the CHANGESET. Also gave permissions if any future changes would be necessary.

LogvinovLeon:
There is one more thing that should be noticed around this PR. I've removed this code written by you:
```
        // sometimes the memSize has the correct value
        // for the memory length, in those cases we increase
        // the memory to reflect this
        if (structLog.memSize > structLog.memory.length) {
          const wordsToAdd = structLog.memSize - structLog.memory.length;
          for (let k = 0; k < wordsToAdd; k++) {
            structLog.memory.push(EMPTY_MEMORY_WORD);
          }
        }
```

Did it because after ethereumjs-vm upgrade it was returning incorrect `memSize` for some of my tests. Expanding memSize before the opcode while correctly expanding memory only after. And this code caused tests to fail as they didn't match the snapshot with this new zero memory. Ideally I would go down `@ethereumjs/vm` and investigate and I've tried, but didn't find anything obvious, so decided that if we have two sources of truth and only one is used and this one is true, then I can remove the part that tries to use false source of truth to break the true one.

But this might be a bad idea and you probably had a reason to have it there, so just pointing it out instead of keeping it under the rug ;)

LogvinovLeon:
The test failure seems to be unrelated to my changes as they were passing before and I can't restart the run. So I've just restarted it with the typo fix commit

alcuadrado:
I noticed that part of the code, which was there because ethereumjs used to behave differently. This PR looks good to me, but I'd like to have @fvictorio's opinion before merging, as he's much more familiar with this area of the codebase than me.

fvictorio:
@LogvinovLeon it seems like I can't push to this branch, is the "Allow edits by maintainers" thing checked?

fvictorio:
I tested this version of `debug_traceTransaction` with an actual transaction (tx `0xbf8863340a7f6615fe1d69eafb805baa81807f8fe69367dbd7ce56ec6eca4d1f` in block 14713415) and I get different results for a `CALL` opcode:

-  The `gasCost` returned from erigon is 136125
- The `gasCost` returned from hardhat is 100

So my guess is that the `dynamicCost` returned from ethereumjs for `CALL` opcodes is returning something different than what we expect (I don't want to say "wrong" because this is a bit subjective).

Checking the same transaction with the code we have in `master` returns the expected values.

Maybe the `dynamicCost` does work as we expect for other opcodes, but checking that would require more testing. So, two options:

- We revert the changes in the debug tracer, and merge the PR
- We check which parts we can replace with the `dynamicCost` returned by ethereumjs (this will take me some time).

@alcuadrado what do you prefer?

LogvinovLeon:
@fvictorio  Ah, sorry. Initially added gave access just to Patricio. You are now also added.
As for the incorrect results - let me add the above mentioned transaction as a test fixture and see for myself.

I've read in the CHANGELOG that all of those (`now taking things like memory usage, address access or storage changes`) are taken into account, so just assumed that tests cover them. If they are not - then it seems like it should be fixed within ethereumjs-vm first. Then I'll just keep the version bump.

LogvinovLeon:
Also - could you please explain your comment about being compatible with Geth more then tests. What if Geth returns something that is obvsly wrong. Should we try to mimic that? Also - by Geth you mean Erigon as in the comment above or Geth?

jochem-brouwer:
Hi guys, I see that there is an issue with memory size, I think this issue https://github.com/ethereum/go-ethereum/issues/24109 is related. Geth's `debug_traceTransaction` usually reports the state *before* executing the opcode, however, with memory, it expands it before actually executing the opcode. In my opinion this is inconsistent and it should be changed. We *cannot* really change this in ethereumjs-vm, since this is a DoS vector: if you would execute a contract where you would expand a gigantic amount of memory (and would thus run out of gas) then the reported `step` event would try to allocate a gigantic amount of memory and thus run out of memory. (This is also an issue in Geth I think)

LogvinovLeon:
After some further debugging - it seems like Geth/Erigon/Hardhat add the gas that is passed down in a call to a `gasCost`.
I would say this is wrong on two notions. 1. You kinda expect that summing all gas costs will give you the total gas cost except for base cost and call data. 2. Even if the call doesn't consume all the gas and some of it is returned - we already recorded it as spent. At the same time - I understand that it's better to be compatible with other clients and it's a bigger discussion. For now I just reverted all the changes except the version bumps and memory expansion.

@jochem-brouwer  I'm nut sure I get your point in full. `We cannot really change this in ethereumjs-vm, since this is a DoS vector`. For me this is exactly the reason why it `should` be changed in both geth and ethereumjs-vm.

jochem-brouwer:
@LogvinovLeon yes, sorry, this is also my point: Geth should change this as well. So, memSize is whatever the memory size is before executing the opcode. In EthereumJS-VM we currently report the state of all variables *before* executing the opcode.

alcuadrado:
Pinging @haltman-at @gnidan @cgewecke who may find this discussion relevant.

alcuadrado:
Also, @fvictorio should we add the transaction you mentioned as a test fixture?

alcuadrado:
> could you please explain your comment about being compatible with Geth more then tests

What I meant is that fixing the values [here](https://github.com/ethereum/tests/tree/develop/GeneralStateTests) is easier than fixing the behavior of `debug_traceTransaction`, as the latter is more widely consumed in the ecosystem.

fvictorio:
To sum up: we should merge this PR as-is, and we should move the whole `debug_traceTransaction` discussion elsewhere. If no one is against that, I will merge the PR.

alcuadrado:
> I think that won't be necessary. 

Awesome! Let's merge it then! 

Thanks @LogvinovLeon! And thanks @jochem-brouwer for chiming in.

PR #2695, opened 2022-05-11
Support predicate in .withArgs()

feuGeneA:
- [x] Support predicate arguments
- [x] Support at least one built-in predicate matcher, like anyArg(), anyValue(), etc. (source: https://linear.app/nomic-foundation/issue/HH-542#comment-de529709)

### DIFF
--- a/packages/hardhat-chai-matchers/src/reverted/revertedWithCustomError.ts
+++ b/packages/hardhat-chai-matchers/src/reverted/revertedWithCustomError.ts
@@ -1,3 +1,5 @@
+import { AssertionError } from "chai";
+
 import { decodeReturnData, getReturnDataFromError } from "./utils";
 
 export const REVERTED_WITH_CUSTOM_ERROR_CALLED = "customErrorAssertionCalled";
@@ -160,7 +162,29 @@ export async function revertedWithCustomErrorWithArgs(
 
   for (const [i, actualArg] of Object.entries(actualArgs) as any) {
     const expectedArg = expectedArgs[i];
-    if (Array.isArray(expectedArg)) {
+    if (typeof expectedArg === "function") {
+      const errorPrefix = `The user-defined predicate for custom error argument #${
+        parseInt(i, 10) + 1
+      }`;
+      try {
+        context.assert(
+          expectedArg(actualArg),
+          `${errorPrefix} returned false`
+          // no need for a negated message, since we disallow mixing .not. with
+          // .withArgs
+        );
+      } catch (e) {
+        if (e instanceof AssertionError) {
+          context.assert(
+            false,
+            `${errorPrefix} threw an AssertionError: ${e.message}`
+            // no need for a negated message, since we disallow mixing .not. with
+            // .withArgs
+          );
+        }
+        throw e;
+      }
+    } else if (Array.isArray(expectedArg)) {
       // we use [...x] here to convert the array-like values used by ethers to
       // represent structs into proper arrays
       new Assertion([...actualArg]).to.deep.equal([...expectedArg]);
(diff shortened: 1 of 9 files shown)


packages/hardhat-chai-matchers/src/reverted/revertedWithCustomError.ts
```diff
@@ -160,7 +162,29 @@ export async function revertedWithCustomErrorWithArgs(
 
   for (const [i, actualArg] of Object.entries(actualArgs) as any) {
     const expectedArg = expectedArgs[i];
-    if (Array.isArray(expectedArg)) {
+    if (typeof expectedArg === "function") {
+      const errorPrefix = `The user-defined predicate for custom error argument #${
```
fvictorio, line 166: The error prefix in `emit` is a bit different (it does not say "user-defined")

packages/hardhat-chai-matchers/src/withArgs.ts
```diff
@@ -1,9 +1,39 @@
...
+
+export function anyUint(i: any): boolean {
+  if (typeof i === "number") {
+    if (i < 0) {
+      throw new AssertionError(
+        `anyUint expected its argument to be an unsigned integer, but it was negative, with value ${i}`
+      );
+    }
+    return true;
+  } else if (isBigNumber(i)) {
```
fvictorio, line 23: This code would be simpler if `isBigNumber` returned true for numbers (although the name wouldn't make sense), because `normalizeToBigInt` does accept numbers.

Nothing actionable here, just something to keep in mind, especially if we make further modifications to our number helpers.

packages/hardhat-chai-matchers/test/fixture-projects/hardhat-project/contracts/Matchers.sol
```diff
@@ -92,6 +93,15 @@ contract Matchers {
     revert CustomErrorWithUint(n);
   }
 
+  function revertWithCustomErrorWithInt(int i) public {
+    x++;
+    revert CustomErrorWithInt(i);
+  }
+
+  function revertWithCustomErrorWithIntView(int i) public {
```
fvictorio, line 101:
```suggestion
  function revertWithCustomErrorWithIntView(int i) public pure {
```

packages/hardhat-chai-matchers/src/reverted/revertedWithCustomError.ts
feuGeneA: Thanks. My first draft had "user-defined", but later I scrapped that, but I neglected to follow up in the reverts. Good catch. Addressed in https://github.com/NomicFoundation/hardhat/pull/2695/commits/c46da127ee74e3931bb5ae84c44c62f01e0aba72

packages/hardhat-chai-matchers/src/withArgs.ts
```diff
@@ -1,9 +1,39 @@
...
+
+import { isBigNumber, normalizeToBigInt } from "hardhat/common";
+
 import { emitWithArgs, EMIT_CALLED } from "./emit";
 import {
   revertedWithCustomErrorWithArgs,
   REVERTED_WITH_CUSTOM_ERROR_CALLED,
 } from "./reverted/revertedWithCustomError";
 
+export function anyValue(): boolean {
```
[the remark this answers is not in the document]
feuGeneA: Never done a jsdoc, and not sure if/how I can validate it (is there a linter for it? or a render job or something?) but addressed in https://github.com/NomicFoundation/hardhat/pull/2695/commits/a075dff56758057f30754b618e08d591de325992

fvictorio (approved):

PR #2705, opened 2022-05-13
Support for Node 18

feuGeneA:


.github/workflows/e2e-ci.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [12, 14, 16]
+        node: [14, 16, 18]
```
alcuadrado, line 34: At first, I was "wait, we still support node 12!", but we are dropping its support in 13 days, so yeah, let's do this 🥳

docs/src/content/config/index.md
```diff
@@ -119,7 +119,7 @@ module.exports = {
 
 ```js
 {
-    localhost: {
+    127.0.0.1: {
```
alcuadrado, line 122:
```suggestion
    localhost: {
```

This is the name of the network.

packages/hardhat-core/CHANGELOG.md
```diff
@@ -146,7 +146,7 @@
 - 8076c43b: Fixed how the cummulative gas is computed for receipts and added a missing field (Thanks @ngotchac!)
 - e6362902: Display similar artifact names in error output if given name is not found (#201)
 - e087bd0b: Improve validation of private keys in the Hardhat config
-- aa1a0080: Fix an issue with new versions of Node.js that prevented clients from connecting to Hardhat's node using 127.0.0.1
+- aa1a0080: Fix an issue with new versions of Node.js that prevented clients from connecting to Hardhat's node using localhost
```
alcuadrado, line 149:
```suggestion
- aa1a0080: Fix an issue with new versions of Node.js that prevented clients from connecting to Hardhat's node using 127.0.0.1
```

Let's not change the changelogs, as they are useful as a record of what happened.

packages/hardhat-core/package.json
```diff
@@ -23,7 +23,7 @@
     "hardhat": "internal/cli/cli.js"
   },
   "engines": {
-    "node": "^12.0.0 || ^14.0.0 || ^16.0.0"
+    "node": "^14.0.0 || ^16.0.0 || ^18.0.0"
```
alcuadrado, line 26:
```suggestion
    "node": "^12.0.0 || ^14.0.0 || ^16.0.0 || ^18.0.0"
```

This is unfortunate, but let's play nicely here.

packages/e2e/package.json
```diff
@@ -32,7 +32,7 @@
     "eslint-plugin-import": "2.24.1",
     "eslint-plugin-prettier": "3.4.0",
     "fs-extra": "^7.0.1",
-    "mocha": "^9.2.0",
+    "mocha": "^10.0.0",
```
alcuadrado, line 35: Was this change required? Why?

packages/hardhat-ethers/test/ethers-provider-wrapper.ts
```diff
@@ -12,7 +12,9 @@ describe("Ethers provider wrapper", function () {
   useEnvironment("hardhat-project");
 
   beforeEach(function () {
-    realProvider = new ethers.providers.JsonRpcProvider();
+    realProvider = new ethers.providers.JsonRpcProvider(
+      "http://127.0.0.1:8545"
```
alcuadrado, line 16: Good catch! 🙌

packages/hardhat-ganache/test/index.ts
```diff
@@ -4,7 +4,7 @@ import { GanacheService } from "../src/ganache-service";
 
 import { useEnvironment } from "./helpers";
 
-describe("Ganache plugin with empty configs", function () {
+describe.skip("Ganache plugin with empty configs", function () {
```
alcuadrado, line 7:
```suggestion
// Note: Skipping as this plugin hasn't been updated to the latest Ganache version,
//   so it doesn't work well with Node 18
describe.skip("Ganache plugin with empty configs", function () {
```

packages/hardhat-ganache/test/index.ts
```diff
@@ -49,7 +49,7 @@ describe("Ganache plugin with empty configs", function () {
   });
 });
 
-describe("Ganache plugin with custom configs", function () {
+describe.skip("Ganache plugin with custom configs", function () {
```
alcuadrado, line 52:
```suggestion
// Note: Skipping as this plugin hasn't been updated to the latest Ganache version,
//   so it doesn't work well with Node 18
describe.skip("Ganache plugin with custom configs", function () {
```

packages/e2e/package.json
alcuadrado: I'm actually ok with it, but I want to understand how this may affect existing users.

alcuadrado:
Great work!

I left a few minor comments.

We need to add changesets to many packages. I can do it.

packages/hardhat-ganache/src/ganache-service.ts
```diff
@@ -202,7 +202,7 @@ export class GanacheService {
 
     // Validate and parse hostname and port from URL (this validation is priority)
     const url = new URL(options.url);
-    if (url.hostname !== "localhost" && url.hostname !== "127.0.0.1") {
+    if (url.hostname !== "127.0.0.1") {
```
alcuadrado, line 205:
```suggestion
    if (url.hostname !== "localhost" && url.hostname !== "127.0.0.1") {
```

I think we can revert this and not publish a new version, as it's not very interesting.

packages/e2e/package.json
morgansliman: I'm not necessarily sure if it was "required" per se, but mocha 10 is basically their version of the same upgrade we're doing here (i.e. deprecating node 12 support and adding node 18 support to their CI), so I felt like it made sense to include it.

alcuadrado: Makes sense

alcuadrado (approved):

PR #2722, opened 2022-05-17
fix console.log() prints undefined to terminal

### LINKED ISSUE #2721: Using console.log() prints undefined to terminal
This issue is regarding `console.sol` found at `packages/hardhat-core/console.sol`

Steps to Reproduce
1. Create or find an existing smart contract which you can run on a Hardhat network.
2. Add a `console.log()` anywhere in the contract.
3. Deploy your contract and call the function in which you added your blank log.

Expected: You see nothing (or a new line)
Actual: You see the word undefined being printed to your screen

Often to keep my JS logs organized, I will put a blank `console.log()` to create a new line in my terminal. However, doing so in Solidity code will result in `undefined` being printed to the screen. Below is a proposed solution that I have tested and found to work. I'll be creating a PR for this as well.

In `packages/hardhat-core/console.sol` on line 17

Replace `_sendLogPayload(abi.encodeWithSignature("log()"));` with `_sendLogPayload(abi.encodeWithSignature("log(string)", ""));`

icepaq:

---

Fixes #2721 2721

### DIFF
--- a/packages/hardhat-core/console.sol
+++ b/packages/hardhat-core/console.sol
@@ -14,7 +14,7 @@ library console {
 	}
 
 	function log() internal view {
-		_sendLogPayload(abi.encodeWithSignature("log()"));
+		_sendLogPayload(abi.encodeWithSignature("log(string)", ""));
 	}
 
 	function logInt(int p0) internal view {


alcuadrado:
Thanks for sending this PR. 

The `console.sol` file is autogenerated though, `packages/hardhat-core/scripts/console-library-generator.js` creates it.

icepaq:
It looks like the generator wasn't used in over 2 years. Over the time, it seems as passing a string into keccak256 became deprecated.

This is why you'll find I added Buffer.from() on a few lines.

fvictorio:
This is already fixed in the latest versions of Hardhat, thanks again @icepaq!

PR #2726, opened 2022-05-17
New structure for docs

fvictorio:


feuGeneA (commented):
Overall I think this is looking great and represents a great step forward!

I like the idea of having the left-hand-pane's "Guides" section roughly correspond to the different sections of the "Quick Start" section. To reinforce that, maybe in the Guides move "Tasks and Scripts" to be just after "Setting up a Project"?

I wonder if, in the left-sidebar the "Hardhat Network Helpers" stuff wouldn't be better placed within the "Hardhat Network" expander.

Also, further regarding the left sidebar, it feels like "first-class products" are what deserves the expander, and Hardhat Network and Hardhat for VSCode both feel like they fit that model, but Hardhat Chai Matchers does _not_ feel like it fits that model, like, the Chai Matchers are not a "first-class product", are they? It kinda feels like the Chai Matchers should be integrated into "testing your contracts," and if it warrants its own page then that's fine and we can link to it from "testing your contracts," but that we shouldn't link to it from the left-hand pane. (Weak opinion...)

I feel like the "Plugins" expander needs as its first sub-item a parallel to "What is it?", perhaps "What are they?" This can have a short blurb about how extensibility is a major design goal of Hardhat, and then about how we have many official plugins and many community plugins as well.

And maybe call them "Hardhat Runner Plugins" ? I'm not as sure about that part haha, but it does feel weird to have the expanders represent "products" but then just have one called "Plugins" without it being clear what the plugins are _for_ (they're not for Network or VSCode, so there's a weird parallelism-not-parallelism going on there).

PR #2728, opened 2022-05-17
fix: close node watcher after task

zzmp:
- [X] Because this PR includes a **bug fix**, relevant tests have been included.

---

Fixes the `node` task by closing the `chokidar` file watcher after the server has closed.

This prevents programmatic usages of the `node` task from hanging. For example, a test (eg `mocha`, `jest`) may run the `node` task for mainnet forking and terminate the server when the tests complete, but the test framework will hang because `chokidar` will not allow the process to quit.

I've included a test to show that the `node` task terminates. The test will still pass without these changes, but (as described) the test framework will hang despite the test passing. I could find no way to make the test fail, as the `chokidar` watcher is a not exposed in the current state of the code.

### DIFF
--- a/.changeset/four-foxes-sleep.md
+++ b/.changeset/four-foxes-sleep.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Updates `node` task to terminate when the server closes.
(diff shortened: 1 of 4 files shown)


alcuadrado:
Thanks! This looks really good

zzmp:
Thanks @alcuadrado. Do you know when https://www.npmjs.com/package/hardhat will be released so that this fix is available?

fvictorio:
Hey @zzmp, you probably already know by now, but [Hardhat v2.9.6](https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.9.6) includes this change :slightly_smiling_face:

PR #2737, opened 2022-05-19
Fix fork cache not working for unknown networks

### LINKED ISSUE #2736: Fork cache not working for Optimism and other networks
## Overview

https://github.com/NomicFoundation/hardhat/blob/e5e4a6e4d79cc79e03be90052059bb321645a542/packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts#L81

When using Hardhat to fork Optimism Mainnet or other networks that do not have a predefined `maxReorg` caching will be disabled. 

Ideally `cacheToDiskEnabled` should use `maxReorg` instead of `actualMaxReorg`. This will enable caching on any chain as long as the block number is more than `FALLBACK_MAX_REORG` behind the latest.

Alternative solution which will require more work: add an option to explicitly set the fork cache settings and if defined always cache.

## Reproduce

```
npx hardhat node --fork 'https://opt-mainnet.g.alchemy.com/v2/KEY' --fork-block-number 8540800
```

bernard-wagner:
Fixes #2736


fvictorio:
Ugh, I fixed this in https://github.com/NomicFoundation/hardhat/pull/3148 and didn't realize that this PR already existed. Sorry about that!

fvictorio:
@gitpoap-bot @bernard-wagner deserves a gitpoap

PR #2741, opened 2022-05-20
Fix problems with chaining `.emit` chai matchers

feuGeneA:


packages/hardhat-chai-matchers/src/emit.ts
```diff
@@ -87,26 +89,35 @@ export function supportEmit(
...
   expectedArgs: any[],
   log: any
 ) {
   const { utils } = require("ethers");
 
-  const actualArgs = (context.contract.interface as Interface).parseLog(
-    log
-  ).args;
+  const actualArgs = (
+    chaiUtils.flag(context, "contract").interface as Interface
```
fvictorio, line 114: I think using the flags inline here (and in some other places in the PR) makes the code a little harder to read, can we extract them to some variables?

fvictorio (approved):
Pre-approving because there's only one really minor comment.

packages/hardhat-chai-matchers/src/emit.ts
feuGeneA: https://github.com/NomicFoundation/hardhat/pull/2741/commits/290ac34ed88bdd1bca38420b82445c5caa2b9c3b

PR #2747, opened 2022-05-24
perf(fs): realpath native support

sambacha:
Subsantiave changes are documented below:

```diff

- import path from "path";
+ import path from 'node:path';



- : files.map((f) => fs.realpathSync(f));
+ : files.map((f) => fs.realpathSync.native(f));


- const configFile = fs.realpathSync(userConfigPath);
+ const configFile = fs.realpathSync.native(userConfigPath);

- fs.realpathSync(resolvedPackageJson) === fs.realpathSync(thisPackageJson)
+ fs.realpathSync.native(resolvedPackageJson) === fs.realpathSync.native(thisPackageJson)



+  // TODO: 
+ fs.mkdtempSync(path.join(os.tmpdir(), `hardhat-tests-${nameHint}`));
- const tmpDir = path.join(tmpDirContainer, `hardhat-tests-${nameHint}`);


+ [path.basename(s)]: { content: fs.realpathSync.native(s, "utf8") },
- [path.basename(s)]: { content: fs.readFileSync(s, "utf8") },


+ projectRoot = fs.realpathSync.native(".");
- projectRoot = fs.realpathSync(".");


+ const projectRoot = fs.realpathSync.native(".");
- const projectRoot = fs.realpathSync(".");


@@ packages/hardhat-core/test/internal/solidity/resolver.ts
+ const realpathSync = fs.realpathSync.native ?? fs.realpathSync;



- fs.realpathSync(path.join("contracts", "fromContracts.sol"))
+ fs.realpathSync.native(path.join("contracts", "fromContracts.sol"))
```


sambacha:
Note this requires NodeJS 14, and typescript 4.6.3

sambacha:
the build was previously failing, fixed that now

sambacha:
### Workflow changes
- Changes made to workflows because this is a breaking change in typescript versioning and besides I do not think yall officially support nodejs 12 anymore.

- by specifying `--prefer-offline`,  yarn will check the cache first, and only try to download the package from the registry if it cannot find it in the cache. 

- updated the nodejs action to v3

### Dependency  changes

- typescript@^4.6.3 (4.7.1 is latest)
- I added turbo by accident , will remove that ofc. 

### Notes

Completely unsolicited ideas, I think some might be ez wins others, well YMMV - would like to know your thoughts ofc!

0. Include certain compiler versions (e.g. 0.6.12) or pre-compiled contracts (OZ, Uni, etc) with hardhat potentially. For reference `solc-typed-ast` compiler does include a cache env for including compilers, [see https://github.com/ConsenSys/solc-typed-ast/blob/master/docker/download.sh](https://github.com/ConsenSys/solc-typed-ast/blob/master/docker/download.sh)

1. [Nx Distributed Task Execution (DTE) Example/Benchmark, https://github.com/vsavkin/interstellar](https://github.com/vsavkin/interstellar)

2. [ tree diffing when restoring files from the cache https://github.com/vsavkin/large-monorepo#why-is-nx-faster-than-turbo](https://github.com/vsavkin/large-monorepo#why-is-nx-faster-than-turbo)

3. [core_d.js launches a server so that your services invoke method is called with the same arguments. Later calls to invoke will be executed on the same instance. ](https://github.com/mantoni/core_d.js#how-does-this-work)
> 3 may be useful for `packages/hardhat-core/src/internal/util/scripts-runner.ts`

4. potentially invoking nodejs directly e.g.

```bash
node --expose-gc node_modules/.bin/jest --coverage --globals "{\"coverage\":true}" src/
```

5. [Utilize a more performant cache system like https://github.com/next-boost/hybrid-disk-cache](https://github.com/next-boost/hybrid-disk-cache) - see [why sqlite3 is faster than the filesystem (below certain file sizes ref: https://www.sqlite.org/fasterthanfs.html](https://www.sqlite.org/fasterthanfs.html)

## Questions 

How to best benchmark this? What sort of test would actually show any statistically significant differences? I would think maybe measuring access time or something as opposed to measuring compilation time.

alcuadrado:
Lots to digest here. Thanks @sambacha! What's the part that requires node 14?

Also, are the `node:` imports faster?

alcuadrado:
This took some time, but we finally got to optimize the compilation process a lot. You can take a look at #3048 to #3053.

Thanks for these ideas <3

alcuadrado:
@gitpoap-bot @sambacha deserves a gitpoap, or a hundred actually

PR #2750, opened 2022-05-24
Add deployContract helper to hardhat-ethers

zemse:

---

Before:

```ts
const YourContract = await ethers.getContractFactory("YourContract");
myContract = await YourContract.deploy();

// or

myContract = await (
  await ethers.getContractFactory("YourContract")
).deploy();
```

After:

```ts
myContract = await ethers.deployContract("YourContract");
```

This is just for improving the convenience a bit.


packages/hardhat-ethers/src/internal/helpers.ts
```diff
@@ -321,6 +321,30 @@ export async function getContractAt(
...
+export async function deployContract(
+  hre: HardhatRuntimeEnvironment,
+  name: string,
+  signerOrOptions?: ethers.Signer | DeployOptions
+): Promise<ethers.Contract> {
+  let libraries: Libraries = {};
+  let signer: ethers.Signer | undefined;
+  let args: any[] = [];
+
+  if (isFactoryOptions(signerOrOptions)) {
```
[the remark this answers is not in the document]
feuGeneA: Aside from the weirdness of reusing `isFactoryOptions` here, I also noticed that much of this logic is duplicated in `getContractFactory()`, and that this function's body can be collapsed to just:
```typescript
{
  const factory = await getContractFactory(hre, name, signerOrOptions);
  const args = isFactoryOptions(signerOrOptions)
    ? signerOrOptions.args ?? []
    : [];
  return factory.deploy(...args);
}
```
While this nicely avoids repetition, it also avoids any handling of the `libraries` property, which in turn makes me feel more comfortable with not having test cases exercising library usage for this new function.

packages/hardhat-ethers/test/index.ts
```diff
@@ -966,15 +942,8 @@ describe("Ethers plugin", function () {
...
-            "setGreeting(string)",
-            "greet()",
-          ]);
-
-          assert.equal(
-            await contract.signer.getAddress(),
-            await signers[0].getAddress()
-          );
+          assert(await contract.greet(), "Hello");
+          await assertContract(contract, signers[0]);
```
feuGeneA, line 946: A nit: probably should `assertContract()` _before_ `contract.greet()`, in order to get a more helpful failure message in the case where `assertContract()` would fail.

packages/hardhat-ethers/src/types/index.ts
```diff
@@ -11,6 +11,9 @@ export interface FactoryOptions {
   signer?: ethers.Signer;
   libraries?: Libraries;
 }
+export interface DeployOptions extends FactoryOptions {
+  args?: any[];
```
feuGeneA, line 15: The team is leaning towards requiring that contract args be typed in order to accept this PR. Perhaps the interface could be changed to support something like `ethers.deployContract("Foo", [arg1, arg2], options)`?

feuGeneA (changes requested):

packages/hardhat-ethers/test/index.ts
zemse: Right makes sense

packages/hardhat-ethers/src/types/index.ts
zemse: Yep, I'm planning to add overloads to typechain as well, in order args to be typed.

So it could generate overloads similar to [this](https://github.com/dethcrypto/TypeChain/blob/master/packages/hardhat/test/fixture-projects/hardhat-project/typechain-types/hardhat.d.ts). Do you think it can work?

```ts
// ArgsType would be generated by Typechain
// following is an example of an overload for a ContractName.
function deployContract(name: 'ContractName', options: DeployOptions<ArgsType>);

interface DeployOptions<T extends Array> extends FactoryOptions {
  args?: T;
}
```

zemse: Or do you prefer to handle this in this way:
```ts
ethers.deployContract("Foo", [arg1, arg2], options) // for constructor args
ethers.deployContract("Foo", options) // for no constructor args
```

I think doing this change can make deployContract one-liner, at least for short `args`. So this can be great, thanks.

zemse: Sorry for the delay. I've made changes to the signature as per your suggestion @feuGeneA. Please let me know or suggest changes. 

Though I didn't understand "requiring that contract args be typed", how do you think args can be typed (better than `any[]`) within the hardhat or what's the typability difference you're seeing in previous approach vs this one? (for me these both approaches seem to be typed).

zemse:
Still interested in this small feature. @feuGeneA Please let me know if any changes are needed.

alcuadrado:
Chiming to confirm that this can be typed by typechain, as checked with Kris.

alcuadrado (approved):

zemse:
Ah. Last night I was just wanting to get a PR up for adding a new plugin. My fork had `.vuepress` thing but list of plugins moved to different place. So I deleted the old fork and created a [PR](https://github.com/zemse/hardhat/pull/1) (and I just realised it was on my fork lol). Moral of the story, no work after hours.

The branch `zemse:deploy-contract-helper` doesn't exist (got deleted when I deleted the fork). From my phone, I thought this PR is gone but I was able to open it, not sure if its mergeable in this condition.

PR #2764, opened 2022-05-26
added hardware wallet example

Chriton:
Added hardware wallet config example


fvictorio:
Hi @Chriton, thanks for sending this. I modified it a bit to make it more concise.

Also, "HD" here means "Hierarchical Deterministic", not "Hardware". Since this is a very common mistake (I know I did it once too), I added a link to an explanation about this.

PR #2766, opened 2022-05-27
Fix `block.chainid` returning 0 in view functions

fvictorio:


feuGeneA (approved):
Just out of curiosity: why use pre-compiled bytecode and ABI instead of compiling just in time?

fvictorio:
> Just out of curiosity: why use pre-compiled bytecode and ABI instead of compiling just in time?

Because I forgot we could do that now :roll_eyes:

Issue #586, opened 2020-05-18
Support for yarn berry (v2)


paulrberg:
[Berry](https://github.com/yarnpkg/berry) is yarn's v2 release and is disruptive in many regards. They replaced "node_modules" with a new [plug-n-play](https://classic.yarnpkg.com/en/docs/pnp/) system that caches dependencies more efficiently.

In the docs for [Typescript support](https://buidler.dev/guides/typescript.html#plugin-type-extensions), you specify that we should add this to our `tsconfig.json` config:

```json
"files": [
  "./buidler.config.ts",
  "./node_modules/@nomiclabs/buidler-ethers/src/type-extensions.d.ts",
  "./node_modules/@nomiclabs/buidler-waffle/src/type-extensions.d.ts"
]
```

But this is sadly not possible in yarn v2. I suppose that this is not a priority, but I wanted to open an issue about it anyways, in case more people stumble upon it and want to signal their support.

PaulRBerg:
It's actually worse than just not being able to use the ethers and waffle plugins. The way Buidler is architected seems to not be compatible \*at all\* with yarn v2 + typescript.

If I run `yarn buidler compile` in a yarn v2 project, I am given the default response:

```sh
? What do you want to do? … 
❯ Create a sample project
  Create an empty buidler.config.js
  Quit
```

But if I copy-and-paste the same source code into a yarn v1, `yarn builder compile` successfully compiles the contract:

```sh
Compiling...
Downloading compiler version 0.6.8
Compiled 1 contract successfully
✨  Done in 4.12s.
```

For the record, this is my `tsconfig.json`:

```json
{
  "compilerOptions": {
    "lib": ["es5"],
    "module": "commonjs",
    "outDir": "dist",
    "sourceMap": true,
    "target": "es5",

    "noImplicitAny": true,
    "strict": true,

    "esModuleInterop": true,
    "moduleResolution": "node",
    "resolveJsonModule": true,

    "forceConsistentCasingInFileNames": true
  },
  "files": [
    "./buidler.config.ts",
    "./node_modules/@nomiclabs/buidler-ethers/src/type-extensions.d.ts",
    "./node_modules/@nomiclabs/buidler-waffle/src/type-extensions.d.ts"
  ],
  "include": ["./scripts", "./test"]
}
```

alcuadrado:
Thanks for opening this issue, @PaulRBerg!

[...]

PR #2775, opened 2022-05-30
Specify ts-node and typescript as optional peerDependencies

kevinji:

I tested the changes locally; not sure how to add this as a test.

---

This fixes a compatibility issue with Yarn strict pnpMode, as hardhat was calling `require.resolve` on both modules.

See also: https://yarnpkg.com/configuration/yarnrc#pnpMode

Improves #586.


alcuadrado:
This is a good point. I just checked and npm, yarn1, and yarn2 support `peerDependenciesMeta`, so this LGTM.

I am running the CI, just in case. It should be ready to be merged after that.

kevinji:
@alcuadrado I force pushed a change to explicit set version bounds, as the lint check from before was failing.

kevinji:
Ah gotcha. I just repushed with the original changes; feel free to fix the script by pushing to the same branch.

PR #2787, opened 2022-06-01
Fix slow tests

fvictorio:
For some reason (which I believe is related to some interaction between sinon an undici), some tests were taking a long time while the server was being closed. This was adding more than 30 seconds to the total time of the test run.

Upgrading undici, plus some other minor changes, fixed it.

Since this is not the first time this happens, I also added a check that fails if the server takes more than one second to close.

### DIFF
--- a/packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts
+++ b/packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts
@@ -871,7 +871,7 @@ describe("Evm module", function () {
             });
 
             it("should allow disabling interval mining", async function () {
-              const interval = 1000;
+              const interval = 100;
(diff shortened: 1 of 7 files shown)


packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts
```diff
@@ -891,7 +891,7 @@ describe("Evm module", function () {
             });
 
             it("should mine block with transaction after the interval", async function () {
-              const interval = 1000;
+              const interval = 100;
```
fvictorio, line 894: This is unrelated, but since I was optimizing the time of this file, I just did it. I think the test is equally valid with a 100ms interval, and this reduces the total test time by another 1800ms.

PR #2789, opened 2022-06-01
Fix incorrect desktop social urls

crgeary:
Fixes incorrect social URLs in the header on desktop due to the `.name` prop being used in the `href={}`.


alcuadrado:
Thanks!

PR #1791, opened 2021-08-12
Add httpHeaders to forking config

TimDaub:
- Fixes #1588
- [x] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.

I haven't tested this except for running `npm run watch` (0 errors). I'm happy to contribute more by adding tests, but I'm helplessly overwhelmed by the code base's size. Pls guide me thx.

PR #2795, opened 2022-06-03
Add httpHeaders to forking config (alternative PR)

### LINKED ISSUE #1238: Hardhat network forking with a proxy
Some users have reported that they can't use mainnet forking if they are behind a proxy. We should add an option for this.

fvictorio:
This is just #1791, with a minor change. I can't push to that PR so I opened this one just to get it merged.

I tested this locally and it worked fine. It would be nice to have tests for this, but I'm not sure how to do it easily. If either Alchemy or Infura accepted the token id as a HTTP header, we could add an extra tests when those URLs are set, extract the token, and pass it via a header instead. But as far as I can tell, they don't support this (I didn't invest too much time looking into this though).

Closes #1238.


TimDaub:
Hey, I just tested this with my full node over at https://rugpullindex.com/docs#AccessToOurEthereumErigonNode and some [hardhat mainnet forking scenario tests](https://github.com/whitecube-online/balot/blob/main/test/testScenarios.ts) and it worked like a charm! Thanks for merging!

fvictorio:
Thanks for testing it! I will close the other PR then.

PR #2800, opened 2022-06-07
Add tabs error handling at build time

RavenInd:
-add tabs error handling at build time


fvictorio:
Hey @RavenInd, this seems to check that the value used in `tab{value=X}` corresponds to one of the options in the tabs group (which is very useful), but we also want to check in build time that the options in the tabs group match what is defined in `global-tabs.tsx`.

Put another way: if you go to `global-tabs.tsx` and replace all "yarn"s with "yrn", the build should fail.

The reason we want this is the scenario where someone adds a new kind of tabs group but they don't know that they also need to update the `global-tabs.tsx` file. If that happens, a build error would let them figure out earlier that they are doing something wrong.

fvictorio:
To give a bit more context: what we want is to have three tabs: "npm 7+", "npm" and "yarn". Having more than 2 tabs seems to be already supported in this PR, so the only thing pending (besides the problem mentioned in [this comment](https://github.com/NomicFoundation/hardhat/pull/2800#issuecomment-1150062536)) would be to support spaces.

PR #2801, opened 2022-06-07
Fr/hotfix/badges same line

RavenInd:
- place badges at the same line


fvictorio:
Hey @RavenInd, thanks a lot for fixing this.

Something else I just realized: the Hardhat badge is not working outside the website. For example, [in npm](https://www.npmjs.com/package/@nomiclabs/hardhat-ethers):

image

And [in GitHub](https://github.com/NomicFoundation/hardhat/tree/master/packages/hardhat-ethers):

image

I think the reason they don't work is that they URL of the image is https://hardhat.org/buidler-plugin-badge.svg?1, but that doesn't exist. I did add it to the repo (see [here](https://github.com/NomicFoundation/hardhat/blob/master/docs/src/assets/buidler-plugin-badge.svg)), but that was probably not the right way to add an asset.

fvictorio:
@RavenInd can you fix the merge conflicts? I would do it but I can't push to your branch for some reason.

PR #2802, opened 2022-06-07
Update tutorial to use Goerli instead of Ropsten

fvictorio:


feuGeneA (approved):
I left a few tweaks that had nothing to do with your changes :sweat_smile: Sorry about that haha.

PR #2809, opened 2022-06-08
hh-ethers: getImpersonatedSigner

feuGeneA:


packages/hardhat-ethers/test/index.ts
```diff
@@ -143,6 +147,16 @@ describe("Ethers plugin", function () {
         });
       });
 
+      describe("getImpersonatedSigner", function () {
+        it("should return the working impersonated signer", async function () {
+          const address = `0x${"ff".repeat(20)}`;
+          await assert.isRejected(
```
fvictorio, line 153: This doesn't seem to match the description of the test

packages/hardhat-ethers/test/index.ts
```diff
@@ -143,6 +147,17 @@ describe("Ethers plugin", function () {
         });
       });
 
+      describe("getImpersonatedSigner", function () {
+        it("should invoke hardhat_impersonateAccount", async function () {
+          const address = `0x${"ff".repeat(20)}`;
+          await assert.isRejected(
```
alcuadrado, line 153:
```suggestion
          // TODO: We are testing this plugin against Ganache, so this fails.
          //  We should test it using Hardhat Network instead.
          await assert.isRejected(
```

alcuadrado (approved):

PR #2810, opened 2022-06-08
Add hardhat-toolbox

fvictorio:
Things not included:

- README
- Description in package.json


packages/hardhat-toolbox/package.json
```diff
@@ -0,0 +1,89 @@
...
+    "mocha": "^9.2.0",
+    "prettier": "2.4.1",
+    "rimraf": "^3.0.2",
+    "solidity-coverage": "^0.7.21",
+    "ts-node": "^8.1.0",
+    "typechain": "^8.0.0",
+    "typescript": "~4.5.2"
+  },
+  "peerDependencies": {
+    "@ethersproject/abi": "^5.1.2",
```
fvictorio, line 68: This dependency is technically wrong. The peer dependency version in typechain is `^5.4.7` (see [here](https://github.com/dethcrypto/TypeChain/blob/015abb28bd22826611051f27e0ec96a00f9a0b61/packages/hardhat/package.json#L51)). I used this one to comply with the `check-dependencies` script. What should we do here?

Same thing for the ethers dependency.

packages/hardhat-toolbox/src/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  // hardhat-gas-reporter doesn't use extendConfig, so
+  // config.gasReporter === userConfig.gasReporter
+  // we use userConfig.gasReporter here for the types
+  const gasReporterConfig = userConfig.gasReporter;
+
+  configAsAny.gasReporter = gasReporterConfig ?? {};
+
+  if (gasReporterConfig?.enabled === undefined) {
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    configAsAny.gasReporter.enabled = process.env.REPORT_GAS ? true : false;
```
alcuadrado, line 23: `gasReporter` can be `undefined` here.

I couldn't cast config to any.

fvictorio: That is handled by the previous line:

```js
configAsAny.gasReporter = gasReporterConfig ?? {};
```

The reason I'm casting the config to any is that hardhat-gas-reporter [doesn't extend the `UserConfig`](https://github.com/cgewecke/hardhat-gas-reporter/blob/8b3ad6d5b8d823faf1c1beee0b900cdc3a6335f2/src/type-extensions.ts), so `gasReporter` is not there at all.

packages/hardhat-toolbox/src/index.ts
```diff
@@ -0,0 +1,34 @@
...
+import "@nomiclabs/hardhat-ethers";
+import "@nomiclabs/hardhat-etherscan";
+import "@typechain/hardhat";
+import "hardhat-gas-reporter";
+import "solidity-coverage";
+/**
+ * If a new official plugin is added, make sure that the tsconfig.json file and
+ * the hardhat-toolbox workflow are updated. The parts of the documentation that
+ * install hardhat-toolbox with npm 6 or yarn must also be updated.
+ */
```
alcuadrado, line 12:
```suggestion
/**
 * If a new official plugin is added, make sure that the tsconfig.json file and
 * the hardhat-toolbox workflow are updated. The parts of the documentation that
 * install hardhat-toolbox with npm 6 or yarn must also be updated. Finally, update
 * the list of dependencies that the sample projects install.
 */
```

PR #2813, opened 2022-06-09
New sample projects creation

alcuadrado:


packages/hardhat-core/sample-projects/javascript/test/Lock.js
```diff
@@ -0,0 +1,107 @@
...
+      it("Shouldn't fail if the unlockTime has arrived", async function () {
+        const { lock, unlockTime } = await loadFixture(deployOneYearLock);
+
+        // We can increase the time in Hardhat Network
+        await time.increaseTo(unlockTime);
+
+        await expect(lock.withdraw()).not.to.be.reverted;
+      });
+
+      it("Shouldn't revert with the right error if called from another account", async function () {
```
feuGeneA, line 67:
```suggestion
      it("Should revert with the right error if called from another account", async function () {
```

packages/hardhat-core/sample-projects/javascript/test/Lock.js
```diff
@@ -0,0 +1,107 @@
...
+        const { lock, unlockTime } = await loadFixture(deployOneYearLock);
+        const [owner, other] = await ethers.getSigners();
+
+        await time.increaseTo(unlockTime);
+
+        // We use lock.connect() to send a transaction from another account
+        await expect(lock.connect(other).withdraw()).to.revertedWith(
```
feuGeneA, line 74:
```suggestion
        const [owner, otherAccount] = await ethers.getSigners();

        await time.increaseTo(unlockTime);

        // We use lock.connect() to send a transaction from another account
        await expect(lock.connect(otherAccount).withdraw()).to.revertedWith(
```

packages/hardhat-core/sample-projects/javascript/test/Lock.js
```diff
@@ -0,0 +1,107 @@
...
+    });
+
+    describe("Transfers", function () {
+      it("Should transfer the funds to the owner", async function () {
+        const { lock, unlockTime, lockedAmount, owner } = await loadFixture(
+          deployOneYearLock
+        );
+        await time.increaseTo(unlockTime);
+
+        await expect(lock.withdraw()).to.changeEtherBalance(
```
feuGeneA, line 100: Consider using `changeEtherBalances` (plural) here, to also check that the contract's Ether balance has changed by `-lockedAmount`.

packages/hardhat-core/sample-projects/typescript/scripts/deploy.ts
```diff
@@ -0,0 +1,23 @@
+import { ethers } from "hardhat";
```
feuGeneA, line 1: Consider adding the same comment header that appears in `sample-projects/javascript/scripts/deploy.ts`.

packages/hardhat-core/src/internal/cli/project-creation.ts
```diff
@@ -142,100 +109,47 @@ async function printWelcomeMessage() {
...
+  const existingFiles = sampleProjectRootFiles
+    .map((f) => path.join(projectRoot, f))
+    .filter((f) => fsExtra.pathExistsSync(f))
+    .map((f) => path.relative(process.cwd(), f));
+
+  if (existingFiles.length > 0) {
+    const errorMsg = `We couldn't initialize the sample project because ${pluralize(
+      existingFiles.length,
+      "this file exists",
+      "these files exist"
```
feuGeneA, line 141:
```suggestion
      "this file already exists",
      "these files already exist"
```

packages/hardhat-core/src/internal/cli/project-creation.ts
```diff
@@ -271,11 +185,9 @@ async function printRecommendedDepsInstallationInstructions(
 async function writeEmptyHardhatConfig() {
   return fsExtra.writeFile(
     "hardhat.config.js",
-    `/**
- * @type import('hardhat/config').HardhatUserConfig
- */
+    `/** @type import('hardhat/config').HardhatUserConfig */
 module.exports = {
-  solidity: "${DEFAULT_SOLC_VERSION}",
+  solidity: "0.8.9",
```
feuGeneA, line 190: This on purpose? Seems like something we'd want to keep parameterized.

feuGeneA (approved):

packages/hardhat-core/src/internal/cli/project-creation.ts
alcuadrado: I de-parametrized it to be consistent with the rest of the sample projects. We need to modify our "update solidity versions" doc to include this.

packages/hardhat-core/sample-projects/typescript/scripts/deploy.ts
alcuadrado: I didn't add the explanation because this is mandatory when you use typescript, which is expected by any ts user.

packages/e2e/test/index.ts
```diff
@@ -128,95 +128,62 @@ describe("e2e tests", function () {
...
-        });
-      });
-
-      for (const suggestedCommand of [
-        // This list should be kept reasonably in sync with
-        // hardhat-core/sample-projects/advanced/README.txt
-        `${hardhatBinary} compile`,
+        // packages/hardhat-core/sample-projects/javascript/README.md
+        `${hardhatBinary} help`,
         `${hardhatBinary} test`,
```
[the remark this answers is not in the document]
alcuadrado: I added it

packages/hardhat-core/sample-projects/javascript/test/Lock.js
alcuadrado: great idea!

packages/hardhat-core/sample-projects/javascript/test/Lock.js
alcuadrado: I implemented this change manually as I had already committed some changes

packages/hardhat-core/src/internal/cli/project-creation.ts
alcuadrado: Note that we can be more reckless here and just bump it whenever we want.

PR #2829, opened 2022-06-14
Update typescript guide

alcuadrado:


docs/src/content/hardhat-runner/docs/guides/typescript.md
```diff
@@ -10,188 +10,131 @@ Hardhat will automatically enable its TypeScript support if your config file end
 
 ### Installing dependencies
 
+:::tip
+
+If you installed [`@nomicfoundation/hardhat-toolbox`](../../plugins/nomicfoundation-hardhat-toolbox) using `npm` 7 or higher, you don't need to follow these steps.
```
fvictorio, line 15: I think we normally don't use backticks for "npm".

```suggestion
If you installed [`@nomicfoundation/hardhat-toolbox`](../../plugins/nomicfoundation-hardhat-toolbox) using npm 7 or higher, you don't need to follow these steps.
```

docs/src/content/hardhat-runner/docs/guides/typescript.md
```diff
@@ -10,188 +10,131 @@ Hardhat will automatically enable its TypeScript support if your config file end
...
-```typescript
-import { ethers } from "hardhat";
-import { Signer } from "ethers";
+```
+mv hardhat.config.js hardhat.config.ts
+```
 
-describe("Token", function () {
-  let accounts: Signer[];
+We need to a single change to your config for it to work with TypeScript: plugins must be loaded with `import` instead of `require`.
```
fvictorio, line 105: Is this correct? You also need to change the `module.exports` to `export default`, so it's not a single change.

docs/src/content/hardhat-runner/docs/guides/typescript.md
```diff
@@ -10,188 +10,131 @@ Hardhat will automatically enable its TypeScript support if your config file end
...
-```typescript
-import { run, ethers } from "hardhat";
+into this:
 
-async function main() {
-  await run("compile");
+<<< @/../packages/hardhat-core/sample-projects/typescript/hardhat.config.ts{1,2,4-6}
 
-  const accounts = await ethers.getSigners();
+Finally, you need to create a [`tsconfig.json`](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) file. Here's our recommended one:
```
fvictorio, line 117: Shouldn't we use `es2020`? Or does `es2017` support native bigints?

docs/src/content/hardhat-runner/docs/guides/typescript.md
```diff
@@ -10,188 +10,131 @@ Hardhat will automatically enable its TypeScript support if your config file end
...
+If you installed [`@nomicfoundation/hardhat-toolbox`](../../plugins/nomicfoundation-hardhat-toolbox) you can skip this section, as it includes [`@typechain/hardhat`](https://github.com/ethereum-ts/TypeChain/tree/master/packages/hardhat).
 
-const config: HardhatUserConfig = {
-  // Your type-safe config goes here
-};
+:::
 
-export default config;
-```
+If you want Hardhat to generate types for your smart contract you should install use [`@typechain/hardhat`](https://github.com/ethereum-ts/TypeChain/tree/master/packages/hardhat). It generates typing files (`*.d.ts`) based on ABI's, and it requires little to no configuration.
```
fvictorio, line 137:
```suggestion
If you want Hardhat to generate types for your smart contract you should install [`@typechain/hardhat`](https://github.com/ethereum-ts/TypeChain/tree/master/packages/hardhat). It generates typing files (`*.d.ts`) based on ABI's, and it requires little to no configuration.
```

fvictorio: Or maybe you meant `install and use`? Otherwise this might seem like only installing typechain is enough.

docs/src/content/hardhat-runner/docs/guides/typescript.md
alcuadrado: You are right!

docs/src/content/hardhat-runner/docs/guides/typescript.md
alcuadrado: We can't use it until node 12 gets EOL'd. ES2020 doesn't downcompiler `{...this}`, and node 12 doesn't support it.

fvictorio: I see, thanks for the explanation.

PR #2831, opened 2022-06-14
Write first version of the "Testing contracts" guide

fvictorio:
Also closes HH-822.


docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,197 @@
-# Testing your contracts
+# Testing contracts
 
-\<New guide, TODO>
+After [compiling your contracts](./compile-contracts.md), the next step is to write some tests to verify that they work as intended. This guide explains our recommended approach for testing contracts in Hardhat. It relies on [ethers](https://docs.ethers.io/v5/) to connect to the Hardhat Network and [Mocha](https://mochajs.org/) and [Chai](https://www.chaijs.com/) to write the tests. It also uses our custom [Chai matchers](/hardhat-chai-matchers) and our [Hardhat Network Helpers](/hardhat-network-helpers) to make it easier to write contracts-related tests.
```
alcuadrado, line 3: Maybe mention that they come installed with hardhat toolbox, and you would have them if you followed the other guides

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,197 @@
...
+
+While this is the way we recommend, Hardhat is flexible and you can customize the approach or take a completely different path with other tools.
+
+### Initial setup
+
+In this guide we’ll write some tests for the sample project so, if you haven’t done it yet, go and [initialize it](./project-setup.md).
+
+We also recommend you use [TypeScript](./typescript.md) to get better autocompletion and catch possible errors earlier. This guide will assume you are using TypeScript, but you can click the tabs of the snippets to switch their language.
+
+The setup includes some example tests in the `test/Lock.ts` file, but ignore them for now. Instead, create a `test/my-tests.ts` file directory and run only those. For example, if your test file is called `test/my-tests.ts`, you can run only that file with `npx hardhat test test/my-tests.ts`
```
alcuadrado, line 13:
```suggestion
The setup includes some example tests in the `test/Lock.ts` file, but ignore them for now. Instead, create a `test/my-tests.ts` file. During this guide we'll only run those, by running `npx hardhat test test/my-tests.ts`, instead of just `npx hardhat test`.
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,197 @@
...
+
+### A simple test
+
+For our first test we’ll deploy the `Lock` contract and assert that the unlock time returned by the `unlockTime()` getter is the same one that we passed in the constructor:
```
alcuadrado, line 17: > The sample project comes with a Lock contract. We'll write tests for it. If you haven't read it yet, please take a look at `contract/Lock.sol`.

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,197 @@
...
+  });
+});
+```
+
+The first thing we do is to import the functions we are going to use: the [`expect`](https://www.chaijs.com/api/bdd/) function from `chai` to write our assertions, the [Hardhat Runtime Environment](../advanced/hardhat-runtime-environment.md), and the [network helpers](/hardhat-network-helpers) to interact with the Hardhat Network. After that we use the `describe` and `it` functions, which are global Mocha's functions used to describe and group your tests (you can read more about Mocha [here](https://mochajs.org/#getting-started)).
+
+The test itself is what’s inside the `it` function. First we set the values for the amount we want to lock (in [wei](https://ethereum.org/en/glossary/#wei)) and the unlock time. For the latter we use [`time.latest`](</hardhat-network-helpers/docs/reference#latest()>), a network helper that returns the timestamp of the last mined block. Then we deploy the contract itself: first we get a [`ContractFactory`](https://docs.ethers.io/v5/single-page/#/v5/api/contract/contract-factory/) for the `Lock` contract and then we deploy it, passing the unlock time as its only constructor argument.
```
alcuadrado, line 43: We also pass it a value

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,197 @@
...
+  // ...deploy the contract...
+
+  await time.increaseTo(unlockTime);
+
+  // this will throw if the transaction reverts
+  await lock.withdraw();
+});
+```
+
+The `time.increaseTo` helper mines a new block with the given timestamp. After doing that, we can call the `withdraw` function again. In ethers, if a function reverts, the method will return a rejected promise. Awaiting the result then means that the test will fail if the function reverts.
```
alcuadrado, line 88: > In ethers, if a function reverts, the method will return a rejected promise. Awaiting the result then means that the test will fail if the function reverts.

This is a bit confusing.

What about something like:

As we mentioned, calling `lock.withdraw()` returns a `Promise`. If the transaction fails, the promise will be rejected. Using `await` will throw in that case, so the test will fail.

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,197 @@
...
+
+### Running tests in parallel
+
+You can run your tests in parallel by using the `--parallel` flag:
+
+```
+npx hardhat test --parallel
+```
+
+Alternatively, use `parallel: true` in the `mocha` section of your Hardhat config, which will enable parallelism for plugins that don't have a `--parallel` flag, like `solidity-coverage`.
```
alcuadrado, line 188: I think this paragraph creates more confusion than anything else.

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,197 @@
...
+    const Lock = await ethers.getContractFactory("Lock");
+    const lock = await Lock.deploy(unlockTime, { value: lockedAmount });
+
+    // assert that the value is correct
+    expect(await lock.unlockTime()).to.equal(unlockTime);
+  });
+});
+```
+
+The first thing we do is to import the functions we are going to use: the [`expect`](https://www.chaijs.com/api/bdd/) function from `chai` to write our assertions, the [Hardhat Runtime Environment](../advanced/hardhat-runtime-environment.md), and the [network helpers](/hardhat-network-helpers) to interact with the Hardhat Network. After that we use the `describe` and `it` functions, which are global Mocha's functions used to describe and group your tests (you can read more about Mocha [here](https://mochajs.org/#getting-started)).
```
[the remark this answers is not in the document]
fvictorio: I don't like the repetition of "thing/things". What about "First we'll import the things we are going to use:"

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
fvictorio: Added a sentence in ec8735dfe79fedc6173deec80d7ea28517b4a382

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
fvictorio: See 1f4ede524fbd3aebc3a68c706b38b2ad89efdb65

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
fvictorio: 0f82fdf3a0a9c51f3faa54ebba1a3afef4ebb04f

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
fvictorio: What about just removing everything after "which will enable..."?

fvictorio: I'm ok with just deleting the whole paragraph though.

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
alcuadrado: Sorry for not being clear. Adding something like that at the beginning

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
alcuadrado: I'd be more explicit. Something like "We also pass it an object with transaction (not constructor) parameters. This is optional, but will use it to send some ETH by setting its `value` field."

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
alcuadrado: > What about removing everything after "which will enable..."?

Yeah, that works

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
fvictorio: Makes sense, added in 61e9fba2002e8ee38626fc4634accef6ec424c9a

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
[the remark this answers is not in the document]
fvictorio: 0ac79e3cc18f65533557e0dd0d1116079dd07aaf

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
fvictorio: Added something like that in 834aa5eac25661b1d40caf699e2442e7aef31978

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+
+While this is the way we recommend, Hardhat is flexible and you can customize the approach or take a completely different path with other tools.
+
+### Initial setup
+
+In this guide we’ll write some tests for the sample project so, if you haven’t done it yet, go and [initialize it](./project-setup.md).
+
+We also recommend you use [TypeScript](./typescript.md) to get better autocompletion and catch possible errors earlier. This guide will assume you are using TypeScript, but you can click the tabs of the snippets to switch their language.
```
feuGeneA, line 11:
```suggestion
We also recommend you [use TypeScript](./typescript.md) to get better autocompletion and catch possible errors earlier. This guide will assume you are using TypeScript, but you can click the tabs of the snippets to switch their language.
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+    const Lock = await ethers.getContractFactory("Lock");
+    const lock = await Lock.deploy(unlockTime, { value: lockedAmount });
+
+    // assert that the value is correct
+    expect(await lock.unlockTime()).to.equal(unlockTime);
+  });
+});
+```
+
+First we'll import the things we are going to use: the [`expect`](https://www.chaijs.com/api/bdd/) function from `chai` to write our assertions, the [Hardhat Runtime Environment](../advanced/hardhat-runtime-environment.md), and the [network helpers](/hardhat-network-helpers) to interact with the Hardhat Network. After that we use the `describe` and `it` functions, which are global Mocha's functions used to describe and group your tests (you can read more about Mocha [here](https://mochajs.org/#getting-started)).
```
feuGeneA, line 43:
```suggestion
First we import the things we are going to use: the [`expect`](https://www.chaijs.com/api/bdd/) function from `chai` to write our assertions, the [Hardhat Runtime Environment](../advanced/hardhat-runtime-environment.md) (`hre`), and the [network helpers](/hardhat-network-helpers) to interact with the Hardhat Network. After that we use the `describe` and `it` functions, which are global Mocha functions used to describe and group your tests. (You can read more about Mocha [here](https://mochajs.org/#getting-started).)
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+
+The test itself is what’s inside the `it` function. First we set the values for the amount we want to lock (in [wei](https://ethereum.org/en/glossary/#wei)) and the unlock time. For the latter we use [`time.latest`](</hardhat-network-helpers/docs/reference#latest()>), a network helper that returns the timestamp of the last mined block. Then we deploy the contract itself: first we get a [`ContractFactory`](https://docs.ethers.io/v5/single-page/#/v5/api/contract/contract-factory/) for the `Lock` contract and then we deploy it, passing the unlock time as its constructor argument. We also pass an object with the transaction parameters. This is optional, but we'll use it to send some ETH by setting its `value` field.
```
feuGeneA, line 45:
```suggestion
The test itself is what’s inside the callback argument to the `it` function. First we set the values for the amount we want to lock (in [wei](https://ethereum.org/en/glossary/#wei)) and the unlock time. For the latter we use [`time.latest`](</hardhat-network-helpers/docs/reference#latest()>), a network helper that returns the timestamp of the last mined block. Then we deploy the contract itself: first we get a [`ContractFactory`](https://docs.ethers.io/v5/single-page/#/v5/api/contract/contract-factory/) for the `Lock` contract and then we deploy it, passing the unlock time as its constructor argument. We also pass an object with the transaction parameters. This is optional, but we'll use it to send some ETH by setting its `value` field.
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+
+Finally, we check that the value returned by the `unlockTime()` getter in the contract matches the value that we used when we deployed it. Since all the functions in a contract are async, we have to use the `await` keyword to get its value, otherwise we would be comparing a promise with a number and this would always fail.
```
feuGeneA, line 47:
```suggestion
Finally, we check that the value returned by the `unlockTime()` [getter](https://docs.soliditylang.org/en/v0.8.13/contracts.html#getter-functions) in the contract matches the value that we used when we deployed it. Since all the functions on a contract are async, we have to use the `await` keyword to get its value; otherwise, we would be comparing a promise with a number and this would always fail.
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+
+```tsx
+it("Should revert with the right error if called too soon", async function () {
+  // ...deploy the contract as before...
+
+  await expect(lock.withdraw()).to.be.revertedWith("You can't withdraw yet");
+});
+```
+
+In the previous test we used `.to.equal`, which is part of Chai and it’s used to compare two values, but here we are using [`.to.be.revertedWith`](/hardhat-chai-matchers/docs/reference#.revertedwith). This matcher asserts that a transaction reverts, and that the reason string of the revert is equal to a given string. The `.to.be.revertedWith` matcher is not part of Chai itself; instead, it’s added by the [Hardhat Chai Matchers](/hardhat-chai-matchers) plugin, which is included in the sample project we are using.
```
feuGeneA, line 69:
```suggestion
In the previous test we used `.to.equal`, which is part of Chai and is used to compare two values. Here we are using [`.to.be.revertedWith`](/hardhat-chai-matchers/docs/reference#.revertedwith), which asserts that a transaction reverts, and that the reason string of the revert is equal to the given string. The `.to.be.revertedWith` matcher is not part of Chai itself; instead, it’s added by the [Hardhat Chai Matchers](/hardhat-chai-matchers) plugin, which is included in the sample project we are using.
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+
+Notice that in the previous test we wrote `expect(await ...)` but now we are doing `await expect(...)`. In the first case we were comparing two values in a synchronous way; the inner await is just there to wait for the value to be retrieved. In the second case, the whole assertion is async, because it has to wait until the transaction is mined. This means that the `expect` call returns a promise that we have to wait.
```
feuGeneA, line 71:
```suggestion
Notice that in the previous test we wrote `expect(await ...)` but now we are doing `await expect(...)`. In the first case we were comparing two values in a synchronous way; the inner await is just there to wait for the value to be retrieved. In the second case, the whole assertion is async because it has to wait until the transaction is mined. This means that the `expect` call returns a promise that we have to await.
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+
+### Manipulating the time of the network
+
+We are deploying our `Lock` contract with an unlock time of one year. If we want to write a test that checks what happens after the unlock time has passed, we can’t wait that amount of time. We could use a shorter amount, like 5 seconds, but that’s a less realistic value and it's still a long time to wait in a test.
```
feuGeneA, line 75:
```suggestion
We are deploying our `Lock` contract with an unlock time of one year. If we want to write a test that checks what happens after the unlock time has passed, we can’t wait that amount of time. We could use a shorter unlock time, like 5 seconds, but that’s a less realistic value and it's still a long time to wait in a test.
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+  // ...deploy the contract...
+
+  await time.increaseTo(unlockTime);
+
+  // this will throw if the transaction reverts
+  await lock.withdraw();
+});
+```
+
+As we mentioned, calling `lock.widthdraw()` returns a Promise. If the transaction fails, the promise will be rejected. Using `await` will throw in that case, so the test will fail if the transaction reverts.
```
feuGeneA, line 90:
```suggestion
As we mentioned, calling `lock.withdraw()` returns a Promise. If the transaction fails, the promise will be rejected. Using `await` will throw in that case, so the test will fail if the transaction reverts.
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+The solution is to simulate the passage of time. This can be done with the [`time.increaseTo`](</hardhat-network-helpers/docs/reference#increaseto(timestamp)>) network helper, which mines a new block with the given timestamp:
+
+```tsx
+it("Should transfer the funds to the owner", async function () {
+  // ...deploy the contract...
+
+  await time.increaseTo(unlockTime);
+
+  // this will throw if the transaction reverts
+  await lock.withdraw();
```
feuGeneA, line 86: how about using `expect(...).not.to.be.reverted` here?

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+});
+```
+
+As we mentioned, calling `lock.widthdraw()` returns a Promise. If the transaction fails, the promise will be rejected. Using `await` will throw in that case, so the test will fail if the transaction reverts.
+
+### Using a different address
+
+The second check done by the `withdraw` function is that the function was called by the owner of the contract. In these tests, all the deployments and function calls are done by default with the first [configured account](/hardhat-network/docs/reference.md#accounts). If we want to check that only the owner can call some function, we need to use a different account and verify that it fails.
```
feuGeneA, line 94:
```suggestion
The second check done by the `withdraw` function is that the function was called by the owner of the contract. By default, deployments and function calls are done with the first [configured account](/hardhat-network/docs/reference.md#accounts). If we want to check that only the owner can call some function, we need to use a different account and verify that it fails.
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+
+The `ethers.getSigners()` returns an array with all the configured accounts. We can then use the `.connect` method of the contract to call the function with a different account and check that the transaction reverts:
```
feuGeneA, line 96:
```suggestion
The `ethers.getSigners()` returns an array with all the configured accounts. We can use the `.connect` method of the contract to call the function with a different account and check that the transaction reverts:
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+});
+```
+
+Here again we are calling a function and asserting that it reverts with the correct reason string. The difference is that we are using `.connect(anotherAccount)` to call the method from a different address.
+
+### Using fixtures
+
+So far we've deployed the `Lock` contract in each test. This means that at the beginning of each test we have to get the contract factory and then deploy the contract. This might be fine for a single contract but, if you have a more complicated setup, each test will have several lines at the begginning just to set up the desired state, and most of the time these lines will be the same.
+
+In Mocha, this duplication of code is handled with a `beforeEach` hook:
```
feuGeneA, line 120:
```suggestion
In a typical Mocha test, this duplication of code is handled with a `beforeEach` hook:
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+  const Lock = await ethers.getContractFactory("Lock");
+  lock = await Lock.deploy(unlockTime, { value: lockedAmount });
+});
+
+it("some test", async function () {
+  // use the deployed contract
+});
+```
+
+There are two problems with this approach:
```
feuGeneA, line 139:
```suggestion
However, there are two problems with this approach:
```

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+
+So far we've deployed the `Lock` contract in each test. This means that at the beginning of each test we have to get the contract factory and then deploy the contract. This might be fine for a single contract but, if you have a more complicated setup, each test will have several lines at the begginning just to set up the desired state, and most of the time these lines will be the same.
+
+In Mocha, this duplication of code is handled with a `beforeEach` hook:
+
+```tsx
+let lock;
+let unlockTime;
+let lockedAmount = 1_000_000_000;
+beforeEach(async function () {
```
feuGeneA, line 126: just for complete accuracy, shouldn't there be a `describe` block here that contains both the `beforeEach` and the `it`?

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -1,3 +1,215 @@
...
+
+The Hardhat Toolbox includes the [`solidity-coverage`](https://github.com/sc-forks/solidity-coverage) plugin to measure the test coverage in your project. Just run the `coverage` task and you'll get a report:
+
+```
+npx hardhat coverage
+```
+
+### Using the gas reporter
+
+The Hardhat Toolbox includes the [`hardhat-gas-reporter`](https://github.com/cgewecke/hardhat-gas-reporter) plugin to get metrics of how much gas is used, based in the execution of your tests. The gas reporter is run when the `test` task is executed and the `REPORT_GAS` environment variable is set:
```
feuGeneA, line 192:
```suggestion
The Hardhat Toolbox also includes the [`hardhat-gas-reporter`](https://github.com/cgewecke/hardhat-gas-reporter) plugin to get metrics of how much gas is used, based on the execution of your tests. The gas reporter is run when the `test` task is executed and the `REPORT_GAS` environment variable is set:
```

feuGeneA (changes requested):

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
fvictorio: Done in 80c22a60f51e341f8a39613a180ceadad892fe20, thanks

docs/src/content/hardhat-runner/docs/guides/test-contracts.md
fvictorio: I'm not sure, I think I personally would use this simpler approach. `.not.to.be.reverted` might give you a more specific error message, but if a tx fails it's easy to figure out what happened. More importantly, you won't get a proper stack trace  if you use the matcher (at least until we get that fixed).

feuGeneA (approved):

Issue #2836, opened 2022-06-15
hardhat-chai-matchers depends on sinon-chai


vimwitch:
The @nomicfoundation/hardhat-chai-matchers package currently includes a `types.d.ts` file with this line:
```
/// <reference types="sinon-chai" />
```

This causes a build error when the `@types/sinon-chai` package isn't available. Ideally we wouldn't install this package, is it possible to remove this from the output?

alcuadrado:
Thanks for this bug report, @vimwitch! I fixed it. You can read how that package ended up there in the PR.

vimwitch:
Thanks @alcuadrado, great response time 🤩

PR #2843, opened 2022-06-15
Remove incorrect sinon-chai reference from hardhat-chai-matchers

alcuadrado:
This PR deserves a thorough explanation:

As reported in #2836, `hardhat-chai-matchers` was incorrectly adding a `/// <reference types="sinon-chai" />`. This was surprising, as we don't use Sinon in that package.

What happened is that tsc loads all the typing packages from the `typeRoots` you declare by default. In our case, those are set to `<repo>/node_modules/@types` and `<repo>/config/typescript/@types`. Inside the main `@types` we have both `chai` and `chai-sinon`, and given how the Chai overload is written (not a declare module construct), it needs to load both.

I fixed it by explicitly telling tsc to only load `@types/node` and `@types/chai`.

### DIFF
--- a/packages/hardhat-chai-matchers/src/types.ts
+++ b/packages/hardhat-chai-matchers/src/types.ts
@@ -1,5 +1,3 @@
-declare module "deep-eql";
-
 // eslint-disable-next-line @typescript-eslint/no-namespace, @typescript-eslint/no-unused-vars
 declare namespace Chai {
   interface Assertion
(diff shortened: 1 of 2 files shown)


packages/hardhat-chai-matchers/src/types.ts
```diff
@@ -1,5 +1,3 @@
-declare module "deep-eql";
```
fvictorio, line 1: Oh, this was necessary when `deep-eql` was a top-level import. After we moved it to a lazy require (not an `await import` because it's used inside a sync function), this module declaration became unnecessary.

fvictorio:
@alcuadrado do we need to do this in the other packages?

PR #2867, opened 2022-06-18
WIP: Update tutorial

alcuadrado:


fvictorio:
Some comments about this:

- We should include an `npm init -y` command in the setup section, right? Otherwise no `package.json` is created.
- Also in the setup part: should we include a mention about npm >= 7? If users just blindly copy the commands and they are using npm 6, then they'll get a very confusing error when they run Hardhat at the end. (For the record, this happened to me because apparently `volta install node@14` gets you npm 6.)
- Add a link to the ethers.js docs when ethers is first mentioned
- The second test that is shown, the one that uses a different account, is under a different `describe` suit. I think we should use the same suite, with a `// ...` comment or something like that indicating that we are adding a new test.

docs/src/content/tutorial/setting-up-the-environment.md
```diff
@@ -1,20 +1,16 @@
...
 [Hardhat for Visual Studio Code](/hardhat-vscode) is the official Hardhat extension that adds advanced support for Solidity to VSCode. If you use Visual Studio Code, give it a try!
 
 :::
 
-Most Ethereum libraries and tools are written in JavaScript, and so is **Hardhat**. If you're not familiar with Node.js, it's a JavaScript runtime built on Chrome's V8 JavaScript engine. It's the most popular solution to run JavaScript outside of a web browser and **Hardhat** is built on top of it.
-
 ## Installing Node.js
 
-You can [skip](./creating-a-new-hardhat-project.md) this section if you already have a working Node.js `>=12.0` installation. If not, here's how to install it on Ubuntu, MacOS and Windows.
+You can [skip](./creating-a-new-hardhat-project.md) this section if you already have a working Node.js `>=14.0` installation. If not, here's how to install it on Ubuntu, MacOS and Windows.
```
alcuadrado, line 13: We should probably use 16, to make sure that they have npm 7+

alcuadrado:
> Some comments about this:
> 
> * We should include an `npm init -y` command in the setup section, right? Otherwise no `package.json` is created.

Makes sense. Maybe without the `-y`, but tell them to follow the steps instead.

> * Also in the setup part: should we include a mention about npm >= 7? If users just blindly copy the commands and they are using npm 6, then they'll get a very confusing error when they run Hardhat at the end. (For the record, this happened to me because apparently `volta install node@14` gets you npm 6.)

Yes, as node 14 comes with npm 6. We should probably use a tabbed snippet.

> * Add a link to the ethers.js docs when ethers is first mentioned

Makes sense

> * The second test that is shown, the one that uses a different account, is under a different `describe` suit. I think we should use the same suite, with a `// ...` comment or something like that indicating that we are adding a new test.

LGTM

docs/src/content/tutorial/boilerplate-project.md
```diff
@@ -73,7 +73,7 @@ What's happening here is that the frontend code to show the current wallet's bal
 npx hardhat --network localhost faucet <your address>
 ```
 
-You'll run a custom **Hardhat** task we included that uses the balance of the deploying account to send 100 MBT and 1 ETH to your address. This will allow you to send tokens to another address.
+You'll run a custom Hardhat task we included that uses the balance of the deploying account to send 100 MBT and 1 ETH to your address. This will allow you to send tokens to another address.
```
feuGeneA, line 76:
```suggestion
You'll run a custom Hardhat task we included that uses the balance of the deploying account to send 100 MHT and 1 ETH to your address. This will allow you to send tokens to another address.
```

feuGeneA (changes requested):
On the "Boilerplate Project" page, the screenshot of the transfer interface is still showing My Buidler Token instead of My Hardhat Token.

feuGeneA (approved):

docs/src/content/tutorial/writing-and-compiling-contracts.md
```diff
@@ -72,9 +78,19 @@ contract Token {
         // transaction will revert.
         require(balances[msg.sender] >= amount, "Not enough tokens");
 
+        // We can print messages and values using console.log
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/2867/commits/71ccfb181a53e8da248a81d3636ebd8c0efa69d4

feuGeneA:
> On the "Boilerplate Project" page, the screenshot of the transfer interface is still showing My Buidler Token instead of My Hardhat Token.

Addressed in https://github.com/NomicFoundation/hardhat/pull/2867/commits/03dcdb445cb5cff18943f4f342885f5722fc8372

docs/src/content/tutorial/setting-up-the-environment.md
alcuadrado, line 13:
```suggestion
You can [skip](./creating-a-new-hardhat-project.md) this section if you already have a working Node.js `>=16.0` installation. If not, here's how to install it on Ubuntu, MacOS and Windows.
```

docs/src/content/tutorial/testing-contracts.md
```diff
@@ -111,90 +117,169 @@ describe("Transactions", function() {
 });
 ```
 
-### Full coverage
+### Reusing common test setups using fixtures
+
+The two tests that we've written begin with the necessary setup, which in this case means deploying the token contract. In more complex projects, this setup could involve multiple deployments and other transactions. Doing that in every test means a lot of code duplication. Plus, executing many transactions at the beginning of each test can make the test suite much slower.
```
alcuadrado, line 122:
```suggestion
The two tests that we wrote begin with their setup, which in this case means deploying the token contract. In more complex projects, this setup could involve multiple deployments and other transactions. Doing that in every test means a lot of code duplication. Plus, executing many transactions at the beginning of each test can make the test suite much slower.
```

docs/src/content/tutorial/testing-contracts.md
```diff
@@ -111,90 +117,169 @@ describe("Transactions", function() {
...
 
-Now that we've covered the basics you'll need for testing your contracts, here's a full test suite for the token with a lot of additional information about Mocha and how to structure your tests. We recommend reading through.
+You can solve these problems by using **fixtures**. A fixture is a setup function that is run only the first time it's invoked. On subsequent invocations, instead of re-running it, Hardhat will reset the state of the network to what it was at the point after the fixture was initially executed.
```
alcuadrado, line 124:
```suggestion
You can avoid code duplication and improve the performance of your test suite by **fixtures**. A fixture is a setup function that is run only the first time it's invoked. On subsequent invocations, instead of re-running it, Hardhat will reset the state of the network to what it was at the point after the fixture was initially executed.
```

docs/src/content/tutorial/testing-contracts.md
```diff
@@ -111,90 +117,169 @@ describe("Transactions", function() {
...
+    // Transfer 50 tokens from addr1 to addr2
+    // We use .connect(signer) to send a transaction from another account
+    await expect(
+      hardhatToken.connect(addr1).transfer(addr2.address, 50)
+    ).to.changeTokenBalances(hardhatToken, [addr1, addr2], [-50, 50]);
+  });
+});
+```
+
+Here we wrote a `deployTokenFixture` function that does the necessary setup and returns the values we might care about in our tests. Then in each test we use `loadFixture` to run the fixture and get those values.
```
alcuadrado, line 169:
```suggestion
Here we wrote a `deployTokenFixture` function that does the necessary setup and returns every value we use later in the tests. Then in each test, we use `loadFixture` to run the fixture and get those values. `loadFixture` will make sure to run the setup the first time, and quickly return to that state in the other tests.
```

alcuadrado: I know this is repeated, but it's worth insisting.

PR #2441, opened 2022-03-02
Add support for Cronoscan and BTTCScan

0xV4L3NT1N3:

PR #2871, opened 2022-06-21
add Cronos ChainID 25 support to hardhat verify

pepperstepper:

---

I added Cronos support to my local project to get verification going and thought I'd add it to upstream

In my local version I had to edit the types.ts file too, as suggested here:
https://github.com/NomicFoundation/hardhat/commit/e04944c6c4a69a52b09f83fce759dcb9cbb32777#diff-450405d52fcde8250684aaf117939e95485b9cf7a80259e0bee6a0f4f3f1e6c5

but that has since been changed to be less strict it seems here so isn't needed in this PR:
https://github.com/NomicFoundation/hardhat/commit/17e95b239e7e53ae46e2f41dde545c3f6a406975

### DIFF
--- a/packages/hardhat-etherscan/src/ChainConfig.ts
+++ b/packages/hardhat-etherscan/src/ChainConfig.ts
@@ -205,4 +205,11 @@ export const chainConfig: ChainConfig = {
       browserURL: "https://explorer.pops.one",
     },
   },
+  cronos: {
+    chainId: 25,
+    urls: {
+      apiURL: "https://api.cronoscan.com/api",
+      browserURL: "https://cronoscan.com",
+    },
+  },
 };


zephinzer:
bumping. this would be very useful for me

fvictorio:
Hi @pepperstepper, thanks for sending this PR, but please read [this comment](https://github.com/NomicFoundation/hardhat/pull/2441#issuecomment-1116382707).

Since many people seem interested in this, I think a `hardhat-cronos` community plugin that adds support for Cronos would be useful, and if someone wants to give it a try, I'd be glad to help. You can use [this](https://github.com/fvictorio/hardhat-custom-chain-plugin-example) as an example to start from.

pepperstepper:
A flag where it prompts the user for the fields via commandline could be a good solution for the core hardhat-etherscan. Really verification is a universally good thing and the lower the bar across chains the better.

PR #2873, opened 2022-06-22
Prompt for installing Hardhat for VSCode extension 

morgansliman:


packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -22,14 +23,20 @@ import { isRunningOnCiServer } from "../util/ci-detection";
...
 
 import { applyWorkaround } from "../util/antlr-prototype-pollution-workaround";
 import { Analytics } from "./analytics";
 import { ArgumentsParser } from "./ArgumentsParser";
 import { enableEmoji } from "./emoji";
-import { confirmTelemetryConsent, createProject } from "./project-creation";
+import { createProject } from "./project-creation";
+import {
+  confirmTelemetryConsent,
+  confirmExtensionInstallation,
```
feuGeneA, line 38: Similarly, I'd like to see this identifier renamed to answer the question "_what_ extension?" Perhaps `confirmHHVSCodeInstallation`?

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -226,6 +233,24 @@ async function main() {
...
+    const alreadyPrompted = hasPromptedForExtension();
+    if (!alreadyPrompted) {
+      const installationConsent = await confirmExtensionInstallation();
+
+      if (installationConsent !== undefined) {
+        if (installationConsent) {
+          execSync("code --install-extension NomicFoundation.hardhat-solidity");
+        } else {
+          console.log(
+            "To learn more about Hardhat for Visual Studio Code, go to https://hardhat.org/hardhat-vscode"
```
feuGeneA, line 247: @alcuadrado do we want to have this banner even after they've denied consent to install? Guessing yes but wanted to have you confirm.

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -22,14 +23,20 @@ import { isRunningOnCiServer } from "../util/ci-detection";
 import {
   hasConsentedTelemetry,
   writeTelemetryConsent,
+  hasPromptedForExtension,
+  writePromptedForExtension,
```
feuGeneA, line 27: I'd like to see these identifiers renamed to answer the question "_what_ extension?" Maybe `hasPromptedForHHVSCode`?

packages/hardhat-core/src/internal/cli/project-creation.ts
```diff
@@ -30,10 +36,6 @@ type SampleProjectTypeCreationAction =
   | Action.CREATE_ADVANCED_SAMPLE_PROJECT_ACTION
   | Action.CREATE_ADVANCED_TYPESCRIPT_SAMPLE_PROJECT_ACTION;
 
-interface Dependencies {
-  [name: string]: string;
-}
-
```
feuGeneA, line 36: Why move this one? It feels more appropriate here than in `prompt.ts`, given its use in the signatures of other functions in this module, and given that "dependencies" aren't really inherent to the "prompt." But I do understand that it would have created a circular reference if you'd have left it here. Maybe it should get pushed into `src/types/index.ts`? Or maybe some other module in `src/types/`?

feuGeneA (changes requested):

packages/hardhat-core/src/internal/cli/cli.ts
morgansliman: he left a comment on the linear issue mentioning that this is the exact scenario when it should be shown

PR #2886, opened 2022-06-30
Hardhat VSCode prompt improvements

alcuadrado:
This PR introduces multiple UX improvements.

You can review it by commit, except for the latest one that introduces multiple wording improvements at once.


feuGeneA (approved):

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -60,6 +64,38 @@ To learn more about which versions of Node.js are supported go to https://hardha
...
 }
 
+async function suggestInstallingHardhatVscode() {
+  const alreadyPrompted = hasPromptedForHHVSCode();
+  if (alreadyPrompted) {
+    return;
+  }
+
+  const isInstalled = isHardhatVSCodeInstalled();
+
```
fvictorio, line 74: I think we should add something like this:

```js
  if (isInstalled === InstallationState.EXTENSION_INSTALLED) {
    writePromptedForHHVSCode();
    return;
  }
```

Otherwise the `test` task will be a bit slower for all Hardhat VSCode users. (I don't know if the difference would be noticeable, but it feels like the right thing to do anyway.)

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -60,6 +64,38 @@ To learn more about which versions of Node.js are supported go to https://hardha
...
+  if (isInstalled !== InstallationState.EXTENSION_NOT_INSTALLED) {
+    return;
+  }
+
+  const installationConsent = await confirmHHVSCodeInstallation();
+  writePromptedForHHVSCode();
```
fvictorio, line 80: This seems to contradict the comment of `confirmHHVSCodeInstallation`, which says that the prompt will be shown again if the functions returns `undefined`.

packages/hardhat-core/src/internal/cli/cli.ts
fvictorio: Update: on my machine, if you already have the extension installed, it adds at least 300-700 ms at the end of the test task.

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -60,6 +64,38 @@ To learn more about which versions of Node.js are supported go to https://hardha
...
+
+  if (installationConsent === true) {
+    const installed = installHardhatVSCode();
```
fvictorio, line 83: I would add a log here saying something like "Installing the extension..."

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -229,20 +265,8 @@ async function main() {
...
-          execSync("code --install-extension NomicFoundation.hardhat-solidity");
-        } else {
-          console.log(
-            "To learn more about Hardhat for Visual Studio Code, go to https://hardhat.org/hardhat-vscode"
-          );
-        }
-        writePromptedForHHVSCode();
-      }
+    if (taskName === "test") {
+      await suggestInstallingHardhatVscode();
```
fvictorio, line 269: I think you are already aware of this, but just in case: this condition also needs to include `!isRunningOnCiServer() && process.stdout.isTTY === true`

packages/hardhat-core/src/internal/cli/prompt.ts
```diff
@@ -154,3 +87,64 @@ export async function confirmProjectCreation(): Promise<{
...
+ * false = don't install and don't ask again
+ * undefined = don't install but maybe ask next time if something changes (i.e. they install VS Code)
+ */
+export async function confirmHHVSCodeInstallation(): Promise<
+  boolean | undefined
+> {
+  return confirmationPromptWithTimeout(
+    "shouldInstallExtension",
+    "Would you like to install the Hardhat for Visual Studio Code extension? It adds advanced editing assistance for Solidity to VSCode",
+    TELEMETRY_CONSENT_TIMEOUT
```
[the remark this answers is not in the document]
alcuadrado: Addressed in 5368b6359

packages/hardhat-core/src/internal/cli/cli.ts
alcuadrado: Done. Now I always store it.

packages/hardhat-core/src/internal/cli/cli.ts
alcuadrado: Updated the comment

packages/hardhat-core/src/internal/cli/cli.ts
alcuadrado: Added that log

fvictorio (approved):
Tested this:

- Without VSCode installed
- With VSCode installed but without the extension
- With VSCode installed and with the extension

I think the way this is implemented means that if you use this version of Hardhat but _don't_ have VSCode installed, but later install VSCode and start using it, you don't get the prompt. I think that trade-off is fine.

PR #2889, opened 2022-07-04
Remove suggestion to report error

fvictorio:
We don't have the bandwidth to process issues that come from this error, so we are removing that part of the message.


alcuadrado:
@fvictorio can you merge this so that it lands on the next release?

PR #2899, opened 2022-07-06
Updat peerDeps in toolbox and sample projects

fvictorio:


alcuadrado (commented):
I think this looks good

PR #2903, opened 2022-07-07
Prompt new users to star our repo

alcuadrado:
I based this PR on https://github.com/NomicFoundation/hardhat/pull/2867

### DIFF
--- a/packages/hardhat-core/src/internal/cli/project-creation.ts
+++ b/packages/hardhat-core/src/internal/cli/project-creation.ts
@@ -247,9 +247,10 @@ async function createPackageJson() {
 
 function showStarOnGitHubMessage() {
   console.log(
-    chalk.cyan("Give Hardhat a star on Github if you're enjoying it") +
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/src/internal/cli/project-creation.ts
```diff
@@ -245,6 +245,14 @@ async function createPackageJson() {
   );
 }
 
+function showStarOnGitHubMessage() {
+  console.log(
+    chalk.cyan("Give Hardhat a star on Github if you're enjoying it") +
+      emoji(" 💞✨")
+  );
```
fvictorio, line 252:
```suggestion
  console.log(
    chalk.cyan("Give Hardhat a star on Github if you're enjoying it!") +
      emoji(" 💞✨")
  );
  console.log();
```

fvictorio: The exclamation mark makes it look a bit better in terminals that don't support emoji.

PR #2907, opened 2022-07-08
Use network helpers in the Hardhat Network docs

fvictorio:
Closes HH-915.

### DIFF
--- a/docs/src/content/hardhat-network-helpers/docs/reference.md
+++ b/docs/src/content/hardhat-network-helpers/docs/reference.md
@@ -268,3 +268,25 @@ it("test", async function () {
   // use token and exchanges contracts
 })
 ```
+
+## Other helpers
+
+### `dropTransaction(txHash)`
+
+Drops the given transaction hash from the mempool, if it exists.
+
+Parameters:
+
+- `txHash`: Transaction hash to be dropped from mempool.
+
+Returns: `true` if successful, otherwise `false`.
+
+Throws: if the transaction was already mined.
+
+Example:
+
+```ts
+await helpers.dropTransaction(
+  "0x1010101010101010101010101010101010101010101010101010101010101010"
+);
+```
--- a/docs/src/content/hardhat-network-helpers/docs/reference.md
+++ b/docs/src/content/hardhat-network-helpers/docs/reference.md
@@ -273,11 +273,11 @@ it("test", async function () {
 
 ### `dropTransaction(txHash)`
 
-Drops the given transaction hash from the mempool, if it exists.
+Removes the given transaction from the mempool, if it exists.
 
 Parameters:
 
-- `txHash`: Transaction hash to be dropped from mempool.
+- `txHash`: Transaction hash to be removed from the mempool.
 
 Returns: `true` if successful, otherwise `false`.
 
--- a/docs/src/content/hardhat-network/docs/explanation/mining-modes.md
+++ b/docs/src/content/hardhat-network/docs/explanation/mining-modes.md
@@ -47,11 +47,11 @@ networks: {
 
 ## Removing and replacing transactions
 
-Transactions in the mempool can be removed using the `hardhat_dropTransaction` method:
+Transactions in the mempool can be removed using the [`dropTransaction`](</hardhat-network-helpers/docs/reference#droptransaction(txhash)>) network helper:
 
 ```js
 const txHash = "0xabc...";
-await network.provider.send("hardhat_dropTransaction", [txHash]);
+await helpers.dropTransaction(txHash);
 ```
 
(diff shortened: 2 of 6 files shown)


docs/src/content/hardhat-network-helpers/docs/reference.md
```diff
@@ -268,3 +268,25 @@ it("test", async function () {
   // use token and exchanges contracts
 })
 ```
+
+## Other helpers
+
+### `dropTransaction(txHash)`
+
+Drops the given transaction hash from the mempool, if it exists.
```
alcuadrado, line 276:
```suggestion
Removes the given transaction from the mempool, if it exists.
```

The tx is dropped, not the hash

docs/src/content/hardhat-network-helpers/docs/reference.md
```diff
@@ -268,3 +268,25 @@ it("test", async function () {
...
+
+Parameters:
+
+- `txHash`: Transaction hash to be dropped from mempool.
```
alcuadrado, line 280:
```suggestion
- `txHash`: Transaction hash to be removed from mempool.
```

docs/src/content/hardhat-network-helpers/docs/reference.md
alcuadrado: also, `removed` is easier to understand

docs/src/content/hardhat-network/docs/explanation/mining-modes.md
```diff
@@ -47,11 +47,11 @@ networks: {
 
 ## Removing and replacing transactions
 
-Transactions in the mempool can be removed using the `hardhat_dropTransaction` method:
+Transactions in the mempool can be removed using the [`dropTransaction`](</hardhat-network-helpers/docs/reference#droptransaction(txhash)>) network helper:
 
 ```js
 const txHash = "0xabc...";
-await network.provider.send("hardhat_dropTransaction", [txHash]);
+await helpers.dropTransaction(txHash);
```
alcuadrado, line 54: `### Using RPC methods` still uses the provider.

alcuadrado:
I think `hardhat_impersonateAccount` should be explained using both the helpers and ethers, in the forking guide.

docs/src/content/hardhat-network-helpers/docs/reference.md
fvictorio: Done in d783f719d113df83fd5b3081d7ac0535ba5e43bf

docs/src/content/hardhat-network/docs/explanation/mining-modes.md
fvictorio: Yes, we don't have helpers to change the mining modes yet. [HH-921](https://linear.app/nomic-foundation/issue/HH-921/add-network-helper-to-configure-mining-behavior) tracks that.

fvictorio:
> I think `hardhat_impersonateAccount` should be explained using both the helpers and ethers, in the forking guide.

What do you mean? For the record, this is the only mention to impersonating accounts we have right now in that guide:

image

I guess we could add a section about impersonating accounts before that section, wdyt?

fvictorio:
Oh, you mean using `ethers.getImpersonatedSigner`? That makes sense, I will add that section.

docs/src/content/hardhat-network/docs/guides/forking-other-networks.md
```diff
@@ -75,16 +75,30 @@ networks: {
...
+
+## Customizing Hardhat Network's behavior
 
-- [`hardhat_impersonateAccount`](../reference/#hardhat-impersonateaccount)
-- [`hardhat_stopImpersonatingAccount`](../reference/#hardhat-stopimpersonatingaccount)
-- [`hardhat_setNonce`](../reference/#hardhat-setnonce)
-- [`hardhat_setBalance`](../reference/#hardhat-setbalance)
-- [`hardhat_setCode`](../reference/#hardhat-setcode)
-- [`hardhat_setStorageAt`](../reference/#hardhat-setstorageat)
+Once you've got local instances of mainnet protocols, setting them in the specific state your tests need is likely the next step. For this, you can use our [Hardhat Network Helpers](/hardhat-network-helpers) library, which allows you to do things like manipulating the time of the network or modify the balance of an account.
```
feuGeneA, line 101:
```suggestion
Once you've got a local instance of the mainnet chain state, setting that state to the specific needs of your tests is likely the next step. For this, you can use our [Hardhat Network Helpers](/hardhat-network-helpers) library, which allows you to do things like manipulating the time of the network or modify the balance of an account.
```
The reason for this change was that I didn't like the use of the word `protocols`, but my attempts to change that one word had a sort of ripple effect.

feuGeneA (approved):

PR #2910, opened 2022-07-08
Add `reset` helper

fvictorio:
Add a `reset` helper without any parameters that resets the network to its initial configured state.

I'm nor sure about one aspect of this helper: if you didn't specify a block number in your config, then `helpers.reset()` will reset the network to the current safe block number, which will probably be different to the initial one. In other words, the behavior is just like killing the process and starting it again.

This makes sense conceptually, but at the same time I could see this being problematic, especially because the cache won't be leveraged in the second run.

For the record, this is the same behavior you get if you use `hardhat_reset` with an URL but without a block number.

Also for the record: I tried to implement the alternative behavior (resetting the network to whatever block number was initially used) but it wasn't an easy thing to do.

### DIFF
--- a/docs/src/content/hardhat-network-helpers/docs/reference.md
+++ b/docs/src/content/hardhat-network-helpers/docs/reference.md
@@ -268,3 +268,11 @@ it("test", async function () {
   // use token and exchanges contracts
 })
 ```
+
+## Other helpers
+
+### `reset()`
+
+Resets the Hardhat Network to its initial configured state.
+
+If you want to reset a forked network to a non-forked state, or a non-forked network to a forked state, use the low-level [`hardhat_reset`](/hardhat-network/docs/reference#hardhat_reset) JSON-RPC method instead.
--- a/docs/src/content/hardhat-network/docs/guides/forking-other-networks.md
+++ b/docs/src/content/hardhat-network/docs/guides/forking-other-networks.md
@@ -88,32 +88,15 @@ Once you've got local instances of mainnet protocols, setting them in the specif
 
 ## Resetting the fork
 
-You can manipulate forking during runtime to reset back to a fresh forked state, fork from another block number or disable forking by calling `hardhat_reset`:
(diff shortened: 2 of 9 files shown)


alcuadrado:
This should be considered an advanced method for people who actually want to change the fork block number and/or URL (e.g. testing against mainnet and l2s). We should warn users in our doc, that if they just want to reset the state without changing the forking config snapshots and revert are much better.

This PR needs a way to pass the URL and block number (both optional) to the RPC.

Also consider splitting the method into two: `resetHardhatNetworkWithoutForking();` and `resetHardhatNetworkForking(url:string,blockNumber?:NumberLike)'` — Not convinced of this names.

0xged:
Hey guys! I've taken an interest in `hardhat-network-helper` in general because I had a util to basically facilitate a lot of this (and love the fact that I don't have to maintain it anymore). 

If you let me add from my experience (working with dApps that are deployed in multiple chains, and also have implemented a yuge amount of integrations test): I've never worked with the fork configuration from `hardhat.config`. I think the only case for that might be just to fork of the network, but not for testing. 

Moreover, I do agree with @alcuadrado here that `helpers.reset()` going into the latest state of the network might be misleading. Sounds more that this behavior should be called `helpers.refresh()` or `reload()` as its not resetting the state to anything previously known.

> Also consider splitting the method into two: resetHardhatNetworkWithoutForking(); and resetHardhatNetworkForking(url:string,blockNumber?:NumberLike)' 

I'd definitely go for this behavior. Not sure about the names, but at least it will be a bit more intuitive. 

Sorry for meddling, hopefully my input helps somewhat! 

You both 🪨

fvictorio:
Thank you @0xged, that's extremely useful!

farwayer:
I agree with @0xged about naming. `reset` to concrete block number helper will be very usefull too. Because now we have to use low-level `hardhat_reset`.

fvictorio:
Closing in favor of #3572.

PR #2931, opened 2022-07-12
Proof of concept for proper traces in chai matchers

fvictorio:
The code is a bit awful but keep in mind that, if we decide on doing this, it won't be _much_ better.

Stack traces sometimes look like this, which is ideal:

image

But sometimes look like this, which I couldn't improve and I guess it's acceptable:

image

### DIFF
--- a/packages/hardhat-chai-matchers/src/internal/emit.ts
+++ b/packages/hardhat-chai-matchers/src/internal/emit.ts
@@ -15,6 +15,18 @@ type Provider = providers.Provider;
 
 export const EMIT_CALLED = "emitAssertionCalled";
 
+const buildAssert =
+  (negated: boolean, ssfi: Function) =>
+  (condition: boolean, messageFalse: string, messageTrue?: string) => {
+    if (!negated && !condition) {
+      throw new AssertionError(messageFalse, undefined, ssfi);
+    }
+
+    if (negated && condition && messageTrue !== undefined) {
+      throw new AssertionError(messageTrue, undefined, ssfi);
+    }
+  };
+
 async function waitForPendingTransaction(
   tx: Promise<Transaction> | Transaction | string,
   provider: Provider
@@ -41,41 +53,46 @@ export function supportEmit(
     "emit",
     function (this: any, contract: Contract, eventName: string) {
       const tx = this._obj;
+      const negated = this.__flags.negate;
 
       const promise = this.then === undefined ? Promise.resolve() : this;
 
+      const onSuccess = (receipt: providers.TransactionReceipt) => {
+        const assert = buildAssert(negated, onSuccess);
+
+        let eventFragment: EventFragment | undefined;
+        try {
+          eventFragment = contract.interface.getEvent(eventName);
+        } catch (e) {
+          // ignore error
+        }
+
+        if (eventFragment === undefined) {
+          throw new AssertionError(
+            `Event "${eventName}" doesn't exist in the contract`
+          );
+        }
+
+        const topic = contract.interface.getEventTopic(eventFragment);
+        this.logs = receipt.logs
+          .filter((log) => log.topics.includes(topic))
+          .filter(
+            (log) =>
+              log.address.toLowerCase() === contract.address.toLowerCase()
+          );
+
+        assert(
+          this.logs.length > 0,
+          `Expected event "${eventName}" to be emitted, but it wasn't`,
+          `Expected event "${eventName}" NOT to be emitted, but it was`
+        );
+        chaiUtils.flag(this, "eventName", eventName);
+        chaiUtils.flag(this, "contract", contract);
+      };
+
       const derivedPromise = promise
         .then(() => waitForPendingTransaction(tx, contract.provider))
-        .then((receipt: providers.TransactionReceipt) => {
-          let eventFragment: EventFragment | undefined;
-          try {
-            eventFragment = contract.interface.getEvent(eventName);
-          } catch (e) {
-            // ignore error
-          }
-
-          if (eventFragment === undefined) {
-            throw new AssertionError(
-              `Event "${eventName}" doesn't exist in the contract`
-            );
-          }
-
-          const topic = contract.interface.getEventTopic(eventFragment);
-          this.logs = receipt.logs
-            .filter((log) => log.topics.includes(topic))
-            .filter(
-              (log) =>
-                log.address.toLowerCase() === contract.address.toLowerCase()
-            );
-
-          this.assert(
-            this.logs.length > 0,
-            `Expected event "${eventName}" to be emitted, but it wasn't`,
-            `Expected event "${eventName}" NOT to be emitted, but it was`
-          );
-          chaiUtils.flag(this, "eventName", eventName);
-          chaiUtils.flag(this, "contract", contract);
-        });
+        .then(onSuccess);
 
       chaiUtils.flag(this, EMIT_CALLED, true);
 
@@ -91,14 +108,20 @@ export async function emitWithArgs(
   context: any,
   Assertion: Chai.AssertionStatic,
   chaiUtils: Chai.ChaiUtils,
-  expectedArgs: any[]
+  expectedArgs: any[],
+  ssfi: Function
 ) {
+  const negated = false; // .withArgs cannot be negated
+  const assert = buildAssert(negated, ssfi);
+
   tryAssertArgsArraysEqual(
     context,
     Assertion,
     chaiUtils,
     expectedArgs,
-    context.logs
+    context.logs,
+    assert,
+    ssfi
   );
 }
 
@@ -107,7 +130,9 @@ function assertArgsArraysEqual(
   Assertion: Chai.AssertionStatic,
   chaiUtils: Chai.ChaiUtils,
   expectedArgs: any[],
-  log: any
+  log: any,
+  assert: ReturnType<typeof buildAssert>,
+  ssfi: Function
 ) {
   const { utils } = require("ethers");
 
@@ -115,12 +140,10 @@ function assertArgsArraysEqual(
     chaiUtils.flag(context, "contract").interface as Interface
   ).parseLog(log).args;
   const eventName = chaiUtils.flag(context, "eventName");
-  context.assert(
+  assert(
     actualArgs.length === expectedArgs.length,
     `Expected "${eventName}" event to have ${expectedArgs.length} argument(s), but it has ${actualArgs.length}`,
-    "Do not combine .not. with .withArgs()",
-    expectedArgs.length,
-    actualArgs.length
+    "Do not combine .not. with .withArgs()"
   );
   for (let index = 0; index < expectedArgs.length; index++) {
     if (typeof expectedArgs[index] === "function") {
@@ -128,7 +151,7 @@ function assertArgsArraysEqual(
         index + 1
       )} event argument`;
       try {
-        context.assert(
+        assert(
           expectedArgs[index](actualArgs[index]),
           `${errorPrefix} returned false`
           // no need for a negated message, since we disallow mixing .not. with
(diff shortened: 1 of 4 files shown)


alcuadrado:
How would those stack traces look with the current version?

fvictorio:
Like this:

image

alcuadrado:
I see, thanks!

> The code is a bit awful but keep in mind that, if we decide on doing this, it won't be much better.

So by this, you meant the code, not the result. The result is a significant improvement.

packages/hardhat-chai-matchers/src/internal/reverted/revertedWith.ts
```diff
@@ -1,45 +1,64 @@
+import { AssertionError } from "chai";
 import { decodeReturnData, getReturnDataFromError } from "./utils";
 
+const buildAssert =
```
feuGeneA, line 4: Maybe de-duplicate this by extracting it to some shared module?

feuGeneA (approved):
I see what you mean about the code being awful, but... I'm torn between on the one hand being elated that you've gotten it working with such a "simple" change and on the other hand being repulsed by the convolution of the huge parameter lists and the ssfi thing and the recursion combinator thing. In any case, bravo for the working solution!

packages/hardhat-chai-matchers/src/internal/reverted/revertedWith.ts
fvictorio: Yes! I will move this to another module and add a long-ass comment explaining why we are doing this monstrosity.

fvictorio:
This is now ready for review. There are three important things in this PR to pay attention.

The first one is the comment explaining what `buildAssert` does. Please read it and tell me if you think that explanation will make sense to our six-months-in-the-future selves.

The second one is the stack traces tests I added. I'm not 100% convinced, but I guess it's better than nothing. My main concern is if they will give us false negatives in, for example, the CI. But I guess we can always remove them if they turn out to be a nuisance.

The third one is that the `assert` function returned by `buildAssert` doesn't accept the 4th and 5th arguments accepted by `this.assert` and used by some (not all) matchers. These are the expected values that will be diffed and shown. More concretely, before we had this:

stack-trace-before

And now we have this:

stack-trace-after

So we lost that part of the error message. As far as I understand, this doesn't add more information, it's just another view of the issue. I think this trade-off is fine, but if we decide we do want that to be shown, then I can take a deeper look at `this.assert` and see if we can replicate that behavior somehow.

packages/hardhat-chai-matchers/src/internal/changeTokenBalance.ts
```diff
@@ -57,8 +67,11 @@ export function supportChangeTokenBalance(Assertion: Chai.AssertionStatic) {
       balanceChanges: EthersT.BigNumberish[]
     ) {
       const ethers = require("ethers") as typeof EthersT;
-      let subject = this._obj;
 
+      // see buildAssert's jsdoc
```
feuGeneA, line 71: I'd like to see a one-line explanation here, something like `capture negated flag before async code executes. see buildAssert's jsdoc.`

packages/hardhat-chai-matchers/src/internal/changeTokenBalance.ts
```diff
@@ -76,12 +89,14 @@ export function supportChangeTokenBalance(Assertion: Chai.AssertionStatic) {
...
-        balanceChangesPromise,
-        addressesPromise,
-        getTokenDescription(token),
-      ]).then(([actualChanges, addresses, tokenDescription]) => {
-        this.assert(
+      const checkBalanceChanges = ([
+        actualChanges,
+        addresses,
+        tokenDescription,
+      ]: [EthersT.BigNumber[], string[], string]) => {
```
feuGeneA, line 96: Just out of curiosity: why does this instance of `checkBalanceChanges` have a return type while the others don't?

packages/hardhat-chai-matchers/src/internal/emit.ts
```diff
@@ -40,42 +41,48 @@ export function supportEmit(
   Assertion.addMethod(
     "emit",
     function (this: any, contract: Contract, eventName: string) {
+      // see buildAssert's jsdoc
```
feuGeneA, line 44: I'd like to see a one-line explanation here, something like `capture negated flag before async code executes. see buildAssert's jsdoc.`

packages/hardhat-chai-matchers/src/internal/emit.ts
```diff
@@ -205,20 +230,21 @@ const tryAssertArgsArraysEqual = (
...
   const eventName = chaiUtils.flag(context, "eventName");
-  context.assert(
+  assert(
     false,
     `The specified arguments (${util.inspect(
       expectedArgs
     )}) were not included in any of the ${
       context.logs.length
-    } emitted "${eventName}" events`,
-    "Do not combine .not. with .withArgs()"
```
feuGeneA, line 222: Did we lose this warning message entirely?

packages/hardhat-chai-matchers/src/internal/reverted/reverted.ts
```diff
@@ -1,15 +1,21 @@
+import { buildAssert } from "../../utils";
 import { decodeReturnData, getReturnDataFromError } from "./utils";
 
 export function supportReverted(Assertion: Chai.AssertionStatic) {
   Assertion.addProperty("reverted", function (this: any) {
+    // see buildAssert's jsdoc
+    const negated = this.__flags.negate;
```
feuGeneA, line 7: I'd like to see a one-line explanation here, something like `capture negated flag before async code executes. see buildAssert's jsdoc.`

packages/hardhat-chai-matchers/src/internal/reverted/revertedWith.ts
```diff
@@ -1,45 +1,53 @@
+import { buildAssert } from "../../utils";
 import { decodeReturnData, getReturnDataFromError } from "./utils";
 
 export function supportRevertedWith(Assertion: Chai.AssertionStatic) {
   Assertion.addMethod(
     "revertedWith",
     function (this: any, expectedReason: unknown) {
+      // see buildAssert's jsdoc
```
feuGeneA, line 8: I'd like to see a one-line explanation here, something like `capture negated flag before async code executes. see buildAssert's jsdoc.`

packages/hardhat-chai-matchers/src/internal/reverted/revertedWithPanic.ts
```diff
@@ -11,6 +12,9 @@ export function supportRevertedWithPanic(Assertion: Chai.AssertionStatic) {
     function (this: any, expectedCodeArg: any) {
       const ethers = require("ethers");
 
+      // see buildAssert's jsdoc
```
feuGeneA, line 15: I'd like to see a one-line explanation here, something like `capture negated flag before async code executes. see buildAssert's jsdoc.`

packages/hardhat-chai-matchers/src/internal/reverted/revertedWithoutReason.ts
```diff
@@ -1,38 +1,46 @@
+import { buildAssert } from "../../utils";
 import { decodeReturnData, getReturnDataFromError } from "./utils";
 
 export function supportRevertedWithoutReason(Assertion: Chai.AssertionStatic) {
   Assertion.addMethod("revertedWithoutReason", function (this: any) {
+    // see buildAssert's jsdoc
```
feuGeneA, line 6: I'd like to see a one-line explanation here, something like `capture negated flag before async code executes. see buildAssert's jsdoc.`

packages/hardhat-chai-matchers/src/internal/reverted/revertedWithoutReason.ts
```diff
@@ -1,38 +1,46 @@
...
+        assert(
           false,
           `Expected transaction to be reverted without a reason, but it reverted with reason '${decodedReturnData.reason}'`
         );
       } else if (decodedReturnData.kind === "Empty") {
-        this.assert(
+        assert(
           true,
-          null,
+          undefined,
```
feuGeneA, line 32: Just out of curiosity, why did these `null`s need to change to `undefined`s?

packages/hardhat-chai-matchers/test/changeEtherBalance.ts
```diff
@@ -579,5 +580,25 @@ describe("INTEGRATION: changeEtherBalance matcher", function () {
...
+      it("includes test file", async function () {
+        try {
+          await expect(() =>
+            sender.sendTransaction({
+              to: receiver.address,
+              value: 200,
+            })
+          ).to.changeEtherBalance(sender, -100);
+        } catch (e: any) {
+          expect(util.inspect(e)).to.include("changeEtherBalance.ts");
```
feuGeneA, line 595: Again, perhaps prefix the file name with `/test/`

packages/hardhat-chai-matchers/src/internal/changeEtherBalance.ts
```diff
@@ -14,20 +15,28 @@ export function supportChangeEtherBalance(Assertion: Chai.AssertionStatic) {
       options?: BalanceChangeOptions
     ) {
       const { BigNumber } = require("ethers");
+
+      // see buildAssert's jsdoc
```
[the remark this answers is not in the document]
fvictorio: Done in 34fe9d5b808d6e77693ac21d0bcd85c5f992621d, but I used a semicolon instead of a period just because I once read [this comic](https://theoatmeal.com/comics/semicolon).

packages/hardhat-chai-matchers/src/internal/reverted/revertedWithoutReason.ts
fvictorio: IIRC, chai was expecting null here, but we tend to use `undefined` as a null value, so I did that when I wrote the `assert` helper just for consistency.

packages/hardhat-chai-matchers/src/internal/emit.ts
fvictorio: I removed it because it was unreachable (that's why no test failed after I did that).

Probably a better option would've been to have a separate method like `assertFalse` that just accepts a single argument, but that would've also made `buildAssert` return multiple functions or something like that...

But now that I think about it, I think the best option is to assert in the assert method (yeah, I know) that the message is defined for the condition that matches. That is: throw an error if the condition is true, negated is true, and the message for that combination does not exist. I will do that.

fvictorio: Done in 7c2ebf745f93487c50167fc710dd1d998e21dced, please let me know what you think.

packages/hardhat-chai-matchers/src/internal/changeTokenBalance.ts
fvictorio: I'm not sure I understand. The other `check*` functions have a similar type too. Notice though that this isn't the return type, it's the type of the parameters of the function, and the reason I had to add it is that it can no longer be inferred because it's not used directly as an argument anymore. Does that make sense?

packages/hardhat-chai-matchers/test/changeEtherBalances.ts
```diff
@@ -348,5 +349,25 @@ describe("INTEGRATION: changeEtherBalances matcher", function () {
...
+      it("includes test file", async function () {
+        try {
+          await expect(() =>
+            sender.sendTransaction({
+              to: receiver.address,
+              value: 200,
+            })
+          ).to.changeEtherBalances([sender, receiver], [-100, 100]);
+        } catch (e: any) {
+          expect(util.inspect(e)).to.include("changeEtherBalances.ts");
```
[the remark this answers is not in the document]
fvictorio: That makes a lot of sense, thank you. Done in 0012bcf4fbfcf880e5ffb4e707caf061c0ccf17f

feuGeneA (approved):

alcuadrado (commented):
Late review: LGTM. Great work.

Issue #2924, opened 2022-07-11
Use blockscout when using `customChains` in the etherscan plugin


clemsos:
I am trying to add Celo to my wokflow, so I added this to my hardhat config:

```js
const etherscan = {
  apiKey: {
    celo: 'api-key',
  },
  customChains: [
    {
      network: 'celo',
      chainId: 42220,
      urls: {
        apiURL: 'https://explorer.celo.org/api',
        browserURL: 'https://explorer.celo.org/',
      },
    },
  ],
}
```

Then when verifying a contract, I got the following error

```
Error in plugin @nomiclabs/hardhat-etherscan: Failed to send contract verification request.
Endpoint URL: https://explorer.celo.org/api
The HTTP server response is not ok. Status code: 400 Response text: {"message":"Unknown action","result":null,"status":"0"}
```

The [Celo explorer is using blockscout](https://explorer.celo.org/), which is already impemented in the plugin so I was wondering if there will be a way to specify it when adding custom chains.

Many thanks !

fvictorio:
Hi @clemsos. Blockscout has a etherscan-compatible API for verification, so this looks like it should work.

Maybe you can open an issue on [their explorer's repo](https://github.com/celo-org/blockscout)?

clemsos:
thank you for the answer, I'll try that !

clemsos:
> Hey checkout the PR #2933

thanks @therealharpaljadeja that seems to work. Actually shame on me... I was using some outdated urls for the api and browser.

Issue #2932, opened 2022-07-12
Add Celo and Alfajores in networks available to verify contracts via `hardhat verify`


therealharpaljadeja:
## Description of the issue
At [CELO](https://github.com/celo-org), we are looking to easily + automatically verify contracts on the Celo and Alfajores network.

Verifying contracts via the RPC endpoint of the Celo Blockscout instance is quite complicated. We currently rely on manual verification by flattening Solidity files.

It would be great if we could add the Celo and Alfajores network to the list of networks available on the [hardhat-etherscan](https://github.com/NomicFoundation/hardhat/tree/master/packages/hardhat-etherscan) package.

This would help us (the team) tremendously and would provide a better developer experience for people building on Celo. Verifying contracts on Celo and Alfajores from any hardhat project would be as simple as running the command

`npx hardhat verify --network celo`

It would then be more developer/user friendly to interact with them via the Celoscan explorer using the contract methods fields.

## Proposed solution
Add the Celo and Alfajores network to the list of networks under the @nomiclabs/hardhat-etherscan package.

#2933

ericnakagawa:
Would love to see this PR accepted. It is the top feature request from the many hackathons we have supported this year.

Improving the verification process for Celo devs helps increase quality and security of our community!

tab00:
I get error "Network alfajores doesn't exist". What's the status of this issue?

fvictorio:
Closing this since in general we are not adding new networks to `hardhat-verify`.

Adding support for a new network is simple, check [here](https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify#adding-support-for-other-networks). This can then be easily packaged in a `hardhat-celo` plugin. Happy to answer questions/light review that if someone implements it.

PR #2933, opened 2022-07-12
Add support for Celoscan

therealharpaljadeja:
--> #2932 

---

## What does this PR introduce?

Add support for Celo and Alfajores network, by adding the Celo and Alfajores chain specifications under`ChainConfig.ts`.

This is to enable to verify contracts deployed on Celo via the command:

`npx hardhat verify --network celo`

### DIFF
--- a/packages/hardhat-etherscan/src/ChainConfig.ts
+++ b/packages/hardhat-etherscan/src/ChainConfig.ts
@@ -205,4 +205,18 @@ export const chainConfig: ChainConfig = {
       browserURL: "https://explorer.pops.one",
     },
   },
+  alfajores: {
+    chainId: 44787,
+    urls: {
+      apiURL: "https://api-alfajores.celoscan.io/api",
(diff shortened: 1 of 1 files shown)


therealharpaljadeja:
Fixes #2924

irzhywau:
@therealharpaljadeja @clemsos Im a bit confused here. how adding the network info have fixed the issue here?

I'm actually experiening the same issue:
```
The HTTP server response is not ok. Status code: 400 Response text: {"message":"Unknown action","result":null,"status":"0"}
```

The network is using bloclscout also, and API endpoint is https://esc-testnet.elastos.io/api

Would be great to have some insights

fvictorio:
In general we are not adding new networks to `hardhat-verify`. Users can do it manually (check [this](https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify#adding-support-for-other-networks) to learn how) and that can be easily packaged in a community plugin.

Issue #2546, opened 2022-03-29
Manually rethrowing reverts impedes custom error decoding


frangio:
We sometimes use a [`verifyCallResult`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/02fcc75bb7f35376c22def91b0fb9bc7a50b9458/contracts/utils/Address.sol#L201-L205) helper to rethrow the returndata if a low level call results in an error:

```solidity
        (bool success, bytes memory returndata) = target.call(data);
        return verifyCallResult(success, returndata, errorMessage);
```

With compiler optimizations turned on, if the returndata contains a custom error, Hardhat is not able to decode the rethrown custom error, presumably because it can't identify where the error actually originates because rethrowing is not the standard solc subroutine.

Things work fine with optimizations turned off.

frangio:
After further testing I believe this isn't specific to manual rethrowing, even for built-in rethrowing optimizations cause custom error decoding to fail. (Sorry I'm not providing reproduction steps.)

alcuadrado:
@fvictorio I'm assigning this issue to you, as it seems important, and you are more familiar with this part of the code.

frangio:
Here's a reproduction that shows the source of an error to be wrongly identified due to manually rethrowing. I haven't included this in the reproduction but if custom errors were involved I can imagine how decoding would be broken.

https://github.com/frangio/hardhat-rethrow-reproduction

frangio:
Still relevant.

fvictorio:
Fixed in Hardhat v2.10.1.

PR #2935, opened 2022-07-12
Loosen HH Network's error propagation heuristic

alcuadrado:
As shown in #2546, Hardhat Network can be too restrictive when deciding if an error is being propagated or not. In particular, it never considers asm-based manual propagation as actual propagation.

This PR loosens our heuristic significantly. If the previous heuristic fails, but the error of the trace and subtrace are equal and not empty, we still consider it to propagate.

There's a minor chance of this resulting in false positives, but I think it's reasonably small.

Another alternative is checking if the `return` opcode is inside of an inline assembly block, but to do that in a performant way we'd need to make a larger change.

### DIFF
--- a/.changeset/fifty-jeans-double.md
+++ b/.changeset/fifty-jeans-double.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Better Solidity errors propagation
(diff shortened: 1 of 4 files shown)


fvictorio (approved):
LGTM. I only have two questions out of curiosity:

1) Do you know why this was only failing when the optimizer was enabled? It looks like it should've failed without optimizations too.
2) In which scenario does this produce a false positive? The only one I can think of is if you manually throw the same error, but without using the data returned by `call`, which would be silly (and you could argue that you are still propagating the error in a roundabout way).

alcuadrado:
For the record: we discussed those things offline.

1. It's not exclusive of optimized code.
2. Also if you call a contract, it throws an error, and then you happen to revert with the same value.

PR #2957, opened 2022-07-17
Fix sample test

andrey-kuprianov:
The sample test doesn't compile due to import errors. Fixed it to use correct imports.

---


fvictorio:
Thanks a lot for this @andrey-kuprianov. I made some small changes to fix the same problem in the javascript snippet, and to match the code with the paragraph that follows.

PR #2959, opened 2022-07-17
hardhat-etherscan support acala & karura & mandala

zjb0807:
hardhat-etherscan support acala & karura & mandala

### DIFF
--- a/packages/hardhat-etherscan/src/ChainConfig.ts
+++ b/packages/hardhat-etherscan/src/ChainConfig.ts
@@ -177,6 +177,41 @@ export const chainConfig: ChainConfig = {
       browserURL: "https://blockscout.com/poa/sokol",
     },
   },
+  mandala: {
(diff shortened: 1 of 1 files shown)


alcuadrado:
Hey @zjb0807,

Thanks for sending this PR. Unfortunately, we aren't merging support for new chains into this plugin, as that means we have to support their integration.

Instead, we created a way for users and plugins to add support themselves. Look at [this seciton](https://github.com/NomicFoundation/hardhat/tree/master/packages/hardhat-etherscan#adding-support-for-other-networks) of its readme.

For example, you can use that config field to create a `hardhat-etherscan-acala` plugin, that uses `extendConfig` to add support for `acala`.

zjb0807:
@alcuadrado 
I use `customChains` to test it locally. The blockchain browser fork from the blockscout, so I think it's fine.

PR #2962, opened 2022-07-19
Use typed version of tsort

ghost:

---

Use the typescript ported version of [tsort](https://github.com/eknkc/tsort) library https://github.com/ayanamitech/tsort-ts, which does the exact same behavior as before.


alcuadrado:
Hey @ayanamidev! Thanks for sending this PR!

I see that you converted `tsort` to TS :) We are currently looking to decrease the number of dependencies that we use, and this can be a good opportunity to do it. What do you think about sending an alternative pr that vendors/imports the sources here instead?

alcuadrado:
Awesome! Thanks 🙌

PR #2964, opened 2022-07-19
Make console.log abi-compliant

fvictorio:
Also closes HH-663, which means that `console.log()` prints an empty line instead of printing "undefined".


fvictorio:
For reference, here's the result of doing different kinds of `console.log` calls between versions of Hardhat.

Notice that an old `console.sol` work fine with the new change, but a new `console.sol` doesn't work with an old hardhat network. So we have backwards-compatibility but not forwards-compatibility. Or is the other way around? Anyway.

### Before this change

```
##########
console.log()
undefined
##########

##########
console.log(3)
3
##########

##########
console.logInt(3)
3
##########

##########
console.logUint(3)
3
##########

##########
console.log(true, 3)
true 3
##########

##########
console.log(3, true)
3 true
##########
```

### After this change

```
##########
console.log()

##########

##########
console.log(3)
3
##########

##########
console.logInt(3)
3
##########

##########
console.logUint(3)
3
##########

##########
console.log(true, 3)
true 3
##########

##########
console.log(3, true)
3 true
##########
```

### Before this change, using a node after this change

```
    ##########
    console.log()
    
    ##########
    
    ##########
    console.log(3)
    3
    ##########
    
    ##########
    console.logInt(3)
    3
    ##########
    
    ##########
    console.logUint(3)
    3
    ##########
    
    ##########
    console.log(true, 3)
    true 3
    ##########
    
    ##########
    console.log(3, true)
    3 true
    ##########
```

### After this change, using a node before this change

```
    ##########
    console.log()
    undefined
    ##########
    
    ##########
    console.log(3)
    ##########
    
    ##########
    console.logInt(3)
    ##########
    
    ##########
    console.logUint(3)
    ##########
    
    ##########
    console.log(true, 3)
    ##########
    
    ##########
    console.log(3, true)
    ##########
```

alcuadrado (approved):
LGTM

PR #2973, opened 2022-07-21
Show selector in "unrecognized custom error" msg

### LINKED ISSUE #2948: Blind "reverted with an unrecognized custom error"
Please add hex representation of "unrecognized custom error". Sometimes revert comes from external smart contract, we have no source code. Especially when we fork mainnet.

feuGeneA:
Fixes #2948

### DIFF
--- a/.changeset/popular-fishes-doubt.md
+++ b/.changeset/popular-fishes-doubt.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+The selector is now shown in the "Unrecognized custom error" message
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/return-data.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/return-data.ts
@@ -64,4 +64,8 @@ export class ReturnData {
 
     return errorCode;
   }
+
+  public getSelector(): string | undefined {
+    return this._selector;
+  }
 }
--- a/packages/hardhat-core/src/internal/hardhat-network/stack-traces/error-inferrer.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/stack-traces/error-inferrer.ts
@@ -524,6 +524,11 @@ export class ErrorInferrer {
 
     let errorMessage = "reverted with an unrecognized custom error";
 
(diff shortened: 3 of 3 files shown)


fvictorio:
Ok, this one was hard to test. I wonder how users run into this, because it's not easy to trigger.

Anyway, this is what I did. I deployed and called `f` on this contract:

```solidity
contract Foo {
  function f() public {
    assembly {
      mstore(0x80, 0x1234567800000000000000000000000000000000000000000000000000000000)
      revert(0x80, 4)
    }
  }
}
```

And got this error message:

```
Error: VM Exception while processing transaction: reverted with an unrecognized custom error with selector 12345678
```

fvictorio (approved):

k06a:
@fvictorio it's easy to reproduce if you have tests with mainnet fork.

fvictorio:
@k06a I will make a release that includes this change tomorrow :slightly_smiling_face:

fvictorio:
@k06a we just released [Hardhat v2.20.2](https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.10.2) which includes this change :tada:

kevincheng96:
This is a great feature! But somehow, I'm not able to hit the specific codepath in `error-inferrer.ts` on my end. Instead, the error I'm getting is from [`solidity-error.ts`](https://github.com/NomicFoundation/hardhat/blob/5a9927dd2ecbb6bbe60b35972c5969981cc080f3/packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-errors.ts#L274) instead, which does not include the custom error selector. Curious why there are two error paths and what's the difference between them. Also, should we also add the custom error selector to the error in `solidity-error.ts`?

fvictorio:
@kevincheng96 do you have reproduction steps? If so, can you please open a new issue about this?

kevincheng96:
> @kevincheng96 do you have reproduction steps? If so, can you please open a new issue about this?

@fvictorio sorry for the delay! Sure, I just opened a [new issue](https://github.com/NomicFoundation/hardhat/issues/3237) for this and it links out to some reproduction steps.

PR #2981, opened 2022-07-21
Support for nested loadFixture()

### LINKED ISSUE #2980: Nested calls to loadFixture() do not always work
When passing a function to `loadFixture()` that itself calls `loadFixture()`, this sometimes fails with an error message saying that loaded fixtures are not supported. This happens especially often if nesting more than once.

```
     FixtureSnapshotError: There was an error reverting the snapshot of the fixture.

This might be caused by using nested loadFixture calls in a test, for example by using multiple beforeEach calls. This is not supported yet.
      at loadFixture (node_modules/@nomicfoundation/hardhat-network-helpers/src/loadFixture.ts:47:15)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async recoverViaDistributor (test/Wallet2.ts:111:9)
```

m1cm1c:

This is a mix between a feature implementation and a bug fix. Confer issue  #2980.

Fixes  #2980.

### DIFF
--- a/packages/hardhat-network-helpers/src/loadFixture.ts
+++ b/packages/hardhat-network-helpers/src/loadFixture.ts
@@ -24,9 +24,6 @@ const snapshots: Array<Snapshot<any>> = [];
  * executing the function again, the blockchain will be restored to that
  * snapshot.
  *
- * _Warning_: don't use `loadFixture` with an anonymous function, otherwise the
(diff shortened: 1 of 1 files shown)


fvictorio:
Thanks for this @m1cm1c, but I think it's better to just document this behavior better, as discussed in #2980.

Issue #2648, opened 2022-04-27
add "files" positional arguments to compile task?


wuya666:
Unlike other built-in tasks like `test` and `flatten`, the `compile` task does not accept files as positional arguments to selectively compile certain solidity source files, which is very annoying for my current collaborative project where other members may leave some source files in non-compilable state. I tried to make a custom task to allow selective compilation, thinking it would be just a couple minutes' work, but end up spending hours on it as I found out the `TASK_COMPILE_SOLIDITY` subtask is kinda one big workflow that requires a lot of other internal functions not modularized into subtasks. Anyway I ended up with something like this to get the job done:

[https://gist.github.com/wuya666/024893e4976246182fa29302a65ac005](https://gist.github.com/wuya666/024893e4976246182fa29302a65ac005)

However itʻs something quite ugly IMO so I forked the repo and tried to modify the original `compile.ts` source file, and indeed itʻs quite simple and easy to add the "files" positional arguments with just less than a dozen lines of code:

[https://github.com/wuya666/hardhat/blob/master/packages/hardhat-core/src/builtin-tasks/compile.ts](https://github.com/wuya666/hardhat/blob/master/packages/hardhat-core/src/builtin-tasks/compile.ts)

Now Iʻm considering if I should create a PR for this change, but Iʻm wondering if thereʻs some good reason behind the decision to omit this kind of "files" arguments for the `compile` task? It has made working on the collaborative project much easier for me, but maybe thereʻs some underlying problem with this kind of "files" arguments to selectively compile specific source files that Hardhat has intentionally chosen to omit them?

And if itʻs a bad thing to have the built-in `compile` task supporting selective compilation, maybe the `TASK_COMPILE_SOLIDITY` subtask could be further modularized so it would be easier to write oneʻs own custom task to support such functionalities when thereʻs such a need?

fvictorio:
Hey @wuya666, thanks for opening this issue. I'm not sure about what's the best option here, but I'll discuss it with the team and get back to you.

fvictorio:
Hi @wuya666, sorry for not getting back to you sooner. I think the way to go is to override a subtask. I added an example [here](https://github.com/fvictorio/hardhat-examples/tree/master/ignore-solidity-files) showing how you can do that.

[...]

PR #2988, opened 2022-07-24
Add `--file` option in `compile` to compile specific script

graykode:
### Motivation
- When executing `hardhat compile`, a dependency graph is created based on the import of all solidity files in hardhat-core and compiled through solc.
- However, if there are a lot of solidity files like [contract-wizard](https://github.com/OpenZeppelin/contracts-wizard), even if only one file is modified, all files are compiled, so it takes a very long time to compile.
- In the current PR, an option called `--file` is added to the compile option and only specific files can be compiled through hardhat:
   ```shell
      npx hardhat compile --file contracts/path/path2/test.sol
   ```
- This is a huge speedup if you have a lot of contract files. For my code, I reduced the compile time by about 30x.

### Features
- [X] Add `--file` option in `compile` to compile specific script

### DIFF
--- a/packages/hardhat-core/src/builtin-tasks/compile.ts
+++ b/packages/hardhat-core/src/builtin-tasks/compile.ts
@@ -1322,7 +1322,7 @@ subtask(TASK_COMPILE_SOLIDITY)
           sourceNames = [file];
         } else {
           throw new HardhatError(ERRORS.RESOLVER.FILE_NOT_FOUND, {
-            file: file,
+            file,
           });
         }
       }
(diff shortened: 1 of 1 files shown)


fvictorio:
Hey @graykode, we talked a bit about this in https://github.com/NomicFoundation/hardhat/issues/2648. This is not something we want to add yet because there are some things we need to figure out. On the interface side, this might be better as a variadic argument, although I'm not really sure. More importantly, on the implementation side, we are not sure how this will interact with things like the build info files and the cache.

Notice that this can be done kind of easily without adding it to core. You can check one example [here](https://github.com/fvictorio/hardhat-examples/tree/master/ignore-solidity-files); that one is more about ignoring files, but I think it illustrates the core idea.

fvictorio:
Closing this for now.

PR #2998, opened 2022-07-28
Add Fig as an autocomplete tip

brendanfalk:

---

[Fig](https://fig.io) provides autocomplete for 300+ CLI tools including hardhat. It looks like this:

image

We think this will help users become more efficient with hardhat CLI, so we would love to have Fig listed as an autocomplete method in your docs. Thanks so much and please let me know if you have any questions!


alcuadrado:
Thanks for submitting this PR, @brendanfalk.

While I do like fig, and it's awesome that it works out of the box with hh, the changes introduced by this pr feel like a Fig ad and we aren't open to merge them.

Hope you understand.

PR #3010, opened 2022-08-03
Update peer dependency to fix `hardhat/common` dep

wchargin:
The `@nomicfoundation/hardhat-network-helpers` module requires the
`hardhat/common` module at runtime. This was added to `hardhat` in
commit 6f5f2f8cd2bd ("Include common directory in the published files"),
which was first released in `hardhat@2.9.5`. The peer dependency should
reflect this so that adding `hardhat-network-helpers` to a project with
(say) `hardhat@2.9.3` doesn't silently produce a broken installation.

wchargin-branch: hardhat-common-peer-dep

### DIFF
--- a/.changeset/big-items-scream.md
+++ b/.changeset/big-items-scream.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-network-helpers": patch
+---
+
+Fix peer dependency on `hardhat` to pull in `hardhat/common`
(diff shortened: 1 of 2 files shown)


fvictorio (approved):

fvictorio:
Thank you @wchargin, you are absolutely right, using `^2.0.0` as the peer dependency was a mistake.

PR #3017, opened 2022-08-05
feat: Support .cts extension with ts-node 10+

phated:
- [x] I talked about the limitations in the ESM PR and suggestion in the ts-node transpile PR

---

Due to the way hardhat registers tasks, v3 can't support ESM (as I outlined at https://github.com/NomicFoundation/hardhat/pull/1994#issuecomment-972459102); however, TypeScript added ESM support and many projects are converting to ESM-only. The ts-node team added support for the `.cts` extension in https://github.com/TypeStrong/ts-node/pull/1694, which allows individual files to be resolved as commonjs (see https://github.com/TypeStrong/ts-node#module-type-overrides).

This change should allow a monorepo to become `"type": "module"` and still use hardhat in commonjs mode. 🎉

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "@changesets/cli": "^2.16.0",
     "prettier": "2.4.1",
     "shelljs": "^0.8.5",
-    "typescript": "~4.5.2",
+    "typescript": "~4.7.4",
     "wsrun": "^5.2.2"
   },
   "scripts": {
--- a/packages/e2e/package.json
+++ b/packages/e2e/package.json
@@ -36,7 +36,7 @@
     "prettier": "2.4.1",
     "rimraf": "^3.0.2",
     "shelljs": "^0.8.5",
-    "ts-node": "^8.1.0",
-    "typescript": "~4.5.2"
+    "ts-node": "^10.8.0",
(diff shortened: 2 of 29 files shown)


fvictorio:
@alcuadrado this looks good to me. Do you want to take a look?

fvictorio:
Hmm, this might need some changes to the Hardhat config loading logic.

robwilkes:
I tried this and it doesn't seem to be detecting my .cts config file. and if I manually pass `--config hardhat.config.cts` I get mocha errors when running `hardhat test`. many modules are starting to move to ESM only, surely the best to way enable support for those is to have the hardhat project marked as ESM but hardhat support/load .cts files. surely this wouldn't take much to get over the line, it feels part way there already.

phated:
@robwilkes this is the very first, minor step into getting some semblance of interopt with `type: module` projects. Please hang tight until we can get things working together. As for this feature specifically, it requires ts-node 10+ and typescript 4.7+ otherwise it won't detect the extension correctly.

fvictorio:
Merging this to the `esm` feature branch, because I think these PRs will be easier to review and test together.

PR #3026, opened 2022-08-09
Support Solidity 0.8.16

feuGeneA:


feuGeneA:
The CI failure is the problem with running the "full block" test against Kovan, and it's also failing in master, so shouldn't be considered an impediment to merging this PR.

fvictorio (approved):

PR #3030, opened 2022-08-10
Migrate Kovan full-block test to Goerli

feuGeneA:


packages/hardhat-core/src/internal/core/config/default-config.ts
```diff
@@ -99,6 +99,16 @@ export const defaultHardhatNetworkParams: Omit<
         ]),
       },
     ],
+    [
+      5, // goerli
+      {
```
fvictorio, line 104:
```suggestion
      {
        // taken from https://github.com/eth-clients/goerli/
```

fvictorio: (I don't know if you took the block numbers from there, but it's what I used to double-check :sweat_smile:)

fvictorio (approved):
Minor comment, pre-approving.

packages/hardhat-core/src/internal/core/config/default-config.ts
feuGeneA: fyi the source is listed at the beginning of that list of hardfork histories for different networks: https://github.com/NomicFoundation/hardhat/pull/3030/files#diff-12aa9a96d746a7b79bc952eea2f359a0b5c12c87f5215689936e26ececde6d1bR61

fvictorio: Oh right, my bad!

PR #3033, opened 2022-08-11
Change stack traces tests to only run against the few most recent Solidity versions, and relegate the older ones to a CI cron job

feuGeneA:
for https://linear.app/nomic-foundation/issue/HH-940/change-stack-traces-tests-to-only-run-against-the-few-most-recent

- [x] CI: only run stack trace tests on latest solc's
- [x] relegate the tests of older solc versions to a scheduled/periodic CI run


.github/workflows/hardhat-network-tracing-scheduled.yml
```diff
@@ -0,0 +1,44 @@
...
+        with:
+          path: |
+            packages/hardhat-core/test/internal/hardhat-network/stack-traces/test-files/**/artifacts
+          key: hardhat-network-stack-traces-tests-${{ hashFiles('yarn.lock') }}-${{ hashFiles('packages/hardhat-core/test/internal/hardhat-network/stack-traces/test-files/**/*.sol') }}-${{ hashFiles('packages/hardhat-core/test/internal/hardhat-network/stack-traces/test-files/**/test.json') }}-${{ hashFiles('packages/hardhat-core/test/internal/hardhat-network/stack-traces/**/*.ts') }}
+      - name: Run tests
+        env:
+          DO_NOT_SET_THIS_ENV_VAR____IS_HARDHAT_CI: true
+          FORCE_COLOR: 3
+          NODE_OPTIONS: "--max-old-space-size=4096"
+        run: yarn test:tracing
```
feuGeneA, line 44: First I started reading ["Reusing workflows"](https://docs.github.com/en/actions/using-workflows/reusing-workflows), but I quickly got annoyed with how much content there was to wade through, and I realized that I could just hack this into place by basically duplicating the existing workflow file and making a few small tweaks, so that's what i did :sweat_smile:

.github/workflows/hardhat-network-tracing-scheduled.yml
```diff
@@ -0,0 +1,44 @@
+name: Hardhat Network Tracing Capabilities, scheduled runs for all solc verisons
+
+on:
+  schedule:
+    - cron: "0 0 * * *"
```
fvictorio, line 5: This is midnight, right? Let's add a comment here, just because cron syntax always gives me pause.

.github/workflows/hardhat-network-tracing-scheduled.yml
```diff
@@ -0,0 +1,44 @@
+name: Hardhat Network Tracing Capabilities, scheduled runs for all solc verisons
```
fvictorio, line 1: Nitpick: I would name use a more descriptive filename like `hardhat-network-tracing-all-solc-versions`

fvictorio (approved):

PR #2985, opened 2022-07-23
(WIP) Upgrade ethereumjs

fvictorio:
Opening this to have a full CI run. There are still several pending things.

PR #3034, opened 2022-08-12
Ugprade EthereumJS

fvictorio:
These are (almost) the same changes as in #2985, but split into commits in a way that hopefully should make it easier to review.


packages/hardhat-core/src/types/config.ts
```diff
@@ -136,15 +134,15 @@ export type HardhatNetworkChainsConfig = Map<
 export interface HardhatNetworkConfig {
   chainId: number;
   from?: string;
-  gas: "auto" | number;
-  gasPrice: "auto" | number;
+  gas: "auto" | bigint;
```
alcuadrado, line 137: Changing `number` for `bigint` is a breaking change.

packages/hardhat-core/src/types/config.ts
```diff
@@ -136,15 +134,15 @@ export type HardhatNetworkChainsConfig = Map<
...
   gasMultiplier: number;
-  initialBaseFeePerGas?: number;
+  initialBaseFeePerGas?: bigint;
   hardfork: string;
   mining: HardhatNetworkMiningConfig;
   accounts: HardhatNetworkAccountsConfig;
-  blockGasLimit: number;
-  minGasPrice: BN;
+  blockGasLimit: bigint;
+  minGasPrice: bigint;
```
alcuadrado, line 145: I think this can be an exception, as we really want to remove `BN`, and this field is not super common

packages/hardhat-core/src/types/config.ts
alcuadrado: Note that reverting this is a pretty large change :(

packages/hardhat-core/src/internal/util/bigint.ts
```diff
@@ -0,0 +1,84 @@
+import {
+  BigIntLike,
+  bufferToBigInt,
+  bufferToHex,
+  toBuffer,
+} from "@ethereumjs/util";
```
alcuadrado, line 6: I'm pretty sure that having this import at the top-level means that we are gonna end up loading ethereumjs/util while booting hardhat, which is slow.

Also, I think we should avoid using ethereumjs here.

packages/hardhat-core/src/internal/util/bigint.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const exhaustiveCheck: never = x;
+  return exhaustiveCheck;
+}
+
+function toHex(x: number | bigint): string {
+  return bufferToHex(toBuffer(x));
+}
+
+export const BigIntUtils = {
```
alcuadrado, line 73: Why this instead of just exporting everything individually?

packages/hardhat-core/src/internal/util/bigint.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  return result;
+}
+
+function cmp(a: bigint, b: bigint): number {
+  return a < b ? -1 : a > b ? 1 : 0;
+}
+
+function mapNumberToBigInt<T>(x: number | T): bigint | T {
```
alcuadrado, line 34: This is a surprising function. Same for the following ones. Maybe the name is surprising?

packages/hardhat-core/src/internal/util/bigint.ts
```diff
@@ -0,0 +1,84 @@
...
+  if (typeof x === "number") {
+    return BigInt(x);
```
alcuadrado, line 36: Shouldn't we validate this?

packages/hardhat-core/src/internal/util/bigint.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  return x;
+}
+
+function toWord(x: bigint | number): string {
+  return x.toString(16).padStart(64, "0");
+}
+
+function fromBigIntLike(x: BigIntLike): bigint {
```
alcuadrado, line 54: I feel uneasy about using an ethereumjs type here.

packages/hardhat-core/src/internal/util/bigint.ts
alcuadrado: I can't find tests for this module. That makes me think that it's even more important to validate this, as it's unclear what happens if there's an overflow.

alcuadrado: PS: These comments were for the other function, the one turning bigints into numbers 😂

packages/hardhat-core/src/internal/util/wei-values.ts
```diff
@@ -10,38 +8,38 @@ import { BN } from "ethereumjs-util";
...
+    wei = BigInt(wei);
   }
 
-  if (wei.eqn(0)) {
+  if (wei === 0n) {
     return "0 ETH";
   }
 
-  if (wei.lt(new BN(10).pow(new BN(5)))) {
+  if (wei < 10_0000n) {
```
alcuadrado, line 20:
```suggestion
  if (wei < 100_000n) {
```

packages/hardhat-core/src/internal/core/jsonrpc/types/input/blockTag.ts
```diff
@@ -1,9 +1,5 @@
 import * as t from "io-ts";
 
-// TS2742 workaround
```
alcuadrado, line 3: YESSSSSS

packages/hardhat-core/src/internal/core/providers/accounts.ts
```diff
@@ -225,47 +224,37 @@ export class LocalAccountsProvider extends ProviderWrapperWithChainId {
...
-        ? new Common({ chain: chainId, hardfork: "london" })
-        : Common.forCustomChain(
-            "mainnet",
-            {
-              chainId,
-              networkId: chainId,
-            },
-            "london"
-          );
+    const common = Common.custom({ chainId, networkId: chainId });
```
alcuadrado, line 257: This is so weird, but I was present in the discord conversation that motivated it, so ok

packages/hardhat-core/src/internal/core/providers/construction.ts
```diff
@@ -72,7 +72,9 @@ export function createProvider(
     ) {
       forkConfig = {
         jsonRpcUrl: hardhatNetConfig.forking?.url,
-        blockNumber: hardhatNetConfig.forking?.blockNumber,
+        blockNumber: BigIntUtils.mapBigIntToNumber(
```
alcuadrado, line 75: I think this function is used exclusively with `| undefined`. Maybe we can make it more specific?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/debug.ts
```diff
@@ -159,7 +159,7 @@ describe("Debug module", function () {
     });
   });
 
-  describe("fork tests", function () {
+  describe("fork tests (pre-berlin)", function () {
```
alcuadrado, line 162: pre berlin or pre london?

packages/hardhat-core/test/internal/hardhat-network/provider/utils/assertEqualTraces.ts
```diff
@@ -3,8 +3,8 @@ import { assert } from "chai";
 import { RpcDebugTraceOutput } from "../../../../../src/internal/hardhat-network/provider/output";
 
 export function assertEqualTraces(
-  expected: RpcDebugTraceOutput,
-  actual: RpcDebugTraceOutput
+  actual: RpcDebugTraceOutput,
```
alcuadrado, line 6: Pretty hard to review this, so I'm asking. Do you also updated the uses of this?

packages/hardhat-core/test/internal/hardhat-network/provider/modules/web3.ts
```diff
@@ -28,7 +29,9 @@ describe("Web3 module", function () {
       describe("web3_sha3", async function () {
         it("Should return the keccak256 of the input", async function () {
           const data = "0x123a1b238123";
-          const hashed = bufferToRpcData(keccak256(toBuffer(data)));
+          const hashed = bufferToRpcData(
+            arrToBufArr(keccak256(bufArrToArr(toBuffer(data))))
```
alcuadrado, line 33: I'm pretty sure this is not needed. `Buffer`s are `Uint8Array`s.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/web3.ts
```diff
@@ -45,6 +46,6 @@ export class Web3Module {
   }
 
   private async _sha3Action(buffer: Buffer): Promise<string> {
-    return bufferToRpcData(keccak256(buffer));
+    return bufferToRpcData(arrToBufArr(keccak256(bufArrToArr(buffer))));
```
alcuadrado, line 49: Maybe we just need a `uint8ArrayToRpcData()` to avoid so many transformations.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-tracer.ts
```diff
@@ -16,14 +15,14 @@ import {
...
-const MAX_PRECOMPILE_NUMBER = Object.keys(precompiles).length + 1;
 const DUMMY_RETURN_DATA = Buffer.from([]);
-const DUMMY_GAS_USED = new BN(0);
+const DUMMY_GAS_USED = 0n;
 
 export class VMTracer {
   private _messageTraces: MessageTrace[] = [];
   private _enabled = false;
   private _lastError: Error | undefined;
+  private _maxPrecompileNumber = getActivePrecompiles(this._vm._common).size;
```
alcuadrado, line 25: Nice!

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-tracer.ts
```diff
@@ -39,19 +38,25 @@ export class VMTracer {
     if (this._enabled) {
       return;
     }
-    this._vm.on("beforeMessage", this._beforeMessageHandler);
-    this._vm.on("step", this._stepHandler);
-    this._vm.on("afterMessage", this._afterMessageHandler);
+    (this._vm.evm as EVM).on("beforeMessage", this._beforeMessageHandler);
```
alcuadrado, line 41: 🥲

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-tracer.ts
```diff
@@ -106,11 +111,11 @@ export class VMTracer {
 
         trace = createTrace;
       } else {
-        const toAsBn = new BN(message.to.toBuffer());
+        const toAsNumber = bufferToBigInt(message.to.toBuffer());
```
alcuadrado, line 114: `toAsBigInt`?

packages/hardhat-core/src/internal/hardhat-network/provider/output.ts
```diff
@@ -274,14 +273,14 @@ export function getRpcTransaction(
   };
 }
 
-function getEffectiveGasPrice(tx: TypedTransaction, baseFeePerGas: BN) {
+function getEffectiveGasPrice(tx: TypedTransaction, baseFeePerGas: bigint) {
   const maxFeePerGas = "maxFeePerGas" in tx ? tx.maxFeePerGas : tx.gasPrice;
   const maxPriorityFeePerGas =
     "maxPriorityFeePerGas" in tx ? tx.maxPriorityFeePerGas : tx.gasPrice;
 
-  // baseFeePerGas + min(maxFeePerGas - baseFeePerGas, maxPriorityFeePerGas)
```
alcuadrado, line 282: no more comments like this :)

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeCommon.ts
```diff
@@ -1,43 +1,23 @@
...
       networkId,
       name: networkName,
-      genesis: {
-        timestamp: `0x${initialBlockTimestamp.toString(16)}`,
-        hash: "0x",
-        gasLimit: blockGasLimit,
-        difficulty: 1,
-        nonce: "0x0000000000000042",
-        extraData: "0x1234",
-        stateRoot: bufferToHex(stateTrie.root),
```
alcuadrado, line 38: what about the state root?

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
```diff
@@ -266,15 +266,6 @@ export class ForkStateManager implements EIP2929StateManager {
       throw notCheckpointedError("commit");
     }
     this._stateCheckpoints.pop();
-
-    const storageMap = this._accessedStorage.pop();
```
alcuadrado, line 270: Not necessary anymore?

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
```diff
@@ -468,4 +428,19 @@ export class ForkStateManager implements EIP2929StateManager {
     this._stateRoot = newRoot;
     this._state = state;
   }
+
+  public async hasStateRoot(root: Buffer): Promise<boolean> {
+    return this._state.has(bufferToHex(root));
+  }
+
+  public async flush(): Promise<void> {
+    // not implemented
```
alcuadrado, line 437: We should probably throw

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
```diff
@@ -468,4 +428,19 @@ export class ForkStateManager implements EIP2929StateManager {
...
+  }
+
+  public async modifyAccountFields(
+    _address: Address,
+    _accountFields: any
+  ): Promise<void> {
+    // not implemented
```
alcuadrado, line 444: We should probably throw

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
```diff
@@ -133,13 +147,20 @@ export class VMDebugTracer {
...
+        "Step handler should be called after at least one beforeMessage handler"
+      );
 
-    const structLog = await this._stepToStructLog(step);
-    this._messages[this._messages.length - 1].structLogs.push(structLog);
+      const structLog = await this._stepToStructLog(step);
+      this._messages[this._messages.length - 1].structLogs.push(structLog);
+    } catch (e: any) {
+      // errors thrown in event handlers are lost, so we save this error to
+      // re-throw it in the `trace` function
```
alcuadrado, line 160: when is the `trace` function called? I guess this function is called from `trace`? Can you document this?

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
```diff
@@ -247,16 +247,25 @@ export class ForkBlockchain
...
     }
 
+    const forkNetworkId = this._jsonRpcClient.getNetworkId();
+    let common: Common;
+    if (Chain[forkNetworkId] !== undefined) {
+      common = new Common({ chain: forkNetworkId });
+    } else {
+      // unknown networks use Goerli's common to prevent the
+      // maxExtraDataSize check
+      common = new Common({ chain: 5 });
```
alcuadrado, line 262: won't this break everything? is this used anywhere during the execution?

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
```diff
@@ -247,16 +247,25 @@ export class ForkBlockchain
       rpcBlock === null ||
       rpcBlock.hash === null ||
       rpcBlock.number === null ||
-      rpcBlock.number.gt(this._forkBlockNumber)
+      rpcBlock.number > this._forkBlockNumber
     ) {
       return undefined;
     }
 
+    const forkNetworkId = this._jsonRpcClient.getNetworkId();
```
alcuadrado, line 255: This is cached internally, right?

alcuadrado: also, it should be chainId, right?

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -539,18 +552,18 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
     }
 
-    const remainingBlockCount = count.subn(mineBlockResults.length);
+    const remainingBlockCount = count - BigInt(mineBlockResults.length);
 
     // There should be at least 2 blocks left for the reservation to work,
     // because we always mine a block after it. But here we use a bigger
     // number to err on the safer side.
-    if (remainingBlockCount.lten(5)) {
+    if (remainingBlockCount <= 5) {
```
alcuadrado, line 564: wait, shouldn't this be `5n`?

packages/hardhat-core/src/internal/util/bigint.ts
fvictorio: I wanted to explicitly have `BigIntUtils.fooBar(...)`, at least in some places. But that can also be done with `import * as BigIntUtils` or similar, and I don't think we use this pattern anywhere else, so I'll change it.

packages/hardhat-core/src/internal/util/bigint.ts
fvictorio: It's mainly about mapping types like `number | undefined` to `bigint | undefined`. The problem is that some of those types have more types (e.g. `number | Buffer | undefined`), so I wrote a generic thing.

Agree that the name is not great though. Any suggestions?

packages/hardhat-core/src/internal/util/bigint.ts
fvictorio: Oh yes, we need tests for this module, my bad.

packages/hardhat-core/src/internal/core/jsonrpc/types/input/blockTag.ts
fvictorio: This is actually on my to-do list. I don't remember why the workaround was necessary, but the code compiles and works without it, so... is that enough? Can we be confident that this is no longer necessary?

packages/hardhat-core/src/internal/core/providers/accounts.ts
fvictorio: The tl;dr is that there isn't a lot of difference between creating a `new Common` for a known chain id, and creating a custom common.

In the "Simplify common usages" PR, `new Common` is used only in one or two places in `src`, which means that we are creating commons in the same way almost everywhere. I believe this will make things simpler for us in the future.

packages/hardhat-core/test/internal/hardhat-network/provider/utils/assertEqualTraces.ts
fvictorio: No, because they were all wrongly passing the actual trace first, which was making the messages in failing tests very confusing :sweat_smile: Plus, I think passing the actual value first and the expected second is idiomatic (at least in chai).

fvictorio: Oh, wait, there is one place where the expected trace is passed first. I'll fix that.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
fvictorio: This is a temporary fix, it has a proper fix in a later PR.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
fvictorio: You can use mix numbers and bigints in comparison operators. But we should probably avoid doing that. Maybe there's an eslint rule for that.

packages/hardhat-core/src/internal/util/bigint.ts
fvictorio: I ended up removing this function. It was barely used after the config types changes.

packages/hardhat-core/src/internal/util/bigint.ts
```diff
@@ -0,0 +1,84 @@
...
+
+function mapBigIntToNumber<T>(x: bigint | T): number | T {
+  if (typeof x === "bigint") {
+    return Number(x);
+  }
+
+  return x;
+}
+
+function toWord(x: bigint | number): string {
```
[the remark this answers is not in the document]
fvictorio: Renamed to `toEvmWord` and added a jsdoc.

packages/hardhat-core/src/internal/util/bigint.ts
fvictorio: Removed.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/debug.ts
fvictorio: The block number is pre-Berlin. Maybe we also need some tests for a block that is post-Berlin but pre-London? I don't think it will add a lot of value though.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/web3.ts
fvictorio: Yes but `Uint8Array`s are not `Buffer`s :sweat_smile: And since `Uint8Array.prototype.toString` doesn't accept an encoding, you have to convert it to Buffer.

We can create some helper for this, but it would be just for readability, the number of conversions will be the same. We could write our own `uint8ArrayToHex` but we should benchmark it to verify that it's indeed faster (and my recent attempt to write a faster `bufferToBigInt` makes me doubt that it will be).

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeCommon.ts
fvictorio: It's used by `putGenesisBlock` to create the genesis block. The `Common` still accepts a genesis block. I think it does because then the `Blockchain` class will use it to create the genesis block on demand... but I'm not super sure.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
fvictorio: Yes. The `_accessedStorageMerge` function doesn't exist anymore in the StateManager classes.

This also made me realize that the `_accessedStorage` and `_accessedStorageReverted` properties are also not necessary anymore.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
fvictorio: These methods are called by the EEI, so we can't throw. No-op'ing is fine though, because they are just called (they don't return anything).

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
fvictorio: Sorry, I don't get what you mean here. The `trace` function is the entry point of this class (the only public method), and that's where errors need to be re-thrown.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
fvictorio: > This is cached internally, right?

It's just a getter for a private field of the client. Maybe making that field `public readonly` would make this clearer, wdyt?

> also, it should be chainId, right?

I'm not sure.It's the value returned by the `getNetworkId` function, which uses `net_version`. Maybe that's a mistake.

packages/hardhat-core/src/internal/util/bigint.ts
fvictorio: Fixed in https://github.com/NomicFoundation/hardhat/pull/3082/commits/767ca3f6f7433e595599933ae30963908bbbf992

packages/hardhat-core/src/internal/util/bigint.ts
fvictorio: Fixed in https://github.com/NomicFoundation/hardhat/pull/3082/commits/159b61cc2b2eb572d0d56bcbc279f0749aa94e2f

packages/hardhat-core/src/internal/util/bigint.ts
fvictorio: Added tests for this module in https://github.com/NomicFoundation/hardhat/pull/3082/commits/2ee5e5e5741ba0731d49c5a5859a03962a40fa2f

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-tracer.ts
fvictorio: Renamed in https://github.com/NomicFoundation/hardhat/pull/3082/commits/a44ca54f18b180e87d3bfe62055eb179f09aa1d0

packages/hardhat-core/src/internal/core/jsonrpc/types/input/blockTag.ts
alcuadrado: Yes, we can, as we are not using it anymore. This happens when there are multiple versions of ethereumjs-util, so that typescript sees multiple definitions of BN. Why in these files? No idea.

packages/hardhat-core/src/internal/core/providers/construction.ts
alcuadrado: (we can ignore this comment)

packages/hardhat-core/test/internal/hardhat-network/provider/modules/web3.ts
alcuadrado: Oh, I was getting it wrong. Thanks

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts
alcuadrado: I still think that this one should throw and kill hardhat network if called.  This nop here would lead to a bug if someone uses it, wouldn't it?

alcuadrado: Actually, it's called here: https://github.com/ethereumjs/ethereumjs-monorepo/blob/8cc931c031798b7ccb5c927286ea17043db907b5/packages/evm/src/interpreter.ts#L966

We need to implement it.

fvictorio: Fixed in 917ba24e9. Added tests for this in ac5be4b4e.

PR #3047, opened 2022-08-16
Add support for the merge

fvictorio:


.changeset/green-lemons-design.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
```
alcuadrado, line 2:
```suggestion
"hardhat": minor
```

docs/src/content/hardhat-network/docs/reference/index.md
```diff
@@ -61,7 +62,7 @@ The block gas limit to use in Hardhat Network's blockchain. Default value: `30_0
 
 #### `hardfork`
 
-This setting changes how Hardhat Network works, to mimic Ethereum's mainnet at a given hardfork. It must be one of `"byzantium"`, `"constantinople"`, `"petersburg"`, `"istanbul"`, `"muirGlacier"`, `"berlin"`, `"london"` and `"arrowGlacier"`. Default value: `"arrowGlacier"`
+This setting changes how Hardhat Network works, to mimic Ethereum's mainnet at a given hardfork. It must be one of `"byzantium"`, `"constantinople"`, `"petersburg"`, `"istanbul"`, `"muirGlacier"`, `"berlin"`, `"london"`, `"arrowGlacier"` and `"merge"`. Default value: `"arrowGlacier"`
```
alcuadrado, line 65: Aren't we missing one here? The latest HF?

Also, the new default should be `merge`.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -1365,6 +1365,12 @@ export class EthModule {
       return 0n;
     }
 
+    if (oldBlockTag === "safe" || oldBlockTag === "finalized") {
+      this._checkPostMergeBlockTags(oldBlockTag);
+
+      return this._node.getLatestBlockNumber();
```
alcuadrado, line 1371: Not obvious what we should do here, but I guess this is a good start.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1738,10 +1749,16 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
       coinbase: this.getCoinbaseAddress(),
-      nonce: "0x0000000000000042",
+      nonce: this.isPostMergeHardfork()
+        ? "0x0000000000000000"
+        : "0x0000000000000042",
       timestamp: blockTimestamp,
     };
 
+    if (this.isPostMergeHardfork()) {
+      headerData.mixHash = this._getNextMixHash();
```
alcuadrado, line 1759: I think we should always call `_getNextMixHash` and place this HF check inside it.

The logic is wrong though, mixHash is 0 post merge according to https://eips.ethereum.org/EIPS/eip-3675

alcuadrado: Oh this other Paris overrides that I guess https://eips.ethereum.org/EIPS/eip-4399

alcuadrado:
LGTM, except maybe I thought we were implementing an RPC method (and its helper) to manipulate the randao value. `hardhat_setPrevRandao(quantity)`.

docs/src/content/hardhat-network/docs/reference/index.md
fvictorio: I added support for gray glacier in other PR: https://github.com/NomicFoundation/hardhat/pull/3087/commits/814a8e107dbeed6e027f5e1402fbc81af366924d

> Also, the new default should be merge.

I thought we were going to change the default after the mainnet merge, didn't we?

fvictorio:
I added the `hardhat_setPrevRandao` method in a separate PR: https://github.com/NomicFoundation/hardhat/pull/3088

docs/src/content/hardhat-network/docs/reference/index.md
alcuadrado: The default will be changed after the merge.

PR #3049, opened 2022-08-17
New compiler downloader

alcuadrado:
Second PR of the compilation times optimization effort.

This PR introduces a simpler compiler downloader. 

Highlights:

* The downloader class only handles a platform. If you need to deal with more you need multiple downloaders.
* The downloader validates the compiler checksums right after downloading. This was a HUGE win in compilation times.
* The downloader checks that the binaries are runnable after downloading.
* The compiler offers a concurrent interface, removing the serial fetching that we had before.

WIP: This requires tests.

Depends on https://github.com/NomicFoundation/hardhat/pull/3048


packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
```diff
@@ -24,301 +19,337 @@ export enum CompilerPlatform {
...
-  download: (url: string, destinationFile: string) => Promise<void>;
-  forceSolcJs: boolean;
-}>;
+/**
+ * Default implementation of ICompilerDownloader.
+ *
+ * Important things to note:
+ *   1. If a compiler version is not found, this downloader may fail.
+ *    1.1. It only re-downloads the list of compilers once every X time.
+ *      1.1.1 If a user tries to download a new compiler after X time has passed
```
fvictorio, line 79: Shouldn't this be _before_ X time has passed?

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
```diff
@@ -24,301 +19,337 @@ export enum CompilerPlatform {
...
-    const compilerKeccak256 = ethereumjsUtil.bufferToHex(
-      ethereumjsUtil.keccak(compiler)
-    );
+  /**
+   * Use
+   * @param _platform
+   * @param _compilersDir
+   * @param _compilerListCachePeriodMs
+   * @param _downloadFunction
+   */
```
fvictorio, line 124: This looks like you started writing the description of the function but didn't get past the first word :stuck_out_tongue:

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -544,91 +529,69 @@ subtask(TASK_COMPILE_SOLIDITY_GET_SOLC_BUILD)
...
-        compilerPath = compilerPathResult.compilerPath;
+      log(
+        "Native solc binary doesn't work, using solcjs instead. Try running npx hardhat clean --global"
+      );
 
-        // when using a native binary, check that it works correctly
-        // it it doesn't, force the downloader to use solcjs
-        if (compilerPathResult.platform !== CompilerPlatform.WASM) {
-          log("Checking native solc binary");
+      const wasmDownloader = new CompilerDownloader(
```
fvictorio, line 569: Why does this use `new CompilerDownloader` instead of `CompilerDownloader.getConcurrencySafeDownloader`?

And, if using the static method is better, maybe the constructor should be private?

fvictorio:
I will fix the tests for this PR.

feuGeneA:
> 

No, I'LL fix the tests on this one :rofl:

packages/hardhat-core/src/builtin-tasks/compile.ts
alcuadrado: Good catch. I'll make it private after completing the tests. It may be needed.

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
alcuadrado: right

alcuadrado:
This PR is now ready

fvictorio (approved):

PR #3051, opened 2022-08-17
Optimized build info generation

alcuadrado:
Fourth PR of the compilation times optimization effort.

This PR optimizes how build info files are stored. The way it works is by incrementally generating the JSON and saving it into the FS. The reason this is faster is that node.js is very slow with huge strings, and the build info can be >100mb.

This saved ~0.5s in Open Zeppelin Contracts.

WIP: This requires tests, as it's not a trivial operation. We should check that the JSON is correctly created, especially under edgecases (e.g. files without contracts).

Depends on https://github.com/NomicFoundation/hardhat/pull/3050

### DIFF
--- a/packages/hardhat-core/src/internal/artifacts.ts
+++ b/packages/hardhat-core/src/internal/artifacts.ts
@@ -2,6 +2,7 @@ import debug from "debug";
 import fsExtra from "fs-extra";
 import * as os from "os";
 import * as path from "path";
+import fsPromises from "fs/promises";
 
 import {
   Artifact,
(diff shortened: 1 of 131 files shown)


packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -173,7 +174,80 @@ export class Artifacts implements IArtifacts {
...
+          await file.write(",");
+        }
+
+        await file.write(`${JSON.stringify(name)}:${JSON.stringify(value)}`);
+      }
+
+      await file.write("}");
+
+      await file.write("}");
+      await file.write("}");
```
fvictorio, line 247:
```suggestion
      // close contracts object
      await file.write("}");
      // close output object
      await file.write("}");
      // close build info object
      await file.write("}");
```

fvictorio:
Added tests for this. They fail now because of [this bug](https://github.com/NomicFoundation/hardhat/pull/3050#discussion_r949427766), but I run the tests with a temporary local fix and they passed.

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -173,7 +174,80 @@ export class Artifacts implements IArtifacts {
...
+    // the JSON in the file.
+    const file = await fsPromises.open(buildInfoPath, "w");
+    try {
+      {
+        const withoutOutput = JSON.stringify({
+          ...buildInfo,
+          output: undefined,
+        });
+        await file.write(withoutOutput.substring(0, withoutOutput.length - 1));
+      }
```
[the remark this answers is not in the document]
alcuadrado: to avoid accidental long-lived allocations of the strings.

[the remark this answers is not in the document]
alcuadrado: added a comment

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -173,7 +174,90 @@ export class Artifacts implements IArtifacts {
...
+          ...buildInfo.output,
+          sources: undefined,
+          contracts: undefined,
+        });
+
+        // We start writing the output
+        await file.write(',"output":');
+
+        // Write the output object except for the last }
+        await file.write(outputWithoutSourcesAndContracts.slice(0, 1));
```
fvictorio, line 208: Uhm, shouldn't this have caused some test to fail?

```suggestion
        await file.write(outputWithoutSourcesAndContracts.slice(0, -1));
```

fvictorio: Maybe in the tests this object is always {} so this was working by coincidence?

alcuadrado: Added enough tests to uncover this and similar bugs (none was present) and fixed it.

PR #3052, opened 2022-08-17
Use solidity analyzer

alcuadrado:
Fith PR of the compilation times optimization effort.

This PR stops using @fvictorio's parser in favor of my `solidity-analyzer` library 😜

I believe this should be ready to review.

Depends on https://github.com/NomicFoundation/hardhat/pull/3051


packages/hardhat-core/src/internal/util/scripts-runner.ts
```diff
@@ -33,7 +33,7 @@ export async function runScript(
     });
 
     childProcess.once("close", (status) => {
-      log(`Script ${scriptPath} exited with status code ${status}`);
+      log(`Script ${scriptPath} exited with status code ${status ?? 0}`);
```
fvictorio, line 36: Weird, the [docs](https://nodejs.org/api/child_process.html#event-close) say that this is always a number. I wonder if this is a bug in `@types/node`.

fvictorio (approved):

fvictorio (approved):

PR #3053, opened 2022-08-17
Misc optimizations

alcuadrado:
Sixth PR of the compilation times optimization effort.

This PR introduces multiple smaller optimizations:

* I created an fs-util module to replace slower dependencies with it (e.g. `glob`)
* Stopped using `loadash` whenever possible, and only load its necessary modules when we still use it.
* Use `execFile` instead of `exec` to run the compiler, as `exec` spawns a new shell (!!!!!!)
* Don't import `hardhat/console.sol` in the sample project as it's slow to compile.
* Some hardhat load time optimizations — Mostly requiring things when needed instead of at the top of the file
* Cache the solidity files resolution results — We were reading each file multiple times (!)
* We were reading some files just to check for their existence (!)

WIP: Parts of this require tests.

Depends on https://github.com/NomicFoundation/hardhat/pull/3052

### DIFF
--- a/packages/hardhat-core/sample-projects/javascript/contracts/Lock.sol
+++ b/packages/hardhat-core/sample-projects/javascript/contracts/Lock.sol
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.9;
 
 // Import this file to use console.log
-import "hardhat/console.sol";
+// import "hardhat/console.sol";
 
 contract Lock {
     uint public unlockTime;
@@ -21,7 +21,7 @@ contract Lock {
     }
 
     function withdraw() public {
-        // Uncomment this line to print a log in your terminal
+        // Uncomment this line and the import to "hardhat/console.sol" to print a log in your terminal
         // console.log("Unlock time is %o and block timestamp is %o", unlockTime, block.timestamp);
 
         require(block.timestamp >= unlockTime, "You can't withdraw yet");
--- a/packages/hardhat-core/sample-projects/javascript/contracts/Lock.sol
+++ b/packages/hardhat-core/sample-projects/javascript/contracts/Lock.sol
@@ -21,7 +21,7 @@ contract Lock {
     }
 
     function withdraw() public {
-        // Uncomment this line and the import to "hardhat/console.sol" to print a log in your terminal
+        // Uncomment this line, and the import of "hardhat/console.sol", to print a log in your terminal
         // console.log("Unlock time is %o and block timestamp is %o", unlockTime, block.timestamp);
 
         require(block.timestamp >= unlockTime, "You can't withdraw yet");
--- a/packages/hardhat-core/sample-projects/javascript/contracts/Lock.sol
+++ b/packages/hardhat-core/sample-projects/javascript/contracts/Lock.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.9;
 
-// Import this file to use console.log
+// Uncomment this line to use console.log
 // import "hardhat/console.sol";
 
 contract Lock {
--- a/packages/hardhat-core/src/internal/artifacts.ts
+++ b/packages/hardhat-core/src/internal/artifacts.ts
@@ -64,12 +64,8 @@ export class Artifacts implements IArtifacts {
   }
 
   public async artifactExists(name: string): Promise<boolean> {
-    try {
-      await this.readArtifact(name);
-      return true;
-    } catch {
-      return false;
-    }
+    const artifactPath = await this._getArtifactPath(name);
+    return fsExtra.pathExists(artifactPath);
   }
 
   public async getAllFullyQualifiedNames(): Promise<string[]> {
--- a/packages/hardhat-core/src/internal/artifacts.ts
+++ b/packages/hardhat-core/src/internal/artifacts.ts
@@ -472,8 +472,6 @@ export class Artifacts implements IArtifacts {
         )
       );
 
-      console.log({ trueCasePath, artifactPath });
-
       if (artifactPath !== trueCasePath) {
         throw new HardhatError(ERRORS.ARTIFACTS.WRONG_CASING, {
           correct: this._getFullyQualifiedNameFromPath(trueCasePath),
--- a/packages/hardhat-core/src/internal/artifacts.ts
+++ b/packages/hardhat-core/src/internal/artifacts.ts
@@ -34,8 +34,8 @@ import {
   FileNotFoundError,
   getAllFilesMatching,
   getAllFilesMatchingSync,
-  getRealCase,
-  getRealCaseSync,
+  getFileTrueCase,
+  getFileTrueCaseSync,
 } from "./util/fs-utils";
 
 const log = debug("hardhat:core:artifacts");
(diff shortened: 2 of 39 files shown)


packages/hardhat-core/src/internal/util/glob.ts
```diff
@@ -12,6 +15,9 @@ export async function glob(
   return files.map(path.normalize);
 }
 
+/**
+ * deprecated
```
fvictorio, line 19:
```suggestion
 * @deprecated
```

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -96,21 +90,32 @@ export class Artifacts implements IArtifacts {
...
+        !f.endsWith(".dbg.json")
+    );
 
     return paths.sort();
   }
 
   public async getBuildInfoPaths(): Promise<string[]> {
-    const paths = await glob(this._buildInfosGlob);
+    const paths = await getAllFilesMatching(
+      path.join(this._artifactsPath, BUILD_INFO_DIR_NAME)
```
fvictorio, line 108: To exactly match the previous implementation, this should filter by `f.endsWith(".json")`

packages/hardhat-core/src/internal/solidity/compilation-job.ts
```diff
@@ -130,7 +129,8 @@ function mergeCompilationJobs(
     }
   }
 
-  return flatten([...jobsMap.values()]);
+  // Array#flat This method defaults to depth limit 1
```
fvictorio, line 132: This answers my previous question. Wow. I had no idea. Should we add a linter rule about not using `.flat` without an argument or something like that? This seems super error-prone.

packages/hardhat-core/src/internal/solidity/resolver.ts
```diff
@@ -116,23 +128,38 @@ export class Resolver {
...
-        return await this.resolveSourceName(normalizeSourceName(imported));
+      let sourceName: string;
+
+      const isRelativeImport = this._isRelativeImport(imported);
+
+      if (!isRelativeImport) {
+        sourceName = normalizeSourceName(imported);
+      } else {
+        sourceName = await this._relativeImportToSourceName(from, imported);
       }
```
fvictorio, line 139:
```suggestion
      if (isRelativeImport) {
        sourceName = await this._relativeImportToSourceName(from, imported);
      } else {
        sourceName = normalizeSourceName(imported);
      }
```

packages/hardhat-core/src/utils/source-names.ts
```diff
@@ -193,25 +186,19 @@ export function replaceBackslashes(str: string): string {
  * `p` doesn't exist. `p` MUST be in source name format.
  */
 async function getPathTrueCase(fromDir: string, p: string): Promise<string> {
-  const { trueCasePath } = await import("true-case-path");
-
   try {
-    const tcp = await trueCasePath(p, fromDir);
+    const absolute = path.join(fromDir, p);
+    const tcp = await getRealCase(absolute);
```
fvictorio, line 191: This variable name is not very descriptive now that we are not using `trueCasePath` anymore.

packages/hardhat-core/src/internal/util/fs-utils.ts
```diff
@@ -0,0 +1,164 @@
...
+    // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
+    throw new FileSystemAccessError(e.message, e);
+  }
+
+  const results = await Promise.all(
+    dir.map(async (file) => {
+      const absolutePathToFile = path.join(absolutePathToDir, file);
+      const stats = await fsPromises.stat(absolutePathToFile);
+      if (stats.isDirectory()) {
+        return getAllFilesMatching(absolutePathToFile, matches);
```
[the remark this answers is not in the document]
alcuadrado: Nice finding! 

I'm fine with that error. Anyone creating a symlink loop should burn in hell anyway

packages/hardhat-core/src/internal/util/fs-utils.ts
```diff
@@ -0,0 +1,164 @@
...
+      } else {
+        if (matches === undefined || matches(absolutePathToFile)) {
```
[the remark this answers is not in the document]
alcuadrado: Disclaimer: this is an after-the-fact justification.

I'd rather not create a new function on each call as this is highly recursive and we know that it's called a TON of times.

packages/hardhat-core/src/internal/util/glob.ts
```diff
@@ -3,6 +3,9 @@ import type { IOptions as GlobOptions } from "glob";
 import * as path from "path";
 import util from "util";
 
+/**
+ * @deprecated
+ */
```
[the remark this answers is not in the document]
alcuadrado: The truffle plugin uses it. My fault 😭

I'll add a comment

packages/hardhat-core/src/internal/solidity/compilation-job.ts
alcuadrado: That'd be useful, but I can't find any rule for that

alcuadrado: Slightly OT: https://developer.chrome.com/blog/smooshgate/

packages/hardhat-core/sample-projects/javascript/contracts/Lock.sol
```diff
@@ -21,7 +21,7 @@ contract Lock {
     }
 
     function withdraw() public {
-        // Uncomment this line to print a log in your terminal
+        // Uncomment this line and the import to "hardhat/console.sol" to print a log in your terminal
```
feuGeneA, line 24:
```suggestion
        // Uncomment this line, and the import of "hardhat/console.sol", to print a log in your terminal
```

packages/hardhat-core/src/internal/artifacts.ts
alcuadrado: done

packages/hardhat-core/src/utils/source-names.ts
alcuadrado: right. renamed it.

packages/hardhat-core/src/internal/util/fs-utils.ts
```diff
@@ -0,0 +1,164 @@
...
+}
+
+/**
+ * Sync version of getRealCase
+ *
+ * @see getRealCase
+ */
+export function getRealCaseSync(absolutePath: string): string {
+  try {
+    return fs.realpathSync.native(absolutePath);
```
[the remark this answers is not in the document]
alcuadrado: done

alcuadrado:
Finally ready! This should have been a few prs.

Handling different filesystems at the same time is so painful!

packages/hardhat-core/test/internal/core/config/config-loading.ts
```diff
@@ -334,13 +342,19 @@ Hardhat plugin instead.`
 
         const files = ctx.getFilesLoadedDuringConfig();
 
+        const filesJson = JSON.stringify(files, undefined, 2);
+
         for (const file of builtinTasksFiles) {
-          // The task names may have been loaded before, so we ignore it.
-          if (file.endsWith("task-names.ts")) {
+          // The task names and the utils may have been loaded before, so we ignore them.
+          if (file.endsWith("task-names.ts") || file.includes("utils")) {
```
fvictorio, line 349: Maybe?

```suggestion
          if (file.endsWith("task-names.ts") || file.includes(path.join(path.sep, "utils", path.sep))) {
```

PR #3057, opened 2022-08-18
docs: fix typo

EduardoBorsa:
Just fixed a typo in the documentation for the testing-contracts that will eventually be used to generate the Lock.ts test.


fvictorio:
Thank you @EduardoBorsa! I also fixed it in the [boilerplate project](https://github.com/NomicFoundation/hardhat-boilerplate/blob/master/test/Token.js).

PR #3065, opened 2022-08-20
(WIP) Simplify Common usages

fvictorio:


alcuadrado:
This looks good, but I fear it may break with a non-standard (as in not natively supported by ethreumjs) network.

PR #3068, opened 2022-08-20
(DONT MERGE) Try ethereumjs master branch

fvictorio:
This PR is not meant to be merged, just to run the CI with the latest code from ethereumjs. To do that, I published all the ethereumjs packages we use under the `@ignored` org, and updated the imports.

After ethereumjs publishes a new beta (or a stable release), we should upgrade and rebase all the commits here except the first one.


fvictorio:
Some of the failing tests are because there's an assertion that expects to see `@ethereumjs` instead of `@ignored`, so we can disregard that one.

The other failing tests are caused by `VM.runBlock` validating the header even when the `skipBlockValidation` flag is set.

alcuadrado:
I'm not sure what I should look for here

fvictorio:
Closing in favor of #3087.

PR #3073, opened 2022-08-24
Add changeset check

fvictorio:
Testing the changeset check


fvictorio:
no changeset needed

fvictorio:
Ok, this seems to work fine. This is what I tried:

- I opened this PR with a changeset. The check passed.
- I removed the changeset. The check failed.
- I commented "no changeset needed" and re-run the workflow. The check passed.

My guess is that the first one worked because the changesets bot commented before the workflow searched the comments. So there's a possible race condition there. The solution might be to run the workflow for each new comment in a PR, but I'd say we use this and see what happens. If it turns out that that race condition is triggered often, we can then improve the workflow.

The only pending thing is that @alcuadrado is concerned about using third-party actions. One way to mitigate this is to use a specific commit instead of a version. The code of that action is really small so it should be easy to audit. Alternatively, we could fork it and use that.

fvictorio:
test

alcuadrado:
I revamped this. Sorry, but not sorry.

It had an error that prevented it from running. Workflows don't support `$default-branch`. Only the templates do. Insane decision from Github.

But while I was at it, I changed how it works. Not it fails if there's no changeset and the pr is not labeled as "no changeset needed". It supports adding/removing both changesets and the label.

I tested it extensively in a private repository.

PR #3082, opened 2022-08-26
EthereumJS upgrade: fix comments after PR review

fvictorio:


packages/hardhat-core/src/internal/hardhat-network/provider/node-types.ts
```diff
@@ -17,17 +17,17 @@ export function isForkedNodeConfig(
 
 interface CommonConfig {
   automine: boolean;
-  blockGasLimit: bigint;
+  blockGasLimit: number;
```
alcuadrado, line 20: These can be `bigint`, as it's an internal type. If it's more convenient to revert to `number` that's fine.

packages/hardhat-core/src/types/config.ts
```diff
@@ -134,15 +134,15 @@ export type HardhatNetworkChainsConfig = Map<
...
   gasMultiplier: number;
-  initialBaseFeePerGas?: bigint;
+  initialBaseFeePerGas?: number;
   hardfork: string;
   mining: HardhatNetworkMiningConfig;
   accounts: HardhatNetworkAccountsConfig;
-  blockGasLimit: bigint;
-  minGasPrice: bigint;
+  blockGasLimit: number;
+  minGasPrice: number;
```
alcuadrado, line 145: this one used to be a BN, so I'd change it to bigint

fvictorio: Changed in 236fb02c0b49e39990a714fe5d3f447e049ab722

PR #3087, opened 2022-08-27
Upgrade ethereumjs to RC versions

fvictorio:


fvictorio:
The only failing test is a false negative because of a timeout. I think this one is ready for review now.

PR #3088, opened 2022-08-28
Add `hardhat_setPrevRandao` method and network helper

fvictorio:


packages/hardhat-network-helpers/src/helpers/setPrevRandao.ts
```diff
@@ -0,0 +1,18 @@
+import type { NumberLike } from "../types";
+import { getHardhatProvider, toPaddedRpcQuantity } from "../utils";
+
+/**
+ * Sets the PREVRANDAO value of the next block.
+ *
+ * @param prevRandao The new PREVRANDAO value to use.
+ */
+export async function setPrevRandao(prevRandao: NumberLike): Promise<void> {
```
alcuadrado, line 9: Shouldn't this be documented somewhere?

fvictorio: It is here: https://github.com/NomicFoundation/hardhat/blob/f598c18e86ce03f75a8885b0710ac6cde848da35/docs/src/content/hardhat-network-helpers/docs/reference.md?plain=1#L302

PR #3094, opened 2022-08-28
Cache artifact paths

alcuadrado:
Seventh PR optimizing the compilation and test loop.

This PR modifies `hre.artifacts` so that it caches the paths to the different files.

Depends on #3053

### DIFF
--- a/packages/hardhat-core/src/internal/artifacts.ts
+++ b/packages/hardhat-core/src/internal/artifacts.ts
@@ -335,35 +335,42 @@ export class Artifacts implements IArtifacts {
     // We clear the cache here, as we want to be sure this runs correctly
     this.clearCache();
 
-    const validArtifactPaths = await Promise.all(
-      this._validArtifacts.flatMap(({ sourceName, artifacts }) =>
-        artifacts.map((artifactName) =>
-          this._getArtifactPath(getFullyQualifiedName(sourceName, artifactName))
-        )
-      )
-    );
-
-    const validArtifactsPathsSet = new Set<string>(validArtifactPaths);
-
-    for (const { sourceName, artifacts } of this._validArtifacts) {
-      for (const artifactName of artifacts) {
-        validArtifactsPathsSet.add(
-          this.formArtifactPathFromFullyQualifiedName(
-            getFullyQualifiedName(sourceName, artifactName)
+    try {
+      const validArtifactPaths = await Promise.all(
+        this._validArtifacts.flatMap(({ sourceName, artifacts }) =>
+          artifacts.map((artifactName) =>
+            this._getArtifactPath(
+              getFullyQualifiedName(sourceName, artifactName)
+            )
           )
-        );
+        )
+      );
+
+      const validArtifactsPathsSet = new Set<string>(validArtifactPaths);
+
+      for (const { sourceName, artifacts } of this._validArtifacts) {
+        for (const artifactName of artifacts) {
+          validArtifactsPathsSet.add(
+            this.formArtifactPathFromFullyQualifiedName(
+              getFullyQualifiedName(sourceName, artifactName)
+            )
+          );
+        }
       }
+
+      const existingArtifactsPaths = await this.getArtifactPaths();
+
+      await Promise.all(
+        existingArtifactsPaths
+          .filter((artifactPath) => !validArtifactsPathsSet.has(artifactPath))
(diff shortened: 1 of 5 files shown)


alcuadrado:
Ready to review!

packages/hardhat-core/src/types/artifacts.ts
```diff
@@ -97,6 +97,27 @@ export interface Artifacts {
...
+
+  /**
+   * Starting with Hardhat 2.11.0, the artifacts object caches the information
+   * about paths that it fetches from the filesystem (e.g. the list of
+   * artifacts, the path that an artifact name resolves to, etc.). The artifacts
+   * and buildInfos themselves are not cached, only their paths.
+   *
+   * This method, if present, clears that cache.
+   */
+  clearCache?: () => void;
```
fvictorio, line 109: Should we be calling this in the `clean` task?

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -40,9 +40,22 @@ import {
...
+  debugFilePaths?: string[];
+  buildInfoPaths?: string[];
+  artifactNameToArtifactPathCache: Map<string, string>;
+  artifactFQNToBuildInfoPathCache: Map<string, string>;
+}
+
 export class Artifacts implements IArtifacts {
   private _validArtifacts: Array<{ sourceName: string; artifacts: string[] }>;
 
+  private _cache?: Cache = {
```
fvictorio, line 54: Let's add a comment saying that an undefined `_cache` signals a disabled cache.

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -265,13 +323,18 @@ export class Artifacts implements IArtifacts {
...
+
     return buildInfoPath;
   }
 
   /**
    * Remove all artifacts that don't correspond to the current solidity files
    */
   public async removeObsoleteArtifacts() {
+    // We clear the cache here, as we want to be sure this runs correctly
+    this.clearCache();
```
fvictorio, line 336: Should we run this at the end of the function too?

packages/hardhat-core/test/internal/artifacts.ts
```diff
@@ -3,12 +3,13 @@ import fsExtra from "fs-extra";
 import * as os from "os";
 import * as path from "path";
 
+import { beforeEach } from "mocha";
```
fvictorio, line 6: ?

packages/hardhat-core/test/internal/artifacts.ts
```diff
@@ -909,6 +910,213 @@ describe("Artifacts class", function () {
...
+        artifacts.clearCache();
+
+        assert.lengthOf(await artifacts.getArtifactPaths(), 2);
+        assert.lengthOf(await artifacts.getDebugFilePaths(), 2);
+        assert.lengthOf(await artifacts.getBuildInfoPaths(), 2);
+      });
+
+      it("Shouldn't re-enable the cache", async function () {
+        artifacts.disableCaching();
+        artifacts.clearCache();
```
fvictorio, line 1008: I guess `disableCaching` should call `clearCache`?

packages/hardhat-core/test/internal/artifacts.ts
alcuadrado: ??

fvictorio: I think this is an accidental auto-import. We don't do this normally (mainly because it prevents the file from being run stand-alone with mocha).

packages/hardhat-core/test/internal/artifacts.ts
alcuadrado: It was already working like that. This test was wrong, so I fixed it, and also added a test verifying that `disableCaching` actually clears it.

packages/hardhat-core/src/internal/artifacts.ts
alcuadrado: done

packages/hardhat-core/src/types/artifacts.ts
alcuadrado: Ohhh! I hadn't thought about it. Yes!

fvictorio (approved):

PR #3105, opened 2022-08-30
Avoid using glob in plugins if possible

fvictorio:


packages/hardhat-truffle4/src/glob.ts
```diff
@@ -0,0 +1,19 @@
+import { join } from "path";
+
+export async function getSolidityFiles(directory: string): Promise<string[]> {
+  try {
+    const { getAllFilesMatching } = await import(
+      "hardhat/internal/util/fs-utils"
+    );
+
+    return await getAllFilesMatching(directory, (f) => f.endsWith(".sol"));
```
fvictorio, line 9: I'm not 100% sure about this because it means we can't delete, move or modify `getAllFilesMatching`.

Maybe we should also check that it's not `undefined`?

alcuadrado: Makes sense. Let's use glob if it's `undefined`.

.changeset/tame-mangos-appear.md
```diff
@@ -0,0 +1,7 @@
+---
+"@nomiclabs/hardhat-truffle4": patch
+"@nomiclabs/hardhat-truffle5": patch
+"@nomiclabs/hardhat-vyper": patch
+---
+
+Avoid using glob when possible
```
alcuadrado, line 7:
```suggestion
Improve the compilation performance by not using the `glob` library whenever we can avoid it.
```

PR #3131, opened 2022-09-04
Add a guide about how to use Chainlink

alcuadrado:
Hey @StephenFluin @PatrickAlphaC,

Moving the conversation here.

I created this PR to get the ball rolling. Please don't consider this final. I just created the necessary boilerplate for this and imported the ethereum.org guide almost as is.

While that guide is nice, it has nothing to do with Hardhat, so it can be confusing. For example, it uses Remix instead of Hardhat.


PatrickAlphaC:
This is awesome!!

Is there a way the docs here can just track the chainlink docs? I feel like that will make life easier so we don't have to update them periodically

alcuadrado:
TBH I have no idea, but can you point me to the docs (including sources) so that we can see if there's something we can do? @PatrickAlphaC

PatrickAlphaC:
Let me check... Yeah this would make life easiest.

alcuadrado:
This PR depends on #3604

PR #3147, opened 2022-09-07
Upgrade solidity-coverage in toolbox and fix sample projects READMEs

fvictorio:


.changeset/mean-berries-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-toolbox": patch
```
alcuadrado, line 2:
```suggestion
"@nomicfoundation/hardhat-toolbox": major
```

I think this needs to be a major, otherwise, people with the previous version will get this update and end up in a broken state. Or, npm 7+ won't even install it.

Being a major, I think we should provide a better description.

alcuadrado:
I left a comment and added a missing changeset

apecollector:
This is needed since currently toolbox contains a broken coverage command that was fixed:

https://github.com/sc-forks/solidity-coverage/issues/737

```
npx hardhat coverage

Version
=======
> solidity-coverage: v0.7.21

Instrumenting for coverage...
=============================

> Test.sol

Compilation:
============

Generating typings for: 19 artifacts in dir: typechain-types for target: ethers-v5
Successfully generated 58 typings!
Compiled 19 Solidity files successfully
Error in plugin solidity-coverage: TypeError: cur._node._vm.on is not a function

For more info run Hardhat with --show-stack-traces
```

fvictorio:
@apecollector we are going to release this today, but you don't need to wait for us to upgrade solidity-coverage, because it's a peer dependency of the toolbox, not a dependency.

Just `npm i --save-dev solidity-coverage@latest` should do the trick.

Issue #680, opened 2020-07-06
Improve colors when printing compiler errors


frangio:
Solidity errors and warnings are currently printed in full red or yellow color. While some color is nice, printing the entire text in a single color isn't comfortable to read.

I'd say only the error category (`Warning`, `ParserError`, ...) should be printed in the corresponding yellow or red color.

It could also be nice to experiment with further formatting of other things like the source code location, the error description itself, or the `^----^` marker that points at the error location, using bold, dimming, or more colors.

alcuadrado:
Thanks for opening this issue, @frangio 

We are currently redoing our compilation pipeline, so it's a great time to implement changes to how errors are displayed. This is being tracked in #553, where printing colorful errors pointing to the exact place were they happened was also suggested.

frangio:
I want to raise this again. :slightly_smiling_face: Given the general verbosity of Solidity error messages it would be very helpful for users if the output colors were improved.

cameel:
We're considering adding a version of these messages with ANSI escapes to Standard JSON output (https://github.com/ethereum/solidity/issues/11507). Would that be usable for Hardhat?

alcuadrado:
Really cool, @cameel !

axic:
Is hardhat output used in any other context than a terminal? Such as in browsers?

fvictorio:
> Is hardhat output used in any other context than a terminal? Such as in browsers?

No. Right now it's not possible to use Hardhat in a browser, so no. And I can't think about any other context where it could be used.

axic:
I assume hardhat also uses the `formattedMessage` of the error list from Solidity? So then if there's an option to make it have colours, it would have ANSI escapes for it. There are some libraries which can be used to transform ANSI escapes into HTML spans, hence you'd need that if hardhat would be used in the browser.

However if you already craft the error messages yourself, by using `message` and the other fields, then you can already colour it.

axic:
Here's a concrete proposal, happy to hear your opinions: https://github.com/ethereum/solidity/issues/11507#issuecomment-919556059

Issue #1846, opened 2021-09-01
Handle `info` messages in solidity


fvictorio:
At the moment, solidity has two kind of "messages" after a compilation: warnings and errors. But a third one, info, will be added. See [this issue](https://github.com/ethereum/solidity/issues/11508).

It seems like this won't break hardhat. The only problem is that these info messages are shown in yellow (as warnings). At the very least we should change that, but we might also have to consider things like allowing users to enable/disable different levels of verbosity, or configure them in some other way.

cameel:
I think it's still relevant. The compiler does emit these and they should be handled (though currently there are not many cases in which they appear, mostly in SMTChecker I think).

fvictorio:
It is indeed, thank you.

fvictorio:
@cameel do you have an example of some code that produces an info message in some solc version?

cameel:
I think that currently the only one is error 1180 from SMT checker. Check out [the command-line test `standard_model_checker_invariants_contract`](https://github.com/ethereum/solidity/tree/develop/test/cmdlineTests/standard_model_checker_invariants_contract).

We do plan to use it more and maybe convert some of the existing warnings into `info` messages but this has been a low priority so far. May take some time until we focus specifically on that unless someone specifically requests some warning to be demoted.

fvictorio:
Thanks for the clarification. I'm going to mark this as blocked until there is some info message in solc that can show up in a normal Hardhat usage.

PR #3150, opened 2022-09-07
Improve colors when printing compiler errors

frangio:
Fixes https://github.com/NomicFoundation/hardhat/issues/680

---

Before:

image

After:

image

Opening as a Draft because as you can see there are other errors which are still printed entirely colored. Changing that would be more effort but it would look more consistent. Let me know your thoughts.

Note this may conflict with a potential future Solidity feature: https://github.com/ethereum/solidity/issues/11507

To-do:
- [ ] Check if we can handle https://github.com/NomicFoundation/hardhat/issues/1846 here

### DIFF
--- a/.changeset/few-sheep-drive.md
+++ b/.changeset/few-sheep-drive.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Stop colorizing the entire message when an error is printed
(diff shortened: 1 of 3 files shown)


alcuadrado:
@fvictorio WDYT? I like this change

alcuadrado:
We should do the same for `Error HHXXX:`

alcuadrado:
👀

PR #3155, opened 2022-09-08
feat: Load hardhat.config.cjs to partially support ESM projects

phated:
- [x] I talked about the limitations in the ESM PR and suggestion in the ts-node transpile PR and discussed this with @fvictorio 

---

Due to the way hardhat registers tasks, v3 can't support ESM (as I outlined at https://github.com/NomicFoundation/hardhat/pull/1994#issuecomment-972459102); however, you can load `.cjs` files inside an ESM project. Since many projects are converting to ESM-only, it is helpful to be able to run hardhat in those contexts. 

This change should allow a monorepo to become `"type": "module"` and still use hardhat in commonjs mode. 🎉 

This support is in parallel to my other PR (#3017) which adds this support for TypeScript. There are follow up PRs to allow loading of ESM tasks and scripts.

### DIFF
--- a/packages/hardhat-core/src/internal/core/project-structure.ts
+++ b/packages/hardhat-core/src/internal/core/project-structure.ts
@@ -7,6 +7,7 @@ import { getPackageRoot } from "../util/packageInfo";
 import { HardhatError } from "./errors";
 import { ERRORS } from "./errors-list";
 const JS_CONFIG_FILENAME = "hardhat.config.js";
+const CJS_CONFIG_FILENAME = "hardhat.config.cjs";
 const TS_CONFIG_FILENAME = "hardhat.config.ts";
 
 export function isCwdInsideProject() {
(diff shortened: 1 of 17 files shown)


phated:
With this `.cjs` change, I found that the `run` task just supports ESM, so I've added those tests to this PR! 🎉

alcuadrado:
👑

alcuadrado:
Should we review and merge this and #3017 in any particular order?

phated:
@alcuadrado I think we should merge this one first. I believe that the TS change should also add tests for run/test scripts after those are done for JS.

phated:
@alvrs was testing this out and pointed out that I didn't update `isCwdInsideProject` 🙈 This fixes the issue and adds a test

@alcuadrado Do you think this can land soon so I can rebase #3156 and update #3017 ?

PR #3156, opened 2022-09-08
feat: Support ESM test running via the test task

phated:
- [x] I talked about the limitations in the ESM PR and suggestion in the ts-node transpile PR and discussed this with @fvictorio 

---

This builds upon #3155 to support loading ESM tests within a `"type": "module"` when using hardhat in commonjs mode. It requires the `.cjs` support from #3155 and then a one-line change to make mocha attempt to load ESM modules.

I can rebase this when the other PR lands to just the last commit (check that out for the smaller diff).

I'm getting some parallel test failures, but those seem to exist for the CJS runs currently (maybe related to #3153?)


fvictorio:
I'm not sure why these tests are failing, but I'll look into it.

PR #3160, opened 2022-09-09
Support Solidity 0.8.17

feuGeneA:
for https://linear.app/nomic-foundation/issue/HH-1159/ensure-support-for-solidity-0817


alcuadrado:
Maybe we should also bump it here:

1. https://github.com/NomicFoundation/hardhat/blob/master/packages/hardhat-core/sample-projects/javascript/hardhat.config.js#L5
2. https://github.com/NomicFoundation/hardhat/blob/master/packages/hardhat-core/sample-projects/typescript/hardhat.config.ts#L5

WDYT @fvictorio @feuGeneA?

fvictorio:
Yes, that's a good idea.

PR #3161, opened 2022-09-09
Fix breaking change introduced in Hardhat 2.9.8

alcuadrado:
This PR fixes a breaking change introduced in bd2bbc0ef3. Thanks @nventuro for reporting it.

### DIFF
--- a/.changeset/hungry-peas-jog.md
+++ b/.changeset/hungry-peas-jog.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed a breaking change in the configuration type that was accidentally introduced in Hardhat 2.9.8.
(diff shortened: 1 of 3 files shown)


alcuadrado:
We need to come up with a way of testing these type changes. For once, the commit that changed the type changed a ton of unrelated tests, which should ring a bell. Maybe having tests that shouldn't be changed?

fvictorio (approved):

fvictorio:
> We need to come up with a way of testing these type changes.

I added a test file that's only meant to test the types of the config. This file will stop compiling if we add a new mandatory field to some type.

It doesn't cover all possible cases. For example, converting some type from `string` to `string | number` would be a somewhat breaking change that won't be caught by that test.

PR #3165, opened 2022-09-12
[Rethnet] Implement and use SyncStateManager

fvictorio:


fvictorio:
@Wodann I think this is no longer necessary because we can make async calls from rethnet, right? I'm going to close it.

Wodann:
Correct

Issue #3173, opened 2022-09-13
`verify:verify` task usage causes duplicated runs of `compile` task


nkuba:
Currently when the `verify:verify` subtask is executed programmatically to verify contracts (e.g. in a deployment script) it runs the `compile` task as well. The compilation is rerun multiple times during the deployment process, which is not necessary as it's assumed the compilation was already called before running the deployment. 

In the example below I want to deploy three contracts (with `hardhat-deploy` plugin) and then verify them. The contracts will be compiled before the deployment and then each `verify:verify` execution will compile them again.
```
const ContractA = await hre.deployments.deploy("ContractA")
const ContractB = await hre.deployments.deploy("ContractB")
const ContractC = await hre.deployments.deploy("ContractC")

await hre.run("verify:verify", { address: ContractA.address })
await hre.run("verify:verify", { address: ContractB.address })
await hre.run("verify:verify", { address: ContractC.address })
```

fvictorio:
Hi @nkuba, the `verify:verify` task now accepts a `noCompile` flag. Do you think that is enough to close this issue?

PR #3174, opened 2022-09-13
Move `compile` task call from `verify:verify` subtask to `verify` task

nkuba:
Currently when the `verify:verify` subtask is executed programmatically to verify contracts (e.g. in a deployment script) it runs the `compile` task as well. The compilation is rerun multiple times during the deployment process, which is not necessary as it's assumed the compilation was already called before running the deployment.

In the example below I want to deploy three contracts (with `hardhat-deploy` plugin) and then verify them. The contracts will be compiled before the deployment and then each `verify:verify` execution will compile them again.
```
const ContractA = await hre.deployments.deploy("ContractA")
const ContractB = await hre.deployments.deploy("ContractB")
const ContractC = await hre.deployments.deploy("ContractC")

await hre.run("verify:verify", { address: ContractA.address })
await hre.run("verify:verify", { address: ContractB.address })
await hre.run("verify:verify", { address: ContractC.address })
```

Here I propose that the `compile` task execution is moved to the main `verify` task with an optional flag `noCompile` to skip the compilation. This is inspired by the way the `test` task handles compilation: https://github.com/NomicFoundation/hardhat/blob/d377c39aee35234b370a90a1a4a1227717061719/packages/hardhat-core/src/builtin-tasks/test.ts#L153-L155

Closes: https://github.com/NomicFoundation/hardhat/issues/3173

### DIFF
--- a/packages/hardhat-etherscan/src/index.ts
+++ b/packages/hardhat-etherscan/src/index.ts
@@ -80,6 +80,9 @@ interface VerificationArgs {
 
   // --list-networks flag
   listNetworks: boolean;
+
+  // Skip contracts compilation; --no-compile flag
+  noCompile: boolean;
 }
 
 interface VerificationSubtaskArgs {
@@ -134,6 +137,7 @@ const verify: ActionType<VerificationArgs> = async (
     contract,
     libraries: librariesModule,
     listNetworks,
+    noCompile,
   },
   { config, run }
 ) => {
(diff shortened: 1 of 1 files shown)


fvictorio:
Hey @nkuba, thanks a lot for sending this. I will do a proper review as soon as I can, but my initial impression is that this makes sense.

JChiaramonte7:
+1 this change, something i use on my own fork that would be good to have merged in

GopherJ:
we also need this

alcuadrado:
Closing this PR as support for `--no-compile` has already been released.

PR #3175, opened 2022-09-13
FakeSenderTransaction: fake a signature

### LINKED ISSUE #1963: impersonating two different accounts got exactly the same tx hash
## How to reproduce
- download files from this [gist](https://gist.github.com/Detoo/94faefe5c41324b8365582952f50193d)
- start hardhat fork node on `rinkeby` block number `9369746`
- execute run.ts

## Expected results
saw 24 different txs with unique hashes

## Actual results
two of the hashes collides:
```typescript
// from: 0xC18F4EfD791e2fa6D34aF94a1B0a6bca65402E58, tx hash: 0x1bc754ec661fd6b64cbbf4b12375f9e3579bfcb30dd6232b1aa75113bd159d8d, blockNumber: 9369757
// from: 0x9D1Cc5b07121EEe7fa1427BDdD6871d2B93536df, tx hash: 0x1bc754ec661fd6b64cbbf4b12375f9e3579bfcb30dd6232b1aa75113bd159d8d, blockNumber: 9369758
```

feuGeneA:
Fixes #1963


packages/hardhat-core/src/internal/hardhat-network/provider/transactions/FakeSenderTransaction.ts
```diff
@@ -100,18 +98,14 @@ export class FakeSenderTransaction extends Transaction {
...
-      .digest();
+    const fakeSignature = makeFakeSignature(data, sender);
 
     super(
       {
         ...data,
-        v: data.v ?? 27,
-        r: data.r ?? fakeSignature.readUInt32LE(),
-        s: data.s ?? fakeSignature.readUInt32LE(4),
+        v: data.v ?? fakeSignature.v,
```
fvictorio, line 106: Just a reminder that we should use the current value here.

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
```diff
@@ -0,0 +1,23 @@
...
+// Produces a signature with r and s values taken from a hash of the inputs.
+export function makeFakeSignature(...inputs: any[]): {
+  v: number;
+  r: number;
+  s: number;
+} {
+  const hash = crypto.createHash("md5");
+
+  for (const input of inputs) {
+    hash.update(Buffer.from(`${util.inspect(input)}`));
```
fvictorio, line 13: This part concerns me a bit, because the `util.inspect` function could return a different value after, say, installing a library that patches the inspect function of some type. I don't know how likely this is to happen, but I think we should avoid it just in case.

The alternative that comes to mind is to just use the sender address and the nonce of the tx here, since those should be unique enough.

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
```diff
@@ -0,0 +1,23 @@
+import crypto from "crypto";
+import util from "util";
+
+// Produces a signature with r and s values taken from a hash of the inputs.
+export function makeFakeSignature(...inputs: any[]): {
+  v: number;
+  r: number;
+  s: number;
+} {
+  const hash = crypto.createHash("md5");
```
alcuadrado, line 10: https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/util/hash.ts#L10

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
alcuadrado: > The alternative that comes to mind is to just use the sender address and the nonce of the tx here, since those should be unique enough.

This can be problematic when you want to replace a pending tx.


The problem here is that `any[]`. I think we can make this type-safe and use appropriate serializations.

packages/hardhat-core/src/internal/hardhat-network/provider/transactions/FakeSenderTransaction.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3175/commits/524af2d7e770ad3663d1660a55b309ac175cb253

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3175/commits/a4118e0f410b44697d5da3a3c8fcbff0acbe8a43

But did I address it entirely? I made it type-safe, but I wasn't sure about the "and use appropriate serializations" part.

.changeset/fresh-forks-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": minor
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3175/commits/6895f01750166147e76951bf81470bdc5c0d7c93

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
feuGeneA: I updated this to use the existing function, in https://github.com/NomicFoundation/hardhat/pull/3175/commits/9d048a68a794f745dea8d10828b0a5787b7eff19

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
```diff
@@ -0,0 +1,30 @@
...
+export function makeFakeSignature(
+  tx: TxData | AccessListEIP2930Transaction | FeeMarketEIP1559Transaction,
+  sender: Address
+): {
+  v: number;
+  r: number;
+  s: number;
+} {
+  const hashDigest = createNonCryptographicHashBasedIdentifier(
+    Buffer.from(`${util.inspect(sender)}${util.inspect(tx)}`)
```
fvictorio, line 22: Instead of `utill.inspect` we should use a proper list of values. This list is slightly different for each transaction type, so I'm not sure if we should have a separate method for each, or just receive an array of `any`s.

feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3175/commits/2c7dfced2963b37662a5d1d6443ddf1234b798f3

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
```diff
@@ -18,8 +16,24 @@ export function makeFakeSignature(
...
+    tx.nonce,
+    tx.gasLimit,
+    tx.value,
+    tx.to,
+    tx.data,
+    "gasPrice" in tx ? tx.gasPrice : "",
+    "chainId" in tx ? tx.chainId : "",
+    "maxPriorityFeePerGas" in tx ? tx.maxPriorityFeePerGas : "",
+    "maxFeePerGas" in tx ? tx.maxFeePerGas : "",
+    // note: accessList omitted because it doesn't serialize readily.
```
feuGeneA, line 29: Add access list.

feuGeneA (commented):
Comments from Pato generated during a call.

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
feuGeneA: And use separate separators for each of the nested pieces of the access list, so we can differentiate them in the serialization. And add an explicit separator to that existing `join` so that it's clear to the reader.

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
```diff
@@ -0,0 +1,46 @@
...
+    tx.value,
+    tx.to,
+    tx.data,
+    "gasPrice" in tx ? tx.gasPrice : "",
+    "chainId" in tx ? tx.chainId : "",
+    "maxPriorityFeePerGas" in tx ? tx.maxPriorityFeePerGas : "",
+    "maxFeePerGas" in tx ? tx.maxFeePerGas : "",
+    "accessList" in tx
+      ? tx.accessList?.map(([buf, bufs]) =>
+          [buf, ...bufs].map((b) => b.toString()).join(";")
```
fvictorio, line 31:
```suggestion
          [buf, ...bufs].map((b) => b.toString("hex")).join(";")
```

Otherwise you get things like this:

```
> Buffer.from([220, 221]).toString()
'��'
```

fvictorio (approved):
Pre-approving, but I think this will need an approval from @alcuadrado too.

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
feuGeneA: Great call, thank you. Note that we're also getting something similar for `tx.data` (similar to `'��'`); however, I tried adding `?.toString("hex")::""` to the end of `tx.data`, but it seems that `BufferLike` doesn't accept that arg the way `Buffer` does.

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3175/commits/39edd2d93bb34742b0db197326c64bb2f3eebd24

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeFakeSignature.ts
fvictorio: I think we should avoid the `.map((a) => a?.toString() ?? "")`. I would try doing something like this:

```js
const hashInputs: Buffer[] = [
  // buffers
]

const hashInputsString = hashInputs
  .map(x => x.toString("hex"))
  .join(",")
```

This means not using a different separator for access lists... but that's probably fine.

PR #3177, opened 2022-09-14
Make eth_getStorageAt more permissive

### LINKED ISSUE #3164: eth_getStorageAt uint256 handling and QUANTITY handling in general
The merge of #2581 is causing problems when using the Foundry project's scripting features (see for example: https://github.com/gakonst/ethers-rs/pull/1687). Meanwhile, the workaround of using an earlier hardhat version (e.g. 2.9.4) also doesn't work due to possibly related problems:

```
eth_feeHistory

  Remote node did not answer to eth_feeHistory correctly

eth_feeHistory

  Errors encountered in param 0: Invalid value 10 supplied to : QUANTITY
```
I say "possibly related" because the discussion in #2581  and #2230 seem to reference how hardhat handles quantity values, and whether leading zeros etc should be allowed. The root cause of the https://github.com/gakonst/ethers-rs/pull/1687 incompatibility appears to be an incorrect spec referenced in #2230 (closed) that indicated a 64 digit hext string was required. I've provided details in #2581 as to why hardhat should probably accept variable length hex strings, as supported by the current open rpc spec (which says {0,64} ), and the go-ethereum implementation (which does not force a length either).

I'm opening this new issue because the discussion in #2581 is on an already merged PR and #2230 is already closed. I've also reported to the ethereum open rpc spec project that {0,64} should probably be {1,64} in their spec. Based on the above further error with eth_feeHistory called from foundry's scripting feature, I'm concerned that QUANTITY handling in hardhat or foundry and/or in the specs may also be an issue.

fvictorio:
Fixes #3164.

### DIFF
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
@@ -49,23 +49,37 @@ function validateStorageSlot(u: unknown, c: t.Context): t.Validation<bigint> {
     );
   }
 
-  if (u.match(/^0x(?:[0-9a-fA-F]*)*$/) === null) {
+  if (u === "") {
+    return t.failure(u, c, "Storage slot argument cannot be an empty string");
+  }
+
+  if (u.startsWith("0x")) {
+    if (u.length > 66) {
+      return t.failure(
+        u,
+        c,
+        `Storage slot argument must have a length of at most 66 ("0x" + 32 bytes), but '${u}' has a length of ${u.length}`
+      );
+    }
+  } else {
+    if (u.length > 64) {
+      return t.failure(
+        u,
+        c,
+        `Storage slot argument must have a length of at most 64 (32 bytes), but '${u}' has a length of ${u.length}`
+      );
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
```diff
@@ -49,23 +49,37 @@ function validateStorageSlot(u: unknown, c: t.Context): t.Validation<bigint> {
...
     return t.failure(
       u,
       c,
-      `Storage slot argument must have a length of 66 ("0x" + 32 bytes), but '${u}' has a length of ${u.length}`
+      `Storage slot argument must be a valid hexadecimal, got '${u}'`
     );
   }
 
-  return t.success(BigInt(u));
+  return t.success(u === "0x" ? 0n : BigInt(u));
```
alcuadrado, line 82: This breaks when `u` doesn't start with `"0x"`.

alcuadrado, line 82:
```suggestion
  return t.success(u === "0x" ? 0n : BigInt(u.startsWith("0x") ? u : `0x${u}`));
```

PR #3186, opened 2022-09-15
Stack traces tests improvements

fvictorio:
Some improvements for the stack traces tests:

- Make testing all solc versions opt-in instead of opt-out. This makes tests take less time locally and in ome workflows (like the pre-release one) that don't need to test all versions.
- Use native solc instead of solcjs
- Put all the artifacts in a common directory instead of having a different one for each test.

### DIFF
--- a/packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
+++ b/packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
@@ -887,7 +887,15 @@ describe("Stack traces", function () {
   }
 
   before("Download solcjs binaries", async function () {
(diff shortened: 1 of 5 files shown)


packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
```diff
@@ -887,12 +889,20 @@ describe("Stack traces", function () {
   }
 
   before("Download solcjs binaries", async function () {
```
alcuadrado, line 891:
```suggestion
  before("Download solc binaries", async function () {
```

alcuadrado (approved):

PR #3189, opened 2022-09-16
Fix two `solidity-analyzer`-related errors

### LINKED ISSUE #3176: v2.11.1-MODULE_NOT_FOUND
#### Win 10 - Node v18.9.0
***Issue: '@nomicfoundation\solidity-analyzer-win32-x64-msvc' Module_Not_Found error present upon a clean install of the hardhat v2.11.1 even though the module is present within node_modules (seems to be a compiling issue)***

***Workaround: Install hardhat v2.10.2 and compile using the previous version first, then proceed to update to v2.11.1 for testing***
> $ npm i hardhat@2.10.2
> $ npx hardhat compile
> $ npm i hardhat@2.11.1
> $ npx hardhat test

#### v2.11.1 ERROR:
hardhat_2 11 1_issue

#### After v2.10.2 Workaround:
hardhat_2 11 1_workaround

alcuadrado:
This small PR fixes two errors related to `solidity-analyzer`:

1. It bumps it to `0.1.0`. As we were using `^` and `0.0.x`, Hardhat wasn't getting any dependency update.
2. It detects NPM's issue #4828 and prints a clear error message explaining how to fix it.

Fixes #3176

### DIFF
--- a/packages/hardhat-core/src/internal/core/errors-list.ts
+++ b/packages/hardhat-core/src/internal/core/errors-list.ts
@@ -245,7 +245,7 @@ Please delete your node_modules, package-lock.json, reinstall your project, and 
       title: "Corrupted lockfile",
       description: `Some versions of NPM are affected [by a bug](https://github.com/npm/cli/issues/4828) that leads to corrupt lockfiles being generated.
 
-This bug can only affect you if you are installing your project without a lockfile, but with an existing node_modules.
+This bug can only affect you if you, or someone at your team, installed the project without a lockfile, but with an existing node_modules.
 
 To avoid it, please delete both your node_modules and package-lock.json, and reinstall your project.
 
--- a/packages/hardhat-core/src/internal/core/errors-list.ts
+++ b/packages/hardhat-core/src/internal/core/errors-list.ts
@@ -237,6 +237,21 @@ Either try using a new directory name, or remove the conflicting files.`,
         "Hardhat attempted to convert the input value to a BigInt, but no known conversion method was applicable to the given value.",
       shouldBeReported: false,
     },
+    CORRUPTED_LOCKFILE: {
+      number: 18,
+      message: `You installed Hardhat with a corrupted lockfile due to the NPM bug #4828.
+
+Please delete your node_modules, package-lock.json, reinstall your project, and try again.`,
+      title: "Corrupted lockfile",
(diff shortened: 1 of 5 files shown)


packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -237,6 +237,21 @@ Either try using a new directory name, or remove the conflicting files.`,
         "Hardhat attempted to convert the input value to a BigInt, but no known conversion method was applicable to the given value.",
       shouldBeReported: false,
     },
+    CORRUPTED_LOCKFILE: {
+      number: 18,
+      message: `You installed Hardhat with a corrupted lockfile due to the NPM bug #4828.
+
+Please delete your node_modules, package-lock.json, reinstall your project, and try again.`,
```
fvictorio, line 244:
```suggestion
Please delete your node_modules, lockfile (package-lock.json or yarn.lock), reinstall your project, and try again.`,
```

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -237,6 +237,21 @@ Either try using a new directory name, or remove the conflicting files.`,
...
+      title: "Corrupted lockfile",
+      description: `Some versions of NPM are affected [by a bug](https://github.com/npm/cli/issues/4828) that leads to corrupt lockfiles being generated.
+
+This bug can only affect you if you are installing your project without a lockfile, but with an existing node_modules.
```
fvictorio, line 248: Maybe I'm missing something here, but I think this is misleading, because this is what the person that generated the corrupted lockfile did, not what the person finding the bug did.

So I guess the message should be something like: "This bug can happen if one of your teammates generated the lockfile using a different operating system or architecture."

packages/hardhat-core/src/internal/core/errors-list.ts
fvictorio: Now that I think about it, if this bug is only present in npm and not in yarn, then the current message is correct.

packages/hardhat-core/src/internal/core/errors-list.ts
alcuadrado: great point!

fvictorio (approved):

PR #3199, opened 2022-09-20
Change ethers.getSigners() to ethers.getSigner()

zeitseeing:
I think this is just a typo. 

The function to get the first account is the singular getSigner()

https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcProvider-getSigner

### DIFF
--- a/docs/src/content/tutorial/testing-contracts.md
+++ b/docs/src/content/tutorial/testing-contracts.md
@@ -17,7 +17,7 @@ const { expect } = require("chai");
 
 describe("Token contract", function () {
   it("Deployment should assign the total supply of tokens to the owner", async function () {
-    const [owner] = await ethers.getSigners();
+    const [owner] = await ethers.getSigner();
 
     const Token = await ethers.getContractFactory("Token");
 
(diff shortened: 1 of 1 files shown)


zeitseeing:
Closing this PR, as this actually seems wrong and the `getSigners()` method is right. I'm just confused, since I couldn't find `getSigners()` in the [ethers.js docs](https://docs.ethers.io/v5/api/signer/). Sorry for the noise!

Any chance you are able to clarify why `getSigner()` does not work?

fvictorio:
Ugh, I understand why this is confusing, and we want to eventually change the API to make it clearer which things are added by Hardhat and which are just part of ethers.

Ethers does have a [`getSigner`](https://docs.ethers.io/v5/single-page/#/v5/api/providers/jsonrpc-provider/-%23-JsonRpcProvider-getSigner) function, but it's in the `provider` object. This returns a signer that is "managed by the node", meaning that it's an unlocked address in the node to which the provider is connected.

Our added helper `ethers.getSigner` is pretty much the same thing. The only difference is that we return a signer with one extra `address` field, which lets you get the address of the signer syncronously. Ethers's signers, on the other hand, only return their addresses via [an async method](https://docs.ethers.io/v5/single-page/#/v5/api/signer/-%23-Signer-getaddress).

`ethers.getSigners` is also added by us, and returns an array with signers for all the accounts available in the node. These also have the type of the expanded signer with the `address` field.

Finally, we also add a `provider` object to ethers, which is a provider connected to whatever network you selected (the in-process network by default). This means that you can use `ethers.provider.getSigner`, and you'll get a "normal" ethers signer.

There are other helpers we add, but hopefully this clarifies the `getSigner`/`getSigners` thing.

zeitseeing:
@fvictorio That makes sense. I appreciate you taking the time explaining this. I shall continue on my hardhat learning journey 😃

PR #3207, opened 2022-09-22
xdai to gnosis

alebanzas:

---

xdaichain is now [Gnosis](https://www.gnosischain.com), with native integration with Etherscan ([gnosisscan.io](https://gnosisscan.io))

### DIFF
--- a/packages/hardhat-etherscan/src/ChainConfig.ts
+++ b/packages/hardhat-etherscan/src/ChainConfig.ts
@@ -163,11 +163,11 @@ export const chainConfig: ChainConfig = {
       browserURL: "https://moonbase.moonscan.io/",
     },
   },
-  xdai: {
+  gnosis: {
     chainId: 100,
     urls: {
-      apiURL: "https://blockscout.com/xdai/mainnet/api",
-      browserURL: "https://blockscout.com/xdai/mainnet",
+      apiURL: "https://gnosisscan.io/api",
+      browserURL: "https://gnosisscan.io",
     },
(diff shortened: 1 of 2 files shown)


clemsos:
hello, fwiw I am getting ```Error in plugin @nomiclabs/hardhat-etherscan: Invalid API URL endpoint, use https://api.gnosisscan.io``` when trying this

alcuadrado:
I'd consider this a breaking change. Can we have both `xdai` and `gnosis`?

alebanzas:
I will add xdai for backward compatibility.

alebanzas:
@clemsos @alcuadrado done with the changes. thanks for your comments.

fvictorio:
The CI is failing because we have a test that checks that there are no duplicate chain ids.

Since xdai/gnosis will be an explicit exception, I think the right thing to do is to reflect that in the test: first get the networks with chain id 100, assert that they are xdai and gnosis, and remove them from the array. Then we check that there are no duplicates, as we are doing now.

@alebanzas would you like to do that?

alebanzas:
@fvictorio agree. I've added the exception in a way that should be easy in the future to add more exceptions and know why is the exception.

fvictorio (approved):

fvictorio:
Ugh, I forgot to add a changeset. I pushed one to `main`: https://github.com/NomicFoundation/hardhat/commit/adb075d4253c2559f548e9892cdb3cd9bba995e4

Issue #3043, opened 2022-08-15
Initial support for ES Modules


fvictorio:
We are working on adding some support for using ES Modules in Hardhat projects. We can't provide **full** support yet because Hardhat's plugin system relies on plugins being loaded synchronously, in the CommonJS model, and ESM works in a different way.

We'll only support ESM in javascript projects, because we rely on `ts-node` for typescript projects and their support for ESM is not quite ready. For example, they rely on an experimental API of Node that is not stable and triggers some annoying warnings.

Hardhat projects that use ESM will need to have a `hardhat.config.cjs` file (a CommonJS config file) and all the files required by it will need to be CJS too. But the rest of the project, including scripts and tests, will be writable in ESM.

**You can try this right now** by installing the `esm` tag of Hardhat:

```
npm i hardhat@esm
```

If you find any problems, please leave a comment here or open an issue about it. Any feedback on this, whether if it's working or not, is super appreciated.

Todo:
- [x] Include `.cjs`/`.mjs` files when running the `test` task
- [x] Create a `.cjs` hardhat config during initialization if the projects has `type: module`
- [x] Write an advanced guide about this
- [x] Throw a proper error when a typescript project uses ESM

vladfaust:
> For example, they rely on an experimental API of Node that is not stable and triggers some annoying warnings.

Personally, I think this reason is not enough to disable TypeScript support.

fvictorio:
It's not that we are _disabling_ it, it's that we don't want to invest significant time adding support for it right now if the result might break anytime soon. But we might re-consider this in some months.

If this is something you want or need, please upvote [this issue](https://github.com/NomicFoundation/hardhat/issues/3385)

Olyno:
Why not use tools such as [tsup](https://github.com/egoist/tsup) (a detailed example available [here](https://antfu.me/posts/publish-esm-and-cjs#tsup)) that allow to build Hardhat in cjs and esm, and thus solve the problem while bringing the support to Typescript at the same time?

dbeal-eth:
> Hardhat projects that use ESM will need to have a hardhat.config.cjs file (a CommonJS config file) and all the files required by it will need to be CJS too. But the rest of the project, including scripts and tests, will be writable in ESM.

[...]

Issue #2927, opened 2022-07-12
Bug: Hardhat.config.cjs or harrdhat.config.mjs not recognized 


sambacha:
Hardhat does not recognize the format of `hardhat.config.cjs` nor `hardhat.config.mjs`

feuGeneA:
This issue has been added to our internal backlog, meaning that we intend to work on it, but that it has not yet been scheduled.

jrynkiew:
What is the status of this issue? It is critical to fix.

PR #3211, opened 2022-09-23
Basic support for ESM

fvictorio:
Closes https://github.com/NomicFoundation/hardhat/issues/3043
Closes https://github.com/NomicFoundation/hardhat/issues/3240
Closes https://github.com/NomicFoundation/hardhat/issues/2927

This PR:

- Adds support for having scripts and tests in ESM
- Adds a new advanced guide about this
- Upgrades typescript and ts-node. This was done in a previous iteration, when we wanted to have support for ESM in TypeScript projects. We decided we won't do that, but I think we can keep the upgraded packages.
- Makes some changes to the project initialization:
  - If the project has a `package.json` with `"type": "module"`, the options shown are a bit different
  - There's a new `javascript-esm` sample project
  - I also made some changes to how we handle the project initialization if the directory already has files. Basically: we skip the `README.md` file if there's already one, and we properly check for clashing _files_, not directories.
- There's also a small change in the Toolbox to handle Hardhat config files with the `.cjs` extension.


fvictorio:
Closing in favor of https://github.com/NomicFoundation/hardhat/pull/3550

PR #3218, opened 2022-09-27
Include task name in HH306 (MISSING_TASK_ARGUMENT)

feuGeneA:
for https://linear.app/nomic-foundation/issue/HH-1152/add-task-name-to-hh306

### DIFF
--- a/packages/hardhat-core/src/builtin-tasks/flatten.ts
+++ b/packages/hardhat-core/src/builtin-tasks/flatten.ts
@@ -43,7 +43,7 @@ function getSortedFiles(dependenciesGraph: DependencyGraph) {
   } catch (error) {
     if (error instanceof Error) {
       if (error.toString().includes("Error: There is a cycle in the graph.")) {
-        throw new HardhatError(ERRORS.BUILTIN_TASKS.FLATTEN_CYCLE, error);
+        throw new HardhatError(ERRORS.BUILTIN_TASKS.FLATTEN_CYCLE, {}, error);
       }
     }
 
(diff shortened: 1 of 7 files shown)


fvictorio:
Sorry for the unrequested review, but I took a look before I realized it was a draft PR :sweat:

I see two (minor) issues with this:

1. If you do `npx hardhat --config`, the message will say `The '--config' parameter of task 'undefined' expects a value, but none was passed.`
2. If you do `npx hardhat compile --config`, the message will say that the `--config` param belongs to the compile task, but it's a global param.

I think we can ignore number 2. It doesn't seem easy to fix, and I don't think it's super important.

With respect to 1, my main concern was that this was even possible. The reason is that, while we added a linter rule to prevent `undefined` values in string interpolation, the ad-hoc string interpolation for Hardhat errors is a different thing. Luckily, fixing it was easy. In the commit I pushed, the type for the values of the `messageArguments` object is now `string | number` instead of `any`. This also uncovered a minor error in the flatten task.

The code won't compile now because `taskName` can be undefined in one of the error usages. It's not obvious how to fix this, but two approaches that come to mind are:

1. Create a new error that doesn't include the part of the sentence that shows the task name, and conditionally throw that one if `taskName` is undefined.
2. Use `taskName ?? "help"`. This kind of works because, when no task name is passed, Hardhat shows the help output.

I don't have a strong preference here. I would probably go with 2 because, while the first approach is slightly better, I don't feel it justifies adding a new error to the list.

fvictorio (approved):

PR #3225, opened 2022-09-29
Correct a small grammatical mistake

hrishibhattu:
- [x] There's no code involved.

---

Just correcting a small grammatical mistake that I noticed while going through the tutorial.


fvictorio:
Thank you!

PR #3228, opened 2022-09-30
Add option to allow mining blocks with the same timestamp

### LINKED ISSUE #2560: It's not possible to mine multiple blocks with the same timestamp

### LINKED ISSUE #3045: Add support for blocks with the same timestamp
There are many reasons to support this, even if it's not consensus-valid. See for example the discussion in #2504.

The way we'll approach this is by adding a new option in the Hardhat network configuration (name TBD) that defaults to false to preserve the current behavior. When this option is enabled:

- Two (or more) blocks mined within the same second will have the same timestamp. That is, we won't add 1 second to generate a different timestamp.
- Some JSON-RPC methods will behave differently: `hardhat_mine` will accept an interval of 0, `evm_setNextBlockTimestamp` will accept the same timestamp as the one from the current block, etc.
- We might need to revisit our assertions in the Network Helpers. We should stop validating timestamp-related things (assuming we are doing it) and just let the hardhat errors bubble up if something fails. We could also capture them and show a different error message, but if there is a better error message that should be done in the hardhat network, not in the helpers.

fvictorio:
This PR adds a new `allowBlocksWithSameTimestamp` option to Hardhat Network that doesn't force the timestamp of new blocks to be different than the last one.

The first commit in this PR should be reviewed on its own to make things simpler. It's a refactor that I think it was long due: it changes the constructor of `HardhatNetworkProvider` so that it receives most of its parameters in an object. This makes the code easier to read and change in several places. I also removed the "network name" parameter, since it wasn't used anywhere.

The rest of the changes adds the new option, modifies the network helpers to not check the timestamp (instead letting Hardhat itself throw an error if there's a wrong argument), update the docs and add tests.

Closes #2560.
Closes #3045.

### DIFF
--- a/packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts
+++ b/packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts
@@ -282,14 +282,14 @@ describe("Evm module", function () {
                 [timestamp - 1],
                 `Timestamp ${
                   timestamp - 1
-                } is lower than or equal to previous block's timestamp ${timestamp}`
+                } is lower than the previous block's timestamp ${timestamp}`
               );
 
               await assertInvalidInputError(
                 this.provider,
                 "evm_setNextBlockTimestamp",
                 [timestamp],
-                `Timestamp ${timestamp} is lower than or equal to previous block's timestamp ${timestamp}`
+                `Timestamp ${timestamp} is equal to the previous block's timestamp`
               );
             });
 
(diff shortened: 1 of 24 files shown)


packages/hardhat-core/test/internal/hardhat-network/provider/modules/evm.ts
```diff
@@ -1437,5 +1437,41 @@ describe("Evm module", function () {
         });
       });
     });
+
+    describe(`${name} provider (allowBlocksWithSameTimestamp)`, function () {
+      setCWD();
+      useProvider({ allowBlocksWithSameTimestamp: true });
```
alcuadrado, line 1443: This is a bit nitpicky, but i think we should have another test that mines two blocks in a row, without manually altering the time, and they should get the same timestamp.

The reason I think this is important, is that this is the most common behavioral change that this pr introduces.

alcuadrado: Please ignore everything I said. This is already present in a different file.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -601,6 +601,54 @@ describe("HardhatNode", () => {
...
+        for (let i = 0; i + 1 < timestamps.length; i++) {
+          if (timestamps[i] !== timestamps[i + 1]) {
+            differentTimestamps++;
+          }
+        }
+
+        // There is a small chance that two of the mined blocks were mined in a
+        // different second.
+        // This number shouldn't be bigger than 1 if we are running these tests
+        // on a computer from this century.
```
alcuadrado, line 628: If we expect hardhat to be here for the long run we should clarify which century is `this century`.

PR #3238, opened 2022-10-04
add synthetix CC plugins to community list

dbeal-eth:
Synthetix has had these very well maintained and portable hardhat plugins for a while now, and its probably time we actually put them on the official "community plugins" list :laughing:


dbeal-eth:
is it allowed to make the link at the end of the `hardhat-cannon` description clickable?

fvictorio:
> is it allowed (or technically feasible) to make the link at the end of the `hardhat-cannon` description clickable?

I don't think that's supported right now, sorry :disappointed: It seems like a useful thing to have though, so I'll add it to our website backlog.

Issue #3223, opened 2022-09-28
Private keys can be derived even if the mnemonic has line breaks


winor30:
Even if a newline character is added to the correct mnemonic, `deriveKeyFromMnemonicAndPath` can still derive private keys.

Following the test (`Should not derive the wrong keys that are added '\\n’`) should be successful. but `deriveKeyFromMnemonicAndPath` returns private key and the test is failed.

https://github.com/NomicFoundation/hardhat/compare/main...winor30:hardhat:fix/add-newline-char-mnemonic-validation?expand=1

fvictorio:
To clarify, you are expecting that adding a newline at the end of the mnemonic should throw an error?

I think this is just us (or the library we are using) trimming the string before deriving the private keys. This doesn't seem super problematic to me, but I could be wrong, can you ellaborate on why this is an issue?

winor30:
thank you for your reply.
I found this problematic because when I used hardhat, I was able to deploy at an address I did not expect when I did a deploy using mnemonic with \n in it.

I felt that if the error occurred when mnemonic with the \n was used, then I would not have to use an unwanted address.

However, if it is not HARDHAT's responsibility to remove the \n, then this may not be a problem.

```ts
const mnemonicFileName = process.env.MNEMONIC_FILE;
// this mnemonic is inclueded \n
const mnemonic = fs.readFileSync(mnemonicFileName, 'ascii')

const config: HardhatUserConfig = {
  ...
  networks: {
    dev: { url: 'http://localhost:8545', accounts: { mnemonic } },
...
  },
```

fvictorio:
Oh wait, I see what you mean now. This:

```js
module.exports = {
  networks: {
    hardhat: {
      accounts: {
        mnemonic: `test test test test test test test test test test test junk`
      }
    },
  },
};
```

produces a different set of keys than this (that ends on a newline):

```
module.exports = {
  networks: {
    hardhat: {
      accounts: {
        mnemonic: `test test test test test test test test test test test junk
`
      }
    },
  },
};
```

I do agree this is problematic. At the very least we should be trimming the mnemonic (which I thought we were doing, but we aren't).

Thanks for reporting it!

fvictorio:
Btw, if you are interested in sending a PR that fixes this, I think it should include:

1. Trimming the mnemonic (probably in `src/internal/util/keys-derivation.ts`, although you could argue that it should be done in the config resolution... maybe both places should do it)
[...]

PR #3246, opened 2022-10-05
possible to derive private key from mnemonic included space character

winor30:
- [X] Because this PR includes a **bug fix**, relevant tests have been included.

---

Fixes https://github.com/NomicFoundation/hardhat/issues/3223

I fixed the logic to derive a private key from the mnemonic that included newline character or space character.
If a newline or space character is included, it is trimmed.


fvictorio:
Thank you @winor30!

PR #3250, opened 2022-10-05
Patches for viaIR

jflatow:
This is a continuation of #2399 and #2400, rebased onto recent `main` (v2.11.2) and including another patch for inferring errors. Opening this PR in case its useful for others, as we are currently relying on it.


fvictorio:
> we are currently relying on it

Hey, I'm working on this on https://github.com/NomicFoundation/hardhat/pull/3205. I think I'll have something finished next week, but if you can give it a try, it would be super useful.

fvictorio:
Hi @jflatow, I just published a version with support for IR under the `ir` npm tag. You can try it out with `npm install hardhat@ir`.

fvictorio:
> Using this patch I still see the 'incorrect parameters' errors

Yes, I didn't manage to fix that one without breaking the error inference for non-IR compilation. I might give it another try, but it's kind of hard without proper support from the solc compiler.

alcuadrado:
Closing in favor of #3415

PR #3253, opened 2022-10-06
Update plugins.ts

TovarishFin:
When trying to use yul, I found that the plugin no longer worked. I did a fairly large rewrite and made a few improvements on top of it. Please consider using my fork instead of the original.

---

### DIFF
--- a/docs/src/content/hardhat-runner/plugins/plugins.ts
+++ b/docs/src/content/hardhat-runner/plugins/plugins.ts
@@ -299,11 +299,11 @@ const communityPlugins: IPlugin[] = [
     tags: ["Deployment", "CREATE2", "Tasks"],
   },
   {
-    name: "@controlcpluscontrolv/hardhat-yul",
-    author: "ControlCplusControlV",
-    authorUrl: "https://github.com/controlCplusControlV/",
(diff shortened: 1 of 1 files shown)


fvictorio:
Hey @TovarishFin, it seems like the upstream project merged a PR from you, do you still want to add your fork to the list?

TovarishFin:
He wanted to chat with me on telegram but never wrote me back... not sure what is going on with him. regardless this is a bit more than i signed up for. Do whatever you think is the best course. 

I would submit one piece of info favoring my fork though. His version never had an NPM package. Mine does. And it matches my name. If the plugin uses his, it will still end up using my NPM package since he just pulled my changes (including the docs). 

Anyways do whatever you want. It doesn't matter much to me either way :)

Either way I will continue to update and use my own fork since I am using it daily.

fvictorio:
In that case I don't mind adding it as a different plugin and saying in the description that it's a more up-to-date fork. We can consider replacing it entirely in the future. I would understand if you don't want to invest more time in this PR though.

TovarishFin:
Alright so I updated it to leave the original and just add my fork. Hopefully this works for you.

fvictorio:
Thank you!

PR #3259, opened 2022-10-08
Replace ropsten and rinkeby with goerli

fvictorio:
Replace ropsten and rinkeby with goerli in our tests.

Since Alchemy is no longer working with these testnets, our tests are failin. These networks are disappearing anyway, so we should make this change.


Wodann (approved):
The changes lgtm, although I can't judge whether you might have missed something based on the diff. A Ctrl+F for the old test nets should suffice :)

PR #3260, opened 2022-10-09
Add ignoreUnknownTxType for forking Arbitrum One

coreycaplan3:

---

This PR adds `ignoreUnknownTxType` for forking Arbitrum One and other networks that have custom tx types. Normally, when forking a live network, getting transactions from Arbitrum One, since the Nitro upgrade, fails because Nitro uses non-standard transaction types. This results in tests that use "mainnet forking" failing.

To fix this problem, this PR adds the `ignoreUnknownTxType` param to the commandline or hardhat config file in the `forking` object. In doing so, developers can optionally allow hardhat to ignore any unknown transaction types, so forking can continue to work normally.


fvictorio:
I'm not sure why our automation closed this, but for the record: we'll include a fix for this (that doesn't need an option) in the next release.

Issue #3200, opened 2022-09-21
We couldn't initialize the sample project because this file already exists ERROR


BerkKaanVAN:
hi, i am having this error in hardhat v2.11.2 version. I get this warning after npx hardhat -> create a javascript project " We couldn't initialize the sample project because this file already exists: README.md" "Please delete or move them and try again."

H20119:
> hi, i am having this error in hardhat v2.11.2 version. I get this warning after npx hardhat -> create a javascript project " We couldn't initialize the sample project because this file already exists: README.md" "Please delete or move them and try again."



> hi, i am having this error in hardhat v2.11.2 version. I get this warning after npx hardhat -> create a javascript project " We couldn't initialize the sample project because this file already exists: README.md" "Please delete or move them and try again."

having the same issue, did you find a solution yet ?

H20119:
having the same issue, did you find a solution yet ?

H20119:
having the same issue, did you find a solution yet ?

fvictorio:
The project initialization tries to avoid overwriting files, and so it stops if there are existing files that clash with the ones that Hardhat would create. In this case, you can do something like this:

```
mv README.md README.md.bak
npx hardhat # initialize project
rm README.md # remove the readme that Hardhat created
mv README.md.bak README.md
```

The project initialization could be smarter here though, but we already have an issue open about that, so I'm going to close this one.

Issue #3240, opened 2022-10-04
"We couldn't initialize the sample project because this file already exists: README.md"


ZahraaHawie:
I'm working on a react app and I want to connect it to a smart contract. Upon hardhat initialization I got this error: 

"We couldn't initialize the sample project because this file already exists: README.md"

15802841343:
+1 I got the same problem

alcuadrado:
Fixed in #3211

PR #3269, opened 2022-10-12
Improve project creation file management

tmigone:
Fixes problems with project creation (#3200  , #3240) by adding a list of files that should not be copied over from the sample project, and another list of files that should not be overwritten if they already exist. Didn't see any tests related to project creation so did not include any for this.

Closes: #3200, #3240


alcuadrado:
Sorry we took so long to reply, and thanks for submitting this issue.

This was already fixed in #3211

PR #3275, opened 2022-10-14
hardhat-foundry plugin

antico5:
Plugin features:
- Set sources path to match the one from foundry
- Avoid name clashing on cache folders by both frameworks
- Parse and use remappings on import resolution
- Task for initializing foundry on an existing hardhat project

Core changes:
- Introduce a subtask for transforming import names passed to `Resolver.resolveImport()`
- Add `settings.remappings` to CompilerInput type

To-do:
- [ ] Change peer dependency of the plugin on Hardhat


packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+          `Command failed. Please continue forge-std installation manually.`
+        )
+      );
+      console.log(error);
+    }
+  }
+);
+
+function isCompilationJobCreationError(
+  x: CompilationJob | CompilationJobCreationError | SolcConfig
```
fvictorio, line 176: The `SolcConfig` type is not necessary here. (I know it's included in the same function from core, but here it's not used.)

packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+      | CompilationJobCreationError;
+
+    if (!pluginActivated || isCompilationJobCreationError(job)) {
+      return job;
+    }
+
+    const remappings = getRemappings();
+    job.getSolcConfig().settings.remappings = Object.entries(remappings).map(
+      (fromTo) => fromTo.join("=")
+    );
```
fvictorio, line 125: Personal preference:

```suggestion
    job.getSolcConfig().settings.remappings = Object.entries(remappings).map(
      ([from, to]) => `${from}=${to}`
    );
```

Two reasons: first, I think destructuring the key/value of `Object.entries` is common practice. Second, `.join` suggests to me an array of arbitrary length, so the meaning of the code is slightly less obvious.

packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+import { HardhatFoundryError } from "./errors";
+
+const TASK_INIT_FOUNDRY = "init-foundry";
+
+let pluginActivated = false;
+
+extendConfig(async (config, userConfig) => {
+  // Check foundry.toml presence. Don't warn when running foundry initialization task
+  if (!existsSync(path.join(config.paths.root, "foundry.toml"))) {
+    if (!process.argv.includes(TASK_INIT_FOUNDRY)) {
```
fvictorio, line 30: Ugh, this is unfortunate, but I can't think of a better alternative.

packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+    throw new HardhatFoundryError(
+      `User-configured sources path (${userSourcesPath}) doesn't match path configured in foundry (${foundrySourcesPath})`
+    );
+  }
+
+  // Set sources path
+  config.paths.sources = foundryConfig.src;
+
+  // Change hardhat's cache path if it clashes with foundry's
+  if (foundryConfig.cache_path === "cache") {
```
fvictorio, line 67: I think we should actually check if the paths clash, just to be super sure. I would do something like this:

```ts
  const foundryCachePath = path.resolve(
    config.paths.root,
    foundryConfig.cache_path
  );
  if (config.paths.cache === foundryCachePath) {
    config.paths.cache = "cache_hardhat";
  }
```

(This actually has a minor bug, where the paths will still clash if both Hardhat and Foundry have set its path to `"cache_hardhat"`, but let's assume that that's unlikely enough, because handling it would be awful.)

packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+    const remappings = getRemappings();
+
+    for (const from in remappings) {
+      if (Object.prototype.hasOwnProperty.call(remappings, from)) {
+        const to = remappings[from];
+        if (importName.startsWith(from)) {
+          return importName.replace(from, to);
+        }
+      }
+    }
```
fvictorio, line 94:
```suggestion
    for (const [from, to] of Object.entries(remappings)) {
      if (importName.startsWith(from)) {
        return importName.replace(from, to);
      }
    }
```

fvictorio: (The reason you were getting the linter warning was because you were using `for .. in` instead of `for .. of`.)

packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+      console.log(`File foundry.toml already exists. Aborting.`);
+      return;
+    }
+
+    console.log(`Creating foundry.toml file...`);
+
+    writeFileSync(
+      "foundry.toml",
+      [
+        `[profile.default]`,
```
fvictorio, line 145: I think this would look better if we use a multi-line string and interpolation? Something like:

```
writeFileSync(
  "foundry.toml",
  `
[profile.default]
src = ${foundrySrc}
out = 'out'
```

etc.

I'm not 100% sure though.

packages/hardhat-foundry/src/errors.ts
```diff
@@ -0,0 +1,28 @@
...
+        return new HardhatFoundryError(
+          "Couldn't run `forge`. Please check that your foundry installation is correct."
+        );
+      case 134:
+        return new HardhatFoundryError(
+          "Running `forge` failed. Please check that your foundry.toml file is correct."
+        );
+      default:
+        return new HardhatFoundryError(
+          `Unexpected error while running \`forge\`:${error.stderr.toString()}`
```
fvictorio, line 22:
```suggestion
          `Unexpected error while running \`forge\`: ${error.stderr.toString()}`
```

packages/hardhat-foundry/src/index.ts
antico5: Normally I don't use that approach because it breaks indentation, but it's just personal preference. Your call

fvictorio: I don't have a strong preference, let's keep it as it is.

packages/hardhat-foundry/src/getRemappings.ts
```diff
@@ -0,0 +1,33 @@
...
+  // Return remappings if they were already loaded
+  if (remappings !== undefined) {
+    return remappings;
+  }
+
+  // Get remappings from foundry
+  const remappingsTxt = runCmdSync("forge remappings");
+
+  remappings = {};
+  const remappingLines = remappingsTxt.split("\n");
```
fvictorio, line 15: I'm not super sure, but I think this should be:

```suggestion
  const remappingLines = remappingsTxt.split(/\r|\n|\r\n/);
```

fvictorio:
Some other comments:

- I'm not convinced about the name and semantics of the new subtask. Especially the comment, which says "It allows to translate a given import name into a source name". I don't think that's accurate. Maybe we should just say that it "maps" the content of the imports?
  - Related to this: should we only call this function of the resolver for non-relative imports? I'm not sure.
- I think we shouldn't separate the `errors.ts` and `runCmd.ts` modules. `runCmdSync` seems like a generic function but the way it maps errors to plugin errors is very specific to running forge commands. So maybe it would be better to have something more explicit? One function for running (and parsing) `forge config --json`, with its own error handling, a different one for remappings, another one for installing a dependency. I believe this will both improve the existing code in `src/index.ts` and make the overall design a bit better.
- When running `forge install`, we are printing an error message and then logging the error. Instead, we should use a specific custom error for this scenario and pass the execution error as a parent error. Something like [this](https://github.com/NomicFoundation/hardhat/blob/a414c5725ccbfd7d751f3777a516ec4d86228587/packages/hardhat-network-helpers/src/errors.ts#L15-L24).
- The error you get when there is a typo in a remapped import is not ideal. I'm not sure what's the right solution here. Maybe we should have a specific error in the resolver when the import is not found and `imported !== importName`.

packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+import chalk from "chalk";
+import { getRemappings } from "./getRemappings";
+import { runCmdSync } from "./runCmd";
+import { HardhatFoundryError } from "./errors";
+
+const TASK_INIT_FOUNDRY = "init-foundry";
+
+let pluginActivated = false;
+
+extendConfig(async (config, userConfig) => {
```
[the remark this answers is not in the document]
antico5: Good catch, the async keyword was left there by mistake.

packages/hardhat-foundry/src/getRemappings.ts
antico5: Good catch, I moved the `\r\n` to the beginning of the regex to match before just `\r`

packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,179 @@
...
+import { existsSync, writeFileSync } from "fs";
+import path from "path";
+import chalk from "chalk";
+import { getRemappings } from "./getRemappings";
+import { runCmdSync } from "./runCmd";
+import { HardhatFoundryError } from "./errors";
+
+const TASK_INIT_FOUNDRY = "init-foundry";
+
+let pluginActivated = false;
```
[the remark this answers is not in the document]
fvictorio: Oh I see, I remember now.

> Is there a more appropiate pattern I should use for this ?

No, I think this is fine.

docs/src/content/hardhat-runner/docs/advanced/hardhat-and-foundry.md
```diff
@@ -0,0 +1,103 @@
+# Integrating with Foundry
+
+This guide explains how to use Hardhat and Foundry in the same project.
```
alcuadrado, line 3: This guide doesn't explain what the plugin does.

What's the expected end result here?

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -157,6 +158,18 @@ subtask(TASK_COMPILE_SOLIDITY_READ_FILE)
     }
   );
 
+/**
+ * This task is meant to be overriden by potential plugins
+ * It allows to transform an import name
+ */
+subtask(TASK_COMPILE_TRANSFORM_IMPORT_NAME)
+  .addParam("importName", undefined, undefined, types.string)
```
alcuadrado, line 166: Is importName a sourceName? Or is it the literal that's included in the import construct?

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -157,6 +158,18 @@ subtask(TASK_COMPILE_SOLIDITY_READ_FILE)
     }
   );
 
+/**
+ * This task is meant to be overriden by potential plugins
+ * It allows to transform an import name
```
alcuadrado, line 163:
```suggestion
 * This task transform the string literal in and import directive.
 * By default it does nothing, but it can be overriden by plugins.
```

packages/hardhat-foundry/README.md
```diff
@@ -0,0 +1,7 @@
+[![npm](https://img.shields.io/npm/v/@nomicfoundation/hardhat-foundry.svg)](https://www.npmjs.com/package/@nomicfoundation/hardhat-foundry) [![hardhat](https://hardhat.org/buidler-plugin-badge.svg?1)](https://hardhat.org)
+
+# hardhat-foundry
+
+This plugin makes it easier to use Hardhat and Foundry in the same project.
```
alcuadrado, line 5: I don't think this is enough. It should explain what is the intended output/behavior of the plugin.

Note that this is what users will see if they land to the plugin page in hardhat.org or npm

packages/hardhat-foundry/DESIGN.md
```diff
@@ -0,0 +1,40 @@
+# Design
```
alcuadrado, line 1: This explanation, but with less technical details, should be in the readme and the guide

alcuadrado: Otherwise it feels like adding an unspecified blob of functionality

packages/hardhat-foundry/DESIGN.md
```diff
@@ -0,0 +1,40 @@
...
+With remappings:
+
+  foo=lib/foo/src
+  bar=lib/foo/src/bar
+  qux=lib/foo/src/qux
+```
+
+In this scenario, `Bar.sol` should be able to import `Qux.sol` in any of two ways:
+
+- `import "../qux/Qux.sol"`
```
alcuadrado, line 37: does this work in foundry?

alcuadrado: self-answer: yes

packages/hardhat-foundry/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "eslint-plugin-prettier": "3.4.0",
+    "hardhat": "^2.11.0",
+    "mocha": "^10.0.0",
+    "prettier": "2.4.1",
+    "rimraf": "^3.0.2",
+    "ts-node": "^8.1.0",
+    "typescript": "~4.5.2"
+  },
+  "peerDependencies": {
+    "hardhat": "^2.12.3"
```
alcuadrado, line 56: Reminder: this needs to be the latest version, the one that it's published with this PR

packages/hardhat-foundry/src/foundry.ts
```diff
@@ -0,0 +1,116 @@
...
+async function runCmd(cmd: string): Promise<string> {
+  try {
+    const { stdout } = await exec(cmd);
+    return stdout;
+  } catch (error: any) {
+    throw buildForgeExecutionError(error.code, error.message);
+  }
+}
+
+function buildForgeExecutionError(
```
alcuadrado, line 98: FYI: it can be more idiomatic throwing from here instead of just returning the error. It may require declaring that this function returns `never`.

I wouldn't change it though, this is fine as is

packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,176 @@
...
+    );
+  }
+
+  // Ensure foundry src path doesnt mismatch user-configured path
+  const userSourcesPath = userConfig.paths?.sources;
+  const foundrySourcesPath = foundryConfig.src;
+
+  if (userSourcesPath !== undefined && userSourcesPath !== foundrySourcesPath) {
+    throw new HardhatFoundryError(
+      `User-configured sources path (${userSourcesPath}) doesn't match path configured in foundry (${foundrySourcesPath})`
```
alcuadrado, line 61: I think this error needs a better explanation. At least a hint about what the users should do.

packages/hardhat-foundry/src/index.ts
```diff
@@ -0,0 +1,176 @@
...
+  TASK_INIT_FOUNDRY,
+  "Initialize foundry setup in current hardhat project",
+  async (_, hre: HardhatRuntimeEnvironment) => {
+    const foundryConfigPath = path.resolve(
+      hre.config.paths.root,
+      "foundry.toml"
+    );
+
+    if (existsSync(foundryConfigPath)) {
+      console.warn(chalk.yellow(`File foundry.toml already exists. Aborting.`));
```
alcuadrado, line 144: This should include info to what they should do to make sure it works ok with hardhat. Possible a link to the website.

alcuadrado: Also, I'd throw an error instead of console and process.exit

alcuadrado (commented):
Left a few comments, mostly about docs, but LGTM.

Great job, @antico5!

packages/hardhat-core/src/builtin-tasks/compile.ts
fvictorio: The string literal in the import statement.

packages/hardhat-core/src/builtin-tasks/compile.ts
fvictorio: Done in 9ada664159680661a76570bc5a720d41b4580d85 because the suggestion has a typo.

packages/hardhat-foundry/src/index.ts
fvictorio: @alcuadrado what about:

```
The sources path in your Hardhat config (...) doesn't match
the path configured in Foundry (...).

Make sure that they use the same path, or remove the custom
sources path in your Hardhat config to let the plugin handle it for you.
```

packages/hardhat-foundry/src/index.ts
fvictorio: Something like this?

```
You've run the foundry-init task, but a foundry.toml file
already exists at /path/to/project. You don't need to
do anything else to use the hardhat-foundry plugin.

Learn more at https://hardhat.org/hardhat-and-foundry
```

And that's a shortlink to the HH and Foundry guide.

packages/hardhat-foundry/README.md
fvictorio: I expanded the README a bit, let me know what you think.

docs/src/content/hardhat-runner/docs/advanced/hardhat-and-foundry.md
fvictorio: I made some improvements in f0763bd8282ac998dbb8deb577cd87c275561b39

alcuadrado (approved):

PR #3262, opened 2022-10-10
add `hardhat-txn-dot-xyz` and `hardhat-bytecode-exporter` and `hardhat-accounts` to plugin list

ItsNickBarry:
New plugin to integrate with ETHBogotá project [txn.xyz](https://www.txn.xyz/).

Also added bytecode exporter and account information output.

PR #3282, opened 2022-10-18
add solidstate plugins to list

ItsNickBarry:
Replaces #3262.

Adds two missing plugins, and three new plugins, from the `@solidstate` namespace.


fvictorio:
Thank you @ItsNickBarry, I also piggybacked on this PR another plugin that was also sent to the `website` branch.

Issue #3226, opened 2022-09-29
Hardhat dependency on local installation


arijoon:
**Issue**

Hardhat requires the real path of the module to be under the current project. This limitation could be in contradiction with some usecases. For example if building each dependency with `nix` as a derivation, hardhat module will be an absolute path defined in `pnp.js` to its derivation (or it'll be a simlink in `node_modules` if not using yarn berry)

I have found and removed [this check](https://github.com/NomicFoundation/hardhat/blob/c89d933b23d09fb14ee87bf133faaf4a8cee0989/packages/hardhat-core/src/internal/cli/cli.ts#L174-L176) for our usecase, however would like to understand the reasoning behind it existing in the first place. Once understood, I can make a PR to allow for additional methods to address any feature that requires this (e.g. if you need to find something else's path relative to hardhat module, we can allow an env variable to be passed as replacement)

fvictorio:
IIRC, it's meant to prevent people from using a globally installed version of Hardhat, since that can cause issues if you then install the plugins locally.

Your use case seems reasonable to me though. I think we could have an escape hatch for this, but it should be a somewhat advanced (and not documented) mechanism. Maybe an environment variable like `__HARDHAT_ALLOW_NON_LOCAL_INSTALLATION=true`?

@alcuadrado wdyt?

alcuadrado:
The first few versions of Buidler tried to allow global installations, which lead to users doing things like:

* install hardhat globally
* install plugins locally
* plugins `require`ing `hardhat`
* error

> Hardhat requires the real path of the module to be under the current project.

I agree that this check is too strict. Any idea on how to remove it and still rule out global installations?

arijoon:
>  agree that this check is too strict. Any idea on how to remove it and still rule out global installations?

Can do something similar to yarn v3's checks. It no longer allows a package to be required if its not defined in `package.json`. So you can read the project's package.json and ensure there is a hardhat dependency or devDependency set. 

If this is only for plugins, I would still recomment this check to be a warning or have an env var that sets the path of `package.json` in case in some edge cases it is not correctly resolved

fvictorio:
What about doing: `require.resolve("hardhat", { paths: [process.cwd()] })` to check if Hardhat is available?

fvictorio:
Oh, that's exactly what we are doing now :man_facepalming: But in that case I don't understand why it works for you after removing the check @arijoon. If you can't `require("hardhat")` then tests/scripts won't work?

[...]

PR #3283, opened 2022-10-19
Allow skipping of local installation check

arijoon:

---

Based on [this discussion ](https://github.com/NomicFoundation/hardhat/issues/3226), the check for local installation is not valid in all cases. In this PR I've simply added an env variable to skip this check. I would highly advise that the check is removed entirely as the current implementation is not correct. Having the hardhat package.json within the sub-directory of the cwd does not capture all cases of local installation. 

Additionally this is purely predicting that there'll be a future error and preventing cli from running. It might be best to allow the cli to run until an error case is hit

### DIFF
--- a/packages/hardhat-core/src/internal/cli/cli.ts
+++ b/packages/hardhat-core/src/internal/cli/cli.ts
@@ -171,7 +171,10 @@ async function main() {
       }
     }
 
-    if (!isHardhatInstalledLocallyOrLinked()) {
+    if (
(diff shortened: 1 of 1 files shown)


arijoon:
@fvictorio [victorio](https://github.com/fvictorio) can you review this please, its quiet simple and as we discussed, just an env variable to skip the check

arijoon:
Thanks a lot @fvictorio , much appreciated. I also tried to solve the issue differently by trying to read the package files but I was unable to find a different solution quickly and opted for simply skipping the check as advised, since the client can know for a fact whether they're using local or global hardhat installation.

PR #3285, opened 2022-10-19
Artifacts class support for extensions

feuGeneA:


packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -40,6 +41,24 @@ import {
 
 const log = debug("hardhat:core:artifacts");
 
+interface IHardhatArtifactSource extends ArtifactSource {
```
fvictorio, line 44: It seems like we have three options here:
- The core logic is just another ArtifactSource and we are dogfooding the interface. Initially I though this was an interesting idea, but now I'm not sure if it's easily doable. Plus, we probably want to have "special powers" here. And it's not like we are going to have dozens of implementations of the interface (unlike the task system, where the fact that we are dogfooding it does seem valuable).
- The core logic for getting the artifacts is "an ArtifactSource just like any other, but not really", which seems to be the current approach.
- The core logic is part of the Artifacts implementation itself and we don't pretend it's not special.

The three options seem sensible, but I think I'd just go with the third one just because it's straightforward and the benefits of the other more indirect options are not super clear to me.

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -867,6 +886,235 @@ Please replace "${contractName}" for the correct contract name wherever you are
   }
 }
 
+export class Artifacts implements IArtifacts {
+  private readonly _sourcesInPriorityOrder: [
+    IHardhatArtifactSource,
+    ...ArtifactSource[]
+  ];
```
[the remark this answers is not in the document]
fvictorio: Also: see my other comment about where to put the core Hardhat logic for getting artifacts. If we go with the option of having that logic as part of the `Artifacts` class (or its sub-layers) then this would become just a single variable with the array of extensions.

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -867,6 +886,235 @@ Please replace "${contractName}" for the correct contract name wherever you are
...
+    this._sourcesInPriorityOrder = [new HardhatArtifactSource(artifactsPath)];
+  }
+
+  public async readArtifact(
+    contractNameOrFullyQualifiedName: string
+  ): Promise<Artifact> {
+    const artifact = this._getFirstValueFromSources(
+      "readArtifact",
+      contractNameOrFullyQualifiedName
+    );
```
fvictorio, line 905: In general I would avoid doing this sort of abstraction, especially if there are only three methods. This is a personal preference (strongish one though), but it would also be more consistent with the rest of the codebase, were we prefer repetition over meta-programming.

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -867,6 +886,235 @@ Please replace "${contractName}" for the correct contract name wherever you are
...
+        if (value === undefined) {
+          continue;
+        }
+        return value;
+      } catch (error) {
+        if (
+          error instanceof HardhatError &&
+          error.number === ERRORS.ARTIFACTS.NOT_FOUND.number
+        ) {
+          caughtError = error;
```
fvictorio, line 1067: I think it would be better to modify the underlying logic to always return undefined and never throw, and remove this try/catch. The result would be the same as what's going on here, but much clearer.

packages/hardhat-core/src/internal/artifacts.ts
feuGeneA: The existing `Artifacts` class is quite tangled, and I opted for the dogfooding approach primarily as a means of verifying the new logic to use `ArtifactSources`. I intend to make some further refactorings to that class (likely separating it into layers), but i felt like this was a stable intermediate step to share for feedback.

Also, note that this current approach does support "special powers", via special-case delegation to only `_sourcesInPriorityOrder[0]` in the `Artifacts` methods that don't exist on `ArtifactSource`s, or, more specifically, those methods listed in this interface you commented on.

To me, the biggest benefit of the indirect options is to avoid growing the existing `Artifacts` class. Yes, it would be straightforward, I think we should attempt to take a step in the direction of breaking that class apart rather than growing it, and this separation of `Source` from `SourceManager` was the best thing I could think of.

packages/hardhat-core/src/internal/artifacts.ts
feuGeneA: Yeah, I'm not super happy with it, but it did seem to me to be better than repeating that whole try/catch block and all that mess yet again. I'm already repeating that mess twice (once for sync and once for async), and I've already had headaches from changing one but forgetting about the other, and I really wanted to avoid having 5 copies of that method instead of 2. But as usual, I'm happy to defer to convention and to your wisdom.

packages/hardhat-core/src/internal/artifacts.ts
feuGeneA: I would love to see this get cleaned up / simplified, but there's a good reason to catch and re-throw here: that lower layer offers typo suggestions in the error, which we can't replicate at this level. (at least, not without exposing more stuff on the interface

packages/hardhat-core/test/internal/artifacts.ts
```diff
@@ -1250,6 +1261,338 @@ describe("Artifacts class", function () {
...
+        for (const CustomSource of [
+          ReadOnlySource,
+          MutableSource,
+          CachingSource,
+          Artifacts,
+        ]) {
+          const customArtifactSource = new Artifacts(
+            "bad artifact path string; internal source will fallback to custom",
+            [new CustomSource(this.tmpDir)]
+          );
```
fvictorio, line 1291: When `CustomSource` is `Artifacts`, this is doing something like:

```js
new Artifacts(
  "...",
  new Artifacts("this.tmpDir")
)
```

I _kind_ of understand why this works, although it surprised me. And if you look at the `Artifacts`'s constructor signature it feels like something that works more by coincidence than by design

packages/hardhat-core/src/internal/artifacts.ts
fvictorio: (Resolving, but see my other comment about this)

packages/hardhat-core/src/internal/artifacts/index.ts
```diff
@@ -0,0 +1,333 @@
...
+    return artifact;
+  }
+
+  public artifactExists(
+    contractNameOrFullyQualifiedName: string
+  ): Promise<boolean> {
+    return this._getFirstValueFromSources(
+      "artifactExists",
+      contractNameOrFullyQualifiedName
+    );
```
fvictorio, line 68: I think this is a bug. The first source will never return `undefined`, so its response will always be used. Instead, this should check all sources and return `true` if any returns `true`, and `false` otherwise.

Btw, I was on the fence about the `_getFirstValueFromSources` approach, but this convinced me that we shouldn't do it. This feels more like incidental duplication than something we should abstract over.

packages/hardhat-core/src/internal/artifacts/index.ts
```diff
@@ -0,0 +1,333 @@
...
+        }
+        return value;
+      } catch (error) {
+        if (
+          error instanceof HardhatError &&
+          error.number === ERRORS.ARTIFACTS.NOT_FOUND.number
+        ) {
+          lastCaughtError = error;
+          continue;
+        }
```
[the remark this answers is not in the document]
feuGeneA: We discussed this on a call. We decided to leave the current approach in place, but to also add a `TODO` comment to make this nicer in the way suggested above.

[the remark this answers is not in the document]
feuGeneA: Also create an Issue and cite it in the `TODO`

packages/hardhat-core/src/internal/artifacts/index.ts
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3285/commits/3647b0444ccb9f122dba864575f15de2b24f1032

packages/hardhat-core/src/internal/artifacts/mutable.ts
```diff
@@ -0,0 +1,325 @@
...
+
+    const debugFile: DebugFile = {
+      _format: DEBUG_FILE_FORMAT_VERSION,
+      buildInfo: relativePathToBuildInfo,
+    };
+
+    return debugFile;
+  }
+
+  private static _getBuildInfoName(
```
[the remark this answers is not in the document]
feuGeneA: We discussed on a call and decided to leave the current approach in place, since it does have the advantage of indicating to the reader that the method doesn't use any instance state data.

packages/hardhat-core/src/internal/artifacts/index.ts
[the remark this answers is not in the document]
feuGeneA: Added TODO's in https://github.com/NomicFoundation/hardhat/pull/3285/commits/ebcd4cd456a24372dbe7059194f15bd52b82c745

packages/hardhat-core/src/internal/artifacts/index.ts
```diff
@@ -0,0 +1,333 @@
...
+
+import { ReadOnlySource } from "./readonly";
+import { MutableSource } from "./mutable";
+import { CachingSource } from "./caching";
+
+type SupportedArtifactSource =
+  | ArtifactSource
+  | ReadOnlySource
+  | MutableSource
+  | CachingSource;
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3285/commits/9717f64f0cb1f794b9e91c56082dbc66a6c0502b

fvictorio (approved):

PR #2916, opened 2022-07-09
Issue #2170: New check for getContractAt()

SfyMantissa:
- [x] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.

---

New feature. Was previously discussed in [Issue #2170](https://github.com/NomicFoundation/hardhat/issues/2170).
Adds a check for input address to `getContractAt()` — now it MUST have code (be a contract), otherwise function throws.

What was done:

- [x] Changes to the module itself
- [x] Test update
- [x] README update 

All tests run fine:

image

**Let me know if anything else is needed for submission :)**

PR #3301, opened 2022-10-26
Don't check that address is a contract in getContractAt

fvictorio:
This PR undoes the changes introduced in #2916. This broke smock and there surely are other scenarios where this is not a good idea.


alcuadrado (approved):
I did like this feature, but i'm ok with removing it if it breaks smock.

PR #3305, opened 2022-10-28
Sort files when creating the solc input

fvictorio:
I manually verified that this fixes the problem in the Balancer repo.

I tried to create a fixture project with a synthetic example that reproduces the issue, but it turned out to be way harder than I expected, so I don't have tests for this.

**Update**: I want to give another try to having a test for this, please don't merge yet.

**Update 2**: I added a test that triggered the bug. This is ready for review now.

### DIFF
--- a/.changeset/mean-mice-tell.md
+++ b/.changeset/mean-mice-tell.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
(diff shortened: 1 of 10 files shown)


alcuadrado (approved):

alcuadrado:
Can we release this ASAP? @nventuro is being blocked by it.

breestemma (approved):

PR #3306, opened 2022-10-28
Fix stack traces test that use custom solc

### LINKED ISSUE #3295: Error while loading custom solidity compiler
### The Problem

When attempting to use a custom solidity compiler that was not released yet the hardhat-core tests fail with `HardhatError: HH11: An internal invariant was violated: Trying to get a compiler before it was downloaded` since the `_getCompilerBuild` function returns `undefined` because the compiler version does not exist in the `list.json`.

https://github.com/NomicFoundation/hardhat/blob/5c31db69a0981440c17dc7d24f8deeeaf5bc8f32/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts#L225

### To reproduce the bug

You can reproduce the bug by setting `HARDHAT_TESTS_SOLC_VERSION` to a non-released version like in the script below.

```sh
#!/bin/bash

(
  wget -c https://binaries.soliditylang.org/bin/soljson-v0.8.16+commit.07a7930e.js -O soljson-v0.8.16.js
  export HARDHAT_TESTS_SOLC_PATH=$(pwd)/soljson-v0.8.16.js
  export HARDHAT_TESTS_SOLC_VERSION=0.8.18 # <---- some version that was not released yet
  mkdir -p tmp && cd tmp
  git clone --depth 1 https://github.com/nomiclabs/hardhat.git
  cd hardhat
  ./scripts/run-tests-with-custom-solc.sh
)
```

### Expected behavior

Setting a custom compiler with `HARDHAT_TESTS_SOLC_PATH` should be possible even though the version is not a solc release. Note that in the script above the version `v0.8.16` was used as an example, but it could be your own solidity local build: `export HARDHAT_TESTS_SOLC_PATH=<your_solidity_repo>/build/solc/solc`

fvictorio:
This is kind of a hack, but it fixes #3295.

#3205 improves the code of the stack traces tests, and I will add a better fix for this there.

Also, @cameel, this only works with solcjs, but we now can also easily support using a custom native solc. Would that be useful for you?

### DIFF
--- a/packages/hardhat-core/test/internal/hardhat-network/stack-traces/compilation.ts
+++ b/packages/hardhat-core/test/internal/hardhat-network/stack-traces/compilation.ts
@@ -1,6 +1,9 @@
 import fs from "fs";
 import path from "path";
-import { NativeCompiler } from "../../../../src/internal/solidity/compiler";
+import {
+  Compiler as SolcJsCompiler,
+  NativeCompiler,
+} from "../../../../src/internal/solidity/compiler";
 import {
   Compiler,
   CompilerDownloader,
@@ -76,12 +79,14 @@ async function compile(
   input: CompilerInput,
   compiler: Compiler
 ): Promise<[CompilerInput, CompilerOutput]> {
+  let runnableCompiler: any;
   if (compiler.isSolcJs) {
-    throw new Error("These tests expect to be able to run native solc");
+    runnableCompiler = new SolcJsCompiler(compiler.compilerPath);
+  } else {
+    runnableCompiler = new NativeCompiler(compiler.compilerPath);
   }
-  const nativeCompiler = new NativeCompiler(compiler.compilerPath);
 
-  const output = await nativeCompiler.compile(input);
+  const output = await runnableCompiler.compile(input);
 
   if (output.errors) {
     for (const error of output.errors) {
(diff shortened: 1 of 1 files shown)


ekpyron:
> Also, @cameel, this only works with solcjs, but we now can also easily support using a custom native solc. Would that be useful for you?

Yes, this would definitely be useful to us.

nikola-matic:
> This is kind of a hack, but it fixes #3295.
> 
> #3205 improves the code of the stack traces tests, and I will add a better fix for this there.
> 
> Also, @cameel, this only works with solcjs, but we now can also easily support using a custom native solc. Would that be useful for you?

Yes please, our external tests actually use the native solc.

fvictorio:
@ekpyron @nikola-matic I added (yet another) environment variable: HARDHAT_TESTS_SOLC_NATIVE. If it's set to `true`, then the given compiler will be executed as a native compiler, not as a js one. This variable is optional; if you omit it, or set it to any value other than `true`, it will assume the given compiler is a js one.

I was tempted to just check if the file ended with `.js`, but I wasn't sure if that will always work, so I erred on the side of explicitness.

Also: I disabled the compilation cache when a custom solc is used. The reason is that the cache uses the given version (and the settings) as the key, but if you are testing nightly versions it could easily happen that the same "version" produces two different results. This is probably not relevant to you because the tests are run in the CI, but I'd rather be extra careful here.

I'm not super happy with the whole approach, and perhaps we can offer you a better mechanism in the future, but hopefully this is enough for now.

fvictorio:
(I'll merge this as soon as the CI finishes)

cameel:
> This is probably not relevant to you because the tests are run in the CI, but I'd rather be extra careful here.

Well, we sometimes have to debug them locally so the cache was an annoying complication in that use case. We actually wasted some time trying to figure out why our workaround for this issue did not work in CI - it turned out that it only worked locally because it was using a cached binary :) Disabling this is definitely helpful.

> I was tempted to just check if the file ended with `.js`, but I wasn't sure if that will always work, so I erred on the side of explicitness.

I think that's the right choice here. It's for testing anyway so being able to specify exactly what you want, without risk of it being misinterpreted is more important than an easy interface.

PR #3324, opened 2022-11-02
Remove redundant symbols

SebastianSupreme:
The text looks much cleaner and more uniform now. 😉


fvictorio:
I don't have a strong opinion about this change. @alcuadrado what do you think?

Before:

image

---

After:

image

Issue #3249, opened 2022-10-05
Load fixture issues


skosito:
Sometimes using same fixture in lot of tests throws this error:

```
 FixtureSnapshotError: There was an error reverting the snapshot of the fixture.

This might be caused by using nested loadFixture calls in a test, for example by using multiple beforeEach calls. This is not supported yet.
      at loadFixture (node_modules/@nomicfoundation/hardhat-network-helpers/src/loadFixture.ts:47:15)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at async recoverViaDistributor (test/Wallet2.ts:111:9)
```

There is no nested loadFixture, tests are working properly until some amount of loadFixture calls for same fixture across multiple files (therefore multiple describe blocks), but after, for example 10-15 usages it starts failing. If I take same fixture and just rename it and start using it, it works. Is there some limit or best practices on how to handle using same fixture across a lot of files and tests?

fvictorio:
@skosito any chance of creating an [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) for this problem? I know that these kinds of issues are hard to reduce to its minimal expression, so the second best thing would be to have access to the repo and to have clear reproduction steps.

skosito:
> @skosito any chance of creating an [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) for this problem? I know that these kinds of issues are hard to reduce to its minimal expression, so the second best thing would be to have access to the repo and to have clear reproduction steps.

thanks for reply, access to repo is not possible unfortunately, i will try to create reproducible example

skosito:
@fvictorio i am sorry for delay, but here is repo  i put up with min reproduction https://github.com/skosito/hardhat-load-fixture-min-repro, to reproduce just run 
```
npm install
npx hardhat test
```

fvictorio:
Thank your for creating the repro @skosito, I'll take a look as soon as I have some time.

fvictorio:
Ugh, I see what's going on now. This seems to be a bug of our current implementation.

Thanks a lot for the reproduction example, it was super useful.

skosito:
> Ugh, I see what's going on now. This seems to be a bug of our current implementation.
> 
> Thanks a lot for the reproduction example, it was super useful.

@fvictorio thank you for checking it out, let me know if there is a way to help more, happy to contribute if possible

[...]

PR #3327, opened 2022-11-03
Fix load fixture issue when reverting to future state

skosito:

---

Please check this issue https://github.com/NomicFoundation/hardhat/issues/3249 for more details and discussion.

The way that loadFixture is working now is that allows situations like this:

```
Load fixture 1
...
Load fixture n - 1
Load fixture n
```

Up until this point every loadFixture takes a snapshot and store it in snapshots array. If loadFixture is called for any `n-i` it will work because it reverts the state to previous state. However, it also allows this, which fails because it tries to revert from previous to future state:

```
Load fixture 1
...
Load fixture n 
Load fixture n - i
Load fixture n -> fails with invalid snapshot error
```

My proposal is to remove all snapshots that have snapshot id bigger than snapshot being restored. This way it will ensure that every time loadFixture is called for future state snapshot is simple re-taken again.
Running tests in hardhat-network-helpers package failed only one test, describing exactly this issue, so i modified it.
I tested this a bit in project I am working on and it fixed the issue.

### DIFF
--- a/.changeset/warm-ghosts-rescue.md
+++ b/.changeset/warm-ghosts-rescue.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-network-helpers": patch
+---
+
+Fixed an error triggered by certain combinations of `loadFixture` calls (#3249, thanks @skosito!)  
--- a/packages/hardhat-network-helpers/src/helpers/takeSnapshot.ts
+++ b/packages/hardhat-network-helpers/src/helpers/takeSnapshot.ts
@@ -7,6 +7,7 @@ export interface SnapshotRestorer {
    * taken.
    */
   restore(): Promise<void>;
+  snapshotId: any;
 }
 
 /**
@@ -38,5 +39,6 @@ export async function takeSnapshot(): Promise<SnapshotRestorer> {
         method: "evm_snapshot",
       });
     },
+    snapshotId
   };
 }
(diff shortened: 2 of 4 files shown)


fvictorio:
Thanks for sending this @skosito! I will try to review this as soon as I have some time.

skosito:
> Thanks for sending this @skosito! I will try to review this as soon as I have some time.

Hey @fvictorio did you have a chance to take a look?

fvictorio:
Hey, sorry, not yet. I'll try to do it this week so we can include this in the next release.

Aminadav:
I confirm that I had the same issue as mentioned, and when I use the patch it fixed it for me.

Aminadav (approved):
I confirm that I had the same issue as mentioned, and when I use the patch it fixed it for me.

skosito:
@fvictorio There were some linter errors in workflows, i fixed them in new commit

fvictorio:
> Looks like now the test is missing to cover scenario where snapshot is invalid for other reasons. Do you have some suggestions how to add that?

Not really, hopefully that simply won't happen? Well, unless someone manually reverts a snapshot with a snapshot id they didn't obtain... but that's too far-fetched. I'm ok with not testing it.

There was a very fun bug related to snapshot ids being strings that were being compared with `<=`. I fixed that. You can take a look at the new test to see how that could cause an error. It was unlikely to happen, but still.

That aside, this is a great PR, thanks for sending it!

skosito:
> > Looks like now the test is missing to cover scenario where snapshot is invalid for other reasons. Do you have some suggestions how to add that?
> 
> Not really, hopefully that simply won't happen? Well, unless someone manually reverts a snapshot with a snapshot id they didn't obtain... but that's too far-fetched. I'm ok with not testing it.
> 
> There was a very fun bug related to snapshot ids being strings that were being compared with `<=`. I fixed that. You can take a look at the new test to see how that could cause an error. It was unlikely to happen, but still.
> 
> That aside, this is a great PR, thanks for sending it!

@fvictorio thanks for info and adding that test, and thanks for checking this!

fvictorio:
Thank you!

Issue #3298, opened 2022-10-26
Cannot adjust EIP-1559 gas mechanism with hardhat deploy


rkdud007:
Hi, I am currently researching about how to set optimize gas with hardhat that can send my deployment more faster. I try to do this by hardhat.config.js file.

goerli: {
      url: `https://eth-goerli.g.alchemy.com/v2/${process.env.ALCHEMY_API_KEY}`,
      accounts: [process.env.PRI_KEY],
      gasPrice: 204369036266, 
    },

I saw these 'gasPrice' [worked well ](https://goerli.etherscan.io/tx/0xb9c5048bbabe3b373afaf6e90cafe34627caf567fd9c7e2352e931d4826869e2)when I deploy my contract, but it is not the gas fee mechanism of eip-1559. 

so what I want to do is, change custom priority fee and set max base fee like we do on eip-1559 on hardhat.config.js when deploying the contract. 

I searched, but it seems that currently I couldn't find out how to do that. 
if this feature is not yet on hardhat, I definitely think this is necessary for gas fee optimization.

rkdud007:
I did find a way to apply 1559 in a roundabout way. But it will be better if hardhat support this on hardhat.config.

`
  const FEE_DATA = {
    maxFeePerGas:         ethers.utils.parseUnits('100', 'gwei'),
    maxPriorityFeePerGas: ethers.utils.parseUnits('5',   'gwei'),
    baseFeePerGas : ethers.utils.parseUnits('20',   'gwei'),
};

  // Wrap the provider so we can override fee data.
  const provider = new ethers.providers.FallbackProvider([ethers.provider], 1);
  provider.getFeeData = async () => FEE_DATA;

  const signer = new ethers.Wallet(process.env.PRI_KEY1, provider);
  const StakingRewards = await ethers.getContractFactory("StakingRewards",signer);
`

brougkr:
Here is a solution for you: 
`  const feeData = await hre.ethers.provider.getFeeData();
  const Contract = await hre.ethers.getContractFactory(CONTRACT_NAME)
  const contract = await Contract.deploy({
    maxPriorityFeePerGas: feeData.maxPriorityFeePerGas,
    maxFeePerGas: feeData.maxFeePerGas,
    type: 2
  })`

rkdud007:
thanks @brougkr:) I want to figure out if we can do this on hardhat.config.js!

fvictorio:
[...]

PR #3331, opened 2022-11-08
feat: adjust eip-1559

rkdud007:
Add option to adjust EIP-1559 gas fields in the Hardhat Network config

I opened an issue few days ago. I changed some values and add one unit-test function. 
issue number is #3298  
hope hardhat team can take a look at my code and give me feedbacks!

---

### DIFF
--- a/packages/hardhat-core/src/internal/core/config/default-config.ts
+++ b/packages/hardhat-core/src/internal/core/config/default-config.ts
@@ -127,6 +127,8 @@ export const defaultHttpNetworkParams = {
   accounts: "remote" as "remote",
   gas: "auto" as "auto",
   gasPrice: "auto" as "auto",
+  maxFeePerGas: "auto" as "auto",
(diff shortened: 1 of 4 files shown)


fvictorio:
Hi @rkdud007, sorry for not reviewing your PR before.

There are a couple of problems here. First, it has compilation errors. Second, I don't think the behavior is correct, and there are some edge cases that are not handled (what happens if you set both gasPrice and maxFeePerGas? I think that should be an error; in fact, I think we should only add the max priority fee option, and that gasPrice works as max fee per gas when the priority is set).

Besides, there are no tests and the linter fails.

I think the first step here is to properly define the feature and the different behaviors. I opened #3418 to track that effort.

In the meantime, I'm going to close this PR for bookkeeping reasons. After we properly define the feature, feel free to open a new PR. But please make sure that 1) it compiles, 2) the linter passes and 3) it has proper tests.

PR #3358, opened 2022-11-18
feat: implement state management in Rethnet

Wodann:
**Depends on:**
- #3357 

Todo:
- [x] Fix Hardhat integration
- [x] Use VMAdapter interface


Wodann:
@fvictorio Ready to review. There are still two failing tests:

- [ ] "automine enabled with pending txs"
- [ ] "when automine is enabled" > "Should throw if the transaction fails"

Both are a result of logging/error message inequality which doesn't seem important to fix at this stage.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/vm-adapter.ts
```diff
@@ -53,6 +53,7 @@ export interface VMAdapter {
 
   // setters
   putAccount(address: Address, account: Account): Promise<void>;
+  putBlock(block: Block): Promise<void>;
```
fvictorio, line 56: If I'm understanding correctly, this method was added because the VM needs access to the last block hashes (for the `BLOCKHASH` opcode). But I think that, if possible, we should use a callback for this, instead of adding the hashes to the state manager, because:

- That way, we don't mix concerns. The blockchain has the block hashes, and the VM (through the state manager/db) access that information. Otherwise, we need to remember to always keep the DB and the blockchain in sync, which seems error-prone.
- It will be easier to do this for the forked db, because in that case using a callback is easier. With this approach, the forked DB should fetch the last 256 blocks during initialization, which is not ideal.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1705,9 +1722,12 @@ Hardhat Network's forking functionality only works with blocks from at least spu
       }
 
       const minerReward = this._common.param("pow", "minerReward");
+      const _stateRoot3 = await this._vm.getStateRoot();
       await blockBuilder.addRewards([[coinbase, minerReward]]);
+      const _stateRoot4 = await this._vm.getStateRoot();
```
fvictorio, line 1727: I guess this is a leftover of some debugging?

packages/hardhat-core/src/internal/hardhat-network/provider/utils/random.ts
```diff
@@ -54,4 +54,7 @@ export const randomAddressString = () => {
   return bufferToHex(randomAddressBuffer());
 };
 
-export const randomAddressBuffer = () => randomHashBuffer().slice(0, 20);
+const addressGenerator = RandomBufferGenerator.create("seed");
+export const randomAddressBuffer = (): Buffer => {
+  return addressGenerator.next().slice(0, 20);
+};
```
fvictorio, line 60: What's the reason for this change?

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
```diff
@@ -89,42 +100,109 @@ export class DualModeAdapter implements VMAdapter {
...
+    if (
+      !(
+        ethereumJSStorageSlot.length === 0 &&
+        rethnetStorageSlot.equals(Buffer.from([0x00]))
+      ) &&
+      !ethereumJSStorageSlot.equals(rethnetStorageSlot)
+    ) {
+      console.trace(ethereumJSStorageSlot);
+      console.trace(rethnetStorageSlot);
+      console.trace(
```
fvictorio, line 148: Just FYI: `console.trace` will print the message and the full trace. For three consecutive calls, the trace is the same, so this output will be unnecessarily noisy. You can either replace all of them with `console.log` or, if you do want to see the call trace, just use one `console.trace`.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/exit.ts
```diff
@@ -34,7 +34,7 @@ export class Exit {
       case 0x51:
       case 0x53:
         return new Exit(ExitCode.INVALID_OPCODE);
-      case 0x65:
+      case 0x64:
```
fvictorio, line 37: Is this correct? [`Return.CreateContractLimit`](https://github.com/bluealloy/revm/blob/90fe01e703dc4c91c3526268bd314745fe82545a/crates/revm/src/instructions.rs#L70) is 0x65, unless I'm missing something

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -982,6 +990,13 @@ Hardhat Network's forking functionality only works with blocks from at least spu
       mixHashGenerator: this._mixHashGenerator.clone(),
     };
 
+    const rethnetSnapshot = await this._rethnetState.makeSnapshot();
```
fvictorio, line 993: Shouldn't there be a matching `this._rethnetState.removeSnapshot` call?

packages/hardhat-core/src/internal/hardhat-network/provider/utils/random.ts
Wodann: I mimicked the random buffer generator in Rust and let them run side by side for generation of state roots (needed for forked databases, so we don't need to fork all blocks). It turned out that in the tests, the address generator was using the same `RandomBufferGenerator` and thus they would become offset.

I split the addressGenerator so that it has its own seed and won't interfere anymore

packages/hardhat-core/src/internal/hardhat-network/provider/vm/exit.ts
Wodann: image

packages/hardhat-core/src/internal/hardhat-network/provider/vm/vm-adapter.ts
Wodann: I added two issues to improve this in the future. It requires a change in `revm`:
https://github.com/NomicFoundation/rethnet/issues/31
https://github.com/NomicFoundation/rethnet/issues/32

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
Wodann: The problem is that I wouldn't know when to actually remove the snapshot. For now, all snapshots keep existing.

I made a tracking issue for this effort: https://github.com/NomicFoundation/rethnet/issues/33

packages/hardhat-core/src/internal/hardhat-network/provider/vm/exit.ts
fvictorio: I think it changed recently: https://github.com/bluealloy/revm/commit/69e302b48248d56dc285631b7f88332017bd4d0b#diff-8ef36919563c5922c56bf87d70a864582b8f2ddefe35e879542c208bbddc5484

I assumed that all new errors would be added at the end, like we do with our error codes in Hardhat, but I guess that's not the case. Good to know.

fvictorio: (To be clear: you are right, is 0x65 now)

fvictorio: So, to be clear: in the latest commit it's 0x65, so we should keep that in mind when we upgrade revm.

Wodann: Yup, I just ran into that problem myself when doing the `revm` update. Corrected is as you suggested. Thanks!

packages/hardhat-core/src/internal/hardhat-network/provider/vm/vm-adapter.ts
Wodann: I already implemented the changes and added them to the PR.

After I had updated a upstream dependency, the build was failing until I resolved this, so figured I might as well combine the two.

crates/rethnet_evm/src/runtime.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub async fn dry_run(
+        &mut self,
+        transaction: TxEnv,
+        mut block: BlockEnv,
+    ) -> (ExecutionResult, State, Trace) {
+        let blockchain = self.blockchain.clone();
+        let db = self.db.clone();
+        let cfg = self.cfg.clone();
+
+        if cfg.spec_id > SpecId::MERGE && block.prevrandao.is_none() {
```
Wodann, line 52: I had to add this as we're not setting the `mixHash` for block headers in Hardhat. This feels like a hack though. What do you think, @fvictorio ?

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
```diff
@@ -445,9 +445,10 @@ export class VMDebugTracer {
       );
     } else if (step.opcode.name === "INVALID") {
       const code = await this._getContractCode(step.codeAddress);
-
```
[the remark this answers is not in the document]
fvictorio: The fix makes sense to me. I just added an else branch to have some defined value for the `op` field of the entry, and an explanatory comment.

crates/rethnet_evm/src/runtime.rs
fvictorio: I'm not sure I understand. We do set the `mixHash` in block headers.

With respect to this code... I think we should panic if we are in a post-merge hardfork and `prevrandao` is not set? I'm not 100% sure though, and I'm fine with leaving it as it is.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
```diff
@@ -89,42 +100,102 @@ export class DualModeAdapter implements VMAdapter {
...
+      key
+    );
+
+    if (
+      !(
+        ethereumJSStorageSlot.length === 0 &&
+        rethnetStorageSlot.equals(Buffer.from([0x00]))
+      ) &&
+      !ethereumJSStorageSlot.equals(rethnetStorageSlot)
+    ) {
```
fvictorio, line 145: I found this very hard to understand. What about doing something like:

```js
if (!ethereumJSStorageSlot.equals(rethnetStorageSlot)) {
  // we only throw if any of the returned values was non-empty
  if (ethereumJSStorageSlot.length !== 0 || !rethnetStorageSlot.equals(Buffer.from([0x00]))) {
    // throw here
  }
}
```

I'm not sure if the comment is 100% accurate, but I guess that's the idea?

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
```diff
@@ -89,42 +100,102 @@ export class DualModeAdapter implements VMAdapter {
...
+      throw new Error("Different contract code");
+    }
+
+    return rethnetCode;
   }
 
   public async putAccount(address: Address, account: Account): Promise<void> {
-    return this._ethereumJSAdapter.putAccount(address, account);
+    await this._ethereumJSAdapter.putAccount(address, account);
+    return this._rethnetAdapter.putAccount(address, account);
```
fvictorio, line 177:
```suggestion
    await this._rethnetAdapter.putAccount(address, account);
```

Personal preference: I would await both calls instead of returning the second. Otherwise, this makes you stop to think "why is the second value returned", when it's not necessary.

Same for the other similar methods.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/block-builder.ts
```diff
@@ -89,7 +89,9 @@ export class BlockBuilder {
       calcDifficultyFromHeader: this._opts.parentBlock.header,
     });
 
+    const _stateRoot = await this._vm.getStateRoot();
     const [result] = await this._vm.runTxInBlock(tx, block);
+    const _stateRoot2 = await this._vm.getStateRoot();
```
fvictorio, line 94: Can we remove these two lines?

packages/hardhat-core/src/internal/hardhat-network/provider/RethnetState.ts
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  public async accountIsEmpty(address: Address): Promise<boolean> {
+    const account = await this._state.getAccountByAddress(address.buf);
+    return (
+      account !== null &&
+      account.balance === 0n &&
+      account.nonce === 0n &&
+      account.codeHash.equals(KECCAK256_NULL)
+    );
```
fvictorio, line 66: Why is this not `return (account === null || (balance === 0 && etc))`?

crates/rethnet_evm_napi/src/state.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    #[napi(ts_return_type = "Promise<void>")]
+    pub fn modify_account(
+        &self,
+        env: Env,
+        address: Buffer,
+        #[napi(
+            ts_arg_type = "(balance: bigint, nonce: bigint, code: Buffer | undefined) => Promise<AccountData>"
+        )]
+        modify_account_fn: JsFunction,
```
fvictorio, line 180: Just to be sure I understand: this function is called with the current state of the account, and it updates the account with the values returned by the function. Is that correct? If so, can we document it? I believe just adding a `///` comment generates the proper jsdoc, like it's done in the `finalize` and `abort` functions of the block builder.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -982,6 +990,13 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+    if (!rethnetSnapshot.equals(snapshot.stateRoot)) {
+      throw new Error(
+        "Snapshot with unequal state roots: ${snapshot.stateRoot} !== ${rethnetSnapshot}."
+      );
+    }
```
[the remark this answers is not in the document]
Wodann: Sounds like the proper fix to do here! Sorry, I forgot to create a discussion topic to ask you 🙏

Feel free to do that

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
Wodann: Yeah, that makes sense. As far as I understand the spec, you're supposed to return zero if a storage slot is empty.

Ethereumjs just returns an empty buffer in that case, hence why the two implementations diverge and we need to check this. I'm not sure if that helps you write a better comment

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
Wodann: Will update

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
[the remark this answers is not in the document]
Wodann: I just implemented it ✅

packages/hardhat-core/src/internal/hardhat-network/provider/RethnetState.ts
Wodann: You're right. Wrong application of negation to get rid of the `||` 😅 

Thanks for catching that!

crates/rethnet_evm_napi/src/state.rs
Wodann: Yes, your understanding is correct.

I added the documentation. Once the API solidifies I'll ensure that every function, object, and class has comments on it.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
```diff
@@ -224,3 +327,39 @@ function assertEqualRunTxResults(
...
+  }
+
+  if (ethereumJSAccount.nonce !== rethnetAccount.nonce) {
+    console.trace(
+      `Different nonce: ${ethereumJSAccount.nonce} !== ${rethnetAccount.nonce}`
+    );
+    throw new Error("Different nonce");
+  }
+
+  // TODO: Add storageRoot to Rethnet
```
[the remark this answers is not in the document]
Wodann: Good call. The question I guess is: Do we want to always know the storage root when we query account data? It's extra computation (for now at least)

crates/rethnet_evm/src/runtime.rs
Wodann: There are about 300 tests (potential half due to the existence of two providers) that are currently failing without this.

If the requirement isn't met I'd just return an `Err`. Panicking would cause a crash so should only happen when an internal assumption about the data model isn't met. E.g. like the `state_root.unwrap()` we discussed the other day.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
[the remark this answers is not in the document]
fvictorio: Hmm, apparently we are not using it anywhere, so... maybe we just don't do this for now? I mean, let's not even open an issue. If it comes up as a necessity later, we can re-evaluate.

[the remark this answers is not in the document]
fvictorio: Or, alternative idea: let's do it just because it lets us compare more things, but we drop it later when we remove ethereumjs?

I'm fine with either option, because the extra confidence this would give us might not be worth it. Up to you!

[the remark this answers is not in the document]
Wodann: I can add it easily. Already have a function for calculating the storage root. I proposed making it a separate API so that we can keep it, because people might want to manually retrieve a storage root for an account? 🤷‍♂️

fvictorio (approved):

crates/rethnet_evm/src/runtime.rs
Wodann: https://github.com/NomicFoundation/hardhat/pull/3443

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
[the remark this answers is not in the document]
fvictorio: Sure, that seems sensible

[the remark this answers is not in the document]
Wodann: Resolved by https://github.com/NomicFoundation/hardhat/pull/3454

PR #3361, opened 2022-11-18
Refactor VMAdapter

fvictorio:
Closes https://github.com/NomicFoundation/rethnet/issues/16
Closes https://github.com/NomicFoundation/rethnet/issues/17

@Wodann I removed the HardhatDB test because that functionality is being indirectly validated just by using the rethnet adapter, so I'd rather not keep that test up to date.


Wodann (approved):

Wodann:
LGTM!

PR #3362, opened 2022-11-19
Upgrade cbor and fix cbor import type

emretepedev:
`@types/cbor` has been deprecated.
`cbor` provides its own type definitions now.


fvictorio:
Great catch, thank you @emretepedev!

PR #3367, opened 2022-11-22
Add getBuildInfoSync

emretepedev:
I needed a synchronized version of **getBuildInfo** (like readArtifactSync) in a Hardhat plugin I am developing.

It would be nice to have a synchronized version to use in the constructor as well.


fvictorio:
Thank you!

PR #3368, opened 2022-11-22
Use proper async/await syntax in sample code

delaaxe:

---

Avoid using deprecate Promise based syntax

### DIFF
--- a/packages/hardhat-core/sample-projects/typescript/scripts/deploy.ts
+++ b/packages/hardhat-core/sample-projects/typescript/scripts/deploy.ts
@@ -17,7 +17,11 @@ async function main() {
 
 // We recommend this pattern to be able to use async/await everywhere
 // and properly handle errors.
-main().catch((error) => {
-  console.error(error);
-  process.exitCode = 1;
-});
+(async () => {
+  try {
+    await main();
+  } catch (error) {
+    console.error(error);
+    process.exitCode = 1;
+  }
+}());


alcuadrado:
Hey! this pr caught my attention. Deprecated? Do you have more info about it?

delaaxe:
> Hey! this pr caught my attention. Deprecated? Do you have more info about it?

Sorry didn't get notified for some reason 

It's just regular old async/await, where rejections are meant to be dealt in try/catch blocks, instead of the old promise based `.then()`/`.catch()`.

The surrounding IIFE is to enable async/await at module level

fvictorio:
@delaaxe "deprecated" usually means "it will be removed in the future so you shouldn't use it", which I don't think is the case with plain promises.

I personally prefer the current version, since it's less verbose and more straightforward. But I wouldn't be against merging your change if there's a good reason that I'm not seeing.

delaaxe:
> @delaaxe "deprecated" usually means "it will be removed in the future so you shouldn't use it", which I don't think is the case with plain promises.
> 
> 
> 
> I personally prefer the current version, since it's less verbose and more straightforward. But I wouldn't be against merging your change if there's a good reason that I'm not seeing.

Less verbose doesn't always mean more understandable and I find a regular try/catch more straightforward.

I agree it's not deprecated per se but languages evolve and it's good practice to use the latest ways of doing this.

I agree it's debatable though.

fvictorio:
I understand your point but I think this is very subjective so, for now, I'd rather keep it as it is.

To clarify why I think the current version is simpler, this is what happens in my head when I read each one:

- "`main` returns a promise and, if it rejects, we print the error and finish the process with an exit code of 1"
- "we create an async IIFE and then we call and await `main` inside a try/catch; if `main` rejects, we catch and print the error and finish the process with an exit code of 1"

I understand though that if you have a different mental model, then the first version can actually be more confusing/harder to process. If we get more feedback suggesting that using an IIFE is clearer, we'll reconsider this.

PR #3372, opened 2022-11-23
Add no-compile flag to etherscan verify task

spalladino:
Adds an optional --no-compile flag to the verify task, which does what you'd expect it to do.


alcuadrado:
Thanks @spalladino! I just added a changeset to the PR nd will merge once the CI passes.

PR #3373, opened 2022-11-24
feat: print debug traces if results don't match

fvictorio:
I only implemented the ethereumjs part of this. I will add relevant comments inline.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
@@ -13,10 +13,18 @@ import { HardhatBlockchainInterface } from "../types/HardhatBlockchainInterface"
 
 import { EthereumJSAdapter } from "./ethereumjs";
 import { RethnetAdapter } from "./rethnet";
-import { VMAdapter } from "./vm-adapter";
+import { Trace, VMAdapter } from "./vm-adapter";
 
 /* eslint-disable @nomiclabs/hardhat-internal-rules/only-hardhat-error */
 /* eslint-disable @typescript-eslint/restrict-template-expressions */
+
+function printEthereumJSTrace(trace: any) {
+  console.log(JSON.stringify(trace, null, 2));
+}
+
+function printRethnetTrace(_trace: any) {
+  // not implemented
+}
 
 export class DualModeAdapter implements VMAdapter {
   constructor(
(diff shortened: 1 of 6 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
```diff
@@ -13,11 +13,19 @@ import { HardhatBlockchainInterface } from "../types/HardhatBlockchainInterface"
...
 
 /* eslint-disable @nomiclabs/hardhat-internal-rules/only-hardhat-error */
 /* eslint-disable @typescript-eslint/restrict-template-expressions */
 
+function printEthereumJSTrace(trace: any) {
+  console.log(JSON.stringify(trace, null, 2));
+}
+
+function printRethnetTrace(_trace: any) {
+  // not implemented
```
fvictorio, line 26: After returning _something_ as a trace in the rethnet adapter, you can implement how it will be printed here.

Depending on the value, the same implementation that `printEthereumJSTrace` uses might be enough.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/rethnet.ts
```diff
@@ -83,7 +83,9 @@ export class RethnetAdapter implements VMAdapter {
       difficulty,
     });
 
-    return rethnetResultToRunTxResult(rethnetResult.execResult);
+    const result = rethnetResultToRunTxResult(rethnetResult.execResult);
+
+    return [result, null];
```
fvictorio, line 88: Replace this `null` with the trace returned by rethnet.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/ethereumjs.ts
```diff
@@ -172,13 +172,28 @@ export class EthereumJSAdapter implements VMAdapter {
...
+            tx,
+            skipNonce: true,
+            skipBalance: true,
+            skipBlockGasLimitValidation: true,
+          });
+        },
+        {
+          disableStorage: true,
+          disableMemory: true,
+          disableStack: true,
```
fvictorio, line 190: If you want extra information in the ethereumjs trace, set any (or all) of these values to `false`.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/ethereumjs.ts
```diff
@@ -373,8 +388,23 @@ export class EthereumJSAdapter implements VMAdapter {
...
+    const vmTracer = new VMDebugTracer(this._vm);
+    let result: RunTxResult | undefined;
+    const trace = await vmTracer.trace(
+      async () => {
+        result = await this._vm.runTx({ tx, block });
+      },
+      {
+        disableStorage: true,
+        disableMemory: true,
+        disableStack: true,
```
fvictorio, line 401: Same comment as above.

Wodann (approved):

PR #3377, opened 2022-11-24
Define minimal types involved in tracing

fvictorio:


packages/hardhat-core/src/internal/hardhat-network/provider/vm/vm-adapter.ts
```diff
@@ -1,17 +1,38 @@
...
+    executionGasUsed: bigint;
+  };
+}
+
+export interface TracingCallbacks {
+  beforeMessage: (message: TracingMessage, next: any) => Promise<void>;
+  step: (step: TracingStep, next: any) => Promise<void>;
+  afterMessage: (result: TracingMessageResult, next: any) => Promise<void>;
+}
+
```
fvictorio, line 35: @Wodann this is the relevant part of this change. These callbacks should be called with those values.

We need to figure out if the `next` callbacks will be called from rethnet or from the adapter, but that's a minor detail.

Wodann: Would you be opposed to defining these in `rethnet_evm_nqpi`? Then they are the same between Rethnet and Hardhat automatically

fvictorio: No problem at all!

fvictorio: Done :slightly_smiling_face:

Wodann (approved):

PR #3382, opened 2022-11-29
Add hardhat_metadata RPC method

fvictorio:
Adds a new `hardhat_metadata` RPC method. This method is meant to be used by other tools, like wallets or deployment systems, to get information about the instance of the network.

### DIFF
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/output/metadata.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/output/metadata.ts
@@ -6,9 +6,6 @@ export interface HardhatMetadata {
   // The chain's id. Used to sign transactions.
   chainId: number;
 
-  // The network's id. Normally matches the chainId. Should not be used to sign transactions.
-  networkId: number;
-
   // A 0x-prefixed hex-encoded 32 bytes id which uniquiely identifies an instance/run
   // of Hardhat Network. Running Hardhat Network more than once (even with the same version
   // and parameters) would always result in different `instanceId`s.
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/output/metadata.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/output/metadata.ts
@@ -1,0 +1,35 @@
+export interface HardhatMetadata {
+  // A string identifying the version of Hardhat, for debugging purposes,
+  // not meant to be displayed to users.
+  clientVersion: string;
+
+  // The chain's id. Used to sign transactions.
+  chainId: number;
+
+  // The network's id. Normally matches the chainId. Should not be used to sign transactions.
+  networkId: number;
+
+  // A 0x-prefixed hex-encoded 32 bytes id which uniquiely identifies an instance/run
+  // of Hardhat Network. Running Hardhat Network more than once (even with the same version
+  // and parameters) would always result in different `instanceId`s.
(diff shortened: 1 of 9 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/modules/web3.ts
```diff
@@ -32,9 +34,7 @@ export class Web3Module {
   }
 
   private async _clientVersionAction(): Promise<string> {
-    const hardhatPackage = await getPackageJson();
-    const ethereumjsVMPackage = require("@nomicfoundation/ethereumjs-vm/package.json");
-    return `HardhatNetwork/${hardhatPackage.version}/@nomicfoundation/ethereumjs-vm/${ethereumjsVMPackage.version}`;
+    return this._node.getClientVersion();
```
fvictorio, line 37: The value returned in the `clientVersion` field of the metadata is the same one that is returned by `web3_clientVersion`. Is that ok?

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
```diff
@@ -88,7 +90,14 @@ Please use block number ${lastSafeBlock} or wait for the block to get ${
     cacheToDiskEnabled ? forkCachePath : undefined
   );
 
-  return { forkClient, forkBlockNumber, forkBlockTimestamp };
+  const forkBlockHash = block.hash;
+
+  assertHardhatInvariant(
+    forkBlockHash !== null,
+    "Forked block should have a hash"
+  );
```
fvictorio, line 98: I think only pending blocks have a null hash, so this should be fine.

packages/hardhat-core/src/internal/core/jsonrpc/types/output/metadata.ts
```diff
@@ -0,0 +1,35 @@
...
+  // The latest block's number in Hardhat Network
+  latestBlockNumber: number;
+
+  // The latest block's hash in Hardhat Network
+  latestBlockHash: string;
+
+  // This field is only present when Hardhat Network is forking another chain.
+  forkedNetwork?: {
+    // The chainId of the network that is being forked
+    chainId: number;
```
alcuadrado, line 27: Should add `networkId` here, for consistency?

packages/hardhat-core/src/internal/hardhat-network/provider/modules/web3.ts
alcuadrado: Yep

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
alcuadrado: Yes, this is a valid assumption

alcuadrado (approved):
LGTM.

I just left a comment about fork network id, mostly to make the API more consistent.

packages/hardhat-core/src/internal/core/jsonrpc/types/output/metadata.ts
fvictorio: Removed `networkId` instead.

alcuadrado:
> I just left a comment about fork network id, mostly to make the API more consistent.

We decided to drop `networkId` instead, as it's not clearly why one would use it for. We can reintroduce it if needed.

PR #3384, opened 2022-11-30
Custom error with selector in `solidity-errors.ts`

vivianjeng:

---

Relates to https://github.com/NomicFoundation/hardhat/issues/2948

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-errors.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-errors.ts
@@ -8,6 +8,7 @@ import {
   SolidityStackTraceEntry,
   SourceReference,
   StackTraceEntryType,
+  UNKNOWN_ERROR_SELECTOR,
   UNKNOWN_FUNCTION_NAME,
   UNRECOGNIZED_CONTRACT_NAME,
   UNRECOGNIZED_FUNCTION_NAME,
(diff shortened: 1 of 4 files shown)


fvictorio:
Thank you @vivianjeng, I think we should also update this test:

https://github.com/NomicFoundation/hardhat/blob/5cecb9692ca8bfa4a57a592e3233b24336da6e1c/packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/sendTransaction.ts#L417-L424

packages/hardhat-core/src/internal/hardhat-network/stack-traces/solidity-errors.ts
```diff
@@ -271,7 +271,8 @@ function getMessageFromLastStackTraceEntry(
       }
 
       if (!stackTraceEntry.message.isEmpty()) {
-        return `VM Exception while processing transaction: reverted with an unrecognized custom error`;
+        const selector = stackTraceEntry.message.getSelector() ?? "";
```
fvictorio, line 274: I think in most cases the selector will exist, but if for some reason it doesn't, the error message will say "with an unrecognized custom error with selector ", which would be ugly and confusing. So I would do something like this:

```suggestion
        const selector = stackTraceEntry.message.getSelector() ?? "<unknown selector>";
```

It's likely that we could do something better here: if there is some return data but it's shorter than 4 bytes, something weird is going on and we should reflect that. But for the time being, this should be enough.

vivianjeng: Thank you for the suggestion
I use a `UNKNOWN_ERROR_SELECTOR` as unknown selector 
Please check 🙏🏻 [15e92d9](https://github.com/NomicFoundation/hardhat/pull/3384/commits/15e92d9c10e9f60f8f45e2a3a3cfd01cdee6d46c)

fvictorio:
Thank you @vivianjeng! I ended up changing my mind about the unrecognized selector. Sorry for making you implement that and then reversing it :sweat_smile: Check my commit to see what I did instead.

PR #3398, opened 2022-12-05
Added Nodereal as another infra provider

ankushnodereal:
Added Nodereal as another infra provider

---


fvictorio:
This seems to be a duplicate of #3397.

alcuadrado:
I've never heard of this provider before, so this looks like a marketing effort, so I'm closing it.

If this provider gets popular, and a similar change is proposed organically, we may accept it.

PR #3415, opened 2022-12-10
Basic support for viaIR

### LINKED ISSUE #1533: Default sourceReferences in ErrorInferrer
There are some places in the `ErrorInferrer` where we do things like this:

https://github.com/nomiclabs/hardhat/blob/6ace4efd4de68bd876330861185b9271dba6ddcf/packages/hardhat-core/src/internal/hardhat-network/stack-traces/error-inferrer.ts#L985-L992

Notice the `this._getLastSourceReference(trace)!`. That is, we try to get the source reference for the revert bytecode and, if that doesn't work, we get the last mapped source reference and **assume that one exists**.

This isn't great for two reasons:

1. We should avoid non-null assertions altogether.
2. We should be smarter about what we default to in these scenarios.

It's not clear what exactly we need to do here. One idea is to add some helpers that default to some source reference and encode their assumptions using `assertHardhatInvariant`. For example:

```js
private _getDefaultLastSourceReference(...): SourceReference {
  const sourceReference = this._getLastSourceReference(trace);
  assertHardhatInvariant(sourceReference !== undefined, "There should be some mapped bytecode in this trace")
  return sourceReference
}
```

(we can also rename `_getLastSourceReference` to `_tryGetLastSourceReference` and name the asserted one `_getLastSourceReference`)

### LINKED ISSUE #2724: Strack traces not compatible with viaIR
Using hardhat 2.9.5, after enabling `viaIR`:
```json
  solidity: {
    compilers: [
      {
        version: "0.8.13",
        settings: {
          viaIR: true,
          optimizer: {
            enabled: true,
            runs: 1000,
          },
        },
      },
    ],
  },
```

Some stack traces cannot be generated properly:
````
/Projects/test/node_modules/hardhat/internal/hardhat-network/stack-traces/solidity-errors.js:108
    return new SolidityCallSite(sourceReference.sourceName, sourceReference.contract, sourceReference.function !== undefined
                                                ^

TypeError: Cannot read property 'sourceName' of undefined
````

Dumping the `stackTraceEmpty` shows it is for a panic, without a source reference:
```
{
  type: 5,
  sourceReference: undefined,
  errorCode: BN {
    negative: 0,
    words: [
      17, 0, 0, 0, 0,
       0, 0, 0, 0, 0,
       0
    ],
    length: 1,
    red: null
  }
}
```

Unfortunately I cannot share a demo repository, but I believe the following should trigger:
1. Code with arithmetic overflow (e.g. `balanceOf[msg.sender] = 1; balanceOf[msg.sender] += type(uint256).max`)
2. A test case calling the above (it can expect a failure or not)

### PR DESCRIPTION
**Pull request by fvictorio**
Closes #2724.
Closes #1533

This PR is (almost) the same as #3205, but with a better commits structure. I recommend reading the commits one by one when reviewing this.

### CODE REVIEW

**Comment by fvictorio** [comment]
Results of using this in the OpenZeppelin repo, enabling viaIR and running their tests:

- If you use solc 0.8.18 (nightly) and set `":"` as the optimizer steps, all tests pass.
- If you use solc 0.8.17 and set "" as the optimizer steps, there are 34 failures, but all of them are caused by https://github.com/ethereum/solidity/issues/13786.
- If you use solc 0.8.17 with the default optimizer steps, there are 44 failures. All of them are caused by Hardhat returning `"Transaction reverted and Hardhat couldn't infer the reason"` instead of the expected error. Since this scenario is using both the new codegen and optimizations enabled, I feel this is acceptable. 

  We _could_ improve these results, though, if we had a better implementation of the `ErrorInferrer` that _always_ returns the correct top stack trace entry when there is return data. I tried to do this but I wasn't able to get it right.

**Comment by ekpyron** [comment]
It may be worth a try for you to use ``"u"`` as optimizer step sequence for 0.8.17 - that should be powerful enough to eliminate the link references, but weak enough not to interfere with anything else.

**Comment by fvictorio** [comment]
Indeed! All OZ tests pass with `optimizerSteps: "u"`. Thanks a lot, @ekpyron!

File: packages/hardhat-core/src/internal/cli/cli.ts [hunk 1]
```diff
@@ -104,6 +104,23 @@ async function suggestInstallingHardhatVscode() {
   }
 }
 
+function showViaIRWarning(resolvedConfig: HardhatConfig) {
+  const viaIREnabled = resolvedConfig.solidity.compilers.some(
+    (compiler) => compiler.settings?.viaIR === true
+  );
```
fvictorio, line 110: We should check the overrides too.

packages/hardhat-core/test/internal/hardhat-network/stack-traces/compilation.ts
```diff
@@ -33,41 +31,70 @@ function getSolcSourceFileMapping(sources: string[]): SolcSourceFileToContents {
...
 ): CompilerInput {
+  const isViaIR = compilerOptions.optimizer?.viaIR ?? false;
+
+  // using ":" as a minimal setting for the yul optimizer steps is only
+  // available starting from version 0.8.18
+  const optimizerDetails =
+    semver.gte(compilerOptions.solidityVersion, "0.8.18") && isViaIR
+      ? {
+          yulDetails: {
+            optimizerSteps: ":",
```
fvictorio, line 44: Use `"u"` and remove `0.8.18` check

alcuadrado:
I think this commit doesn't work well with overloaded functions 1249a530410e80c11d8ee12c997b7ccb4b6a08f0, apart from that it LGTM

alcuadrado:
Should the guide recommend  using `"u"` for solc < 0.8.18 and `":"` otherwise?

alcuadrado:
>  it depends on your heuristics if "u" or "u:" works better (the latter has fewer optimizations than the former), but it's probably a good idea to keep including the u in any case.

We'll try it out and let you know.

> We're by the way considering [ethereum/solidity#13972](https://github.com/ethereum/solidity/issues/13972) for 0.8.20 - that may make this smoother.

That's great!

btw, we are planning to release this next week.

If anyone want's to test it, I just published a build under `hardhat@viair`

fvictorio:
> I think this commit doesn't work well with overloaded functions

Ugh, I'm glad you caught that, thank you.

I fixed it by passing all the function abis that match the name, and then checking if any of those matches the selector of the function node. I also added a test that triggered the problem. See https://github.com/NomicFoundation/hardhat/pull/3415/commits/dea67989e2d9f9ada6b2b225f3465b00095a129e.

PR #3432, opened 2022-12-16
Improve proxy support

fvictorio:
This PR:

- Adds support for the `http_proxy` environment variable both for downloading the compiler and for the HTTP provider. This means that both forking and connecting to networks via JSON-RPC should use the proxy.
- Adds support for using `http_proxy` in `hardhat-etherscan`.

As part of this, I'm technically introducing a breaking change: the `HTTP_PROXY` and `HTTPS_PROXY` variables don't work anymore for the compiler download. The first one (all uppercase) is not the most commonly used, for [fun historical reasons](https://everything.curl.dev/usingcurl/proxies/env#http_proxy-in-lower-case-only). And the second one is misleading: we use the same proxy for both http and https.

### DIFF
--- a/.changeset/few-flies-drum.md
+++ b/.changeset/few-flies-drum.md
@@ -1,0 +1,7 @@
+---
+"hardhat": patch
+---
+
+Added support for the `http_proxy` environment variable. When this variable is set, Hardhat will send its requests through the given proxy for things like JSON-RPC requests, mainnet forking and downloading compilers.
+
+We also removed support for the `HTTP_PROXY` and `HTTPS_PROXY` environment variables, since `http_proxy` is the most commonly used environment variable for this kind of thing. Those variables could only be used for downloading compilers. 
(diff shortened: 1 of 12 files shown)


umakanth-fmr:
@fvictorio , any tentative timeline or release this PR planned ?

fvictorio:
Can't promise a date but hopefully soon? It will almost surely be included in our next release.

.changeset/few-flies-drum.md
```diff
@@ -0,0 +1,7 @@
+---
+"hardhat": patch
+---
+
+Added support for the `http_proxy` environment variable. When this variable is set, Hardhat will send its requests through the given proxy for things like JSON-RPC requests, mainnet forking and downloading compilers.
+
+We also removed support for the `HTTP_PROXY` and `HTTPS_PROXY` environment variables, since `http_proxy` is the most commonly used environment variable for this kind of thing. Those variables could only be used for downloading compilers.
```
alcuadrado, line 7:
```suggestion
We also removed support for the `HTTP_PROXY` and `HTTPS_PROXY` environment variables, since `http_proxy` is the most commonly used environment variable for this kind of thing. Those variables could only be used for downloading compilers.

Finally, we also added support for `no_proxy`, which accepts a comma separated list of hosts or `"*"`. By default its value is `"127.0.0.1"`.
```

alcuadrado (approved):
I only left a comment in a changeset.

PR #3443, opened 2022-12-22
improvement: return error when prevrandao is missing for post-merge tx

Wodann:


crates/rethnet_evm_napi/Cargo.toml
```diff
@@ -9,7 +9,7 @@ crate-type = ["cdylib"]
 [dependencies]
 anyhow = "1.0.64"
 crossbeam-channel = { version = "0.5.6", default-features = false }
-napi = { version = "2.10.1", default-features = false, features = ["async", "error_anyhow", "napi8", "serde-json"] }
+napi = { version = "= 2.10.2", default-features = false, features = ["async", "error_anyhow", "napi8", "serde-json"] }
```
Wodann, line 12: I had to pin the `napi` version as a bug was introduced in `2.10.3` that breaks our build:
https://github.com/napi-rs/napi-rs/issues/1401

packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToRethnet.ts
```diff
@@ -164,6 +45,46 @@ export function ethereumjsBlockHeaderToRethnet(
...
+    case HardforkName.LONDON:
+      return SpecId.London;
+    case HardforkName.ARROW_GLACIER:
+      return SpecId.ArrowGlacier;
+    case HardforkName.GRAY_GLACIER:
+      return SpecId.GrayGlacier;
+    case HardforkName.MERGE:
+      return SpecId.Merge;
+    default:
+     const _exhaustiveCheck: never = hardfork;
```
fvictorio, line 81: I added this so that we get a compilation error if we add anew `HardforkName` but forget to update this switch.

fvictorio (approved):

PR #3459, opened 2022-12-23
Remove an unused variable

ilovehackathons:

---

In JavaScript, the array destructuring syntax allows you to skip variables at certain positions. The variable at position 0 is unused in the code example, therefore it should be skipped.

### DIFF
--- a/docs/src/content/tutorial/testing-contracts.md
+++ b/docs/src/content/tutorial/testing-contracts.md
@@ -100,7 +100,7 @@ describe("Token contract", function () {
   // ...previous test...
 
   it("Should transfer tokens between accounts", async function() {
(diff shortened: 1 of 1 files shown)


fvictorio:
Hmm, this makes sense, but I wonder if it wouldn't be better to keep the tutorial simpler since some people reading it might not be super familiar with javascript. What do you think?

alcuadrado:
Thanks for submitting this PR, @ilovehackathons! Unfortunately, I think we shouldn't merge it. Here's why:

> Hmm, this makes sense, but I wonder if it wouldn't be better to keep the tutorial simpler since some people reading it might not be super familiar with javascript. What do you think?

Agreed with this. As the purpose of the tutorial is educational, the easier it is to understand, the better.

PR #3483, opened 2022-12-28
feat: generate receipt for Rethnet transactions and validate

Wodann:


crates/rethnet_evm_napi/src/receipt.rs
```diff
@@ -0,0 +1,12 @@
+use napi::bindgen_prelude::{BigInt, Buffer};
+use napi_derive::napi;
+
+use crate::Log;
+
+#[napi(object)]
+pub struct Receipt {
```
fvictorio, line 7: This doesn't seem to be used, am I missing something?

Wodann: You're 100% right. We don't actually output the receipt from Rethnet (yet). I can remove it if you'd like?

fvictorio: No, it's fine if it's going to be used :)

fvictorio (approved):

PR #3498, opened 2022-12-30
Allow plugins to provide more artifacts

### LINKED ISSUE #1040: Allow plugins to provide more artifacts

fvictorio:
Closes #1040.

If someone wants to give this a try, please install `hardhat@2.13.0-dev.2`.


alcuadrado (approved):
LGTM! Great PR. Now this code is manageable again :)

I added a comment explaining a subtle thing in the CachingSource, and changed the order of priority between multiple sources. The latest should win.

Issue #3370, opened 2022-11-23
Relax node engines requirement


fvictorio:
While oficcially [we only support LTS versions of node.js](https://hardhat.org/hardhat-runner/docs/reference/stability-guarantees#node.js-versions-support), non-LTS versions are likely to work. But sometimes users can't use some of these versions (see #3341) because we have a strict list of engines in our `package.json`.

We should:
- [x] Drop the engines field from the `package.json`, or change it to something more permissive like `>=10.0.0` (I'm not sure about the trade-offs here)
- [x] Print a warning when a non-supported version is used (see [this comment](https://github.com/NomicFoundation/hardhat/issues/3341#issuecomment-1322529104) for an important implementation detail).

iamrekas:
Hey, @fvictorio i just open new PR for this issue. Please take a look

fvictorio:
This was fixed by #3507.

PR #3507, opened 2022-12-31
"Relaxing" node strict version

iamrekas:

---

This PR include fix for strict node version "relaxing" related to #3370 issue. 

Now it's strict to minimum version 14 LTS of node. 

Then running cli, it's show one of the errors:
- If using >=14.0.0 - what is using not LTS version (hardcoded LTS versions)
- If using <14.0.0 - the old warning about using not a supported node version.

Thank you for this project. It's awesome!

P.S. sorry for few my faults  - pure english and i don't test code, because i don't have for this moment node v18, only v19, so i need this to be included into production.


yannickrocks (approved):
LGTM

alcuadrado:
Thanks for submitting this PR. I changed a bit, so that this validation gets run before importing any Hardhat file. This way, if we use syntax that's not supported in the version of node that the user has, the warning still gets printed.

Issue #3500, opened 2022-12-30
Add a validation phase for the resolved configuration


fvictorio:
We validate the values provider in the user config, but not the ones that result after resolving the configuration (using our resolution and the config extenders). This is a blocker for #3323 and makes #3488 harder to do.

We should add something like `validateResolvedConfig(resolved)` after this:

https://github.com/NomicFoundation/hardhat/blob/a1a36f8dd03ab99e04c3edfc071eb32d6b906b39/packages/hardhat-core/src/internal/core/config/config-loading.ts#L107-L109

Validating the number of optimizer runs (#3488) is easy and low risk, so maybe that could be the first validation included there.

PR #3512, opened 2023-01-02
feat: validation added for the optimizer runs

konarshankar07:

---

https://github.com/NomicFoundation/hardhat/issues/3500

Validation added for resolved configs for optimizer runs

### DIFF
--- a/packages/hardhat-core/src/internal/core/config/config-validation.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-validation.ts
@@ -13,6 +13,7 @@ import { HardhatError } from "../errors";
 import { ERRORS } from "../errors-list";
 import { hardforkGte, HardforkName } from "../../util/hardforks";
 import { HardhatNetworkChainUserConfig } from "../../../types/config";
+import { HardhatConfig } from "../../../types";
 import { defaultHardhatNetworkParams } from "./default-config";
 
 function stringify(v: any): string {
@@ -572,4 +573,18 @@ export function getValidationErrors(config: any): string[] {
 
   const ioTsErrors = DotPathReporter.report(result);
   return [...errors, ...ioTsErrors];
+}
+
+export function validateResolvedConfig(resolvedConfig: HardhatConfig) {
+  console.log(resolvedConfig.solidity.compilers);
+  const soliditySettings: any[] = resolvedConfig.solidity.compilers.map(
+    (s) => s.settings
+  );
+  const optimiser: any[] = soliditySettings.map((o) => o.optimizer);
+  const runs: number[] = optimiser.map((r): number => r.runs);
+  if (runs[0] > 2 ** 32 - 1) {
+    throw new HardhatError(ERRORS.GENERAL.INVALID_CONFIG, {
+      errors: "Exceeded max number of optimizer runs.",
+    });
+  }
 }
--- a/packages/hardhat-core/src/internal/core/config/config-loading.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-loading.ts
@@ -19,7 +19,7 @@ import { getUserConfigPath } from "../project-structure";
 
 import { SUPPORTED_SOLIDITY_VERSION_RANGE } from "../../hardhat-network/stack-traces/constants";
 import { resolveConfig } from "./config-resolution";
-import { validateConfig } from "./config-validation";
+import { validateConfig, validateResolvedConfig } from "./config-validation";
 import { DEFAULT_SOLC_VERSION } from "./default-config";
 
 const log = debug("hardhat:core:config");
@@ -107,6 +107,8 @@ export function loadConfigAndTasks(
   for (const extender of HardhatContext.getHardhatContext().configExtenders) {
     extender(resolved, frozenUserConfig);
   }
+
+  validateResolvedConfig(resolved);
 
   if (showSolidityConfigWarnings) {
     checkUnsupportedSolidityConfig(resolved);


packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -573,3 +574,17 @@ export function getValidationErrors(config: any): string[] {
   const ioTsErrors = DotPathReporter.report(result);
   return [...errors, ...ioTsErrors];
 }
+
+export function validateResolvedConfig(resolvedConfig: HardhatConfig) {
+  console.log(resolvedConfig.solidity.compilers);
```
fvictorio, line 579: I guess you added this for debugging? We should remove it.

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -573,3 +574,17 @@ export function getValidationErrors(config: any): string[] {
...
+  const soliditySettings: any[] = resolvedConfig.solidity.compilers.map(
+    (s) => s.settings
+  );
+  const optimiser: any[] = soliditySettings.map((o) => o.optimizer);
+  const runs: number[] = optimiser.map((r): number => r.runs);
+  if (runs[0] > 2 ** 32 - 1) {
```
fvictorio, line 585: Personal preference, slightly more readable

```suggestion
  if (runs[0] >= 2 ** 32) {
```

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -573,3 +574,17 @@ export function getValidationErrors(config: any): string[] {
...
   return [...errors, ...ioTsErrors];
 }
+
+export function validateResolvedConfig(resolvedConfig: HardhatConfig) {
+  console.log(resolvedConfig.solidity.compilers);
+  const soliditySettings: any[] = resolvedConfig.solidity.compilers.map(
+    (s) => s.settings
+  );
+  const optimiser: any[] = soliditySettings.map((o) => o.optimizer);
+  const runs: number[] = optimiser.map((r): number => r.runs);
```
fvictorio, line 584: I believe this will throw if there are no optimizer settings. We should use a single map that does something like `settings?.optimizer?.runs` and returns `Array<number | undefined>`, and then we should handle the possible undefined value.

packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -573,3 +574,17 @@ export function getValidationErrors(config: any): string[] {
...
+  if (runs[0] > 2 ** 32 - 1) {
+    throw new HardhatError(ERRORS.GENERAL.INVALID_CONFIG, {
```
fvictorio, line 586: We should use a different error type here. I can do that for you if you want, since it's a bit more involved.

fvictorio (changes requested):
Thanks for the PR. I left some comments. Also, we should add tests for this.

konarshankar07:
Sorry @fvictorio  I forgot to convert to draft this PR and still working on the Unit test

packages/hardhat-core/src/internal/core/config/config-validation.ts
konarshankar07: @fvictorio I will try from my end and let you know if needed any help

packages/hardhat-core/src/internal/core/config/config-validation.ts
konarshankar07: Yes, My bad forgot to remove

packages/hardhat-core/src/internal/core/config/config-validation.ts
konarshankar07: I have checked and if we don't add the optimizer settings then the solidity compiler adding default values so I don't think that this will throw any error

fvictorio: The compiler will add default values when you execute `solc`, but this is Hardhat's config, that can have empty settings.

konarshankar07:
@fvictorio 
I have added the very basic error message if you could help me to optimize then it would be great

konarshankar07:
@fvictorio 

Any update on this?

packages/hardhat-core/test/internal/core/config/config-validation.ts
```diff
@@ -1812,4 +1815,25 @@ describe("Config validation", function () {
       });
     });
   });
+
+  describe("Resolved Config validation", function () {
```
fvictorio, line 1819: I think there should be several other test cases here:

- A similar test as the existing one, but iwth a valid number of runs
- One with `{ optimizer: enabled }` (no runs)
- One with the settings object but without the `optimizer` entry
- One where the invalid number of runs is in the overrides part of the config

There are other combinations, but that should be good enough.

fvictorio:
> I have added the very basic error message if you could help me to optimize then it would be great

I'll handle this part, thanks :+1:

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -252,6 +252,13 @@ To avoid it, please delete both your node_modules and package-lock.json, and rei
 Note that you don't need to do this every time you install a new dependency, but please make sure to delete your node_modules every time you delete your package-lock.json.`,
       shouldBeReported: false,
     },
+    INVALID_RESOLVED_CONFIG: {
```
fvictorio, line 255: Reminder for myself: update this.

packages/hardhat-core/test/internal/core/config/config-validation.ts
konarshankar07: @fvictorio  
I have added all the remaining tests

fvictorio:
Hey @konarshankar07, thanks a lot for working on this. We are going to make a couple of big releases in the coming days, and I want to postpone merging this until then, to minimize risk. I'll make a final review after those releases are out.

fvictorio:
Btw, there seem to be some linter errors, can you fix those?

fvictorio:
Thank you!

I ended up re-using the `INVALID_CONFIG` error code. I initally thought that having a different code could be useful, but then I realized that it would be confusing in cases like this, where the validation is only done for the resolved config.

I think ideally validation should be like this:

- Validate property X of the user config. If it's invalid, raise an `INVALID_CONFIG` error
- Resolve the config
- Check if property X is still valid. If not, raise a _different_ `INVALID_RESOLVED_CONFIG` error, explaining that this was probably caused by a plugin or an `extendConfig`

But this seems overkill in this case, since the thing we are validating is a very edge case.

Issue #3523, opened 2023-01-03
Add `version` field to the HRE


fvictorio:
It should be possible to get the version of Hardhat in runtime. Exposing it in the HRE seems like the obvios way to do it.

PR #3547, opened 2023-01-05
feat: hardhat version added in the HRE

konarshankar07:

---

https://github.com/NomicFoundation/hardhat/issues/3523

Added HardHat version in the HRE (HardHat runtime environment)

### DIFF
--- a/packages/hardhat-core/src/internal/core/runtime-environment.ts
+++ b/packages/hardhat-core/src/internal/core/runtime-environment.ts
@@ -20,6 +20,7 @@ import { Artifacts } from "../artifacts";
 import { MessageTrace } from "../hardhat-network/stack-traces/message-trace";
 import { lazyObject } from "../util/lazy";
 
+import { getHardhatVersion } from "../util/packageInfo";
 import { analyzeModuleNotFoundError } from "./config/config-loading";
 import { HardhatError } from "./errors";
 import { ERRORS } from "./errors-list";
@@ -31,8 +32,6 @@ import {
   createTaskProfile,
   TaskProfile,
 } from "./task-profiling";
-
-import { getHardhatVersion } from "../util/packageInfo";
 
 const log = debug("hardhat:core:hre");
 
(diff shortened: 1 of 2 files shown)


packages/hardhat-core/src/internal/core/runtime-environment.ts
```diff
@@ -50,6 +51,10 @@ export class Environment implements HardhatRuntimeEnvironment {
 
   public entryTaskProfile?: TaskProfile;
 
+  public version: string | null;
```
fvictorio, line 54: Why don't you just do `public version: getHardhatVersion()`?

fvictorio:
Thank you @konarshankar07. Can you add a test for this? Just checking that `hre.version` exists and it's a string would be enough, no need to check the actual value (since that will change all the time). 

I think you could add it as a couple of assertions [here](https://github.com/NomicFoundation/hardhat/blob/3461c8f119f7ed784a8324298fd565f6ff755c0e/packages/hardhat-core/test/internal/core/runtime-environment.ts#L144-L148).

packages/hardhat-core/src/internal/core/runtime-environment.ts
konarshankar07: Sorry @fvictorio  not able to understand how the return type will be function

fvictorio: Yeah, sorry, I meant `public version: string = getHardhatVersion()`.

This made me realize that it doesn't make sense for that function to return `string | null`, so I updated that. I also changed the builtin `help` task to use this new field instead of reading it from the package.json.

fvictorio:
@konarshankar07 please edit the description to say `Closes #3523`. That way, when this PR is merged, the issue will be automatically closed.

fvictorio (approved):

fvictorio:
Thank you @konarshankar07!

konarshankar07:
Thanks @fvictorio for helping me out :)

PR #3550, opened 2023-01-06
Basic support for ESM (rebased)

fvictorio:
This PR is almost the same as #3211, but with a better commit structure.

Commits:
- The first commit upgrades typescript and ts-node. This is an artifact of a previous attempt, where we added some basic support for typescript with ES modules. We rolled that back, but I kept the upgrades since it's useful to do it now and then anyway. This also includes two minor fixes related to these new versions.
- The second commit has the most important logic of the PR. It adds support for `hardhat.config.cjs` files and makes the necessary changes in the `test` task.
- The third commit has the tests for the changes in the second commit. I included them in a separate commit because there are several new fixture projects and that adds a lot of noise.
- The fourth commit improves how Hardhat's project creation works and shows different options when the `package.json` has `type: module`.

The other two commits add a new guide and the changeset.

To-do:
- [ ] ~Update Foundry guide to say that any option can be used. That is enabled by this PR because of the improvements to project initialization.~ Canceled, this didn't make sense in retrospect.
- [x] Use top-level await in the ESM+JS sample project


packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -252,6 +252,23 @@ To avoid it, please delete both your node_modules and package-lock.json, and rei
 Note that you don't need to do this every time you install a new dependency, but please make sure to delete your node_modules every time you delete your package-lock.json.`,
       shouldBeReported: false,
     },
+    ESM_PROJECT_WITHOUT_CJS_CONFIG: {
+      number: 19,
+      message: `Your project is an ESM project (you have "type": "module" set in your package.json) but your Hardhat config file uses a .js extension.
```
alcuadrado, line 257:
```suggestion
      message: `Your project is an ESM project (you have "type": "module" set in your package.json) but your Hardhat config file uses .js extension.
```

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -252,6 +252,23 @@ To avoid it, please delete both your node_modules and package-lock.json, and rei
...
+
+Rename the file to use a .cjs to fix this problem.`,
```
alcuadrado, line 259:
```suggestion
Rename the file to use .cjs to fix this problem.`,
```

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -252,6 +252,23 @@ To avoid it, please delete both your node_modules and package-lock.json, and rei
...
+      title: "Hardhat config with .js extension in an ESM project",
+      description:
+        "Your project is an ESM project but your Hardhat config uses an .js extension. Hardhat config files cannot be an ES module. To fix this, rename your Hardhat config to use the .cjs extension.",
```
alcuadrado, line 262:
```suggestion
        "Your project is an ESM project but your Hardhat config uses .js extension. Hardhat config files cannot be an ES module. To fix this, rename your Hardhat config to use the .cjs extension.",
```

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -983,6 +1000,14 @@ The first supported version is %firstSupportedVersion%`,
 Please use a newer, supported version.`,
       shouldBeReported: true,
     },
+    TEST_TASK_ESM_TESTS_RUN_TWICE: {
+      number: 609,
+      message: `Your project uses ESM and you've programmatically run your tests twice. This is not supported yet.`,
+      title: "Running tests twice in an ESM project",
+      description:
+        'You have run your tests twice programmatically and your project is an ESM project (you have `"type": "module"` in your `package.json`, or some of your files have an `.mjs` extension). This is not supported by Mocha yet.',
```
alcuadrado, line 1008:
```suggestion
        'You have run your tests twice programmatically and your project is an ESM project (you have `"type": "module"` in your `package.json`, or some of your files have the `.mjs` extension). This is not supported by Mocha yet.',
```

PR #3575, opened 2023-01-13
feat: add ForkDatabase

feuGeneA:
fixes https://github.com/NomicFoundation/rethnet/issues/9
- [x] In-memory caching


crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,241 @@
...
+use revm::{db::DatabaseRef, Account, AccountInfo, Bytecode};
+
+use rethnet_eth::{Address, B256, U256};
+
+use crate::db::{
+    layered_db::{LayeredDatabase, RethnetLayer},
+    remote::{RemoteDatabase, RemoteDatabaseError},
+};
+
+struct ForkDatabase {
```
Wodann, line 11: Based on linguistics, should this be `ForkingDatabase` or `ForkedDatabase`?

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,241 @@
...
+    layered_db: LayeredDatabase<RethnetLayer>,
+    remote_db: RemoteDatabase,
+    account_info_cache: HashMap<Address, AccountInfo>,
+    code_by_hash_cache: HashMap<B256, Bytecode>,
+    storage_cache: HashMap<(Address, U256), U256>,
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum ForkDatabaseError {
```
Wodann, line 20: Same as above, w.r.t. naming

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+impl revm::Database for ForkDatabase {
+    type Error = ForkDatabaseError;
+
+    fn basic(&mut self, address: Address) -> Result<Option<AccountInfo>, Self::Error> {
+        if let Some(cached) = self.account_info_cache.get(&address) {
+            Ok(Some(cached.clone()))
+        } else {
+            let to_be_cached = self
```
Wodann, line 51: If I'm not mistaken, the cache is only needed for the remote database.

The layered database is used for local changes and thus is already in cache

crates/rethnet_evm/src/db/fork.rs
feuGeneA: I'm happy to defer to your preference, but I felt that "forked database" sounds like you forked a database, but really you forked a node/chain, not a database, and it's your fork (noun, not past tense verb) of that node/chain that's serving as the backing for the database.

crates/rethnet_evm/src/db/fork.rs
feuGeneA: Right, this was an earlier draft, the latest draft only uses the cache for the remote database.

Wodann:
I merged your other PR. Would you be able to merge the latest `rethnet/main` to reduce the list of changes?

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,271 @@
...
+        if let Some(cached) = self.account_info_cache.get(&address) {
+            Ok(Some(cached.clone()))
+        } else if let Some(layered) = self
+            .layered_db
+            .basic(address)
+            .map_err(ForkDatabaseError::LayeredDatabase)?
+        {
+            Ok(Some(layered))
+        } else if let Some(remote) = self
+            .layered_db
```
feuGeneA, line 75: I had said there wasn't anything in [the n-api PR](https://github.com/NomicFoundation/hardhat/pull/3612) that touched `ForkDatabase`, but I remembered that there is this one small (but glaring haha) thing. This should be `remote_db`, not `layered_db`. I fixed it over [here](https://github.com/NomicFoundation/hardhat/pull/3612/commits/fc3047457fb1b50b732c34a3a358c1cdae375fe8) but haven't ported it back into this PR yet.

Wodann: Is it cherry pickable?

feuGeneA:
Yup just haven't gotten to it yet, will ASAP

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,271 @@
...
+        if self.layered_db.iter().next().is_some() {
+            Ok(self
+                .layered_db
+                .state_root()
+                .map_err(ForkDatabaseError::LayeredDatabase)?)
+        } else if let Some(cached) = self.fork_block_state_root_cache {
+            Ok(cached)
+        } else {
+            self.fork_block_state_root_cache = Some(
+                self.remote_db
```
Wodann, line 203: We might have to create a workaround for this. `state_root`s in Hardhat forked databases are pseudorandomly generated hashes.

This struct mimics the Hardhat implementation: https://github.com/NomicFoundation/hardhat/blob/rethnet/main/crates/rethnet_evm/src/random.rs#L8

And this is the original TypeScript implementation that you can track to see how it's used in Hardhat: https://github.com/NomicFoundation/hardhat/blob/rethnet/main/packages/hardhat-core/src/internal/hardhat-network/provider/utils/random.ts

Wodann (changes requested):

crates/rethnet_evm/src/db/layered_db.rs
```diff
@@ -250,12 +250,7 @@ impl Database for LayeredDatabase<RethnetLayer> {
         log::debug!("account with address `{}`: {:?}", address, account);
 
         // TODO: Move this out of LayeredDatabase when forking
-        Ok(account.or(Some(AccountInfo {
-            balance: U256::ZERO,
-            nonce: 0,
-            code_hash: KECCAK_EMPTY,
-            code: None,
-        })))
+        Ok(account)
```
Wodann, line 253: Feel free to remove the `TODO` comment.

You might have to move the `.or(..)` one layer up, in the `ForkDatabase`

crates/rethnet_evm/src/db/fork.rs
feuGeneA: yup, picked in https://github.com/NomicFoundation/hardhat/pull/3575/commits/359d3e67cdae3cba56e9355ddd0cf0e51e7bf7cf

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,271 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use std::str::FromStr;
+
+    fn get_alchemy_url() -> Result<String, String> {
```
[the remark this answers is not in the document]
feuGeneA: love it, thanks, addressed in https://github.com/NomicFoundation/hardhat/pull/3575/commits/4c4cf6b488a9b2dae35bdb7dcc2f633ac01a32a0

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,271 @@
...
+            fork_block_state_root_cache: None,
+        }
+    }
+}
+
+impl revm::Database for ForkDatabase {
+    type Error = ForkDatabaseError;
+
+    fn basic(&mut self, address: Address) -> Result<Option<AccountInfo>, Self::Error> {
+        if let Some(cached) = self.account_info_cache.get(&address) {
```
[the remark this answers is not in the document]
feuGeneA: Great catch, thank you, addressed in https://github.com/NomicFoundation/hardhat/pull/3575/commits/cc84b5306d5ac6cf6461584c2ec784fae2c84aee

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,271 @@
...
+            self.account_info_cache.insert(address, remote.clone());
+
+            if remote.code.is_some() {
+                self.code_by_hash_cache
+                    .insert(remote.code_hash, remote.code.clone().unwrap());
+            }
+
+            Ok(Some(remote))
+        } else {
+            Err(ForkDatabaseError::AccountNotFound(address))
```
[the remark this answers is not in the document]
feuGeneA: thanks, addressed in https://github.com/NomicFoundation/hardhat/pull/3575/commits/ea0cef498bfed77cfa57071d08cde36055a2dfc2

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,271 @@
...
+        }
+    }
+
+    fn code_by_hash(&mut self, code_hash: B256) -> Result<Bytecode, Self::Error> {
+        if let Some(cached) = self.code_by_hash_cache.get(&code_hash) {
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3575/commits/08af04b8e5c1df34cdde654e57003c0e548fab80

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,271 @@
...
+                .layered_db
+                .code_by_hash(code_hash)
+                .map_err(ForkDatabaseError::LayeredDatabase)?)
+
+            // remote_db doesn't support code_by_hash, so there's no delegation to it here.
+        }
+    }
+
+    fn storage(&mut self, address: Address, index: U256) -> Result<U256, Self::Error> {
+        if let Some(cached) = self.storage_cache.get(&(address, index)) {
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3575/commits/5e8fae89af86f58e343e16c80c60b608ccc56118

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,271 @@
...
+            .map_err(ForkDatabaseError::LayeredDatabase)
+    }
+
+    /// Modifies the account at the specified address using the provided function.
+    fn modify_account(
+        &mut self,
+        address: Address,
+        modifier: crate::debug::ModifierFn,
+    ) -> Result<(), Self::Error> {
+        self.layered_db
```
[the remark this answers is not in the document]
feuGeneA: in https://github.com/NomicFoundation/hardhat/pull/3575/commits/bacae573f6efc351ba913b8249a1a584f86b6249 i added some logic to first check if it's already in the layered db, and if not then check if it's in the remote db, and if it IS in the remote db then insert it into the layered db before calling modify. does that cover the situation you had in mind?

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,271 @@
...
+            .modify_account(address, modifier)
+            .map_err(ForkDatabaseError::LayeredDatabase)
+    }
+
+    /// Removes and returns the account at the specified address, if it exists.
+    fn remove_account(&mut self, address: Address) -> Result<Option<AccountInfo>, Self::Error> {
+        crate::DatabaseDebug::remove_account(&mut self.layered_db, address)
```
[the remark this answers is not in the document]
feuGeneA: That calls the `remove_account` method that's inherent to `LayeredDatabase<RethnetLayer>`, but it seemed to me that we wanted to delegate to the method on the `DatabaseDebug` trait, given that that's the trait we're implementing here, but honestly I'm a little confused about the difference between the two. Is it clear to you which one we want to be delegating to here? If not, I can dig a little to get a better understanding of that difference, and then make an informed decision.

crates/rethnet_evm/src/db/fork.rs
feuGeneA: Are you sure we need to do anything with that here in this PR? It seems to me that perhaps we'd want to use that within the `LayeredDatabase`, when new layers are created, but that when referring to historical blocks (fork block or prior) we'd want to just use the actual state root from the block.

crates/rethnet_evm/src/db/layered_db.rs
feuGeneA: thanks, removed in https://github.com/NomicFoundation/hardhat/pull/3575/commits/3607c46a64fe31c608a501413cdbfa4c108951fb.

i'm not sure what you meant by moving up the `.or(..)`. is that not covered by the `else { Ok(None) }` branch in `basic()`?

crates/rethnet_evm/src/db/fork.rs
```diff
@@ -0,0 +1,298 @@
...
+    fn commit(&mut self, changes: HashMap<Address, Account>) {
+        self.layered_db.commit(changes)
+    }
+}
+
+impl crate::DatabaseDebug for ForkDatabase {
+    type Error = ForkDatabaseError;
+
+    fn account_storage_root(&mut self, address: &Address) -> Result<Option<B256>, Self::Error> {
+        self.layered_db
```
Wodann, line 136: I just noticed that we're not handling cases where the account only exists in the fork block.

You can handle that by checking `if let Some(..)` and otherwise returning the storage root for the fork block.

crates/rethnet_evm/src/db/fork.rs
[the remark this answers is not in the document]
Wodann: That's perfect! It exactly handles the situation I mentioned.

crates/rethnet_evm/src/db/fork.rs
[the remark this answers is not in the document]
Wodann: Thanks to you I just realised that the two methods diverge in implementation. I've created a fix for that in a separate PR, so feel free to leave your implementation as is for now.

crates/rethnet_evm/src/db/fork.rs
Wodann: The logic should be independent from `LayeredDatabase` as we only need pseudorandom root hashes when using `ForkDatabase`. When just using `LayeredDatabase` we can always calculate our state root hashes.

When combining the `LayeredDatabase` with a remote database, it needs to somehow create a unique root hash for each layer that takes into account the "history" as well.

This is what the `ForkStateManager` does in Hardhat:

https://github.com/NomicFoundation/hardhat/blob/rethnet/main/packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkStateManager.ts#L272-L278

crates/rethnet_evm/src/db/layered_db.rs
Wodann: Feel free to look into this in your `napi` PR if you run into that problem there

feuGeneA: I'll leave this as-is for now, and if it's a problem in the broader test run (`yarn test:forking`) then I'll fix it over in [the n-api PR](https://github.com/NomicFoundation/hardhat/pull/3612)

crates/rethnet_evm/src/db/fork.rs
feuGeneA: The reason I left this out is because there is no way to retrieve the storage _root_ via the JSON-RPC API. One can obtain the storage value at a particular index (via `eth_getStorageAt`), but from what I can tell there is no API for retrieving the storage root for an account (only the broader state root for a block).

Wodann: I just checked the fork state manager and that leaves it undefined as well, so indeed no need to implement it.

crates/rethnet_evm/src/state/layered_state.rs
```diff
@@ -255,12 +256,7 @@ impl State for LayeredState<RethnetLayer> {
             .flatten();
 
         // TODO: Move this out of LayeredState when forking
```
Wodann, line 258: Please remove the `// TODO` as it has been handled

Wodann (approved):

PR #3578, opened 2023-01-13
Fix #1324 (WebSocket BatchRequest) supports Batch requests via WebSockets

### LINKED ISSUE #1324: Support batch JsonRPC requests for Websockets
Hi. I noticed that batch requests seem to only be supported when using http and not websockets. Looking at [handler.ts](https://github.com/nomiclabs/hardhat/blob/master/packages/hardhat-core/src/internal/hardhat-network/jsonrpc/handler.ts) it seems like a trivial fix. Since I need this for my project I will fork hardhat and implement it. Are you interested in a PR for this?

tenbits:
Hello Team, 

fixing here the #1324 issue. BatchRequests are now supported also via WebSockets. Unit tests are attached.

### DIFF
--- a/.changeset/plenty-comics-bow.md
+++ b/.changeset/plenty-comics-bow.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Add BatchRequest support for WebSocket server
+Added support for sending batch requests through WebSocket to the Hardhat node (thanks @tenbits!)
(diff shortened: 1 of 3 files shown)


fvictorio:
Thanks a lot @tenbits! I'll try to review this next week.

fvictorio:
This is great, thank you @tenbits!

tenbits:
Thanks, Franco, for your time and for looking into this.

PR #3591, opened 2023-01-18
chore: run clippy & rustfmt in CI

Wodann:


.github/workflows/rethnet-ci.yml
```diff
@@ -6,13 +6,15 @@ on:
...
     paths:
+      - ".github/workflows/rethnet-ci.yml"
       - "config/**"
       - "crates/**"
       - "Cargo.toml"
       - "rust-toolchain"
   pull_request:
     branches: ["**"]
     paths:
+      - ".github/workflows/rethnet-ci.yml"
```
fvictorio, line 17: Wait, is this necessary? I mean, it does make sense. Maybe we should do it in the other workflows too.

Wodann: Yes, otherwise it would not re-run the CI for rethnet-ci

PR #3593, opened 2023-01-18
feat: tracing in Rethnet using JS callbacks

Wodann:
The stack trace implementation has some caveats:

- `traceTransaction` is not implemented for Rethnet. This means that there are still 10 failing tests in `HARDHAT_EXPERIMENTAL_VM_MODE=rethnet` due to `Error: traceTransaction not implemented for Rethnet`
- The `"Before byzantium"` test is skipped
- There are 7 failing tests in dual adapter mode: 5 in stack traces, 2 (1x Hardhat Network Provider & 1x JSON-RPC provider) in `eth_sendTransaction > when automine is enabled > Shoud throw if the transaction fails`; all due to differences in the number of stack trace steps. We agreed that it's a time sync to try and resolve them now

In the future we need to revisit the `JsTracer` implementation to remove ethereumjs specific fixes. This will be a breaking change for the Hardhat tracer ABI, but makes the implementation more "sane".


crates/rethnet_evm_napi/src/tracer/js_tracer.rs
```diff
@@ -0,0 +1,664 @@
...
+        _is_static: bool,
+    ) -> (Return, Gas, Bytes) {
+        let (sender, receiver) = channel();
+
+        let status = self.after_message_fn.call(
+            AfterMessageHandlerCall {
+                result: rethnet_evm::ExecutionResult {
+                    exit_reason: ret,
+                    out: rethnet_evm::TransactOut::Call(out.clone()),
+                    gas_used: if ret == Return::InvalidOpcode || ret == Return::OpcodeNotFound {
```
Wodann, line 507: I had to do this to match ethereumjs (same pattern below). What do you think, @fvictorio, should this be handled in `revm` or here?

fvictorio: Ugh, good question, I don't know. It seems to me like something revm should do, but maybe there's a good reason to do it this way. It's worth opening an issue to discuss it though.

Wodann: https://github.com/bluealloy/revm/issues/335

crates/rethnet_eth/src/remote.rs
```diff
@@ -393,24 +393,23 @@ impl RpcClient {
...
+
+    fn get_alchemy_url() -> String {
+        std::env::var_os("ALCHEMY_URL")
             .expect("ALCHEMY_URL environment variable not defined")
             .into_string()
-            .expect("couldn't convert OsString into a String"))
+            .expect("Couldn't convert OsString into a String")
     }
 
+    #[test_with::env(ALCHEMY_URL)]
```
Wodann, line 409: @feuGeneA I added a macro that ignores tests when the `ALCHEMY_URL` environment variable is not set. This allows people have local tests succeed even if they don't have alchemy setup

fvictorio (approved):

PR #3601, opened 2023-01-21
Website UI improvements: hero x-index order and sidebar scrollbar

ajmeese7:

---

I changed the `z-index` of the hero element so this doesn't happen:

ethereum logo over text

And configured the sidebar to not have a scrollbar when one isn't necessary, like so:

unnecessary sidebar

### DIFF
--- a/docs/src/components/landingBlocks/HeroBlock.tsx
+++ b/docs/src/components/landingBlocks/HeroBlock.tsx
@@ -45,7 +45,7 @@ const Block = styled.div`
   padding: 0 0 24px;
   min-height: 100px;
   &.content {
-    z-index: 1;
+    z-index: 2;
   }
   & svg {
     margin: 0 auto;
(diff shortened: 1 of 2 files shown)


fvictorio:
Hey @ajmeese7, thanks a lot for this! The removal of the scrollbar is great, but the other change seems to make things worse for some widths.

Before:

image

After:

image

ajmeese7:
Good catch! Do you want a height-based media query rule for this or would you rather the change be reverted entirely?

docs/src/components/landingBlocks/HeroBlock.tsx
```diff
@@ -45,7 +45,7 @@ const Block = styled.div`
   padding: 0 0 24px;
   min-height: 100px;
   &.content {
-    z-index: 1;
+    z-index: 2;
```
alcuadrado, line 48:
```suggestion
    z-index: 1;
```

alcuadrado:
They, thanks for making this change! I reverted the z-index bit to get this merged quickly.

I don't understand how the media query solution would work, but if it does improve, feel free to send another PR :)

PR #3602, opened 2023-01-23
improvement: update to latest revm

Wodann:


packages/hardhat-core/src/internal/hardhat-network/stack-traces/message-trace.ts
```diff
@@ -91,3 +98,34 @@ export type MessageTraceStep = MessageTrace | EvmStep;
 export interface EvmStep {
   pc: number;
 }
+
+export function isCallOutput(
```
Wodann, line 102: What's the best place for these helper functions?

Wodann:
Merged into PR #3593

PR #3612, opened 2023-01-25
feat: implement fork state management

feuGeneA:
- Resolves https://github.com/NomicFoundation/rethnet/issues/9
- Resolves https://github.com/NomicFoundation/rethnet/issues/30


packages/hardhat-core/test/internal/hardhat-network/provider/utils/runFullBlock.ts
```diff
@@ -69,57 +67,27 @@ export async function runFullBlock(
...
-
-async function runBlockAndGetAfterBlockEvent(
-  vm: VM,
-  runBlockOpts: RunBlockOpts
-): Promise<AfterBlockEvent> {
-  let results: AfterBlockEvent;
-
-  function handler(event: AfterBlockEvent) {
-    results = event;
+  const vm = forkedNode["_vm"];
```
Wodann, line 70: @fvictorio could you please have a look at this too?

packages/hardhat-core/test/internal/hardhat-network/provider/utils/runFullBlock.ts
```diff
@@ -69,57 +67,27 @@ export async function runFullBlock(
...
+
+  const blockBuilder = new BlockBuilder(vm, common, {
+    parentBlock,
+    headerData: block.header,
+  });
+  await blockBuilder.startBlock();
+
+  for (const [i, tx] of block.transactions.entries()) {
+    console.log(
```
Wodann, line 79: Still has `console.log`

crates/rethnet_evm/src/trace.rs
```diff
@@ -17,7 +17,7 @@ pub struct Trace {
 /// A single EVM step.
 pub struct Step {
     /// The executed op code
-    pub opcode: u8,
+    pub op: u8,
```
Wodann, line 20: Is there any reason for renaming this?

crates/rethnet_evm/src/state/fork.rs
```diff
@@ -1,82 +1,112 @@
...
+                    .expect("failed to get latest block")
+                    .number
+                    .map(U256::from)
+            })
+            .unwrap();
+
+        let remote_state = RemoteState::new(url, fork_block_number);
+
+        let mut initialized_accounts = accounts.clone();
+        for (address, account_info) in accounts.iter() {
```
Wodann, line 66:
```suggestion
        let initialized_accounts = accounts.into_iter().map(|(address, account_info)| {
            (*address, AccountInfo { ... })
        }).collect();
```

crates/rethnet_evm/src/state/fork.rs
```diff
@@ -1,82 +1,112 @@
...
+                    nonce: remote_account_info.nonce,
+                    ..account_info.clone()
+                },
+            );
+        }
 
-        let layered_db = LayeredDatabase::<RethnetLayer>::default();
+        let mut layered_state =
+            LayeredState::with_layer(RethnetLayer::with_genesis_accounts(initialized_accounts));
+        crate::state::StateDebug::checkpoint(&mut layered_state).unwrap();
```
Wodann, line 83: This introduces a discrepancy with how the `LayeredState` works. It assumes that the starting layer is initialized with the genesis accounts and no checkpoint is created. The `napi` wrapper calls checkpoint to ensure that genesis block is always checkpointed. We should do the same in both places.

In the case of the `LayeredState` it's documented. `with_layer` creates the database with the provided layer as its only layer - ergo no checkpoint.

crates/rethnet_evm/src/state/fork.rs
```diff
@@ -123,19 +168,17 @@ impl revm::Database for ForkDatabase {
...
+impl revm::DatabaseCommit for ForkState {
     fn commit(&mut self, changes: HashMap<Address, Account>) {
-        self.layered_db.commit(changes)
+        self.layered_state.commit(changes)
     }
 }
 
-impl crate::DatabaseDebug for ForkDatabase {
-    type Error = ForkDatabaseError;
+impl crate::state::debug::StateDebug for ForkState {
```
Wodann, line 177: `use crate::state::debug::StateDebug` at the top of the file. This also removes the necessity for the verbose usage of full paths to call trait functions

crates/rethnet_evm/src/state/fork.rs
```diff
@@ -203,60 +224,80 @@ impl crate::DatabaseDebug for ForkDatabase {
...
+        block_number: U256,
+        state_root: &B256,
+    ) -> Result<(), Self::Error> {
+        self.state_root_to_block_number
+            .insert(*state_root, block_number);
+        self.set_state_root(state_root)?;
+        Ok(())
+    }
+
+    fn restore_fork_block_context(&mut self, state_root: &B256) -> Result<(), Self::Error> {
```
Wodann, line 242: Why not just call `set_state_root` instead of adding a new function that does the same?

Wodann (changes requested):

crates/rethnet_eth/src/remote.rs
```diff
@@ -231,24 +231,70 @@ impl RpcClient {
         Ok(success.result)
     }
 
+    fn hash_string(input: &str) -> String {
+        use std::hash::Hasher;
+        let mut hasher = std::collections::hash_map::DefaultHasher::new();
```
Wodann, line 236: The hashbrown hashmap is faster, so I recommend using their hashing algorithm - if it suffices the requirements of your use case

crates/rethnet_eth/src/remote.rs
```diff
@@ -231,24 +231,70 @@ impl RpcClient {
...
+        hasher.write(input.as_bytes());
+        hasher.finish().to_string()
+    }
+
+    fn make_cache_path(&self, request_body: &str) -> std::path::PathBuf {
+        // TODO: consider using a better path for this directory. currently, for test runs,
```
Wodann, line 242: I'd let the end user optionally specify the caching directory. If they don't specify anything use your default. That should be straightforward to explain in the comments of the API

crates/rethnet_eth/src/remote.rs
```diff
@@ -231,24 +231,70 @@ impl RpcClient {
...
+            .create(directory.clone())
+            .expect("failed to create on-disk RPC response cache");
+
+        std::path::Path::new(&directory).join(format!("{}.json", Self::hash_string(request_body)))
+    }
+
+    fn read_response_from_cache(&self, request_body: &str) -> Option<String> {
+        if let Ok(mut file) = std::fs::File::open(&self.make_cache_path(request_body)) {
+            let mut response = String::new();
+            if std::io::Read::read_to_string(&mut file, &mut response).is_ok() {
```
Wodann, line 259: https://doc.rust-lang.org/std/fs/fn.read_to_string.html

crates/rethnet_eth/src/remote.rs
```diff
@@ -231,24 +231,70 @@ impl RpcClient {
...
+        std::fs::DirBuilder::new()
+            .recursive(true)
+            .create(directory.clone())
+            .expect("failed to create on-disk RPC response cache");
+
+        std::path::Path::new(&directory).join(format!("{}.json", Self::hash_string(request_body)))
+    }
+
+    fn read_response_from_cache(&self, request_body: &str) -> Option<String> {
+        if let Ok(mut file) = std::fs::File::open(&self.make_cache_path(request_body)) {
```
Wodann, line 257: `use std::fs::File;`

Wodann (changes requested):

.github/workflows/hardhat-network-forking-rethnet-ci.yml
```diff
@@ -0,0 +1,84 @@
...
+
+jobs:
+  # We should run this tests with multiple configurations
+  # but somehow the requests to Alchemy take much longer when doing that.
+  # As a temporary workaround, we run them with a single config.
+  #
+  # Once we properly refactor Hardhat Network's tests we should
+  # add more configurations (different OS and node versions).
+  #
+  # We should also understand what's going on before blindly refactoring them.
```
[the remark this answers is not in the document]
feuGeneA: This is just an old comment from the duplication of `hardhat-network-forking-ci.yml`. See https://github.com/NomicFoundation/hardhat/pull/3612#discussion_r1146431954

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
```diff
@@ -610,91 +927,108 @@ function assertEqualRunTxResults(
...
-
-  if (!ethereumJSAccount.codeHash.equals(rethnetAccount.codeHash)) {
-    console.trace(
-      `Different codeHash: ${ethereumJSAccount.codeHash} !== ${rethnetAccount.codeHash}`
+      `Different balance: ${ethereumJSAccount.balance} (ethereumjs) !== ${rethnetAccount.balance} (rethnet)`
     );
-    throw new Error("Different codeHash");
+    differences.push("balance");
   }
+  const inGetBalanceContext =
```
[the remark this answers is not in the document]
feuGeneA: This is sort of a hack but I thought potentially a reasonable one. This is determining whether this check is running within the context of an `eth_getBalance` call. If it is, it skips all of the account checks except for balance, since in this case the user really only cares about the balance, not any of the other fields.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
```diff
@@ -59,6 +69,24 @@ export class DualModeAdapter implements VMAdapter {
       selectHardfork
     );
 
+    // if the fork node config doesn't specify a fork block number, then the
+    // ethereum-js VM used the latest block number from the forked chain. here
+    // we get that number from that VM so it can be passed into the rethnet VM;
+    // otherwise, the rethnet VM's determination of "latest" might be a later
+    // number, resulting in a different fork block.
```
fvictorio, line 76: I found this explanation a bit hard to understand. This is my attempt to re-write it, but feel free to ignore it.

```suggestion
    // If the fork node config doesn't specify a block number, then the
    // ethereumjs adapter will fetch and use the latest block number. We re-use
    // that value here; otherwise, rethnet would also fetch it and we could have
    // a race condition if the latest block changed in the meantime.
```

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
```diff
@@ -59,6 +69,24 @@ export class DualModeAdapter implements VMAdapter {
...
+    if (
+      isForkedNodeConfig(config) &&
+      config.forkConfig.blockNumber === undefined
+    ) {
+      const forkBlockNumber = ethereumJSAdapter.getForkBlockNumber();
+      if (forkBlockNumber !== undefined) {
```
fvictorio, line 82: Can this happen? Shouldn't this be an assertion error?

packages/hardhat-core/src/internal/hardhat-network/stack-traces/opcodes.ts
```diff
@@ -306,6 +306,10 @@ export enum Opcode {
   SELFDESTRUCT = 0xff,
 }
 
+export function opcodeName(opcode: number): string {
+  return opcodeNames[opcode];
```
fvictorio, line 310: Typescript enums are two-way, so you can remove the `opcodeNames` array and do:

```suggestion
  return Opcode[opcode] ?? `<unrecognized opcode ${opcode}>`;
```

Wodann: 🤦 

So much work 😭

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -0,0 +1,777 @@
...
+            RpcClientError::InterpretationError {
+                msg: err.to_string(),
+                expected_type: std::any::type_name::<BatchResult>(),
+                body: response.text.to_owned(),
+            }
+        })?;
+
+        debug_assert_eq!(results.0.id, response.request_ids[0]);
+        debug_assert_eq!(results.1.id, response.request_ids[1]);
+        debug_assert_eq!(results.2.id, response.request_ids[2]);
```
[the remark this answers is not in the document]
Wodann: Thanks! I fixed it by sorting the response IDs. That should guarantee that the received responses are mapped to the right requests.

packages/hardhat-core/test/internal/hardhat-network/provider/utils/runFullBlock.ts
Wodann: This was originally done by Gene. I just commented that he left a `console.log` entry. I'll just remove it myself 😁

packages/hardhat-core/test/internal/hardhat-network/provider/utils/runFullBlock.ts
Wodann: I was unfamiliar with the previous test and since Gene changed it, I wanted you to validate the TS

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
Wodann: I instead did `forkBlockNumber`. I'll trust you that it's impossible within the control flow of the hardhat fork code 😅

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
Wodann: Replace Gene's explanation with yours. I believe it's correct, but @feuGeneA could you please verify that Franco's comments still carry the same meaning as you intended.

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -0,0 +1,777 @@
...
+    fn extract_response<T>(response: &str, request_id: &jsonrpc::Id) -> Result<T, jsonrpc::Error>
+    where
+        T: for<'a> serde::Deserialize<'a>,
+    {
+        let response: jsonrpc::Response<T> =
+            serde_json::from_str(response).unwrap_or_else(|error| {
+                panic!(
+                    "Response `{response}` failed to parse with expected type `{expected_type}`, due to error: {error}",
+                    expected_type = std::any::type_name::<T>()
+                )
```
feuGeneA, line 76: With this new response handling, it seems there's some information getting lost.

If you tweak the `get_tx_by_hash_success` test to have a bogus hash (just arbitrarily change one of the digits), the server responds with just `{"jsonrpc":"2.0","id":0,"result":null}`.

The way I had it before, that response was conveyed as: `failed to get transaction by hash: InterpretationError { msg: "invalid type: null, expected struct Transaction at line 1 column 37", request_body: "{\"version\":\"2.0\",\"method\":\"eth_getTransactionByHash\",\"params\":[\"0xf008e9f9bb92057dd0035496fbf4fb54f66b4b18b370928e46d6603933054d5a\"],\"id\":0}", expected_type: "rethnet_eth::remote::jsonrpc::Success<rethnet_eth::remote::eth::Transaction>", response_text: "{\"jsonrpc\":\"2.0\",\"id\":0,\"result\":null}" }', crates/rethnet_eth/src/remote.rs:516:14`

With the revised error handling, now you just get: ```thread 'remote::client::tests::get_tx_by_hash_success' panicked at 'Response `{"jsonrpc":"2.0","id":0,"result":null}` failed to parse with expected type `rethnet_eth::remote::eth::Transaction`, due to error: data did not match any variant of untagged enum ResponseData at line 1 column 38', crates/rethnet_eth/src/remote/client.rs:73:17```

With the new way, we can't see what the request was, nor where it originated (in the new way, the panic originates from the internal function rather than the site of usage (eg the test)). I think it would be good to restore those two pieces of information to the error message.

feuGeneA: I added a test just to be clear what i mean. it seems to me that the case exercised by that test should be a recoverable error rather than a `panic`. let me know what you think.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
feuGeneA: LGTM!

crates/rethnet_evm/src/block/builder.rs
```diff
@@ -165,6 +165,12 @@ where
                 AccountModifierFn::new(Box::new(move |balance, _nonce, _code| {
                     *balance += reward;
                 })),
+                &|| {
+                    Ok(AccountInfo {
+                        code: None,
+                        ..AccountInfo::default()
```
fvictorio, line 171: What's (conceptually) the difference between `code: None` and `code: Some(Bytecode::new())`? Or, in other words, why don't we just do `Ok(AccountInfo::default())` here?

crates/rethnet_eth/src/withdrawal.rs
```diff
@@ -0,0 +1,41 @@
+use revm_primitives::{Address, U256};
+use ruint::aliases::U128;
+
+/// Ethereum withdrawal
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct Withdrawal {
```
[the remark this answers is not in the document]
Wodann: Moved to the `remote` module in https://github.com/NomicFoundation/hardhat/pull/3612/commits/e289615ff95d233d828540810bf84cd074d76178

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -0,0 +1,828 @@
...
+        let mut hasher = std::collections::hash_map::DefaultHasher::new();
+        hasher.write(input.as_bytes());
+        hasher.finish().to_string()
+    }
+
+    fn make_cache_path(&self, request_body: &str) -> std::path::PathBuf {
+        // TODO: consider using a better path for this directory. currently, for test runs,
+        // it's going to crates/rethnet_eth/remote_node_cache. shouldn't it be rooted somewhere
+        // more accessible/convenient?
+        let directory = format!("remote_node_cache/{}", Self::hash_string(&self.url));
```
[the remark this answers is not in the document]
Wodann: Created an issue for this: https://github.com/NomicFoundation/rethnet/issues/82

.github/workflows/hardhat-core-ci.yml
```diff
@@ -24,64 +38,280 @@ concurrency:
...
+
+      - name: Install package
         run: yarn --frozen-lockfile
+
       - name: Build
         run: yarn build
+
       - name: Run tests
         env:
+          HARDHAT_EXPERIMENTAL_VM_MODE: ${{ matrix.vm }}
```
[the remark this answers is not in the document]
Wodann: Thanks :)

crates/rethnet_evm/Cargo.toml
```diff
@@ -12,6 +12,7 @@ cita_trie = { git = "https://github.com/Wodann/cita-trie", rev = "60efef5", vers
 hashbrown = { version = "0.13", default-features = false, features = ["ahash", "serde"] }
 hasher = { git = "https://github.com/Wodann/hasher", rev = "89d3fc9", version = "0.1.4", default-features = false, features = ["hash-keccak"] }
 log = { version = "0.4.17", default-features = false }
+once_cell = "1.17.1"
```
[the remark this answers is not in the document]
Wodann: No, thanks for spotting that!

Removed in https://github.com/NomicFoundation/hardhat/pull/3612/commits/dc581d1dd05bcbf4c0124978d9a9478c8b07aa48

crates/rethnet_evm/src/block/builder.rs
Wodann: The `code` field is intended to signal whether we're locally storing the code or - if `None` whether we're caching it elsewhere. We always store it separately, so I just want to make sure to adhere to this convention.

I am planning to propose a change to `revm` to make the separation clearer using an enum:

```rust
pub enum CodeRef {
  Local(Bytecode),
  Ref(B256),
}
```

that way we also avoid the duplication of having a stored `code_hash` when it's also stored in the `Bytecode`

packages/hardhat-core/src/internal/hardhat-network/provider/RethnetState.ts
```diff
@@ -29,14 +30,17 @@ export class RethnetStateManager {
...
     forkConfig: ForkConfig,
     genesisAccounts: GenesisAccount[]
-  ): RethnetStateManager {
+  ): Promise<RethnetStateManager> {
     let blockNumber;
     if (forkConfig.blockNumber !== undefined) {
       blockNumber = BigInt(forkConfig.blockNumber);
+    } else {
+      const { forkBlockNumber } = await makeForkProvider(forkConfig);
+      blockNumber = forkBlockNumber;
```
feuGeneA, line 43: Just out of curiosity: why get the latest fork block number out here instead of within the rust side of things?

crates/rethnet_eth/src/remote.rs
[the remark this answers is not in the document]
Wodann: Did this refactor myself

crates/rethnet_evm/src/state/fork.rs
```diff
@@ -1,219 +1,154 @@
...
-            Ok(Some(layered))
-        } else if let Some(remote) = self.get_remote_account_info(&address, *block_number)? {
-            Ok(Some(remote))
+        if let Some(local) = self.local_state.basic(address)? {
+            Ok(Some(local))
         } else {
-            Ok(None)
+            self.remote_state.lock().basic(address)
         }
     }
```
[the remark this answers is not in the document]
Wodann: Yes, in that case the `local_state` will be set to the initial state and result in no result found. Thus it will always fall back to the `else` statement

crates/rethnet_evm/src/state/fork.rs
```diff
@@ -1,219 +1,154 @@
...
 
     fn storage(&self, address: Address, index: U256) -> Result<U256, Self::Error> {
-        let layered = self.layered_state.storage(address, index)?;
-
-        if layered != U256::from(0) {
-            Ok(layered)
-        } else if let Some(cached) = self.storage_cache.read().get(&(address, index)).cloned() {
-            Ok(cached)
+        if let Ok(local) = self.local_state.storage(address, index) {
+            Ok(local)
```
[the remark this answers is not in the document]
Wodann: From what I can see, you're pointing to an outdated version of the code. I'm using a `HashSet` of zeroed (address, index) pairs to lookup which storage slots were specifically set to zero

packages/hardhat-core/src/internal/hardhat-network/provider/RethnetState.ts
Wodann: The logic that's used to determine the block number is very specific and already exists on the EthereumJS side. It was safer to just re-use that instead of trying to reproduce the code to do so.

The previous code was causing issues with block number mismatches

crates/rethnet_evm/src/block/builder.rs
fvictorio: Makes sense, thanks!

crates/rethnet_evm/src/state.rs
```diff
@@ -31,14 +33,22 @@ pub enum StateError {
...
     InvalidCodeHash(B256),
     /// Specified state root does not exist
-    #[error("State root `{0:?}` does not exist.")]
-    InvalidStateRoot(B256),
+    #[error("State root `{state_root:?}` does not exist (fork: {fork_identifier}).")]
+    InvalidStateRoot {
+        /// Requested state root
+        state_root: B256,
+        /// Whether the state root was intended for a fork
+        fork_identifier: bool,
```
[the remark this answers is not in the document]
Wodann: Good call! Applied change

crates/rethnet_evm/src/state.rs
```diff
@@ -31,14 +33,22 @@ pub enum StateError {
...
+    },
+    /// Storage slot with specified index does not exist
+    #[error("Storage slot with index `{0}` does not exist.")]
+    InvalidStorageSlot(U256),
```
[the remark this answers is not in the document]
Wodann: This error wasn't being used any more, so I removed it.

crates/rethnet_evm/src/state/remote/cached.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::state::{account::RethnetAccount, StateError};
+
+use super::RemoteState;
+
+/// A cached version of [`RemoteState`].
+#[derive(Debug)]
+pub struct CachedRemoteState {
+    remote: RemoteState,
+    account_cache: HashMap<U256, HashMap<Address, RethnetAccount>>,
+    code_cache: HashMap<U256, HashMap<B256, Bytecode>>,
```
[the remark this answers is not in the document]
Wodann: I added comments

crates/rethnet_evm_napi/src/context.rs
```diff
@@ -32,11 +36,20 @@ impl RethnetContext {
...
         })
     }
+
+    /// Skips a hash in the generator
+    #[napi]
+    pub fn set_hash_generator_seed(&self, seed: Buffer) {
+        let seed = B256::from_slice(&seed);
+
+        self.inner.hash_generator.lock().set_next(seed);
+    }
```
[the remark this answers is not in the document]
Wodann: Good catch. Updated

Wodann (approved):

Wodann:
I addressed all outstanding comments and resolved merge conflicts.

The only way to get rid of my original "suggestions requested" status was to change it into approved. Waiting for others to finalise any additional comments before merging

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -372,66 +287,441 @@ impl RpcClient {
...
+    ) -> Result<Option<eth::Block<eth::Transaction>>, RpcClientError> {
+        self.call(&MethodInvocation::BlockByHash(*hash, include_transactions))
+            .await
+    }
+
+    /// eth_getBlockByNumber
+    pub async fn get_block_by_number(
+        &self,
+        number: BlockSpec,
+        include_transactions: bool,
```
feuGeneA, line 336: passing `false` for `include_transactions` causes a `panic`

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -372,66 +287,441 @@ impl RpcClient {
...
+            code,
+            nonce,
+        })
+    }
+
+    /// eth_getBlockByHash
+    pub async fn get_block_by_hash(
+        &self,
+        hash: &B256,
+        include_transactions: bool,
```
feuGeneA, line 326: passing `false` for `include_transactions` causes a `panic`

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -372,66 +287,441 @@ impl RpcClient {
...
+        let account_info = RpcClient::new(&alchemy_url)
+            .get_account_info(&dai_address, BlockSpec::Tag("latest".to_string()))
+            .await
+            .expect("should have succeeded");
+
+        assert_eq!(account_info.balance, U256::from(0));
+        assert_eq!(account_info.nonce, 1);
+        assert_ne!(account_info.code_hash, KECCAK_EMPTY);
+        assert!(account_info.code.is_some());
+    }
```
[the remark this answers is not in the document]
Wodann: https://ethereum.org/en/developers/docs/smart-contracts/upgrading/

Contracts are immutable

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -752,77 +965,66 @@ mod tests {
...
-            .get_account_info(&dai_address, BlockSpec::Number(U256::from(16220843)))
+        let total_supply = RpcClient::new(&alchemy_url)
+            .get_storage_at(
+                &dai_address,
+                U256::from_str_radix(
+                    "0000000000000000000000000000000000000000000000000000000000000001",
+                    16,
+                )
+                .expect("failed to parse storage location"),
+                BlockSpec::Tag("latest".to_string()),
```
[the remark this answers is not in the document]
Wodann: I removed the test, as storage slots can indeed change

crates/rethnet_eth/src/remote/client.rs
Wodann: Good catch. I added the proper parsing code for dealing with transactions being hashes

.github/workflows/hardhat-core-ci.yml
```diff
@@ -24,64 +38,280 @@ concurrency:
...
+      - name: Install package
         run: yarn --frozen-lockfile
+
+      - name: Cache network requests
+        uses: actions/cache@v2
+        with:
+          path: |
+            packages/hardhat-core/test/internal/hardhat-network/provider/.hardhat_node_test_cache
+          key: hardhat-network-forking-tests-${{ hashFiles('yarn.lock') }}-${{ hashFiles('packages/hardhat-core/test/internal/hardhat-network/provider/node.ts') }}
+
```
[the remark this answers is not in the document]
Wodann: Created an issue for it: https://github.com/NomicFoundation/rethnet/issues/87

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -217,13 +209,10 @@ impl RpcClient {
...
-        let responses: Vec<serde_json::Value> =
-            serde_json::from_str(&response.text).map_err(|err| {
-                RpcClientError::DeserializationError {
-                    msg: err.to_string(),
-                    body: response.text.clone(),
-                }
-            })?;
+        let responses: Vec<serde_json::Value> = serde_json::from_str(&response.text)
+            .unwrap_or_else(|error| {
+                panic!("Batch response `{response:?}` failed to parse due to error: {error}")
```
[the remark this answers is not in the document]
Wodann: The [JSON-RPC spec](https://www.jsonrpc.org/specification#error_object) states:

> When a rpc call is made, the Server MUST reply with a Response, except for in the case of Notifications.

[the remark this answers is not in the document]
Wodann: It will be interesting to see whether we can break it though. We can improve our test suite by adding integration tests that mock the server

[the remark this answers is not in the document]
Wodann: Just to clarify, the thing I'd like to make a distinction between here are:
1) parsing errors of the JSON schema
2) server failures

crates/rethnet_eth/src/remote/client.rs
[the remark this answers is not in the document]
feuGeneA: well sure, the contract itself is immutable, so the `code_hash` check here could never fail, but...

the balance of a contract account definitely can change over time: "You can send Ether to other contracts" ([source](https://solidity-by-example.org/sending-ether/)). anyone could send ether to this contract, from either an Externally Owned Account or from another contract account, and that would change the balance in this test. (why anyone would do that, though, is another matter entirely haha.)

Also, about the nonce, i just looked it up, and it _can_ change: if contract A's code is induced to deploy another contract, that increases the nonce of contract account A. again, i'm not sure whether the DAI contract code is capable of doing this, but...

just an FYI that in theory this test is inherently unstable

crates/rethnet_eth/src/remote/client.rs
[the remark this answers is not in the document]
feuGeneA: just to be clear, i've personally experienced HTTP 500's coming back from Alchemy from time to time. maybe that's handled by the underlying `reqwest` library or whatever, and never bubbles up this far? but it definitely happens

crates/rethnet_eth/src/remote/client.rs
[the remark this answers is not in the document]
Wodann: Thanks!

I adapted it to only check the code for that one.

crates/rethnet_eth/src/remote/client.rs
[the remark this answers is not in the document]
Wodann: I've bubbled up the `reqwest` error for HTTP error statuses and added a test to validate it: https://github.com/NomicFoundation/hardhat/pull/3612/commits/41e4b466beae30a11addd4afcf5170b6e1d4de10

PR #3626, opened 2023-02-01
explain how to persist blocks using localhost

ncimino:
This change adds an explanation of how to persist blocks using `hardhat` with `localhost` configurations.

### DIFF
--- a/docs/src/content/hardhat-network/docs/reference/index.md
+++ b/docs/src/content/hardhat-network/docs/reference/index.md
@@ -21,6 +21,31 @@ You can set the following fields on the `networks.hardhat` config:
 
 The chain ID number used by Hardhat Network's blockchain. Default value: `31337`.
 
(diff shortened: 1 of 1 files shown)


alcuadrado:
Thanks for submitting this. I like the explanation, but I'm confused about where it should be placed.

WDYT @fvictorio ?

fvictorio:
I agree that this shouldn't be in the reference. We should have an explainer that clearly outlines the difference between the in-process Hardhat network, `hh node`, and external networks.

I wrote about that [here](https://hackmd.io/@fvictorio/hardhat-networks-and-providers). That explanation is too verbose and personal, but I think it could work as inspiration.

That being said, I don't want to ignore this PR until we have the time to write a full guide. Maybe we can just start with a separate page with this content and expand it in the future? I feel it's better than nothing.

@ncimino do you want to do that?

fvictorio:
I think I'm going to close this for the time being. I'm pretty sure this shouldn't be in the reference, but sadly I can't figure out a better place to move it. Happy to re-open if someone has a good idea.

PR #3637, opened 2023-02-03
Overriding subtask parameters

zemse:

---

This PR adds an optional third argument to the `hre.run` function which enables to override subtask arguments.

### DIFF
--- a/packages/hardhat-core/src/internal/core/runtime-environment.ts
+++ b/packages/hardhat-core/src/internal/core/runtime-environment.ts
@@ -276,29 +276,32 @@ export class Environment implements HardhatRuntimeEnvironment {
     const previousRunSuper: any = globalAsAny.runSuper;
     globalAsAny.runSuper = runSuper;
 
-    let modifiedHreWithParentTaskProfile: any | undefined;
-    if (
-      this.hardhatArguments.flamegraph === true ||
-      subtaskArguments !== undefined
-    ) {
-      // We create a modified version of `this`, as we want to keep track of the
-      // `taskProfile` and use it as `callerTaskProfile` if the action calls
-      // `run`, and add a few utility methods.
-      //
-      // Note that for this to work we need to set the prototype later
+    // We create a modified version of `this`, as we want to keep track of the
+    // `subtaskArguments` and use it to generate updated `subtaskArguments` if
+    // the action calls `run`.
+    //
+    // Note that for this to work we need to set the prototype later
+    let modifiedHreWithParentTaskProfile: any = {
+      ...this,
+      run: (
+        _name: string,
+        _taskArguments: TaskArguments,
+        _subtaskArguments: SubtaskArguments
+      ) =>
+        (this as any).run(
+          _name,
+          _taskArguments,
+          { ..._subtaskArguments, ...subtaskArguments }, // parent subtask args take precedence
+          taskProfile
+        ),
+    };
+
+    if (this.hardhatArguments.flamegraph === true) {
+      // To generate flamegraph, we want to keep track of the `taskProfile` and
+      // use it as `callerTaskProfile` if the action calls `run`, and add a few
+      // utility methods.
       modifiedHreWithParentTaskProfile = {
-        ...this,
-        run: (
-          _name: string,
-          _taskArguments: TaskArguments,
-          _subtaskArguments: SubtaskArguments
-        ) =>
-          (this as any).run(
-            _name,
-            _taskArguments,
-            { ..._subtaskArguments, ...subtaskArguments }, // parent subtask args take precedence
-            taskProfile
-          ),
+        ...modifiedHreWithParentTaskProfile,
         adhocProfile: async (_name: string, f: () => Promise<any>) => {
           const adhocProfile = createTaskProfile(_name);
           taskProfile!.children.push(adhocProfile);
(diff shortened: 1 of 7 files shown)


packages/hardhat-core/src/types/runtime.ts
```diff
@@ -183,7 +187,8 @@ export interface TasksMap {
 
 export type RunTaskFunction = (
   name: string,
-  taskArguments?: TaskArguments
+  taskArguments?: TaskArguments,
+  subtaskArguments?: { [subtaskName: string]: TaskArguments }
```
fvictorio, line 191: Shouldn't this be

```suggestion
  subtaskArguments?: SubtaskArguments
```

?

packages/hardhat-core/src/internal/core/runtime-environment.ts
```diff
@@ -334,9 +354,14 @@ export class Environment implements HardhatRuntimeEnvironment {
    */
   private _resolveValidTaskArguments(
     taskDefinition: TaskDefinition,
-    taskArguments: TaskArguments
+    taskArguments: TaskArguments,
+    subtaskArguments: SubtaskArguments = {}
```
fvictorio, line 358: This is a private function so the default value is not necessary.

packages/hardhat-core/src/types/runtime.ts
```diff
@@ -146,6 +146,10 @@ export interface TaskDefinition extends ConfigurableTaskDefinition {
  */
 export type TaskArguments = any;
 
+export interface SubtaskArguments {
+  [subtaskName: string]: TaskArguments;
+}
+
 export interface RunSuperFunction<ArgT extends TaskArguments> {
   (taskArguments?: ArgT): Promise<any>;
```
fvictorio, line 154: Should this signature be updated?

packages/hardhat-core/src/internal/core/runtime-environment.ts
```diff
@@ -267,16 +275,28 @@ export class Environment implements HardhatRuntimeEnvironment {
     globalAsAny.runSuper = runSuper;
 
     let modifiedHreWithParentTaskProfile: any | undefined;
-    if (this.hardhatArguments.flamegraph === true) {
+    if (
+      this.hardhatArguments.flamegraph === true ||
+      subtaskArguments !== undefined
+    ) {
```
fvictorio, line 281: Hmm, I don't think this is right. This seems to mean that we will generate a flamegraph when subtask arguments are defined, which we don't want to do.

packages/hardhat-core/src/internal/core/runtime-environment.ts
```diff
@@ -218,20 +223,23 @@ export class Environment implements HardhatRuntimeEnvironment {
...
     if (taskDefinition instanceof OverriddenTaskDefinition) {
       runSuperFunction = async (
-        _taskArguments: TaskArguments = taskArguments
+        _taskArguments: TaskArguments = taskArguments,
+        _subtaskArguments: SubtaskArguments = subtaskArguments
       ) => {
         log("Running %s's super", taskDefinition.name);
 
         if (taskProfile === undefined) {
           return this._runTaskDefinition(
```
fvictorio, line 239: There is another `this._runTaskDefinition` call after this that is not receiving the `_subtaskArguments`. I can't add a comment on that line, so I'm mentioning it here.

packages/hardhat-core/src/internal/core/runtime-environment.ts
alcuadrado: I guess we can always make the `subtaskArguments !== undefined`, by defaulting to `{}`

packages/hardhat-core/src/types/runtime.ts
alcuadrado: yep

packages/hardhat-core/src/types/runtime.ts
alcuadrado: Good catch. It looks better.

packages/hardhat-core/src/internal/core/runtime-environment.ts
alcuadrado: Agreed. It's not undefined when it's called.

packages/hardhat-core/src/internal/core/runtime-environment.ts
```diff
@@ -260,23 +269,39 @@ export class Environment implements HardhatRuntimeEnvironment {
...
+
     if (this.hardhatArguments.flamegraph === true) {
-      // We create a modified version of `this`, as we want to keep track of the
-      // `taskProfile` and use it as `callerTaskProfile` if the action calls
-      // `run`, and add a few utility methods.
-      //
-      // Note that for this to work we need to set the prototype later
+      // To generate flamegraph, we want to keep track of the `taskProfile` and
+      // use it as `callerTaskProfile` if the action calls `run`, and add a few
+      // utility methods.
```
alcuadrado, line 302:
```suggestion
      // We modify the `this` again to add  a few utility methods.
```

packages/hardhat-core/src/internal/core/runtime-environment.ts
```diff
@@ -260,23 +269,39 @@ export class Environment implements HardhatRuntimeEnvironment {
...
     const previousRunSuper: any = globalAsAny.runSuper;
     globalAsAny.runSuper = runSuper;
 
-    let modifiedHreWithParentTaskProfile: any | undefined;
+    // We create a modified version of `this`, as we want to keep track of the
+    // `subtaskArguments` and use it to generate updated `subtaskArguments` if
+    // the action calls `run`.
+    //
+    // Note that for this to work we need to set the prototype later
+    let modifiedHreWithParentTaskProfile: any = {
```
alcuadrado, line 284:
```suggestion
    let modifiedHre: any = {
```

This name is outdated.

alcuadrado (approved):

Issue #3653, opened 2023-02-08
Support Solidity 0.8.18


taxio:
### Describe the feature

Solidity v0.8.18 has been released.
https://github.com/ethereum/solidity/releases/tag/v0.8.18
https://blog.soliditylang.org/2023/02/01/solidity-0.8.18-release-announcement/

Is there a schedule to support this?

fvictorio:
Fixed by #3654.

PR #3654, opened 2023-02-08
Support Solidity 0.8.18

taxio:

---

for https://github.com/NomicFoundation/hardhat/issues/3653

Bump up solidity version to [v0.8.18](https://github.com/ethereum/solidity/releases/tag/v0.8.18)


packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
```diff
@@ -809,6 +809,11 @@ const solidityCompilers: SolidityCompiler[] = [
     compilerPath: "soljson-v0.8.17+commit.8df45f5f.js",
     latestSolcVersion: true,
   },
+  {
+    solidityVersion: "0.8.18",
+    compilerPath: "soljson-v0.8.18+commit.87f61d96.js",
+    latestSolcVersion: true,
```
alcuadrado, line 815: This is the only thing that caught my attention.

alcuadrado (commented):
Thanks for this PR! I only left a minor comment. Maybe @fvictorio can chime in.

packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
fvictorio: Not sure what you mean. The only thing that comes to mind about this is that maybe we should remove the `latestSolcVersion` flag from a previous version (e.g., 0.8.14).

Right now we don't have a policy about this. Maybe something like "the latest 3 versions of the current minor, and the last version of the previous minor" is enough.

fvictorio: To clarify: this flag is used to filter the tested versions by default, but we also have a workflow that runs the stack traces tests with all the configured versions. So this is a sort of `fast` vs `full` tests config.

fvictorio: I removed the `latestSolcVersion` flag from 0.8.14 and 0.8.15.

fvictorio (approved):

PR #3655, opened 2023-02-08
feat: support importing a sol from a project in the monorepo

zouguangxian:

---

### DIFF
--- a/packages/hardhat-core/src/internal/solidity/resolver.ts
+++ b/packages/hardhat-core/src/internal/solidity/resolver.ts
@@ -299,7 +299,6 @@ export class Resolver {
       const packageRoot = path.dirname(packageJsonPath);
       const pattern = new RegExp(`^${libraryName}/?`);
       const fileName = sourceName.replace(pattern, "");
-      absolutePath = path.join(packageRoot, fileName);
 
       await this._validateSourceNameExistenceAndCasing(
         packageRoot,
@@ -308,6 +307,7 @@ export class Resolver {
         fileName,
         true
       );
+      absolutePath = path.join(packageRoot, fileName);
     } else {
       await this._validateSourceNameExistenceAndCasing(
         nodeModulesPath,
--- a/packages/hardhat-core/src/internal/solidity/resolver.ts
+++ b/packages/hardhat-core/src/internal/solidity/resolver.ts
@@ -287,15 +287,35 @@ export class Resolver {
       }
     }
 
-    let packageRoot = path.dirname(packageJsonPath);
-    let pattern = new RegExp(`^${libraryName}/?`);
-    let fileName = sourceName.replace(pattern, "");
+    let nodeModulesPath = path.dirname(path.dirname(packageJsonPath));
+    if (this._isScopedPackage(sourceName)) {
+      nodeModulesPath = path.dirname(nodeModulesPath);
+    }
 
-    await this._validateSourceNameExistenceAndCasing(
-      packageRoot,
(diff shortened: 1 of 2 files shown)


alcuadrado:
Thanks for submitting this PR. Can you further explain its changes? I'm not sure if I get them, as monorepos already work.

zouguangxian:
If there are two project packages/A(@examples/A) and packages/B(@examples/B) in a monorepo.
````
monorepo
   - packages
      -  A
      -  B  
````

`import "@examples/A/contracts/access.sol"` in contracts/resource.sol of project packages/B.

The dependencies in packages.json of project packages/B would like:
```
  "devDependencies": {
    "@examples/A": "workspace:^",
  }
```

When executing `yarn hardhat compile` in project packages/B, it will report
````
Error HH404: File @examples/A/contracts/access.sol, imported from contracts/resource.sol, not found.

Caused by: FileNotFoundError: File /Users/gx/monorepo/@examples/A/contracts/access.sol not found
````

By outputting the variables in packages/hardhat-core/src/internal/solidity/resolver.ts, it shows that
````
packageJsonPath = /Users/gx/monorepo/packages/A/package.json 
sourceName = @examples/A/contracts/access.sol 
libraryName = @examples/A
nodeModulesPath = /Users/gx/monorepo/
````
The reason is it uses nodeModulesPath + sourceName to locate the file, but there is an assumption that the package.json will be placed under node_modules/@scope/package.

Actually, if you know the packageJsonPath, the source code should be in the same directory with that file; you only need to access the file whose name is the sourceName without the libraryName prefix.

**Comment by alcuadrado** [comment]
Do you mind creating a reproduction repository? I've tested the setup you describe and it works for me, so we are probably doing something slightly differently.

**Comment by zouguangxian** [comment]
I created a reproduction repository at :
https://github.com/zouguangxian/yarn-monorepo-hardhat-example

After you clone the repository, execute these instructions to reproduce the issue:
```
yarn install
cd packages/B
yarn build --show-stack-traces
```

**Comment by zouguangxian** [comment]
The issue exists when using **Yarn PNP** + monorepo + hardhat.

**Comment by alcuadrado** [comment]
Thanks! I see your point now. It's an issue with PnP. I was running Yarn 1.x

File: packages/hardhat-core/src/internal/solidity/resolver.ts [hunk 1]
```diff
@@ -287,14 +287,13 @@ export class Resolver {
       }
     }
 
-    let nodeModulesPath = path.dirname(path.dirname(packageJsonPath));
-    if (this._isScopedPackage(sourceName)) {
-      nodeModulesPath = path.dirname(nodeModulesPath);
-    }
+    let packageRoot = path.default.dirname(packageJsonPath);
```
alcuadrado, line 290: Is this a Yarn 2+ specific API? I can't find any docs about `path.default`.

zouguangxian: I corrected it to `path.dirname`.

fvictorio:
Hey @zouguangxian, I adapted your solution so that it's only used in the special case where the imported file is not inside a `node_modules` file. I think this makes this change much less risky, while still solving your issue.

The reason I didn't want to merge the previous version is that the `_validateSourceNameExistenceAndCasing` function expects a source name, but what you were passing (in all scenarios) was no longer technically a source name. Perhaps this wasn't a big deal, but it might've led to confusing error messages in some scenarios.

@alcuadrado could you review my version of the fix, please?

PR #3658, opened 2023-02-09
add eslint rule for hardhat plugin errors

zoeyTM:
Added an ESLint rule for HardhatPluginErrors too, mostly for Ignition.

### DIFF
--- a/config/eslint/eslintrc.js
+++ b/config/eslint/eslintrc.js
@@ -13,6 +13,7 @@ module.exports = {
     "@typescript-eslint",
   ],
   rules: {
+    "@nomiclabs/only-hardhat-plugin-error": "off",
     "@typescript-eslint/adjacent-overload-signatures": "error",
     "@typescript-eslint/array-type": [
       "error",
--- a/config/eslint/eslintrc.js
+++ b/config/eslint/eslintrc.js
@@ -13,7 +13,6 @@ module.exports = {
     "@typescript-eslint",
   ],
   rules: {
-    "@nomiclabs/only-hardhat-plugin-error": "off",
     "@typescript-eslint/adjacent-overload-signatures": "error",
     "@typescript-eslint/array-type": [
       "error",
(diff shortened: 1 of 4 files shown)


packages/eslint-plugin/onlyHardhatErrorRule.js
```diff
@@ -20,6 +20,26 @@ function onlyHardhatErrorRule(context) {
...
+  return {
+    ThrowStatement(node) {
+      const expression = parserServices.esTreeNodeToTSNodeMap.get(node.argument);
+
+      if (!isHardhatPluginError(expression, checker)) {
+        const exceptionName = getExpressionClassName(expression, checker);
+
+        context.report({
+          node,
+          message: `Only HardhatError must be thrown, ${exceptionName} found.`,
```
alcuadrado, line 36:
```suggestion
          message: `Only HardhatPluginError must be thrown, ${exceptionName} found.`,
```

config/eslint/eslintrc.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
     "@typescript-eslint",
   ],
   rules: {
+    "@nomiclabs/only-hardhat-plugin-error": "off",
```
alcuadrado, line 16: I'm a bit confused about this line. I'd say it's not necessary. Maybe you added it while debugging something?

zoeyTM: My assumption was that all rules were "on" by default when using a plugin that exports multiple rules, so this would trigger in the hardhat repo unless I turned it off. That may be an incorrect assumption, I'm not sure.

alcuadrado: I don't think so, as we explicitly enable it here: https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/.eslintrc.js#L8

zoeyTM: OH I missed that entirely. Okay I'll remove that line.

PR #3659, opened 2023-02-09
Set "resolveJsonModule": true in sample tsconfig

mlshv:
I think `"resolveJsonModule": true` should be added to sample project tsconfig because it's a pretty common case to import ABI from JSON files in tests. By far, i've been adding this setting manually to all the projects bootstrapped with hardhat. I believe that it would be better include it by default.

### DIFF
--- a/.changeset/dull-files-dream.md
+++ b/.changeset/dull-files-dream.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Set "resolveJsonModule": true in sample tsconfig
+The `resolveJsonModule` compiler option is now enabled by default in the sample tsconfig (thanks @mlshv!)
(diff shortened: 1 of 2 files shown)


alcuadrado:
This is a good idea.

Does this work well with ESM projects? @fvictorio may know.

mlshv:
It looks like if you set `"module"` to `"ESNext"`, for example, you should also add `"moduleResolution": "node"` and add type assertion to imports to make it work on Node 17.5 and higher:

```js
import ERC20Json from "../artifacts/contracts/ERC20.sol/ERC20.json" assert { type: "json" };
```

alcuadrado:
I just merged our ESM support, which doesn't include TS as of now, so this should be good to go.

mlshv:
@alcuadrado, is any action required from my side?

fvictorio (approved):

fvictorio:
Thanks!

PR #3665, opened 2023-02-10
[WIP] Upgrade undici and make sure to always set its timeout fields

alcuadrado:
This PR also bumps the timeout used by hardhat-vyper to download the compiler. This tiemout wasn't used, but it happened to match unidici's default, and that wasn't enough.

### DIFF
--- a/.changeset/fresh-rings-dress.md
+++ b/.changeset/fresh-rings-dress.md
@@ -1,0 +1,5 @@
+---
+"@nomiclabs/hardhat-vyper": patch
+---
+
+Increase the timeout used to download the compiler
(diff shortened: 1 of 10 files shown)


alcuadrado:
UPDATE

I noticed that we weren't handling a timeout, which can happen when establishing a connection. It has to be handled in the agent, not per-request, so this fix is outdated.

I implemented the proper fix in the HTTP provider, but I think this really calls for a `@nomicfoundation/hardhat-utils` package.

hawkofsky:
is this pr not merged yet?

fvictorio:
Not yet, I'll try to prioritize it.

insider89:
Why is this one closed? I still have time out issue

PR #3668, opened 2023-02-12
Use our mirror to download the Vyper compilers

alcuadrado:
Previously we were downloading directly from Github, which is unstable and rate-limited. 

I created [a mirror of the releases page](https://github.com/NomicFoundation/vyper-releases-mirror) using github actions and vercel, and this PR uses it.

### DIFF
--- a/packages/hardhat-vyper/src/downloader.ts
+++ b/packages/hardhat-vyper/src/downloader.ts
@@ -9,11 +9,11 @@ import {
   CompilerPlatform,
   CompilerRelease,
 } from "./types";
-import { GITHUB_RELEASES_URL } from "./constants";
 import { VyperPluginError, getLogger } from "./util";
 
 const log = getLogger("downloader");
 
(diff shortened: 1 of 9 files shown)


packages/hardhat-vyper/src/downloader.ts
```diff
@@ -73,7 +61,7 @@ export class CompilerDownloader {
   }
 
   public async initCompilersList(
-    { forceDownload } = { forceDownload: true }
+    { forceDownload } = { forceDownload: false }
```
alcuadrado, line 64: This was redownloading the list every time, which is not necessary as it's already redownlaoded when a release is not found.

alcuadrado:
@zoeyTM can you take a quick look at this? You were the latest person that modified the downloader

zoeyTM (approved):
LGTM!

fvictorio (approved):

Issue #3691, opened 2023-02-17
Bug: If compiled with unintalled npm depencency


N-dcool:
### Version of Hardhat

2.12.7

### What happened?

 when I compile a smart contract with an uninstalled npm dependency it errors out  

 ---> Error HH411: The library `library name with last character missing`, imported from `contract path`, is not installed. Try installing it using npm 

 in terminal.

<img width="1432" alt="bugReport" src="https://user-images.githubusercontent.com/107427909/219768770-61d5c59d-eb4b-42fc-a6e9-d7ffe83ae257.png">

## another example : 
<img width="964" alt="bug2" src="https://user-images.githubusercontent.com/107427909/219770112-3dd75af2-f28f-4ecd-adba-d1ee1a306410.png">

### Minimal reproduction steps

try with any uninstalled npm dependencies having `@` at the prefix
compile smart contract: `npx hardhat compile`

N-dcool:
I open pull request I think I got the solution #3692

fvictorio:
Thank you @N-dcool. Just to double-check, this is about the error message having that issue, right?

fvictorio:
Ok, I looked into this a bit more. Some comments about this.

First, I think this only happens if you are making a mistake in your imports. Hardhat uses node to resolve dependencies, and in node a package can be either something like `foo` or something like `@foo/bar`, but you can't have a package named `@foo`. That means that you can never import a contract doing `@foo/contract.sol`. In your example, you are doing `@hardhat/console.sol` instead of `hardhat/console.sol`.

That being said, we should show a better error message when a user makes that mistake. Plus, that assumption might break in a project that uses our `hardhat-foundry` plugin.

I think the problem is in this line:

https://github.com/NomicFoundation/hardhat/blob/68ea1ad47bdfdd0440390974b09ec544d7b58474/packages/hardhat-core/src/internal/solidity/resolver.ts#L405

This is basically saying "if the import starts with `@`, then it will have two slashes", which as I explained might not be the case.

I'm not sure what's the right fix here. Maybe just check if the resulting index is `-1` before slicing, and use `sourceName.length` instead in that case. Feel free to send a PR fixing this, but this should probably include a new test.

N-dcool:
> Thank you @N-dcool. Just to double-check, this is about the error message having that issue, right?

Yes I thought it should fix issue.

N-dcool:
> Ok, I looked into this a bit more. Some comments about this.
> 
[...]

PR #3692, opened 2023-02-17
Fixing Bug: If compiled with unintalled npm depencency 

N-dcool:

---

I tried figuring out where is there error after 4-5 hrs i got this line suspisious


fvictorio:
As explained [here](https://github.com/NomicFoundation/hardhat/issues/3691#issuecomment-1438295601), this doesn't fix the issue. Plus, `.substr` and `.substring` have the same behavior when only one parameter is used, so this change shouldn't have any impact.

PR #3709, opened 2023-02-22
fix: ethers.Signer check in getContractFactory

jtakalai:
- [ X ] Because this PR includes a **bug fix**, relevant tests have been included.

---

The current instanceof check will fail for ethers.Wallets that are imported from another ethers outside hardhat.

Consider this code that deploys a test contract from a clean wallet (to achieve deterministic addresses):
```typescript
import { ethers } from "hardhat"
import { Wallet, utils } from "ethers"

async function deployMyContract(deployer: Wallet) {
    return (await ethers.getContractFactory("MyContract", deployer)).deploy()
}

async function deployContracts() {
    const [admin] = await ethers.getSigners() as unknown as Wallet[]
    const cleanWallet = new Wallet(utils.id("testing"), admin.provider)
    await (await admin.sendTransaction({ to: cleanWallet.address, value: parseEther("1") })).wait()
    await deployMyContract(cleanWallet)
}
```

The problem is, if deployer is created like this, then the currently modified check fails to recognize that deployer is in fact an ethers.Signer. This means that (very confusingly!) the deployer address will be something else (probably the first in the `getSigners()` array).

### DIFF
--- a/.changeset/fix-getContractFactory-signer.md
+++ b/.changeset/fix-getContractFactory-signer.md
@@ -2,4 +2,4 @@
 "@nomiclabs/hardhat-ethers": patch
 ---
 
-fix: ethers.Signer check in getContractFactory
+Make getContractFactory's params validation more flexible.
--- a/.changeset/fix-getContractFactory-signer.md
+++ b/.changeset/fix-getContractFactory-signer.md
@@ -1,0 +1,5 @@
+---
+"@nomiclabs/hardhat-ethers": patch
+---
+
+fix: ethers.Signer check in getContractFactory
--- a/packages/hardhat-ethers/src/internal/helpers.ts
+++ b/packages/hardhat-ethers/src/internal/helpers.ts
@@ -116,8 +116,7 @@ export async function getContractFactory(
 function isFactoryOptions(
   signerOrOptions?: ethers.Signer | FactoryOptions
 ): signerOrOptions is FactoryOptions {
-  const { Signer } = require("ethers") as typeof ethers;
-  if (signerOrOptions === undefined || signerOrOptions instanceof Signer) {
+  if (signerOrOptions === undefined || ["Wallet", "SignerWithAddress", "Signer"].includes(signerOrOptions.constructor.name)) {
     return false;
   }
 
--- a/packages/hardhat-ethers/src/internal/helpers.ts
+++ b/packages/hardhat-ethers/src/internal/helpers.ts
@@ -117,7 +117,7 @@ function isFactoryOptions(
   signerOrOptions?: ethers.Signer | FactoryOptions
 ): signerOrOptions is FactoryOptions {
   const { Signer } = require("ethers") as typeof ethers;
-  return !Signer.isSigner(signerOrOptions);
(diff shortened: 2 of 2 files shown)


alcuadrado:
Thanks for sending this PR!

> The current instanceof check will fail for ethers.Wallets that are imported from another ethers outside hardhat.

This is interesting! Can you tell us more about when this happens? I understand how the check fails, but curious about the context.

jtakalai:
> Thanks for sending this PR!
> 
> > The current instanceof check will fail for ethers.Wallets that are imported from another ethers outside hardhat.
> 
> This is interesting! Can you tell us more about when this happens? I understand how the check fails, but curious about the context.

Well the context in short is I use `ethers` from my own node_modules to get the Wallet; I imagine if I were to use that particular ethers can't encapsulated inside hardhat, it would recognize it with instanceof. I think any version of ethers should work there though, coming from anywhere, and there really isn't much else to tell except "well, it's calling itself a Wallet/Signer" (or SignerWithAddress... what's the story behind that one? ethers.Signer does expose `address` already...).

Workaround clearly is replacing `deployer` with `{ signer }` (which is what I'm doing now) but it's still bad to leave people looking for bugs elsewhere when in fact they're only confused about who signed the deployment tx!

jtakalai:
@alcuadrado is that changeset thing necessary? I seemed like it's blocking the PR, so I added it... you can remove it if it's unnecessary noise.

alcuadrado:
> Well the context in short is I use `ethers` from my own node_modules to get the Wallet;

Yes, but the plugin has a peerDependency on ethers, it doesn't have its own version of ethers. That's why I'm asking. Are you using two incompatible versions? do you mind sharing your project or package jsons and lockfiles?

> or SignerWithAddress... what's the story behind that one? ethers.Signer does expose address already...

It has a synchronous `.address`, while `Signer` has an async `getAddress`. `Wallet` does have `.address`.

> @alcuadrado is that changeset thing necessary? I seemed like it's blocking the PR, so I added it... you can remove it if it's unnecessary noise.

Yes, it's needed. Thank you!

BTW, I'm not opposed to this change, but just trying to understand if there's an alternative to achieve this, or if there's some other problem waiting to happen.

jtakalai:
@alcuadrado sure! Here's the code where I had to do the `{ signer }` workaround: https://github.com/streamr-dev/network-contracts/pull/468/files#diff-7507020c4f2dc3b7a6d4f052c2b758c0def70cb4098eedcd27dbc65340dc4508L49

alcuadrado:
Thanks @jtakalai!

Can you please give me permissions to commit to your repo? I think I have a more robust solution.

jtakalai:
> Can you please give me permissions to commit to your repo? I think I have a more robust solution.

Sorry for the delay! I didn't consider there was more to do  :D   I sent you invite, although not sure what the value of working in my repo is vs. working in a branch in this repo...

alcuadrado:
> I sent you invite, although not sure what the value of working in my repo is vs. working in a branch in this repo...

Oh, now I realized that it's `main` in your repo. That's why it was different. Normally you just check a setting in the PR, and no invite is needed.

PR #3719, opened 2023-02-26
Scoped hardhat tasks

zemse:
### Changes

- Adds a `hre.scopedTasks` bucket of TasksMap under a scope. Legacy `hre.tasks` still exists and is referred to when scope name is undefined.

### Help Printer changes

<details>
<summary>npx hardhat help</summary>

```
% npx hardhat help
Hardhat version 2.14.0

Usage: hardhat [GLOBAL OPTIONS] <TASK> [TASK OPTIONS]

GLOBAL OPTIONS:

  --config              A Hardhat config file. 
  --emoji               Use emoji in messages. 
  --flamegraph          Generate a flamegraph of your Hardhat tasks 
  --help                Shows this message, or a task's help if its name is provided 
  --max-memory          The maximum amount of memory that Hardhat can use. 
  --network             The network to connect to. 
  --show-stack-traces   Show stack traces (always enabled on CI servers). 
  --tsconfig            A TypeScript config file. 
  --typecheck           Enable TypeScript type-checking of your scripts/tests 
  --verbose             Enables Hardhat verbose logging 
  --version             Shows hardhat's version. 


AVAILABLE TASKS:

  check                 Check whatever you need
  clean                 Clears the cache and deletes all artifacts
  compile               Compiles the entire project, building all artifacts
  console               Opens a hardhat console
  coverage              Generates a code coverage report for tests
  flatten               Flattens and prints contracts and their dependencies. If no file is passed, all the contracts in the project will be flattened.
  gas-reporter:merge
  help                  Prints this message
  node                  Starts a JSON-RPC server on top of Hardhat Network
  run                   Runs a user-defined script after compiling the project
  test                  Runs mocha tests
  typechain             Generate Typechain typings for compiled contracts
  verify                Verifies contract on Etherscan


AVAILABLE SCOPES:

  scope1                des1
  scope2                des2

To get help for a specific task run: npx hardhat help [scope/task] [task]
```
</details>

<details>
<summary>npx hardhat scope1 help</summary>

```
% npx hardhat scope1 help


AVAILABLE TASKS UNDER SCOPE scope1:

  task1	
  task2	

```
</details>


packages/hardhat-core/src/builtin-tasks/help.ts
```diff
@@ -1,30 +1,75 @@
...
     "task",
     "An optional task to print more info about"
   )
   .setAction(
-    async ({ task: taskName }: { task?: string }, { tasks, version }) => {
+    async (
+      {
+        scopeOrTask: first,
+        task: second,
+      }: { scopeOrTask?: string; task?: string },
```
schaable, line 24: The names `first` and `second` don't tell me too much about the variables and make the code a bit harder to read. I'd stick with `scopeOrTask` and `task`.

packages/hardhat-core/src/internal/core/tasks/util.ts
```diff
@@ -0,0 +1,16 @@
+import { TaskIdentifier } from "../../../types";
+
+export function parseTaskIdentifier(taskIdentifier: TaskIdentifier): {
+  scope: string | undefined;
+  task: string;
+} {
+  let scope: string | undefined;
```
schaable, line 7: This may be easier to read if we do:
```
  if (typeof taskIdentifier === "string") {
    return { scope: undefined, task: taskIdentifier };
  } else {
    const { scope, task } = taskIdentifier;
    return { scope, task };
  }
```

packages/hardhat-core/src/internal/core/tasks/task-definitions.ts
```diff
@@ -36,30 +44,52 @@ export class SimpleTaskDefinition implements TaskDefinition {
...
    *
-   * @param name The task's name.
+   * @param taskIdentifier The task's identifier.
    * @param isSubtask `true` if the task is a subtask, `false` otherwise.
+   * @param _setScopeCallback fn to call while setting the scope.
    */
   constructor(
-    public readonly name: string,
+    taskIdentifier: TaskIdentifier,
+    private readonly _setScopeCallback: (
```
schaable, line 62: shouldn't we default this parameter to an empty function? that way we can avoid having to pass it every time we create a new instance

packages/hardhat-core/src/internal/cli/ArgumentsParser.ts
```diff
@@ -102,6 +106,71 @@ export class ArgumentsParser {
...
+  public parseScopeAndTaskNames(
+    allUnparsedCLAs: string[],
+    taskDefinitions: TasksMap,
+    scopeDefinitions: ScopesMap
+  ): {
+    scopeName?: string;
+    taskName: string;
+    unparsedCLAs: string[];
+  } {
+    let scopeName: string | undefined;
```
schaable, line 123: I think this is good, but it could use a little bit of cleanup. For example you could put `const [firstCLA, secondCLA, ...restCLAs] = allUnparsedCLAs;` at the top to avoid repetition. Same with `const scopeDefinition = scopeDefinitions[firstCLA];` and maybe `taskDefinitions[firstCLA]`. Also what do you think of using early return here?

fvictorio (approved):

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -613,8 +613,26 @@ Please double check your task definitions.`,
 What makes these types special is that they can be represented as strings, so you can write them down in the terminal.`,
       shouldBeReported: false,
     },
-    DEPRECATED_TRANSFORM_IMPORT_TASK: {
+    TASK_SCOPE_CLASH: {
```
kanej, line 616: Do you not usually append to the end of the errors list, rather than reassigning error codes?

PR #3723, opened 2023-02-27
Fix #3721

### LINKED ISSUE #3721: Modifications to the HRE don't persist between tasks
Due to the changes in https://github.com/NomicFoundation/hardhat/pull/3637, if a task adds a property to the HRE, this won't be reflected in other tasks. This happen because we are creating a new (modified) HRE between task executions.

This breaks Nexus's setup, and it's very likely that it affects other projects too.

This is a blocker for the release of v2.13.0.

alcuadrado:
This PR fixes #3721 by using a Proxy instead of creating new instances of the HRE every time.

### DIFF
--- a/packages/hardhat-core/src/internal/core/runtime-environment.ts
+++ b/packages/hardhat-core/src/internal/core/runtime-environment.ts
@@ -324,7 +324,7 @@ export class Environment implements HardhatRuntimeEnvironment {
       };
     }
 
-    const uninjectFromGlobal = proxiedHre?.injectToGlobal();
+    const uninjectFromGlobal = proxiedHre.injectToGlobal();
 
     try {
       return await taskDefinition.action(taskArguments, proxiedHre, runSuper);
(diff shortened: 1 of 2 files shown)


packages/hardhat-core/src/internal/core/runtime-environment.ts
```diff
@@ -275,62 +275,59 @@ export class Environment implements HardhatRuntimeEnvironment {
...
+          return f();
+        } finally {
+          completeTaskProfile(adhocProfile);
+        }
+      };
+    }
 
-    const uninjectFromGlobal =
-      modifiedHre?.injectToGlobal() ?? this.injectToGlobal();
+    const uninjectFromGlobal = proxiedHre?.injectToGlobal();
```
fvictorio, line 327:
```suggestion
    const uninjectFromGlobal = proxiedHre.injectToGlobal();
```

fvictorio (approved):

PR #3726, opened 2023-02-28
Fix hardhat-solhint plugin error with solhint 3.4.0

### LINKED ISSUE #3724: @nomiclabs/hardhat-solhint does not work
### Version of Hardhat

2.12.7

### What happened?

```
NomicLabsHardhatPluginError: An error occurred loading the solhint formatter stylish
    at getFormatter (.../node_modules/@nomiclabs/hardhat-solhint/src/index.ts:21:11
```

`Caused by: Error: Cannot find module 'eslint/lib/formatters/stylish'`

calling:

https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-solhint/src/index.ts#L12

- create hardhat project using typescript and default settings
- add hardhat-solhint and enable in config
- run npx hardhat check
- 

### Search terms

solhint hardhat-solhint

yhuard:

---

Fixes #3724.

The error described in the aforementioned issue comes from https://github.com/protofire/solhint/pull/380, which updated ESLint from v5 to v8 in Solhint 3.4.0. In ESLint v8, `eslint/lib/formatters/${formatterName}` is no longer a valid path. However, Solhint is now shipping the formatters, so we can use `solhint/lib/formatters/${formatterName}` instead.

### DIFF
--- a/.changeset/strange-drinks-cry.md
+++ b/.changeset/strange-drinks-cry.md
@@ -1,0 +1,5 @@
+---
+"@nomiclabs/hardhat-solhint": patch
+---
+
+Update solhint to v3.4.0 and fix error with ESLint formatter not found.
--- a/packages/hardhat-solhint/package.json
+++ b/packages/hardhat-solhint/package.json
@@ -34,7 +34,7 @@
     "README.md"
   ],
   "dependencies": {
-    "solhint": "^3.0.0"
+    "solhint": "^3.4.0"
   },
   "devDependencies": {
     "@types/chai": "^4.2.0",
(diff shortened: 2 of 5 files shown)


alcuadrado:
Hey, thanks for sending this PR. Unfortunately, the CI doesn't pass on windows. Do you have any idea why?

yhuard:
Hi @alcuadrado. Since the failing tests are not related to the line I've changed, I guess the issue comes from the latest version of solhint (3.4.0).
I don't have a computer with Windows though :(

yhuard:
I've managed to track down the issue to this change: https://github.com/isaacs/node-glob/blob/main/changelog.md#80

> Glob pattern paths must use forward-slashes as path separators

Solthint 3.4.0 updated a lot of dependencies, including `glob`.

The PR should be ok now :)

alcuadrado:
Thanks for digging it further and fixing it !

PR #3735, opened 2023-03-03
fix: hardhat-foundry plugin user-configured sources path not match pa…

### LINKED ISSUE #3734: hardhat-foundry plugin path bug
### Version of Hardhat

2.13.0

### What happened?

```
🤣  hardhat-foundry-template git:(master) 🌹 npx hardhat test                                                                                                             
Error in plugin hardhat-foundry: User-configured sources path (./src) doesn't match path configured in foundry (src)

For more info run Hardhat with --show-stack-traces
```

`foundry.toml`

```
[default]
src = 'src'
out = 'out'
libs = ['lib']
```

`hardhat.config.ts`

```ts
  solidity: {
    version: "0.8.13",
    settings: {
      optimizer: {
        enabled: true,
        runs: 200,
      },
    },
  },
  paths: {
    sources: "./src", // Use ./src rather than ./contracts as Hardhat expects
    cache: "./cache_hardhat", // Use a different cache for Hardhat than Foundry
  },
};

```

read the source code and find that there is a logic error(the condition should be two paths, not two strings):

https://github.com/NomicFoundation/hardhat/blob/57c56d2353da8828d6b2c167b24bd70b6298a8cc/packages/hardhat-foundry/src/index.ts#L59

### Search terms

hardhat-foundry

bingryan:

---

Fixes #3734.

this bug is in the conditional logic (the condition should be two paths, not two strings)

### DIFF
--- a/.changeset/sharp-experts-play.md
+++ b/.changeset/sharp-experts-play.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-foundry": patch
+---
+
+Fixed a bug in how the configured sources paths are compared (thanks @bingryan!)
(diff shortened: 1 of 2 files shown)


fvictorio:
Thank you!

PR #3741, opened 2023-03-05
feat(@nomiclabs/hardhat-solhint): load .solhintignore file

### LINKED ISSUE #1665: .solhintignore file is not being recognized by @nomiclabs/hardhat-solhint plugin
Add `.solhintignore` file that looks like this:

```
External/ 
```

Run solhint with:

```
npx hardhat check
```

Get the following output:

Screenshot 2021-07-14 at 12 03 59

Expected everything in `External/` to be ignored but it's still being sent to solhint

yhuard:
- [x] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team

---

Fixes #1665. 
I'm not sure whether this is to be considered as a bugfix or a new feature.

### DIFF
--- a/.changeset/clever-hairs-happen.md
+++ b/.changeset/clever-hairs-happen.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
+Improved error handling and messaging for errors from the block explorer
--- a/packages/hardhat-verify/src/internal/errors.ts
+++ b/packages/hardhat-verify/src/internal/errors.ts
@@ -9,6 +9,7 @@ import { TASK_VERIFY_VERIFY } from "./task-names";
 export class HardhatVerifyError extends NomicLabsHardhatPluginError {
   constructor(message: string, parent?: Error) {
     super("@nomicfoundation/hardhat-verify", message, parent);
+    Object.setPrototypeOf(this, this.constructor.prototype);
   }
 }
 
@@ -421,25 +422,23 @@ Encoder error reason: ${reason} fault in ${operation}`,
   }
 }
 
+/**
+ * `VerificationAPIUnexpectedMessageError` is thrown when the block explorer API
+ * does not behave as expected, such as when it returns an unexpected response message.
+ */
 export class VerificationAPIUnexpectedMessageError extends HardhatVerifyError {
   constructor(message: string) {
     super(`The API responded with an unexpected message.
-Please report this issue to the Hardhat team.
 Contract verification may have succeeded and should be checked manually.
 Message: ${message}`);
   }
 }
 
-export class UnexpectedError extends HardhatVerifyError {
-  constructor(e: unknown, functionName: string) {
-    const defaultErrorDetails = `Unexpected error in ${functionName}`;
-    const errorDetails =
-      e instanceof Error
-        ? e.message ?? defaultErrorDetails
-        : defaultErrorDetails;
-    super(`An unexpected error occurred during the verification process.
-Please report this issue to the Hardhat team.
-Error Details: ${errorDetails}`);
+export class NetworkRequestError extends HardhatVerifyError {
+  constructor(e: Error) {
+    super(
+      `A network request failed. This is an error from the block explorer, not Hardhat. Error: ${e.message}`
+    );
   }
 }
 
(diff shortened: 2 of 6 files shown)


packages/hardhat-solhint/src/index.ts
```diff
@@ -84,7 +105,10 @@ function printReport(reports: any) {
 subtask("hardhat-solhint:run-solhint", async (_, { config }) => {
   const { processPath } = require("solhint/lib/index");
   return processPath(
-    join(config.paths.sources, "**", "*.sol").replace(/\\/g, "/"),
+    relative(
```
yhuard, line 108: Because [ignore](https://github.com/kaelzhang/node-ignore#1-pathname-should-be-a-pathrelatived-pathname) requires relative paths to properly work (used by Solhint [here](https://github.com/protofire/solhint/blob/master/lib/index.js#L65))

yhuard:
I'm investigating the issue on Windows (again...)

Oni-giri:
Thanks for taking the time to check it @yhuard !

alcuadrado:
Nice PR! Thanks 🙌

I just left a minor comment

packages/hardhat-solhint/src/index.ts
yhuard: Done. Let me know if it's clear enough :)

fvictorio: I think this would be a bit more readable if it was done in two or three statements, with interspersed comments. Sometihng like this:

```
// we need a relative path because...
const sourcesPath = relative(config.paths.root, config.paths.sources);

// we create a glob that matches all .sol files
const solFilesGlob = join(sourcesPath, "**", "*.sol");

// globs only use forward slashes, even on windows
const normalizedGlob = solFilesGlob.replace(...)
```

Wdyt @yhuard?

yhuard: Thanks for your feedback @fvictorio I've refactored the code as you suggested.

packages/hardhat-solhint/src/index.ts
```diff
@@ -104,26 +104,19 @@ function printReport(reports: any) {
...
-    ).replace(/\\/g, "/"),
-    await getSolhintConfig(config.paths.root)
-  );
+
+  // Create a glob pattern that matches all the .sol files within the sources folder
+  const solFilesGlob = join(config.paths.sources, "**", "*.sol");
+
+  // Make glob pattern relative to the current working directory
+  // See https://github.com/kaelzhang/node-ignore/tree/5.2.4#1-pathname-should-be-a-pathrelatived-pathname
+  const relativeGlob = relative(process.cwd(), solFilesGlob);
```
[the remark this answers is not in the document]
yhuard: To be honest, I'm not 100% sure. First, I was using `config.paths.root` but I've changed it to `process.cwd()` in my last commit because it made more sense to me. [Solhint calls `fs.readFileSync(file)`](https://github.com/protofire/solhint/blob/master/lib/index.js#L55) where `file` is the relative path of the file to lint. I.e. in the context of Solhint, `process.cwd()` is the reference directory from which relative paths start.

In the tests, both `relative(config.paths.root, solFilesGlob)` and `relative(process.cwd(), solFilesGlob)` work. 
An edge case would be if Hardhat's `root` folder is different from the current working directory.
For instance:
```js
  paths: {
    root: "..",
    sources: "solhintignore-project/contracts",
  },
```
In this case, with paths relative to `process.cwd()`, the tests would pass, while with paths relative to `config.paths.root` the tests wouldn't pass.

But this is a really unusual case. In most cases, `config.paths.root` work fine.

If we decide to use `process.cwd()`, then I'll have to make sure that all the ignore globs are relative to the current working directory as well.

If we decide to ignore this edge case (hardhat.root ≠ cwd), then I'll be happy to replace `process.cwd()` with `config.paths.root`. 

I think I'm overthinking it 😅  Let me know what's best in your opinion.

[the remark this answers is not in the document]
yhuard: Hey @fvictorio! I've replaced `process.cwd()` with `config.paths.root` so as not to block this PR. It's highly unlikely that users will complain about the edge case I mentioned yesterday anyway.

PR #3767, opened 2023-03-16
Add support for the shanghai hardfork

fvictorio:
Adds support for the shanghai hardfork. The two main changes are:

- Syncing our fork of ethereumjs with upstream, and making some changes to make `allowUnlimitedContractSize` to work with the EIP-3860 initcode limit.
- Add `withdrawals` and `withdrawalsRoot` fields to RPC blocks if Shanghai is enabled


docs/src/content/hardhat-network/docs/reference/index.md
```diff
@@ -7,9 +7,12 @@
 - petersburg
 - istanbul
 - muirGlacier
+- berlin
 - london
 - arrowGlacier
+- grayGlacier
```
fvictorio, line 13: These two hardforks were missing in the list, so I just added them.

packages/hardhat-core/package.json
```diff
@@ -99,16 +99,16 @@
...
+    "@nomicfoundation/ethereumjs-block": "4.2.1",
+    "@nomicfoundation/ethereumjs-blockchain": "6.2.1",
+    "@nomicfoundation/ethereumjs-common": "3.1.1",
+    "@nomicfoundation/ethereumjs-evm": "1.3.1",
+    "@nomicfoundation/ethereumjs-rlp": "4.0.2",
+    "@nomicfoundation/ethereumjs-statemanager": "1.0.4",
+    "@nomicfoundation/ethereumjs-trie": "5.0.4",
+    "@nomicfoundation/ethereumjs-tx": "4.1.1",
+    "@nomicfoundation/ethereumjs-util": "8.0.5",
+    "@nomicfoundation/ethereumjs-vm": "6.4.1",
```
fvictorio, line 111: I used pinned versions this time (I should've done that the previous time). The fork also has pinned versions in their intra-dependencies.

packages/hardhat-core/src/internal/hardhat-network/provider/BlockchainBase.ts
```diff
@@ -63,7 +63,7 @@ export abstract class BlockchainBase {
 
   public async getBlock(
     blockHashOrNumber: Buffer | bigint | number
-  ): Promise<Block | null> {
+  ): Promise<Block> {
```
fvictorio, line 66: `BlockchainInterface` has this breaking change.

packages/hardhat-core/src/internal/hardhat-network/provider/BlockchainBase.ts
```diff
@@ -73,22 +73,32 @@ export abstract class BlockchainBase {
...
 
     if (typeof blockHashOrNumber === "number") {
-      return this._data.getBlockByNumber(BigInt(blockHashOrNumber)) ?? null;
+      const blockByNumber = this._data.getBlockByNumber(
+        BigInt(blockHashOrNumber)
+      );
+      if (blockByNumber === undefined) {
+        throw new Error("Block not found");
+      }
+      return blockByNumber;
```
fvictorio, line 82: We now throw instead of returning null when the block doesn't exist.

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
```diff
@@ -48,7 +48,8 @@ export function serializeTransaction(
 
 export function deserializeTransaction(
   tx: SerializedTransaction,
-  common: Common
+  common: Common,
+  { allowUnlimitedContractSize }: { allowUnlimitedContractSize: boolean }
```
fvictorio, line 52: A lot of the changes in this PR come from the fact that we need to pass the value of `allowUnlimitedContractSize` to be able to create transactions with the correct `disableMaxInitCodeSizeCheck` value.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
```diff
@@ -254,13 +261,17 @@ export class ForkBlockchain
...
+    // prevent ethereumjs from throwing if unsupported fields are passed.
+    // We use "berlin" for pre-EIP-1559 blocks (blocks without baseFeePerGas),
+    // "merge" for blocks that have baseFeePerGas but not withdrawals,
+    // and "shanghai" for blocks with withdrawals
+    if (rpcBlock.baseFeePerGas === undefined) {
       common.setHardfork("berlin");
+    } else if (rpcBlock.withdrawals === undefined) {
+      common.setHardfork("merge");
+    } else {
+      common.setHardfork("shanghai");
```
fvictorio, line 274: This is an important and risky change. When processing a remote block, we try to set the "right" hardfork that will accept the fields of the remote block. This doesn't mean that the hardfork exactly matches the hardfork enabled in the remote block, but that it will be "compatible".

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -1424,10 +1448,18 @@ Hardhat Network's forking functionality only works with blocks from at least spu
         if (txHash.equals(hash)) {
           const vmDebugTracer = new VMDebugTracer(vm);
           return vmDebugTracer.trace(async () => {
-            await vm.runTx({ tx: txWithCommon, block });
+            await vm.runTx({
+              tx: txWithCommon,
+              block,
+              skipHardForkValidation: true,
```
fvictorio, line 1454: This is a new check that we need to disable in the `runTx` calls. If this is not enabled, `runTx` will fail if the VM and the block have a different hardfork. To be honest, I'm not sure in exactly which scenarios this happens, but several tests passed again with this change :man_shrugging:

packages/hardhat-core/src/internal/hardhat-network/provider/transactions/ReadOnlyValidEIP1559Transaction.ts
```diff
@@ -68,7 +72,7 @@ export class ReadOnlyValidEIP1559Transaction extends FeeMarketEIP1559Transaction
       }
     );
 
-    super(data, { freeze: false, common: fakeCommon });
+    super(data, { ...opts, freeze: false, common: fakeCommon });
```
fvictorio, line 75: This kind of change is needed to pass the `disableMaxInitCodeSizeCheck` flag around.

packages/hardhat-core/test/internal/hardhat-network/helpers/contracts.ts
```diff
@@ -712,3 +712,38 @@ contract CallSelfDestruct {
   },
   topics: {},
 };
+
+export const EXAMPLE_TOUCH_ADDRESS_CONTRACT = {
```
fvictorio, line 716: This new example contract is used to test EIP-3651 (warm COINBASE)

packages/hardhat-core/test/internal/hardhat-network/provider/HardhatBlockchain.ts
```diff
@@ -79,9 +79,12 @@ describe("HardhatBlockchain", () => {
       assert.equal(await blockchain.getBlock(1n), one);
     });
 
-    it("returns undefined non-existent block", async () => {
-      assert.equal(await blockchain.getBlock(0), undefined);
-      assert.equal(await blockchain.getBlock(randomHashBuffer()), undefined);
+    it("throws error for non-existent block", async () => {
+      await assert.isRejected(blockchain.getBlock(0), "Block not found");
```
fvictorio, line 83: Some tests changed because `getBlock` now throws instead of returning null.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/hardforks.ts
```diff
@@ -1532,4 +1537,289 @@ describe("Eth module - hardfork dependant tests", function () {
       });
     });
   });
+
+  describe("shanghai hardfork", function () {
```
fvictorio, line 1541: These are the most interesting tests, please review them carefully.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
```diff
@@ -728,16 +728,21 @@ describe("HardhatNode", () => {
...
+        networkName: "sepolia",
+        url: INFURA_URL.replace("mainnet", "sepolia"),
+        blockToRun: 3095000n, // this block is post-shanghai
+        chainId: 11155111,
+      },
     ];
 
     for (const { url, blockToRun, networkName, chainId } of forkedBlocks) {
       const remoteCommon = new Common({ chain: chainId });
-      const hardfork = remoteCommon.getHardforkByBlockNumber(blockToRun);
```
fvictorio, line 735: This doesn't work for shanghai because `getHardforkByBlockNumber` needs the block timestamp to correctly detect that a block is in that hardfork. Now the hardfork is detected inside `runFullBlock` because we have fetched the block there and can pass the timestamp.

packages/hardhat-core/src/internal/hardhat-network/provider/fork/ForkBlockchain.ts
alcuadrado: I think it should be fine, as we were already setting incorrect HFs.

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
alcuadrado: I think this is a validation that is meant to be run only when receiving a new transaction, isn't it?

If that's the case we can run it in the RPC layer / Node, and use `true` everywhere else. WDYT?

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
alcuadrado: I'll double check this

alcuadrado: Oh, this can trace remote transactions, so see https://github.com/NomicFoundation/hardhat/pull/3767/files#r1143994004

packages/hardhat-core/src/internal/hardhat-network/provider/transactions/ReadOnlyValidEIP1559Transaction.ts
alcuadrado: See https://github.com/NomicFoundation/hardhat/pull/3767/files#r1143996647

packages/hardhat-core/src/internal/hardhat-network/provider/TxPool.ts
fvictorio: You mean always using `disableMaxInitCodeSizeCheck: true`, and making the initcode check ourselves in an upper layer? Makes sense to me, I'll do that.

packages/hardhat-core/test/internal/hardhat-network/provider/node.ts
alcuadrado: Not sure if I follow this, but the rest LGTM

alcuadrado (approved):

PR #3808, opened 2023-04-02
Update hardhat-and-foundry.md

Fabien-Morrow:
This tutorial doesn't work if a git repository isn't initialized before running `npx hardhat init-foundry`
- consider adding my commit or
- add this step to the init-foundry script for user convenience

Thx a lot for your hardwork guys :)

---

### DIFF
--- a/docs/src/content/hardhat-runner/docs/advanced/hardhat-and-foundry.md
+++ b/docs/src/content/hardhat-runner/docs/advanced/hardhat-and-foundry.md
@@ -10,7 +10,7 @@ How to set up a project that combines Hardhat and Foundry depends on whether you
 
 :::tip
 
-Foundry relies on Git to work properly. Make sure your project is already imported in Git, or type `git init` in your project directory.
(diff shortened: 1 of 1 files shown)


fvictorio:
Hi @Fabien-Morrow, thanks a lot for this.

I don't think we should run `git init` ourselves, so I agree that we should explain this somewhere in the guide.

I'm not sure if that's the right place though. My guess is that most people will already have a git repository, so adding that there seems noisy (plus, I'm not even sure what happens if you run `git init` in an existing repository).

What do you think about adding a "Troubleshooting" section at the end with this scenario? That way people googling the error message will land there and see the explanation. You can see an example of that [here](https://github.com/NomicFoundation/hardhat/blob/dfc4465026cc4ccfadced5cffe84a53ad8acdc50/docs/src/content/hardhat-network/docs/guides/forking-other-networks.md?plain=1#L226-L230).

fvictorio:
> However it's a mandatory step. The tutorial doesn't work if a git repo doesn't exist

That's a great point, and I like the idea of adding a tip block. I just changed the wording a bit and ran prettier. Thank you!

PR #3818, opened 2023-04-06
Ensure concurrency safety across when downloading compilers for multiple hardhat projects

HenryNguyen5:
Introduces a new wrapper around the compiler downloader that ensures multiple downloader calls within a transaction are guaranteed to execute sequentially.

This fixes a concurrency issue that arises from the separation of the `isCompilerDownloaded` and `downloadCompiler` calls, where the previous call to `downloadCompiler` was safe, but `isCompilerDownloaded` was not.

By allowing the composition of calls via `transaction`, and _only_ exposing the underlying downloader via the `transaction` method, we can expose a much safer API surface for managing the downloads of multiple compilers.

This also has the nice side effect of downloading the same compiler once and only once. Before this bug fix, we'd be seeing multiple re-downloads of the same compiler within our project, https://github.com/smartcontractkit/chainlink

I ended up creating a wrapper class rather than adding this functionality to the existing compiler class to preserve the current interface that it adheres to, so we can keep using the same interfaces for future compiler implementations without having to think about concurrency handling at the same time.

---

### DIFF
--- a/.changeset/nasty-moose-watch.md
+++ b/.changeset/nasty-moose-watch.md
@@ -1,0 +1,9 @@
+---
+"hardhat": patch
+---
+
+Ensure concurrency safety across when downloading compilers for multiple hardhat projects
+
+This fixes a concurrency issue that arises from the separation of the `isCompilerDownloaded` and `downloadCompiler` calls, where the previous call to `downloadCompiler` was safe, but `isCompilerDownloaded` was not.
+
+This also has the nice side effect of downloading the same compiler once and only once. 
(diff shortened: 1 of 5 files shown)


HenryNguyen5:
@fvictorio Friendly ping on this as our CI pipelines are pretty flaky when we cache bust the compilers. LMK if you need anything else from me to expedite this!

ChristopherDedominici:
The issue has been resolved in [this PR](https://github.com/NomicFoundation/hardhat/pull/4550)

Issue #3812, opened 2023-04-04
Clarify that hardhat-chai-matchers is meant to be used with ethers


fvictorio:
Our docs are not clear to the fact that the chai matchers are meant to be used with ethers, and that adding support for web3 is out of scope.

Saty248:
@fvictorio is anyone else working on this issue ,else i can start working on this.

PR #3821, opened 2023-04-07
modifying doc to solve the issue#3812

Saty248:

---

hey @fvictorio @alcuadrado I have made these changes in the webpage . I have edited the docs regarding the issue #3812 .
Kindly see the changes and recommend me if any changes are required..

### DIFF
--- a/docs/src/content/docs-landing/index.md
+++ b/docs/src/content/docs-landing/index.md
@@ -33,7 +33,14 @@ Hardhat for Visual Studio Code is a VS Code extension that adds language support
 
 ### Hardhat Chai Matchers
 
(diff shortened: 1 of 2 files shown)


fvictorio:
Hey @Saty248, thanks for the PR but I don't think a tip is the right way to do this.

To be honest, I think this is a change we'll need to do ourselves. Doing it externally is a bit hard because we know our preferences on how to approach documentation changes, but they are hard to explain. Or, put another way: the work I'll have to do to tell you how to do this is the same as just doing the changes :sweat_smile:

I'm going to close now, but we have a "good first issue" label if you are interested in contributing.

Saty248:
> Hey @Saty248, thanks for the PR but I don't think a tip is the right way to do this.
> 
> To be honest, I think this is a change we'll need to do ourselves. Doing it externally is a bit hard because we know our preferences on how to approach documentation changes, but they are hard to explain. Or, put another way: the work I'll have to do to tell you how to do this is the same as just doing the changes 😅
> 
> I'm going to close now, but we have a "good first issue" label if you are interested in contributing.

thanks for the update.. i will look into the good first issues label .

Issue #2730, opened 2022-05-18
Better error message when URL is an empty string


fvictorio:
An empty string in the forking URL or in a network URL produces this error message:

```
Uncaught TypeError [ERR_INVALID_URL]: Invalid URL:
```

We should handle this as a special case and print a better error message. Also, this is not caught during config validation; instead, it happens the first time a request is sent.

[This comment](https://github.com/NomicFoundation/hardhat/issues/2244#issuecomment-1121306153) also seems to suggest that an "only absolute URLs are supported" message can be shown when an empty string is used; maybe that depends on the OS or it happens in other scenarios.

ZainGulbaz:
Hi @fvictorio , Can you please assign this issue to me? Thanks.

ZainGulbaz:
Hi @fvictorio , I have made a pull request for it. You can have a look. Thanks.

ZainGulbaz:
#3825

ZainGulbaz:
#3825 better message when url is an empty string.

kshyun28:
Hello @fvictorio, I'm looking at the list of good first issues and am interested in taking a stab at this issue if its okay.

P.S. your evm-puzzles were really helpful in making me understand evm at a practical level :)

kshyun28:
I have a question, am trying to run tests in the `hardhat-core` package. It's requiring me both `ALCHEMY_URL` and `INFURA_URL`.
![image](https://github.com/NomicFoundation/hardhat/assets/56709653/cd358968-503f-41e0-821c-34f3e29676fc)

I've added a file `packages/hardhat-core/.env` and populated with the following, but still shows the same error. 
```
INFURA_URL=https://sepolia.infura.io/v3/[REDACTED]
ALCHEMY_URL=https://eth-sepolia.g.alchemy.com/v2/[REDACTED]
```

- Where should I be setting the environment variables? 
- And is using Sepolia HTTP url correct, or should I use another network (Mainnet, Goerli, etc.)?

fvictorio:
Hey @kshyun28, sorry for not responding sooner and thanks for the PR!

> Have used mainnet rpc urls after seeing tests fail on Sepolia rpc. Might be obvious, but wasn't clear to me as a first time potential contributor.

That's a great point. Do you want to send a separate PR improving that warning? Ideally we would change the environment variables names too, but that will be a huge change and maybe it's not worth it

> your evm-puzzles were really helpful in making me understand evm at a practical level

[...]

PR #3825, opened 2023-04-08
Better error message when URL is an empty string #2730

ZainGulbaz:
**bug fix**
1- This PR deals with the error if an empty url is passed in config.js.
2- The test is also written for this bug fix.

### DIFF
--- a/packages/hardhat-core/src/internal/core/errors-list.ts
+++ b/packages/hardhat-core/src/internal/core/errors-list.ts
@@ -437,6 +437,13 @@ Please double check your transactions' parameters.`,
 Please check that you are sending an \`address\` parameter.`,
       shouldBeReported: false,
     },
+    EMPTY_URL:{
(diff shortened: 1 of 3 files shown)


fvictorio:
Hi @ZainGulbaz, thanks for sending this, but there are several problems with this PR:

1. There are many linter errors
2. The test you wrote doesn't really work. I think this is causd by the inner await.
3. I think the error messages are wrong: the "only absolute URLs are supported" message is equally confusing as what we have today, and the description mentions a `deploy.js` file for some reason.
4. The error has `shouldBeReported` to true, but this is a user error that can be fixed, we don't want it to be reported to us
5. I don't think the http provider is the right place to do that. It would be better to do it during config validation, so we can tell the user exactly which network(s) have the problem.

I'm going to close this now because there are too many issues, but you can open a new one if you can get all of those things solved.

PR #3828, opened 2023-04-08
Improve error message when URL is empty string

MarioRodrigues10:

---

This PR was created in order to solve #2730


fvictorio:
Hi @MarioRodrigues10, I doubt this PR does what you think it does. The issue is about showing a better error in certain scenarios, and related to how we validate the config. This is changing the value used in a comparison (and it probably introduces a bug).

PR #3838, opened 2023-04-11
fix: remove self references in Hardhat objects

Wodann:

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -239,11 +239,7 @@ export class HardhatNode extends EventEmitter {
       );
     }
 
-    const txPool = new TxPool(
-      (address) => vm.getAccount(address),
-      BigInt(blockGasLimit),
-      common
-    );
+    const txPool = new TxPool(BigInt(blockGasLimit), common);
 
     const instanceId = bufferToBigInt(randomBytes(32));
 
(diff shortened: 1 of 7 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -239,11 +239,7 @@ export class HardhatNode extends EventEmitter {
       );
     }
 
-    const txPool = new TxPool(
-      (address) => vm.getAccount(address),
-      BigInt(blockGasLimit),
-      common
-    );
+    const txPool = new TxPool(BigInt(blockGasLimit), common);
```
fvictorio, line 242: I think this one doesn't leak? If it does, there's something really wrong with my mental model. I also ran an experiment with the (very useful) [FinalizationRegistry](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry) to confirm that this pattern shouldn't leak:

```js
// index.js
const registry = new FinalizationRegistry(key => console.log(`Freeing ${key}`));

class TxPool {
  constructor(cb) {
    this.cb = cb;
  }
}

class Node {
  constructor(vm, txPool) {
    this.vm = vm;
    this.txPool = txPool
  }
}

function create() {
  const vm = {}
  const txPool = new TxPool(() => console.log(vm))
  registry.register(vm, "vm")
  registry.register(txPool, "txPool")
  return new Node(vm, txPool)
}

let node = create()

registry.register(node, "node")

console.log("Run GC");
gc()

node = null;

console.log("Run GC again");
gc()
```

Run it with `node --expose-gc index.js` and you'll see that all the relevant objects are correctly freed.

fvictorio (approved):

PR #3839, opened 2023-04-11
feat: add tracing instrumentation

Wodann:

### DIFF
--- a/crates/rethnet_evm/src/state/debug.rs
+++ b/crates/rethnet_evm/src/state/debug.rs
@@ -1,9 +1,38 @@
+use std::{fmt::Debug, ops::Deref};
+
 use auto_impl::auto_impl;
 use rethnet_eth::{Address, B256, U256};
 use revm::primitives::{AccountInfo, Bytecode};
 
-/// Function type for modifying account information.
-pub type AccountModifierFn = Box<dyn Fn(&mut U256, &mut u64, &mut Option<Bytecode>) + Send>;
+/// Debuggable function type for modifying account information.
+pub struct AccountModifierFn {
+    inner: Box<dyn Fn(&mut U256, &mut u64, &mut Option<Bytecode>) + Send>,
+}
+
+impl AccountModifierFn {
+    /// Constructs an [`AccountModifierDebuggableFn`] from the provided function.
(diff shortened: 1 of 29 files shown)


crates/rethnet_evm/src/state/debug.rs
```diff
@@ -1,9 +1,38 @@
...
+        Self { inner: func }
+    }
+}
+
+impl Debug for AccountModifierFn {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            std::any::type_name::<dyn Fn(&mut U256, &mut u64, &mut Option<Bytecode>)>()
```
fvictorio, line 24: I'm curious why this is useful instead of just formatting it as, say, `rethnet_evm::state::debug::AccountModifierFn` or something.

crates/rethnet_evm_napi/src/state.rs
```diff
@@ -114,15 +125,19 @@ impl StateManager {
...
     #[napi]
+    #[cfg_attr(feature = "tracing", tracing::instrument)]
     pub async fn checkpoint(&self) -> napi::Result<()> {
         self.state
             .checkpoint()
             .await
-            .map_err(|e| napi::Error::new(Status::GenericFailure, e.to_string()))
+            .map_err(|e| napi::Error::new(Status::GenericFailure, e.to_string()))?;
+
+        Ok(())
```
fvictorio, line 135: Is this necessary? This seems redundant compared with the previous version, which still compiles fine AFAICT.

fvictorio (approved):

crates/rethnet_evm_napi/src/state.rs
Wodann: You're right. It's not necessary. The old code was fine

crates/rethnet_evm/src/state/debug.rs
Wodann: I just wanted to see the internal function handle: parameter and return types

crates/rethnet_evm_napi/src/state.rs
Wodann: You're right. It's not necessary. The old code was fine

PR #3842, opened 2023-04-11
refactor: replace background with short-lived tasks

Wodann:


crates/rethnet_evm_napi/src/context.rs
```diff
@@ -0,0 +1,79 @@
...
+    Status,
+};
+use napi_derive::napi;
+use tracing_subscriber::{prelude::*, EnvFilter, Registry};
+
+#[napi]
+#[derive(Debug)]
+pub struct RethnetContext {
+    inner: Arc<Context>,
+}
```
fvictorio, line 14: What does NAPI-RS do with `Arc`s? I searched their docs but couldn't find it.

fvictorio (approved):
This one went over my head, so I'm just leaving some random questions from skimming the code

crates/rethnet_evm_napi/src/context.rs
Wodann: Nothing. This is a native Rust object that is exposed to JS using a handle.

PR #3851, opened 2023-04-11
Fix gas-provider error caused by empty blocks

LEAFERx:

---

On a chain that supports EIP-1559, when there's an empty block, the reward returned from `eth_feeHistory` will be 0. However, `AutomaticGasPriceProvider` should set `maxPriorityFeePerGas` to be at least 1 to avoid `ProviderError: FeeTooLow, EffectivePriorityFeePerGas too low 0 < 1`.

This bug is found on [Chiado](https://docs.gnosischain.com/about/networks/chiado/), a minimal reproducible example and detailed analysis can be found [here](https://github.com/LEAFERx/chiado-ethers-mre).


alcuadrado:
Hey, sorry we never merged this, and thanks for sending this PR. I'm closing it as I believe it has already been fixed.

PR #3868, opened 2023-04-18
feat: assorted synthetic benchmarks for `SyncState`

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/71

### DIFF
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -127,8 +127,5 @@ fn bench_basic(c: &mut Criterion) {
     }
 }
 
-    name = benches;
-    config = Criterion::default().significance_level(0.1).sample_size(10);
-    targets = bench_insert_account, bench_checkpoint, bench_basic
-}
+criterion_group!(benches, bench_insert_account, bench_checkpoint, bench_basic);
 criterion_main!(benches);
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -3,9 +3,7 @@ use revm::{db::StateRef, primitives::AccountInfo};
 use std::{clone::Clone, str::FromStr};
 
 use rethnet_eth::Address;
-use rethnet_evm::state::{
-    HybridState, LayeredState, RethnetLayer, StateDebug, StateError, SyncState,
-};
+use rethnet_evm::state::{HybridState, LayeredState, RethnetLayer, StateError, SyncState};
 
 #[derive(Default)]
 struct RethnetStates {
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -17,7 +17,7 @@ impl RethnetStates {
         for state in states.iter_mut() {
             let number_of_checkpoints = number_of_accounts / number_of_accounts_per_checkpoint;
             for _ in 0..=number_of_checkpoints {
-                for i in 1..=number_of_accounts {
+                for i in 1..=number_of_accounts_per_checkpoint {
                     state
                         .insert_account(
                             Address::from_str(&format!("0x{:0>40x}", i)).unwrap(),
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -1,9 +1,9 @@
-use criterion::{criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
-use revm::{db::StateRef, primitives::AccountInfo};
 use std::{clone::Clone, str::FromStr};
 
+use criterion::{criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
 use rethnet_eth::Address;
 use rethnet_evm::state::{HybridState, LayeredState, RethnetLayer, StateError, SyncState};
+use revm::{db::StateRef, primitives::AccountInfo};
 
 #[derive(Default)]
 struct RethnetStates {
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -44,8 +44,15 @@ impl RethnetStates {
     }
 }
 
-const ADDRESS_SCALES: [u64; 4] = [100, 500, 1000, 2000];
-const CHECKPOINT_SCALES: [u64; 4] = [1, 2, 4, 8];
+const NUM_SCALES: usize = 4;
+const CHECKPOINT_SCALES: [u64; NUM_SCALES] = [1, 5, 10, 20];
+const MAX_CHECKPOINT_SCALE: u64 = CHECKPOINT_SCALES[NUM_SCALES - 1];
+const ADDRESS_SCALES: [u64; 4] =  [
+    MAX_CHECKPOINT_SCALE * 5,
+    MAX_CHECKPOINT_SCALE * 25,
+    MAX_CHECKPOINT_SCALE * 50,
+    MAX_CHECKPOINT_SCALE * 100,
+];
 
 fn bench_sync_state_method<O, R>(c: &mut Criterion, method_name: &str, mut method_invocation: R)
 where
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -12,7 +12,7 @@ struct RethnetStates {
 }
 
 impl RethnetStates {
-    fn fill(&mut self, number_of_accounts: u64, number_of_layers_per_account: u64) {
+    fn fill(&mut self, number_of_accounts: u64, number_of_accounts_per_checkpoint: u64) {
         let mut states: [&mut dyn SyncState<StateError>; 2] = [&mut self.layered, &mut self.hybrid];
         for state in states.iter_mut() {
             for i in 1..=number_of_accounts {
@@ -22,7 +22,7 @@ impl RethnetStates {
                         AccountInfo::default(),
                     )
                     .unwrap();
-                for _ in 1..=number_of_layers_per_account {
+                if i % number_of_accounts_per_checkpoint == 0 {
                     state.checkpoint().unwrap();
                 }
             }
@@ -44,18 +44,18 @@ impl RethnetStates {
 }
 
 const ADDRESS_SCALES: [u64; 4] = [100, 500, 1000, 2000];
-const LAYER_SCALES: [u64; 4] = [1, 2, 4, 8];
+const CHECKPOINT_SCALES: [u64; 4] = [1, 2, 4, 8];
 
 fn bench_insert_account(c: &mut Criterion) {
     for number_of_accounts in ADDRESS_SCALES.iter() {
-        for layers_per_account in LAYER_SCALES.iter() {
+        for accounts_per_checkpoint in CHECKPOINT_SCALES.iter() {
             let mut rethnet_states = RethnetStates::default();
-            rethnet_states.fill(*number_of_accounts, *layers_per_account);
+            rethnet_states.fill(*number_of_accounts, *accounts_per_checkpoint);
 
             for (label, state_factory) in rethnet_states.make_clone_factories().into_iter() {
                 c.benchmark_group(format!(
-                    "StateRef::insert_account() with {} accounts with {} layer(s) per account",
-                    *number_of_accounts, layers_per_account,
+                    "StateRef::insert_account() with {} accounts and with {} account(s) per checkpoint",
+                    *number_of_accounts, accounts_per_checkpoint,
                 ))
                 .bench_function(label, |b| {
                     b.iter_batched(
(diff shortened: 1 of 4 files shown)


crates/rethnet_evm/benches/state_ref_basic.rs
```diff
@@ -0,0 +1,94 @@
+use criterion::{black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
+use revm::primitives::AccountInfo;
+use std::str::FromStr;
+
+use rethnet_eth::Address;
+use rethnet_evm::state::{LayeredState, RethnetLayer};
+
+fn make_state(number_of_addresses: u64) -> LayeredState<RethnetLayer> {
+    let mut state = LayeredState::<RethnetLayer>::default();
+    for i in 1..number_of_addresses {
```
Wodann, line 10:
```suggestion
    for i in 1..=number_of_addresses {
```

This makes it an inclusive range, leading to `number_of_addresses` entries

crates/rethnet_evm/benches/state_ref_basic.rs
```diff
@@ -0,0 +1,94 @@
...
+            .unwrap();
+        state.checkpoint().unwrap();
+    }
+    state
+}
+
+const ADDRESS_SCALES: [u64; 4] = [100, 500, 1000, 2000];
+
+fn bench_layered_create(c: &mut Criterion) {
+    for number_of_addresses in ADDRESS_SCALES.iter() {
```
Wodann, line 26: Could you set it up such that there is an additional for loop over the different types of `StateRef` implementations? That way we only need one function per benchmark and can reuse it for all (state type, no. addresses) pairs.

You can achieve this by having a closure that returns a `Box<dyn StateRef>`: `Vec<Box<dyn Fn(u64) -> Box<dyn StateRef>>>`

For now, we'll just have the `LayeredState`

crates/rethnet_evm/benches/state_ref_basic.rs
```diff
@@ -0,0 +1,94 @@
...
+                "layered_basic_with_clone_with_{}_addresses",
+                *number_of_addresses,
+            ),
+            number_of_addresses,
+            |b, number_of_addresses| {
+                b.iter_batched(
+                    || state.clone(),
+                    |mut state| {
+                        for i in *number_of_addresses..1 {
+                            use revm::db::State;
```
Wodann, line 76: Could you please move all of the local `use` statements to the top of the file. There is no benefit in Rust to have them locally, in this case.

Use cases where it would be helpful is:
- if you have conflicting naming; or
- if you are including `use` statement in a scope that's behind a feature flag, as you'd also need to put the `use` statement itself behind the feature flag if you do it at the top of the file.

Wodann (changes requested):

crates/rethnet_evm/benches/state_ref_basic.rs
feuGeneA: I'm sorry I've been stubborn about this, and I realize that there is no benefit *in Rust*, but as a reader of the code I actually find it immensely beneficial. Given that such imported traits have no direct mention in the file, I appreciate having their placement restricted to the scope they're actually needed in. If that scope is the entire file, due to repeated uses within the file, then at the top of the file is the right place. But if the use is only in a single inner scope, then I feel that the placement should be within that single inner scope. (Note that in this particular case I do have a repetition of `revm::db::State` within this file, but that's because I expect one of those usages to go away depending on which version of that function we end up keeping.)

Wodann: A language server provides you that information if you need a reminder though

crates/rethnet_evm/benches/state.rs
```diff
@@ -0,0 +1,137 @@
...
+                    b.iter_batched(
+                        || state_factory(),
+                        |mut state| {
+                            state.insert_account(
+                                Address::from_str(&format!("0x{:0>40x}", number_of_accounts + 1))
+                                    .unwrap(),
+                                AccountInfo::default(),
+                            )
+                        },
+                        BatchSize::LargeInput,
```
feuGeneA, line 72: For this and all other mentions of `BatchSize`, consider whether `LargeInput` really is the best thing to be using.

crates/rethnet_evm/benches/state.rs
```diff
@@ -0,0 +1,137 @@
...
+                    BatchSize::LargeInput,
+                )
+            });
+        }
+    }
+}
+
+criterion_group! {
+    name = benches;
+    config = Criterion::default().significance_level(0.1).sample_size(10);
```
feuGeneA, line 134: See if we really need this sample size and significance level modification. We probably only need to do it for particular benchmarks (eg `checkpoint`) rather than ALL of them like it's coded for here.

crates/rethnet_evm/benches/state_ref_basic.rs
feuGeneA: true, but what if for example i'm just browsing the code on github? etc. maybe i'm being overly principled, but it doesn't seem right to *depend* on tooling for readability. in my latest draft all of the `use` are at the top, by virtue of using `SyncState`, so it's a moot point for this PR, but this does remain one of the (very) few smelly things in my limited Rust experience.

crates/rethnet_evm/benches/state.rs
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3868/commits/7b6790bda940a4141880d3a63a7101d72eeda4d4

crates/rethnet_evm/benches/state.rs
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3868/commits/1c0c5c3b61ab5ac1246e83eb0a702846e27c7f34

crates/rethnet_evm/benches/state.rs
```diff
@@ -0,0 +1,131 @@
...
+                        BatchSize::SmallInput,
+                    );
+                });
+            }
+        }
+    }
+}
+
+fn bench_checkpoint(c: &mut Criterion) {
+    for number_of_accounts in ADDRESS_SCALES.iter() {
```
Wodann, line 79: A lot of this code repeats itself for every benchmark. Could you refactor it to remove code duplication?

Wodann (changes requested):

crates/rethnet_evm/benches/state.rs
```diff
@@ -0,0 +1,131 @@
...
+use rethnet_evm::state::{HybridState, LayeredState, RethnetLayer, StateError, SyncState};
+
+#[derive(Default)]
+struct RethnetStates {
+    layered: LayeredState<RethnetLayer>,
+    hybrid: HybridState<RethnetLayer>,
+}
+
+impl RethnetStates {
+    fn fill(&mut self, number_of_accounts: u64, number_of_layers_per_account: u64) {
```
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3868/commits/b9040fb8bc7ac3b1b369ac4889d003e564c3db76

crates/rethnet_evm/benches/state.rs
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3868/commits/b1f392822bb51b937ce7008af7b6d1c5f65c019d

I wanted to take it even further, by having a single `c.benchmark_group(...).bench_function(...)`, within a loop over an array of `(group_label, state_action)` tuples, where a `state_action` conveys the second parameter to `iter_batched()`, but I started running into Boxes and return type differences and couldn't figure it out in a reasonable amount of time. If you'd like to see that happen, maybe we can pair on it to expedite things.

crates/rethnet_evm/benches/state_ref_basic.rs
Wodann: GitHub these days also offers redirects to type definitions. It seems like you're trying to tailor to a niche while cluttering the code with unnecessary noise for the majority of use cases.

crates/rethnet_evm/benches/state.rs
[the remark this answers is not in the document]
Wodann: This might be the perfectionist in me, but this approach has two caveats:
1. Not all checkpoints have an equal number of accounts
2. It's more costly to set up

Alternatively, you could do

```rust
for _ in 0..number_of_checkpoints {
  for i in 1..=number_of_accounts_per_checkpoint {
    ..
  }
}
```

which avoids the `%` operator and has an equal number of accounts per checkpoint

crates/rethnet_evm/benches/state.rs
Wodann: I don't really understand this change. We should still have separate benchmarks, but they should be reusing the parts of the code that are shared, split off into a separate function.

I'm happy to pair on it tomorrow.

crates/rethnet_evm/benches/state.rs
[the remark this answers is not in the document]
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3868/commits/b9040fb8bc7ac3b1b369ac4889d003e564c3db76

crates/rethnet_evm/benches/state.rs
feuGeneA: Reverted the previous change in https://github.com/NomicFoundation/hardhat/pull/3868/commits/102e7cb25cdc8cce1363be283e962727ee61ba5a and re-did it in https://github.com/NomicFoundation/hardhat/pull/3868/commits/26fcefb1976bcbb561749ea0bfed60f93500154b

crates/rethnet_evm/benches/state.rs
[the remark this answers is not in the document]
feuGeneA: Actually that latest change is no good. it drastically increased the test setup time (like over 90 minutes just for the `insert` benchmark). I need to look further at this, so moving the PR back to Draft status.

crates/rethnet_evm/benches/state.rs
```diff
@@ -0,0 +1,133 @@
...
+    )> {
+        vec![
+            ("LayeredState", Box::new(|| Box::new(self.layered.clone()))),
+            ("HybridState", Box::new(|| Box::new(self.hybrid.clone()))),
+        ]
+    }
+}
+
+const ADDRESS_SCALES: [u64; 4] = [100, 500, 1000, 2000];
+const CHECKPOINT_SCALES: [u64; 4] = [1, 2, 4, 8];
```
Wodann, line 48:
```suggestion
const NUM_SCALES: usize = 4;
const CHECKPOINT_SCALES: [u64; NUM_SCALES] = [1, 5, 10, 20];
const MAX_CHECKPOINT_SCALE: u64 = CHECKPOINT_SCALES[NUM_SCALES - 1];
const ADDRESS_SCALES: [u64; 4] =  [
    MAX_CHECKPOINT_SCALE * 5,
    MAX_CHECKPOINT_SCALE * 25,
    MAX_CHECKPOINT_SCALE * 50,
    MAX_CHECKPOINT_SCALE * 100,
];
```

This ensures we can always divide the address scale by the checkpoint scale, avoiding flooring by division.

crates/rethnet_evm/benches/state.rs
```diff
@@ -0,0 +1,133 @@
...
+impl RethnetStates {
+    fn fill(&mut self, number_of_accounts: u64, number_of_accounts_per_checkpoint: u64) {
+        let mut states: [&mut dyn SyncState<StateError>; 2] = [&mut self.layered, &mut self.hybrid];
+        for state in states.iter_mut() {
+            let number_of_checkpoints = number_of_accounts / number_of_accounts_per_checkpoint;
+            for _ in 0..=number_of_checkpoints {
+                for i in 1..=number_of_accounts {
+                    state
+                        .insert_account(
+                            Address::from_str(&format!("0x{:0>40x}", i)).unwrap(),
```
Wodann, line 23: One of the `from_low_u64_*` methods will likely be a lot faster than creating and parsing a string.

crates/rethnet_evm/benches/state.rs
```diff
@@ -0,0 +1,133 @@
...
+fn bench_checkpoint(c: &mut Criterion) {
+    bench_sync_state_method(
+        c,
+        "StateHistory::checkpoint()",
+        |mut state, _number_of_accounts| state.checkpoint(),
+    );
+}
+
+fn bench_basic(c: &mut Criterion) {
+    let mut group = c.benchmark_group("StateRef::basic()");
```
Wodann, line 105: Why does `bench_basic` not use `bench_sync_state_method`?

Wodann (changes requested):

crates/rethnet_evm/benches/state.rs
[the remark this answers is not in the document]
feuGeneA: It was a simple mistake, fixed in https://github.com/NomicFoundation/hardhat/pull/3868/commits/9b6d427b5526625bd786b6e0a84fa58af14d6ea4

crates/rethnet_evm/benches/state.rs
feuGeneA: I quickly whipped up a couple of benchmarks to confirm your hunch, and you're right:
image
Addressed in https://github.com/NomicFoundation/hardhat/pull/3868/commits/9e7c6bed702a8d753e487e953197b2e17336bfbc

crates/rethnet_evm/benches/state.rs
feuGeneA: Good question. I guess I thought that it wouldn't work for some reason, but I just tried it and it works fine, so I made that change in https://github.com/NomicFoundation/hardhat/pull/3868/commits/20bf651e9d2473b83ac690017fabe7ddfc693379

Wodann (approved):

Issue #3771, opened 2023-03-17
"function call to a non-contract account" heuristic should include the address


fvictorio:
This message would be way more useful if it included the address that was called.

ZainGulbaz:
@fvictorio  Kindly let me know if you want any change or addition regarding my pull request for this improvement. Thanks <3.

PR #3870, opened 2023-04-19
 "function call to a non-contract account" heuristic should include the address

ZainGulbaz:
1- The pull request includes the improvement.

There is the improvement in the error message that is shown when a non contract account function is called.
Previous Message:
**Transaction reverted: function call to a non-contract account**
Improved Message:
**Transaction reverted: function call to a non-contract account 0x009fd6e51aa188f6f4ce6ab8827279cfffb92200**

The PR is related to #3771


alcuadrado:
Sorry we never got to merge this. Unfortuantely the code has changed so much that it can't be merged now.

PR #3873, opened 2023-04-21
feat(core): return forkUrl in hardhat_metadata

0xdapper:
- [x] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team. Created after opening the PR: https://github.com/NomicFoundation/edr/issues/442

---

Expose `forkUrl` so users are able to reset a forked state back to original fork block number with the same RPC queried from `hardhat_metadata`.

### DIFF
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/output/metadata.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/output/metadata.ts
@@ -28,5 +28,8 @@ export interface HardhatMetadata {
 
     // The hash of the block that the network forked from.
     forkBlockHash: string;
+
+    // The json rpc URL of the network that is being forked.
+    forkUrl: string;
   };
 }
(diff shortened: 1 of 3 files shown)


fvictorio:
Interesting, thanks for this @0xdapper.

I wonder a bit about the security implications of this, since the URL can have a secret key. This seems far-fetched, but to give you a real example: some people have used exposed Hardhat nodes in workshops, and in those cases they might not want to have this information exposed.

I'll discuss this with @alcuadrado and get back to you.

0xdapper:
I see, yeah, I didn't consider anyone was using non-local/semi-public hardhat instances and sharing it and if they are I assume they wouldn't have a problem sharing the secrets too?

alcuadrado:
Hey, sorry this was never merged. Unfortunately this can't be merged now, as this is now implemented in EDR. I moved the issue to its repo.

Issue #3583, opened 2023-01-13
Remove unnecessary "VM Exception while processing transaction" prefix in error messages


fvictorio:
Several error messages look like this:

```
Error: VM Exception while processing transaction: reverted with reason string 'reason'
```

which is super noisy.

Notice that a `require(false)`, for some reason, doesn't include that:

```
Error: Transaction reverted without a reason string
```

This is even worse when ethers estimate gas, because the first line is huge and the noisy prefix doesn't help:

```
Error: cannot estimate gas; transaction may fail or may require manual gas limit [ 
See: https://links.ethers.org/v5-errors-UNPREDICTABLE_GAS_LIMIT ]
(reason="VM Exception while processing transaction: reverted with reason string 'reason'", ...continues...
```

So the error message is there, but hard to see.

PR #3875, opened 2023-04-23
Removed VM Exception string where not necessary #3583

lucapdr01:

---

As stated in issue #3583 I removed "VM Exception while processing transaction" prefix in error messages where not relevant, also modified the corresponding test cases


alcuadrado:
Thanks for submitting this PR, and sorry we didn't get to it sooner. Unfortunately this change is too risky, as some libraries have historically depend on this. We may do this change in the new major version that we are working on.

PR #3881, opened 2023-04-25
feat: add worst-case benchmark for `StateRef::code_by_hash`

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/72

### DIFF
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -1,9 +1,12 @@
 use std::{clone::Clone, str::FromStr};
 
 use criterion::{criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion};
-use rethnet_eth::Address;
+use rethnet_eth::{Address, Bytes, U256};
 use rethnet_evm::state::{HybridState, LayeredState, RethnetLayer, StateError, SyncState};
-use revm::{db::StateRef, primitives::AccountInfo};
+use revm::{
+    db::StateRef,
+    primitives::{AccountInfo, Bytecode},
+};
 
 #[derive(Default)]
 struct RethnetStates {
@@ -16,10 +19,20 @@ impl RethnetStates {
         let mut states: [&mut dyn SyncState<StateError>; 2] = [&mut self.layered, &mut self.hybrid];
         for state in states.iter_mut() {
             let number_of_checkpoints = number_of_accounts / number_of_accounts_per_checkpoint;
-            for _ in 0..=number_of_checkpoints {
-                for i in 1..=number_of_accounts_per_checkpoint {
+            for checkpoint_number in 0..number_of_checkpoints {
+                for account_number in 1..=number_of_accounts_per_checkpoint {
+                    let account_number =
+                        (checkpoint_number * number_of_accounts_per_checkpoint) + account_number;
+                    let address = Address::from_low_u64_ne(account_number);
                     state
-                        .insert_account(Address::from_low_u64_ne(i), AccountInfo::default())
+                        .insert_account(
+                            address,
+                            AccountInfo::new(
+                                U256::from(account_number),
+                                account_number,
+                                Bytecode::new_raw(Bytes::copy_from_slice(address.as_bytes())),
+                            ),
+                        )
                         .unwrap();
                 }
                 state.checkpoint().unwrap();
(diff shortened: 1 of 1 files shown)


crates/rethnet_evm/benches/state.rs
```diff
@@ -16,10 +19,19 @@ impl RethnetStates {
...
+                    let address =
+                        (checkpoint_number * number_of_accounts_per_checkpoint) + account_number;
                     state
-                        .insert_account(Address::from_low_u64_ne(i), AccountInfo::default())
+                        .insert_account(
+                            Address::from_low_u64_ne(address),
+                            AccountInfo::new(
+                                U256::from(address),
+                                address,
+                                Bytecode::new_raw(Bytes::from(format!("0x{address:x}"))),
```
Wodann, line 32: address is a byte hash, so you can literally use those bytes directly.

crates/rethnet_evm/benches/state.rs
```diff
@@ -115,5 +127,21 @@ fn bench_basic(c: &mut Criterion) {
     });
 }
 
-criterion_group!(benches, bench_insert_account, bench_checkpoint, bench_basic);
+fn bench_code_by_hash(c: &mut Criterion) {
+    bench_sync_state_method(c, "StateRef::code_by_hash", |state, number_of_accounts| {
+        for i in (1..=number_of_accounts).rev() {
+            state
+                .code_by_hash(Bytecode::new_raw(Bytes::from(format!("0x{i:x}"))).hash())
```
Wodann, line 134: Using the improvement mentioned above, combined with `Address::from_low_u64_ne`, you can make this more performant, reducing the overhead that string operations will have on your benchmark

crates/rethnet_evm/benches/state.rs
feuGeneA: I'm not sure what you have in mind here. You're talking about line 32, right?

Using `address` directly for this parameter doesn't work because in this context it's a `u64`: ```expected struct `Bytecode`, found `u64` ```

but it seems like you're talking about an `Address`, which i understand is a byte hash, but using `Address::from_low_u64_ne(address)` for this parameter doesn't work either: ```expected struct `Bytecode`, found struct `B160` ```

i also tried using `Bytecode::new_raw(Address::from_low_u64_ne(address))` but that didn't work either: ```expected struct `rethnet_eth::Bytes`, found struct `B160` ```

are you perhaps confusing `code` with `code_hash` and thinking this argument is expecting a hash?

crates/rethnet_evm/benches/state.rs
feuGeneA: Similar to the other comment, I'm confused about what you're looking for here. It's important that the object on which `.hash()` is called here is constructed the same exact way as done in `fill()` (where you left your other comment), so that the input will be a hash whose code is actually present.

crates/rethnet_evm/benches/state.rs
Wodann: `Address` is just a 20-byte static array. `Bytecode` can be constructed from a byte slice. So - one way of avoiding the string generation, it being the most expensive way of converting - you can generate the address from the u64 and then grab the byte slice of it to instantiate the Bytecode

feuGeneA: I think I achieved what you wanted (thought it didn't work out _exactly_ as you suggested) in https://github.com/NomicFoundation/hardhat/pull/3881/commits/2074b056344a742c870a3afd12c4efd5b9c8a559. Let me know what you think.

crates/rethnet_evm/benches/state.rs
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3881/commits/2074b056344a742c870a3afd12c4efd5b9c8a559.

Wodann: This is what I had in mind:

```rust
Bytecode::new_raw(Bytes::copy_from_slice(address.as_bytes()))
```

crates/rethnet_evm/benches/state.rs
```diff
@@ -16,10 +19,27 @@ impl RethnetStates {
         let mut states: [&mut dyn SyncState<StateError>; 2] = [&mut self.layered, &mut self.hybrid];
         for state in states.iter_mut() {
             let number_of_checkpoints = number_of_accounts / number_of_accounts_per_checkpoint;
-            for _ in 0..=number_of_checkpoints {
-                for i in 1..=number_of_accounts_per_checkpoint {
+            for checkpoint_number in 0..=number_of_checkpoints {
```
Wodann, line 22:
```suggestion
            for checkpoint_number in 0..number_of_checkpoints {
```
This should be an exclusive range, as we're starting at 0

Wodann (changes requested):

Wodann (approved):

PR #3887, opened 2023-04-26
fix: store code when inserting account in `LayeredState`

feuGeneA:


crates/rethnet_evm/src/state/layered/changes.rs
```diff
@@ -344,6 +344,13 @@ impl LayeredChanges<RethnetLayer> {
     pub fn remove_code(&mut self, code_hash: &B256) {
         self.last_layer_mut().contracts.remove(code_hash);
     }
+
+    pub fn insert_account(&mut self, address: &Address, account_info: &AccountInfo) {
+        self.account_or_insert_mut(address).info = account_info.clone();
+        if let Some(code) = account_info.code.clone() {
+            self.insert_code(code);
+        }
+    }
```
Wodann, line 353:
```suggestion
    pub fn insert_account(&mut self, address: &Address, mut account_info: AccountInfo) {
        if let Some(code) = account_info.code.take() {
            self.insert_code(code);
        }

        self.account_or_insert_mut(address).info = account_info;
    }
```

Wodann: 1) As we're inserting, we can just explicitly state the intention in the API that we're moving the ownership of the `AccountInfo` instead of requesting a reference.
2) We don't want to store the code in the account itself but store it in a separate data structure. The method to move ownership of a value in an `Option` is by using the `take` method. We do this before assigning the `account_info` to the account

crates/rethnet_evm/src/state/layered.rs
```diff
@@ -94,7 +94,7 @@ impl StateDebug for LayeredState<RethnetLayer> {
         address: Address,
         account_info: AccountInfo,
     ) -> Result<(), Self::Error> {
-        self.changes.account_or_insert_mut(&address).info = account_info;
+        self.changes.insert_account(&address, &account_info);
```
Wodann, line 97:
```suggestion
        self.changes.insert_account(&address, account_info);
```

crates/rethnet_evm/src/state/hybrid.rs
```diff
@@ -88,7 +88,7 @@ impl StateDebug for HybridState<RethnetLayer> {
         account_info: AccountInfo,
     ) -> Result<(), Self::Error> {
         self.trie.insert_account(address, account_info.clone())?;
-        self.changes.account_or_insert_mut(&address).info = account_info;
+        self.changes.insert_account(&address, &account_info);
```
Wodann, line 91:
```suggestion
        self.changes.insert_account(&address, account_info);
```

crates/rethnet_evm/src/state/layered.rs
```diff
@@ -275,3 +275,24 @@ impl StateHistory for LayeredState<RethnetLayer> {
...
+    use rethnet_eth::Bytes;
+
+    #[test]
+    fn code_by_hash_success() {
+        let mut state = LayeredState::<RethnetLayer>::default();
+        let inserted_bytecode = Bytecode::new_raw(Bytes::from("0x11"));
+        state
+            .insert_account(
+                Address::from_low_u64_ne(1234),
+                AccountInfo::new(U256::from(0), 0, inserted_bytecode.clone()),
```
Wodann, line 292:
```suggestion
                AccountInfo::new(U256::ZERO, 0, inserted_bytecode.clone()),
```

Wodann (changes requested):

Wodann (approved):

PR #3895, opened 2023-05-03
Handle 429 responses withouto retry-after header

### LINKED ISSUE #3501: Handle 429 (Too many requests) errors without Retry-After headers and in batch transactions
The error manifests with the following error when calling `debug_traceTransaction`:
```
	err: Error: restoreForkBlockContext called when checkpointed
	at checkpointedError (/usr/src/app/node_modules/hardhat/internal/hardhat-network/provider/fork/ForkStateManager.js:15:39)
```

It took me a bit of debugging but I found out that the error happens because `HttpProvider` doesn't handle rateLimitting errors correctly.
I'm getting two types of rateLimitting errors from Alchemy.

1. Normal 429 response. This is supposed to be handled in https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/core/providers/http.ts#L184, however the implemntation assumes that there is `Retry-After` response header. Alchemy docs (https://docs.alchemy.com/reference/throughput#option-2-retry-after) specifically say that this response should be available, but these are the response headers that I'm actually getting:
```
{
  date: 'Thu, 29 Dec 2022 12:23:16 GMT',
  'content-type': 'application/json',
  'content-length': '248',
  connection: 'keep-alive',
  'x-alchemy-trace-id': '469ec7ad3503e37d37fb6a5e2e4c91cf',
  'cf-cache-status': 'DYNAMIC',
  server: 'cloudflare',
  'cf-ray': '78128188e9265b4d-IAD'
}
```

As a result the error is just returned.

To work around this issue I implemented an exponential back-off as suggested by the later part of the Alchemy docs (https://docs.alchemy.com/reference/throughput#option-4-exponential-backoff)

My suggested fix for this issue goes as follows:

```
diff --git a/packages/hardhat-core/src/internal/core/providers/http.ts b/packages/hardhat-core/src/internal/core/providers/http.ts
index bff610414..9cfad9462 100644
--- a/packages/hardhat-core/src/internal/core/providers/http.ts
+++ b/packages/hardhat-core/src/internal/core/providers/http.ts
@@ -189,7 +189,7 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
         // https://undici.nodejs.org/#/?id=garbage-collection
         // It's not clear how to "cancel", so we'll just consume:
         await response.body.text();
-        const seconds = this._getRetryAfterSeconds(response);
+        const seconds = this._getRetryAfterSeconds(response, retryNumber);
         if (seconds !== undefined && this._shouldRetry(retryNumber, seconds)) {
           return await this._retry(request, seconds, retryNumber);
         }
@@ -260,12 +260,13 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
   }
 
   private _getRetryAfterSeconds(
-    response: Dispatcher.ResponseData
+    response: Dispatcher.ResponseData,
+    retryNumber: number
   ): number | undefined {
     const header = response.headers["retry-after"];
 
     if (header === undefined || header === null) {
-      return undefined;
+      return Math.min(2 ** retryNumber, MAX_RETRY_AWAIT_SECONDS);
     }
 
     const parsed = parseInt(header, 10);
```

2. Rate limiting error when using `HttpProvider.sendBatch(..)` method. 

ForkStateManager is using batch requests. Sadly Alchemy sometimes returns a 200 response with a response body like:
```
[
  {
    "jsonrpc": "2.0",
    "id": 236,
    "result": "0x"
  },
  {
    "jsonrpc": "2.0",
    "id": 237,
    "error": {
      "code": 429,
      "message": "Your app has exceeded its compute units per second capacity. If you have retries enabled, you can safely ignore this message. If not, check out https://docs.alchemyapi.io/guides/rate-limits"
    }
  },
  {
    "jsonrpc": "2.0",
    "id": 238,
    "result": "0x33fe288aee4bf"
  }
]
```

HttpProvider has no logic  to handle an error returned like that. To workaround this issue I've made the following changes to the implementation:
```
diff --git a/packages/hardhat-core/src/internal/core/providers/http.ts b/packages/hardhat-core/src/internal/core/providers/http.ts
index bff610414..b3fe22f04 100644
--- a/packages/hardhat-core/src/internal/core/providers/http.ts
+++ b/packages/hardhat-core/src/internal/core/providers/http.ts
@@ -120,16 +120,32 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
       return jsonRpcRequest;
     });
 
-    const jsonRpcResponses = await this._fetchJsonRpcResponse(requests);
-
-    for (const response of jsonRpcResponses) {
-      if (isErrorResponse(response)) {
-        error.message = response.error.message;
-        error.code = response.error.code;
-        error.data = response.error.data;
-        // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
-        throw error;
+    let jsonRpcResponses;
+    let retryNumber = 0;
+    const seconds = 1;
+    let needsToRetry = false;
+    do {
+      if (retryNumber > 0) {
+        await new Promise((resolve) => setTimeout(resolve, 1000 * seconds));
+      }
+      jsonRpcResponses = await this._fetchJsonRpcResponse(requests);
+      needsToRetry = false;
+      for (const response of jsonRpcResponses) {
+        if (isErrorResponse(response)) {
+          error.message = response.error.message;
+          error.code = response.error.code;
+          error.data = response.error.data;
+          if (response.error.code === 429) {
+            needsToRetry = true;
+          } else {
+            // eslint-disable-next-line @nomiclabs/hardhat-internal-rules/only-hardhat-error
+            throw error;
+          }
+        }
       }
+    } while (needsToRetry && this._shouldRetry(++retryNumber, seconds));
+    if (needsToRetry) {
+      throw error;
     }

```

The idea of this changes is to retry the batch request if any of partial responses comes back with error.code = 429. Since there is no Retry-After header or other mechanism of specifying how long the client should wait, I decided to just wait 1 second (for simplicity sake). 

With these two fixes in place I'm able to consistently get debug_traceTransaction output from ethereum mainnet using Alchemy endpoint. I'd love these changes to be incorporated into upstream, so that I don't have to maintain my own patches.

fvictorio:
Closes #3501

### DIFF
--- a/.changeset/ten-hairs-attend.md
+++ b/.changeset/ten-hairs-attend.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed a bug caused by nodes returning 429 responses without a `Retry-After` header (thanks @kowalski!)
--- a/packages/hardhat-core/src/internal/core/providers/http.ts
+++ b/packages/hardhat-core/src/internal/core/providers/http.ts
@@ -203,7 +203,7 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
         // https://undici.nodejs.org/#/?id=garbage-collection
         // It's not clear how to "cancel", so we'll just consume:
         await response.body.text();
-        const seconds = this._getRetryAfterSeconds(response);
+        const seconds = this._getRetryAfterSeconds(response, retryNumber);
         if (seconds !== undefined && this._shouldRetry(retryNumber, seconds)) {
           return await this._retry(request, seconds, retryNumber);
         }
@@ -272,12 +272,15 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
   }
 
   private _getRetryAfterSeconds(
-    response: Undici.Dispatcher.ResponseData
+    response: Undici.Dispatcher.ResponseData,
+    retryNumber: number
   ): number | undefined {
     const header = response.headers["retry-after"];
 
     if (header === undefined || header === null || Array.isArray(header)) {
-      return undefined;
+      // if the response doesn't have a retry-after header, we do
+      // an exponential backoff
+      return Math.min(2 ** retryNumber, MAX_RETRY_AWAIT_SECONDS);
     }
 
     const parsed = parseInt(header, 10);
--- a/packages/hardhat-core/test/internal/core/providers/http.ts
+++ b/packages/hardhat-core/test/internal/core/providers/http.ts
@@ -4,8 +4,6 @@ import { MockAgent, MockPool } from "undici";
 import { HttpProvider } from "../../../../src/internal/core/providers/http";
 import { SuccessfulJsonRpcResponse } from "../../../../src/internal/util/jsonrpc";
 
-import { expectErrorAsync } from "../../../helpers/errors";
-
 const TOO_MANY_REQUEST_STATUS = 429;
 
 function makeMockPool(url: string): MockPool {
--- a/packages/hardhat-core/test/internal/core/providers/http.ts
+++ b/packages/hardhat-core/test/internal/core/providers/http.ts
@@ -40,16 +40,26 @@ describe("HttpProvider", function () {
       assert.equal(result, successResponse.result);
     });
 
-    it("should throw upon receiving a rate-limit response that lacks a retry-after header", async function () {
+    it("should retry even if the rate-limit response lacks a retry-after header", async function () {
       const mockPool = makeMockPool(url);
+      let tooManyRequestsReturned = false;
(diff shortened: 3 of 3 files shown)


Phips0812 (approved):

Phips0812:
Hi @fvictorio just reviewed the code, looks good to me. Are there any plans when this will get merged?

schaable (approved):

fvictorio:
We need to adapt some tests to this change.

alcuadrado:
Looks good

PR #3913, opened 2023-05-10
chore(ci): temporarily replace nextest with cargo test

Wodann:


.github/workflows/hardhat-core-ci.yml
```diff
@@ -267,26 +266,41 @@ jobs:
...
 
-      - name: Doctests
+      - name: Run cargo test
         uses: actions-rs/cargo@v1
         env:
           RUSTFLAGS: ${{ matrix.RUSTFLAGS }}
         with:
           command: test
-          args: --doc --workspace --all-features
+          args: --workspace --all-targets --features tracing,bench-once
```
feuGeneA, line 275: This needs the ALCHEMY_URL env var in order to run the remote tests: https://github.com/NomicFoundation/hardhat/actions/runs/4939180231/jobs/8829714657#step:5:208

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -694,15 +692,10 @@ mod tests {
...
-                error.message,
-                "invalid 1st argument: block_number value was not valid block tag or block number"
-            );
-                assert!(error.data.is_none());
+            if let RpcClientError::HttpStatus(error) = error {
+                assert_eq!(error.status(), Some(StatusCode::from_u16(400).unwrap()));
             } else {
-                unreachable!("Invalid error");
+                unreachable!("Invalid error: {error}");
             }
```
feuGeneA, line 699: It's really too bad that the error is now being reported this way. So much less information. It makes me wonder if it's a transient thing (if it might switch back to a legit RPC response at some point). And it also makes me wonder whether Infura responds "properly" (with an actual RPC response) or with an HTTP status code like this. We should make sure our stuff is compatible with (gives reasonable error messages for) Infura too. I'm eager to get this PR merged, but maybe we need a ticket to check on that in the future?

feuGeneA (changes requested):

crates/rethnet_eth/src/remote/client.rs
Wodann: I think it responds both with a `400` error and a text body. We could iterate on this in the future and provide both the status + response body?

.github/workflows/hardhat-core-ci.yml
Wodann: Good catch. I added it to the doc tests by accident instead

crates/rethnet_eth/src/remote/client.rs
Wodann: I created an issue to address this in the future: https://github.com/NomicFoundation/rethnet/issues/100

PR #3915, opened 2023-05-10
test: when there's an invalid error, show it

feuGeneA:


Wodann:
Duplicates changes in: https://github.com/NomicFoundation/hardhat/pull/3913

PR #3916, opened 2023-05-10
Merge main into rethnet/main

fvictorio:
This PR syncs `main` with `rethnet/main`. Besides the merge conflicts, some things I had to do are:

- There was an OOM in one test when `ethereumjs` mode was used. [This commit](https://github.com/NomicFoundation/hardhat/pull/3916/commits/12fa6a175a9fe2243c21f5b5a7398c53a43b41f8) fixes that. The fact that this wasn't happening with the previous version is a bit worrying. I think this is caused by a change in how the memory for each step is generated. I don't know if this is a bug or a fix in ethereumjs to be honest, because the content debug_tt traces is super underspecified. I think we can just ignore that for now.
- The stack traces test are now running with the shanghai hardfork. [This commit](https://github.com/NomicFoundation/hardhat/pull/3916/commits/a038dab28b8b59bd7620accd51fedb166e1ef6a5) sets that up for rethnet mode.
- There are some tests that fail because we don't have a way to disable the EIP-3860 max initcode size check. [This commit](https://github.com/NomicFoundation/hardhat/pull/3916/commits/d4f6c0a2aae7164681f0a5a68970b45dcdfc6379) disables those tests except for ethereumjs mode. I opened [this issue](https://github.com/NomicFoundation/rethnet/issues/103) for fixing that and re-enabling those tests. @Wodann can you apply the proper labels to it? I just set the milestone.

Even after that there are a few stack traces tests failing in dual mode. But ideally I would like to get this merged and to get that fixed in a separate PR (by me or by anyone else).


Wodann:
These failing forking tests are new: https://github.com/NomicFoundation/hardhat/actions/runs/4956863372/jobs/8867848141?pr=3916#step:9:4925

It seems like something might have changed in the `mixHash` / `prevrandao` setup?

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2342,7 +2364,7 @@ Hardhat Network's forking functionality only works with blocks from at least spu
   private async _persistIrregularWorldState(): Promise<void> {
     this._irregularStatesByBlockNumber.set(
       this.getLatestBlockNumber(),
-      await this._vm.makeSnapshot()
```
Wodann, line 2345: This breaks one of the forking tests: https://github.com/NomicFoundation/hardhat/actions/runs/4956863372/jobs/8867848141?pr=3916#step:9:5021

PR #3917, opened 2023-05-10
chore: sync pre-commit hook to CI

feuGeneA:
given that we've just made a bunch of changes to the CI test scripts, it was bothering me that the `pre-commit` git hook wasn't matching up with those recent changes.

### DIFF
--- a/.cargo-husky/hooks/pre-commit
+++ b/.cargo-husky/hooks/pre-commit
@@ -1,0 +1,11 @@
+#!/bin/sh -x
+#
+# Intended to be in sync with .github/workflows/hardhat-core-ci.yml's
+# test-rethnet-rs
+
+set -e
+
+cargo test --workspace --all-targets --features tracing,bench-once
+cargo test --doc --workspace --features tracing
+cargo clippy --all -- -D warnings
+cargo fmt --all -- --check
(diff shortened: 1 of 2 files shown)


.cargo-husky/hooks/pre-commit
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh -x
```
Wodann, line 1: Upon checkout, is this guaranteed to have executable permission?

> Give an executable permission to the files (on *nix OS).

https://stackoverflow.com/questions/40978921/how-to-add-chmod-permissions-to-file-in-git

feuGeneA: My understanding is that cargo-husky should handle that. This file isn't used directly. It's "installed" to `.git/hooks` by cargo-husky upon `cargo test` in the `rethnet` crate.

feuGeneA: Oh I see now in the docs that you're expected to do that. I didn't do that here. I'll do that.

feuGeneA: Looks like it was already done, presumably because the `+x` was already set locally when i `git add`ed it:

```$ git ls-files -s .cargo-husky/hooks/pre-commit 
100755 8bbfeb0d4328f955308e1597697a3c3d45b7d270 0	.cargo-husky/hooks/pre-commit
```

"100755, which means it’s an executable file" ([source](https://git-scm.com/book/sv/v2/Git-Internals-Git-Objects))

Wodann (approved):

PR #3918, opened 2023-05-11
rethnet/state: Support valid `code` in `AccountInfo` outputs on `StateDebug` methods

feuGeneA:


feuGeneA:
See test failures in CI run here: https://github.com/NomicFoundation/hardhat/actions/runs/4943618267/jobs/8838291925#step:5:110

Wodann:
By design, `revm` exposes two methods: `StateRef::basic` and `StateRef::code_by_hash`. The former receives basic account information: balance, nonce, and code hash. The latter is meant to lookup code by its hash, but **only** if it is needed. Most of the time we are not dealing with the code but just basic info, so it's a waste to store it alongside the basic info and keep it in memory while doing operations that don't need it.

The optional code on the on the `AccountInfo`  struct is documented: https://docs.rs/revm-primitives/latest/revm_primitives/state/struct.AccountInfo.html 

Closing this PR

PR #3921, opened 2023-05-12
test: `StateHistory` worst case benchmarks

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/78

### DIFF
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -8,19 +8,6 @@ use rethnet_eth::{Address, Bytes, U256};
 mod util;
 use util::{bench_sync_state_method, prep_no_op, STORAGE_SCALES};
 
-fn bench_checkpoint(c: &mut Criterion) {
-    bench_sync_state_method(
-        c,
-        "StateHistory::checkpoint()",
-        prep_no_op,
-        |mut state, _number_of_accounts| {
-            let result = state.checkpoint();
-            debug_assert!(result.is_ok());
-        },
-        &[0],
-    );
-}
-
 fn bench_basic(c: &mut Criterion) {
     bench_sync_state_method(
         c,
@@ -71,11 +58,5 @@ fn bench_storage(c: &mut Criterion) {
     );
 }
 
-criterion_group!(
-    benches,
-    bench_checkpoint,
-    bench_basic,
-    bench_code_by_hash,
-    bench_storage,
-);
+criterion_group!(benches, bench_basic, bench_code_by_hash, bench_storage);
 criterion_main!(benches);
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -13,14 +13,14 @@ fn bench_basic(c: &mut Criterion) {
         c,
         "StateRef:basic",
         prep_no_op,
-        |state, number_of_accounts| {
+        |state, number_of_accounts, _, _| {
             for i in (1..=number_of_accounts).rev() {
                 let result = state.basic(Address::from_str(&format!("0x{:0>40x}", i)).unwrap());
                 debug_assert!(result.is_ok());
             }
         },
         &[0],
-        &[0],
+        &[1],
     );
 }
 
@@ -29,7 +29,7 @@ fn bench_code_by_hash(c: &mut Criterion) {
         c,
         "StateRef:code_by_hash",
         prep_no_op,
-        |state, number_of_accounts| {
+        |state, number_of_accounts, _, _| {
             for i in (1..=number_of_accounts).rev() {
                 let result = state.code_by_hash(
                     Bytecode::new_raw(Bytes::copy_from_slice(
@@ -41,7 +41,7 @@ fn bench_code_by_hash(c: &mut Criterion) {
             }
         },
         &[0],
-        &[0],
+        &[1],
     );
 }
 
@@ -50,14 +50,14 @@ fn bench_storage(c: &mut Criterion) {
         c,
         "StateRef:storage",
         prep_no_op,
-        |state, number_of_accounts| {
+        |state, number_of_accounts, _, _| {
             for i in (1..=number_of_accounts).rev() {
                 let result = state.storage(Address::from_low_u64_ne(i), U256::from(i));
                 debug_assert!(result.is_ok());
             }
         },
         &STORAGE_SCALES,
-        &[0],
+        &[1],
     );
 }
 
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -11,7 +11,7 @@ use util::{bench_sync_state_method, prep_no_op, STORAGE_SCALES};
 fn bench_basic(c: &mut Criterion) {
     bench_sync_state_method(
         c,
-        "StateRef::basic()",
+        "StateRef:basic",
         prep_no_op,
         |state, number_of_accounts| {
             for i in (1..=number_of_accounts).rev() {
@@ -27,7 +27,7 @@ fn bench_basic(c: &mut Criterion) {
 fn bench_code_by_hash(c: &mut Criterion) {
     bench_sync_state_method(
         c,
-        "StateRef::code_by_hash",
+        "StateRef:code_by_hash",
         prep_no_op,
         |state, number_of_accounts| {
             for i in (1..=number_of_accounts).rev() {
@@ -48,7 +48,7 @@ fn bench_code_by_hash(c: &mut Criterion) {
 fn bench_storage(c: &mut Criterion) {
     bench_sync_state_method(
         c,
-        "StateRef::storage",
+        "StateRef:storage",
         prep_no_op,
         |state, number_of_accounts| {
             for i in (1..=number_of_accounts).rev() {
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -20,6 +20,7 @@ fn bench_basic(c: &mut Criterion) {
             }
         },
         &[0],
+        &[0],
     );
 }
 
@@ -40,6 +41,7 @@ fn bench_code_by_hash(c: &mut Criterion) {
             }
         },
         &[0],
+        &[0],
     );
 }
 
@@ -55,6 +57,7 @@ fn bench_storage(c: &mut Criterion) {
             }
         },
         &STORAGE_SCALES,
+        &[0],
     );
 }
 
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -6,7 +6,7 @@ use revm::{db::StateRef, primitives::Bytecode};
 use rethnet_eth::{Address, Bytes, U256};
 
 mod util;
-use util::{bench_sync_state_method, prep_no_op, STORAGE_SCALES};
+use util::{bench_sync_state_method, prep_no_op, Permutations};
 
 fn bench_basic(c: &mut Criterion) {
     bench_sync_state_method(
@@ -56,7 +56,7 @@ fn bench_storage(c: &mut Criterion) {
                 debug_assert!(result.is_ok());
             }
         },
-        &STORAGE_SCALES,
+        &Permutations::storage_scales(),
         &[1],
     );
 }
--- a/crates/rethnet_evm/benches/state_history.rs
+++ b/crates/rethnet_evm/benches/state_history.rs
@@ -1,7 +1,7 @@
 use criterion::{criterion_group, criterion_main, Criterion};
 
 mod util;
-use util::{bench_sync_state_method, prep_no_op};
+use util::{bench_sync_state_method, prep_no_op, SNAPSHOT_SCALES};
 
 fn bench_checkpoint(c: &mut Criterion) {
     bench_sync_state_method(
@@ -17,5 +17,39 @@ fn bench_checkpoint(c: &mut Criterion) {
     );
 }
 
-criterion_group!(state_history_benches, bench_checkpoint);
+fn bench_make_snapshot(c: &mut Criterion) {
+    bench_sync_state_method(
+        c,
+        "StateHistory::make_snapshot",
+        prep_no_op,
+        |mut state, _number_of_accounts| {
+            state.make_snapshot();
+        },
+        &[0],
+        &SNAPSHOT_SCALES,
+    );
+}
+
+fn bench_revert(c: &mut Criterion) {
+    bench_sync_state_method(
+        c,
+        "StateHistory::revert",
+        |state, _| {
+            state.checkpoint().unwrap();
+        },
+        |mut state, _number_of_accounts| {
+            let result = state.revert();
+            debug_assert!(result.is_ok());
+        },
+        &[0],
+        &SNAPSHOT_SCALES,
+    );
+}
+
+criterion_group!(
+    state_history_benches,
+    bench_checkpoint,
+    bench_make_snapshot,
+    bench_revert
+);
 criterion_main!(state_history_benches);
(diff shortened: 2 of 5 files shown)


crates/rethnet_evm/benches/util.rs
```diff
@@ -48,6 +49,9 @@ impl RethnetStates {
                 }
                 state.checkpoint().unwrap();
             }
+            for _ in 0..number_of_snapshots {
```
Wodann, line 52: Make the same snapshots over and over will not result in multiple snapshots, as the state root will be the same. You'd need to employ the same approach as is done for checkpoints: every X changes, create a snapshot

crates/rethnet_evm/benches/state_history.rs
```diff
@@ -0,0 +1,100 @@
...
+fn bench_make_snapshot(c: &mut Criterion) {
+    bench_sync_state_method(
+        c,
+        "StateHistory::make_snapshot",
+        prep_no_op,
+        |mut state, _number_of_accounts| {
+            state.make_snapshot();
+        },
+        &[0],
+        &SNAPSHOT_SCALES,
```
Wodann, line 31: The number of snapshots doesn't matter. No need to vary this for snapshot creation. The thing that would be of interest is how the performance scales depending on the other 3 variables:
- number of accounts
- number of checkpoints
- number of storage slots per account

crates/rethnet_evm/benches/state_history.rs
```diff
@@ -0,0 +1,100 @@
...
+        |state, _| {
+            snapshot
+                .borrow_mut()
+                .assign_from_slice(state.make_snapshot().as_bytes());
+        },
+        |mut state, _number_of_accounts| {
+            state.remove_snapshot(&snapshot.borrow());
+        },
+        &[0],
+        &SNAPSHOT_SCALES,
```
Wodann, line 49: Same as `bench_make_snapshot`

crates/rethnet_evm/benches/state_history.rs
```diff
@@ -0,0 +1,100 @@
...
+        "StateHistory::revert",
+        |state, _| {
+            state.checkpoint().unwrap();
+        },
+        |mut state, _number_of_accounts| {
+            let result = state.revert();
+            debug_assert!(result.is_ok());
+        },
+        &[0],
+        &SNAPSHOT_SCALES,
```
Wodann, line 88: Revert does not use snapshots, so no need to include them.

crates/rethnet_evm/benches/state_history.rs
```diff
@@ -0,0 +1,100 @@
...
+    bench_sync_state_method(
+        c,
+        "StateHistory::checkpoint()",
+        prep_no_op,
+        |mut state, _number_of_accounts| {
+            let result = state.checkpoint();
+            debug_assert!(result.is_ok());
+        },
+        &[0],
+        &[0],
```
Wodann, line 18: The number of storage slots are of interest for creating a checkpoint, as the state root is calculated

crates/rethnet_evm/benches/state_history.rs
```diff
@@ -0,0 +1,100 @@
...
+        },
+        |mut state, _number_of_accounts| {
+            state.remove_snapshot(&snapshot.borrow());
+        },
+        &[0],
+        &SNAPSHOT_SCALES,
+    );
+}
+
+fn bench_set_block_context_to_snapshot(c: &mut Criterion) {
```
Wodann, line 53:
```suggestion
fn bench_set_block_context_to_latest_snapshot(c: &mut Criterion) {
```

Wodann (changes requested):
I left several requested changes

crates/rethnet_evm/benches/util.rs
feuGeneA: see how i'm now varying the balance in the new function `prep_snapshots()`

crates/rethnet_evm/benches/state_history.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3921/commits/797e797c542cda249f9e0bac462d0649ebb27cf9

crates/rethnet_evm/benches/state_history.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3921/commits/a65aa9bcf22849445a0ef338ca9f2cc9a181afa4

crates/rethnet_evm/benches/state_history.rs
feuGeneA: I removed the snapshots part in https://github.com/NomicFoundation/hardhat/pull/3921/commits/2e7966fe3a7760ad5e8c27c63cb20ab672bac1ef, but i forgot to bring in the storage slots part, i'll circle back to that soon.

crates/rethnet_evm/benches/state_history.rs
```diff
@@ -0,0 +1,100 @@
...
+        |mut state, _number_of_accounts| {
+            let result = state.set_block_context(&snapshot.borrow(), None);
+            debug_assert!(result.is_ok());
+        },
+        &[0],
+        &SNAPSHOT_SCALES,
+    );
+}
+
+// TODO: consider whether we should have an additional benchmark of set_block_context that restores
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3921/commits/31287478e7cf20a30173970a140823e37d49a1d2

crates/rethnet_evm/benches/state_history.rs
```diff
@@ -69,6 +73,110 @@ fn bench_set_block_context_to_latest_snapshot(c: &mut Criterion) {
     );
 }
 
+fn prep_snapshots(
```
Wodann, line 76: Please correct me if I'm wrong, but what this does is:

1. for a varied number of `a`  accounts, `c` checkpoints, and `s` storage slots; fill the benchmark
  - The intend of this is such that we can see how `a`, `c`, and `s` influence the performance
2. In addition to 1, for a varied number of `p` snapshots, make additional modifications and checkpoints

This means that the parameters in a, c, and s don't actually influence the outcome of benchmarks related to snapshots anymore.

E.g. the "revert to earliest layer" benchmark should revert to the first stored layer in a database (created the first time `checkpoint` is called). Your code is currently reverting to the `c + 1`th layer.

Wodann: This is just a suggestion (the might be better solutions)!

One possible solution I see is to write:

Pseudocode:
```rust
fn benchmark(
  number_of_accounts: usize, // divisable by number_of_checkpoints without rounding
  number_of_checkpoints: usize, // divisable by number_of_snapshots without rounding
  number_of_snapshots: usize,
  number_of_storage_slots_per_account: usize,
) {
  let number_of_accounts_per_checkpoint = number_of_accounts / number_of_checkpoints;
  let number_of_checkpoints_per_snapshot = number_of_checkpoints / number_of_snapshots;
  
  for snapshot_idx in 0..number_of_snapshots {
    for checkpoint_idx in 0..number_of_checkpoints_per_snapshot {
      for account_idx in 0..number_of_accounts_per_checkpoint {
        // setup basic account
        
        for storage_slot_idx in 0..number_of_storage_slots_per_account {
          // setup storage slot
        }
      }
    }
  }
}
```

Wodann: At that point, you just need to add some logic to retrieve the state root of a particular checkpoint or snapshot

crates/rethnet_evm/benches/state.rs
```diff
@@ -11,7 +11,7 @@ use util::{bench_sync_state_method, prep_no_op, STORAGE_SCALES};
 fn bench_basic(c: &mut Criterion) {
     bench_sync_state_method(
         c,
-        "StateRef::basic()",
```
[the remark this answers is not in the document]
feuGeneA: I think we probably should keep the trait names, because it seems that the "group name" is combined with the "function name" when the benchmark is titled on the CLI and in the HTML report. This line you have highlighted here is used as the group name. Here's an example of a full title as it's printed in the CLI output: `Benchmarking StateDebug:storage_root exist acct/Layered,20 accts per chkpt,10 slots per acct,0 snapshots/2000`

crates/rethnet_evm/benches/state_history.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3921/commits/def0b581c0600ae20418a51422367108e0a9b528

crates/rethnet_evm/benches/state_history.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3921/commits/b4957daec0d5e96dc62f348a33ad0f36dc262164

crates/rethnet_evm/benches/util.rs
```diff
@@ -1,75 +1,118 @@
...
+                Box::new(|| Box::new(self.hybrid.clone())),
+                &self.hybrid_checkpoints,
+                &self.hybrid_snapshots,
+            ),
         ]
     }
 }
 
 #[cfg(feature = "bench-once")]
+#[allow(dead_code)]
```
Wodann, line 110: Why does this throw a dead code warning?

crates/rethnet_evm/benches/util.rs
```diff
@@ -86,61 +129,79 @@ mod config {
     ];
 
     pub const STORAGE_SCALES: [u64; 4] = [1, 10, 100, 1000];
+
+    pub const SNAPSHOT_SCALES: [u64; 4] = [1, 10, 100, 1000];
```
Wodann, line 133: `CHECKPOINT_SCALES` needs to be dividable without remainder by every scale of  `SNAPSHOT_SCALES` or the benchmark will end up having fluctuations in number of accounts, etc. due to rounding.

To do so,

```rust
pub const SNAPSHOT_SCALES: [u64; NUM_SCALES] = [1, 5, 10, 20];
const MAX_SNAPSHOT_SCALE: u64 = SNAPSHOT_SCALES[NUM_SCALES - 1];

pub const CHECKPOINT_SCALES: [u64; NUM_SCALES] = [
  MAX_SNAPSHOT_SCALE,
  MAX_SNAPSHOT_SCALE * 2
  MAX_SNAPSHOT_SCALE * 4,
  MAX_SNAPSHOT_SCALE * 8
];
const MAX_CHECKPOINT_SCALE: u64 = CHECKPOINT_SCALES[NUM_SCALES - 1];

pub const ADDRESS_SCALES: [u64; NUM_SCALES] = [
  MAX_CHECKPOINT_SCALE,
  MAX_CHECKPOINT_SCALE * 5,
  MAX_CHECKPOINT_SCALE * 25,
  MAX_CHECKPOINT_SCALE * 50,
];
```

What that means is that we vary:

- no. of snapshots: 1, 5, 10, 20
- no. of checkpoints: 20, 40, 80, 160
- no. of accounts: 160, 800, 4000, 8000

(no. storage is not taken into consideration as it's unrelated

crates/rethnet_evm/benches/util.rs
```diff
@@ -86,61 +129,79 @@ mod config {
...
 pub fn bench_sync_state_method<O, R, Prep>(
     c: &mut Criterion,
     method_name: &str,
     mut prep: Prep,
     mut method_invocation: R,
     storage_scales: &[u64],
+    snapshot_scales: &[u64],
 ) where
-    R: FnMut(Box<dyn SyncState<StateError>>, u64) -> O,
+    R: FnMut(Box<dyn SyncState<StateError>>, u64, &Vec<B256>, &Vec<B256>) -> O,
```
Wodann, line 147: Nice!

crates/rethnet_evm/benches/util.rs
```diff
@@ -86,61 +129,79 @@ mod config {
...
-                            format!(
-                                "{},{} account(s) per checkpoint, {} storage slots per account",
-                                label, *accounts_per_checkpoint, *storage_slots_per_account
+                    for (label, state_factory, checkpoints, snapshots) in
+                        rethnet_states.make_state_refs().into_iter()
+                    {
+                        group.bench_with_input(
+                            BenchmarkId::new(
+                                format!(
+                                    "{},{} chkpts,{} slots per acct,{} snapshots",
```
Wodann, line 169: no. of accounts?

Wodann (changes requested):
Some more questions and comments

crates/rethnet_evm/benches/util.rs
feuGeneA: it's because of the `SNAPSHOT_SCALES` member of this module, which is used in some of the benchmark binaries but not all. (each benchmark (eg `state_history` vs `state_debug`) is a separate binary and raises its own set of `dead_code` warnings/errors). i moved the `dead_code` directive, to apply specifically to `SNAPSHOT_SCALES` rather than this whole module, in https://github.com/NomicFoundation/hardhat/pull/3921/commits/776d41b48a0bb33af00592bcdb8996a7d9784a40

crates/rethnet_evm/benches/util.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3921/commits/a59897b66f25e6c4bd0e249afc1087ca84fef3e9

crates/rethnet_evm/benches/util.rs
feuGeneA: i removed "number of accounts" from the "function name" because "number of accounts" is the value we're using as the "input" to that function, so that value is automatically appended to every permutation of the "benchmark name"

fvictorio (approved):

crates/rethnet_evm/benches/util.rs
```diff
@@ -1,146 +1,246 @@
...
+            for snapshot_scale in SNAPSHOT_SCALES {
+                assert!(checkpoint_scale % snapshot_scale == 0, "all checkpoint scales must be evenly divisible by all snapshots scales, but checkpoint scale {checkpoint_scale} is not evenly divisible by snapshot scale {snapshot_scale}");
+            }
+        }
+    }
+}
+
+impl Default for RethnetStates {
+    fn default() -> Self {
+        Self::assert_scale_divisibility();
```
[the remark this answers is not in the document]
feuGeneA: Yes, I did do this in a weird way, so thank you for pushing back on it.

What I had wanted to do was a file level assert, just after the declarations, but Rust doesn't support that without a 3rd party crate. I chose to do it here rather than in `fill()` because this struct could in theory be used by some other benchmark without that benchmark using `fill()`. But I agree this isn't ideal. I took your idea and then took it a little further, using a `struct` and `impl` rather than the `mod`. Check it out in https://github.com/NomicFoundation/hardhat/pull/3921/commits/9d1de6d50964a9a650fdb5fdd63f3dc12ce2d7e5

crates/rethnet_evm/benches/state_history.rs
```diff
@@ -0,0 +1,141 @@
...
+use util::{bench_sync_state_method, prep_no_op, SNAPSHOT_SCALES, STORAGE_SCALES};
+
+fn bench_checkpoint(c: &mut Criterion) {
+    bench_sync_state_method(
+        c,
+        "StateHistory:checkpoint()",
+        prep_no_op,
+        |mut state, _number_of_accounts, _, _| {
+            let result = state.checkpoint();
+            debug_assert!(result.is_ok());
```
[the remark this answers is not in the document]
feuGeneA: Yes, our CI and git pre-commit hook are both configured to run `cargo test --workspace --all-targets --features tracing,bench-once`. That command, due I believe to `--all-targets`, also runs the benchmarks as a test, and the `bench-once` bit ensures that we just run through a single permutation of the benchmarks, so that it can finish quickly.

PR #3924, opened 2023-05-12
Add support for ethers v6

fvictorio:
This PR adds support for ethers v6 in `hardhat-ethers` and `hardhat-chai-matchers`. I'm going to give a high-level explanation of the changes here, but I'll also add some inline explanatory comments.

## hardhat-ethers

The main change here is that we now have our own classes fully implementing the (new) `Provider` and `Signer` interfaces. This gives us more control and avoids things that ethers does by default that don't go well with a devnet, like certain caches or some long polling intervals.

Some of the changes here also come from the fact that `ContractFactory` has type params, and I added those to some of our helpers.

## hardhat-chai-matchers

The main change here is related to dropping support for `BigNumber`s and using bigints in our implementation.

There are also two fixes here for a couple of problems that are not strictly related to the ethers upgrade but that for some reason surfacted during this work.

The first problem is this:

```js
await expect(something).to.emit(...).withArgs(/* invalid params */)
```

If there is a validation issue in the _sync_ part of `withArgs`, then we shouldn't do anything in the _async_ part of `.to.emit`. Otherwise there could be issues if the (async) assertion fails. To do this, we set an `ASSERTION_ABORTED` flag when `.withArgs` fails, check for that in the async handlers of `.to.emit` (and `.revertedWithCustomError`, the other assertion that supports `withArgs`). and just return if the assertion was aborted.

The second problem is similar:

```js
await expect(txThatRejects).to.changeEtherBalances(/* invalid params */)
```

Here again we have a sync validation that will throw an error. But if that happens, then `txThatRejects` will be an unhandled rejection. To fix this, you'll see some `Promise.resolve(this._obj).catch(() => {});` that consume the promise and handle a possible rejection.


packages/hardhat-ethers/src/internal/ethers-utils.ts
```diff
@@ -0,0 +1,459 @@
+// these helpers functions were copied verbatim from ethers
```
fvictorio, line 1: Functions copied from ethers's codebase, and slightly adapted to comply with our linter. Is this comment enough? Probably not.

packages/hardhat-ethers/src/dist/src/signer-with-address.ts
```diff
@@ -1 +1 @@
-export { SignerWithAddress } from "../../signers";
+export { CustomEthersSigner as SignerWithAddress } from "../../signers";
```
fvictorio, line 1: Two questions here:

1. I don't like the `CustomEthersProvider` and `CustomEthersSigner` name. What do you think about `HardhatEthersProvider` and `HardhatEthersSigner`?
2. Should we keep exporting `SignerWithAddress` as an alias for backwards compatibility or should we drop it?

PR #3932, opened 2023-05-15
Add support for extending providers

nicosantangelo:

---

# Extend providers

This is a first implementation to allow for the provider to be wrapped and extended with extra functionality.
The DSL mimics the way DSL `extendConfig` and `extendEnvironment` work, but it allows for an `async` wrapping to be made.

Hardhat will "compile" the provider into `hre.network.provider` from the combination of all extender functions it finds, in order.
The API looks like this:

```ts
extendProvider(async (provider: EIP1193Provider) => {
  const wrappedProvider = new ExtendedProvider(provider);
  await wrappedProvider.doSomething();
  return wrappedProvider;
});
```

### Considerations

The new composed provider will run after the first call to `request` is made, unlike before, where it was built when first accessed (`lazyObject`). Because of this it was (currently) decided that any call to a sync method, i.e. EventEmitter methods, will throw informing the user that she needs to call `request` first.
This should be documented.

# ⚠️  Problem with hardhat-ethers

Currently, we have a problem with `hardhat-ethers` regarding this new implementation. `hardhat-ethers` extends the environment and hooks two events to the provider by calling [`createProviderProxy`](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-ethers/src/internal/index.ts#L36), [here](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-ethers/src/internal/provider-proxy.ts#L25).
The provider will now be a `LazyInitializationProvider` which causes that line [to throw](https://github.com/NomicFoundation/hardhat/pull/3932/files#diff-582b4cc2358c0e8a40fd6a7e8e90234ceb85d7881709a4a8d4cf6719d6c532e3R53) because `.request()` hasn't been called yet

It's a bit tricky to fix, the options I have in mind are:
- Having an implementation that doesn't throw on EventEmitter calls if `.request()` hasn't been called. This will work but doing a full implementation might be tricky. For example, should we support `setMaxListeners`? and should we delete listeners when `.off()` is called? Maybe we could have a `private EventEmitter` and then just move the listeners by looping though `Object.keys(privateEmitter.eventNames())`
- Expose a public async `.init()` method on our providers that can be called to force what `.request()` does underneath. This will make either `EthereumProvider` or `EIP1193Provider` muddy in my opinion
- Use an async `extendProvider` method to extend the ethers provider instead of doing it on `extendEnvironment`. This sounds good on paper but it poses two problems. First what method to call to initialize the provider, second that hre.ethers is a `lazyObject` and this would initialize it immediately

I think the first option is the best one but I'm not completely sure (and I might be missing some other easier version!), I've implemented it locally already and seems to work well

# TODO
- [x] Automated tests
- [x] Write documentation
- [x] Consider making `ProviderWrapper` part of the public API. Maybe best for a different PR?


packages/hardhat-core/src/internal/core/providers/lazy-initialization.ts
```diff
@@ -0,0 +1,145 @@
...
+} from "../../../types";
+import { CustomError } from "../errors";
+
+/**
+ * A class that delays the (async) creation of its internal provider until the first call
+ * to a JSON RPC method via request/send/sendAsync.
+ * Trying to use the EventEmitter API without calling request first (initializing the provider)
+ * will throw.
+ */
+export class LazyInitializationProvider implements EthereumProvider {
```
nicosantangelo, line 16: This class re-implements the methods from `EventEmitter` like `EventEmitterWrapper` does. This could be done automagically by maybe looping through `Object.keys(EventEmitter.prototype)`.

I prefer this more explicit solution but we can discuss it further!

packages/hardhat-core/src/internal/core/providers/lazy-initialization.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  private async ensureProvider(): Promise<EthereumProvider> {
+    if (!this.provider) {
+      this.provider = await this.providerFactory();
+    }
+    return this.provider;
+  }
+}
+
+export class UninitializedProviderError extends CustomError {
```
nicosantangelo, line 140: I'm not entirely sure this is the correct place for this error! I searched around but adding a new HardHat Error seemed like too much.

Let me know if it should be done differently!

fvictorio:
> Use an async extendProvider method to extend the ethers provider instead of doing it on extendEnvironment. This sounds good on paper but it poses two problems. First what method to call to initialize the provider, second that hre.ethers is a lazyObject and this would initialize it immediately

I think we can rule out this option because the provider created in `hardhat-ethers` is in `hre.ethers.provider`, but `extendProvider` will affect `hre.network.provider` (the provider that is always available, no matter which plugins you are using).

> Expose a public async .init() method on our providers that can be called to force what .request() does underneath. This will make either EthereumProvider or EIP1193Provider muddy in my opinion

Agree that it's not worth it, especially because, worst case, you can just use a dummy `request` to trigger that.

> Having an implementation that doesn't throw on EventEmitter calls if .request() hasn't been called. This will work but doing a full implementation might be tricky.

What if we don't do a full implementation, and only implement the minimal things we need to prevent breaking `hardhat-ethers`? I know it doesn't feel great, but it seems like the most reasonable option. This would mean only implementing the `.on` callback. I would even be fine with a super restrictive implementation (e.g., a given event can only be registered once, if you register a second listener it throws).

packages/hardhat-core/src/internal/core/providers/lazy-initialization.ts
fvictorio: I also much prefer the explicit solution

packages/hardhat-core/src/internal/core/providers/lazy-initialization.ts
fvictorio: I added a new error for this, thanks for pointing this out

packages/hardhat-core/src/types/provider.ts
```diff
@@ -56,3 +56,5 @@ export interface EthereumProvider extends EIP1193Provider {
     callback: (error: any, response: JsonRpcResponse) => void
   ): void;
 }
+
+export type ProviderFactory = () => Promise<EthereumProvider>;
```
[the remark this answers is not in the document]
nicosantangelo: Hmm, I think that even thought this looks like an innocuous enough type, there's no real reason to make it public. Maybe if the plan was to make `LazyInitializationProvider` a core functionality (and to make it public) it'd make more sense.

Would it make sense to move the type to `src/internal/core/providers/lazy-initialization.ts`? where `LazyInitializationProvider` lives

nicosantangelo:
> What if we don't do a full implementation, and only implement the minimal things we need to prevent breaking hardhat-ethers? I know it doesn't feel great, but it seems like the most reasonable option. This would mean only implementing the .on callback. I would even be fine with a super restrictive implementation (e.g., a given event can only be registered once, if you register a second listener it throws).
> Also, I think that's the only option that doesn't need a change in hardhat-ethers. Otherwise, people upgrading Hardhat but not the plugin would get an error.

Agree! the second point most of all I think makes it the better option by far.
What do you think of [this implementation](https://github.com/NomicFoundation/hardhat/pull/3932/commits/f245b7c978794585491f5fbeab27d88802bf83b7). It doesn't need the error we created earlier and allows for all EventEmitter methods to be called without initialization moving them over from the `_emmiter` to the provider after `_initProvider`

Now I'm writing this, releasing some memory and making `_emitter` undefined after might be useful but a bit tricky in terms of typings. If you think it's a good idea we can do that too!

packages/hardhat-core/src/types/provider.ts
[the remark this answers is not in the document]
fvictorio: Yes, that makes sense. I'm also fine with just inlining this type, since it's short enough. Whatever you prefer!

fvictorio:
> What do you think of [this implementation](https://github.com/NomicFoundation/hardhat/pull/3932/commits/f245b7c978794585491f5fbeab27d88802bf83b7).

I like it. Just a minor comment: we try to avoid using non-null assertinos (the `!` operator). But I think that's easily fixable if `this._initProvider` returns the provider (maybe we should rename it to `this._getOrInitProvider` or something like that).

fvictorio:
> Now I'm writing this, releasing some memory and making `_emitter` undefined after might be useful but a bit tricky in terms of typings. If you think it's a good idea we can do that too!

I'm not sure tbh. Since you are removing all listeners, I think keeping it it's fine. There's only one instance of the provider per execution so memory-wise this doesn't seem problematic.

nicosantangelo:
> I like it. Just a minor comment: we try to avoid using non-null assertinos (the ! operator). But I think that's easily fixable if this._initProvider returns the provider (maybe we should rename it to this._getOrInitProvider or something like that).

Perfect!  I didn't like having "to know" that `this._initProvider` had side-effects either, it looked a bit hard to guess. I'll change it

packages/hardhat-core/src/internal/core/providers/lazy-initialization.ts
```diff
@@ -0,0 +1,180 @@
...
+import { HardhatError } from "../errors";
+import { ERRORS } from "../errors-list";
+
+export type ProviderFactory = () => Promise<EthereumProvider>;
+export type Listener = (...args: any[]) => void;
+
+/**
+ * A class that delays the (async) creation of its internal provider until the first call
+ * to a JSON RPC method via request/send/sendAsync.
+ * Trying to use the EventEmitter API without calling request first (initializing the provider)
```
fvictorio, line 17: I think this line is out of date now, isn't it?

fvictorio:
Closing in favor of #4008, which includes these changes.

PR #3949, opened 2023-05-23
Gelato Web3 Functions hardhat plugin 

donoso-eth:

---

Gelato Web3 Function hardhat plugin link and text correction


donoso-eth:
The plugin has a broken link. This PR corrected the broken link, yet the check returns:
```
No changeset found
The PR is not labeled as "no changeset needed"
```
Should I do something different?
Thks for helping!

PR #3953, opened 2023-05-24
`DatabaseCommit` worst case benchmarks

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/80

The benchmark can be run with `ALCHEMY_URL=... cargo bench --bench state --features test-remote`. Though, that will run all of the benchmarks in the `state` module, which is more than what's new here. To run ONLY what's new here, first make this edit:
```diff
--- a/crates/rethnet_evm/benches/state.rs
+++ b/crates/rethnet_evm/benches/state.rs
@@ -152,9 +152,11 @@ fn bench_storage(c: &mut Criterion) {
 
 criterion_group!(
     benches,
+    /*
     bench_basic,
     bench_code_by_hash,
     bench_storage,
+    */
     bench_database_commit
 );
 criterion_main!(benches);
```

(In theory you should be able to pass a "FILTER" on the command line to `Skip benchmarks whose names do not contain FILTER`, instead of hacking this edit into place, and while I think that may work, it seems to take a very long time for the executable to start up and run the requested benchmark, so making the edit is more expedient. That long startup time, and in general a long run time for the entire set of benchmarks, is something we've noticed before, and we need to address it, but that's outside the scope of this PR.)


packages/hardhat-core/scripts/test-run-forked-block.ts
```diff
@@ -30,7 +30,7 @@ async function main(rpcUrl: string, blockNumber: bigint) {
   const remoteCommon = new Common({ chain: chainId });
   const hardfork = remoteCommon.getHardforkByBlockNumber(blockNumber);
```
fvictorio, line 31: I think we can remove this line (yes, I know, the linter should run in this file, but that turns out to be surprisingly hard :disappointed:)

crates/rethnet_eth/src/remote.rs
```diff
@@ -69,6 +69,18 @@ where
     s.serialize_str(&result)
 }
 
+/// for use with serde's deserialize_with on fields of hexadecimal strings that should be
+/// parsed as Option<u64>
+pub fn optional_u64_from_hex<'de, D>(deserializer: D) -> Result<Option<u64>, D::Error>
```
fvictorio, line 74: Is it common to write code under the `src` directory that's only used in test code?

I'm pretty sure this isn't a big deal because this will not be part of the binary (unlike in javascript, where the code will probably end up in the user's `node_modules` if you write it under `src`). Just curious if it's something that you guys are already doing elsewhere.

crates/rethnet_evm/benches/state.rs
```diff
@@ -61,5 +155,13 @@ fn bench_storage(c: &mut Criterion) {
...
 }
 
-criterion_group!(benches, bench_basic, bench_code_by_hash, bench_storage);
+criterion_group!(
+    benches,
+    /*
+    bench_basic,
+    bench_code_by_hash,
+    bench_storage,
+    */
```
fvictorio, line 164: I think you committed this accidentally?

crates/rethnet_evm/benches/state.rs
```diff
@@ -45,6 +45,100 @@ fn bench_code_by_hash(c: &mut Criterion) {
...
+
+    bench_sync_state_method(
+        c,
+        "DatabaseCommit:commit",
+        prep_no_op,
+        |state, _number_of_accounts, _, _| {
+            #[allow(unused_variables)]
+            let mut accounts_to_commit: HashMap<Address, Account> = HashMap::new();
+            let json_accounts: HashMap<Address, AccountState> = serde_json::from_str(
+                &std::fs::read_to_string("benches/accounts_changed_in_mainnet_block_17295357.json")
```
fvictorio, line 74: Perhaps we should use a directory for this, something like `benches/fixtures`?

crates/rethnet_evm/benches/state.rs
```diff
@@ -45,6 +45,100 @@ fn bench_code_by_hash(c: &mut Criterion) {
...
+                        storage,
+                        storage_cleared: account_state.storage_cleared,
+                        is_destroyed: false,
+                        is_touched: false,
+                        is_not_existing: false,
+                    },
+                );
+            }
+
+            state.commit(accounts_to_commit);
```
fvictorio, line 118: I'm a bit confused about the fact that the whole function is being benchmarked. Shouldn't only this line be? Or, at the very least, I think the whole deserializing of the fixture shouldn't be included in the benchmark, and I'm not sure about the insertion part.

crates/rethnet_eth/src/remote.rs
fvictorio: I realized after writing this that you just moved an already existing function. I'm still curious about this, but there's no need to do anything about it.

packages/hardhat-core/scripts/test-run-forked-block.ts
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3953/commits/724b96c4e94c1e2d227bac7bd42d5b46972fe8a8

crates/rethnet_eth/src/remote.rs
feuGeneA: It's a good question, and I'm not sure what the answer is.

crates/rethnet_evm/benches/state.rs
feuGeneA: Yes! Thank you! Addressed in https://github.com/NomicFoundation/hardhat/pull/3953/commits/357a0f3f0199c50161fa8a629f995dfadcc712e7

crates/rethnet_evm/benches/state.rs
feuGeneA: I like that idea, thank you for suggesting it, done in https://github.com/NomicFoundation/hardhat/pull/3953/commits/250d3d794e7ad656df672f14ad9c56c00b89b5b1

crates/rethnet_evm/benches/state.rs
feuGeneA: You're right. I think I was just taking Remco's spec too literally (see the issue this PR fixes). In the other benchmark issues he had used a "Setup" section separate from "Measurement", so I guess I just assumed everything under "Benchmark" was the measurement target, but I should have thought about it more. Addressed in https://github.com/NomicFoundation/hardhat/pull/3953/commits/5f2b419cf561933d6e05af2de66a76a32f0e34df

fvictorio (approved):

crates/rethnet_evm/benches/state.rs
```diff
@@ -45,11 +45,106 @@ fn bench_code_by_hash(c: &mut Criterion) {
     );
 }
 
+fn bench_database_commit(c: &mut Criterion) {
```
Wodann, line 48: Given that all 3 other traits have their own file (state, state_debug, and state_history), I feel like it would make sense to also have a db_commit file

crates/rethnet_evm/benches/state.rs
```diff
@@ -45,11 +45,106 @@ fn bench_code_by_hash(c: &mut Criterion) {
...
+    }
+
+    bench_sync_state_method(
+        c,
+        "DatabaseCommit:commit",
+        state_prep_no_op,
+        |state, _number_of_accounts, _, _| {
+            state.commit(accounts_to_commit.clone());
+
+            debug_assert!(json_accounts.iter().all(|(address, json)| {
```
Wodann, line 121: Use `.for_each` if you want to execute something for each element in the iterator

crates/rethnet_evm/benches/state.rs
```diff
@@ -45,11 +45,106 @@ fn bench_code_by_hash(c: &mut Criterion) {
...
+            Account {
+                info: AccountInfo {
+                    balance: account_state.balance.unwrap(),
+                    nonce: account_state.nonce.unwrap(),
+                    code: code.clone(),
+                    code_hash: code.map_or(KECCAK_EMPTY, |code| code.hash()),
+                },
+                storage,
+                storage_cleared: account_state.storage_cleared,
+                is_destroyed: false,
```
Wodann, line 107: Does your script also generate deletions of accounts (i.e. self-destruct)?

crates/rethnet_evm/benches/state.rs
feuGeneA: I did it this way because i wanted this whole statement to be optimized out if the build doesn't include `debug_assert`s, and i was suspicious of choosing a profile (`dev`, `test`, etc), because i didn't want to depend on any particular profile continuing to have `debug_assert`s enabled, but i just discovered that i can do `#[cfg(debug_assert)]`, so i did that and used `for_each` in https://github.com/NomicFoundation/hardhat/pull/4037/commits/49ed70c5d167fa20ff47dc41a7377c250136d242

crates/rethnet_evm/benches/state.rs
feuGeneA: the script just captures whatever transactions happened in a particular block. i haven't delved into what types of transactions are present in the block i picked. (i just picked an arbitrary recent one.) do you think it's important to have self-destruct transactions? if so, we could dig a bit to find such a block and use that one as the fixture instead of what's there now.

Wodann: The idea of this benchmark is for the `fn commit` is to handle a variety of changes that matches real-world use cases, hence why we capture and replay the block. An ideal block would be one that has some of all cases:
- self-destruct account
- insert account
- modify accounts
- insert storage slots
- modify storage slots
- remove storage slots

Naturally, it'll be hard to find a block that matches all of these, so potentially replaying a sequence of blocks would do the trick?

Were there any qualifying characteristics for the block you chose?

feuGeneA: No, there were no qualifying characteristics for the block I chose. I asked the team [here](https://nomicfoundation.slack.com/archives/CTTF55XPT/p1687302433214939)

Wodann: Given that the expectation is that it'll be tricky to find one block to rule them all, I'd go with the approach of selecting a range of blocks to replay. At least statistically you're more likely to get a little of everything.

crates/rethnet_eth/src/remote.rs
Wodann: Often you'll create a separate module (that is never published to crates.io) which contains test helpers. You can include that as a dev-dependency in your Cargo.toml

PR #3959, opened 2023-05-29
[ADD] Support for fuse and fuse spark

MayankMittal1:
## Description
Add Support for fuse and fuse spark chain for hardhat-verify


schaable:
Hi @MayankMittal1. In general, we are not adding new networks to `hardhat-verify`. 

Adding support for a new network is simple, check [here](https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify#adding-support-for-other-networks). This can then be easily packaged in a plugin.

PR #3965, opened 2023-05-31
Make console.sol assembly "memory-safe"

ZumZoom:
Add "memory-safe" annotation to assembly code, to allow console.sol usage in stack-too-deep prone contracts.

Notes on backwards compatibility:
- constants support in inline assembly was added in 0.5.11 hence the need of local var `consoleAddress`
- assembly flag "memory-safe" was added in 0.8.13 hence the need of Natspec tag


fvictorio:
Hey @ZumZoom, thanks a lot for this! The `console.sol` library is auto-generated using [this script](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/scripts/console-library-generator.js) though. Would you like to implement this change there? If not, I can do it for you.

ZumZoom:
Uh-oh. Didn’t notice that it is autogenerated. Will do 👌

fvictorio:
Hey @ZumZoom, do you have an example of a contract that doesn't compile with the current version of `console.sol` but it does with this version? I can probably devise one, but if you have one at hand, it would be very helpful.

ZumZoom:
The issue arose in this code:
https://github.com/1inch/limit-order-settlement/commit/f3a6ce864f85fc0e809ac5da57dd706a9957fa94

packages/hardhat-core/scripts/console-library-generator.js
```diff
@@ -81,10 +80,7 @@ for (let i = 0; i < singleTypes.length; i++) {
     typeAliasedInt.charAt(0).toUpperCase() + typeAliasedInt.slice(1);
 
   const sigInt = eutil.bufferToInt(
-    keccak256(eutil.bufArrToArr(Buffer.from("log" + "(" + type + ")"))).slice(
```
fvictorio, line 84: Thanks for cleaning up this! :slightly_smiling_face:

fvictorio:
Hey @ZumZoom, sorry for not merging this sooner. I wanted to write a test for this (something that compiles with the change but it doesn't without it), but it's pretty involved, and it depends on the version of solc so... I'm going to merge this as is.

Thanks!

ZumZoom:
I can try to come up with the small test for that later today.

PR #3967, opened 2023-06-01
Rethnet: Implement `eth_` requests for JSON-RPC server

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/91


crates/rethnet_eth/src/remote.rs
```diff
@@ -37,6 +39,15 @@ where
     seq.end()
 }
 
+fn sequence_to_single<'de, D, T>(deserializer: D) -> Result<T, D::Error>
```
fvictorio, line 42: I'm a bit confused about `single_to_sequence` and `sequence_to_single`. Why are they used in some methods and not in others? I can't see a pattern.

crates/rethnet_eth/src/remote.rs
```diff
@@ -97,72 +110,553 @@ impl BlockSpec {
...
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+struct TransactionInput {
+    from: Option<Address>,
+    to: Option<Address>,
+    gas: Option<U256>,
+    #[serde(rename = "gasPrice")]
+    gas_price: Option<U256>,
+    value: Option<U256>,
+    data: Option<ZeroXPrefixedBytes>,
+}
```
fvictorio, line 177: This is not complete (e.g., EIP-1559 is not supported). But I feel that it's better to have some basic things in place and implement the JSON-RPC server, because that will make it easier to re-use Hardhat tests and properly polish this.

crates/rethnet_eth/src/remote.rs
```diff
@@ -97,72 +110,553 @@ impl BlockSpec {
...
     }
 }
 
-#[derive(serde::Serialize)]
+impl serde::Serialize for ZeroXPrefixedBytes {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        serializer.serialize_str(&format!("0x{}", hex::encode(self.inner.clone()),))
```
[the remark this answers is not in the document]
feuGeneA: Thanks for noticing this. I added the `.clone()` because without it I got error message `cannot move out of `self.inner` which is behind a shared reference`...````move occurs because `self.inner` has type `bytes::Bytes`, which does not implement the `Copy` trait````, and I'm used to reaching for `Clone` when there's no `Copy`, but I didn't realize that a reference would do just fine here. Thanks! Addressed in https://github.com/NomicFoundation/hardhat/pull/3967/commits/0cbf7cc858f83806e7700e0cfb515db05bb5bc5a

crates/rethnet_eth/src/remote.rs
```diff
@@ -82,9 +93,11 @@ where
 }
 
 /// For specifying a block
-#[derive(Clone)]
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+#[serde(untagged)]
 pub enum BlockSpec {
     /// as a block number
+    #[serde(serialize_with = "serialize_u256")]
```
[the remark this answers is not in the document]
feuGeneA: we implement a custom serializer/deserializer for data types that we define. we use *`_with` for data types that we don't deine but that we want to handle in a special way (eg serializing an integer to hex, with the special handling of leading zeroes, a `0x` prefix, etc).

crates/rethnet_eth/src/remote.rs
```diff
@@ -97,72 +110,553 @@ impl BlockSpec {
...
+
+            fn expecting(&self, formatter: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+                formatter.write_str("a 0x-prefixed string of hex digits")
+            }
+
+            fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                if &value[0..1] == "0x" {
```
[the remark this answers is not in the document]
feuGeneA: Thanks for noticing this. Yes, we do have tests of the serde logic in general (see all the tests i added in this PR), but we were not testing this particular branch. I added a test of this case in [this commit](https://github.com/NomicFoundation/hardhat/pull/3967/commits/f7953e759c1b1eb4d41f7475d15ca9d054eef2f4), and you can see the failure result for that commit in CI [here](https://github.com/NomicFoundation/hardhat/actions/runs/5180437673/jobs/9334716497#step:5:173), and you can see the "fix" [here](https://github.com/NomicFoundation/hardhat/pull/3967/commits/2cbf6f1aa2a0d77642332be347b954751b2b7b5e). Interestingly, note that the problem wasn't just the inverted boolean, but also the range operation: the way it was written before was exclusive of the ending range specifier; if it were *just* the inverted boolean, it would have been failing on all of the "working" test cases haha.

crates/rethnet_eth/src/remote.rs
```diff
@@ -97,72 +110,553 @@ impl BlockSpec {
...
+                    Ok(Bytes::from(
+                        hex::decode(&value[2..])
+                            .unwrap_or_else(|_| panic!("failed to decode hex string \"{value}\"")),
+                    )
+                    .into())
+                }
+            }
         }
+
+        deserializer.deserialize_identifier(ZeroXPrefixedBytesVisitor)
```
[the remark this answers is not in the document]
feuGeneA: counterintuitively, the function name here relates to the target of the deserialization, not the source, and in this case we want to deserialize the data as to the identifier `ZeroXPrefixedBytes` (which is the `Value` type associated with the `*Visitor` type argument here).

crates/rethnet_eth/src/remote.rs
```diff
@@ -97,72 +110,553 @@ impl BlockSpec {
...
+                E: serde::de::Error,
+            {
+                if &value[0..1] == "0x" {
+                    Err(serde::de::Error::custom(
+                        "string does not have a '0x' prefix",
+                    ))
+                } else {
+                    Ok(Bytes::from(
+                        hex::decode(&value[2..])
+                            .unwrap_or_else(|_| panic!("failed to decode hex string \"{value}\"")),
```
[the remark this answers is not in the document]
feuGeneA: yes. I originally wrote it the way you said here, but `clippy` made me to do this instead.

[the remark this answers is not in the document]
feuGeneA: If i remember correctly, the explanation for the reason is that, with the way you and i were inclined to write it, the `format!()` call would have always been executed, regardless of whether the expect call actually fails or not.

crates/rethnet_eth/src/remote.rs
feuGeneA: Thanks for pointing this out. The pattern is _supposed to be_ that we use these helpers for every method that accepts only a single parameter, so that that parameter will be encoded/decoded as an array in the JSON. But I wasn't applying the pattern consistently, and there was an underlying bug/problem here because of that. I updated the tests to verify that there's always a sequence, in [this commit](https://github.com/NomicFoundation/hardhat/pull/3967/commits/7832956e8c2c669fb98069fe792c1a74089a9ab9), and you can see in the CI run for that commit [here](https://github.com/NomicFoundation/hardhat/actions/runs/5179422686/jobs/9332394337#step:5:185) that this test fails with the code as-is, and you can see how i fixed it (made the tests pass) in [this commit](https://github.com/NomicFoundation/hardhat/pull/3967/commits/7a4c1432507f26b0c01dc900040f95de4478147f).

crates/rethnet_eth/src/remote.rs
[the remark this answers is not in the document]
fvictorio: Oh good catch, I missed that

crates/rethnet_eth/src/remote.rs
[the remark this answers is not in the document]
fvictorio: Ohhh right! That makes sense, thanks.

crates/rethnet_eth/src/remote.rs
fvictorio: I still don't quite understand why the single parameter is a special case. All JSON-RPC parameters are inside an array, so having 0, 1, or more shouldn't be too different IMO.

For example, given these two methods:

```rust
    GetTransactionByHash(B256),
    GetTransactionCount(Address, BlockSpec),
```

In the first case the params are `["0x123..."]` and in the second case the params are something like `["0x123...", "latest"]`. In both cases you have an array in the JSON-RPC side, and in both cases you have an enum variant in the rust side. So I don't quite see why the single parameter is a special case. (Not saying that this is wrong, I'm probably missing something here, perhaps related to some serde's automagic behavior, but I'd like to understand it better.)

feuGeneA: It's because, by default, `serde` will serialize to an array only if it finds a `Vec`; and, if you tell it to deserialize to a data structure that isn't a `Vec`, and then it finds an array in the JSON, it will throw an error saying it didn't get what it expected. `single_to_sequence` makes `serde` serialize to a JSON array even though the source data type is _not_ a `Vec`. And `sequence_to_single` makes `serde` deserialize from an array even though our `enum` variant only told it to expect a single element.

feuGeneA: Also, for our `enum` variants that have multiple parameters (eg `GetTransactionCount(Address, BlockSpec)`), serde is automatically treating that list of parameters as a "sequence" to be encoded/decoded as an array in the JSON.

feuGeneA: > It's because, by default, `serde` will serialize to an array only if it finds a `Vec`; and, if you tell it to deserialize to a data structure that isn't a `Vec`, and then it finds an array in the JSON, it will throw an error saying it didn't get what it expected. `single_to_sequence` makes `serde` serialize to a JSON array even though the source data type is _not_ a `Vec`. And `sequence_to_single` makes `serde` deserialize from an array even though our `enum` variant only told it to expect a single element.

I guess I too strongly worded this. Everywhere I said `Vec` I guess I should have said "sequence", which could be a `Vec`, or alternatively could be a tuple, which covers the case of our enum variants with multiple parameters. (And maybe there are other forms of "sequence" as well? Not sure).

crates/rethnet_eth/src/remote.rs
```diff
@@ -97,72 +110,611 @@ impl BlockSpec {
...
+        let call_decoded: MethodInvocation = serde_json::from_str(&json).expect(&format!(
+            "should have successfully deserialized json {json}"
+        ));
+        assert_eq!(call, call_decoded);
+    }
+
+    #[test]
+    #[should_panic(expected = "string \\\"deadbeef\\\" does not have a '0x' prefix")]
+    fn test_zero_x_prefixed_bytes_deserialization_without_0x_prefix() {
+        serde_json::from_str::<ZeroXPrefixedBytes>("\"deadbeef\"").unwrap();
```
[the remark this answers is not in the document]
feuGeneA: https://github.com/NomicFoundation/hardhat/pull/3967/commits/273fd9e3e044c975f05ca8c194c37c64ab1a4e68

crates/rethnet_eth/src/remote.rs
fvictorio: Ah, ok, that makes sense. Especially this:

> Also, for our enum variants that have multiple parameters (eg GetTransactionCount(Address, BlockSpec)), serde is automatically treating that list of parameters as a "sequence" to be encoded/decoded as an array in the JSON.

So variants with a single value _are_ special (for serde).

Thanks for the explanation!

crates/rethnet_eth/src/remote.rs
```diff
@@ -97,72 +110,617 @@ impl BlockSpec {
...
+        let json = serde_json::json!(call).to_string();
+
+        #[derive(Debug, serde::Deserialize)]
+        struct MethodInvocationWithUntypedParams {
+            #[allow(dead_code)]
+            method: String,
+            #[allow(dead_code)]
+            params: Vec<serde_json::Value>,
+        }
+        serde_json::from_str::<MethodInvocationWithUntypedParams>(&json).expect(
```
Wodann, line 360: What does this validate?

crates/rethnet_eth/src/remote/eth.rs
```diff
@@ -175,3 +175,43 @@ where
     let opt = Option::deserialize(deserializer)?;
     Ok(opt.unwrap_or_default())
 }
+
+pub mod eip712 {
```
Wodann, line 179: Is this code ever used? If so, consider moving it into a separate file in `src/remote/eth/`

feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/1f5203b6358c13dc87969761d5db201345b4d639

crates/rethnet_eth/src/remote.rs
feuGeneA: i wrote a comment to explain in [this commit](https://github.com/NomicFoundation/hardhat/pull/3998/commits/4e29b86397af028139e8ce9ed37f2b33a30cff05). let me know what you think.

PR #3996, opened 2023-06-06
Rename error class & avoid compiling

### LINKED ISSUE #3969: hardhat-verify pending things
These are the things we didn't get to do for v1.0.0:

- Don't compile under any circumstance. This also means dropping the `--no-compile` flag, which is a breaking change that should've been included in the first version. We are going to accept this and do it as 1.1.0. The way this works is that we try to get the bytecode that matches the deployed bytecode and, if we can't find it, we instruct the user to compile the project (maybe with `--force`? the exact error message is TBD).
- Rename `HardhatEtherscanError` to `HardhatVerifyError`

schaable:
Changes:
 - Update the error class name to match the plugin's name.
 - Do not compile when running the task and remove the `--no-compile` flag. Update the tests accordingly.
 
 Closes #3969

### DIFF
--- a/packages/hardhat-verify/src/errors.ts
+++ b/packages/hardhat-verify/src/errors.ts
@@ -6,13 +6,13 @@ import {
 } from "./abi-validation-extras";
 import { TASK_VERIFY_VERIFY } from "./task-names";
 
-export class HardhatEtherscanError extends NomicLabsHardhatPluginError {
+export class HardhatVerifyError extends NomicLabsHardhatPluginError {
   constructor(message: string, parent?: Error) {
     super("@nomicfoundation/hardhat-verify", message, parent);
   }
 }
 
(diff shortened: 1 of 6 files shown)


packages/hardhat-verify/src/errors.ts
```diff
@@ -288,20 +288,20 @@ contract: "contracts/Example.sol:ExampleContract"
   }
 }
 
-export class DeployedBytecodeDoesNotMatchFQNError extends HardhatEtherscanError {
+export class DeployedBytecodeDoesNotMatchFQNError extends HardhatVerifyError {
```
fvictorio, line 291: We are going to dedup this.

schaable: Done!

fvictorio (approved):

PR #3998, opened 2023-06-06
feat: `rethnet_rpc_server` crate

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/95
Fixes https://github.com/NomicFoundation/rethnet/issues/92

- [x] address [this comment](https://github.com/NomicFoundation/hardhat/pull/3998/files#r1235610872) by moving the compiler input and expected output out of the code and into fixture files.
- [x] address [this comment](https://github.com/NomicFoundation/hardhat/pull/3998#discussion_r1235612107) by extracting the test helper method `help_test_method_invocation_serde` into a place commonly accessible to both `rethnet_eth::tests` and `rethnet_rpc_server::tests`, since each of those contains a nearly identical version.
- [x] address [this comment](https://github.com/NomicFoundation/hardhat/pull/3998#discussion_r1248490986) by reconciling the block tag resolution logic with what exists in hardhat, and also by examining all of the hardhat logic we should be replicating, to see if anything else doesn't match.


crates/rethnet_rpc_server/src/methods/compiler_io.rs
```diff
@@ -0,0 +1,110 @@
...
+struct CompilerSettings {
+    #[serde(rename = "viaIR")]
+    via_ir: Option<bool>,
+    optimizer: OptimizerSettings,
+    metadata: Option<MetadataSettings>,
+    /// maps a source name to a mapping from contract names to a vector of output
+    /// selections:
+    output_selection: HashMap<String, HashMap<String, Vec<String>>>,
+    evm_version: Option<String>,
+    /// maps a library file name to a mapping from library name to library content:
```
Wodann, line 29:
```suggestion
    /// Mapping of: file name -> (library name -> library content)
```

I am finding these hard to read. Just an idea, but would the above format make it easier to comprehend?

crates/rethnet_rpc_server/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "rethnet_rpc_server"
+version = "0.1.0"
```
Wodann, line 3:
```suggestion
version = "0.1.0-dev"
```

crates/rethnet_eth/src/remote.rs
```diff
@@ -28,7 +28,9 @@ impl From<u64> for U64 {
     }
 }
 
-fn single_to_sequence<S, T>(val: &T, s: S) -> Result<S::Ok, S::Error>
+/// for use with serde's serialize_with on a single value that should be serialized as a
+/// sequence
+pub fn single_to_sequence<S, T>(val: &T, s: S) -> Result<S::Ok, S::Error>
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/515cab7cc4cb75c494c7d3b65f186757237d74ff, though i chose to use a more specific name than `util`

crates/rethnet_rpc_server/src/methods.rs
```diff
@@ -0,0 +1,781 @@
...
+    #[serde(
+        rename = "hardhat_stopImpersonatingAccount",
+        serialize_with = "single_to_sequence",
+        deserialize_with = "sequence_to_single"
+    )]
+    StopImpersonatingAccount(Address),
+}
+
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+struct RpcHardhatNetworkConfig {
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/7edf75ec9dc52682479b08a4f8e5d82256a72a71, though i chose to use a different name than `rpc`, since that's such an overloaded term in this crate.

crates/rethnet_rpc_server/src/methods/compiler_io.rs
```diff
@@ -0,0 +1,110 @@
...
+
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+struct CompilerSettings {
+    #[serde(rename = "viaIR")]
+    via_ir: Option<bool>,
+    optimizer: OptimizerSettings,
+    metadata: Option<MetadataSettings>,
+    /// maps a source name to a mapping from contract names to a vector of output
+    /// selections:
```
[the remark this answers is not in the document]
feuGeneA: i stole the terminology from this context: https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/types/artifacts.ts#L196

crates/rethnet_rpc_server/src/methods/compiler_io.rs
feuGeneA: thanks for pushing to improve this. addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/5790528b22e19b3bf6ecd0e93d092f35a62c2334

crates/rethnet_rpc_server/Cargo.toml
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/3221c121a58a8833cf2c5540cc0e0deb0b551244

crates/rethnet_eth/src/remote/methods/hardhat/compiler_io.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(missing_docs)]
```
Wodann: Why don't you add docs? New code should be written with docs. I've only included this for old code as we're slowly moving towards all code having docs

crates/rethnet_eth/src/remote/methods.rs
```diff
@@ -0,0 +1,15 @@
+use crate::remote::MethodInvocation as EthMethodInvocation;
+
+/// RPC methods specific to Hardhat
+pub mod hardhat;
+
+/// an RPC method with its parameters
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+#[serde(untagged)]
+#[allow(clippy::large_enum_variant)]
+pub enum MethodInvocation {
```
Wodann, line 11: hardhat methods are not part of `eth`. They are our own custom methods. As such, they and `MethodInvocation` should live outside of the `rethnet_eth` crate. They were fine where they were, or if you want in a separate `rethnet_rpc` crate

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -56,6 +59,24 @@ pub async fn router(state: StateType) -> Router {
                                     }
                                 }))
                             }
+                            MethodInvocation::Hardhat(HardhatMethodInvocation::SetBalance(address, balance)) => {
```
Wodann: Instead of this becoming an enormous `match` statement with all logic, could you split it using function calls?

feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/7ab3fd0879636121fb43112fc6a0c19f3b91981c

crates/rethnet_eth/src/remote/methods/hardhat.rs
```diff
@@ -11,6 +9,9 @@ use crate::{
 /// Compiler input and output structures used as parameters to Hardhat RPC methods
 pub mod compiler_io;
 
+/// input types for use with hardhat_reset
+pub mod reset;
```
Wodann, line 13: As long as the `mod` is not public, you don't need to add the docs.

Does it need to be `pub`?

crates/rethnet_eth/src/remote/methods/hardhat.rs
```diff
@@ -7,7 +7,7 @@ use crate::{
 };
 
 /// Compiler input and output structures used as parameters to Hardhat RPC methods
-pub mod compiler_io;
+pub mod add_compilation_result;
```
Wodann, line 10: Does it need to be public?

crates/rethnet_eth/src/remote/methods/hardhat.rs
feuGeneA: it needed to be public back when it lived in `rethnet_eth`, but now that it's co-located in `rethnet_rpc_server` it doesn't. i removed the "allow missing docs" directive in https://github.com/NomicFoundation/hardhat/pull/3998/commits/c86e0d6425c249174b9ddfa0630dff480663b03b

crates/rethnet_eth/src/remote/methods/hardhat.rs
feuGeneA: it needed to be public back when it lived in `rethnet_eth`, but now that it's co-located in `rethnet_rpc_server` it doesn't. i removed the "allow missing docs" directive in https://github.com/NomicFoundation/hardhat/pull/3998/commits/c86e0d6425c249174b9ddfa0630dff480663b03b

crates/rethnet_eth/src/remote/methods.rs
feuGeneA: i moved them back to `rethnet_rpc_server` in https://github.com/NomicFoundation/hardhat/pull/3998/commits/37b288e819d30d5913a8cb7425bc47657ec59912

crates/rethnet_eth/src/remote/methods/hardhat/compiler_io.rs
feuGeneA: these types were public back when they were in `rethnet_eth`, but now that i've moved them to `rethnet_rpc_server` (in https://github.com/NomicFoundation/hardhat/pull/3998/commits/f2fd69a0b195853f6019abac23e8ad3e6e22152c), i removed this directive in https://github.com/NomicFoundation/hardhat/commit/c86e0d6425c249174b9ddfa0630dff480663b03b

crates/rethnet_eth/src/remote/methods.rs
feuGeneA: sorry, disregard the previous comment, i actually moved them back to `rethnet_rpc_server` in [THIS commit](https://github.com/NomicFoundation/hardhat/pull/3998/commits/f2fd69a0b195853f6019abac23e8ad3e6e22152c)

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+                })
+                .expect("shouldn't happen"),
+            ),
+        )
+    }
+}
+
+/// returns the state root in effect BEFORE setting the block context, so that the caller can
+/// restore the context to that state root.
+async fn set_block_context<T>(
```
feuGeneA, line 72: This function isn't good enough. It should actually support block tags. Right now it only supports `"latest"` (which effectively requires no effort haha). If we're in the context of a forked node, this function should be using an RPC client to use the remote node to resolve the block tag to a number, and THEN set the block context. I've started making these changes but haven't completed them yet.

crates/rethnet_rpc_server/src/methods/compiler_io.rs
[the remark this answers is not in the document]
Wodann: @fvictorio Could you elaborate for us what it means?

crates/rethnet_eth/src/remote/methods.rs
```diff
@@ -0,0 +1,616 @@
...
+#[cfg(test)]
+mod tests {
+    use bytes::Bytes;
+
+    use super::*;
+
+    fn help_test_method_invocation_serde(call: MethodInvocation) {
+        let json = serde_json::json!(call).to_string();
+
+        // validate that variations of MethodInvocation which have single values can still be
```
Wodann, line 241: Is this testing Rethnet behaviour or serde behaviour?

crates/rethnet_rpc_server/src/hardhat_methods.rs
```diff
@@ -0,0 +1,766 @@
...
+
+        let call_decoded: HardhatMethodInvocation = serde_json::from_str(&json)
+            .unwrap_or_else(|_| panic!("should have successfully deserialized json {json}"));
+        assert_eq!(call, call_decoded);
+    }
+
+    #[test]
+    fn test_serde_hardhat_add_compilation_result() {
+        // these were taken from a run of TypeScript function compileLiteral
+        let compiler_input_json = r#"{
```
Wodann, line 161: I'm curious what your thought process was in including the input as a literal here.

One possible alternative would be to use move this to a fixture and use https://doc.rust-lang.org/std/macro.include_str.html to include it instead?

Curious to hear your thoughts

crates/rethnet_rpc_server/src/hardhat_methods.rs
```diff
@@ -0,0 +1,766 @@
...
+                .unwrap(),
+            serde_json::from_str::<add_compilation_result::CompilerOutput>(compiler_output_json)
+                .unwrap(),
+        );
+
+        help_test_method_invocation_serde(call.clone());
+
+        match call {
+            HardhatMethodInvocation::AddCompilationResult(_, ref input, ref output) => {
+                assert_eq!(
```
Wodann, line 616: Is there a reason you're  no longer using the same helper method to do this - as for the Eth tests?

crates/rethnet_rpc_server/src/hardhat_methods/add_compilation_result.rs
```diff
@@ -0,0 +1,107 @@
...
+    source_map: String,
+    /// mapping: source name -> (mapping: library name -> LinkReferences)
+    link_references: HashMap<String, HashMap<String, Vec<LinkReference>>>,
+}
+
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct LinkReference {
+    start: usize,
+    length: usize, // TODO: this should always be 20; can we enforce that
```
Wodann, line 105: One option to achieve this is to write a manual implementation of serde, instead of using the derive macro

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+    state: &StateType,
+    block_spec: BlockSpec,
+) -> Result<B256, ResponseData<T>> {
+    match (*state).read().await.state_root() {
```
Wodann, line 76: I don't think you need to dereference state
```suggestion
    match state.read().await.state_root() {
```

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+pub async fn run(
+    address: SocketAddr,
+    config: RpcHardhatNetworkConfig,
+    genesis_accounts: HashMap<Address, AccountInfo>,
+) -> Result<SocketAddr, std::io::Error> {
+    let listener = TcpListener::bind(address)?;
+    let address = listener.local_addr()?;
+
+    let rethnet_state: StateType = if let Some(config) = config.forking {
+        Arc::new(RwLock::new(Box::new(ForkState::new(
```
Wodann, line 417: Consider splitting this into multiple lines with variable assignment to make it easier to read.

```rust
let runtime = tokio::runtime::Builder::new_multi_thread()
                    .enable_io()
                    .enable_time()
                    .build()
                    .expect("failed to construct async runtime");

let generator = { .. };
let state = { .. };

// etc.
```

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+        )
+    }
+}
+
+/// returns the state root in effect BEFORE setting the block context, so that the caller can
+/// restore the context to that state root.
+async fn set_block_context<T>(
+    state: &StateType,
+    block_spec: BlockSpec,
+) -> Result<B256, ResponseData<T>> {
```
Wodann, line 75: Aren't the return types of all functions known? Why are they a generic `T`?

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/e3a009f9b8d68719e849b423c009f01ddcf9a4c9

crates/rethnet_eth/src/remote/methods.rs
feuGeneA: It's verifying that we're using `sequence_to_single`/`single_to_sequence` on all of the `MethodInvocation` variants that only have a single parameter. see more context [here](https://github.com/NomicFoundation/hardhat/pull/3967#discussion_r1218522112).

crates/rethnet_rpc_server/src/hardhat_methods.rs
feuGeneA: I didn't put much thought into it at all, to be honest. It took some effort to get it this far, so I was just going for the most expedient path. Honestly I didn't even notice that the `compiler_output_json` was so huge, because my editor was folding it for me. I'm up for moving it to a fixture if you think that's the right thing to do.

crates/rethnet_rpc_server/src/hardhat_methods.rs
feuGeneA: Only that I wasn't sure of the best way to share it, since the `tests` submodules that contain each of these helpers are private. Where would be a good place for such a common helper to live?

crates/rethnet_rpc_server/src/hardhat_methods/add_compilation_result.rs
feuGeneA: Thanks, I like that idea. Addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/725fb498fc6d1fc3a9c5be471f8ded000cbe5b88

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: Thanks for suggesting this. I did it your way in my recent changes to block spec handling [here](https://github.com/NomicFoundation/hardhat/pull/3998/commits/e3a009f9b8d68719e849b423c009f01ddcf9a4c9), and i revised the rest of the instances in https://github.com/NomicFoundation/hardhat/pull/3998/commits/e6ee53f1af9d0e089bd75454f44111d4c86143d0

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: thanks, i like that. addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/ebeedf15fb102b05d7f642a4d4f2b1d20392fc92

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: the return types of the specific method handler functions are known, but these helpers are used with various different handlers with different return types.

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -67,67 +74,200 @@ where
...
+            .write()
+            .await
+            .set_block_context(
+                &KECCAK_EMPTY,
+                Some(match block_spec {
+                    BlockSpec::Number(n) => Ok(n),
+                    BlockSpec::Tag(tag) => {
+                        if let Some(fork_client) = state.fork_client.clone() {
+                            if let Some(block_number) =
+                                (*fork_client)
```
Wodann: If you query this, you'll potentially get a block number that's older than the latest block number supported in the `ForkState`.

If I'm not mistaken, both Rethnet & Hardhat's forked states make the block number immutable?

I had a look at the spec of `eth_getBlockByNumber` and our own implementation in Hardhat and only 5 different tags are supported. You can model that using an enum and make it easier on yourself:

https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/core/jsonrpc/types/input/blockTag.ts#L15

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -175,7 +315,7 @@ async fn handle_set_code(
 async fn handle_set_nonce(state: StateType, address: Address, nonce: U256) -> ResponseData<()> {
     match TryInto::<u64>::try_into(nonce) {
         Ok(nonce) => {
-            match (*state).write().await.modify_account(
+            match (*state.rethnet).write().await.modify_account(
```
Wodann: Can't you do `state.rethnet.write()` here too?

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -201,7 +341,7 @@ async fn handle_set_storage_at(
     position: U256,
     value: U256,
 ) -> ResponseData<()> {
-    match (*state)
+    match (*state.rethnet)
```
Wodann: Does `state.rethnet.write` work?

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -39,7 +39,14 @@ pub enum MethodInvocation {
     Hardhat(HardhatMethodInvocation),
 }
 
-type StateType = Arc<RwLock<Box<dyn SyncState<StateError>>>>;
+type RethnetStateType = Arc<RwLock<Box<dyn SyncState<StateError>>>>;
+
+struct AppState {
+    rethnet: RethnetStateType,
+    fork_client: Option<Arc<Mutex<rethnet_eth::remote::client::RpcClient>>>,
```
Wodann, line 46: You can simplify this by using a `use` statement

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: i think you're looking at an old/outdated revision? as explained [here](https://github.com/NomicFoundation/hardhat/pull/3998#discussion_r1236001000), i made these changes across two different commits (one in the "block spec handling" commit, and another specific to the remaining instances that didn't have anything to do with block spec handling).

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: i think you're looking at an old/outdated revision? as explained [here](https://github.com/NomicFoundation/hardhat/pull/3998#discussion_r1236001000), i made these changes across two different commits (one in the "block spec handling" commit, and another specific to the remaining instances that didn't have anything to do with block spec handling).

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/63a0f5532e6ed156d79ff4cf3468575b69c93cd7

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -39,7 +39,14 @@ pub enum MethodInvocation {
     Hardhat(HardhatMethodInvocation),
 }
 
-type StateType = Arc<RwLock<Box<dyn SyncState<StateError>>>>;
+type RethnetStateType = Arc<RwLock<Box<dyn SyncState<StateError>>>>;
+
+struct AppState {
+    rethnet: RethnetStateType,
```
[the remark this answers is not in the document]
feuGeneA: i was envisioning `AppState::rethnet` to later be of a different type (eg the rethnet runtime) once the blockchain stuff is ready. Also, the word "state" is already an overloaded concept in this file, so I was avoiding calling the field `state` for that reason. I renamed the field to `rethnet_state` for now, in https://github.com/NomicFoundation/hardhat/pull/3998/commits/ef8b2c2f80657866c32092be6f3a07dbf5fc850f

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: > If you query this, you'll potentially get a block number that's older than the latest block number supported in the `ForkState`.

After some discussion, we realized that this should have said NEWER, not older.

feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/1b2980892569f3bd219d2546635a3939c4393849

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -44,6 +44,7 @@ type RethnetStateType = Arc<RwLock<Box<dyn SyncState<StateError>>>>;
 struct AppState {
     rethnet_state: RethnetStateType,
     fork_client: Option<Arc<Mutex<RpcClient>>>,
+    fork_block_number: Option<usize>,
```
Wodann, line 47: Is a `U256` not supported?

crates/rethnet_eth/src/remote/methods.rs
Wodann: That's already validated by request -> json -> request though, right?

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -0,0 +1,526 @@
...
+        state
+            .rethnet_state
+            .write()
+            .await
+            .set_block_context(
+                &KECCAK_EMPTY,
+                Some(match block_spec {
+                    BlockSpec::Number(n) => Ok(n),
+                    BlockSpec::Tag(tag) => {
+                        if let Some(fork_client) = state.fork_client.clone() {
```
Wodann, line 101: This is still not the same logic as currently exists in Hardhat: https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts#L1388

So far, all of our code is a port from Hardhat's logic to Rust. As such, we need to make sure that we replicate all logic. When writing code, the ground truth should be the TypeScript code. The Hardhat tests that we run and validate afterwards merely serve as confirmation that we did it correctly.

If you haven't done this yet, I'd recommend validating that all of implementations actually follow the same logic as the [Eth](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts) and [Hardhat](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/hardhat-network/provider/modules/hardhat.ts) modules.

crates/rethnet_rpc_server/src/methods/compiler_io.rs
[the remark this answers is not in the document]
fvictorio: Sorry, I completely missed this mention.

`outputSelection`, in the context of compiling solidity with solc, means selecting which information will be included in the compilation output. You can read about it in the comment above that setting here: https://docs.soliditylang.org/en/v0.8.20/using-the-compiler.html#input-description

crates/rethnet_eth/src/remote/methods.rs
feuGeneA: No because if the variant failed to specify both of those helpers, then the request->json part would yield `params` as a single value and not an array, and then json->request part would deserialize from the single (non-array) `params` value back to the single value of the variant.

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: I modeled this after [the existing `HardhatNetworkForkingUserConfig`](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/types/config.ts#L91), which specifies it as a `number` rather than a `bigint`.

Wodann: In the blockchain it's a `U256`, so I'd perform an "upgrade" here and just use that type to support larger numbers too

crates/rethnet_rpc_server/src/hardhat_methods.rs
feuGeneA: Make that helper function generic over the method invocation type in order to avoid the circular dependency.

feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/3f8b866e69434a869e9bdbc78084c445928699b9

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: Addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/2cda55e1d9064f13c00e607ec7558131f7c08dad

crates/rethnet_rpc_server/src/hardhat_methods.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/cf97805377e5fcc749a8bf71905d04073a7bc6be

crates/rethnet_eth/src/remote.rs
```diff
@@ -39,6 +39,53 @@ impl From<u64> for U64 {
...
+}
+
+///
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+#[serde(untagged)]
+pub enum Eip1898BlockSpec {
+    /// to represent the Object { blockHash, requireCanonical } in EIP-1898
+    Hash(Eip1898BlockHash),
+    /// to represent the Object { blockNumber } in EIP-1898
+    Number(Eip1898BlockNumber),
```
Wodann, line 66: Why not inline the arguments?

To show the usage difference in Rust:

```
match spec {
-  Eip1898BlockSpec::Number(Eip1898BlockNumber { block_number }) => ...,
+ Eip1898BlockSpec::Number { block_number } => ...,
}
```

or

```
match spec {
-  Eip1898BlockSpec::Number(block_number) => block_number.block_number,
+ Eip1898BlockSpec::Number{ block_number => block_number,
}
```

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -75,77 +74,76 @@ where
     }
 }
 
+async fn get_latest_block_number<T>(state: &StateType) -> Result<U256, ResponseData<T>> {
+    if let Some(fork_block_number) = state.fork_block_number {
+        // TODO: when we're able to mint local blocks, and there are some newer than
```
Wodann, line 79: Could you please create a GitHub issue for this and link to it?

Wodann (changes requested):
I think there is just one last thing to address. Thanks for incorporating the changes

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/15c7cc92d0a9bbd3bcd435f52c12ad97bb684904

Wodann (approved):

PR #3999, opened 2023-06-07
fix aurora browser urls

ZumZoom:
Aurora team dropped support of the previous domain names. This change is needed to fix verification on Aurora.


fvictorio:
Thank you! This also made me realize that we had a wrong changeset (a leftover of renaming hardhat-etherscan to hardhat-verify), so I fixed that too.

PR #4005, opened 2023-06-08
Pin `web3` version in installation guides

pcaversaccio:
This PR pins the `web3` version in the installation guide of `hardhat-truffle5` and `hardhat-web3`. The reason is that `web3` version `4.0.1` just got released (https://github.com/web3/web3.js/releases/tag/v4.0.1) and `npm` targets this version now which I think is incompatible with the current `hardhat-truffle5` and `hardhat-web3` versions.

I don't think this PR deserves a changeset, but lmk if you want one.


fvictorio:
Good catch, thanks!

TrejGun:
@fvictorio can you please share your official opinion on the new web3 version? are you going to support it? may be timeline?

fvictorio:
I don't have an opinion, but it's not in our plans right now. At the very least, supporting viem is something we are going to work on first.

fvictorio:
That being said, feel free to open a feature request so we can gauge interest 🙂

PR #4008, opened 2023-06-08
Ledger Wallet provider package

nicosantangelo:
- [x] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team

---

# LedgerProvider

This PR adds a new package that extends provider allowing it to connect and interact with a Ledger wallet. To use it you'd just install `@nomicfoundation/hardhat-ledger` in your project and import it in your hardhat config:

```ts
import "@nomicfoundation/hardhat-ledger";

import { HardhatUserConfig } from "hardhat/config";

// Possible configuration
  solidity: "0.8.18",
  networks: {
    hardhat: {
      ledgerAccounts: [
        "0x18225dbbd263d5a01ac537db4d1eefc12fae8b24",
        "0x1f8c3cc4dd1fdf560cb2560a9b5f80dbde55c9d8",
      ],
    },
  },
};
```

For more information on how it works you can check the [README](https://github.com/NomicFoundation/hardhat/tree/feat/hh-ledger/packages/hardhat-ledger)


packages/hardhat-ledger/src/internal/create-ledger-provider.ts
```diff
@@ -0,0 +1,17 @@
...
+import { withSpinners } from "./with-spinners";
+
+export function createLedgerProvider(
+  provider: EIP1193Provider,
+  networkConfig: HardhatNetworkUserConfig
+): LedgerProvider {
+  const accounts =
+    networkConfig.ledgerAccounts === undefined
+      ? []
+      : networkConfig.ledgerAccounts;
```
fvictorio, line 13: This can be written as `networkConfig.ledgerAccoutns ?? []`. Maybe we should add the [`prefer-nullish-coalescing`](https://typescript-eslint.io/rules/prefer-nullish-coalescing/) rule (with `ignoreTernaryTests` set to false) to our eslint config. Do you want to do that? It will also mean changing a lot of code in other packages where we are not doing that, but the auto-fix of that rule seems to work well, so it's just a matter of running `yarn lint:fix` from the root once after adding the rule.

packages/hardhat-ledger/src/provider.ts
```diff
@@ -0,0 +1,437 @@
...
+    this.emit("derivation_start");
+
+    let path = "";
+    try {
+      for (
+        let account = 0;
+        account <= LedgerProvider.MAX_DERIVATION_ACCOUNTS;
+        account++
+      ) {
+        path = `44'/60'/${account}'/0'/0`;
```
fvictorio, line 356: I think the last apostrophe is wrong? At least that's what BIP-44 says, and also I couldn't use an account in my ledger until I removed it.

If it is indeed wrong, we also need to update all the error messages that use explicit derivation paths.

fvictorio (commented):
I haven't reviewed the code in depth yet, I just skimmed some parts and then tried to use the plugin locally. Some comments:

- If I just import the plugin and run a script that just does `console.log(await network.provider.send("eth_accounts", []))`, I get an error. Importing the plugin or not shouldnt' cause any significant changes if the user doesn't do anything with the ledger.
- We should check that the values set by the user in `ledgerAccounts` are valid addresses.
- I couldn't send a transaction without explicitly setting a gas price
- The normal accounts don't seem to be merged with the ledger accounts. That is, if I set `ledgerAccounts: ["0x123..."]`, then that (by default) should be the first account _after_ the "normal" configured accounts (in the case of the Hardhat network, which starts with 20 accounts, this would mean that the first ledger account would be the 21st).
- What do you think about changing the spinner message during derivation to something like `Trying address #1 (derivation path: "m/44'/60'/0'/0/0")`?

nicosantangelo:
> I couldn't send a transaction without explicitly setting a gas price

Regarding this, you meant a transaction with or without using your Ledger? The transaction we send with the ledger is a signed (raw) transaction that needs an explicit `gas` to work when sent to the internal hardhat-network provider (unless I'm mistaken).

If it's a transaction without using the ledger, it should be fixed with the latest commits!

packages/hardhat-ledger/src/provider.ts
nicosantangelo: It is 🤦 , testing residue, great catch. Changed to `44'/60'/0/0/0`

packages/hardhat-ledger/src/internal/create-ledger-provider.ts
nicosantangelo: Sure! only thing'd be that you'd have to review code from other packages here. I'll do it in a single commit so we can remove it if you feel like it messes up the review :)

packages/hardhat-ledger/src/index.ts
```diff
@@ -0,0 +1,8 @@
+import { extendProvider } from "hardhat/config";
+import { HardhatNetworkUserConfig } from "hardhat/types";
+import { createLedgerProvider } from "./internal/create-ledger-provider";
+
+extendProvider(async (provider, config, network) => {
+  const networkConfig = config.networks[network] as HardhatNetworkUserConfig;
```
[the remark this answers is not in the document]
nicosantangelo: Ohh I see. I think this was a gap in my understanding of the types. I thought at first glance that `Hardhat` in `HardhatNetwork(..)` was just a prefix and not _the_ `hardhat` network.

I'll re-read it, try to fix it, but if I run into trouble I'll let you know

packages/hardhat-core/src/types/config.ts
```diff
@@ -46,6 +46,7 @@ export interface HardhatNetworkUserConfig {
   hardfork?: string;
   mining?: HardhatNetworkMiningUserConfig;
   accounts?: HardhatNetworkAccountsUserConfig;
+  ledgerAccounts?: HardhatLedgerAccountsConfig;
```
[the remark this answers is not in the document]
nicosantangelo: that's waay cleaner haha. Missed that, on it!

packages/hardhat-ledger/README.md
```diff
@@ -0,0 +1,90 @@
...
+
+This plugin adds nothing to the Hardhat Runtime Environment.
+
+### Provider extensions
+
+The provider supplied by Hardhat will be extended using [`extendProvider`](https://hardhat.org/hardhat-runner/docs/advanced/building-plugins#extending-the-hardhat-provider), decorating it to be a `LedgerProvider`. Any other calls to extendProvider will be added on top of this.
+
+A `LedgerProvider` knows how to connect and interact with a Ledger wallet
+
+## Usage
```
nicosantangelo, line 43: I wanted to write an "advanced usage" section detailing how to extend the provider by hand, importing `LedgerProvider` but I'm not sure that's encouraged. Let me know if you think is a good idea.

Nevertheless I've exported the errors from `index.ts` just in case the use-case of catching a particular derivation/connection/etc error arises.

packages/hardhat-ledger/src/internal/create-ledger-provider.ts
fvictorio: Yes, doing that in a single commit would be ideal.

packages/hardhat-toolbox/package.json
```diff
@@ -33,21 +33,21 @@
...
     "@nomiclabs/hardhat-ethers": "^2.0.0",
     "@nomiclabs/hardhat-etherscan": "^3.0.0",
     "@typechain/ethers-v5": "^10.1.0",
     "@typechain/hardhat": "^6.1.2",
     "@types/chai": "^4.2.0",
     "@types/mocha": ">=9.1.0",
     "@types/node": "^14.0.0",
-    "@typescript-eslint/eslint-plugin": "4.29.2",
-    "@typescript-eslint/parser": "4.29.2",
+    "@typescript-eslint/eslint-plugin": "5.53.0",
```
nicosantangelo, line 48: Had to update these dependencies (`@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser`) from `4.29.2` to `5.53.0` as the other packages cause the old version does not support `ignoreTernaryTests`.

That broke the `@typescript-eslint/experimental-utils` import but I feel like I I'm just monkey patching by installing it here to make it work:

```bash
$ npx yarn lint
 | Error: Failed to load plugin '@nomiclabs/eslint-plugin-hardhat-internal-rules' declared in '.eslintrc.js » ../packages/hardhat-toolbox/../../config/eslint/eslintrc.js': Cannot find module '@typescript-eslint/experimental-utils'
 | Require stack:
 | - ../packages/eslint-plugin/onlyHardhatErrorRule.js
 | - ../packages/eslint-plugin/index.js```

Comment (fvictorio) [reply packages/hardhat-toolbox/package.json:48]: Oh, I didn't realize it was going to be that much of a hassle. Don't worry, I opened https://github.com/NomicFoundation/hardhat/issues/4028 to remember to do that at some point.

File: packages/hardhat-ledger/src/provider.ts [hunk 7]
```diff
@@ -0,0 +1,445 @@
...
+      }
+    } catch (error) {}
+  }
+
+  public async request(args: RequestArguments): Promise<unknown> {
+    const params = this._getParams(args);
+
+    if (
+      args.method === "eth_accounts" ||
+      args.method === "eth_requestAccounts"
```
fvictorio, line 118: Hardhat doesn't support `eth_requestAccounts`, so I think we shouldn't add it here (otherwise it would be weird to have a new supported method after adding the plugin).

packages/hardhat-ledger/src/provider.ts
```diff
@@ -0,0 +1,445 @@
...
+
+    if (!isControlledAddress) {
+      throw new HardhatLedgerNotControlledAddressError(
+        "Tried to send a transaction with an address we don't control.",
+        hexAddress
+      );
+    }
+  }
+
+  private _toHex(value: string | Buffer): string {
```
fvictorio, line 440: I would add a check here in case an already 0x-prefixed string is passed. Since this function is called several times with many kinds of values, I think it makes sense to be more defensive. Wdyt?

packages/hardhat-ledger/src/provider.ts
```diff
@@ -0,0 +1,445 @@
...
+
+    await this._requireControlledInit(txRequest.from);
+
+    if (txRequest.gas === undefined) {
+      throw new HardhatError(ERRORS.NETWORK.MISSING_TX_PARAM_TO_SIGN_LOCALLY, {
+        param: "gas",
+      });
+    }
+
+    if (txRequest.from === undefined) {
```
fvictorio, line 257: This check doesnt' seem necessary. `validateParams` already validates that it's not undefined, and anyway the `this._requireControlledInit` is already assuming that it's not undefined.

packages/hardhat-ledger/src/provider.ts
```diff
@@ -0,0 +1,445 @@
...
+  private async _derivePath(addressToFindAsBuffer: Buffer): Promise<string> {
+    const addressToFind = this._toHex(addressToFindAsBuffer).toLowerCase();
+
+    if (this.paths[addressToFind] !== undefined) {
+      return this.paths[addressToFind];
+    }
+
+    this.emit("derivation_start");
+
+    let path = "";
```
fvictorio, line 345: When I saw this, I tried to convert it into a `string | undefined` variable, but I see that that makes the code unnecessarily more complex, because we "know" this variable is going to be immediately set.

That being said, in these cases I prefer to use a more recognizable value, in case this ends up being shown in an error message for any reason. So what do you think about something like:

```suggestion
    let path = "<unset-path>";
```

packages/hardhat-ledger/src/provider.ts
```diff
@@ -0,0 +1,445 @@
...
+    const provider = new LedgerProvider(options, wrappedProvider);
+    await provider.init();
+    return provider;
+  }
+
+  constructor(
+    public readonly options: LedgerOptions,
+    protected readonly _wrappedProvider: EIP1193Provider
+  ) {
+    super(_wrappedProvider);
```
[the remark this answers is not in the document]
nicosantangelo: That's much better! Good eye

packages/hardhat-ledger/src/provider.ts
```diff
@@ -0,0 +1,445 @@
...
+
+  public get eth(): EthWrapper {
+    if (this._eth === undefined) {
+      throw new HardhatError(ERRORS.GENERAL.UNINITIALIZED_PROVIDER);
+    }
+    return this._eth;
+  }
+
+  public async init(): Promise<void> {
+    if (this._eth === undefined && !this._isCreatingTransport) {
```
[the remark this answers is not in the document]
nicosantangelo: What do you think about removing it and leaving a comment explaining the situation (and maybe linking to this comment)?
Just because I think we can easily forget about it and in case someone is confused by an error the lib is throwing and decides to read the code.

(I'll [just remove it](https://github.com/NomicFoundation/hardhat/pull/4008/commits/ceb7dfa5911533512926fb283219d7b3be1dc962) for now and add the comment later if you think it's a good idea)

packages/hardhat-ledger/src/internal/cache.ts
```diff
@@ -0,0 +1,44 @@
...
+export async function write<T extends CacheableJson>(json: T) {
+  const ledgerCacheFile = await getLedgerCacheFile();
+  await fs.writeJSON(ledgerCacheFile, json);
+}
+
+export async function read<T>(): Promise<T | undefined> {
+  const ledgerCacheFile = await getLedgerCacheFile();
+  try {
+    await fs.access(ledgerCacheFile, fs.constants.F_OK);
+    const file: T = await fs.readJSON(ledgerCacheFile);
```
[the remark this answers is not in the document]
nicosantangelo: Honestly you're right. It was kinda muscle memory, but the only reason why i'd leave it there is "performance", which I can't really measure. Removing it!

packages/hardhat-ledger/src/internal/cache.ts
```diff
@@ -0,0 +1,44 @@
...
+  return compilersCache;
+}
+
+async function getCacheDir(): Promise<string> {
+  const { cache } = await generatePaths();
+  await fs.ensureDir(cache);
+  return cache;
+}
+
+async function generatePaths(packageName = "hardhat"): Promise<Paths> {
```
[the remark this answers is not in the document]
nicosantangelo: Makes sense! I toyed with the idea of just not taking a param really (as they are internal/private methods) but ended up with this implementation.

I'll change it from [here](https://github.com/NomicFoundation/hardhat/pull/4008/files#diff-b9d3db0dc755826c3c3a8070393addbedd62d6d1497e724804ee9430fa2dff5fR23) too! Let me know what you think.

‼️ This is completely out of scope but I copied the idea from `/hardhat-core/src/internal/util/global-dir.ts` in case you keep track of things like this to change later (or you prefer I change it there too ofc)

packages/hardhat-ledger/src/provider.ts
nicosantangelo: Oh did not realize. Removed it, it'll error out with `MethodNotFoundError: Method eth_requestAccounts not found` by default, when forwarded to the inner provider

‼️ Just in case, `eth_requestAccounts` is supported on `LocalAccountsProvider` [here](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/core/providers/accounts.ts#L51)

packages/hardhat-ledger/src/provider.ts
nicosantangelo: I've never thought of that but I like it! It can also act as a failsafe if the variable is not properly set later (and there are tests)

The wonders if having to move variables out of scope to be able to use them in errors haha, I'll change it and keep it in mind for future cases

packages/hardhat-ledger/src/provider.ts
nicosantangelo: Right you are, good catch!

packages/hardhat-ledger/src/provider.ts
nicosantangelo: Oh good point. I'm having to resist a very strong urge to move it to a `utils.ts` file and properly test it, but I tend to be very careful when adding "bag of methods" files like `utils`.

What do you think? I see a few `utils.ts` and `utilities.ts` in the project but I don't know if you have a preference/way of handling this (besides just leaving it as a private method). For example, I could add a `address.ts` file to `internal` and keep `toHex` there, instead of just going with `utils`

I'll add the check for `0x` locally anyways!

packages/hardhat-ledger/src/provider.ts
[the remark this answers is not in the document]
fvictorio: Makes sense, let's do that.

packages/hardhat-ledger/src/internal/cache.ts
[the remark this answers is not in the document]
fvictorio: Ah, I see. Yeah, I'm tempted to change it, but it feels out of scope. Thanks for pointing that out though.

packages/hardhat-ledger/src/provider.ts
fvictorio: Moving it to a separate module to be able to test it is fine by me, yes. My only nitpick is that the file is called `address.ts` but this is used for non-address buffers too :sweat_smile:

Tbh I'm fine with calling it `utils.ts`. But I understand it feels like the lazy option.

PR #4011, opened 2023-06-09
This removes an useless array

anajuliabit:

---

Hi, there seems to be an array with no purpose.


fvictorio:
Nice catch!

I think it _technically_ has a purpose: you can have multiple compilation jobs with the same version, therefore doing this means that you only call `semver.lt` once. But that seems like a micro-optimization attempt (and I doubt it has an impact), so I am ok with simplifying the code.

PR #4018, opened 2023-06-12
Add HARDHAT_DISABLE_TELEMETRY_PROMPT envvar

### LINKED ISSUE #4012: there should be a way to disable the prompt for telemetry without asking the user
### Version of Hardhat

2.14.1

### What happened?

I have a template, that setup everything on one command
It run several process in parallele including one process to watch and compile contracts and other watching for artifact and deploying 

The issue is that if someone use this template and have never used hardhat before, the 2 processes will ask the user the same question

Furthermore, it is not that clear to the user why its process is waiting for.

I tried to set in my template's config but it still ask the user the question

```
analytics: {enabled: false},
```

This assume you have zellij installed but you could follow the README instruction and run each process separately

```bash
rm ~/.config/hardhat-nodejs/telemetry-consent.json
npx degit wighawag/jolly-roger your-app-folder
cd your-app-folder
# edit contracts/hardhat.config.ts to add `analytics: {enabled: false},`
pnpm i
pnpm start
```

Actually if you want to test without zellij, pnpm offer a parallel mode to execute multiple task and here the issue would be even worse, as there would be no way for the user to interact with the questions

### Search terms

_No response_

fvictorio:
Closes #4012.

Adds a new `HARDHAT_DISABLE_TELEMETRY_PROMPT` variable that, if set to true, will prevent Hardhat from showing the telemetry consent prompt the first time it's run, or during the project initialization.

I also wrote an "Environment variables" reference page, which was long overdue.

### DIFF
--- a/.changeset/fluffy-papayas-reflect.md
+++ b/.changeset/fluffy-papayas-reflect.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Added a HARDHAT_DISABLE_TELEMETRY_PROMPT environment variable that can be set to `true` to prevent Hardaht from showing the telemetry consent prompt.
--- a/.changeset/fluffy-papayas-reflect.md
+++ b/.changeset/fluffy-papayas-reflect.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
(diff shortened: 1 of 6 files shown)


.changeset/fluffy-papayas-reflect.md
schaable, line 1: > Hardaht 
typo

schaable (approved):

PR #4021, opened 2023-06-13
Implement JSON-RPC server library

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/92

### DIFF
--- a/crates/rethnet_rpc_server/Cargo.toml
+++ b/crates/rethnet_rpc_server/Cargo.toml
@@ -4,10 +4,16 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+axum = "0.6.18"
 hashbrown = { version = "0.13", default-features = false, features = ["ahash"] }
+hyper = "0.14.26"
 rethnet_eth = { version = "0.1.0-dev", path = "../rethnet_eth", features = ["serde"] }
+rethnet_evm = { version = "0.1.0-dev", path = "../rethnet_evm" }
+rethnet_evm_napi = { version = "0.1.0-dev", path = "../rethnet_evm_napi" }
+revm = { git = "https://github.com/Wodann/revm", rev = "a4550cc", version = "3.1", default-features = false, features = ["dev", "secp256k1", "serde", "std"] }
 serde = { version = "1.0.147", default-features = false, features = ["derive"] }
 serde_json = { version = "1.0.89" }
+tokio = { version = "1.21.2", default-features = false, features = ["macros"] }
 
 [dev-dependencies]
 bytes = { version = "1.2.1", default-features = false }
--- a/crates/rethnet_rpc_server/src/lib.rs
+++ b/crates/rethnet_rpc_server/src/lib.rs
@@ -1,1 +1,200 @@
+use std::sync::Arc;
+
+use axum::{
+    extract::{Json, State},
+    Router,
+};
+use revm::db::components::state::StateRef;
+use tokio::sync::Mutex;
+
+use rethnet_eth::{
+    remote::{client::Request as RpcRequest, jsonrpc, MethodInvocation},
+    U256,
+};
+use rethnet_evm::state::{HybridState, RethnetLayer};
+
 mod methods;
+
+type StateType = Arc<Mutex<HybridState<RethnetLayer>>>;
+
+pub async fn router(state: StateType) -> Router {
+    Router::new().route(
+        "/",
+        axum::routing::post(
+            |State(rethnet_state): State<StateType>, payload: Json<serde_json::Value>| async move {
+                let request: RpcRequest = serde_json::from_value(payload.0.clone()).unwrap();
+                match request {
+                    RpcRequest {
+                        version,
+                        id,
+                        method: _,
+                    } if version != jsonrpc::Version::V2_0 => {
+                        Json(serde_json::json!(jsonrpc::Response {
+                            jsonrpc: jsonrpc::Version::V2_0,
+                            id,
+                            data: jsonrpc::ResponseData::<serde_json::Value>::Error { error: jsonrpc::Error {
+                                code: 0,
+                                message: String::from("unsupported JSON-RPC version"),
+                                data: Some(serde_json::to_value(version).unwrap()),
+                            }}
+                        }))
+                    },
+                    RpcRequest { version: _, id, method } => {
+                        match method {
+                            MethodInvocation::GetBalance(address, _block_spec) => {
+                                Json(serde_json::json!(jsonrpc::Response {
+                                    jsonrpc: jsonrpc::Version::V2_0,
+                                    id,
+                                    data: match (*(rethnet_state.lock().await)).basic(address) {
+                                        Ok(Some(account_info)) => {
+                                            jsonrpc::ResponseData::<U256>::Success { result: account_info.balance }
+                                        },
+                                        Ok(None) => {
+                                            jsonrpc::ResponseData::<U256>::Error { error: jsonrpc::Error {
+                                                code: 0, message: String::from("No such account"), data: None,
+                                            }}
+                                        }
+                                        Err(e) => {
+                                            jsonrpc::ResponseData::<U256>::Error { error: jsonrpc::Error {
+                                                code: 0, message: e.to_string(), data: None,
+                                            }}
+                                        }
+                                    }
+                                }))
+                            }
+                            _ => { // TODO: after adding all the methods here, eliminate this
+                                   // catch-all match arm.
+                                Json(serde_json::json!(jsonrpc::Response {
+                                    jsonrpc: jsonrpc::Version::V2_0,
+                                    id,
+                                    data: jsonrpc::ResponseData::<serde_json::Value>::Error { error: jsonrpc::Error {
+                                        code: 0,
+                                        message: String::from("unsupported JSON-RPC method"),
+                                        data: Some(serde_json::to_value(method).unwrap()),
+                                    }}
+                                }))
+                            }
+                        }
+                    }
+                }
+            }
+        )
+    )
+    .with_state(state)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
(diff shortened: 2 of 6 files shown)


crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -1 +1,200 @@
...
+    remote::{client::Request as RpcRequest, jsonrpc, MethodInvocation},
+    U256,
+};
+use rethnet_evm::state::{HybridState, RethnetLayer};
+
 mod methods;
+
+type StateType = Arc<Mutex<HybridState<RethnetLayer>>>;
+
+pub async fn router(state: StateType) -> Router {
```
Wodann, line 20: I recommend following the pattern used in the readme of axum: https://crates.io/crates/axum/

You can add types to the function signature that tell it what type of Json object you're expecting, reducing boilerplate code

image

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -1 +1,200 @@
...
+    use axum::{body::Body, http::Request};
+    use rethnet_eth::{
+        remote::{jsonrpc, BlockSpec, MethodInvocation},
+        Address,
+    };
+    use rethnet_evm::KECCAK_EMPTY;
+    use revm::primitives::state::AccountInfo;
+    use std::net::{SocketAddr, TcpListener};
+
+    fn start_server() -> SocketAddr {
```
Wodann, line 98: I'd move part of this function into the lib as `async fn run(addr: SocketAddr, genesis_accounts: HashMap<Address, AccountInfo>)`, which runs the `axum::Server`. We'll need that once we wrap this in the `rethnet` CLI too.

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -1 +1,200 @@
...
+                    .into_body(),
+            )
+            .await
+            .unwrap()
+            .to_vec(),
+        )
+        .expect("should decode response from UTF-8")
+    }
+
+    #[tokio::test]
```
Wodann, line 147: Given that these are integration tests - using the public API of rethnet_rpc_server - you can use the integration test folder: `crates/rethnet_rpc_server/tests`

See https://doc.rust-lang.org/cargo/guide/project-layout.html

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -1 +1,200 @@
...
+                .await
+                .unwrap();
+        });
+
+        address
+    }
+
+    async fn submit_request(address: &SocketAddr, request: &RpcRequest) -> String {
+        String::from_utf8(
+            hyper::body::to_bytes(
```
Wodann, line 128: Why not use `reqwest`, which we already have as a dependency?

crates/rethnet_rpc_server/Cargo.toml
```diff
@@ -4,10 +4,16 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
+axum = "0.6.18"
 hashbrown = { version = "0.13", default-features = false, features = ["ahash"] }
+hyper = "0.14.26"
 rethnet_eth = { version = "0.1.0-dev", path = "../rethnet_eth", features = ["serde"] }
+rethnet_evm = { version = "0.1.0-dev", path = "../rethnet_evm" }
+rethnet_evm_napi = { version = "0.1.0-dev", path = "../rethnet_evm_napi" }
```
Wodann, line 12: We should not need napi here

crates/rethnet_rpc_server/Cargo.toml
```diff
@@ -4,10 +4,16 @@ version = "0.1.0"
...
+revm = { git = "https://github.com/Wodann/revm", rev = "a4550cc", version = "3.1", default-features = false, features = ["dev", "secp256k1", "serde", "std"] }
```
Wodann, line 13: What do we need `revm` for?

Wodann (changes requested):

feuGeneA:
@Wodann the overlap of changes between this PR and #3998 was too much to manage them independently, so i just pushed all of the changes from here over to there. i addressed some of your comments here with commits in that other PR, and i'll link to them from here for your convenience. some of your comments here haven't been addressed with changes yet, but i'm going to continue the conversation here where you started them. let's add any new comments to #3998 instead of here.

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -1 +1,200 @@
...
+
+use rethnet_eth::{
+    remote::{client::Request as RpcRequest, jsonrpc, MethodInvocation},
+    U256,
+};
+use rethnet_evm::state::{HybridState, RethnetLayer};
+
 mod methods;
+
+type StateType = Arc<Mutex<HybridState<RethnetLayer>>>;
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/d39c5c522deb6c1f5c725ff0d6e4cf5d2f437a1b

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/07c267afd1162739fbe13cc16ce74af6c5e56238

crates/rethnet_rpc_server/Cargo.toml
feuGeneA: https://github.com/NomicFoundation/hardhat/pull/3998/commits/394d688b2f2f5e705615ac46e6f5eb6a16147b48

crates/rethnet_rpc_server/Cargo.toml
feuGeneA: i was using it in a few places but it was unnecessary and i eliminated those references (and this dependency) in https://github.com/NomicFoundation/hardhat/pull/3998/commits/6c036a9c5c614a8f227d2a8ecaebc10bc73cab78

crates/rethnet_rpc_server/Cargo.toml
feuGeneA: in an early iterations i had tried out a direct instantiation of the runtime object in rethnet_evm_napi, but i quickly realized that wasn't right, but this old dependency was left unnoticed. thanks for pointing it out.

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: i could use some more input on what exactly you'd like to see happen here. my understanding is that we'll only have a single "handler", because we'll only ever route to one "endpoint": `/`. so, i'm not sure what benefit there would be to breaking it out into a separate function like this. i did make a small change to have `axum` do the JSON parsing instead of doing it ourselves. see that [here](https://github.com/NomicFoundation/hardhat/pull/3998/commits/afc1a6d0be2cc29bf3303d6ba27568371053856b). is that what you had in mind?

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: but, _are_ these integration tests? i'm happy to defer to your judgement, but these actually felt like unit tests to me, and that's why i kept them here. i know that sometimes people say things like, "if it's doing network interaction then it's an integration test." but to me, an integration test is a test that exercises the integration of multiple different components. and to me that seems to be the reason for having the integration test folder in the rust project hierarchy: it provides a place to put tests that combine multiple different components, because such tests would be out of place within the `mod tests` sub-module of a single component, because they're not specific to any one component. here, however, all of the tests are specific to the contents of this `lib` module, so i felt they belonged within that module. again, i'm happy to defer to your preference, but that was my rationale.

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: i actually forgot about that. i just modeled this after the `axum` examples. but, if i understand correctly, this isn't introducing an additional dependency, because `hyper` is already a dependency of `axum`. do you think there's a good reason to pick one over the other for these tests?

Wodann: You do actually add the `hyper` dependency in the cargo toml. If not for this test, would you need it there?

feuGeneA: you're right, thanks for pushing. addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/77ee710ca3d23139066c1ede9c62e0fe0091ccc8

crates/rethnet_rpc_server/src/lib.rs
Wodann: I generally agree with what you said, but based on your description I'd recognise four components at work here:

1. Serialization
2. Client (reqwest)
3. Server (listener + handler)
4. Deserialization

As such, I'd say that it qualifies as an integration test under your definition too.

The Rust doc says something along similar lines: https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

crates/rethnet_rpc_server/src/lib.rs
Wodann: Yes, we'll have a single handler, but by using the code as described in the example, it automatically handles and responds with an error for the failure case when parsing. It reduces boilerplate for you.

feuGeneA: i see, but i'm still not seeing how you'd like to see it applied to the code here, especially now that i've introduced all the `handle_*` functions. if you think there's still room for further improvement beyond the introduction of those functions, let's discuss on a call.

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/3998/commits/37b288e819d30d5913a8cb7425bc47657ec59912

PR #4032, opened 2023-06-15
feat: user args instead of conf in compilation workflow

adjisb:

The compilation subtasks get configuration paths as parameters (so they can be overriden from plugins that change the compilation workflow).


fvictorio:
This looks really good @adjisb. I'll be a bit busy today, but I'll try to make a proper review as soon as I can.

fvictorio:
Thanks!

PR #4041, opened 2023-06-17
Fix broken link to hardhat-ethers

ethosdev:
Fix broken link to hardhat-ethers


fvictorio:
Thank you! I also added a redirect so that any existing link like that keeps working.

PR #4043, opened 2023-06-19
Add test for stack-too-deep error with console.log

ZumZoom:
Adds a test for a fix made in #3965


packages/hardhat-core/test/internal/hardhat-network/stack-traces/compilation.ts
```diff
@@ -34,14 +34,6 @@ function getSolcInput(
 ): CompilerInput {
   const isViaIR = compilerOptions.optimizer?.viaIR ?? false;
 
-  const optimizerDetails = isViaIR
-    ? {
-        yulDetails: {
-          optimizerSteps: "u",
```
ZumZoom, line 40: So the issue with using `"u"` as `optimizerSteps` is that it does not allow stack-to-memory movement for some reason.

However using default `optimizerSteps` leads to failure of the `console-logs::error::assert`. It seems the reason is that source mappings of viaIR compiled files are a mess and mostly unusable.

@fvictorio if you have any insights on the issue with the source mappings, I'd appreciate that.

packages/hardhat-core/test/internal/hardhat-network/stack-traces/test-files/0_8/console-logs/error/assert/c.sol
```diff
@@ -9,4 +9,10 @@ contract C {
     assert(b);
   }
 
+  // test2 is needed as workaround for the source mappings issue
```
ZumZoom, line 12: So for now I've fixed the issue with the `console-logs::error::assert` test by adding another function in the contract which forces solc to leave jumps from the `console` contract back to the `C` contract (to differentiate between `test` and `test2` calls) and also to generate proper source mappings for those jumps.

With single `test` function I believe those jumps were "optimized" by inliner and source mappings were unusable because of that optimization.

fvictorio:
(I'm not sure if your comments are still relevant if we take that testing approach, let me know if they are)

fvictorio:
> However I think that it'll be good to keep my changes to the [compilation.ts](https://github.com/NomicFoundation/hardhat/pull/4043#diff-079d38479fdf11ba357044ece4a41e867645d19bb2dfa5417b0a82deb10eb2ee) flow. It now uses default optimization pipeline instead of `u` which might be handy in case of new tests in the future.

That is intentional because that's [our current recommended setting](https://hardhat.org/hardhat-runner/docs/reference/solidity-support#support-for-ir-based-codegen) for viaIR. Maybe the latest versions of solc don't need that though. In any case, I think I'd rather keep it as it is right now. Do you want to open an issue both for that and for the cache invalidation problem? They are separate things, but I'm pretty sure it will make sense to work on both together.

> If you agree, I can leave only those two changes here, rename this PR to smth like "stack-traces tests optimization"

Assuming you are fine with the previous paragraph, I think we can merge the `clean` script fix. Thanks for noticing that, I can see how it made you lose some hours :disappointed:

ZumZoom:
Ok, so I am closing this PR then. I've made separate PR #4049 for the cleanup script and also created issues #4050 and #4051. Tbh I don't feel that they are related enough to keep them in one issue.

PR #4047, opened 2023-06-19
bad source mappings example

ZumZoom:
This draft is only created to show that there exist some internal dependencies between the tests which allow the `console-logs::error::assert` test to pass in main branch.

The test fails if run in isolation though.


ZumZoom:
Strange enough it works on CI, but fails on my machine. I am on OS X and the test that is failing on me is the `soljson-v0.8.17+commit.8df45f5f.js` one.

ZumZoom:
The issue was with the artifacts not getting properly cleaned up after the test.

PR #4057, opened 2023-06-22
feat: rethnet CLI `node` command

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/93


Wodann (changes requested):
Partial review. TBC

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -1,6 +1,8 @@
-use std::net::{SocketAddr, TcpListener};
+pub use std::net::SocketAddr;
```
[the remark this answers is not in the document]
feuGeneA: thanks, addressed in https://github.com/NomicFoundation/hardhat/pull/4057/commits/92e29ac45b6ec22cb2e79947c7feca990f9a9dd1

crates/rethnet/src/lib.rs
```diff
@@ -31,13 +64,54 @@ impl From<bool> for ExitStatus {
...
+
+                #[cfg(not(unix))]
+                let terminate = std::future::pending::<()>();
+
+                tokio::select! {
+                    _ = ctrl_c => {},
+                    _ = terminate => {},
+                }
+
+                println!("Shutting down");
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4057/commits/6752317cd797c3aab6a8cb73fa7c964b5252e100

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -458,13 +539,39 @@ async fn router(state: StateType) -> Router {
...
 /// returns the address that the server is listening on (with any input wildcards resolved).
-pub async fn run(
+pub async fn serve(
     address: SocketAddr,
     config: RpcHardhatNetworkConfig,
-    genesis_accounts: HashMap<Address, AccountInfo>,
-) -> Result<SocketAddr, std::io::Error> {
+    genesis_accounts: Option<HashMap<Address, AccountInfo>>,
+) -> Result<
+    axum::Server<hyper::server::conn::AddrIncoming, axum::routing::IntoMakeService<Router>>,
```
Wodann, line 547: I think not leaking the `axum::Server` type would also avoid needing the hyper dependency

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -458,13 +539,39 @@ async fn router(state: StateType) -> Router {
...
+    std::io::Error,
+> {
+    let genesis_accounts = genesis_accounts.unwrap_or(TEST_ACCOUNTS.iter().fold(
+        HashMap::default(),
```
Wodann, line 551: What made you decide to default to `TEST_ACCOUNTS` if the user doesn't specify `genesis_accounts`?

crates/rethnet/tests/cli.rs
```diff
@@ -0,0 +1,179 @@
...
+use rethnet_eth::{
+    remote::{
+        client::Request as RpcRequest, jsonrpc, methods::MethodInvocation as EthMethodInvocation,
+        BlockSpec,
+    },
+    Address, Bytes, U256,
+};
+use rethnet_rpc_server::{HardhatMethodInvocation, MethodInvocation, TEST_ACCOUNTS};
+
+fn shutdown(child: &mut Child) -> Result<(), std::io::Error> {
```
Wodann, line 18: Is there a reason you're not using: https://doc.rust-lang.org/std/process/struct.Child.html#method.kill

crates/rethnet/tests/cli.rs
```diff
@@ -0,0 +1,179 @@
...
+    // from gracefully terminating the server:
+    let send_result = reqwest::Client::new()
+        .post("http://127.0.0.1:8545/")
+        .header(reqwest::header::CONTENT_TYPE, "application/json")
+        .body(request_body)
+        .send()
+        .await;
+
+    // signal the server to shut down gracefully:
+    shutdown(&mut server)?;
```
Wodann, line 131: If this is a repeating pattern for tests, you can create a struct that implements the `Drop` trait and calls  shutdown` on drop. It makes it more explicit and doesn't require the comments to explain what's going on, aka self-explanatory code.

Wodann (changes requested):

crates/rethnet/src/lib.rs
```diff
@@ -31,13 +65,54 @@ impl From<bool> for ExitStatus {
...
 {
-    let args = Args::parse_from(args);
+    let args = Cli::parse_from(args);
     match args.command {
         Command::Start => Ok(ExitStatus::Success),
+        Command::Node(node_args) => {
+            tracing_subscriber::fmt::Subscriber::builder().init();
+
+            let server =
+                rethnet_rpc_server::serve("127.0.0.1:8545".parse()?, node_args.try_into()?, None)
```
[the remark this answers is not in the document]
Wodann: For reference:

```
Hardhat version 2.13.0

Usage: hardhat [GLOBAL OPTIONS] node [--fork <STRING>] [--fork-block-number <INT>] [--hostname <STRING>] [--port <INT>]

OPTIONS:

  --fork                The URL of the JSON-RPC server to fork from
  --fork-block-number   The block number to fork from
  --hostname            The host to which to bind to for new connections (Defaults to 127.0.0.1 running locally, and 0.0.0.0 in Docker)
  --port                The port on which to listen for new connections (default: 8545)

node: Starts a JSON-RPC server on top of Hardhat Network
```

crates/rethnet/src/lib.rs
```diff
@@ -13,6 +16,37 @@ struct Args {
 #[allow(clippy::large_enum_variant)]
 enum Command {
     Start,
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4057/commits/f4f8da6d3938061169b0688a6f2fa699b8b61e74

crates/rethnet/src/lib.rs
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4057/commits/e45e994d8b0ca07c10c0cb3913160603e4487ee6

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 use std::net::{SocketAddr, TcpListener};
 use std::sync::Arc;
 
+pub use axum::Server;
```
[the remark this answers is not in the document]
feuGeneA: yes we were using it externally but not very extensively, and abstracting it is a good idea, so i did it in https://github.com/NomicFoundation/hardhat/pull/4057/commits/77e9cc09bf764b2d9448fc35548afe72b6f61cff

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -514,13 +621,7 @@ pub async fn run(
...
-    tokio::spawn(async move {
-        axum::Server::from_tcp(listener)
-            .unwrap()
-            .serve(router(rethnet_state).await.into_make_service())
-            .await
-            .unwrap();
-    });
-
-    Ok(address)
+    Ok(axum::Server::from_tcp(listener)
```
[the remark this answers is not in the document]
feuGeneA: we do use the instance for a few things but it was easy enough to abstract it, so i did that, in https://github.com/NomicFoundation/hardhat/pull/4057/commits/77e9cc09bf764b2d9448fc35548afe72b6f61cff

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4057/commits/77e9cc09bf764b2d9448fc35548afe72b6f61cff
though, it didn't eliminate the dependency, as we're still keeping a private instance.

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -49,30 +54,33 @@ struct AppState {
...
-                serde_json::to_value(Response {
-                    jsonrpc: jsonrpc::Version::V2_0,
-                    id,
-                    data: ResponseData::<T>::new_error(0, "serde_json::to_value() failed", None),
-                })
-                .expect("shouldn't happen"),
-            ),
-        )
-    }
+pub const TEST_ACCOUNTS: [&str; 20] = [
```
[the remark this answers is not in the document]
feuGeneA: in retrospect i misnamed this. these addresses are the default addresses in Hardhat Network.

they're derived from https://github.com/NomicFoundation/hardhat/blob/62c61e139e96711b3de42ec00d2e7ce29d5536ff/packages/hardhat-core/src/internal/core/config/default-config.ts#L20

and they're visible on stdout when you run `npx hardhat node`.

i renamed this variable to `DEFAULT_ACCOUNTS` in https://github.com/NomicFoundation/hardhat/pull/4057/commits/56cfafd3f5c9e4d5eecf06180550ae6f9a235906

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -363,91 +372,163 @@ async fn handle_set_storage_at(
...
-                                ) => response(
-                                    id,
-                                    handle_get_transaction_count(state, address, block).await,
-                                ),
-                                MethodInvocation::Hardhat(HardhatMethodInvocation::SetBalance(
-                                    address,
-                                    balance,
-                                )) => {
-                                    response(id, handle_set_balance(state, address, balance).await)
+                |State(state): State<StateType>, payload: Json<serde_json::Value>| async move {
```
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4057/commits/74d8cbc6895a59608e4fba1ba9691839f77ba296

crates/rethnet/tests/cli.rs
feuGeneA: i think at first i was trying to simulate a ctrl+c(windows)/SIGINT(unix) rather than SIGTERM, thinking that that was more proper, and then along the way in my research i think i was led further astray by some crate that was offering a unix-only wrapper around the unix `kill` command, but eventually i settled on SIGTERM, but i hadn't realized `Child::kill` was a thing, but now i'm using it, as of [this commit](https://github.com/NomicFoundation/hardhat/commit/5ecc2971149d2938976e5c64f0cc4e279ed7105f)

crates/rethnet/tests/cli.rs
feuGeneA: the thing is that we need to use that `server` object even after the shutdown (see the next line) in order to get the standard output of the process, so we can validate its log entries, so i think shutting down the process via `Drop` doesn't fit what we need here. or do you see some other way to utilize `Drop`?

crates/rethnet_rpc_server/src/lib.rs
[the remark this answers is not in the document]
Wodann: It's important to remember that Rethnet is a general-purpose Eth Runtime Development library and as such should not tailor to one specific end-user, though.

Is there an easy way that you can still pass in these default accounts without hard-coding it in Rethnet?

Wodann (changes requested):
Almost there! Still one discussion thread that I responded to, which needs to be resolved

crates/rethnet_rpc_server/src/lib.rs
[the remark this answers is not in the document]
feuGeneA: i also raised this issue to track this going forward: https://github.com/NomicFoundation/hardhat/issues/4126

crates/rethnet/src/lib.rs
```diff
@@ -1,18 +1,68 @@
...
 #[allow(clippy::large_enum_variant)]
 enum Command {
-    Start,
+    Node(NodeArgs),
+}
+
+#[derive(Args)]
+struct NodeArgs {
+    #[clap(long, default_value = "127.0.0.1")]
+    host: String,
```
Wodann, line 29:
```suggestion
    /// Host IP address [default: 127.0.0.1]
    #[clap(long)]
    host: Option<String>,
```

crates/rethnet/src/lib.rs
```diff
@@ -1,18 +1,68 @@
...
+    type Error = anyhow::Error;
+
+    fn try_from(node_args: NodeArgs) -> Result<Config, Self::Error> {
+        Ok(Config {
+            address: match node_args.host.as_str() {
+                "127.0.0.1" => {
+                    SocketAddr::new(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)), node_args.port)
+                }
+                _ => format!("{}:{}", node_args.host, node_args.port).parse()?,
+            },
```
Wodann, line 48:
```suggestion
        let ip = node_args.host.map_or_else(
            || Ok(IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1)))
            |host| IpAddr::from_str(&host)
        )?;

        Ok(Config {
            address: SocketAddr::new(ip, node_args.port),
```

Wodann: This is what I had in mind. We're no longer constructing a temporary string using `format!`, which is expensive, nor parsing the IP address if it wasn't specified

Wodann (changes requested):

crates/rethnet/src/lib.rs
feuGeneA: Thanks for pushing this further. Re-addressed in https://github.com/NomicFoundation/hardhat/pull/4057/commits/8df18fc042d806758d2d12f46b7240dbf22bbc66

I did it a little different than your suggestion, because `node_args.host` actually isn't an option, since if it was left unspecified on the command line, at this point it has been resolved to it's default value. But my latest revision eliminates the `format!()` as well as the parsing of the port.

Wodann: Maybe this implicit detail was not clear based on my suggested code, but my point was that, from the user perspective, there is no difference between:

1. a manual mention of [default: 127.0.0.1]
```
host: Option<String>
```

and

2.
```
#[clap(..., default = "127.0.0.1")]
host: String,
```

Internally, `Clap` with just do the `unwrap_or_default` for you. Converting to a string and then having to do a `match` to detect a string equality is just a waste when we know that the default value when a `None` occurs should be a `const` constructed `IpV4Address`

feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4057/commits/8f61a479918d8edb63aacb5064ca161fcb0bc9a6

Wodann (approved):
LGTM! Nice find on the clap `default_value_t`

Feel free to merge once CI succeeds (excluding the dual mode runs)

PR #4075, opened 2023-06-27
Consume downloads instead of piping them

### LINKED ISSUE #3877: `hardhat compile` randomly and quietly stops before compiling contracts
Edit by @fvictorio: this is caused by a bug in Node v18.16.0. For a temporary workaround, downgrade your Node version to 18.15.0.

---

### Version of Hardhat

2.14.0

### What happened?

`npx hardhat compile` sometimes goes on with compiling contracts and sometimes (seems totally random) silently stops at the `Downloading compiler ...` step before any contracts are actually compiled. This issue is reproducible with no contracts already compiled and in docker.

Here's a simple repository where you can observe the problem: [repo](https://github.com/sebastiendan/hardhat-compile-issue).

This CI workflow run shows in the logs that the compilation didn't happen: [log](https://github.com/sebastiendan/hardhat-compile-issue/actions/runs/4788710757/jobs/8515780386#step:7:271).

In addition, one can run the docker container in interactive mode and manually run the `build` command to see that, again, the command sometimes doesn't build nor throws an error but simply quietly stops without compiling (same output as the CI log above):
- Run the container in interactive mode: `docker run --rm -it ghcr.io/sebastiendan/hardhat-compile-issue:main bash`
- Run the compile command: `npm run build`
- Observe sometimes an actual compilation, sometimes a quiet stop in the command execution
(run the above multiple times by exiting the container and re-running the two commands above)

### Search terms

no compilation

fvictorio:
This is a workaround for an issue related to node v18.16.0 (and v18.16.1), the stream module and undici. See https://github.com/nodejs/undici/issues/2173 for the details.

Fixes #3877.

### DIFF
--- a/.changeset/tricky-parents-warn.md
+++ b/.changeset/tricky-parents-warn.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed an issue related to compiler downloads in node v18.16.x
--- a/packages/hardhat-core/src/internal/util/download.ts
+++ b/packages/hardhat-core/src/internal/util/download.ts
@@ -1,9 +1,7 @@
 import type { Dispatcher } from "undici";
 
-import fs from "fs";
 import fsExtra from "fs-extra";
 import path from "path";
-import util from "util";
 
 import { getHardhatVersion } from "./packageInfo";
 import { shouldUseProxy } from "./proxy";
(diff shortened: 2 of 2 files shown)


wei3erHase:
THANKS! 🥕

PR #4091, opened 2023-06-30
Drop support for node v14 and add support for node v20

fvictorio:


packages/hardhat-core/src/internal/cli/bootstrap.ts
```diff
@@ -2,12 +2,12 @@
...
 import chalk from "chalk";
 
-const SUPPORTED_NODE_VERSIONS = ["^14.0.0", "^16.0.0", "^18.0.0"];
+import { SUPPORTED_NODE_VERSIONS } from "./constants";
 
 if (!semver.satisfies(process.version, SUPPORTED_NODE_VERSIONS.join(" || "))) {
   console.warn(
     chalk.yellow.bold(`WARNING:`),
-    `You are using a version of Node.js that is not supported, and it may work incorrectly, or not work at all. See https://hardhat.org/nodejs-versions`
+    `You are using a version of Node.js (${process.version}) that is not supported, and it may work incorrectly, or not work at all. See https://hardhat.org/nodejs-versions`
```
schaable, line 10: the error message does look a bit odd. What about something like: `You are currently using Node.js ${process.version}, which is not supported by Hardhat. This can lead to unexpected behavior or complete failure. See https://hardhat.org/nodejs-versions`

PR #4095, opened 2023-07-01
test: Add Rethnet provider as JSON-RPC provider in Hardhat tests

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/94

- [x] figure out why the `eth_getBalance` test `should result in a modified balance` is failing with `Failed to restore previous block context`
- [x] figure out why `hardhat_setStorageAt` is responding with `null` in some cases:
  - [x] sending `0xff` for the storage value is expected to yield error `[-32000] Storage value must be exactly 32 bytes long. Received 0xff, which is 1 bytes long`, but we’re responding with just `null`
  - [x] sending `0xffff` for the storage value is expected to yield error `[-32000] Storage value must be exactly 32 bytes long. Received 0xffff, which is 2 bytes long.`, but we’re responding with just `null`
  - [x] sending `0x` followed by 31 `f`s as the storage value is expected to yield error `[-32000] Storage value must be exactly 32 bytes long. Received 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff, which is 31 bytes long.` but we’re responding with just `null`

### DIFF
--- a/crates/rethnet_eth/src/remote.rs
+++ b/crates/rethnet_eth/src/remote.rs
@@ -62,6 +62,12 @@ pub enum Eip1898BlockSpec {
     },
 }
 
+impl Display for Eip1898BlockSpec {
+    fn fmt(&self, formatter: &mut Formatter) -> Result<(), fmt::Error> {
+        formatter.write_str(&serde_json::to_string(self).map_err(|_| fmt::Error)?)
+    }
+}
+
 /// possible block tags as defined by the Ethereum JSON-RPC specification
 #[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
 pub enum BlockTag {
@@ -121,12 +127,11 @@ impl BlockSpec {
 
 impl Display for BlockSpec {
     fn fmt(&self, formatter: &mut Formatter) -> Result<(), fmt::Error> {
-        let s = match self {
-            BlockSpec::Number(n) => n.to_string(),
-            BlockSpec::Tag(t) => t.to_string(),
-            BlockSpec::Eip1898(e) => serde_json::to_string(e).map_err(|_| fmt::Error)?,
-        };
-        formatter.write_str(&s)
+        match self {
+            BlockSpec::Number(n) => n.fmt(formatter),
+            BlockSpec::Tag(t) => t.fmt(formatter),
+            BlockSpec::Eip1898(e) => e.fmt(formatter),
+        }
     }
 }
 
--- a/crates/rethnet_eth/src/remote.rs
+++ b/crates/rethnet_eth/src/remote.rs
@@ -1,3 +1,6 @@
+use std::fmt;
+use std::fmt::{Display, Formatter};
+
 /// an Ethereum JSON-RPC client
 pub mod client;
 
@@ -79,6 +82,18 @@ pub enum BlockTag {
     Finalized,
 }
 
+impl Display for BlockTag {
+    fn fmt(&self, formatter: &mut Formatter) -> Result<(), fmt::Error> {
+        formatter.write_str(match self {
+            BlockTag::Earliest => "earliest",
+            BlockTag::Latest => "latest",
+            BlockTag::Pending => "pending",
+            BlockTag::Safe => "safe",
+            BlockTag::Finalized => "finalized",
+        })
+    }
+}
+
 /// For specifying a block
 #[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
 #[serde(untagged)]
@@ -101,6 +116,17 @@ impl BlockSpec {
     /// Constructs a `BlockSpec` for the pending transactions.
     pub fn pending() -> Self {
         Self::Tag(BlockTag::Pending)
+    }
+}
+
+impl Display for BlockSpec {
+    fn fmt(&self, formatter: &mut Formatter) -> Result<(), fmt::Error> {
+        let s = match self {
+            BlockSpec::Number(n) => n.to_string(),
+            BlockSpec::Tag(t) => t.to_string(),
+            BlockSpec::Eip1898(e) => serde_json::to_string(e).map_err(|_| fmt::Error)?,
+        };
+        formatter.write_str(&s)
     }
 }
 
--- a/crates/rethnet_eth/src/remote.rs
+++ b/crates/rethnet_eth/src/remote.rs
@@ -189,7 +189,14 @@ impl serde::Serialize for ZeroXPrefixedBytes {
     where
         S: serde::Serializer,
     {
-        serializer.serialize_str(&format!("0x{}", hex::encode(&self.inner)))
+        let encoded = hex::encode(&self.inner);
+        serializer.serialize_str(&format!(
+            "0x{}",
+            match encoded.as_str() {
+                "00" => "",
+                other => other,
+            }
+        ))
     }
 }
 
--- a/crates/rethnet_evm/src/state/hybrid.rs
+++ b/crates/rethnet_evm/src/state/hybrid.rs
@@ -363,7 +363,7 @@ mod tests {
     use rethnet_eth::Bytes;
 
     #[test]
-    fn repro_provider_test_should_result_in_a_modified_balance() {
+    fn test_irregular_state_modification() {
         // this test reproduces the sequence of calls that are induced by (the current
         // implementation of rethnet_rpc_server) being used as a provider in the hardhat tests and
         // running the hardhat_setBalance test entitled "should result in a modified balance".
--- a/crates/rethnet_evm/src/state/hybrid.rs
+++ b/crates/rethnet_evm/src/state/hybrid.rs
@@ -414,6 +414,7 @@ mod tests {
                 },
             )
             .unwrap();
+        state.make_snapshot();
 
         /* [eth_getBalance]
          * state_root
--- a/crates/rethnet_evm/src/state/hybrid.rs
+++ b/crates/rethnet_evm/src/state/hybrid.rs
@@ -355,3 +355,74 @@ impl StateHistory for HybridState<RethnetLayer> {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use rethnet_eth::Bytes;
+
+    #[test]
+    fn repro_provider_test_should_result_in_a_modified_balance() {
+        // this test reproduces the sequence of calls that are induced by (the current
+        // implementation of rethnet_rpc_server) being used as a provider in the hardhat tests and
+        // running the hardhat_setBalance test entitled "should result in a modified balance".
+
+        let mut state: HybridState<RethnetLayer> = Default::default();
+
+        let seed = 1;
+        let address = Address::from_low_u64_ne(1);
+        state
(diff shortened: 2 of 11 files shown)


packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
```diff
@@ -67,6 +67,7 @@ export const PROVIDERS = [
     isJsonRpc: false,
     networkId: DEFAULT_NETWORK_ID,
     chainId: DEFAULT_CHAIN_ID,
+    rethnetBinary: undefined as string | undefined,
```
fvictorio, line 70: I think you can drop this property from the binaries; it's not used anywhere. The rethnet binary is passed to the (inner) `useProvider` call in the definition of the `useProvider` method of the rethnet provider (sorry, I know this is super confusing), so that's enough.

packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
```diff
@@ -194,3 +197,22 @@ if (INFURA_URL !== undefined) {
     },
   });
 }
+
+if (RETHNET_BINARY !== undefined) {
+  PROVIDERS.push({
+    name: "Rethnet",
+    isFork: false,
+    isJsonRpc: false,
```
fvictorio, line 205: I think this should be true?

packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
```diff
@@ -115,6 +122,79 @@ export function useProvider({
...
+      this.rethnetProcess.stderr.on("data", (data: any) => {
+        stderr += data.toString();
+        console.log(`rethnet subprocess ${this.rethnetProcess.pid}: ${data}`);
+      });
+
+      function outputForError() {
+        return `stdout:\n${stdout}\nstderr:\n${stderr}`;
+      }
+
+      const wait = new Promise((resolve) => setTimeout(resolve, 2000));
```
fvictorio, line 145: There is a sleep helper you can use, in the `sleep.ts` module in the same directory.

packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
```diff
@@ -115,6 +122,79 @@ export function useProvider({
...
+          } else {
+            reject(
+              new Error(`Rethnet subprocess error: ${err}. ${outputForError()}`)
+            );
+          }
+        });
+      });
+
+      // sleep a moment to let the server initialize
+      await new Promise((resolve) => setTimeout(resolve, 250));
```
fvictorio, line 179: Same comment about the sleep helper.

feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4095/commits/368202e32b4226e46905d22cb08899d4f2cbab97

packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4095/commits/368202e32b4226e46905d22cb08899d4f2cbab97

packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4095/commits/cb61fb95c774d3c7204a72d1de250a18d62c03f2

packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
fvictorio: Yes, Hardhat includes the hash of the transaction in the stack trace, but that only works for the in-process hardhat network IIRC. That's why those particular tests are skipped.

crates/rethnet_evm/src/state/hybrid.rs
```diff
@@ -355,3 +355,75 @@ impl StateHistory for HybridState<RethnetLayer> {
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use rethnet_eth::Bytes;
+
+    #[test]
+    fn repro_provider_test_should_result_in_a_modified_balance() {
```
Wodann, line 366:
```suggestion
    fn test_irregular_state_modification() {
```

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -204,7 +211,7 @@ async fn handle_get_balance(
     block: Option<BlockSpec>,
 ) -> ResponseData<U256WithoutLeadingZeroes> {
     event!(Level::INFO, "eth_getBalance({address:?}, {block:?})");
-    match set_block_context(&state, block).await {
+    match set_block_context(&state, block.clone()).await {
```
Wodann, line 214: What made the clone necessary?

packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
```diff
@@ -194,3 +194,21 @@ if (INFURA_URL !== undefined) {
     },
   });
 }
+
+if (RETHNET_BINARY !== undefined) {
+  PROVIDERS.push({
+    name: "Rethnet",
+    isFork: false,
```
Wodann, line 201: @fvictorio should we add two prodivers? One with fork and one without?

packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
```diff
@@ -115,6 +123,77 @@ export function useProvider({
         this.server.getProvider()
       );
     }
+
+    if (rethnetBinary !== undefined) {
```
Wodann, line 127: Is there a reason why you didn't put this code in a dedicated class that gets constructed there and can also be reused elsewhere?

Wodann (changes requested):

crates/rethnet_eth/src/remote.rs
```diff
@@ -104,6 +119,17 @@ impl BlockSpec {
     }
 }
 
+impl Display for BlockSpec {
+    fn fmt(&self, formatter: &mut Formatter) -> Result<(), fmt::Error> {
+        let s = match self {
+            BlockSpec::Number(n) => n.to_string(),
+            BlockSpec::Tag(t) => t.to_string(),
+            BlockSpec::Eip1898(e) => serde_json::to_string(e).map_err(|_| fmt::Error)?,
```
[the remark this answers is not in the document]
feuGeneA: It felt appropriate since EIP-1898 specifies everything in JSON. `e.fmt(formatter)` would require `impl Display for Eip1898BlockSpec`, right? Do you think we should do a custom formatting besides just JSON? If so, what do you think that should look like?

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: I suspect that at some point I was doing some debug logging in here that required this clone, but clearly it's unnecessary at this point so i removed it in https://github.com/NomicFoundation/hardhat/pull/4095/commits/be61d53fe80b7aafb3484d84a8c4eb6631f87796

packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
feuGeneA: Only that I didn't foresee any specific other usage for such an abstraction. Usually I avoid preparing for unforeseen re-use as a premature optimization.

Wodann: Leaving aside other use cases, how would you ensure encapsulation of your code in that case? (https://en.wikipedia.org/wiki/Single-responsibility_principle)

crates/rethnet_rpc_server/src/lib.rs
Wodann: Did clippy not raise an error about this?

feuGeneA: Nope, it ran cleanly both locally and in CI.

crates/rethnet_eth/src/remote.rs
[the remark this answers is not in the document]
Wodann: This was implied, but my comment requires that you implement `trait Display` for `Eip1898BlockSpec`.

My question specifically boils down to: Why is the struct itself not responsible for defining its output format?

This again comes down to single-responsibility principle: every unit is responsible for their own behaviour and it should be encapsulated within.

packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4095/commits/dfb80375229c0901963db000953f65fd39c3ce5e

crates/rethnet_eth/src/remote.rs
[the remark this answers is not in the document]
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4095/commits/5ff2c6a9b46269161e97f44abd1016ecc4fbfd1b

packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4095/commits/50378a0b033586fe478d4ed35d91a5148566153f

packages/hardhat-core/test/internal/hardhat-network/helpers/providers.ts
fvictorio: `isFork` means mainnet providers, so in the case of the rethnet provider is probably not useful. The rethnet provider is more akin to the Hardhat Node provider that is started in background.

packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
```diff
@@ -115,6 +119,20 @@ export function useProvider({
...
+      const { childProcess, processExit, httpProvider } =
```
fvictorio, line 124: I would rename `processExit` to `isReady`, and document that it's a promise that resolves if the process doesn't fail in N milliseconds.

Wodann (commented):
No more comments on my part. Ceding to Franco as he still had some unresolved comments

packages/hardhat-core/test/internal/hardhat-network/helpers/useProvider.ts
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4095/commits/dda3c487a083ea2a719b430a7c09be9596ded35f

fvictorio (approved):

PR #4096, opened 2023-07-01
Add config validation for empty network and forking url

### LINKED ISSUE #2730: Better error message when URL is an empty string

kshyun28:

---

Considering the discussions in the related issue, this PR adds empty string validation such as:
- empty strings `""`
- empty strings with at least 1 whitespace `" "`

for the following configs:
- [CUSTOM_NETWORK].url
- [HARDHAT_NETWORK_NAME].forking.url

To improve the error message, instead of using the default `invalid` error message, I also added an `empty` error message. This can be up for debate since we can just use `invalid` error message which makes the code simpler. 

This closes #2730 

Screenshot of sample config error
Hardhat empty url config error

### DIFF
--- a/packages/hardhat-core/src/internal/core/config/config-validation.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-validation.ts
@@ -401,6 +401,18 @@ export function getValidationErrors(config: any): string[] {
         );
       }
 
+      if (hardhatNetwork.forking !== undefined) {
+        if (isEmptyString(hardhatNetwork.forking.url)) {
+          errors.push(
+            getEmptyErrorMessage(
+              `HardhatConfig.networks.${HARDHAT_NETWORK_NAME}.forking.url`,
+              hardhatNetwork.forking.url,
+              "string"
+            )
+          );
+        }
+      }
+
       // Validating the accounts with io-ts leads to very confusing errors messages
       const { accounts, ...configExceptAccounts } = hardhatNetwork;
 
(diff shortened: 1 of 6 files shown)


packages/hardhat-core/src/internal/core/config/config-validation.ts
```diff
@@ -51,6 +51,12 @@ function getMessage(e: ValidationError): string {
       );
 }
 
+function getEmptyErrorMessage(path: string, value: any, expectedType: string) {
+  return `Empty value ${stringify(
+    value
+  )} for ${path} - Expected a non-empty value of type ${expectedType}.`;
+}
+
```
fvictorio, line 59: I think I would rename this to `getEmptyStringErrorMessage`, drop the `expectedType` parameter, and use a more specific error message like:

```
Empty string ${stringify(value)} for ${path} - Expected a non-empty string.
```

Wdyt?

kshyun28:
Hello @fvictorio, I understand that if this implementation of the fix would break some existing projects, better not go with this approach, and no worries, it's not really discarding since the issue is moving and allows us to see potential issues. 

I'll try to reimplement the fixes in the `HttpProvider`. Actually that was the first thing that came to mind for me, but considered the config validation approach. At least we now know that config validation won't work. :)

PR #4100, opened 2023-07-03
[Docs]: Add Tenderly to Forking other networks page

dzimiks:

---

## Description

Updated [Forking other networks page](https://hardhat.org/hardhat-network/docs/guides/forking-other-networks) with the code snippets using Tenderly node.


fvictorio:
Hey @dzimiks, thanks for this but I think we'll keep the list of providers as is to make things more concise, otherwise the docs will get noisy if we add every service out there.

Issue #841, opened 2020-10-02
Lost of precision when using %d in console.log


alcuadrado:
If a user logs a number using "%d" in `console.log` they will see an incorrect result. This is because EVM numbers are represented using BN.js, and casting those to `number` looses precision.

We should somehow detect this case and turn those numbers into strings.

alcuadrado:
This is a pretty minimal and isolated reproduction script:
```
const util = require("util");
const BN = require("BN.js");

util.format("%d", (new BN(10).pow(new BN(18)).subn(1)))
```

The expected result is `"999999999999999999"`, but we get `"1000000000000000000"`.

Note that this doesn't fix it, `util.format("%d", (new BN(10).pow(new BN(18)).subn(1)).toString())`, as the string is also coerced into a `number`.

PR #4115, opened 2023-07-05
Loss of precision when using %d and %i in console.log

ChristopherDedominici:
Very large number lose precision when they are console logged using %d or %i, [see issue here](https://github.com/NomicFoundation/hardhat/issues/841).

Solution: convert %d and %i into %s so even very large numbers can be logged without losing precision.


schaable (approved):
LGTM!

packages/hardhat-core/src/internal/hardhat-network/stack-traces/consoleLogger.ts
```diff
@@ -134,7 +134,24 @@ export class ConsoleLogger {
...
+  }
+
+  private _replaceNumberFormatSpecifiers(consoleLogs: ConsoleLogs) {
+    // Replace the occurrences of %d and %i with %s. This is necessary because if the arguments passed are numbers,
+    // they could be too large to be formatted as a Number or an Integer, so it is safer to use a String.
+    // %d and %i are replaced only if there is an odd number of % before the d or i.
+    // If there is an even number of % then it is assumed that the % is escaped and should not be replaced.
+    if (consoleLogs.length > 0) {
+      consoleLogs[0] = String(consoleLogs[0]).replace(
+        /((?<!%)(?:%%)*)(%[di])/g,
```
alcuadrado, line 151: I used to think I was good with regexes 😅

packages/hardhat-core/src/internal/hardhat-network/stack-traces/consoleLogger.ts
```diff
@@ -134,7 +134,24 @@ export class ConsoleLogger {
...
+    return consoleLogs;
+  }
+
+  private _replaceNumberFormatSpecifiers(consoleLogs: ConsoleLogs) {
+    // Replace the occurrences of %d and %i with %s. This is necessary because if the arguments passed are numbers,
+    // they could be too large to be formatted as a Number or an Integer, so it is safer to use a String.
+    // %d and %i are replaced only if there is an odd number of % before the d or i.
+    // If there is an even number of % then it is assumed that the % is escaped and should not be replaced.
+    if (consoleLogs.length > 0) {
+      consoleLogs[0] = String(consoleLogs[0]).replace(
```
alcuadrado, line 150: I always find this `ConsoleLogs` type confusing. Cant's `consoleLogs[0]` be an array?

Or is it safe to cast to string because if the length >0 we are formatting, and hence the first one must be a string?

ChristopherDedominici: I spent some time on it and I think no, but correct me if I'm wrong because maybe I missed something.

My  findings: the only way that console.log can be used in a .sol file are these:

- 1 argument passed: console.log(123);
- 2 arguments passed: console.log("%d", 123456789123456789123456789);
- more than 2 arguments passed: console.log("%d %%d %%%d %%%%d %%%%%d %%%%%%d", 1, 2, 3);

In all the scenarios the consoleLogs[0]  it's always a single value, not an array, even if multiple console.logs are called at the same time, [see test here](https://github.com/NomicFoundation/hardhat/pull/4115/files/923b72ddceee4cc3e45602a779be78576b883b83#diff-167324083ac14fceebeee729d324066ff06f0eb908f1e6d1cdd25edf2796cd3d).

ChristopherDedominici: done

ChristopherDedominici: I think it will always be a string, because the function **_decode**, where I take the consoleLogs, it returns a variable that is of type ConsoleLogs.
ConsoleLogs is this:

image

So string or array of strings ("in a recursive way").

So even if I remove the "String(consoleLogs[0])" in my code to don't force the first element to be a string and I run a test with console.log(1) I still get a "1" instead of a 1.

But yours it's still a good and valid observation, to be 100% safe I could add the check like this:

if (consoleLogs.length > 0 && typeof consoleLogs[0] === "string") {

fvictorio: +1 to adding that check, I came here to suggest exactly that before reading this conversation.

fvictorio: I agree that the first element is always a string right now, but that could change in the future.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/consoleLogger.ts
fvictorio: I think we should add an explanatory comment. I wrote this, in part to validate my understanding of what's going on here:

```
        /**
         * This regex matches a '%d' or an '%i' that has an even number of
         * '%' behind it (including 0). This group of pairs of '%' is captured
         * and preserved, while the '%[di]' is replaced with '%s'.
         * Naively doing (%%)* doesn't work; we also have to use the
         * (?<!%) negative look-behind to make this work.
         * The (?:) is just to avoid capturing that inner group.
         */
```

fvictorio: (I wanted to include the full explanation of why the negative look-behind is not necessary but that will make the comment even longer and it will be too noisy. Anyone landing here will have the elements to figure it out.)

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
```diff
@@ -0,0 +1,73 @@
+import { assert } from "chai";
```
fvictorio, line 1: Personal preference, feel free to ignore, but I'd rather name this file `hardhat-network/console-log.ts` than use a directory with a single file. Super weak preference though.

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
```diff
@@ -0,0 +1,73 @@
...
+  // Set up the test environment
+  useFixtureProject("console-log");
+  useEnvironment();
+
+  it("should print all the numbers without losing precision", async function () {
+    await this.env.run(TASK_COMPILE, { quiet: true });
+
+    // Retrieve the artifact of the solidity file compiled in the previous command
+    const artifacts = new Artifacts(ARTIFACTS_FOLDER_PATH);
+    const artifact = artifacts.readArtifactSync("Test");
```
fvictorio, line 30: You can drop the previous line and the `ARTIFACTS_FOLDER_PATH` variable, and to this instead:

```suggestion
    const artifact = this.env.artifacts.readArtifactSync("Test");
```

The environment is created with an `artifacts` field that is initialized with the right directory. Using this is both more concise and closer to what's going on in a real project.

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
```diff
@@ -0,0 +1,73 @@
...
+    // Modify console.log to store the messages that are gonna be printed when executing the smart contract function
+    const originalConsoleLog = console.log;
+    const capturedLogs: string[] = [];
+    console.log = (s: string) => {
+      capturedLogs.push(s);
+    };
+
+    // Call the contract function
+    await this.env.network.provider.send("eth_sendTransaction", [
+      { from: deployer, to: receipt.contractAddress, data: "0xf8a8fd6d" },
```
fvictorio, line 51:
```suggestion
      {
        from: deployer,
        to: receipt.contractAddress,
        // selector of 'test()'
        data: "0xf8a8fd6d",
      },
```

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
```diff
@@ -0,0 +1,73 @@
...
+    ]);
+
+    // Restore the original console.log
+    console.log = originalConsoleLog;
+
+    // Process the captured logs as needed
+    assert.equal("123456789123456789123456789", capturedLogs[0]);
```
fvictorio, line 58: I think we should also assert the length of `capturedLogs` here.

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
```diff
@@ -0,0 +1,73 @@
...
+    assert.equal("123456789123456789123456789", capturedLogs[1]);
+    assert.equal("121314", capturedLogs[2]);
```
fvictorio, line 60: For these tests to be 100% valid, I think we should use unsafely large values in all these scenarios. Otherwise we don't really know that the right modifier transformation is being done under the hood.

ChristopherDedominici: Good point, I'll change all of them but I'll still keep 2 "shorter" numbers, one for %d and %i. Just to also be sure that everything works fine with "shorter" numbers (it should always be the case, but just to be safe).

packages/hardhat-core/src/internal/hardhat-network/stack-traces/consoleLogger.ts
ChristopherDedominici: Done

packages/hardhat-core/src/internal/hardhat-network/stack-traces/consoleLogger.ts
ChristopherDedominici: Done

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
ChristopherDedominici: Done

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
ChristopherDedominici: Done

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
ChristopherDedominici: Done

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
ChristopherDedominici: Done

packages/hardhat-core/test/internal/hardhat-network/console-log/test.ts
ChristopherDedominici: Done

packages/hardhat-core/test/internal/hardhat-network/console-log.ts
```diff
@@ -0,0 +1,77 @@
...
+    console.log = (v: any) => {
+      capturedLogs.push(v);
+    };
+
+    // Call the contract function
+    await this.env.network.provider.send("eth_sendTransaction", [
+      {
+        from: deployer,
+        to: receipt.contractAddress,
+        data: "0xf8a8fd6d",
```
fvictorio, line 46: I still think we should include this comment, or assign the string to a descriptive variable.

```suggestion
        // selector of 'test()'
        data: "0xf8a8fd6d",
```

fvictorio (approved):
Pre-approving, just a minor comment. Feel free to merge.

packages/hardhat-core/test/internal/hardhat-network/console-log.ts
ChristopherDedominici: Done

PR #4128, opened 2023-07-10
Fix deletion of valid build-info file

### LINKED ISSUE #1117: hardhat-etherscan breaks on hardhat/console.sol

schaable:
Bug reproduction steps:

1. Create a contract that imports a dependency (can be local or external).
2. Compile the project.
3. Rename the contract, only changing the contract name (filename remains the same).
4. Compile again without using `--force` or `hh clean`.
5. Go to the artifacts folder. The debug file of the renamed contract will reference a valid build-info file. However, the imported dependency's debug file will point to a missing build-info file.

This issue occurs because the renamed contract is recompiled while the imported contract remains unchanged (which is correct). However, during the artifact cleanup, the original build-info file, which is still pointed by the imported contract, is mistakenly deleted. Two possible fixes are:

1. Ensure that every debug file points to the correct build-info file by adjusting the debug file pointers.
2. **Modify the cleanup process to delete build-info files only if they are not associated with any debug files. Although this fix was partially implemented, the build-info files were unintentionally deleted along with the invalid artifacts in `_removeArtifactFiles`.**

Closes #1117

### DIFF
--- a/packages/hardhat-core/test/builtin-tasks/compile.ts
+++ b/packages/hardhat-core/test/builtin-tasks/compile.ts
@@ -852,10 +852,10 @@ Read about compiler configuration at https://hardhat.org/config
     it("should not remove the build-info if it is still referenced by an external library", async function () {
       await this.env.run("compile");
 
-      const pathContractA = path.join("contracts", "A.sol");
-      let contractA = fsExtra.readFileSync(pathContractA, "utf-8");
+      const pathToContractA = path.join("contracts", "A.sol");
+      let contractA = fsExtra.readFileSync(pathToContractA, "utf-8");
       contractA = contractA.replace("contract A", "contract B");
-      fsExtra.writeFileSync(pathContractA, contractA, "utf-8");
+      fsExtra.writeFileSync(pathToContractA, contractA, "utf-8");
 
       /**
        * The _validArtifacts variable is not cleared when running the compile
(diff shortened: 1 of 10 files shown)


packages/hardhat-core/test/builtin-tasks/compile.ts
```diff
@@ -844,4 +844,96 @@ Read about compiler configuration at https://hardhat.org/config
...
+      const buildInfoPathE = path.join(
+        "artifacts",
+        "contracts",
+        "E.sol",
+        "E.dbg.json"
+      );
+      assertBuildInfoExists(buildInfoPathE);
+    });
+
+    afterEach(() => {
```
fvictorio, line 927: I think I'd slightly prefer to do this in the tests themselves, because they are the ones doing this change so it makes sense that they also do the cleanup. An `afterEach` that performs a cleanup of something that wasn't done in a `beforeEach` feels a bit weird to me.

I understand that this means the files won't be reverted if the test fails, but I'm ok with that trade-off.

fvictorio (approved):
Pre-approved because the comments I left are very minor. Feel free to ignore the one about the `afterEach` if you don't agree. The other one about removing the unnecessary import seems a bit more important.

packages/hardhat-core/test/builtin-tasks/compile.ts
schaable: Yeah, my main worry was something failing during the test and breaking future runs. But it's ok if you're good with the trade-off.

packages/hardhat-core/test/fixture-projects/compilation-contract-with-deps/contracts/E.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.19;
+
+import "dependency/contracts/console.sol";
```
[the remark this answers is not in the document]
schaable: Good catch.

PR #4131, opened 2023-07-11
Negative block number queried when forked network block number is near zero

ChristopherDedominici:
The code line: `const lastSafeBlock = latestBlock - maxReorg;` could result in a negative value when maxReorg > latestBlock.
Fix: add check, when the difference is a negative number then use latestBlock.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
@@ -44,7 +44,7 @@ export async function makeForkClient(
   const maxReorg = actualMaxReorg ?? FALLBACK_MAX_REORG;
 
   const latestBlock = await getLatestBlockNumber(provider);
-  const lastSafeBlock = latestBlock - maxReorg;
+  const lastSafeBlock = getLastSafeBlock(latestBlock, maxReorg);
 
   let forkBlockNumber;
   if (forkConfig.blockNumber !== undefined) {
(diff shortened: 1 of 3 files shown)


fvictorio:
@ChristopherDedominici can you add a changeset?

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
```diff
@@ -127,3 +127,11 @@ async function getLatestBlockNumber(provider: HttpProvider) {
   const latestBlock = BigInt(latestBlockString);
   return latestBlock;
 }
+
+export function getLastSafeBlock(
+  latestBlock: bigint,
+  maxReorg: bigint
+): bigint {
+  // Design choice: if latestBlock - maxReorg gives a negative number than the newest block will be used
```
fvictorio, line 135: Let's explain that we do this because the forked network is almost surely a devnet

packages/hardhat-core/test/internal/hardhat-network/provider/fork/ForkBlockchain.ts
```diff
@@ -66,6 +69,18 @@ describe("ForkBlockchain", () => {
     assert.instanceOf(fb, ForkBlockchain);
   });
 
+  // TODO: remove regex code
```
fvictorio, line 72: What does this comment mean?

ChristopherDedominici: Ugh, nothing, I forgot to cancel it. I'll remove it

packages/hardhat-core/src/internal/hardhat-network/provider/utils/makeForkClient.ts
ChristopherDedominici: Done

fvictorio (approved):

PR #4139, opened 2023-07-12
debug_traceTransaction ignores "tracer" param

ChristopherDedominici:
When calling `debug_traceTransaction` passing a `tracer` parameter that is not supported an error should be thrown.  [See issue here](https://github.com/NomicFoundation/hardhat/issues/3973).

### DIFF
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/input/debugTraceTransaction.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/input/debugTraceTransaction.ts
@@ -5,6 +5,8 @@ import { optionalOrNullable } from "../../../../util/io-ts";
 export const rpcDebugTracingConfig = optionalOrNullable(
   t.type(
     {
+      tracer: optionalOrNullable(t.string),
+      callTracer: optionalOrNullable(t.boolean),
       disableStorage: optionalOrNullable(t.boolean),
       disableMemory: optionalOrNullable(t.boolean),
       disableStack: optionalOrNullable(t.boolean),
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/input/debugTraceTransaction.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/input/debugTraceTransaction.ts
@@ -6,7 +6,6 @@ export const rpcDebugTracingConfig = optionalOrNullable(
   t.type(
     {
       tracer: optionalOrNullable(t.string),
-      callTracer: optionalOrNullable(t.boolean),
       disableStorage: optionalOrNullable(t.boolean),
       disableMemory: optionalOrNullable(t.boolean),
       disableStack: optionalOrNullable(t.boolean),
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/modules/debug.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/modules/debug.ts
@@ -4,13 +4,18 @@ import {
   RpcDebugTracingConfig,
 } from "../../../core/jsonrpc/types/input/debugTraceTransaction";
 import { validateParams } from "../../../core/jsonrpc/types/input/validation";
-import { MethodNotFoundError } from "../../../core/providers/errors";
+import {
+  InvalidArgumentsError,
+  MethodNotFoundError,
+} from "../../../core/providers/errors";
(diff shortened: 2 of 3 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/modules/debug.ts
```diff
@@ -32,7 +37,31 @@ export class DebugModule {
...
+      rpcHash,
+      rpcDebugTracingConfig
+    );
+
+    this._throwErrIfTracerIsNotSupported(validatedParams[1]);
+
+    return validatedParams;
+  }
+
+  private _throwErrIfTracerIsNotSupported(config: RpcDebugTracingConfig) {
```
schaable, line 51: Nitpicking, but I would name the function according to its purpose, maybe something like `_validateTracer`.

schaable (approved):

packages/hardhat-core/src/internal/core/jsonrpc/types/input/debugTraceTransaction.ts
```diff
@@ -5,6 +5,8 @@ import { optionalOrNullable } from "../../../../util/io-ts";
 export const rpcDebugTracingConfig = optionalOrNullable(
   t.type(
     {
+      tracer: optionalOrNullable(t.string),
+      callTracer: optionalOrNullable(t.boolean),
```
fvictorio, line 9: I don't hink there's a `callTracer` option?

You can see the supported options [here](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-debug#debugtracetransaction).

packages/hardhat-core/src/internal/hardhat-network/provider/modules/debug.ts
```diff
@@ -4,13 +4,18 @@ import {
...
+  InvalidArgumentsError,
+  MethodNotFoundError,
+} from "../../../core/providers/errors";
 import { HardhatNode } from "../node";
 import { RpcDebugTraceOutput } from "../output";
 
 /* eslint-disable @nomicfoundation/hardhat-internal-rules/only-hardhat-error */
 
 export class DebugModule {
+  public static readonly supportedTracers = ["structLogs"];
```
fvictorio, line 17: I don't think there's a `structLogs` value for the tracer option in geth.

I will look into this more in depth tomorrow.

packages/hardhat-core/src/internal/hardhat-network/provider/modules/debug.ts
```diff
@@ -32,7 +37,31 @@ export class DebugModule {
...
+
+    this._validateTracerParam(validatedParams[1]);
+
+    return validatedParams;
+  }
+
+  private _validateTracerParam(config: RpcDebugTracingConfig) {
+    if (
+      config !== undefined &&
+      config.tracer !== undefined &&
```
fvictorio, line 54:
```suggestion
      config?.tracer !== undefined &&
```

fvictorio:
@ChristopherDedominici can you add a changeset for this change

packages/hardhat-core/src/internal/hardhat-network/provider/modules/debug.ts
ChristopherDedominici: Done

fvictorio:
I took another look at this and I think this approach won't work, because geth's default tracer doesn't have a name :disappointed: That is, there's no value that you can use in `{ tracer: XXX }` to get the same behavior that you get when you omit the `tracer` key. `callTracer`, the one you are using, does a different thing from the opcode logger.

Since we only implement the default tracer for now, the correct thing to do would be to set `supportedTracers` to `[]`, which would result in a very confusing error message. So I think the right thing to do is to just show a specific error when the `tracer` option is passed, saying something like "Hardhat only supports the default tracer at the moment".

.changeset/three-maps-judge.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Added logic to throw an error when the debug_traceTransaction method is called with a tracer parameter that is not supported.
```
fvictorio, line 5:
```suggestion
Added logic to throw an error when the debug_traceTransaction method is called with a tracer parameter.
```

fvictorio (approved):
Pre-approving

PR #4140, opened 2023-07-12
feat: add net_version JSON-RPC API

Wodann:


crates/rethnet_eth/src/remote/methods.rs
```diff
@@ -155,6 +155,9 @@ pub enum MethodInvocation {
     /// eth_mining
     #[serde(rename = "eth_mining")]
     Mining(),
+    /// net_version
+    #[serde(rename = "net_version")]
+    NetworkId(),
```
feuGeneA, line 160: Why not `NetVersion` instead of `NetworkId`? I can see some logic in naming the variant based on what it returns, but so far the convention among the other variants is to name them based on the name of the method you're invoking (keeping with the concept of the enum, `MethodInvocation`).

Also, this is the first method that doesn't start with `eth_`, and I had been thinking about how when such a method was introduced, we should probably add an `Eth` prefix to all the other methods.
```suggestion
    /// net_version
    #[serde(rename = "net_version")]
    NetVersion(),
```

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -415,6 +415,11 @@ impl RpcClient {
         self.call(&MethodInvocation::GetStorageAt(*address, position, block))
             .await
     }
+
+    /// Calls `net_version` API
+    pub async fn network_id(&self) -> Result<U256, RpcClientError> {
```
feuGeneA, line 420:
```suggestion
    pub async fn net_version(&self) -> Result<U256, RpcClientError> {
```

So far here the convention has been to name these associated methods according to the method being invoked (at least for the methods that are really just wrappers around RPC methods (`get_account_info` being the exception)).

feuGeneA (changes requested):

crates/rethnet_eth/src/remote/methods.rs
Wodann: The reason for naming it `NetworkId` is to keep consistency with our codebase. Within our code, we call that concept `NetworkId`. The JSON-RPC spec can't change once a method has been named a certain way, but `serde` allows us to decouple the concepts

Internal naming reference: https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/hardhat-network/jsonrpc/client.ts#L37

feuGeneA: I think that's a solid argument for keeping the client method named the way you did it, but I still think the `MethodInvocation` variant should be named based on the method being invoked.

Wodann: If you prefix or suffix everything with `Eth`, I expect clippy to complain to you about adding unnecessary characters that are repeated. The goal is to keep the name as simple as possible while conveying still conveying identifying information.

reference: https://rust-lang.github.io/rust-clippy/master/index.html#/enum_variant_names

feuGeneA: I wouldn't prefix EVERYTHING with `Eth`, only the variants that represent methods that start with `eth_`. This variant in question would be prefixed with `Net`, the `web_*` methods would be prefixed with `Web3`, etc

feuGeneA: > like here: https://github.com/foundry-rs/foundry/blob/044f680a5c2541b0fdbd69eb8f33abd57e161569/anvil/core/src/eth/mod.rs#L47

I'm not sure why there they named this variant `EthNetworkId`. it seems to be corresponding to a method `eth_networkId` mentioned there, but that method is not part of the JSON-RPC spec.

Wodann: I'm happy to rename this to `NetVersion` and leave the client method as `network_id`

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -415,6 +415,11 @@ impl RpcClient {
...
+        self.call(&MethodInvocation::NetworkId()).await
```
Wodann, line 421:
```suggestion
        self.call(&MethodInvocation::NetVersion()).await
```

crates/rethnet_eth/src/remote/methods.rs
feuGeneA: perfect! thanks for hearing me out!

PR #4141, opened 2023-07-12
chore: ensure CI runs  bench targets

Wodann:


.github/workflows/hardhat-core-ci.yml
```diff
@@ -56,13 +56,13 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --workspace --all-features --all-targets
+          args: --workspace --all-features --all-targets --profile bench
```
feuGeneA, line 59: I think this means that after this change we're _not_ doing `cargo check --profile dev`, which we _were_ doing before. Is that okay?

image
https://doc.rust-lang.org/cargo/reference/profiles.html#profile-selection

feuGeneA (approved):

.github/workflows/hardhat-core-ci.yml
Wodann: Bench is the most complete profile for this case, if I understand correctly. I could be wrong though:
- test: extends dev but runs #[test]
- bench: extends release but runs #[test] and #[bench]
- dev: unoptimised code
- release: optimised code

Issue #2540, opened 2022-03-28
[hardhat-vyper] Warn when test mode directive is present in contract sources


alcuadrado:
See https://github.com/NomicFoundation/hardhat/issues/2534

We should warn the users during compilation if something like that is present.

feuGeneA:
throw, don't just warn. let there be false positives and we'll deal with that when it actually happens.

PR #4148, opened 2023-07-13
[hardhat-vyper] Warn when test mode directive is present in contract sources

ChristopherDedominici:
Our vyper plugin only compiles things in test mode, which is dangerous, as it can leave things like this in production deployments.
[See issue here](https://github.com/NomicFoundation/hardhat/issues/2540).

Example:
```
#@ if mode == "test":
@external
def _mint_for_testing(_to: address, _token_id: uint256):
    self._mint(_to, _token_id)
#@ endif
```

Solution: 
throw an error explaining that we currently do not support this feature if a test directive is found in the code.

### DIFF
--- a/packages/hardhat-vyper/README.md
+++ b/packages/hardhat-vyper/README.md
@@ -74,7 +74,7 @@ There are no additional steps you need to take for this plugin to work.
 
 ## Test directives
 
-Brownie allows you to use the test directive `#@ if mode == "test":` to specify when a portion of code should be compiled only for testing purposes.
+Brownie allows you to use the test directive `#@ if mode == "test":` to specify when a portion of code should be included only for testing purposes.
 
 Example:
 
--- a/packages/hardhat-vyper/README.md
+++ b/packages/hardhat-vyper/README.md
@@ -86,7 +86,7 @@ def _mint_for_testing(_to: address, _token_id: uint256):
 #@ endif
 ```
 
-We currently do NOT support this feature. An error will be thrown every time that, when compiling a contract, the directive `#@ if mode == "test":` is found.
+We do NOT support this feature. An error will be thrown every time that, when compiling a contract, the directive `#@ if mode == "test":` is found.
 
 ### Additional notes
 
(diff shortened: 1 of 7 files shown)


schaable (approved):

packages/hardhat-vyper/README.md
```diff
@@ -72,6 +72,22 @@ module.exports = {
 
 There are no additional steps you need to take for this plugin to work.
 
+## Test directives
+
+Brownie allows you to use the test directive `#@ if mode == "test":` to specify when a portion of code should be compiled only for testing purposes.
```
fvictorio, line 77:
```suggestion
Brownie allows you to use the test directive `#@ if mode == "test":` to specify when a portion of code should be included only for testing purposes.
```

packages/hardhat-vyper/src/parser.ts
```diff
@@ -31,6 +33,16 @@ export class Parser {
     return result;
   }
 
+  private _validateTestModeNotUsed(fileContent: string, absolutePath: string) {
+    if (fileContent.includes('#@ if mode == "test":')) {
+      throw new Error(
```
fvictorio, line 38: Let's throw a `VyperPluginError` instead. This error extends `NomicLabsHardhatPluginError`, which is then caught and handled by the Hardhat CLI.

packages/hardhat-vyper/src/parser.ts
```diff
@@ -31,6 +33,16 @@ export class Parser {
...
+        `We found a test directive in the file at path ${absolutePath}.` +
+          ` Test directives are a Brownie feature not supported by Hardhat.` +
+          ` Learn more at https://hardhat.org/hardhat-runner/plugins/nomiclabs-hardhat-vyper#test-directives.`
```
fvictorio, line 41: Let's remove the dot just in case, maybe it messes with the link in some terminals

```suggestion
          ` Learn more at https://hardhat.org/hardhat-runner/plugins/nomiclabs-hardhat-vyper#test-directives`
```

fvictorio:
@ChristopherDedominici this does need a changeset, otherwise it won't be included in the next release.

packages/hardhat-vyper/src/parser.ts
ChristopherDedominici: done

PR #4150, opened 2023-07-14
Test slow-imports rule

fvictorio:
Add a workflow that runs a simple smoke test to validate that our slow-imports eslint rule works as expected.

This will run once per day. The alternative was to try to figure out which file changes could break this (probably any change in the rules source and any change in `yarn.lock`), but I think once per day is simpler and good enough.

### DIFF
--- a/.github/workflows/test-slow-imports-rule.yml
+++ b/.github/workflows/test-slow-imports-rule.yml
@@ -1,0 +1,39 @@
+# This workflow checks that our slow-imports rule works correctly.
+# This is done by intentionally breaking the rule, running the linter, and checking
+# that we got a non-zero exit code. It's not a completely fault-proof strategy,
+# but it's an useful smoke test.
+name: Slow imports rule check
+
+on:
+  schedule:
+    - cron: "0 * * * *"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-slow-imports-rule.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - uses: actions/checkout@v2
+      - name: Install
+        run: yarn
+      - name: Add slow import
+        run: echo 'import "lodash";' >> src/internal/constants.ts
+      - name: Build
```
schaable, line 32: Not that it does something wrong, but I don't think you need this step.

schaable (approved):

.github/workflows/test-slow-imports-rule.yml
fvictorio: You do because several of the rules we use from `typescript-eslint` need [type information](https://typescript-eslint.io/linting/typed-linting).

PR #4152, opened 2023-07-15
Don't run changesets check in changesets PR

fvictorio:
This will prevent the changesets check to run in PRs like [this one](https://github.com/NomicFoundation/hardhat/pull/3857).


alcuadrado:
Good catch!

ChristopherDedominici (approved):

Issue #4110, opened 2023-07-04
Consider dropping the `abort-controller` dependency


fvictorio:
I'm not really sure why we use this dependency. It might be because node v12 doesn't have a global `AbortController` class.

We should check if having the dependency is still necessary given we only support node >= 16.

orlandoortegajr:
Hi, @fvictorio. Could you assign this to me? It's my first time trying to contribute to this project, and I figured this is probably something I can tackle.

alcuadrado:
Go ahead, @orlandoortegajr! 😍

I think @fvictorio's reasoning is correct. When we added that we wanted to abort slow requests if we already finished running the main task. The only way we found that worked well with `node-fetch` (which we used to use) was that.

orlandoortegajr:
@alcuadrado I see that makes a lot sense, I just scanned node-fetch's documentation pretty quickly and saw that they recommend adding this dependency. I'll see what I can do, like @fvictorio mentioned in node 15 the AbortController was introduced and it looks like it might be a direct replacement shouldn't take long.

alcuadrado:
Fixed by @orlandoortegajr 🚀

PR #4155, opened 2023-07-16
removed abort-controller dependency from core

orlandoortegajr:

---

The solution to https://github.com/NomicFoundation/hardhat/issues/4110. Removed abort-controller dependency that was recommended by docs of the package node-fetch, which is no longer being used. Now core will use the native AbortController API included in node versions 15+, which is a direct replacement. No additional changes are needed (see https://developer.mozilla.org/en-US/docs/Web/API/AbortController).

### DIFF
--- a/.changeset/chilly-meals-enjoy.md
+++ b/.changeset/chilly-meals-enjoy.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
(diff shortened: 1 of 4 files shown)


alcuadrado (approved):
Wow, this was surprisingly simple! Thanks for sending the PR.

Will merge once the CI passes.

PR #4162, opened 2023-07-18
docs: add `cargo doc` to CI & git hook

feuGeneA:


.cargo-husky/hooks/pre-commit
```diff
@@ -20,3 +21,4 @@ cargo test --workspace --all-targets ${ALL_FEATURES}
 cargo test --doc --workspace ${ALL_FEATURES}
 cargo clippy --all --all-targets ${ALL_FEATURES} -- -D warnings
 cargo fmt --all -- --check
+cargo doc --workspace --no-deps ${ALL_FEATURES}
```
Wodann, line 24: Is there a benefit to running this?

I've noticed that the pre-commit hook is already taking a lot more time with the additions we've made over time. The pre-commit hook is not intended to be a catch-all; that's what CI is for. It should be a sanity check that's run before you push things upstream.

feuGeneA: I'm happy to defer to your opinion on this, but here are my thoughts.

To me it's important that we _do_ have a catch-all, that's identical to CI, that can easily be run locally, and I do think it should include this line. However, I too dislike the amount of time that the pre-commit hook is taking now, so I'm not sure that this hook is the right place for that local catch-all to live.

Maybe these things belong in a pre-push hook rather than a pre-commit one? I'm not sure whether that would satisfy you here, but for my personal workflow it feels more appropriate.

Or maybe we should just have a non-hook script that can be run locally on a manual basis? And maybe CI should just invoke that script itself, so that it can never be out of sync with the github workflow definition?

Wodann: The plan is to move to something similar to what Slang does. They run a script, both locally in CI to build all of these things. Previously they had a bash script, but it's not cross-platforms. They're moving to writing their code in `build.rs` files for reasons specific to their code generation setup.

We could do something similar for a CI build script. You could then run that locally `cargo run edr-ci`. This will be extremely costly though, as it will also contain cross-compilation going forward, so I'd still propose having a separate sanity check as a pre-commit hook.

No need to implement this now though. This will be one of our tasks as we start on the release workflow. If you want, I'd recommend creating a bash script for now. Run that when you want to, but try and keep the pre-commit hook minimal.

Wodann (approved):

.cargo-husky/hooks/pre-commit
```diff
@@ -5,6 +5,7 @@
 
 set -e
 
+export RUSTDOCFLAGS="-D warnings"
```
Wodann, line 8: Is this still necessary?

Wodann (approved):

PR #4172, opened 2023-07-19
feat: XDC and Apothem network in hardhat-verify

0xbeny:


schaable:
Hi @0xbeny, thanks you submitting this PR. Unfortunately, we are not adding new networks to the core of hardhat-verify anymore. Take a look at: https://github.com/NomicFoundation/hardhat/blob/7daf60e743da4e6b9c38384ae317b47735227767/packages/hardhat-verify/src/chain-config.ts#L290-L292

Issue #1499, opened 2021-06-01
Make `flatten` task smarter


fvictorio:
The `flatten` task only generates the dependency graph of the project's files and merges them into a single file (removing all the `import` statements). This works for some scenarios, but in others, it generates a file that cannot be compiled. Examples of this are having multiple SPDX comments in the same file, or having multiple `pragma abicoder`, even if they are compatible.

We want to make this task smarter so that it generates compilable files most of the time.

[This document](https://www.notion.so/nomiclabs/New-flatten-task-spec-44f2282a0f9c4386bfd2d6f254c96e1c) has a more in-depth explanation of what we want to do. At the end there is a list of test cases that should make the expected behavior clearer.

alcuadrado:
Hey, @nachomazzara and @boringcrypto, wdyt about this document?

boringcrypto:
This is great! The only thing I'd like to add would be a way to select where the output goes. I'm on windows, so I can only pipe *everything* to a file, including console.logs and stdout messages, etc.

fvictorio:
I think we can add a `--output` param

nachomazzara:
Hey all! thanks for taking this seriously. It has been a big pain for me (A paranoid man who likes to deploy contracts manually)

I like the proposal! let's make it happen 🚀

ihorbond:
@fvictorio was this ticket ever resolved? Came here because of duplicate licenses too

boringcrypto:
I've been using this: https://github.com/nomiclabs/hardhat/issues/1050#issuecomment-840444798 - hacky, but solved my problem

TrejGun:
guys please, this is really annoying

partylikeits1983:
quick and dirty way with python: 

`$ npx hardhat flatten contracts/bank.sol > contracts/flat.sol`

```
import os

bad_words = ['SPDX-License-Identifier', 'pragma']

with open('/home/ubuntu/Desktop/defiBank/contracts/flat.sol') as oldfile, open('/home/ubuntu/Desktop/defiBank/contracts/flat1.sol', 'w') as newfile:
    for line in oldfile:
        if not any(bad_word in line for bad_word in bad_words):
            newfile.write(line)
            
filename = "/home/ubuntu/Desktop/defiBank/contracts/flatten1.sol"

string = "// SPDX-License-Identifier: MIT \n pragma solidity ^0.8.0; \n"
            
def insert(originalfile,string):
    with open(originalfile,'r') as f:
[...]

PR #4173, opened 2023-07-19
Make flatten task smarter

ChristopherDedominici:
Link to issue: [see issue here](https://github.com/NomicFoundation/hardhat/issues/1499)

Make the **flatten task** smarter, it should be able to handle multiple _SPDX licenses_ and multiple _pragma abicoder directives_.

**SPDX licenses**
Add a single license at the top, merging all the licenses in the sources. Comment out all the existing licenses to make solc ignore them, but leaving them in place, to communicate the actual license of each part of the flattened output. 

**Pragma abicoder directives**
Example of possible scenarios:
1. One of the files has the abi v2 pragma (i.e. `pragma experimental ABIEncoderV2` or `pragma abicoder v2`) and another file has no abi related pragma.
2. One of the files has the abi v2 pragma and another one has the abi v1 pragma (i.e. `pragma abicoder v1`)
3. Multiple files have the same ABI pragma

Like with the licenses, we don't want to remove the ABI pragmas, but just comment them out instead, and add one to the top of the output.
In the case (1) and (2) above, there's no correct abi pragma that could be added, but we default to adding an abi v2 pragma. 
A warning to stderr is generated after that the flattened output is printed to stdout.
In the case (3) we would just add the repeated abi pragma at the top.


packages/hardhat-core/src/builtin-tasks/flatten.ts
```diff
@@ -93,9 +169,17 @@ subtask(
 
     const sortedFiles = getSortedFiles(dependencyGraph);
 
+    flattened += addLicensesHeader(sortedFiles);
+    flattened += addPragmaAbicoderDirectiveHeader(sortedFiles);
```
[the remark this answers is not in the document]
ChristopherDedominici: done

packages/hardhat-core/src/builtin-tasks/flatten.ts
```diff
@@ -15,6 +15,13 @@ import {
   TASK_FLATTEN_GET_FLATTENED_SOURCE,
 } from "./task-names";
 
+// Match every line where a SPDX license is defined. The first captured group is the license.
+const SPDX_LICENSES_REGEX =
+  /^(?:\/\/|\/\*) *SPDX-License-Identifier:\s*([\w\d._-]+) *(\*\/)?(\n|$)/gim;
```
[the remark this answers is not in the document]
ChristopherDedominici: Regex have been modified

packages/hardhat-core/src/builtin-tasks/flatten.ts
```diff
@@ -135,5 +258,27 @@ task(
...
+    const [flattenedFile, metadata] = await run(
+      TASK_FLATTEN_GET_FLATTENED_SOURCE_AND_METADATA,
+      { files }
+    );
+
+    console.log(flattenedFile);
+
+    // Warn the user when SPDX licenses or pragma abicoder directives are not specified
+    if (metadata?.filesWithoutLicenses.length > 0) {
+      console.warn(
```
fvictorio, line 270: This should use `chalk.yellow`, otherwise the warnings are hard to see.

packages/hardhat-core/src/builtin-tasks/flatten.ts
```diff
@@ -135,5 +258,27 @@ task(
...
-    console.log(await run(TASK_FLATTEN_GET_FLATTENED_SOURCE, { files }));
+    const [flattenedFile, metadata] = await run(
+      TASK_FLATTEN_GET_FLATTENED_SOURCE_AND_METADATA,
+      { files }
+    );
+
+    console.log(flattenedFile);
+
+    // Warn the user when SPDX licenses or pragma abicoder directives are not specified
+    if (metadata?.filesWithoutLicenses.length > 0) {
```
fvictorio, line 269: If I understand correctly, the `?` operator is here because `metadata` could be null. But we can just use an object where the files are empty arrays.

packages/hardhat-core/test/builtin-tasks/flatten.ts
```diff
@@ -105,4 +129,213 @@ describe("Flatten task", () => {
...
+      });
+    });
+
+    describe("Flatten files that contain SPDX licenses", () => {
+      describe("Files contain one single license per file", () => {
+        describe("Files contain same licenses", () => {
+          useFixtureProject("flatten-task/contracts-spdx-same-licenses");
+
+          it("should successfully flatten and compile the files", async function () {
+            const [flattenedFiles, _metadata] = await this.env.run(
```
fvictorio, line 152: Should we be asserting the content of the returned metadata in these tests?

packages/hardhat-core/test/fixture-projects/flatten-task/contracts-regex-spdx-licenses-and-pragma-directives/contracts/A.sol
```diff
@@ -0,0 +1,10 @@
+/* SPDX-License-Identifier: MIT2  this part should be ignored*/
+// SPDX-License-Identifier: Linux2-man-pages-1-para this part should be ignored
+
+// NOTE: leave on the same line, the regex matcher should be able to handle the scenario
```
fvictorio, line 4: Outdated comment.

packages/hardhat-core/src/builtin-tasks/flatten.ts
```diff
@@ -135,5 +258,27 @@ task(
...
+      console.warn(
+        `The following file(s) do NOT specify SPDX licenses: ${metadata.filesWithoutLicenses.join(
+          ", "
+        )}`
+      );
+    }
+
+    if (metadata?.filesWithoutPragmas.length > 0) {
```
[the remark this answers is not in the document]
ChristopherDedominici: done

packages/hardhat-core/src/builtin-tasks/flatten.ts
ChristopherDedominici: done

packages/hardhat-core/src/builtin-tasks/flatten.ts
ChristopherDedominici: The check was to avoid a null value because in this part of the code the value might be null:

image

But in this part of the code the value should be always populated so I can remove the check

packages/hardhat-core/test/builtin-tasks/flatten.ts
ChristopherDedominici: done

packages/hardhat-core/test/fixture-projects/flatten-task/contracts-regex-spdx-licenses-and-pragma-directives/contracts/A.sol
ChristopherDedominici: removed

packages/hardhat-core/src/builtin-tasks/flatten.ts
ChristopherDedominici: The check is needed because in this part of the code a null value could be returned:

image

packages/hardhat-core/test/builtin-tasks/flatten.ts
```diff
@@ -105,4 +129,213 @@ describe("Flatten task", () => {
...
+      });
+    });
+  });
+
+  describe("TASK_FLATTEN", () => {
+    let spyFunctionConsoleLog: SinonSpy;
+    let spyFunctionConsoleWarn: SinonSpy;
+
+    beforeEach(() => {
+      spyFunctionConsoleLog = sinon.spy(console, "log");
```
schaable, line 364: I think you could use stubs to avoid cluttering the console output.

packages/hardhat-core/src/builtin-tasks/flatten.ts
```diff
@@ -13,8 +13,16 @@ import {
   TASK_FLATTEN,
   TASK_FLATTEN_GET_DEPENDENCY_GRAPH,
   TASK_FLATTEN_GET_FLATTENED_SOURCE,
+  TASK_FLATTEN_GET_FLATTENED_SOURCE_AND_METADATA,
 } from "./task-names";
 
+// Match every group where a SPDX license is defined. The fourth captured group is the license.
+const SPDX_LICENSES_REGEX =
+  /^(\/\/|\/\*)(\s|\t)*SPDX-License-Identifier:(\s|\t)*([\w\d._-]+).*/gm;
```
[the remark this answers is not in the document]
ChristopherDedominici: Yes, in this specific case I want to avoid the new line, but for the other regex you're right

schaable (approved):

PR #4179, opened 2023-07-20
hardhat-verify changes for OpenZeppelin & Sourcify

schaable:
Summary:
 - Moved the plugin's code to the `./src/internal` path. This was omitted in the original implementation. Only the public "interface" was left under `./src`, which is: `index.ts`, `etherscan.ts`, and `types.ts`
 - Re-export `Etherscan` class from `./src/internal`.
 - Change the typescript configuration to emit files to the root of the plugin. This allows to use imports from the root, for example: `import { Etherscan } from @nomicfoundation/hardhat-verify/etherscan` instead of `import { Etherscan } from @nomicfoundation/hardhat-verify/src/etherscan`. Test files are emitted to `build-test`.
 - Created a `.eslintignore` file to avoid linting errors in the IDE in files that are not linted anyway by the `eslint` npm script. Specifically this annoying one at the start of the files: `Parsing error: ESLint was configured to run on <tsconfigRootDir>/packages/hardhat-verify/... using parserOptions.project: <tsconfigRootDir>/packages/hardhat-verify/tsconfig.json. However, that TSConfig does not include this file.`
 - Use `import type` when possible.
 - Make `getCurrentChainConfig` a static method of the `Etherscan` class. Also improve it's signature and error messages.
 - Simplify the `Etherscan` class constructor and add a `fromChainCoinfig` method.
 - Make the `verify` method signature more similar to the Etherscan API.
 - There were some places still referencing the old `hardhat-etherscan` plugin. Updated to `hardhat-verify`.
 - Add enabled field to etherscan & sourcify config.
 - Add toggle logic to `TASK_VERIFY_GET_VERIFICATION_SUBTASKS`.
 - Add more tests.
 - Output errors at the end of the verification process
 - Show sourcify warning after the etherscan validation
 - Move etherscan & sourcify subtasks to their own files
 - Rename resolve arguments taks & add sourcify placeholder


fvictorio:
Something that I don't quite like is that the sourcify warning is shown immediately, instead of being shown after the etherscan verification. I think showing it first will give it too much weight.

The problem is that showing it in the right position is a bit harder to do. The only reasonable way to do it that comes to mind is that, instead of printing the warning where we are doing it now,  we push a different `TASK_VERIFY_SOURCIFY_DISABLED_WARNING` subtask. I think this should work and it's not inconsistent with similar things we do in the compile task.

packages/hardhat-verify/src/index.ts
```diff
@@ -211,9 +216,42 @@ subtask(TASK_VERIFY_RESOLVE_ARGUMENTS)
...
+  async (_, { config }): Promise<string[]> => {
+    const verificationSubtasks = [];
+
+    if (config.etherscan.enabled) {
+      verificationSubtasks.push(TASK_VERIFY_ETHERSCAN);
+    }
+
+    if (config.sourcify.enabled) {
+      verificationSubtasks.push(TASK_VERIFY_SOURCIFY);
+    } else {
```
fvictorio, line 230: Shouldn't this be something like:

```js
} else if (userConfig.sourcify?.enabled !== undefined) {
```

packages/hardhat-verify/src/index.ts
```diff
@@ -39,27 +43,27 @@ import {
...
   ExtendedContractInformation,
   extractInferredContractInformation,
   extractMatchingContractInformation,
   getLibraryInformation,
   LibraryToAddress,
-} from "./solc/artifacts";
+} from "./internal/solc/artifacts";
 
-import "./type-extensions";
+import "./internal/type-extensions";
```
fvictorio, line 66: I wonder if we should move the etherscan/sourcify subtasks to their own module, and here we just add `import "./internal/tasks/etherscan";` and `import ./internal/tasks/sourcify";`. It's not something we are doing anywhere else, and it would make code navigation harder in certain scenarios, but maybe it's worth it? I'm on the fence, so up to you if you want to do it.

packages/hardhat-verify/src/index.ts
schaable: The `config.sourcify` key is initialized in the config extender. I don't think we need to use `userConfig` after that. Am I missing your point?

packages/hardhat-verify/src/internal/etherscan.ts
```diff
@@ -0,0 +1,251 @@
...
+  public isBytecodeMissingInNetworkError() {
+    return this.message.startsWith("Unable to locate ContractCode at");
+  }
+
+  public isOk() {
+    return this.status === 1;
+  }
+}
+
+function resolveApiKey(apiKey: ApiKey | undefined, network: string) {
```
ericglau, line 235: Can you expose this function as well (e.g. make this a `public static` function in the `Etherscan` class, or provide a way to get the `_apiKey` field from the `Etherscan` class with the resolved value)?

We would like to use the resolved API key to make some separate calls to the Etherscan API, outside of the Etherscan class.

schaable: Would it be enough if we make `_apiKey` public?

packages/hardhat-verify/src/index.ts
schaable: Self note: this is to distinguish between the default (`false` value, we show the warning) and the explicit `enabled: false` in the config (we do not show the warning).

packages/hardhat-verify/src/internal/etherscan.ts
ericglau: Yes, that would work.

schaable: We had the same thought 😄

schaable: By the way, I'm going to split this PR into two to avoid blocking you, since the Sourcify part will likely require a major release. You can continue reviewing [here](https://github.com/NomicFoundation/hardhat/pull/4198).

PR #4181, opened 2023-07-21
fix: test-remote feature flag

Wodann:


feuGeneA (commented):
For my own education if nothing else: why was this needed? what indication was there that this change should be made? It seems harmless, since CI and everything is running fine, but then again it was also running fine before this change (...right?) so I'm hesitant to Approve without understanding the context.

Wodann:
If I want to run my remote tests, I'd do `cargo t --all --all-targets --features test-remote`

However, because `serde` is not enabled it will never run tests in modules that require serde, which is practically all of them

feuGeneA (approved):

PR #4182, opened 2023-07-21
feat: add helper functions for determining spec

Wodann:


crates/rethnet_eth/src/spec.rs
```diff
@@ -0,0 +1,180 @@
...
+        hardforks.insert(U256::from(3), ropsten_config());
+        hardforks.insert(U256::from(4), rinkeby_config());
+        hardforks.insert(U256::from(5), goerli_config());
+        hardforks.insert(U256::from(42), kovan_config());
+
+        hardforks
+    })
+}
+
+/// Returns the `SpecId` corresponding to the provided chain ID and block number, if the chain ID is supported.
```
feuGeneA, line 162: With just backticks, `cargo doc` won't link this to the doc for that datatype. I tried it (before and after my suggested change), and I confirmed it in https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html#linking-to-items-by-name

```suggestion
/// Returns the [SpecId] corresponding to the provided chain ID and block number, if the chain ID is supported.
```

feuGeneA (changes requested):

crates/rethnet_eth/src/spec.rs
Wodann: Good catch! Backticks still add the formatting though, right?

crates/rethnet_eth/src/spec.rs
```diff
@@ -0,0 +1,180 @@
...
+
+const MAINNET_HARDFORKS: &[(u64, SpecId)] = &[
+    (0, SpecId::FRONTIER),
+    (200_000, SpecId::FRONTIER_THAWING),
+    (1_150_000, SpecId::HOMESTEAD),
+    (1_920_000, SpecId::DAO_FORK),
+    (2_463_000, SpecId::TANGERINE),
+    (2_675_000, SpecId::SPURIOUS_DRAGON),
+    (4_370_000, SpecId::BYZANTIUM),
+    (7_280_000, SpecId::PETERSBURG),
```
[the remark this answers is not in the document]
Wodann: It was superseded and since we can only do one version per block I went with the latest one

crates/rethnet_eth/src/spec.rs
Wodann, line 162:
```suggestion
/// Returns the [`SpecId`] corresponding to the provided chain ID and block number, if the chain ID is supported.
```

feuGeneA: yup just not the link

crates/rethnet_eth/src/spec.rs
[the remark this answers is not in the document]
feuGeneA: Doesn't feel right to leave it out, though maybe my gut feeling there is unfounded. Why can we only do one version per block? Note that over in Hardhat it was pretty [recently that Franco V added](https://github.com/NomicFoundation/hardhat/commit/ad5d50bbd19bb93df0d07cca3b2855d8d8f44384) some verification that all hardforks are represented.

feuGeneA (approved):

PR #4184, opened 2023-07-21
Upgrade to typescript v5

fvictorio:
We'll need this to be able to use viem.

@alcuadrado important things here:
- I'm using `~5.0.0` as the version range in all packages (even the toolbox).
- [This commit](https://github.com/NomicFoundation/hardhat/pull/4184/commits/999c5ea64f15b2873650676c8279e2807721c037) is the one that fixes the two compilation errors that showed up after upgrading.
- I added a workflow to compile the whole repository with typescript v4. I also pushed a temporary commit to double-check that the workflow fails as expected (see [here](https://github.com/NomicFoundation/hardhat/actions/runs/5620956258/job/15230842149)).

### DIFF
--- a/.github/workflows/compile-with-typescript-v4.yml
+++ b/.github/workflows/compile-with-typescript-v4.yml
@@ -1,0 +1,22 @@
+name: Compile with typescript v4
+
+on:
+  push:
+    branches: main
+  pull_request:
+    branches:
+      - "**"
+
+jobs:
+  compile_with_typescript_v4:
+    name: Compile with typescript v4
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - name: Install typescript v4 in all packages
+        run: sed -i 's/"typescript": "~5.0.0"/"typescript": "^4.0.0"/' package.json packages/*/package.json && yarn
+      - name: Build
+        run: yarn build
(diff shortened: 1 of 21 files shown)


alcuadrado:
I think the workflow is overly restrictive. I believe that what we want is to check is that Hardhat's generated `.d.ts` are still compatible with TS 4.x, that's not the same as the sources being compatible.

.github/workflows/compile-with-typescript-v4.yml
```diff
@@ -0,0 +1,22 @@
...
+  compile_with_typescript_v4:
+    name: Compile with typescript v4
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16
+      - name: Install typescript v4 in all packages
+        run: sed -i 's/"typescript": "~5.0.0"/"typescript": "^4.0.0"/' package.json packages/*/package.json && yarn
```
schaable: don't you need two stars in the pattern `packages/*/package.json`?

alcuadrado: I don't think it's needed, as every package is right in `packages/`, like `packages/asd/package.json`

schaable (approved):

fvictorio:
@alcuadrado that makes sense, but since it's passing, do you mind if we merge it? I think testing this properly would require e2e tests. I can update the e2e issue to include that, and we can remove that workflow when we do it.

alcuadrado:
sure!

PR #4191, opened 2023-07-21
fix: Block & Transaction RPC types

Wodann:
Depends on #4189

### DIFF
--- a/crates/rethnet_eth/src/remote/client.rs
+++ b/crates/rethnet_eth/src/remote/client.rs
@@ -750,11 +750,61 @@ mod tests {
         }
 
         #[tokio::test]
+        async fn get_earliest_block() {
+            let alchemy_url = get_alchemy_url();
+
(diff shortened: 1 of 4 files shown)


crates/rethnet_eth/src/remote/client.rs
```diff
@@ -749,6 +749,26 @@ mod tests {
...
+        #[tokio::test]
+        async fn get_earliest_block_with_transaction_data() {
+            let alchemy_url = get_alchemy_url();
+
+            let _block = RpcClient::new(&alchemy_url)
+                .get_block_by_number_with_transaction_data(BlockSpec::earliest())
+                .await
+                .expect("should have succeeded");
+        }
+
```
feuGeneA, line 771: For these, why not validate something in the return data? Since it will always be the same. Perhaps block number or block hash?

feuGeneA (approved):

PR #4198, opened 2023-07-25
Hardhat verify external api

schaable:
- Separate the public API from the internal code. While most of the code was moved into `src/internal`, the `index.ts`, `etherscan.ts`, and `types.ts` files are still in the root and can be considered public.
 - The Etherscan class is accessible through `@nomicfoundation/hardhat-verify/etherscan`. Most of the properties and methods were updated to be better suited for external use.


ericglau:
This looks good to me, thanks!

packages/hardhat-verify/src/internal/config.ts
```diff
@@ -1,7 +1,8 @@
 import type LodashCloneDeepT from "lodash.clonedeep";
+import type { HardhatConfig, HardhatUserConfig } from "hardhat/types";
+import type { EtherscanConfig } from "../types";
+
 import chalk from "chalk";
-import { HardhatConfig, HardhatUserConfig } from "hardhat/types";
-import { EtherscanConfig } from "./types";
```
fvictorio, line 4: Why are these not `import type` anymore?

fvictorio (approved):

PR #4199, opened 2023-07-25
Hardhat verify add sourcify boilerplate

schaable:
- Added the ability to deactivate the verification providers through the `enabled` field in the plugin's configuration.
 - Display errors at the end of the verification process and return with a non-zero exit code, allowing multiple verification providers to be executed without blocking each other.
 - Display a warning when the Sourcify configuration is missing.

### DIFF
--- a/packages/hardhat-verify/src/internal/tasks/sourcify.ts
+++ b/packages/hardhat-verify/src/internal/tasks/sourcify.ts
@@ -1,7 +1,20 @@
 import chalk from "chalk";
 import { subtask } from "hardhat/config";
 
-import { TASK_VERIFY_SOURCIFY_DISABLED_WARNING } from "../task-names";
+import {
+  TASK_VERIFY_SOURCIFY,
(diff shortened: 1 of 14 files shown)


packages/hardhat-verify/src/internal/tasks/sourcify.ts
```diff
@@ -0,0 +1,31 @@
...
+} from "../task-names";
+
+/**
+ * Main Sourcify verification subtask.
+ *
+ * Verifies a contract in Sourcify by coordinating various subtasks related
+ * to contract verification.
+ */
+subtask(TASK_VERIFY_SOURCIFY, async () => {
+  // code here
```
ChristopherDedominici, line 16: Is this placeholder on purpose?

schaable: It is. The Sourcify team will implement the task, this PR covers only the boilerplate to make sure it works seamlessly with Etherscan.

PR #4201, opened 2023-07-26
Make flatten task smarter: improve warning messages and add check for null metadata

ChristopherDedominici:
[Link to main issue](https://github.com/NomicFoundation/hardhat/issues/1499).

- Improve the formatting for warning messages
- Add a check to avoid errors when the flattened file metadata are null

### DIFF
--- a/.changeset/chilly-llamas-sniff.md
+++ b/.changeset/chilly-llamas-sniff.md
@@ -1,5 +1,0 @@
----
-"hardhat": patch
----
-
-Improved the formatting for warning messages and added a check to avoid errors when the flattened file metadata are null.
--- a/.changeset/chilly-llamas-sniff.md
+++ b/.changeset/chilly-llamas-sniff.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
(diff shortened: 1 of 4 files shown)


.changeset/chilly-llamas-sniff.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Improved the formatting for warning messages and added a check to avoid errors when the flattened file metadata are null.
```
fvictorio, line 5: I think in this case we don't need to add a changeset because it's just a fix for the unreleased things in the previous PR.

If you prefer to keep it anyway, then please re-write it to indicate that it's about the flatten task, and remove the implementation-related notes. So, something like "Improved warnings of the flatten task and fixed and error when the flatten task was run in a project with no contracts". I'd just remove the file though.

fvictorio (approved):
Pre-approving, feel free to merge after removing/updating the changeset.

.changeset/chilly-llamas-sniff.md
ChristopherDedominici: Deleted

PR #4202, opened 2023-07-26
Add support for solc 0.8.21

schaable:


packages/hardhat-core/test/fixture-projects/solidity-config-warnings/multiple-unsupported-solc.js
```diff
@@ -1,11 +1,19 @@
+const {
+  getLatestSupportedVersion,
+  increasePatch,
+} = require("../../internal/hardhat-network/stack-traces/compilers-list");
+
+const nextUnsupportedVersion = increasePatch(getLatestSupportedVersion());
+const nextNextUnsupportedVersion = increasePatch(nextUnsupportedVersion);
```
fvictorio, line 7: :open_mouth: :100:

packages/hardhat-core/test/internal/hardhat-network/stack-traces/compilers-list.ts
```diff
@@ -209,4 +211,26 @@ export const solidityCompilers: SolidityCompiler[] = [
...
+    latestSolcVersion: true,
+  },
 ];
+
+export const getLatestSupportedVersion = () =>
+  solidityCompilers.map((sc) => sc.solidityVersion).sort(semver.compare)[
+    solidityCompilers.length - 1
+  ];
+
+export const increasePatch = (version: string): string =>
```
fvictorio, line 235: Personal preference: I would either inline this where it's used (because `semver.inc(version, "patch")` it's pretty self-explanatory IMO) or I would generate and export the `nextVersion` and `nextNextVersion` variables here. This feels like a weird middle ground where we are exporting a generic unrelated functionality.

fvictorio (approved):

packages/hardhat-core/test/internal/hardhat-network/stack-traces/compilers-list.ts
schaable: Yeah, this started with a self-made function to increase the patch version and then I remembered semver already had that built-in. I'll import it directly on the config files.

Issue #4151, opened 2023-07-14
Improve experience of mixing hardhat_reset and fixtures


fvictorio:
If you use fixtures and `hardhat_reset`, you will likely run into an error like:

```
FixtureSnapshotError: There was an error reverting the snapshot of the fixture.

This might be caused by using nested loadFixture calls in a test, for example by using multiple beforeEach calls. This is not supported yet.
```

That message is misleading, and also a lie: we support nested loadFixture calls now.

Two things to do here:
- Update that error message
- Check if there's an easy way to detect that there was a reset and act accordingly (just clear the fixtures array in that case I guess?)

zdenham:
I'm also seeing a similar behavior when using `hardhat_setStorageAt` and `hardhat_setCode` with fixtures

Saty248:
hi @fvictorio ,am working on this issue.. 

1. first of all ,in my pr, am changing the the error message to 

There was an error reverting the snapshot of the fixture.

This might be caused by using hardhat_reset and loadFixture calls in a testcase.  

2. what we also can do is, when hardhat_reset is called, we create a function at the loadFixtures which sets the fixtures array to empty.

any thoughts about this idea?

julianmrodri:
I think I facing this issue. What is strange in my case is that i works locally but not in CI. When the test runs in Github actions I get: What could be the reason Im only seeing the error in CI?  Is there a workaround? I do not have nested beforeEach, but I do a network reset and then a beforeEach that deploys a lot of contracts, and performs impersonation, etc.

``` 
 FixtureSnapshotError: There was an error reverting the snapshot of the fixture.

This might be caused by using nested loadFixture calls in a test, for example by using multiple beforeEach calls. This is not supported yet.
      at loadFixture (/home/runner/work/protocol/protocol/node_modules/@nomicfoundation/hardhat-network-helpers/src/loadFixture.ts:51:15)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (node:internal/process/task_queues:96:5)
      at runNextTicks (node:internal/process/task_queues:65:3)
      at processImmediate (node:internal/timers:437:9)
      at async Context.<anonymous> (/home/runner/work/protocol/protocol/test/integration/FacadeMonitorNew.test.ts:159:9)
  
      Caused by: InvalidSnapshotError: Trying to restore an invalid snapshot.
[...]

PR #4208, opened 2023-07-28
adding a function and editing the error message for the issue #4151

Saty248:
- [x ] Because this PR includes a **bug fix**, relevant tests have been included.

---

i have changed the error message for the issue #4151 and i've also added a function deleteOnReset() which resets the snapshots array.

let me know what other changes I need to make on this issue.

### DIFF
--- a/packages/hardhat-network-helpers/src/loadFixture.ts
+++ b/packages/hardhat-network-helpers/src/loadFixture.ts
@@ -70,6 +70,6 @@ export async function loadFixture<T>(fixture: Fixture<T>): Promise<T> {
 }
 
 // function to clear the snapshots array when hardhat_reset is called..
-export async function deleteOnReset(){
-  snapshots=[];
+export async function clearSnapshots() {
+  snapshots = [];
 }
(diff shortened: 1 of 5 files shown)


fvictorio:
Hey @Saty248, thanks for this. This is a really interesting idea. It won't work for raw `hardhat_reset` calls, so we should still detect those kind of problems, but that could be a separate efforts.

This needs tests though, and I dont' think the code is formatted correctly.

packages/hardhat-network-helpers/src/loadFixture.ts
```diff
@@ -68,3 +68,8 @@ export async function loadFixture<T>(fixture: Fixture<T>): Promise<T> {
     return data;
   }
 }
+
+// function to clear the snapshots array when hardhat_reset is called..
+export async function deleteOnReset(){
```
fvictorio, line 73: This function name is overspecified. I think `clearSnapshots` would be better.

Saty248:
hey @fvictorio,
 i've changed the function name from deleteOnRequest to clearSnapshots and formated the code using lint.Also I've added a test case for the clearsnapshot on reset.ts under test. Do I need to added additional test cases for hardhat reset with a url and block number?

PR #4209, opened 2023-07-28
feat: Implement config json rpc apis

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/116

### DIFF
--- a/crates/rethnet/src/config.rs
+++ b/crates/rethnet/src/config.rs
@@ -36,10 +36,12 @@ pub struct ConfigFile {
     pub accounts: Vec<AccountConfig>,
     pub chain_id: U256,
     pub coinbase: Address,
+    pub network_id: U256,
 }
 
 impl Default for ConfigFile {
     fn default() -> Self {
+        let chain_id = U256::from(31337);
         Self {
             accounts: DEFAULT_PRIVATE_KEYS
                 .into_iter()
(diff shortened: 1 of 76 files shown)


feuGeneA:
Upon running the Hardhat tests with `rethnet node` as a provider, I found the following failures that need to be fixed. However, the existing commits here are still ready to be reviewed.

image
image

(Note that that `0x7a69` is `31337` in decimal, which is the default `network_id` and `chain_id` in the current implementation.)

alcuadrado:
`net_version` is a weird method in that it returns a number in its normal decimal string representation.

The other `eth_chainId` error has two parts: (1) it should be represented as `QUANTITY`*, (2) the value itself.

* `QUANTITY` means `"0x"`-prefixed hex encoded number without leading zeros, except for the value `0`, which's encoded as `0x0`. 

Most numbers in the RPC are represented as `QUANTITY`.

feuGeneA:
> Upon running the Hardhat tests with `rethnet node` as a provider, I found the following failures that need to be fixed. However, the existing commits here are still ready to be reviewed. ... [[source](https://github.com/NomicFoundation/hardhat/pull/4209#issuecomment-1656435253)]

With https://github.com/NomicFoundation/hardhat/pull/4209/commits/5e928ff253a73ca9fbb7a375cff55b38abebf60c, these failures are no longer occurring.

crates/rethnet/src/config.rs
```diff
@@ -34,10 +34,14 @@ pub const DEFAULT_PRIVATE_KEYS: [&str; 20] = [
 pub struct ConfigFile {
     // TODO: expand this per https://github.com/NomicFoundation/rethnet/issues/111
     pub accounts: Vec<AccountConfig>,
+    pub chain_id: U64,
```
Wodann, line 37: What's the reason you're using a `U64` instead of a `u64` for the config file?

Wodann (changes requested):

crates/rethnet/src/config.rs
feuGeneA: Good question. I guess I was just being consistent everywhere, but clearly it makes sense for this to be a `u64`, so I did that in https://github.com/NomicFoundation/hardhat/pull/4209/commits/26610b346e327832760e47491b4df78f3e8cb8a6

Wodann (approved):

PR #4217, opened 2023-07-29
Update plugins.ts, added hardhat-generate-function-selectors

3obby:
added my plugin, 'hardhat-generate-function-selectors', which I use to create synthetix-router-style contract expansions.


fvictorio:
Hi @3obby, is this published to npm? Because https://www.npmjs.com/package/hardhat-generate-function-selectors `404`s.

PR #4218, opened 2023-07-29
feat: Implement miscelaneous net web3 rpc methods

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/137

### DIFF
--- a/crates/rethnet_eth/Cargo.toml
+++ b/crates/rethnet_eth/Cargo.toml
@@ -16,7 +16,6 @@ primitive-types = { version = "0.11.1", default-features = false, features = ["r
 reqwest = { version = "0.11", features = ["blocking", "json"] }
 revm-primitives = { git = "https://github.com/Wodann/revm", rev = "9fdf446", version = "1.1", default-features = false }
 # revm-primitives = { path = "../../../revm/crates/primitives", version = "1.0", default-features = false }
-ruint = { version = "1.7.0", default-features = false }
 rlp = { version = "0.5.2", default-features = false, features = ["derive"] }
 secp256k1 = { version = "0.24.0", default-features = false, features = ["alloc", "recovery"] }
 serde = { version = "1.0.147", default-features = false, features = ["derive"], optional = true }
--- a/crates/rethnet_eth/Cargo.toml
+++ b/crates/rethnet_eth/Cargo.toml
@@ -16,6 +16,7 @@ primitive-types = { version = "0.11.1", default-features = false, features = ["r
 reqwest = { version = "0.11", features = ["blocking", "json"] }
 revm-primitives = { git = "https://github.com/Wodann/revm", rev = "9fdf446", version = "1.1", default-features = false }
 # revm-primitives = { path = "../../../revm/crates/primitives", version = "1.0", default-features = false }
+ruint = { version = "1.7.0", default-features = false }
 rlp = { version = "0.5.2", default-features = false, features = ["derive"] }
 secp256k1 = { version = "0.24.0", default-features = false, features = ["alloc", "recovery"] }
 serde = { version = "1.0.147", default-features = false, features = ["derive"], optional = true }
--- a/crates/rethnet_eth/Cargo.toml
+++ b/crates/rethnet_eth/Cargo.toml
@@ -14,14 +14,13 @@ itertools = { version = "0.10.5", default-features = false, features = ["use_all
 open-fastrlp = { version = "0.1.2", default-features = false, features = ["derive"], optional = true }
 primitive-types = { version = "0.11.1", default-features = false, features = ["rlp"] }
 reqwest = { version = "0.11", features = ["blocking", "json"] }
-revm-primitives = { git = "https://github.com/Wodann/revm", rev = "a4550cc", version = "1.1", default-features = false }
+revm-primitives = { git = "https://github.com/Wodann/revm", rev = "9fdf446", version = "1.1", default-features = false }
 # revm-primitives = { path = "../../../revm/crates/primitives", version = "1.0", default-features = false }
 rlp = { version = "0.5.2", default-features = false, features = ["derive"] }
-ruint = { version = "1.7.0", default-features = false }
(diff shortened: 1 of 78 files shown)


crates/rethnet_rpc_server/build.rs
```diff
@@ -0,0 +1,35 @@
...
+    String(String),
+    Object {
+        version: String,
+        git: Option<String>,
+        rev: Option<String>,
+    },
+}
+
+fn main() {
+    let cargo_toml: CargoToml = toml::from_str(include_str!("../rethnet_evm/Cargo.toml"))
```
Wodann, line 21: Does this work when you run `cargo b --all` from a different directory?

crates/rethnet_rpc_server/build.rs
```diff
@@ -0,0 +1,35 @@
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+#[derive(Debug, Serialize, Deserialize)]
+struct CargoToml {
```
Wodann, line 5: A crate exists for this: https://crates.io/crates/cargo_toml

Wodann (changes requested):
Are you also able to validate that the Hardhat tests are succeeding?

crates/rethnet_rpc_server/build.rs
feuGeneA: I tried it from both the top-level directory of the repo and from a directory deep within the tree (specifically `crates/rethnet_eth/src/remote/methods`) and both worked fine.

crates/rethnet_rpc_server/build.rs
feuGeneA: good call, thanks, i used it in https://github.com/NomicFoundation/hardhat/commit/257f7c8c1f10cb969a2db5355ecf5aef6c72261e

feuGeneA:
> Are you also able to validate that the Hardhat tests are succeeding?

Thanks for double checking me on that. I had run it for some but not all!

The tests for `net_listening` and `web3_sha3` are passing.

The test for `net_peerCount` is passing with https://github.com/NomicFoundation/hardhat/pull/4218/commits/da4ea298916ba6241d087008a45487e642801e96

As of https://github.com/NomicFoundation/hardhat/pull/4218/commits/af82c96fff399298b003bc9f2dc66ab3575caa99, the test for `web3_clientVersion` is failing (as expected) with: `expected 'edr/0.1.0-dev/revm/(https://github.com/Wodann/revm@9fdf446)3.3' to match /^HardhatNetwork\/.*\/@nomicfoundation\/ethereumjs-vm/`

Wodann:
> As of https://github.com/NomicFoundation/hardhat/commit/af82c96fff399298b003bc9f2dc66ab3575caa99, the test for web3_clientVersion is failing (as expected) with: expected 'edr/0.1.0-dev/revm/(https://github.com/Wodann/revm@9fdf446)3.3' to match /^HardhatNetwork\/.*\/@nomicfoundation\/ethereumjs-vm/

@feuGeneA could you please make that test run-time dependent on whether we're running with EDR? That way we can ensure that test succeeds too, such that down the line we can enable this provider too.

fvictorio:
We normally disable tests by checking a property of the provider. I don't remember if the setup added a sort of "isEDR" flag, but I think it didn't. That seems to me like the easiest way to do it, and it's consistent with the current approach. See for example:

https://github.com/NomicFoundation/hardhat/blob/3e4f4b36d4df88214f897d71feb37f1b51b8ee94/packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/sendRawTransaction.ts#L118-L120

crates/rethnet_eth/Cargo.toml
```diff
@@ -16,6 +16,7 @@ primitive-types = { version = "0.11.1", default-features = false, features = ["r
 reqwest = { version = "0.11", features = ["blocking", "json"] }
 revm-primitives = { git = "https://github.com/Wodann/revm", rev = "9fdf446", version = "1.1", default-features = false }
 # revm-primitives = { path = "../../../revm/crates/primitives", version = "1.0", default-features = false }
+ruint = { version = "1.7.0", default-features = false }
```
Wodann, line 19: `revm-primitives` re-exports `ruint`,  so there is no need to add it as a dependency. Using re-exports avoids us growing out of sync with `revm` and introducing compiler errors.

https://docs.rs/revm-primitives/latest/revm_primitives/index.html

crates/rethnet_eth/src/remote/serde_with_helpers.rs
```diff
@@ -1,7 +1,7 @@
 /// helper utilities for use with serde's serialize_with and deserialize_with
 use std::fmt::Write;
 
-use crate::U256;
+use ruint::Uint;
```
Wodann, line 4:
```suggestion
use revm_primitives::ruint::Uint;
```

Wodann (changes requested):

feuGeneA:
> @feuGeneA could you please make that test run-time dependent on whether we're running with EDR? That way we can ensure that test succeeds too, such that down the line we can enable this provider too.

With https://github.com/NomicFoundation/hardhat/pull/4218/commits/98acf988707dd59175f79fc5bbf2ae8a5834f161, the test is passing for both the existing providers and the `rethnet node` provider.

crates/rethnet_eth/Cargo.toml
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4218/commits/6d1e0ba16866a04340148603d71fd79c6a95023b

crates/rethnet_eth/src/remote/serde_with_helpers.rs
feuGeneA: addressed in https://github.com/NomicFoundation/hardhat/pull/4218/commits/6d1e0ba16866a04340148603d71fd79c6a95023b

PR #4219, opened 2023-07-30
feat: Impersonation of accounts

feuGeneA:
Fixes https://github.com/NomicFoundation/rethnet/issues/133

### DIFF
--- a/crates/rethnet/tests/cli.rs
+++ b/crates/rethnet/tests/cli.rs
@@ -53,6 +53,7 @@ async fn node() -> Result<(), Box<dyn std::error::Error>> {
             address,
             bytes::Bytes::from(hex::decode("deadbeef").unwrap()).into(),
         )),
+        MethodInvocation::Hardhat(HardhatMethodInvocation::ImpersonateAccount(address)),
         MethodInvocation::Hardhat(HardhatMethodInvocation::SetBalance(address, U256::ZERO)),
         MethodInvocation::Hardhat(HardhatMethodInvocation::SetCode(
             address,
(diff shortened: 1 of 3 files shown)


Wodann (commented):
LGTM! Are tests for these APIs also working in `Hardhat`?

feuGeneA:
> LGTM! Are tests for these APIs also working in `Hardhat`?

Here's a quick view of the tests that exist:
image

(that shows a result from the `Hardhat Network provider, thus all the green check marks.)

When running with `rethnet node` as a provider, the only test that passes is `returns true`. The `validates invalid input parameter` test fails with an unintuitive error message that we've discussed before and decided that we're okay with for the time being (`Invalid JSON-RPC response received: Failed to deserialize the JSON body into the target type: data did not match any variant of untagged enum Request`).

The remaining tests fail for lack of support for the following RPC methods:
- `Hardhat(SetNextBlockBaseFeePerGas`
- `Eth(SendTransaction`

feuGeneA:
> LGTM! Are tests for these APIs also working in `Hardhat`?

In short, all of the tests we would expect to pass at this point are passing.

In more detail:

Here's a quick view of the tests that exist:
image

(that shows a result from the `Hardhat Network provider, thus all the green check marks.)

When running with `rethnet node` as a provider, the only test that passes is `returns true`. The `validates invalid input parameter` test fails with an unintuitive error message that we've discussed before and decided that we're okay with for the time being (`Invalid JSON-RPC response received: Failed to deserialize the JSON body into the target type: data did not match any variant of untagged enum Request`).

The remaining tests fail for lack of support for the following RPC methods:
- `Hardhat(SetNextBlockBaseFeePerGas`
- `Eth(SendTransaction`

Wodann (approved):

PR #4227, opened 2023-08-02
evm_increaseTime

feuGeneA:


feuGeneA:
I was able to run all of the methods implemented herein, via the hardhat provider tests, and some tests passed, and the ones that did fail were failing due to expected reasons (eg RPC methods not implemented yet)

crates/rethnet/src/config.rs
```diff
@@ -34,6 +34,7 @@ pub const DEFAULT_PRIVATE_KEYS: [&str; 20] = [
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct ConfigFile {
     // TODO: expand this per https://github.com/NomicFoundation/rethnet/issues/111
+    pub allow_blocks_with_same_timestamp: Option<bool>,
```
Wodann, line 37:
```suggestion
    #[serde(default)]
    pub allow_blocks_with_same_timestamp: bool,
```

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -294,6 +307,51 @@ async fn handle_coinbase(state: StateType) -> ResponseData<Address> {
...
+}
+
+async fn handle_evm_set_next_block_timestamp(
+    state: StateType,
+    timestamp: U256OrUsize,
+) -> ResponseData<String> {
+    event!(Level::INFO, "evm_setNextBlockTimestamp({timestamp:?})");
+    match state.blockchain.read().await.last_block() {
+        Ok(latest_block) => {
+            match Into::<U256>::into(timestamp.clone()).checked_sub(latest_block.header.timestamp) {
```
Wodann, line 330:
```suggestion
            match U256::from(timestamp.clone()).checked_sub(latest_block.header.timestamp) {
```

Wodann (changes requested):
Are timestamp-related tests succeeding in Hardhat?

crates/rethnet/src/config.rs
feuGeneA: In [the other PR](https://github.com/NomicFoundation/hardhat/pull/4221) I settled on the pattern of just doing `#[serde(default)]` at the struct level rather than at the field level, since many of these fields need specific values as their defaults rather than just the default value for their type. For that reason, I'd prefer to leave out the `#[serde(default)]` here, instead specifying the default value in `ConfigFile::default()`.  However, the `Option` removal part of your suggestion I did take, when I [merged the `rethnet/142-instantiate-blockchain-mempool-and-miner-in-server` branch into this branch](https://github.com/NomicFoundation/hardhat/pull/4227/commits/a44f2f68ddd9f0cced4c1309322746cbcefe627b).

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: I had tried it this way at first, but I didn't like the result. It gives this error:
image
If I understand correctly, this is because the `ruint` code is set up to generically require a `try_from` for _any_ conversion from an integer. Even though I code it with `::from(_)`, not `::try_from(_)`, it still complains about wanting a `TryFrom`. That's a little quirky, isn't it? Or am I missing something?

In any case, if I'm following the code correctly, the only type of error that could occur is `FromUintError::Overflow`, aka `The Uint value is too large for the target type`. And, in this case, I _know_ that `timestamp` will fit into a `U256`, so I wanted to avoid implementing `TryFrom<U256OrUsize>`.

I'm happy to code this however you'd like, just let me know, but I thought I'd explain first that I _did_ try it your suggested way first, and to explain my rationale for why I settled on the way you see it written here.

Wodann: My understanding of the error is that `From` is not implemented for the `timestamp` type. Is it a `U256OrSize`?

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -284,6 +294,51 @@ async fn handle_coinbase(state: StateType) -> ResponseData<Address> {
     }
 }
 
+async fn handle_evm_increase_time(
+    state: StateType,
+    increment: U256OrUsize,
+) -> ResponseData<String> {
+    event!(Level::INFO, "evm_increaseTime({increment:?})");
+    let increment: U256 = increment.into();
```
Wodann, line 302: Can this be turned into `U256::from(increment)` without getting the same compiler error?

Wodann (changes requested):
I can also check this out locally and have a look if it is up-to-date with `rethnet/main`?

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: Yes, it's a `U256OrSize`, and I actually did already `impl From<U256OrUsize> for U256` alongside `enum U256OrUsize`.

Note that, at least in my IDE, jumping to the definition of `U256::from()` goes to a method on `impl<...> Uint`, not to an implementation of `From<_>`, and that implementation simply delegates to `UintTryFrom`, panicking upon failure. I'm fairly confident that the error is that `TryFrom<U256OrUsize>` is the complaint here.

I went ahead and implemented `TryFrom<U256OrUsize> for U256`, and was able to convert this into `U256::from(_)` (and was required to convert a few other places too, since it told me that I couldn't _also_ implement `From<U256OrUsize>` due to a conflict). I did that in [this commit](https://github.com/NomicFoundation/hardhat/pull/4227/commits/14e602714da7bd480d2a1071990c568ef0325811). It had seemed counterintuitive to me to implement `TryFrom` when I know for sure that converting a `u64` to a `U256` will _never_ induce an `Overflow` error, but if you think the code is cleaner this way then I'm good with that. Let me know if you like that or if you'd like me to revert it so we can do something different.

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: No, it would cause that same error, but since implementing `TryFrom<_> for U256` in [this commit](https://github.com/NomicFoundation/hardhat/pull/4227/commits/14e602714da7bd480d2a1071990c568ef0325811) I was able to convert this instance and a few others into `U256::from(_)`.

crates/rethnet_rpc_server/src/lib.rs
Wodann: That's very odd. I can only imagine that for some reason `U256` doesn't have the conversion version for `usize` because I've used the pattern `U256::from(value)` for a `u64`, IIRC.

Feel free to revert it to what you had originally then, with the `Into::into` and leave it at that. That prevents the unnecessary creation of an error path

feuGeneA:
> Are timestamp-related tests succeeding in Hardhat?

Yes.

For `evm_increaseTime`, two of the three tests are passing, and one is failing with `Method not found for invocation 'Eth(GetBlockByNumber(Tag(Pending), false))'`, because that method isn't supported yet.

For `evm_setNextBlockTimestamp`, of the 25 tests, 7 are passing, 13 are failing due to methods known to be as-of-yet unsupported, and... 5 are failing with a new problem that seems to have appeared upon merging `rethnet/main` into this branch (via the base branch `rethnet/142...`), seemingly involving the changes to `serde` stuff:

```
rethnet subprocess 2232935: 2023-08-08T22:42:31.461858Z  INFO rethnet_rpc_server: evm_setNextBlockTimestamp(U256(0x0000000000000000000000000000000000000000000000000000000064d2c514_U256))

rethnet subprocess 2232935: thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 0', /home/gene/dev/nomiclabs/rethnet/crates/rethnet_eth/src/serde.rs:144:8
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

          9) should accept a hex string param
```

I'm looking into that now.

feuGeneA:
That `panic` turned out to be related not to the merging in of the base branch, but rather to the not-thoroughly-tested introduction of the `evm_mine` request type in this PR. I fixed it in https://github.com/NomicFoundation/hardhat/pull/4227/commits/3267b6c76e95e3a8d80959f17fbcffb7474cd80e. Now, the `evm_setNextBlockTimestamp` tests have (still just) 7 of the 25 passing, with the remainder failing with clean "unsupported method" messages.

Wodann (approved):

PR #4233, opened 2023-08-02
improvement: add clippy warnings

Wodann:


crates/rethnet_eth/src/remote.rs
```diff
@@ -16,7 +16,7 @@ pub mod jsonrpc;
 /// RPC methods
 pub mod methods;
 
-/// helper utilities for use with serde's serialize_with and deserialize_with
+/// helper utilities for use with serde's `serialize_with` and `deserialize_with`
```
feuGeneA, line 19: did clippy catch this or did you? i'd be impressed haha

crates/rethnet_eth/src/remote/eth.rs
```diff
@@ -88,7 +88,7 @@ where
     Ok(u64::from_str_radix(&s[2..], 16).expect("failed to parse u64"))
 }
 
-/// log object used in TransactionReceipt
+/// log object used in `TransactionReceipt`
```
feuGeneA, line 91: This one should have the `[]` brackets, since it's actually a code object (right?)

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -214,7 +214,7 @@ impl RpcClient {
     }
 
     /// Submit a consolidated batch of RPC method invocations in order to obtain the set of data
-    /// contained in AccountInfo.
+    /// contained in [`AccountInfo`].
```
feuGeneA, line 217: fyi ["Backticks around the link will be stripped"](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html) so i don't _think_ they have any effect here

crates/rethnet_eth/src/lib.rs
```diff
@@ -1,4 +1,77 @@
...
+    clippy::unnested_or_patterns,
+    clippy::unused_self,
+    clippy::useless_transmute,
+    clippy::verbose_file_reads,
+    clippy::zero_sized_map_values,
+    future_incompatible,
+    missing_docs,
+    nonstandard_style,
+    rust_2018_idioms
+)]
```
feuGeneA, line 74: I'm confused about why there's the huge list here and also in the .cargo/config file. does the top-level one not apply to all of the different crates in the workspace? also, is it intended that the two lists differ? because I see several entries that appear in one list but not the other. if having two lists is needed/preferred, it would be ideal to de-duplicate them somehow.

crates/eth_execution_api/src/lib.rs
feuGeneA, line 1: Just for my own education, why was this needed? Did clippy demand it? Does it really expect to find an empty file? does it recognize that this is the sample/demo output and recommend an empty file instead?

crates/rethnet_eth/src/signature.rs
```diff
@@ -113,7 +113,8 @@ impl fmt::Display for Signature {
 }
 
 impl Signature {
-    /// to obtain message_hash consider rethnet_eth::src::utils::hash_message().
+    /// Constructs a new signature from a message and private key.
+    /// To obtain the hash of a message consider [`rethnet_eth::src::utils::hash_message`].
```
feuGeneA, line 117:
```suggestion
    /// To obtain the hash of a message consider [`rethnet_eth::utils::hash_message`].
```

feuGeneA (changes requested):

crates/rethnet_eth/src/remote.rs
Wodann: Clippy did  :slight_smile:

crates/eth_execution_api/src/lib.rs
Wodann: It was angry about something else in that file and instead of fixing it, I just decided to remove the code as it was default code anyway

crates/rethnet_eth/src/lib.rs
Wodann: That's my bad. I started by adding it to each crate and then learned of an alternative: the cargo.toml file

crates/rethnet_eth/src/remote/client.rs
Wodann: There is actually a difference:

image

feuGeneA (approved):

PR #4236, opened 2023-08-03
Forbid chaining matchers

schaable:


packages/hardhat-chai-matchers/README.md
```diff
@@ -47,3 +47,39 @@ await expect(token.transfer(recipient, 1000))
...
+
+To work around this limitation, write separate assertions for each matcher:
+
+```js
+const tx = contract.f(...);
+await expect(tx).to.changeEtherBalance(...)
+await expect(tx).to.changeTokenBalance(...)
+```
+
+The decision to forbid chaining for these async matchers is due to the complexity involved in implementing proper chaining behavior. Since async matchers perform asynchronous operations and return Promises, chaining them in a meaningful way requires handling Promise resolutions and rejections correctly.
```
fvictorio, line 83: I think I would drop this paragraph, most people won't care about our ~excuses~ reasons.

packages/hardhat-chai-matchers/src/internal/errors.ts
```diff
@@ -23,3 +23,11 @@ export class HardhatChaiMatchersAssertionError extends HardhatChaiMatchersError
     super(`Assertion error: ${message}`);
   }
 }
+
+export class HardhatChaiMatchersNonChainableMatcherError extends HardhatChaiMatchersError {
+  constructor(matcherName: string) {
+    super(
+      `${matcherName} is not chainable. Please refer to the [README](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-chai-matchers#chaining-async-matchers) for more information.`
```
fvictorio, line 30: We can't use markdown here. The way we normally do this is to add a shortlink like [this](https://github.com/NomicFoundation/hardhat/blob/f324b3a335587d24da3c1901ed2b5728d6e4cebb/docs/redirects.config.js#L64-L68) and then use that in the error message.

packages/hardhat-chai-matchers/src/internal/emit.ts
```diff
@@ -38,7 +39,7 @@ export function supportEmit(
   chaiUtils: Chai.ChaiUtils
 ) {
   Assertion.addMethod(
-    "emit",
+    EMIT_MATCHER,
     function (this: any, contract: Contract, eventName: string) {
```
fvictorio, line 43: This matcher doesn't have the chaining check.

packages/hardhat-chai-matchers/src/internal/utils.ts
```diff
@@ -10,3 +14,14 @@ export function assertIsNotNull<T>(
...
+
+export function preventAsyncMatcherChaining(
+  context: object,
+  matcherName: string,
+  chaiUtils: Chai.ChaiUtils
+) {
+  if (chaiUtils.flag(context, ASYNC_MATCHER_CALLED) === true) {
+    throw new HardhatChaiMatchersNonChainableMatcherError(matcherName);
+  }
+  chaiUtils.flag(context, ASYNC_MATCHER_CALLED, true);
```
fvictorio, line 26: Random idea, we could use the matcher name as the flag value:

```suggestion
  const previousMatcherName = chaiUtils.flag(context, ASYNC_MATCHER_CALLED);
  if (previousMatcherName !== undefined) {
    throw new HardhatChaiMatchersNonChainableMatcherError(matcherName, previousMatcherName);
  }
  chaiUtils.flag(context, ASYNC_MATCHER_CALLED, matcherName);
```

and then change the error message to say something like `${matcherName} was called after ${previousMatcherName}`. Wdyt?

packages/hardhat-chai-matchers/test/helpers.ts
```diff
@@ -153,3 +156,20 @@ export async function runFailedAsserts({
     failedAssert(matchers[method].staticCall(...args))
   ).to.be.rejectedWith(AssertionError, failedAssertReason);
 }
+
+/**
+ * Used to test that non-chainable matchers like changeEtherBalance,
+ * changeTokenBalance and reverted throw an error when used along with other
+ * non-chainable matchers.
+ */
+function supportNonChainableMatcher(
```
fvictorio, line 165: I'm not sure about this. I'd rather use realistic tests, because this feels too white-boxy to me.

packages/hardhat-chai-matchers/src/internal/utils.ts
schaable: I like it 👍

packages/hardhat-chai-matchers/test/helpers.ts
schaable: I think the transparency is needed here. Chaining it to other matchers that could or could not be incompatible is less clear. Instead, this expresses clearly the test intent.

fvictorio (approved):

VVander:
Thank you 
image

PR #4243, opened 2023-08-05
feat: implement mining json rpc methods

feuGeneA:
- [x] `evm_mine` (with stubbed logging)
- [x] `hardhat_intervalMine` (with stubbed logging)
- [x] `hardhat_mine` (with stubbed logging)
- [x] QA
- [x] review

### DIFF
--- a/crates/rethnet_rpc_server/src/hardhat_methods.rs
+++ b/crates/rethnet_rpc_server/src/hardhat_methods.rs
@@ -52,7 +52,7 @@ pub enum HardhatMethodInvocation {
     #[serde(rename = "hardhat_mine")]
     Mine(
         /// block count:
-        #[serde(default, skip_serializing_if = "Option::is_none")]
+        #[serde(default)]
         Option<U256>,
         /// interval:
         #[serde(default, skip_serializing_if = "Option::is_none")]
@@ -213,6 +213,7 @@ mod tests {
             Some(U256::from(1)),
         ));
         help_test_method_invocation_serde(HardhatMethodInvocation::Mine(Some(U256::from(1)), None));
+        help_test_method_invocation_serde(HardhatMethodInvocation::Mine(None, Some(U256::from(1))));
         help_test_method_invocation_serde(HardhatMethodInvocation::Mine(None, None));
     }
 
--- a/crates/rethnet_rpc_server/src/hardhat_methods.rs
+++ b/crates/rethnet_rpc_server/src/hardhat_methods.rs
@@ -52,9 +52,11 @@ pub enum HardhatMethodInvocation {
     #[serde(rename = "hardhat_mine")]
     Mine(
         /// block count:
-        U256,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        Option<U256>,
         /// interval:
-        U256,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        Option<U256>,
     ),
     /// hardhat_reset
     #[serde(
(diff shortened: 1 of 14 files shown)


crates/rethnet_rpc_server/src/hardhat_methods.rs
```diff
@@ -52,9 +52,11 @@ pub enum HardhatMethodInvocation {
     #[serde(rename = "hardhat_mine")]
     Mine(
         /// block count:
-        U256,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        Option<U256>,
         /// interval:
-        U256,
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        Option<U256>,
```
Wodann, line 59: Can this ever specify just the interval but not the block count?

In that case, would it interpret the first value that appears to be the block count regardless?

Could you please add a test to validate this?

crates/rethnet_rpc_server/src/lib.rs
```diff
@@ -320,14 +460,14 @@ async fn handle_evm_set_next_block_timestamp(
                         error_response_data(0, &format!("Timestamp {timestamp:?} is equal to the previous block's timestamp. Enable the 'allowBlocksWithSameTimestamp' option to allow this"))
                     } else {
                         let mut next_block_timestamp = state.next_block_timestamp.write().await;
-                        *next_block_timestamp = timestamp.into();
+                        *next_block_timestamp = Some(timestamp.into());
                         ResponseData::Success {
-                            result: next_block_timestamp.to_string(),
+                            result: next_block_timestamp.unwrap().to_string(),
```
Wodann, line 465: Why not use `timestamp` or an intermediate result thereof? The `.unwrap` seems unnecessary here

Wodann (changes requested):
I left some comments

crates/rethnet_rpc_server/src/hardhat_methods.rs
feuGeneA: > Can this ever specify just the interval but not the block count?

Because the input is JSON, the ways to decline to specify an argument for an optional parameter are either (a) to simply leave it out, which effectively renders it `undefined`, or (b) to actually pass `null` or `undefined` for the value. So yes, in theory someone could pass `null` or `undefined` for the first value in order to specify only the second value.
 
> In that case, would it interpret the first value that appears to be the block count regardless?

If only one value appears, it will always be interpreted as the block count. However, again, one _can_ specify solely the interval, by passing `null` or `undefined` for the block count.

I hadn't considered this case, so thank you for raising it. I realized that `skip_serializing_if = "Option::is_none"` is incorrect on the first parameter, since that might cause an arg list of `undefined, 1` to be interpreted as just `1`, so I removed that, and added the test you requested below, in https://github.com/NomicFoundation/hardhat/pull/4243/commits/8426fe6db416f58fa3fd20afb917e9fdc74fbbec

> 
> Could you please add a test to validate this?
> 
>     1. serialize `Mine(None, Some(U256::from(1)))`
> 
>     2. deserialize the output of 1
> 
>     3. `assert_eq!(deserialized.0, None)` && `assert_eq!(deserialized.1, Some(U256::from(1)))`

crates/rethnet_rpc_server/src/lib.rs
feuGeneA: Thanks, addressed in https://github.com/NomicFoundation/hardhat/pull/4243/commits/015bf7d7b1bc43dcbb6b6e3fe5c0524bc972ae00

crates/rethnet_rpc_server/src/hardhat_methods.rs
Wodann: In that case, should we ever default a value when it's missing? Or should we always require both values to be specified to `null`?

feuGeneA: I think that, due to the nature of JSON and this being a JSON-RPC interface, we shouldn't require both values to be specified to `null`. For good measure, I added a test that exercises the case of providing an empty `params` array, in https://github.com/NomicFoundation/hardhat/pull/4243/commits/948f871760f019b3983724327f3775acc4ae8e38; I copied the input request verbatim from what a particular Hardhat provider test sends.

Wodann (approved):

PR #4244, opened 2023-08-05
fix: use node v18 for hardhat-ledger

Wodann:


Wodann:
@fvictorio why is the ledger test failing, whereas the other ones are not?

fvictorio:
I have no idea why it was working before and then it stopped working. But the fix I sent is something we'll probably do in all packages: the build script runs the root's build script, and this is run before running the tests.

That being said, can you explain why is this change necessary? And why it's done only in `hh-ledger`?

Wodann:
> That being said, can you explain why is this change necessary? And why it's done only in `hh-ledger`?

The change I made was to update the hardhat-ledger CI to require node 18. This was a change that was already done to all other CI workflows previously, as a bug existed in older node versions when marshalling a buffer of data.

fvictorio:
> This was a change that was already done to all other CI workflows previously, as a bug existed in older node versions when marshalling a buffer of data.

Ah, sorry, I completely forgot about that.

fvictorio (approved):

PR #4248, opened 2023-08-07
docs: add book with getting started chapter

Wodann:


agostbiro (approved):
Thanks @Wodann, LGTM. I managed to build Hardhat core based on the build instruction. 

I'm on MacOS and I already had all the dependencies installed, so I didn't try the setup steps. One thing we could add there if you think it make sense is a manual option that just lists the dependencies to install (rustup, node 18, yarn if I'm not mistaken).

Wodann:
> Thanks @Wodann, LGTM. I managed to build Hardhat core based on the build instruction.
> 
> One thing we could add there if you think it make sense is a manual option that just lists the dependencies to install (rustup, node 18, yarn if I'm not mistaken).

I'm happy to add a manual install section for MacOS and Windows. I can do the Windows one. Would you mind creating a PR for the instructions for MacOS? Bonus points if you can create a script that automatically does it on macOS 😃

PR #4252, opened 2023-08-07
docs: add automated macOS and manual setup instructions

agostbiro:
Adds automated macOS and manual setup instructions to the book.

It’s difficult to write a setup script for MacOS as people have different preferred installation methods, so this PR adds a [Brewfile](https://github.com/Homebrew/homebrew-bundle) to the repo which lets users install the dependencies with a single command using the [Homebrew](https://brew.sh) package manager which is the most popular one.


.gitignore
```diff
@@ -117,3 +117,7 @@ Cargo.lock
 
 # Rethnet remote node cache
 remote_node_cache/
+
+# Brew package manager. 
+# Lock file is not recommended to be commited: https://github.com/Homebrew/homebrew-bundle#versions-and-lockfiles
+Brewfile.lock.json
```
Wodann, line 123:
```suggestion
Brewfile.lock.json
```

Wodann, line 123: git show this error when a EOL (end of line) is not present at the end of the file. To avoid it, I always set my IDE to automatically end a newline at the end of my file on save.

Wodann (changes requested):

.gitignore
agostbiro: Thanks, fixed in e0eb4862fddce9b89c7dbd5d5d234405d9ae578e

Wodann (approved):

PR #4254, opened 2023-08-07
refactor: integrate rethnet blockchain, miner, and mempool in hardhat

Wodann:


Wodann:
@agostbiro could you please review the Rethnet files? I've had to do multiple merges from `rethnet/main` into this branch over time and I might have made some mistakes, so it would be good to have a second pair of eyes.

@fvictorio could you please pair with me on this to make sure that the Hardhat refactor is valid?

crates/rethnet_evm/src/blockchain/storage/contiguous.rs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /// Reverts to the block with the provided number, deleting all later blocks.
+    pub fn revert_to_block(&mut self, block_number: &U256) -> bool {
+        let block_number = usize::try_from(block_number)
+            .expect("No blocks with a number larger than usize::MAX are inserted");
+
+        let block_index = if let Some(first_block) = self.blocks.first() {
+            let first_block_number = usize::try_from(first_block.header().number)
+                .expect("No blocks with a number larger than usize::MAX are inserted");
```
agostbiro, line 51: I think it'd be better to encode the invariant that the block number cannot exceed `usize::MAX` in the block type if it's not too much trouble. Also, I'm wondering about using `usize` vs `u64` here? I think the latter is safer for portability, since it's not inconceivable that block numbers will exceed 2^32 in the future.

crates/rethnet_rpc_server/src/node.rs
```diff
@@ -238,15 +291,16 @@ impl NodeData {
...
             &mut *self.blockchain,
-            &mut *self.state,
+            self.state.clone(),
             &mut self.mem_pool,
             &self.evm_config,
             block_timestamp,
-            block_gas_limit,
             self.beneficiary,
+            self.min_gas_price,
+            // TODO: make this configurable
```
agostbiro, line 302: Could you add an issue for this?

crates/rethnet_evm/src/blockchain.rs
```diff
@@ -110,6 +107,18 @@ pub trait Blockchain {
...
     ) -> Result<Option<Arc<BlockReceipt>>, Self::BlockchainError>;
 
+    /// Retrieves the hardfork specification of the block at the provided number.
+    async fn spec_at_block_number(&self, number: &U256) -> Result<SpecId, Self::BlockchainError>;
+
+    /// Retrieves the hardfork specification used for new blocks.
+    fn spec_id(&self) -> SpecId;
+
+    /// Retrieves the state at a given block
+    async fn state_at_block(
```
agostbiro, line 117: Don't forget to rename this please

crates/rethnet_evm/src/blockchain/storage/contiguous.rs
Wodann: We intend to convert the type of block numbers from `U256` to `u64`. I've created an issue for it here: https://github.com/NomicFoundation/rethnet/issues/185

crates/rethnet_rpc_server/src/node.rs
Wodann: Resolved in https://github.com/NomicFoundation/hardhat/pull/4254/commits/5f1e7d44273bf2bd16303c1873d9ccb71614bf1f

packages/hardhat-core/src/internal/hardhat-network/provider/vm/ethereumjs.ts
```diff
@@ -318,7 +317,10 @@ export class EthereumJSAdapter implements VMAdapter {
         gasUsed: ethereumJSResult.totalGasSpent,
         receipt: ethereumJSResult.receipt,
         returnValue: ethereumJSResult.execResult.returnValue,
-        createdAddress: ethereumJSResult.createdAddress,
+        createdAddress:
+          ethereumJSError === undefined
+            ? ethereumJSResult.createdAddress
+            : undefined,
```
Wodann, line 323: In the future, we should add something to match Geth in the JSON-RPC layer.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/ethereumjs.ts
```diff
@@ -501,7 +503,10 @@ export class EthereumJSAdapter implements VMAdapter {
       gasUsed: ethereumJSResult.totalGasSpent,
       receipt: ethereumJSResult.receipt,
       returnValue: ethereumJSResult.execResult.returnValue,
-      createdAddress: ethereumJSResult.createdAddress,
+      createdAddress:
+        ethereumJSError === undefined
+          ? ethereumJSResult.createdAddress
+          : undefined,
```
Wodann, line 509: In the future, we should add something to match Geth in the JSON-RPC layer.

fvictorio (approved):

PR #4255, opened 2023-08-07
docs: add test page

Wodann:


agostbiro (approved):

Wodann:
I'll make a new PR to improve those points; thanks!

book/src/01_getting_started/03_test.md
```diff
@@ -0,0 +1,42 @@
+# Test
+
+Functionality in EDR is tested in two way:
```
feuGeneA, line 3: Suggestion: `two ways`. (The original is singular.)

book/src/01_getting_started/03_test.md
```diff
@@ -0,0 +1,42 @@
...
+```bash
+cargo t --features bench-once,serde,std,tracing
+```
+
+The `bench-once` feature flag is used to ensure that benchmarks only run one iteration, to avoid taking too long.
+
+## Hardhat
+
+To validate that the port of Hardhat Node to EDR did not break any functionality, we implemented the EDR integration alongside the existing TypeScript code.
+Each system in hidden behind an interface that allows us to execute the original Hardhat implementation, EDR, or a dual-mode adapter that executes both implementations side-by-side and asserts that outputs are equal.
```
feuGeneA, line 31: Suggestion: "that allows us to execute _either_", so that when you start reading the list you know it's an "or list" rather than an "and list."

book/src/01_getting_started/03_test.md
```diff
@@ -0,0 +1,42 @@
...
+
+To switch modes, set the `HARDHAT_EXPERIMENTAL_VM_MODE` environment variable to one of: `ethereumjs`, `rethnet`, or `dual` (default). E.g.:
```
feuGeneA, line 33: consider placing after `rethnet` something like "(representing EDR)"

Wodann: Addressed in https://github.com/NomicFoundation/hardhat/pull/4262

book/src/01_getting_started/03_test.md
Wodann: Addressed in https://github.com/NomicFoundation/hardhat/pull/4262

book/src/01_getting_started/03_test.md
Wodann: Addressed in https://github.com/NomicFoundation/hardhat/pull/4262

Wodann:
Addressed remaining comments in Addressed in https://github.com/NomicFoundation/hardhat/pull/4262

Issue #3490, opened 2022-12-29
Improve error message when a directory is imported


fvictorio:
If you do something like (notice that it's a directory and not a file):

```
import "@openzeppelin/contracts/proxy";
```

you get this error:

```
An unexpected error occurred:

[Error: EISDIR: illegal operation on a directory, read] {
  errno: -21,
  code: 'EISDIR',
  syscall: 'read'
}
```

We should show a better error in these cases.

mwawrusch:
This should bumped in priority. I have a project with 50+ contracts. Will take forever to figure out what the problem is.

alcuadrado:
@fvictorio this could be a good issue to introduce someone to HH's codebase.

PR #4258, opened 2023-08-08
Improve error message when a directory is imported

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/3490).

**ISSUE DESCRIPTION:**
If you do something like (notice that it's a directory and not a file):

import "@openzeppelin/contracts/proxy";
you get this error:

An unexpected error occurred:

[Error: EISDIR: illegal operation on a directory, read] {
  errno: -21,
  code: 'EISDIR',
  syscall: 'read'
}
We should show a better error in these cases.

**TODO:**
- improve error message when a directory is trying to be imported
- improve error message if a sol file that does not exist is trying to be imported


packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -161,9 +161,22 @@ subtask(TASK_COMPILE_SOLIDITY_READ_FILE)
...
+        if (!fsExtra.existsSync(absolutePath)) {
+          throw new HardhatError(ERRORS.RESOLVER.FILE_NOT_FOUND);
+        }
+
+        if (fsExtra.lstatSync(absolutePath).isDirectory()) {
+          throw new HardhatError(ERRORS.RESOLVER.INVALID_IMPORT_OF_DIRECTORY);
+        }
+
+        // eslint-disable-next-line @nomicfoundation/hardhat-internal-rules/only-hardhat-error
+        throw e;
```
schaable, line 178: Shouldn't we have a custom error instead of re-throwing?

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -161,9 +161,22 @@ subtask(TASK_COMPILE_SOLIDITY_READ_FILE)
...
     async ({ absolutePath }: { absolutePath: string }): Promise<string> => {
-      return fsExtra.readFile(absolutePath, {
-        encoding: "utf8",
-      });
+      try {
+        return await fsExtra.readFile(absolutePath, {
+          encoding: "utf8",
+        });
+      } catch (e) {
+        if (!fsExtra.existsSync(absolutePath)) {
```
schaable, line 169: I'd put these checks before the try-catch block so that it follows the error-first approach, making it a bit more readable. It's a bit of a personal preference, so if you think your way is better, feel free to ignore this comment.

packages/hardhat-core/test/fixture-projects/compilation-import-folder/node_modules/some-plugin/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "some-plugin",
+  "version": "1.2.3"
+}
```
schaable, line 4:
```suggestion
}

```

packages/hardhat-core/test/fixture-projects/compilation-import-folder/node_modules/some-plugin/index.js
```diff
@@ -0,0 +1 @@
+task('some-task').setAction(() => {})
```
schaable, line 1:
```suggestion
task('some-task').setAction(() => {})

```

packages/hardhat-core/test/fixture-projects/compilation-import-folder/contracts/A.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.19;
+
+// The file has to not exist for test purposes
+import "some-plugin"
+
+contract A {}
```
schaable, line 7:
```suggestion
contract A {}

```

fvictorio: Our `.editorconfig` file has `insert_final_newline = true`. @ChristopherDedominici can you double-check that your IDE is using the `.editorconfig` settings? Maybe you need to install [this](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig).

packages/hardhat-core/src/internal/core/errors-list.ts
```diff
@@ -855,6 +855,16 @@ Use a relative import instead of referencing the package's name.`,
...
       shouldBeReported: false,
     },
+    INVALID_IMPORT_OF_DIRECTORY: {
+      number: 414,
+      message:
+        "Invalid import %imported% from %from%. Attempting to import a directory. Directories cannot be imported.",
+      title: "Invalid import: a directory cannot be imported",
+      description: `A Solidity file is attempting to import a directory, which is not possible.
+
+      Please double check your imports.`,
```
fvictorio, line 865: Text inside backticks is used verbatim and interpreted as markdown, so this shouldn't be indented. In practice I think this doesn't change anything because whatever markdown pre-processor we are using right now doesn't seem to handle [indented code blocks](https://spec.commonmark.org/0.30/#indented-code-blocks), but that could easily change.

```suggestion
Please double check your imports.`,
```

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -161,9 +161,22 @@ subtask(TASK_COMPILE_SOLIDITY_READ_FILE)
...
+          throw new HardhatError(ERRORS.RESOLVER.FILE_NOT_FOUND);
+        }
+
+        if (fsExtra.lstatSync(absolutePath).isDirectory()) {
+          throw new HardhatError(ERRORS.RESOLVER.INVALID_IMPORT_OF_DIRECTORY);
```
fvictorio, line 174: Sorry, I think I didn't explain this correctly in our call. We shouldn't throw resolver-related errors here, because this subtask is just about reading files. We should throw a more general "tried to read a file, but found a directory" error.

To be explicit, this is what I would do:
- Add a new error in the `GENERAL` category about trying to read a directory as a file
- In the resolver, use `isHardhatErrorType` to detect those kind of errors, and re-throw the `RESOLVER.INVALID_IMPORT_OF_DIRECTORY` one.
- Remove the previous `if` statement that also throws a resolver error. That was (I guess) handled correctly before, and so the `throw e` at the end should keep whatever behavior we have now.

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -161,9 +161,22 @@ subtask(TASK_COMPILE_SOLIDITY_READ_FILE)
...
+      try {
+        return await fsExtra.readFile(absolutePath, {
+          encoding: "utf8",
+        });
+      } catch (e) {
+        if (!fsExtra.existsSync(absolutePath)) {
+          throw new HardhatError(ERRORS.RESOLVER.FILE_NOT_FOUND);
+        }
+
+        if (fsExtra.lstatSync(absolutePath).isDirectory()) {
```
fvictorio, line 173: We should use the async version of this method. In general we always use async functions for file-system stuff except in certain cases where the function _has_ to be sync.

packages/hardhat-core/src/builtin-tasks/compile.ts
fvictorio: I agree that that would be more readable but I think the current approach is better performance-wise. I don't have numbers to prove it, but adding an extra file system operation for each solidity file in the project is almost surely wasteful.

packages/hardhat-core/src/builtin-tasks/compile.ts
ChristopherDedominici: As mentioned in our chat I'd go sync here because;
- it's an error so the process will stop anyway, no reason to go async
- sync is more readable

packages/hardhat-core/test/fixture-projects/compilation-import-folder/contracts/A.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.19;
+
+// The file has to not exist for test purposes
+import "some-plugin"
```
fvictorio, line 5: This is very minor, but "plugin" normally refers to Hardhat plugins, not to solidity libraries. So I would rename this to `some-lib` and would remove the `index.js` from `node_modules/some-plugin`.

Also, I think we should have a different test that imports a directory too but using a relative path.

packages/hardhat-core/test/internal/solidity/resolver.ts
```diff
@@ -594,3 +594,33 @@ describe("Resolver regression tests", function () {
...
+      async () => {
+        await this.env.run(TASK_COMPILE);
+      },
+      ERRORS.RESOLVER.INVALID_IMPORT_OF_DIRECTORY,
+      "HH414: Invalid import some-plugin from contracts/A.sol. Attempting to import a directory. Directories cannot be imported."
+    );
+  });
+});
+
+describe("TASK_COMPILE: the file to compile is trying to import a non existing file", function () {
```
fvictorio, line 613: I'm not against adding this if we didn't have such a test, but just to make sure I understand: this is unrelated to the PR changes, right?

packages/hardhat-core/test/fixture-projects/compilation-import-non-existing-file/contracts/A.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.19;
+
+// For test purposes this file must not exist
+import "some-plugin/nonExistingFile.sol"
```
fvictorio, line 5: Same comment about the dependency here. I'd rename it to `some-lib` to avoid confusions, and remove the `index.js` that adds a task.

packages/hardhat-core/test/internal/solidity/resolver.ts
ChristopherDedominici: Correct

schaable (approved):

packages/hardhat-core/test/fixture-projects/compilation-import-folder-from-path/contracts/A.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.19;
+
+// For test purposes this import must point to a folder
+import "../contracts";
```
fvictorio, line 5: Ugh, I know I'm being a pain in the ass, but this feels too unrealistic to me. It's probably fine, but I'd rather do something like adding a `dir` directory with a `dir/B.sol` file, and import `./dir`.

fvictorio (approved):

PR #4265, opened 2023-08-09
Fix escaped double quote in vyper docs

ChristopherDedominici:
Change the ```vy into ```py in the vyper docs.
With ```vy the " character is escaped into &quot;

Error:
image

Expected:
image

### DIFF
--- a/.changeset/little-pianos-suffer.md
+++ b/.changeset/little-pianos-suffer.md
@@ -1,5 +1,0 @@
----
-"@nomiclabs/hardhat-vyper": patch
----
-
-Fixed escaped double quote in vyper docs.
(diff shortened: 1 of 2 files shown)


.changeset/little-pianos-suffer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomiclabs/hardhat-vyper": patch
+---
+
+Fixed escaped double quote in vyper docs.
```
fvictorio, line 5: Doc changes don't need changesets because there is nothing to be released. The website is automatically updated with each change in `main`.

fvictorio (approved):
Pre-approving, feel free to merge after removing the changeset.

PR #4269, opened 2023-08-09
fix: `U256OrUsize` always being deserialized as U256

Wodann:


crates/rethnet_eth/src/remote/methods.rs
```diff
@@ -322,10 +322,10 @@ pub enum MethodInvocation {
 #[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
 #[serde(untagged)]
 pub enum U256OrUsize {
-    /// U256
-    U256(U256),
     /// usize
     Usize(usize),
+    /// U256
+    U256(U256),
```
Wodann, line 328: It seems like `U256` is implemented to deserialize both numbers and hex strings. The implementation of an enum in serde always tries deserializing the different types from top to bottom. The first one that succeeds is accepted.

This ensures that we first try to deserialize a number. If that fails, we deserialize a `U256`.

This is raising the question whether the `U256OrUsize` type is even needed, @feuGeneA ?

feuGeneA: The type is needed because the RPC methods `evm_mine`, `evm_increaseTime` and `evm_setNextBlockTimestamp` all take a number that can be specified as either an integer or as a hex-encoded string, and the `U256` type cannot be deserialized from a simple integer. (I manually confirmed this via the Hardhat provider tests before responding here: I changed the input types for those methods to `U256`, and the server failed to deserialize the request when the provider tests sent a simple integer.)

feuGeneA: @Wodann and I paired on this. For some reason, the test failure that this PR fixes, as reproduced in CI on all platforms and as reproduced by @agostbiro as well, does NOT fail on my machine. (I'm running `Linux xps13 6.0.0-3-rt-amd64 #1 SMP PREEMPT_RT Debian 6.0.7-1 (2022-11-05) x86_64 GNU/Linux`.) So at this point it's not clear whether the type in question really is needed generally, as indicated by my local confirmation mentioned in my last message, or whether that confirmation really indicates that the type is only needed on my machine. 🙄 In any case, we decided to leave the type in place for now.

feuGeneA (approved):

PR #4270, opened 2023-08-09
feat: cache rpc per chain id instead of per provider

agostbiro:
This PR introduces caching RPC requests per chain id instead of per provider (resolves https://github.com/NomicFoundation/rethnet/issues/99). 

To achieve this, the RPC client needs to know the chain id that it's connected to. I solved this by lazily fetching the chain id from the provider on the first request. This has two drawbacks:

1. The first request with the RPC client will be slower, as it'll send two requests.
2. If the RPC client is used to fetch requests from a local node simulating an existing chain, that can pollute the cache with invalid responses. As far as I can tell the client is only used with remote nodes and I added a warning against using the RPC client with local nodes. This might be insufficient though if the url is set by a user with an env var to a local node.

As an alternative solution the RPC client could require the chain id as input, but if the input comes from the user, this wouldn't solve the potentially polluted cache problem either, it could only make the first request faster.

The PR also includes some refactoring:

1. Use `tokio::fs` instead of `std::fs` to avoid blocking in async functions.
2. I changed some panics to errors to make error handling more consistent. 
3. I added the [preserve order](https://docs.rs/serde_json/latest/serde_json/enum.Value.html#variant.Object) feature to `serde` to make sure the requests are consistently serialized. This comes with a small performance penalty, but it should increase cache hit rate.

A future improvement could be to use human-readable chain id in the directory name of the cache. I didn't see an easy way to achieve this without introducing new dependencies, so I left it out of the PR.


agostbiro:
I realized the preserve order in `serde_json` doesn't do what I thought it was -- gonna look into this more + the failing tests tomorrow.

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -212,7 +291,24 @@ impl RpcClient {
...
     pub async fn chain_id(&self) -> Result<U256, RpcClientError> {
-        self.call(&MethodInvocation::ChainId()).await
+        // It's important to drop the read guard here, otherwise we'll deadlock when trying to
+        // acquire the write guard.
+        let chain_id = { *self.chain_id.read().await };
+
+        if let Some(chain_id) = chain_id {
+            Ok(chain_id)
+        } else {
+            // It's important to call without cache here to avoid infinite recursion as this method
```
Wodann, line 301: This made me realise that there are more methods that we'd like to call without caching, as not all methods are immutable. E.g. think about the RPC methods that return the server's version, etc.

crates/rethnet_eth/Cargo.toml
```diff
@@ -19,9 +20,10 @@ revm-primitives = { git = "https://github.com/Wodann/revm", rev = "9fdf446", ver
 rlp = { version = "0.5.2", default-features = false, features = ["derive"] }
 secp256k1 = { version = "0.24.0", default-features = false, features = ["alloc", "recovery"] }
 serde = { version = "1.0.147", default-features = false, features = ["derive"], optional = true }
-serde_json = { version = "1.0.89", optional = true }
+serde_json = { version = "1.0.89", optional = true, features = ["preserve_order"] }
```
Wodann, line 23: Good find!

Wodann (changes requested):
Do you think you'd be able to add some tests that validate a transition from alchemy to infura, but for the same chain id?

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -62,111 +79,177 @@ struct BatchResponse {
...
+            .await
+            .map_err(RpcClientError::FailedToSend)?
+            .error_for_status()
+            .map_err(RpcClientError::HttpStatus)?
+            .text()
+            .await
+            .map_err(RpcClientError::CorruptedResponse)
+    }
+
+    fn request_to_string(&self, input: &MethodInvocation) -> (jsonrpc::Id, String) {
```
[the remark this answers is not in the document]
agostbiro: Thanks, tried to address in [72ccfa5](https://github.com/NomicFoundation/hardhat/pull/4270/commits/72ccfa54abc29a22a4ff85632650a5fa2f24801b)

crates/rethnet_eth/src/remote/client.rs
agostbiro: Thanks, will address in https://github.com/NomicFoundation/rethnet/issues/88

agostbiro:
> Do you think you'd be able to add some tests that validate a transition from alchemy to infura, but for the same chain id?

Yes, but this would need the cache directory to be configurable to make sure it's ran in a clean environment. I made a new PR for that: https://github.com/NomicFoundation/hardhat/pull/4280. Will merge that branch into this and add Alchemy/Infura tests here once the other PR is approved. 

Wdyt should be tested? Just a smoke test that it doesn't blow up or some specific invariants? I can't think of any right now.

>  realized the preserve order in serde_json doesn't do what I thought it was

I originally added the `preserve_order` feature to `serde_json` to make sure we serialize the request bodies consistently as we derive the cache keys from this. I learned two things since then:

1. The `preserve_order` feature won't achieve this, because it's only effective when converting between `String` and `serde_json::Value`.
2. We currently derive the RPC cache key from the serialized JSON-RPC request which includes the request id, so we need to rethink our approach here anyway. I'll do this in https://github.com/NomicFoundation/rethnet/issues/158

agostbiro:
I'll investigate the failing tests.

Wodann:
> I'll investigate the failing tests.

FYI, the failing dual tests are a known issue. There is a bug in the TypeScript that causes memory to leak and the GC to report an out-of-memory exception.

Wodann:
> Wdyt should be tested? Just a smoke test that it doesn't blow up or some specific invariants? I can't think of any right now.

For this particular task, I'd mainly like to validate that switching a provider doesn't invalidate the cache.

agostbiro:
I merged https://github.com/NomicFoundation/hardhat/pull/4294 into this branch and added a test to make sure that switching from Alchemy to Infura doesn't invalidate the RPC cache.

The PR is now ready for review @Wodann with 3 caveats:

1. https://github.com/NomicFoundation/hardhat/pull/4294 should be reviewed and merged before this one.
2. As consequence of the new cache test, now both the Alchemy and Infura API url environment variables must be set in order to run remote EDR tests. CI might fail because of this, but I haven't gotten that far yet due to the next caveat.
3. This PR has the same weird lint issue as here: https://github.com/NomicFoundation/hardhat/pull/4294#issuecomment-1682778466

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -242,9 +242,13 @@ impl RpcClient {
     where
         T: for<'a> serde::Deserialize<'a>,
     {
+        let cache_key = CacheableMethodInvocation::try_from(input)
+            .ok()
+            .and_then(|v| v.cache_key());
```
Wodann, line 247: I expect clippy to complain about this, but this can be:

```suggestion
            .and_then(CacheableMethodInvocation::cache_key);
```

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -553,15 +557,17 @@ mod tests {
         client: RpcClient,
 
         // Need to keep the tempdir around to prevent it from being deleted
-        _cache_dir: TempDir,
+        // Only accessed when feature = "test-remote", hence the allow.
+        #[allow(dead_code)]
```
Wodann, line 561:
```suggestion
        #[cfg(feature = "test-remote")]
```

How about?

Wodann (changes requested):

crates/rethnet_eth/src/remote/client.rs
agostbiro: We need the the cache dir when `test-remote` is not set as well, but the attribute itself is only accessed in code gated behind `test-remote`.

crates/rethnet_eth/src/remote/client.rs
agostbiro: I like this pattern, but it doesn't work here unfortunately, bc `CacheableMethodInvocations::cache_key` takes `&CacheableMethodInvocations` but the value is owned.

agostbiro:
I added the Infura url to the CI in [92a9c8a](https://github.com/NomicFoundation/hardhat/pull/4270/commits/92a9c8aba9012c90659eac79943da5436daaedd0). The remaining CI failures seem to be failing in `rethnet/main` as well.

Wodann (approved):

agostbiro:
CI failures are known ones 👍

PR #4279, opened 2023-08-10
Fix: artifactExists returns false for missing artifacts

### LINKED ISSUE #3758: artifactExists throws an error if the artifact doesn't exist
We should fix this as part of #3498, or after merging it.

schaable:
Closes #3758

### DIFF
--- a/.changeset/ten-trees-smoke.md
+++ b/.changeset/ten-trees-smoke.md
@@ -1,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+"hardhat": patch
+---
+
+Fixed an issue where artifactExists would throw an error for missing artifacts.
--- a/.changeset/ten-trees-smoke.md
+++ b/.changeset/ten-trees-smoke.md
@@ -1,5 +1,4 @@
 ---
-"@nomicfoundation/hardhat-verify": patch
 "hardhat": patch
 ---
 
--- a/packages/hardhat-core/src/internal/artifacts.ts
+++ b/packages/hardhat-core/src/internal/artifacts.ts
@@ -78,7 +78,18 @@ export class Artifacts implements IArtifacts {
   }
 
   public async artifactExists(name: string): Promise<boolean> {
-    const artifactPath = await this._getArtifactPath(name);
+    let artifactPath = "";
+    try {
+      artifactPath = await this._getArtifactPath(name);
+    } catch (e) {
+      if (e instanceof HardhatError) {
+        return false;
+      }
+
+      // eslint-disable-next-line @nomicfoundation/hardhat-internal-rules/only-hardhat-error
+      throw e;
+    }
+
     return fsExtra.pathExists(artifactPath);
   }
 
@@ -407,6 +418,7 @@ export class Artifacts implements IArtifacts {
 
   /**
    * Returns the absolute path to the given artifact
+   * @throws {HardhatError} If the name is not fully qualified.
    */
   public formArtifactPathFromFullyQualifiedName(
     fullyQualifiedName: string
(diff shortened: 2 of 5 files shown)


.changeset/ten-trees-smoke.md
```diff
@@ -0,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-verify": patch
```
fvictorio, line 2: The change in `hh-verify` is just a refactor without user impact (except maybe saving a couple of CPU cycles), so we shouldn't include it in the changeset.

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -78,7 +78,18 @@ export class Artifacts implements IArtifacts {
   }
 
   public async artifactExists(name: string): Promise<boolean> {
-    const artifactPath = await this._getArtifactPath(name);
+    let artifactPath = "";
```
fvictorio, line 81: This is not necessary, and it could hide problems that can be caught by the compiler.

```suggestion
    let artifactPath;
```

To understand why this is the case, try commenting out the `throw e` line. If `artifactPath` is not pre-assigned, you'll (correctly) get a compilation error. But if you pre-assign it, then the code will compile, and the runtime behavior will be to call `pathExists` with an empty string.

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -78,7 +78,18 @@ export class Artifacts implements IArtifacts {
...
+    try {
+      artifactPath = await this._getArtifactPath(name);
+    } catch (e) {
+      if (e instanceof HardhatError) {
```
fvictorio, line 85: In general we use:

```suggestion
      if (HardhatError.isHardhatError(e)) {
```

To be honest, I always forget in which cases `instanceof` doesn't work. @alcuadrado do you remember this?

packages/hardhat-core/src/internal/artifacts.ts
```diff
@@ -490,6 +502,11 @@ export class Artifacts implements IArtifacts {
    * If the name is fully qualified, the path is computed from it.  If not, an
    * artifact that matches the given name is searched in the existing artifacts.
    * If there is an ambiguity, an error is thrown.
+   *
+   * @throws {HardhatError} with descriptor:
+   * - {@link ERRORS.ARTIFACTS.WRONG_CASING} if the path case doesn't match the one in the filesystem.
+   * - {@link ERRORS.ARTIFACTS.MULTIPLE_FOUND} if there are multiple artifacts matching the given contract name.
+   * - {@link ERRORS.ARTIFACTS.NOT_FOUND} if the artifact is not found.
```
fvictorio, line 509: Thanks for adding these jsdocs, they will be very helpful.

packages/hardhat-core/test/internal/artifacts.ts
```diff
@@ -1309,6 +1309,40 @@ describe("Artifacts class", function () {
...
+      assert.isTrue(await artifacts.artifactExists("source.sol:Interface"));
+      // non-FQN
+      assert.isTrue(await artifacts.artifactExists("Interface"));
+    });
+
+    it("should check that the artifact doesn't exist", async function () {
+      const artifacts = new Artifacts(this.tmpDir);
+
+      // FQN, missing artifact
+      assert.isFalse(await artifacts.artifactExists("invalid/source.sol:A"));
```
fvictorio, line 1339: This assertion doesn't match the comment, does it? Maybe you meant this?

```suggestion
      // FQN, missing file
      assert.isFalse(await artifacts.artifactExists("invalid/source.sol:Interface"));
      // FQN, missing artifact
      assert.isFalse(await artifacts.artifactExists("source.sol:A"));
```

schaable: That's a valid difference, but for it to be valid we'd need to generate an artifact like we do in the successful case. And it actually makes sense to add it for the rest of the assertions as well, as they don't make too much sense otherwise. Thanks!

fvictorio (approved):

Issue #3858, opened 2023-04-13
Elongated memory data for debug_traceTransaction


antazoey:
### Version of Hardhat

2.13.1

### What happened?

When upgrading from 2.13.0 to 2.13.1 or greater, I get lots of empty memory, and realistic or larger txns, I get this bug: https://github.com/NomicFoundation/hardhat/issues/3272

I am using HTTP to send txns to a Hardhat node. The `memory` field of the revert struct log is extremely massive and filled with 0-bytes. Just a bunch of 0s all the way down.

I do the same thing with ganache and anvil and the return value and struct memory make a lot more sense.

### Minimal reproduction steps

This example only works on small txns. Large one cause this bug: https://github.com/NomicFoundation/hardhat/issues/3858

* Invoke a transaction to a Hardhat node using HTTP and store the txn hash
* Call `debug_traceTransaction` using HTTP to observe the trace. Look at the last frame. It looks like this:

```
{
    'pc': 93, 
    'op': 'REVERT', 
    'gas': 29976633, 
    'gasCost': 0, 
    'depth': 1, 
    'stack': [
        '0x000000000000000000000000000000000000000000000000000000003ccfd60b', 
        '0x0000000000000000000000000000000000000000000000000000000000000033', 
        '0x0000000000000000000000000000000000000000000000000000000000000004', 
        '0x0000000000000000000000000000000000000000000000000000000000000080'], 
    'memory': [
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000080', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x82b4290000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
        '0x0000000000000000000000000000000000000000000000000000000000000000', 
[...]

PR #4286, opened 2023-08-11
Elongated memory data for debug_traceTransaction

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/3858)

When calling the method debug_traceTransaction a large amount of 0s is retuned inside the memory property. This is wrong.

E.g.: 
The memory should stop after the 5th line, the one that starts with "0x82b429...."

image

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
@@ -319,11 +319,13 @@ export class VMDebugTracer {
   }
 
   private _getMemory(step: InterpreterStep): string[] {
-    const memory = Buffer.from(step.memory)
-      .toString("hex")
-      .match(/.{1,64}/g);
+    const rawMemory =
+      Buffer.from(step.memory)
+        .toString("hex")
+        .match(/.{1,64}/g) ?? [];
 
-    return memory === null ? [] : memory.slice(0, Number(step.memoryWordCount)); // Remove the additional non allocated memory
+    // Remove the additional non allocated memory
+    return rawMemory.slice(0, Number(step.memoryWordCount));
   }
 
   private _getStack(step: InterpreterStep): string[] {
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
```diff
@@ -319,9 +323,9 @@ export class VMDebugTracer {
       .toString("hex")
       .match(/.{1,64}/g);
 
-    const result = memory === null ? [] : [...memory];
-
-    return result;
+    return memory === null
+      ? []
+      : memory?.slice(0, Number(step.memoryWordCount)); // Remove the additional non allocated memory
```
schaable, line 328: if `memory` is not `null`, [it has to be an array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#return_value). You can drop the optional chaining.

schaable (approved):

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-debug-tracer.ts
```diff
@@ -319,9 +323,7 @@ export class VMDebugTracer {
       .toString("hex")
       .match(/.{1,64}/g);
 
-    const result = memory === null ? [] : [...memory];
-
-    return result;
+    return memory === null ? [] : memory.slice(0, Number(step.memoryWordCount)); // Remove the additional non allocated memory
```
fvictorio, line 326: I can't add a code suggestion for the whole block, but what do you think about this? I think its more readable

```js
    const rawMemory =
      Buffer.from(step.memory)
        .toString("hex")
        .match(/.{1,64}/g) ?? [];

    // Remove the additional non allocated memory
    return rawMemory.slice(0, Number(step.memoryWordCount));
```

fvictorio (approved):

PR #4289, opened 2023-08-11
fix: fix EDR RPC cache key (WIP)

agostbiro:
This is a WIP PR to make sure that I'm opening to make sure that the direction for solving the problem makes sense. Tracking issue: https://github.com/NomicFoundation/rethnet/issues/158. I branched out from https://github.com/NomicFoundation/hardhat/pull/4280.

There are two problems currently with how we derive the RPC cache key:

1. It [includes](https://github.com/NomicFoundation/hardhat/blob/01414c524b711991a3360540a170c7d2356abf6d/crates/rethnet_eth/src/remote/client.rs#L110) the request id which makes it ineffective.
2. It derives the key from the hash of the JSON serialization of the request parameters. 

The first problem is easy to solve, but the second is more subtle. The problem with using the JSON serialization as part of the cache key is that there is no guarantee that two objects that are logically equivalent serialize to the same JSON representation. More specifically, JSON serialization doesn't guarantee that

`a == b => json(a) == json(b)`

Because of this, it's possible that the cache works as expected in our tests, but some users experience more frequent cache misses with some specific configuration in the wild. This could lead to hard to diagnose performance degradation so we should avoid it if possible.

In Rust, the [`Hash`](https://doc.rust-lang.org/std/hash/trait.Hash.html) trait provides the guarantee that 

`a == b => hash(a) == hash(b)`

so ideally we derive our RPC cache key from the hash of the method name and the parameters. Since these are jointly implemented by the `MethodInvocation` enum, my first attempt was to derive `Hash` for this enum and than use a hash of the `MethodInvocation` as the RPC cache key.

This didn't compile at first as some parameters didn't implement `Hash`. I could fix this mostly by adding `Hash` derives to the parameters (I also added `Eq` where I added `Hash` as they should be implemented hand-in-hand).

There were two cases where deriving `Hash` was not possible: the `f64` parameters of `eth_feeHistory` and the `serde_json::Value` parameter of `eth_signTypedData`.

`f64` doesn't implement `Hash` because of the equality issues with `NaN`. I solved this by introducing a new `Percentile` type that checks that the percentile value is in the supported range [0, 100]. Because of this check we can safely implement `Hash` for `Percentile`.

The `serde_json::Value` parameter of `eth_signTypedData` is more difficult to handle. We could introduce a new type around it and implement hashing based on a canonical JSON serializer that guarantees `a == b => canonical_json(a) == canonical_json(b)`, but there is no canonical canonical json (hehe) crate in the Rust ecosystem. The existing crates seem barely used and canonical json serialization is tricky to even define properly, and even more tricky to implement correctly, so we couldn't have high confidence that our solution is correct if we went down this path.

Taking a step back, `eth_signTypedData` shouldn't be cached so we don't really need to be able to implement `Hash` for it. We need a way to differentiate methods that can be cached anyway for https://github.com/NomicFoundation/rethnet/issues/88, so I took the approach of separating methods in `MethodInvocation` into cacheable and non-cacheable:

```rust
/// For an invoking a method on a remote ethereum node
#[derive(Clone, Debug, PartialEq, Eq, serde::Deserialize, serde::Serialize)]
#[serde(untagged)]
pub enum MethodInvocation {
    /// A potentially cacheable method invocation
    Cacheable(CacheableMethodInvocation),
    /// A non-cacheable method invocation
    NonCacheable(NonCacheableMethodInvocation),
}
```

Where only `CacheableMethodInvocation` implements `Hash`. 

This change is in this draft PR, but it's a pretty one so I wanted to ask your opinion @Wodann @feuGeneA to make sure it makes sense to you before I continue implementing the cache key fix based on this.

Also, I wanted to ask if the `evm_*` method should stay in `MethodInvocation` as they cannot be invoked on a remote node.

### DIFF
--- a/crates/rethnet_eth/src/remote/methods.rs
+++ b/crates/rethnet_eth/src/remote/methods.rs
@@ -384,9 +384,6 @@ pub enum PercentileError {
     /// Out of bounds value error
     #[error("Percentile must be between 0 and 100 inclusive, instead it is {0}")]
     OutOfBounds(f64),
-    /// Non-finite value error
-    #[error("Percentile must be a finite value, instead it is {0}")]
-    NonFiniteValue(f64),
 }
 
 /// The `rewardPercentiles` argument for `eth_feeHistory`.
@@ -427,9 +424,7 @@ impl TryFrom<f64> for Percentile {
     type Error = PercentileError;
 
     fn try_from(value: f64) -> Result<Self, Self::Error> {
-        if !value.is_finite() {
-            Err(PercentileError::NonFiniteValue(value))
-        } else if !(0. ..=100.).contains(&value) {
+        if !(0. ..=100.).contains(&value) {
             Err(PercentileError::OutOfBounds(value))
         } else {
             Ok(Self(value))
@@ -471,11 +466,12 @@ mod tests {
     #[test]
     fn test_infinite_percentile() {
         assert!(Percentile::try_from(f64::INFINITY).is_err());
+        assert!(Percentile::try_from(f64::NEG_INFINITY).is_err());
     }
 
     #[test]
     fn test_nan_percentile() {
-        assert!(Percentile::try_from(f64::INFINITY).is_err());
+        assert!(Percentile::try_from(f64::NAN).is_err());
     }
 
     #[test]
--- a/crates/rethnet_eth/src/remote/methods.rs
+++ b/crates/rethnet_eth/src/remote/methods.rs
@@ -10,10 +10,11 @@ use crate::{
     },
     Address, B256, U256,
 };
+use std::hash::{Hash, Hasher};
 
 /// for specifying input to methods requiring a transaction object, like `eth_call`,
 /// `eth_sendTransaction` and `eth_estimateGas`
-#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Hash, serde::Deserialize, serde::Serialize)]
 pub struct TransactionInput {
     /// the address from which the transaction should be sent
     pub from: Option<Address>,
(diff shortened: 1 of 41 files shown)


agostbiro:
Fixed the merge conflicts after the base branch got merged, diff view is back to normal 👍

crates/rethnet_eth/src/remote/methods.rs
```diff
@@ -349,9 +378,78 @@ pub struct GetLogsInput {
...
+            .into_iter()
+            .map(Percentile::try_from)
+            .collect::<Result<Vec<_>, _>>()?;
+        Ok(Self(percentiles))
+    }
+}
+
+/// A reward percentile value for `eth_feeHistory`. It's a floating point value in the range [0,
+/// 100] as per the Ethereum docs: https://ethereum.github.io/execution-apis/api-documentation/
+#[derive(Clone, Copy, Debug, serde::Deserialize, serde::Serialize)]
```
agostbiro, line 409: Deriving `Deserialize` here is a mistake as it'd allow constructing an instance bypassing the range check. Will fix this

Wodann:
I'm a big proponent of the single responsibility principle and try to enforce that in our codebase. It reduces complexity, thus improving testability, increasing readability, and reducing bug complexity.

With the proposal to incorporate the notion of cachability into the `enum MethodInvocation`, we're polluting that API with an implementation detail. To avoid this, I'd split the cached type from the JSON-RPC type and keep that private.

I was thinking about how to resolve the issue with caching of a call to `RpcClient` that uses a tag. One option is to pass in a resolver that uses the response message when resolving the input. Potentially we can combine this with the internal hash type representation?

E.g. if we have

```rs
fn call<ResponseT, ResolverT>(input: &MethodInvocation, resolver: ResolverT) -> Result<ResponseT, RpcClientError>
  where
    ResponseT: for<'a> serde::Deserialize<'a>,
    ResolverT: FnOnce(ResponseT) -> impl Hash // you'd need to fill in the impl Hash with your own type
```

you could have a closure:

```rs
move |response: eth::Block<B256>| {
  CachedMethod::GetBlockByNumber(response.number, request.1)
}
```

This should naturally only run if the response is a success type. Also, consider that the response might not be resolvable, in which case we should not cache

Issue #4249, opened 2023-08-07
Project doesn't compile with latest dependencies


fvictorio:
The latest dependency workflow is failing because of some compilation error. I don't know the underlying reason  yet.

PR #4293, opened 2023-08-14
Add explicit any type to some undici responses

fvictorio:
Fixes https://github.com/NomicFoundation/hardhat/issues/4249.

In the latest version of `undici`, the type of `body.json()` is `Promise<unknown>` instead of `Promise<any>`. This change just uses an explicit `any` type in the places that were relying on that behavior.

### DIFF
--- a/packages/hardhat-verify/src/internal/etherscan.ts
+++ b/packages/hardhat-verify/src/internal/etherscan.ts
@@ -97,7 +97,7 @@ export class Etherscan {
     url.search = parameters.toString();
 
     const response = await sendGetRequest(url);
-    const json = await response.body.json();
+    const json: any = await response.body.json();
 
     if (json.message !== "OK") {
       return false;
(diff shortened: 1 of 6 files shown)


packages/hardhat-verify/src/internal/etherscan.ts
```diff
@@ -97,7 +97,7 @@ export class Etherscan {
     url.search = parameters.toString();
 
     const response = await sendGetRequest(url);
-    const json = await response.body.json();
+    const json: any = await response.body.json();
```
schaable, line 100: are these lines working? Because the `EtherscanResponse` constructor takes an `any`.

```
const etherscanResponse = new EtherscanResponse(await response.body.json());
```

https://github.com/NomicFoundation/hardhat/pull/4293/files#diff-87b22bea7502ad3ea606dc40db78f8e4e4837e1fcb849bb3ffa183976a4cd0f6L163

https://github.com/NomicFoundation/hardhat/pull/4293/files#diff-87b22bea7502ad3ea606dc40db78f8e4e4837e1fcb849bb3ffa183976a4cd0f6L216

schaable (approved):
just one question, feel free to merge it if you think it's not an issue.

packages/hardhat-verify/src/internal/etherscan.ts
fvictorio: Yes beause you can assign anything to an `any` value, including an `unknown` value.

PR #4294, opened 2023-08-14
fix: make EDR RPC cache key unique

agostbiro:
This PR is a follow up from https://github.com/NomicFoundation/hardhat/pull/4289 with a new approach to making remote method invocations cacheable: by introducing a new internal `CacheableMethodInvocation` type. It's still WIP.


crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
```diff
@@ -0,0 +1,154 @@
+use crate::remote::methods::{GetLogsInput, MethodInvocation, Percentiles, TransactionInput};
```
Wodann, line 1: In a next PR, I'd recommend moving the RPC client and caching related files out into a `rethnet_rpc_client` crate

crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
```diff
@@ -0,0 +1,154 @@
...
+use revm_primitives::{Address, B256};
+
+/// These method invocations are hashable and can be potentially cached.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+// Using repr(u8) to let us assign a number to each variant which is important to make sure hashing
+// remains consistent if new variants are added or if variants are reordered.
+#[repr(u8)]
+pub(super) enum CacheableMethodInvocation<'a> {
+    /// eth_call
+    Call(&'a TransactionInput, &'a Option<BlockSpec>) = 0,
```
Wodann, line 13: `BlockSpec` is not deterministic. When it is used, the "earliest", "safe", "latest", etc. tags need to be resolved based on the received response.

I'd recommend using an `Option<U256>` for the block number here instead. Also, see my comment on the previous PR about resolvers

crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
```diff
@@ -0,0 +1,154 @@
...
+    /// eth_chainId
+    ChainId() = 1,
+    /// eth_estimateGas
+    EstimateGas(&'a TransactionInput, &'a Option<BlockSpec>) = 2,
+    /// eth_feeHistory
+    FeeHistory(
```
Wodann, line 19: This wasn't possible for the RPC variants, as it needed to support a specific JSON format that required sequences, but here we can using names for the variant members

crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
```diff
@@ -0,0 +1,154 @@
...
+/// Error type for [`CacheableMethodInvocation::try_from`].
+#[derive(thiserror::Error, Debug)]
+#[error("Method is not cacheable: {0:?}")]
+pub(super) struct MethodNotCacheableError(MethodInvocation);
+
+impl<'a> TryFrom<&'a MethodInvocation> for CacheableMethodInvocation<'a> {
+    type Error = MethodNotCacheableError;
+
+    fn try_from(value: &'a MethodInvocation) -> Result<Self, Self::Error> {
+        use CacheableMethodInvocation as CMI;
```
Wodann, line 80: Even though I understand the desire to avoid retyping of the same long type name here, we are actively moving away from non-standard acronyms in our codebase to keep it more readable. Even our generic types and lifetimes are becoming full names.

I say non-standard because things like `Eth` or `IO` are completely acceptable.

crates/rethnet_eth/src/remote/methods.rs
```diff
@@ -364,4 +441,98 @@ mod tests {
...
+        assert!(Percentile::try_from(0.).is_ok());
+    }
+
+    #[test]
+    fn test_hundred_percentile() {
+        assert!(Percentile::try_from(100.).is_ok());
+    }
+
+    #[test]
+    fn test_percentile_serialization() {
```
Wodann, line 476: Couldn't you combine the two tests into a round-trip test where you serialize an input, deserialize its output, and compare it to the source input?

crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
agostbiro: Thanks, I'm planning to address this in https://github.com/NomicFoundation/rethnet/issues/88, so I'd say it's out of scope now unless you want me to combine the two PRs into one.

crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
agostbiro: Ok, added follow up issue: https://github.com/NomicFoundation/rethnet/issues/161

Wodann (approved):

agostbiro:
This is now ready for review @Wodann . The PR addresses three issues:

1. Makes sure that the request id is not included in the cache key.
2. It switches from hashing a JSON string to hashing Rust types to ensure that equivalent requests always hash to the same value irrespective of JSON key order, but without potential errors due to `None` being ill-defined in JSON.
3. It switches the hashing algorithm from one with 64-bit output space to one with 256-bit output space as the former can have collisions with tens of millions of blocks.

I tried to minimize changes in the client to avoid conflicts with https://github.com/NomicFoundation/hardhat/pull/4270

agostbiro:
There is a weird caching issue going on with the CI. The lints fails with the following error:

```
pattern `&MethodInvocation::EvmSetIntervalMining(_)` not covered
```

But the `MethodInvocation::EvmSetIntervalMining` variant is not in this branch.

Wodann:
> There is a weird caching issue going on with the CI. The lints fails with the following error:
> 
> ```
> pattern `&MethodInvocation::EvmSetIntervalMining(_)` not covered
> ```
> 
> But the `MethodInvocation::EvmSetIntervalMining` variant is not in this branch.

I cleared the caches to validate your suspicion, but the CI is still failing. Then I looked at the code and that call does in fact exist [here](https://github.com/NomicFoundation/hardhat/blob/rethnet/main/crates/rethnet_eth/src/remote/methods.rs#L315).

crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
```diff
@@ -0,0 +1,680 @@
...
+            | MethodInvocation::EvmSetNextBlockTimestamp(_)
+            | MethodInvocation::EvmSnapshot() => Err(MethodNotCacheableError(value.clone())),
+        }
+    }
+}
+
+/// Potentially cacheable Ethereum JSON-RPC method invocations for a batch call.
+#[derive(Clone, Debug)]
+#[repr(transparent)]
+pub(super) struct CacheableMethodInvocations<'a>(Vec<CacheableMethodInvocation<'a>>);
```
Wodann, line 191: I'd consider caching individual parts of the batch call separate when you introduce a resolution phase. That will allow you to get more cache hits and simplifies the cache

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -108,11 +111,11 @@ impl RpcClient {
             .create(directory.clone())
             .expect("failed to create on-disk RPC response cache");
 
-        Path::new(&directory).join(format!("{}.json", Self::hash_string(request_body)))
+        Path::new(&directory).join(format!("{}.json", cache_key))
     }
 
-    fn read_response_from_cache(&self, request_body: &str) -> Option<String> {
-        if let Ok(mut file) = std::fs::File::open(self.make_cache_path(request_body)) {
+    fn read_response_from_cache(&self, cache_key: Option<&CacheKey>) -> Option<String> {
```
Wodann, line 117: Why would you pass in an option to a function if it is always guaranteed to fail?

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -124,14 +127,20 @@ impl RpcClient {
...
     /// returns response text
-    async fn send_request_body(&self, request_body: &str) -> Result<String, RpcClientError> {
-        if let Some(cached_response) = self.read_response_from_cache(request_body) {
+    async fn send_request_body(
+        &self,
+        request_body: &str,
+        cache_key: Option<CacheKey>,
+    ) -> Result<String, RpcClientError> {
+        if let Some(cached_response) = self.read_response_from_cache(cache_key.as_ref()) {
             Ok(cached_response)
```
Wodann, line 144:
```suggestion
        if let Some(cache_key) = &cache_key {
            self.read_response_from_cache(cache_key)
```

Wodann (changes requested):

crates/rethnet_eth/src/remote/client.rs
agostbiro: In order to minimize conflicts with https://github.com/NomicFoundation/hardhat/pull/4270. It's fixed there.

crates/rethnet_eth/src/remote/client.rs
agostbiro: I would like to avoid making this change to minimize conflicts with #4270. It's fixed there.

crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
agostbiro: Good idea, I'll try to include it 👍

crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
```diff
@@ -0,0 +1,680 @@
...
+
+        for method_invocation in methods {
+            this = this.hash_method_invocation(method_invocation)?;
+        }
+
+        Some(this)
+    }
+
+    fn finalize(self) -> Option<CacheKey> {
+        // Returns Option for chaining convenience.
```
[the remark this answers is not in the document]
agostbiro: Thanks good catch! Fixed in [0b44369](https://github.com/NomicFoundation/hardhat/pull/4294/commits/0b44369b890ae09c519bd4283c635aa2122b4ef4). (The comment and the `Option` return type there are a left over from when the code had different structure and otherwise would've had to wrap `finalize` calls in `Some` all the time in this module.)

In other places like [this](https://github.com/NomicFoundation/hardhat/pull/4294/files#diff-cf941b21bd00797395a8590c975818931f27bb21a221df240c931cfd0b317a9cR402) where the return value is `Some(result)` we are using short-circuiting with `?`, so those methods can return `None`.

agostbiro:
> I cleared the caches to validate your suspicion, but the CI is still failing. Then I looked at the code and that call does in fact exist [here](https://github.com/NomicFoundation/hardhat/blob/rethnet/main/crates/rethnet_eth/src/remote/methods.rs#L315).

Thanks for looking into this! Figured out what happened:

1. I finished working on this PR and merged my local `rethnet/main` into it.
2. https://github.com/NomicFoundation/hardhat/pull/4243 got merged into `rethnet/main` introducing `MethodInvocation::EvmSetIntervalMining`
3. I pushed my changes to this PR.
4. The CI failed for this PR as it runs on the result of merging this branch into `rethnet/main` (as opposed to the latest commit in this branch).

I'll keep the last point in mind next time 👍

crates/rethnet_eth/src/remote/cacheable_method_invocation.rs
```diff
@@ -0,0 +1,680 @@
...
+            GetTransactionReceipt { transaction_hash } => this.hash_b256(transaction_hash),
+            NetVersion => this,
+        };
+
+        Some(this)
+    }
+
+    fn hash_method_invocations(self, methods: &CacheableMethodInvocations<'_>) -> Option<Self> {
+        // Make sure it's prefix-free
+        let mut this = self.hash_usize(methods.len());
```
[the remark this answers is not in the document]
Wodann: I'd always go with the most descriptive, self-documenting name, but I guess there is an argument to be made that `hasher` might be construed as the inner one, so I'm fine either way

Wodann:
This seems to have uncovered a bug in `ForkState`, as we're not storing irregular state changes separately: https://github.com/NomicFoundation/hardhat/actions/runs/5901750446/job/16008533360#step:9:1

I don't know why the broken cache would have hidden this in the past, though 🤔

agostbiro:
Thanks for the review @Wodann , created the follow up issue here: https://github.com/NomicFoundation/rethnet/issues/170

PR #4298, opened 2023-08-15
refactor: add EIP-155 to transaction types

Wodann:


crates/rethnet_eth/src/remote/eth.rs
```diff
@@ -172,19 +172,37 @@ impl TryFrom<Transaction> for (SignedTransaction, Address) {
...
-                value: value.value,
-                input: value.input,
-                signature: Signature {
-                    r: value.r,
-                    s: value.s,
-                    v: value.v,
-                },
-            }),
+            0 => {
+                if value.v > 36 {
```
agostbiro, line 176: I saw this check in two places, maybe we could add a function/method for deciding whether `v` is an EIP-155 value?

crates/rethnet_eth/src/transaction/signed/eip155.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /// Recovers the Ethereum address which was used to sign the transaction.
+    pub fn recover(&self) -> Result<Address, SignatureError> {
+        self.signature
+            .recover(EIP155TransactionRequest::from(self).hash())
+    }
+
+    pub fn chain_id(&self) -> u64 {
+        (self.signature.v - 35) / 2
```
agostbiro, line 37: Is this correct if the parity is odd? Would be good to have some tests here

crates/rethnet_eth/src/transaction/signed/eip155.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /// See <https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md>
+    /// > If you do, then the v of the signature MUST be set to {0,1} + `CHAIN_ID` * 2 + 35 where
+    /// > {0,1} is the parity of the y value of the curve point for which r is the x-value in the
+    /// > secp256k1 signing process.
+    pub fn meets_eip155(&self, chain_id: u64) -> bool {
```
agostbiro, line 44: Should we reject `chain_id == 0`? Would be good to have some tests here as well for edge cases

crates/rethnet_evm_napi/src/block.rs
```diff
@@ -308,7 +308,10 @@ impl Block {
             .iter()
             .map(|transaction| match transaction {
                 rethnet_eth::transaction::SignedTransaction::Legacy(transaction) => {
-                    LegacySignedTransaction::new(&env, transaction).map(Either3::A)
+                    LegacySignedTransaction::from_legacy(&env, transaction).map(Either3::A)
+                }
+                rethnet_eth::transaction::SignedTransaction::EIP155(transaction) => {
+                    LegacySignedTransaction::from_eip155(&env, transaction).map(Either3::A)
```
agostbiro, line 314: I found this a bit confusing. Are we returning a `LegacySignedTransaction` here on purpose?

crates/rethnet_evm_napi/src/transaction/pending.rs
```diff
@@ -82,7 +82,10 @@ impl PendingTransaction {
     > {
         match &*self.inner {
             rethnet_eth::transaction::SignedTransaction::Legacy(transaction) => {
-                LegacySignedTransaction::new(&env, transaction).map(Either3::A)
+                LegacySignedTransaction::from_legacy(&env, transaction).map(Either3::A)
+            }
+            rethnet_eth::transaction::SignedTransaction::EIP155(transaction) => {
+                LegacySignedTransaction::from_eip155(&env, transaction).map(Either3::A)
```
agostbiro, line 88: I'm also confused here if returning `LegacySignedTransaction` is correct

agostbiro (changes requested):
Thanks for the PR @Wodann ! I left some comments

crates/rethnet_evm_napi/src/block.rs
Wodann: Yes, javascript doesn't have strongly typed enums. Instead Typescript determines which variant something is based on the presence of variant fields. Given that both the EIP-155 and Legacy versions have the same fields, but different interpretations, we cannot express this in Typescript/Javascript.

I originally did, but it introduced a bug where both Legacy and EIP-155 signed transactions were interpreted as the former (or rather, the first in the list of enum variants that matches all fields)

crates/rethnet_evm_napi/src/transaction/pending.rs
Wodann: https://github.com/NomicFoundation/hardhat/pull/4298#discussion_r1295003527

crates/rethnet_eth/src/transaction/signed/eip155.rs
Wodann: I'm happy to add tests, but would you mind if I do that in https://github.com/NomicFoundation/hardhat/pull/4300/files/3c85628c4f99db9b9274efecce0558d937756f5b#diff-6ba4eabcc2f6a81547196c40a9e9a9ecdc9a53d1a3b6a75012953e6f1960fe0b, since that adds a bunch of tests and would cause conflicts if I also add a module here 😅

crates/rethnet_eth/src/remote/eth.rs
Wodann: Addressed in https://github.com/NomicFoundation/hardhat/pull/4298/commits/99653bcb79aee565cae47c479caa4a7c03121da9

crates/rethnet_eth/src/transaction/signed/eip155.rs
Wodann: I removed these functions in https://github.com/NomicFoundation/hardhat/pull/4298/commits/99653bcb79aee565cae47c479caa4a7c03121da9, as they are no longer necessary.

crates/rethnet_eth/src/transaction/signed/eip155.rs
agostbiro: Sure, sounds good!

Wodann:
Closed in favour of a combined PR: https://github.com/NomicFoundation/hardhat/pull/4300

PR #4301, opened 2023-08-15
fix: next base fee calculation

Wodann:


crates/rethnet_evm/src/miner.rs
```diff
@@ -177,11 +181,8 @@ fn calculate_next_base_fee(parent: &Header) -> U256 {
             let delta = parent_base_fee * gas_used_delta
                 / parent_gas_target
                 / base_fee_max_change_denominator;
-            if delta > U256::from(1) {
-                delta
-            } else {
-                parent_base_fee + U256::from(1)
-            }
+
+            parent_base_fee + delta.max(U256::from(1))
```
agostbiro, line 185: Code looks correct, but it would good to add some tests for it if you have time. We could adapt the test vector from Reth: https://github.com/paradigmxyz/reth/blob/0840acd87b4c784542d9200a050d79194c4e5031/crates/primitives/src/basefee.rs#L34

agostbiro (approved):
Approving in the interest of moving things along, but tests would be good to have if you have time: https://github.com/NomicFoundation/hardhat/pull/4301#discussion_r1294604669

crates/rethnet_evm/src/miner.rs
Wodann: Addressed in https://github.com/NomicFoundation/hardhat/pull/4301/commits/f585ff722e34c7c3f6349afed08e165c31d30916

PR #4303, opened 2023-08-16
fix: broken CI tests

Wodann:


agostbiro:
Thanks for this, @Wodann ! Code looks good, but looks like there are still some tests failing in dual mode. Is there anything we can do about those?

agostbiro (approved):
Ah got it, saw some error traces in the failing tests that looked like they were a failing expectation, but I see now that those are just logs from succeeding tests. After scrolling down I found the out of memory exception. 👍

PR #4305, opened 2023-08-16
Docs Improvement : add some more information for user knowledge.

VivekKadvani:

---


fvictorio:
Hi @VivekKadvani, thanks for the PR but I think it's safe to assume that most people know what "coverage" is and how reports are read.

Adding explanations and context is useful, but we have to draw a line somewhere and assume which things are too basic (otherwise we would end up explaining, say, what a variable is). I think this falls on the side of "too basic", and so adding an explanation is not worth the extra text.

PR #4315, opened 2023-08-21
feat: cache batch requests per method call

agostbiro:
This PR contains changes that are a prerequisite for https://github.com/NomicFoundation/rethnet/issues/88:

- Per call cache resolution of batch requests
- Store only the call results instead of the JSON-RPC responses in the on-disk cache
- Some refactoring to simplify things

### DIFF
--- a/crates/rethnet_eth/src/remote/client.rs
+++ b/crates/rethnet_eth/src/remote/client.rs
@@ -6,6 +6,7 @@ use std::{
 };
 
 use bytes::Bytes;
+use futures::{stream, StreamExt};
 use itertools::{izip, Itertools};
 use revm_primitives::{AccountInfo, Address, Bytecode, B256, KECCAK_EMPTY, U256};
 use serde::de::DeserializeOwned;
@@ -25,6 +26,9 @@ use super::{
 };
 
 const RPC_CACHE_DIR: &str = "rpc_cache";
+// More than 16 parallel reads does not significantly improve performance on any disk/workload, but
+// it can cause slow downs based on this article <https://pkolaczk.github.io/disk-parallelism/>.
+const PARALLEL_DISK_READS: usize = 16;
 
 /// Specialized error types
 #[derive(Debug, thiserror::Error)]
@@ -376,12 +380,13 @@ impl RpcClient {
             })
             .collect::<Vec<_>>();
 
-        let mut results: Vec<Option<serde_json::Value>> =
-            Vec::with_capacity(method_invocations.len());
-        for cache_key in cache_keys.iter() {
-            let result = self.try_from_cache(cache_key.as_ref()).await?;
-            results.push(result)
-        }
+        let mut results: Vec<Option<serde_json::Value>> = stream::iter(&cache_keys)
+            .map(|cache_key| async { self.try_from_cache(cache_key.as_ref()).await })
+            .buffered(PARALLEL_DISK_READS)
+            .collect::<Vec<Result<_, RpcClientError>>>()
+            .await
+            .into_iter()
+            .collect::<Result<Vec<_>, RpcClientError>>()?;
 
         let mut requests: Vec<SerializedRequest> = Vec::new();
         let mut id_to_index = HashMap::<&Id, usize>::new();
--- a/crates/rethnet_eth/src/remote/client.rs
+++ b/crates/rethnet_eth/src/remote/client.rs
@@ -1,4 +1,4 @@
-use std::collections::VecDeque;
+use std::collections::{HashMap, VecDeque};
 use std::path::{Path, PathBuf};
 use std::{
     io,
@@ -72,6 +72,8 @@ pub enum RpcClientError {
         method_invocation: MethodInvocation,
         /// The id of the request iwth the missing response
         id: Id,
+        /// The response text
+        response: String,
     },
 
     /// The JSON-RPC returned an error.
@@ -382,11 +384,15 @@ impl RpcClient {
         }
 
         let mut requests: Vec<SerializedRequest> = Vec::new();
-        for (id, method_invocation, cache_response) in izip!(&ids, method_invocations, &results) {
+        let mut id_to_index = HashMap::<&Id, usize>::new();
+        for (index, (id, method_invocation, cache_response)) in
+            izip!(&ids, method_invocations, &results).enumerate()
+        {
             if cache_response.is_none() {
                 let request =
                     Self::serialize_request_with_id(method_invocation.clone(), id.clone())?;
                 requests.push(request);
+                id_to_index.insert(id, index);
             }
         }
 
@@ -398,9 +404,9 @@ impl RpcClient {
             Self::parse_response_str(&remote_response)?;
 
         for response in remote_responses {
-            let index = ids
-                .iter()
-                .position(|id| id == &response.id)
+            let index = id_to_index
+                // Remove to make sure no duplicate ids in response
+                .remove(&response.id)
                 .ok_or_else(|| RpcClientError::InvalidId {
                     response: remote_response.clone(),
                     id: response.id,
@@ -429,6 +435,7 @@ impl RpcClient {
                 result.ok_or_else(|| RpcClientError::MissingResponse {
                     method_invocation: method_invocations[index].clone(),
                     id: ids[index].clone(),
+                    response: remote_response.clone(),
                 })
             })
             .collect()
--- a/crates/rethnet_eth/src/remote/client.rs
+++ b/crates/rethnet_eth/src/remote/client.rs
@@ -178,8 +178,8 @@ impl RpcClient {
     async fn make_cache_path(&self, cache_key: &CacheKey) -> Result<PathBuf, RpcClientError> {
         let chain_id = self.chain_id().await?;
 
-        // TODO We should use a human readable name for the chain id as directory name once
-        // available.
+        // TODO We should use a human readable name for the chain id
+        // Tracking issue: <https://github.com/NomicFoundation/rethnet/issues/172>
         let directory = self
             .rpc_cache_dir
             .join(Self::hash_bytes(chain_id.as_le_bytes().as_ref()));
--- a/crates/rethnet_eth/src/remote/client.rs
+++ b/crates/rethnet_eth/src/remote/client.rs
@@ -193,16 +193,6 @@ impl RpcClient {
         Ok(path)
     }
 
-    /// Don't fail the request, just log an error if we fail to read from cache.
-    fn log_cache_error(cache_key: CacheKey, message: &'static str, error: impl Into<CacheError>) {
-        let cache_error = RpcClientError::CacheError {
-            message: message.to_string(),
-            cache_key: cache_key.to_string(),
-            error: error.into(),
-        };
-        log::error!("{cache_error}");
-    }
-
     async fn read_response_from_cache(
         &self,
         cache_key: &CacheKey,
@@ -212,7 +202,7 @@ impl RpcClient {
             Ok(contents) => match serde_json::from_str(&contents) {
                 Ok(value) => Ok(Some(value)),
                 Err(error) => {
-                    Self::log_cache_error(
+                    log_cache_error(
                         cache_key.clone(),
                         "failed to deserialize item from RPC response cache",
                         error,
@@ -224,7 +214,7 @@ impl RpcClient {
             Err(error) => {
                 match error.kind() {
                     io::ErrorKind::NotFound => (),
-                    _ => Self::log_cache_error(
+                    _ => log_cache_error(
                         cache_key.clone(),
                         "failed to read from RPC response cache",
                         error,
@@ -240,7 +230,7 @@ impl RpcClient {
         match tokio::fs::remove_file(path).await {
             Ok(_) => Ok(()),
             Err(error) => {
-                Self::log_cache_error(
+                log_cache_error(
                     cache_key.clone(),
                     "failed to remove from RPC response cache",
                     error,
@@ -273,7 +263,7 @@ impl RpcClient {
         match tokio::fs::write(cache_path, contents).await {
             Ok(_) => (),
             Err(error) => {
-                Self::log_cache_error(
+                log_cache_error(
                     cache_key.clone(),
                     "failed to write to RPC response cache",
                     error,
@@ -602,6 +592,16 @@ impl RpcClient {
     pub async fn network_id(&self) -> Result<U256, RpcClientError> {
         self.call(MethodInvocation::NetVersion()).await
     }
+}
+
+/// Don't fail the request, just log an error if we fail to read/write from cache.
+fn log_cache_error(cache_key: CacheKey, message: &'static str, error: impl Into<CacheError>) {
+    let cache_error = RpcClientError::CacheError {
+        message: message.to_string(),
+        cache_key: cache_key.to_string(),
+        error: error.into(),
+    };
+    log::error!("{cache_error}");
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
--- a/crates/rethnet_eth/src/remote/client.rs
+++ b/crates/rethnet_eth/src/remote/client.rs
@@ -1,18 +1,21 @@
+use std::collections::VecDeque;
 use std::path::{Path, PathBuf};
 use std::{
     io,
     sync::atomic::{AtomicU64, Ordering},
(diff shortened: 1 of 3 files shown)


crates/rethnet_eth/src/remote/client.rs
```diff
@@ -226,67 +300,148 @@ impl RpcClient {
...
+            serde_json::to_value(&requests).map_err(RpcClientError::InvalidJsonRequest)?,
+        );
+        let remote_response = self.send_request_body(&request_body).await?;
+        let remote_responses: Vec<jsonrpc::Response<serde_json::Value>> =
+            Self::parse_response_str(&remote_response)?;
+
+        for response in remote_responses {
+            let index = ids
+                .iter()
+                .position(|id| id == &response.id)
```
agostbiro, line 413: Currently we have `n = 3`, so the quadratic complexity here is fine, but we might wanna fetch a lot of blocks at once in batch requests later on, so I'll rewrite this tomorrow to use a better data structure.

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -141,36 +184,76 @@ impl RpcClient {
...
+                cache_key: cache_key.to_string(),
+                error: error.into(),
             })?;
 
         let path = Path::new(&directory).join(format!("{}.json", cache_key));
         Ok(path)
     }
 
+    /// Don't fail the request, just log an error if we fail to read from cache.
+    fn log_cache_error(cache_key: CacheKey, message: &'static str, error: impl Into<CacheError>) {
```
Wodann, line 197: As this is not linked to the struct, make it a free function within the file.

I follow this heuristic: https://drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -226,67 +300,148 @@ impl RpcClient {
...
-        SerializedRequest { id, request }
+        Ok(SerializedRequest(request))
     }
 
-    async fn call<T>(&self, input: &MethodInvocation) -> Result<T, RpcClientError>
-    where
-        T: for<'a> serde::Deserialize<'a>,
-    {
-        let cache_key = CacheableMethodInvocation::try_from(input)
+    async fn call<T: DeserializeOwned + Serialize>(
```
Wodann, line 331: Is there a reason that you're moving the bounds up here?

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -226,67 +300,148 @@ impl RpcClient {
...
+            })
+            .collect::<Vec<_>>();
 
-        let request_body = format!("[{}]", request_strings.join(","));
+        let mut results: Vec<Option<serde_json::Value>> =
+            Vec::with_capacity(method_invocations.len());
+        for cache_key in cache_keys.iter() {
+            let result = self.try_from_cache(cache_key.as_ref()).await?;
+            results.push(result)
+        }
```
Wodann, line 392: I don't know of any, but I figured I'd ask, in case you know: is there a way to do async code in functional code, such that you can do: `.iter().map(..).collect::<Future<Result<Vec<Option<serde_json::Value>>>>`?

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -226,67 +300,148 @@ impl RpcClient {
...
+            Self::parse_response_str(&remote_response)?;
+
+        for response in remote_responses {
+            let index = ids
+                .iter()
+                .position(|id| id == &response.id)
+                .ok_or_else(|| RpcClientError::InvalidId {
+                    response: remote_response.clone(),
+                    id: response.id,
+                })?;
```
Wodann, line 417: Would it help if we are returning the "expected" and "actual" IDs as variables?

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -527,20 +594,25 @@ impl RpcClient {
...
 }
 
-#[derive(Debug)]
-struct SerializedRequest {
-    id: jsonrpc::Id,
-    request: String,
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[repr(transparent)]
+#[serde(transparent)]
+struct SerializedRequest(serde_json::Value);
```
Wodann, line 610: What is the benefit of wrapping?

Wodann (changes requested):
Thank you for the improvements. This is starting to look a lot cleaner 😃

crates/rethnet_eth/src/remote/client.rs
agostbiro: We need `Serialize` here now because we extract the result from the response before writing it to the cache and as a consequence `write_response_to_cache` takes `impl Serialize` as argument instead of `&str`.  

(`DeserializeOwned` is just sugar for `for<'a> serde::Deserialize<'a>`)

crates/rethnet_eth/src/remote/client.rs
agostbiro: It'd be possible with [`futures::stream`](https://docs.rs/futures/latest/futures/stream/index.html), but imo it's more difficult to read than the imperative version. Would make sense though if we wanted to parallelize the disk reads.

crates/rethnet_eth/src/remote/client.rs
agostbiro: Just to make the purpose of the generic value clear. It has no runtime cost, but it guarantees that we don't accidentally pass a response in place of a request or vice versa.

crates/rethnet_eth/src/remote/client.rs
agostbiro: Thanks, good call! Fixed in [9664ac8](https://github.com/NomicFoundation/hardhat/pull/4315/commits/9664ac82d6170abef9ced65f23d7722c24b87b73)

crates/rethnet_eth/src/remote/client.rs
Wodann: I was moreso referring to the move from the where clause to the generic type definition

crates/rethnet_eth/src/remote/client.rs
Wodann: I thought it would make some of this code cleaner with regard to the pre-allocations as those can be `filter_map`ped and `collect`ed respectively

crates/rethnet_eth/src/remote/client.rs
agostbiro: Ah just thought the where clause was an overkill for this. I typically use it with more complex generic bounds. Happy to put it back though

crates/rethnet_eth/src/remote/client.rs
agostbiro: Ok, want me to give it a try? (We need the `results` to be be as long as as the input `method_invocations` as we'll replace the `None` values with the results from the remote node based on the indices.

Wodann: If it's trivial to do, give it a try :)

crates/rethnet_eth/src/remote/client.rs
Wodann: Fair enough. Feel free to leave it as is then

Wodann (approved):
There is just one minor nitpick to convert to a functional style of programming for the iterator, but otherwise it LGTM

agostbiro:
@Wodann could you look at the [failures:](https://github.com/NomicFoundation/hardhat/actions/runs/5941490569/job/16115895376)


```
  21) Hardhat module
       Alchemy Forked provider
         hardhat_mine
           should reflect timestamps properly
             with two consecutive hardhat_mine invocations:

      Expected block 17972137 to have a timestamp increase of 1, but got 2 instead
      + expected - actual

      -2
      +1
      
      at assertTimestampIncrease (test/internal/hardhat-network/provider/modules/hardhat.ts:514:20)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Context.<anonymous> (test/internal/hardhat-network/provider/modules/hardhat.ts:562:13)

  22) Hardhat module
       Alchemy Forked provider
         hardhat_mine
           should reflect timestamps properly
             with two consecutive hardhat_mine invocations with different intervals:

      Expected block 17972137 to have a timestamp increase of 1, but got 2 instead
      + expected - actual

      -2
      +1
      
      at assertTimestampIncrease (test/internal/hardhat-network/provider/modules/hardhat.ts:514:20)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Context.<anonymous> (test/internal/hardhat-network/provider/modules/hardhat.ts:599:13)

  23) Hardhat module
       Alchemy Forked provider
         hardhat_reset
           can reset the forked provider to the latest block number:
     AssertionError: expected 17972156 to be close to 17972136 +/- 4
      at Context.<anonymous> (test/internal/hardhat-network/provider/modules/hardhat.ts:1466:20)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  24) HardhatNode
       should run calls in the right hardfork context
         when forking with a default hardfork activation history
           should throw when given post-EIP-1559 gas semantics and when running in the context of a pre-EIP-1559 block:
     AssertionError: Async error should have matched regex /Cannot run transaction: EIP 1559 is not activated./ but got "restoreForkBlockContext called when checkpointed"
      at expectErrorAsync (test/helpers/errors.ts:15:26)
      at Context.<anonymous> (test/internal/hardhat-network/provider/node.ts:937:31)
      at processImmediate (node:internal/timers:476:21)

  25) HardhatNode
       should run calls in the right hardfork context
         when forking with a hardfork activation history that indicates London happened one block early
           should throw when given post-EIP-1559 gas semantics and when running in the context of the block before EIP-1559 activation:
     AssertionError: Async error should have matched regex /Cannot run transaction: EIP 1559 is not activated./ but got "restoreForkBlockContext called when checkpointed"
      at expectErrorAsync (test/helpers/errors.ts:15:26)
      at Context.<anonymous> (test/internal/hardhat-network/provider/node.ts:983:31)
      at processImmediate (node:internal/timers:476:21)

  26) HardhatNode
       should run calls in the right hardfork context
         when forking with a hardfork activation history that indicates London happened one block early
           should accept pre-EIP-1559 gas semantics when running in the context of the block of the EIP-1559 activation:
     Error: Different snapshot state root
      at DualModeAdapter.makeSnapshot (src/internal/hardhat-network/provider/vm/dual.ts:310:13)
      at async HardhatNode._runInBlockContext (src/internal/hardhat-network/provider/node.ts:2101:22)
      at async HardhatNode.runCall (src/internal/hardhat-network/provider/node.ts:594:19)
      at async runCall (test/internal/hardhat-network/provider/node.ts:876:7)
      at async Context.<anonymous> (test/internal/hardhat-network/provider/node.ts:1002:11)

````

Note sure if these are known

File: crates/rethnet_eth/src/remote/client.rs [hunk 7]
```diff
@@ -376,12 +380,13 @@ impl RpcClient {
...
 
-        let mut results: Vec<Option<serde_json::Value>> =
-            Vec::with_capacity(method_invocations.len());
-        for cache_key in cache_keys.iter() {
-            let result = self.try_from_cache(cache_key.as_ref()).await?;
-            results.push(result)
-        }
+        let mut results: Vec<Option<serde_json::Value>> = stream::iter(&cache_keys)
+            .map(|cache_key| async { self.try_from_cache(cache_key.as_ref()).await })
+            .buffered(PARALLEL_DISK_READS)
```
Wodann, line 385: Should this be using something like https://doc.rust-lang.org/std/thread/fn.available_parallelism.html to be system-specific?

Wodann (changes requested):

Wodann:
> @Wodann could you look at the [failures:](https://github.com/NomicFoundation/hardhat/actions/runs/5941490569/job/16115895376)
> Note sure if these are known

I looked into this and am a little surprised that this is failing on Ubuntu but not Windows. One possible issue that I could see playing a role here is that the `test-core` and `test-provider` jobs are competing for limited query counts on alchemy.

A problem with our current client implementation is that we never retry on failures. My plan was to adopt a similar pattern with a `Provider` and `ProviderBuilder` to what Foundry [already has](https://github.com/foundry-rs/foundry/blob/master/crates/common/src/provider.rs) for robustness.

Maybe this resolves our issues? But I'm not sure.

Note: I tried re-running tests and [this one](https://github.com/NomicFoundation/hardhat/actions/runs/5941490569/job/16123562623?pr=4315#step:9:1599) failed with a panic. This might at least be avoided using the above method

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -226,67 +296,154 @@ impl RpcClient {
...
+                    .ok()
+                    .and_then(|v| v.cache_key())
+            })
+            .collect::<Vec<_>>();
 
-        self.send_request_body_with_cache(&request_body, cache_key)
+        let mut results: Vec<Option<serde_json::Value>> = stream::iter(&cache_keys)
+            .map(|cache_key| async { self.try_from_cache(cache_key.as_ref()).await })
+            .buffered(PARALLEL_DISK_READS)
+            .collect::<Vec<Result<_, RpcClientError>>>()
```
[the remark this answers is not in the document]
agostbiro: No unfortunately, with the futures, we have an iterator over `Future<Result>` and we need to collect it first to `Vec<Result>` and then need another collect to transpose this to `Result<Vec>`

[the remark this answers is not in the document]
agostbiro: Did simplify the code a bit though by removing the unnecessary async block: [92c107f](https://github.com/NomicFoundation/hardhat/pull/4315/commits/92c107ffe70c785f7839addc397764192c3d59b8)

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -131,6 +177,9 @@ impl RpcClient {
 
     async fn make_cache_path(&self, cache_key: &CacheKey) -> Result<PathBuf, RpcClientError> {
         let chain_id = self.chain_id().await?;
+
+        // TODO We should use a human readable name for the chain id as directory name once
```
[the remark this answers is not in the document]
agostbiro: Sure, done here: [9b81dc1](https://github.com/NomicFoundation/hardhat/pull/4315/commits/9b81dc185df96f0114aad509dedbd3ab98b693bc)

crates/rethnet_eth/src/remote/client.rs
agostbiro: This is about the number of futures we allow to be executed concurrently and as a result the max number of parallel disk reads we allow. I updated the code to use concurrent instead of parallel in the naming.

Wodann: I might be missing something, but that would still benefit from using a system-adapted parallelism or concurrency limit. If you're using more parallel readers than a system's multi-thread setup natively supports, you'll be unnecessarily saturating the runtime.

agostbiro: I looked into this some more and I think it should be ok:

The [available_parallelism](https://doc.rust-lang.org/std/thread/fn.available_parallelism.html) function tells us the number of logical cores on the machine. This is the number of threads that can perform computation in parallel in a program and our Tokio runtime will spawn [this many workers threads](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.worker_threads) that are responsible for progressing futures. So if we block in async code, we're blocking the worker threads and there is indeed a problem.

But since my refactoring we're using `tokio::fs` instead of `std::fs` for file system i/o in the RPC client. These both perform blocking i/o for portability, but `tokio::fs` will spawn the blocking operations on a special thread pool for blocking operations and return a future that can be awaited without blocking the worker threads. The [default value](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.max_blocking_threads) for the size of the blocking thread pool is 512, so spawning max 16 concurrent blocking threads here keeps us well within this limit. The default size can be so large, because the various operating systems are very efficient at handling threads that are blocking on some syscall for i/o.

My concern with picking the `PARALLEL_DISK_READS` (since renamed to `CONCURRENT_DISK_READS`) was not so much about saturating the blocking thread pool, but  about performance degradation of disk throughput when issuing too many parallel reads. Based on the article [linked](https://pkolaczk.github.io/disk-parallelism/) in the comment 16 parallel disk reads is a very conservative parameter.

Wodann (approved):

Issue #2513, opened 2022-03-21
Support state overrides in eth_call


WyseNynja:
Geth has support for "state overrides" that make eth_call even more powerful than it already is.

Geth's docs go into detail: https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set.

Ganache is likely to merge their PR before the next version: https://github.com/trufflesuite/ganache/pull/2565

A similar change in hardhat would bring it to feature parity.

Brownie already supports state overrides, and I am working on a PR that uses them so that multicall is available on any block on any chain without needing to deploy anything.

WyseNynja:
As-is, if you try to set the override argument in eth_call, the error is: `ValueError: {'code': -32602, 'message': 'Expected between 1 and 2 arguments and got 3'}`

fvictorio:
Still relevant, although we probably won't be able to work on this at least for some months.

drortirosh:
Any chance state overrides in eth_call is implemented in the near future?
It is already available in ganache, foundry/anvil, and of course on production nodes (geth, erigon, bor, etc)

PR #4316, opened 2023-08-22
Support state overrides in eth_call

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/2513)

Add support to state overrides, [see GETH docs for more info](https://geth.ethereum.org/docs/rpc/ns-eth#3-object---state-override-set)


packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
```diff
@@ -40,6 +41,14 @@ export const rpcStorageSlot = new t.Type<bigint>(
   t.identity
 );
 
+export const rpcStorageSlotHexStr = new t.Type<string>(
+  "Storage slot hex string",
+  isString,
```
fvictorio, line 46: We try to avoid using lodash because we'll probably remove it as a dependency in the future. In this particular case, `x => typeof x === "string"` is enough for our purposes. Lodash's implementations also accept String objects, but that's not relevant for us.

packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts
```diff
@@ -40,6 +41,14 @@ export const rpcStorageSlot = new t.Type<bigint>(
   t.identity
 );
 
+export const rpcStorageSlotHexStr = new t.Type<string>(
```
fvictorio, line 44: Personal preference: `rpcStorageSlotHexString` (if you run `rg 'Str\b' src/` you'll see that we don't use this abbreviation often).

fvictorio, line 44: Let's add a comment explaining why we need this separate codec

packages/hardhat-core/src/internal/core/jsonrpc/types/input/callRequest.ts
```diff
@@ -21,3 +28,29 @@ export const rpcCallRequest = t.type(
...
+    nonce: optionalOrNullable(rpcQuantity),
+    code: optionalOrNullable(rpcData),
+    state: optionalOrNullable(stateProperties),
+    stateDiff: optionalOrNullable(stateProperties),
+  },
+  "stateOverrideOptions"
+);
+
+export const stateOverrideSet = t.record(address, stateOverrideOptions);
+export const optionalStateOverrideSet = t.union([
```
fvictorio, line 47: You can use the `optional` helper, from the same module that `optionalOrNullable` is imported.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2354,17 +2366,96 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+      await this._overrideCode(address, stateOverrideOptions);
+      await this._overrideStateAndStateDiff(address, stateOverrideOptions);
+    }
+  }
+
+  private async _overrideBalanceAndNonce(
+    address: Address,
+    stateOverrideOptions: StateOverrideOptions
+  ) {
+    const MAX_NONCE = 18446744073709551615n; // The max allowed hex value is 8 bytes: 0xFFFFFFFFFFFFFFFF
```
fvictorio, line 2386:
```suggestion
    const MAX_NONCE = 0xffffffffffffffffn; // A nonce value has to fit in 8 bytes
```

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2354,17 +2366,96 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+    }
+
+    if (stateOverrideOptions.state !== undefined) {
+      // Clear all storage except specified properties which are override
+      await this._stateManager.clearContractStorage(address);
+    }
+
+    const tmpStateProperties = state ?? stateDiff;
+
+    for (const [storageAddr, value] of Object.entries(tmpStateProperties!)) {
```
fvictorio, line 2434: I understand why this non-null assertion makes sense, but I've seen so many bugs caused by it that I normally go out of my way to avoid them.

In this case, you can do something like this, which is way more verbose but it should work:

```js
    let newState;
    if (state !== undefined && stateDiff === undefined) {
      newState = state;
    } else if (state === undefined && stateDiff !== undefined) {
      newState = stateDiff;
    } else if (state === undefined && stateDiff === undefined) {
      // nothing to do
      return;
    } else {
      throw new InvalidInputError(
        `The properties 'state' and 'stateDiff' cannot be used simultaneously when configuring the state override set passed to the eth_call method.`
      );
    }
```

packages/hardhat-core/src/internal/hardhat-network/provider/modules/eth.ts
```diff
@@ -353,7 +363,11 @@ export class EthModule {
       trace,
       error,
       consoleLogMessages,
-    } = await this._node.runCall(callParams, blockNumberOrPending);
+    } = await this._node.runCall(
+      callParams,
+      blockNumberOrPending,
+      stateOverrideSet
```
fvictorio, line 369: I would do `stateOverrideSet ?? {}` here, use a mandatory state override object in the rest of the code, and remove the `optionalStateOverrideSet !== undefined` check in `_runTxAndRevertMutations`.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2354,17 +2366,96 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
+      await this._stateManager.putContractStorage(
+        address,
+        Buffer.from(storageAddr.replace("0x", ""), "hex"),
```
fvictorio, line 2437:
```suggestion
        toBuffer(storageAddr),
```

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
fvictorio, line 2434: I think `Addr` here is misleading. This is not an address, it's an storage key.

docs/src/content/hardhat-network/docs/reference/index.md
```diff
@@ -289,6 +289,34 @@ const trace = await hre.network.provider.send("debug_traceTransaction", [
 
 #### `eth_call`
 
+The method allows you to call a contract's function without making a transaction on the blockchain. This is particularly useful for querying contract data and performing read-only operations. See the [GETH documentation](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth) for more info.
```
ChristopherDedominici, line 292: How the doc looks like in the website:

image

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -624,7 +629,8 @@ Hardhat Network's forking functionality only works with blocks from at least spu
 
   public async runCall(
```
ChristopherDedominici, line 630: @fvictorio does it make sense to have here the optional parameter 'stateOverrideSet'?
Because this parameter is only used with the RPC method eth_call and I have the feeling that I'm 'polluting" the function definition with an additional parameter just for one specific scenario

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -624,7 +629,8 @@ Hardhat Network's forking functionality only works with blocks from at least spu
...
     call: CallParams,
-    blockNumberOrPending: bigint | "pending"
+    blockNumberOrPending: bigint | "pending",
+    stateOverrideSet: OptionalStateOverrideSet = {}
```
ChristopherDedominici, line 633: @fvictorio  does it make sense to have here the optional parameter 'stateOverrideSet'?
Because this parameter is only used with the RPC method eth_call and I have the feeling that I'm 'polluting" the function definition with an additional parameter just for one specific scenario

fvictorio (approved):

PR #4318, opened 2023-08-23
chore(hooks): simplify pre-commit hooks

Wodann:
This removes the run of `--no-default-features` from the commit hooks and changes the order to run lightweight commands first


agostbiro (approved):
Thanks, LGTM!

PR #4320, opened 2023-08-23
hardhat-viem initial implementation supporting client getters

### LINKED ISSUE #4273: hardhat-viem: initial implementation with a mechanism to get clients
Basic implementation just with support for creating public, wallet and test clients with sane defaults.

schaable:
Closes #4273


packages/hardhat-viem/src/internal/chains.ts
```diff
@@ -0,0 +1,94 @@
...
+
+const _memoizedGetChainId = memoize(async (provider: EthereumProvider) => {
+  return Number(await provider.send("eth_chainId"));
+});
+
+async function isHardhatNetwork(provider: EthereumProvider) {
+  return _memoizedIsHardhatNetwork(provider);
+}
+
+const _memoizedIsHardhatNetwork = memoize(
```
ChristopherDedominici, line 70: Just a personal preference:

What about declare a single method that handle all the scenarios? (especially if we expect more values in the future(?))

```
async function isNetworkWithProvider(
  provider: EthereumProvider,
  methodName: string
) {
  try {
    await provider.send(methodName);
    return true;
  } catch {
    return false;
  }
}
```

And then use it like this:

```
const _memoizedIsHardhatNetwork = memoize(
  async (provider: EthereumProvider) => {
    return isNetworkWithProvider(provider, "hardhat_metadata");
  }
```

"hardhat_metadata" and "anvil_nodeInfo" could be enums too

packages/hardhat-viem/src/internal/chains.ts
```diff
@@ -0,0 +1,94 @@
...
+      throw new UnknownDevelopmentNetworkError();
+    }
+  }
+
+  const matchingChains = Object.values(chains).filter(
+    ({ id }) => id === chainId
+  );
+
+  if (matchingChains.length === 0) {
+    throw new NetworkNotFoundError(chainId);
```
fvictorio, line 32: I feel like here we should repeat the `isHardhatNetwork`/`isFoundryNetwork` checks, to handle the case where someone is using a development network with a different chain id. This is not an uncommon practice.

packages/hardhat-viem/src/internal/accounts.ts
```diff
@@ -0,0 +1,13 @@
+import type { EthereumProvider } from "hardhat/types";
+import type { Address } from "viem";
+
+import memoize from "lodash.memoize";
+
+export async function getAccounts(provider: EthereumProvider) {
+  return _memoizedGetChainId(provider);
+}
+
+const _memoizedGetChainId = memoize(
```
fvictorio, line 10: Wrong function name.

packages/hardhat-viem/test/clients.ts
```diff
@@ -0,0 +1,165 @@
...
+describe("clients", () => {
+  describe("innerGetPublicClient", () => {
+    it("should return a public client", async () => {
+      const provider: EthereumProvider = new EthereumMockedProvider();
+
+      const client = await innerGetPublicClient(provider, chains.mainnet);
+
+      assert.isDefined(client);
+      assert.equal(client.type, "publicClient");
+      assert.equal(client.chain?.id, chains.mainnet.id);
```
fvictorio, line 22: Unnecessary?

```suggestion
      assert.equal(client.chain.id, chains.mainnet.id);
```

There are other places where this is happening too. You can add `/* eslint @typescript-eslint/no-unnecessary-condition: "error" */` at the beginning of the file to find them.

packages/hardhat-viem/test/chains.ts
```diff
@@ -0,0 +1,138 @@
...
+import { EthereumMockedProvider } from "./mocks/provider";
+
+describe("chains", () => {
+  describe("getChain", () => {
+    afterEach(sinon.restore);
+
+    it("should return the chain corresponding to the chain id", async () => {
+      const provider: EthereumProvider = new EthereumMockedProvider();
+      const sendStub = sinon.stub(provider, "send");
+      sendStub.withArgs("eth_chainId").returns(Promise.resolve("0x1")); // mainnet chain id
```
fvictorio, line 21: We should somehow assert that this is the only JSON-RPC call done, shouldn't we? That would help prevent performance regressions where we do unnecessary http requests.

I think we just need to add an `assert.equal(sendStub.callCount, 1);` statement.

I realize that that would make these tests even more whitebox, but I think it's worth it in this case since we are asserting certain guarantees (a minimal amount of requests) that we want to preserve.

fvictorio (approved):

PR #4323, opened 2023-08-24
feat: remappings support rework on core and hardhat-foundry plugin

antico5:
With this change, the solc input generation changes when remappings are involved.

We stop using settings.remappings, and instead use the importPaths without transformation as source names.

The task `TASK_COMPILE_TRANSFORM_IMPORT_NAME` was deprecated in favor of a new one, `TASK_COMPILE_GET_REMAPPINGS`, as now the resolver will make use of the remappings directly, if there are any. But the deprecated task was not removed, since we have to detect and handle 2 possible incompatible scenarios: an up-to-date hardhat core and an outdated hardhat-foundry plugin, and the opposite case.

For detecting an outdated hardhat-foundry plugin,  we check that the transform task doesn't make any changes to import path. For detecting an outdated hardhat core from an updated plugin, the new version of core sends an extra parameter to the deprecated transform task. The suggested approach of checking the presence of the transform task on `hre.tasks` using `extendEnvironment` did not work, because since the plugin itself was implementing it, it was present at the moment of checking it. I also tried other approaches so let me know in a comment if there are better alternatives.

The DependencyGraph was slightly changed to use source names instead of absolute paths as keys on the `visited` sources, because now different source names can point to the same file on disk.

I didn't iterate a lot over the changes, so feedback is welcome. Also testing on a few scenarios that are failing with previous versions work well with this change.

Also any recommendations on automated testing for this is welcome.


packages/hardhat-core/src/internal/solidity/resolver.ts
```diff
@@ -534,4 +559,14 @@ export class Resolver {
...
+
+  private _applyRemapping(sourceName: string): string {
+    for (const [from, to] of Object.entries(this._remappings)) {
+      if (sourceName.startsWith(from)) {
+        return sourceName.replace(from, to);
+      }
+    }
+
+    return sourceName;
+  }
```
fvictorio, line 571: This is incorrect. The full algorithm applies the remapping with the longest matching prefix and, if there is a tie, the first one wins.

We should extract this to a function or something and unit-test it.

fvictorio (approved):

PR #4324, opened 2023-08-24
bump: revm

Wodann:
Bumps our `revm` dependency to be in sync with the latest commit on revm's `main` branch


agostbiro (approved):
Thanks @Wodann ! Code LGTM, but I'm not sure about the CI failures. Could you have a look at those before merging?

PR #4325, opened 2023-08-24
fix: rpc result serde and avoid rate limits

agostbiro:
This PR started out as a quickfix for `Block<Transaction>` deserialization that was blocking @Wodann , but then some new issues popped up for which I also included fixes. Specifically, this PR contains:

- Make it possible for `Block<Transaction>` to be deserialized from `serde_json::Value`
- Make it possible for `TypedReceipt<Log>` to be deserialized from `serde_json::Value`
- Fix deserialization for `yParity` in `Block`. Geth recently [started to return](https://github.com/ethereum/go-ethereum/pull/27744) `yParity` alongside the `v` parameter for EIP-2930 and EIP-1559 transactions even though these should only contain the `yParity` parameter in theory. I discovered this when fetching blocks from Infura.
- Join some remote tests in the rpc client to reduce requests to avoid getting rated limited
- Limit concurrent requests for transaction receipts to avoid getting rate limited
- Use shared cache for integration tests and make them sequential to avoid getting rate limited

### DIFF
--- a/crates/rethnet_eth/src/remote/client.rs
+++ b/crates/rethnet_eth/src/remote/client.rs
@@ -1070,13 +1070,27 @@ mod tests {
         }
 
         #[tokio::test]
-        async fn get_earliest_block() {
+        async fn get_block_with_transaction_data() {
             let alchemy_url = get_alchemy_url();
+            let client = TestRpcClient::new(&alchemy_url);
 
-            let _block = TestRpcClient::new(&alchemy_url)
-                .get_block_by_number(BlockSpec::earliest())
+            let block_spec = BlockSpec::Number(U256::from(16220843));
+
+            assert_eq!(client.files_in_cache().len(), 0);
+
+            let block_from_remote = client
+                .get_block_by_number_with_transaction_data(block_spec.clone())
                 .await
-                .expect("should have succeeded");
+                .expect("should have from remote");
+
+            assert_eq!(client.files_in_cache().len(), 1);
+
+            let block_from_cache = client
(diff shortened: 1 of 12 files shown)


crates/rethnet_eth/src/remote/client.rs
```diff
@@ -1070,13 +1070,27 @@ mod tests {
         }
 
         #[tokio::test]
-        async fn get_earliest_block() {
+        async fn get_block_with_transaction_data() {
```
Wodann, line 1073: Should this be renamed to signal that you're testing caching?

crates/rethnet_eth/src/remote/eth.rs
```diff
@@ -52,14 +52,23 @@ pub struct Transaction {
     #[serde(with = "crate::serde::bytes")]
     pub input: Bytes,
     /// ECDSA recovery id
-    #[serde(alias = "yParity", with = "crate::serde::u64")]
+    #[serde(with = "crate::serde::u64")]
     pub v: u64,
+    /// Y-parity for EIP-2930 and EIP-1559 transactions. In theory these transactions types
```
Wodann, line 57: Did this cause a crash when deserializing?

Wodann (approved):

Wodann:
I merged as I only had one tiny nitpick and it seemed prudent to get this merged. It's still worth having a look at this question, @agostbiro : https://github.com/NomicFoundation/hardhat/pull/4325#discussion_r1304751501

PR #4328, opened 2023-08-25
feat: retry for rpc client

agostbiro:
Add retry logic for the rpc client to help without rate limiting. 

In addition to 429 responses, it also handles 5xx responses and request timeouts.

Testing strategy: without this change the local test suite fails on my machine with rate limit errors, with this change, it succeeds.

### DIFF
--- a/crates/rethnet_eth/src/remote/client.rs
+++ b/crates/rethnet_eth/src/remote/client.rs
@@ -831,7 +831,7 @@ mod tests {
         let error = TestRpcClient::new(&server.url())
             .call::<Option<eth::Transaction>>(MethodInvocation::GetTransactionByHash(hash))
             .await
-            .expect_err("should have failed to interpret response as a Transaction");
+            .expect_err("should have failed to due to a HTTP status error");
 
         if let RpcClientError::HttpStatus(error) = error {
             assert_eq!(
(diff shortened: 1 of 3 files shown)


crates/rethnet_eth/src/remote/client.rs
```diff
@@ -698,8 +812,8 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn send_request_body_500_status() {
-        const STATUS_CODE: u16 = 500;
+    async fn send_request_body_400_status() {
```
Wodann, line 815: Why 400?

crates/rethnet_eth/src/remote/client.rs
```diff
@@ -1088,7 +1202,7 @@ mod tests {
         }
 
         #[tokio::test]
-        async fn get_block_with_transaction_data() {
+        async fn get_block_with_transaction_data_cached() {
```
Wodann, line 1205: Was this just attempting to reduce the number of tests (and thus queries) by merging the tests for `get_block_with_transaction_data` and caching?

Wodann (changes requested):

crates/rethnet_eth/src/remote/client.rs
agostbiro: Yes, renaming is follow up from previous PR. There are a bunch of other block number with transaction data tests too

crates/rethnet_eth/src/remote/client.rs
agostbiro: If I understand correctly, this test was a smoke test that things don't blow up when the server returns an http error code. We're retrying 5xx requests now, so we can't test that here. I picked 400, because it's a common error that isn't retried.

Wodann (approved):

PR #4333, opened 2023-08-28
fix: disable in-memory cache for unsafe block number

agostbiro:
This PR disables in-memory caching for unsafe block numbers. 

I'm not 100% sure if this is the right approach, but figured it's easiest to discuss in the context of a PR.


Wodann (commented):
This seems reasonable to me. What did you want to discuss about the design tomorrow? We can also go over these two reviews on our call tomorrow

agostbiro:
> This seems reasonable to me. What did you want to discuss about the design tomorrow? We can also go over these two reviews on our call tomorrow

Ok, it works for me too. I just wasn't familiar with this part of the code, so wasn't sure about my approach. 👍

Wodann (approved):

PR #4336, opened 2023-08-29
Set default evmVersion to paris for solc gte 0.8.20

### LINKED ISSUE #4232: Set default evmVersion to paris
Edit by @fvictorio, original issue below.

We should:
- [ ] Use paris as the evmVersion targeted by solc unless the user explicitly set one. This should work for all possible solc configurations: the version shortcut, an explicit version, multiple versions, and overrides.
- [ ] Tests for all those scenarios.
- [ ] Update the docs to explain the situation, what Hardhat does, and how to change it.
- [ ] We should manually test this. It's easy to do: we just compile a contract with the default version, and check that no PUSH0 opcodes are included. Pasting the bytecode in the [EVM Codes Playground](https://www.evm.codes/playground) and changing the view to mnemonic is the easiest way to do this.
- [ ] Add a test which does the same that the previous step does. It's probably enough to have a fixture project where we just assert the resulting bytecode is a previously manually validated one. This will use a fixed solc version, but that's fine since this is mainly a smoke test.
- [ ] We should at least check that there are no surprises around the cache. The way to do this is to compile a project with the current version of Hardhat, then install the version with these changes, and compile again. This should trigger a re-compilation because the solc config changed.

---

### Describe the feature

Solc defaults the `evmVersion` parameter to shanghai since version 0.8.20. The main issue caused by this is that the compiler emits PUSH0 as an opcode. This opcode is supported on Ethereum L1 but as far as I know it hasn't been adopted on any other chain so far (except perhaps Gnosis Chain?). Most notably, it hasn't been adopted by L2s. On all these other chains, the PUSH0 opcode will be seen as an invalid opcode and cause the code to revert.

Hardhat relies on the solc default for `evmVersion`. Foundry, for reference, [defaults to paris](https://github.com/foundry-rs/foundry/blob/e05b9c75b4501d5880764948b61db787f3dd7fe0/config/src/lib.rs#L1732).

There is some knowledge about the potential issues caused by using shanghai, but I wouldn't count on it being widely known. At some point we will need to face EVM differences across chains and layers, and setting the default to paris simply delays this. However, setting the default to shanghai at this moment seems premature and is just asking for someone to deploy a broken contract to production.

I suggest to set the default evmVersion to paris if the compiler version is >=0.8.20 and the user hasn't configured it themselves.

### Search terms

evmVersion paris shanghai push0

schaable:
Closes #4232


docs/src/content/hardhat-runner/docs/config/index.md
```diff
@@ -208,3 +208,20 @@ You can configure how your tests are run using the `mocha` entry, which accepts
 ## Quickly integrating other tools from Hardhat's config
 
 Hardhat's config file will always run before any task, so you can use it to integrate with other tools, like importing `@babel/register`.
+
+## Default EVM Version
+
+The default EVM version is determined by solc's choice for a given compiler version. To specify a different EVM version, modify your `hardhat.config.js`:
+
+```js
+module.exports = {
```
ChristopherDedominici, line 217: Just an idea: I saw that in one of the tests you do a "more complex override":

```
{
        solidity: {
          compilers: [
            { version: "0.8.20", settings: { evmVersion: "istanbul" } },
          ],
          overrides: {
            "contracts/ERC20.sol": {
              version: "0.8.21",
              settings: { evmVersion: "shanghai" },
            },
          },
        },
      }
```

Could it be useful for the users to also have a similar example in the docs?

fvictorio: I believe that's already covered in the [multiple solidity versions](https://hardhat.org/hardhat-runner/docs/advanced/multiple-solidity-versions) guide, and adding it here might be a bit distracting. Using overrides is an advanced use case, and changing the evm version is also an advanced use case, so the combination of both things doesn't seem worth being explicitly mentioned.

fvictorio (approved):

PR #4338, opened 2023-08-29
fix: use insertion ordering for MemPool

Wodann:


agostbiro:
Code looks good, but I'm wondering about the context here? By default we probably want to order transactions in the mempool based on fees.

Wodann:
> Code looks good, but I'm wondering about the context here? By default we probably want to order transactions in the mempool based on fees.

This is merely intended to make tests succeed in Hardhat tests.

PR #4340, opened 2023-08-29
Fix/header defaults

Wodann:


agostbiro:
For some reason the CI didn't run at all on this PR

PR #4341, opened 2023-08-29
feat(blockchain): add reservation of blocks

Wodann:


agostbiro (approved):
Thanks, looks good overall. Two questions:

- Can `reserve_blocks` have a default implementation in the trait? It looks like the two implementations are the same.
- Is this tested from the hardhat side? Would be good to have tests for edge cases.

Feel free to merge if these are not concerns.

Wodann:
> - Can `reserve_blocks` have a default implementation in the trait? It looks like the two implementations are the same.

They are mostly the same, yes. I can see about moving shared code to a file.

> - Is this tested from the hardhat side? Would be good to have tests for edge cases.

Yes, at the moment most of our codebase relies on tests in Hardhat. Once we have released, we'll start porting over more tests so that we can test standalone.

Wodann:
I tried splitting the shared code into a free function but ran into lifetime violations, so left the code as is for now. In the future, if we have a `trait BlockchainStorage`, we might be able to resolve it through a default implementation on the `trait Blockchain`

Wodann:
CI is failing due to https://github.com/NomicFoundation/hardhat/actions/runs/6026825025/job/16350667120#step:9:3168

PR #4342, opened 2023-08-30
Update tasks-and-scripts.md with importing the task  function.

HMFazleRabbi:

---


fvictorio:
This is not necessary in javascript projects, that's why we don't include these required in our examples. You can import them if you prefer being explicit (as explained elsewhere; I don't remember where now), but to keep the snippets concise we don't do it.

PR #4345, opened 2023-08-30
fix: avoid disk cache write race conditions

agostbiro:
This PR introduces a change to avoid disk cache write race conditions by writing to a temporary file first and then moving it to the cache file location.

Resolves https://github.com/NomicFoundation/rethnet/issues/164


crates/rethnet_eth/src/remote/client.rs
```diff
@@ -366,16 +371,53 @@ impl RpcClient {
...
         result: impl Serialize,
     ) -> Result<(), RpcClientError> {
-        let cache_path = self.make_cache_path(cache_key).await?;
         let contents = serde_json::to_string(&result).expect(
             "result serializes successfully as it was just deserialized from a JSON string",
         );
-        match tokio::fs::write(cache_path, contents).await {
+
+        // 1. Write to a random temporary file first to avoid race conditions.
+        let tempfile = match tokio::task::spawn_blocking(NamedTempFile::new).await? {
```
Wodann, line 379: Is the overhead of spawning a task worth it for the creation of the file? This feels like premature optimisation.

Wodann (changes requested):

crates/rethnet_eth/src/remote/client.rs
agostbiro: It's required unfortunately, otherwise it'd be blocking the worker threads as `NamedTempFile::new` does blocking i/o. 

(The `tokio::fs` module does the same under the hood to wrap blocking `std::fs` operations.)

Wodann (approved):

PR #4347, opened 2023-08-30
refactor: make blockchain traits async

Wodann:


agostbiro (approved):
LGTM thanks! I checked and the rethnet mode failures are due to this: https://github.com/NomicFoundation/rethnet/issues/175

PR #4349, opened 2023-08-31
fix: roll back cached remote state block number check

agostbiro:
Attempt at fixing https://github.com/NomicFoundation/hardhat/actions/runs/6030065963/job/16361112955


agostbiro:
Closing in favor of https://github.com/NomicFoundation/hardhat/pull/4356

PR #4357, opened 2023-09-01
docs: add rust backtrace instructions

agostbiro:
Add instructions to generate a Rust stack trace from Hardhat tests.


Wodann:
I'm not a fan of adding "dead dependencies" to our `Cargo.toml`. It increases the number of crates that need to be downloaded and built. It feels like this is just one of the possible ways to debug, so it's an option for people to use when they are debugging.

Wodann (commented):
I'm not a fan of adding "dead dependencies" to our `Cargo.toml`. It increases the number of crates that need to be downloaded and built. It feels like this is just one of the possible ways to debug, so it's an option for people to use when they are debugging.

agostbiro:
I understand the concern about dead dependencies, but we already have `backtrace` as a transitive dependency (see Cargo.lock), so it doesn't increase download size/build time. 

It is however a nice quality of life improvement (at least for me), because if the `backtrace` crate is not in the dependencies, my IDE will try to import `std::backtrace::Backtrace` which doesn't have the required functionality. This is a bit of a gotcha that has already caused me to lose some time.

Wodann:
Could you please add a comment to clarify this, as an automated tool for pruning dependencies will flag this.

agostbiro:
> Could you please add a comment to clarify this, as an automated tool for pruning dependencies will flag this.

Thanks! I learned in the meantime that since Rust 1.65 the standard library has this functionality, so I removed the `backtrace` crate and updated the instructions in [33ed5bb.](https://github.com/NomicFoundation/hardhat/pull/4357/commits/33ed5bb6014beb61602191d8483bfa6e383d007c)

Wodann (approved):

PR #4360, opened 2023-09-04
fix: cache block numbers for block time duration

agostbiro:
Fixes failure 2 in https://github.com/NomicFoundation/rethnet/issues/175 by caching block numbers for validation purposes for the block time interval.

This change brings down the running time of the tests from ~2 minutes to ~75 seconds (before the CI broke, these tests took around ~45 seconds). This is still over the initial 50 second timeout. If I disable the block number checks for in memory caches, the running time goes back to the initial ~45 seconds. Some slow down is to be expected here since the previous version of the code cached all block numbers, so I increased the timeout to 80 seconds. As an alternative we could pin the tests to a safe block number.

Do note however that these [tests run in <5 seconds](https://github.com/NomicFoundation/hardhat/actions/runs/6013413577/job/16310837774#step:6:2880) in the main branch, so more work remains here.

### DIFF
--- a/crates/rethnet_eth/src/block/reorg.rs
+++ b/crates/rethnet_eth/src/block/reorg.rs
@@ -1,4 +1,5 @@
 use crate::U256;
+use std::time::Duration;
 
 /// Test whether a block number is safe from a reorg for a specific chain based on the latest block
 /// number.
@@ -57,7 +58,7 @@ pub fn largest_possible_reorg(chain_id: &U256) -> U256 {
         1 | 4 | 5 | 42 => 32,
         // Ropsten
         3 => 100,
-        // xDai
+        // Gnosis/xDai
         100 => 38,
         _ => {
             log::warn!(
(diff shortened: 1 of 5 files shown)


crates/rethnet_eth/src/block/reorg.rs
```diff
@@ -69,3 +70,23 @@ pub fn largest_possible_reorg(chain_id: &U256) -> U256 {
...
+        // Ethereum mainnet, Ropsten, Rinkeby, Goerli and Kovan testnets
+        // 32 blocks is one epoch on Ethereum mainnet
+        1 | 3 | 4 | 5 | 42 => Duration::from_secs(12),
+        // Gnosis/xDai
+        // https://gnosisscan.io/chart/blocktime
+        100 => Duration::from_secs(5),
+        _ => {
+            log::warn!(
+                "Unknown chain id {chain_id}, using default block time of {} seconds",
+                rethnet_defaults::DEFAULT_SAFE_BLOCK_TIME.as_secs(),
```
Wodann, line 87: In my opinion, the hardcoded Rethnet defaults should only be used in the Rethnet binaries and not be forced upon users.

The only constants we should be using in rethnet_eth are Eth-specific constants. This also avoids circular dependencies, as rethnet_defaults does or would likely depend on rethnet_eth for types.

We already made one temporary addition in a previous PR of yours that will be replaced when Rethnet supports proper config of chains, but we shouldn't keep breaking this pattern

Wodann (changes requested):
Instead of still querying the block number before performing the actual request, which adds a serial delay, why not similtaneously run them in a batch? That avoids delays

agostbiro:
> Instead of still querying the block number before performing the actual request, which adds a serial delay, why not similtaneously run them in a batch? That avoids delays

There are two use cases for `cached_block_number()`:

1. Resolve write key for RPC cache. In this case `cached_block_number()` is called after the request. The block number is not necessary to resolve all write keys, and I think the best solution here would be to not to wait for writing to the RPC cache before returning the RPC result, but write to the RPC cache in a background thread. I tried to this earlier, but ran into life time problems and it'd take a larger refactoring to fix those.
2. Check whether a block number is cacheable for the in-memory cache. In this case `cached_block_number()` is called after the request as well, and it's not possible to know from the RPC client whether it'll be called.

crates/rethnet_eth/src/block/reorg.rs
agostbiro: I agree in principle, but I see no other way to do it in this PR. Can we fix this at the same time as the other default (safety depth)?

Wodann: Sure

Wodann:
> I think the best solution here would be to not to wait for writing to the RPC cache before returning the RPC result, but write to the RPC cache in a background thread

@agostbiro, could you please add an issue for this to the backlog?

Wodann (approved):
LGTM.

For the future, it would be good to benchmark this using flame charts. I'll add a section to the book about how to use it. That hopefully gives us an indication of where most of our time is being spent

PR #4369, opened 2023-09-11
fix: MemPool update logic

Wodann:


crates/rethnet_evm/src/mempool.rs
```diff
@@ -224,7 +233,27 @@ impl MemPool {
...
             *old_transactions = transaction;
         } else {
+            let caller = *transaction.caller();
+            let mut next_pending_nonce = transaction.nonce() + 1;
+
             pending_transactions.push(transaction);
+
+            // Move as many future transactions as possible to the pending status
+            if let Some(future_transactions) = self.future_transactions.get_mut(&caller) {
+                while let Some((idx, _)) = future_transactions
```
agostbiro, line 243: It looks like this might be quadratic in the length of `future_transactions`. I'm wondering if we could maybe partition in one go?

crates/rethnet_evm/src/mempool.rs
```diff
@@ -148,21 +148,30 @@ impl MemPool {
...
+
             if let Some((idx, _)) = transactions
                 .iter()
                 .enumerate()
                 .find(|(_, transaction)| !is_valid_tx(transaction, &self.block_gas_limit, &sender))
             {
-                // Question: Do we ever need to consider tx.nonce < sender.nonce, due to manual modifications?
-
+                // Move all consequent transactions to the future queue
                 let mut invalidated_transactions = transactions.split_off(idx);
```
agostbiro, line 175: Tbh I don't understand why we can split off at that index. Why are all the transactions after that are guaranteed to be invalid?

agostbiro (changes requested):

crates/rethnet_evm/src/mempool.rs
Wodann: Transactions are sorted by insertion order and transaction can only be inserted when it's an increment of one nonce. Thus, if we remove one transaction of an author, we can also removed all consequent ones.

crates/rethnet_evm/src/mempool.rs
agostbiro: Sure, shouldn't be a problem as long as the number of transactions in the mempool is <100 which is I think a safe assumption for now.

agostbiro (approved):

crates/rethnet_evm/src/mempool.rs
Wodann: https://github.com/NomicFoundation/rethnet/issues/181

PR #4370, opened 2023-09-11
fix: WETH contract test

Wodann:
`ForkState` does not currently calculate an accurate account storage root as we have no easy way of combining historic account storage with the local one.

As this is line with Hardhat, we'll for now opt to just return a KECCAK hash of the null RLP-encoding.

If people request the feature for Hardhat (or EDR), we might add proper support in the future.


agostbiro (approved):
Thanks, LGTM

PR #4373, opened 2023-09-11
fix: use the same fork block number between the blockchain and VM

Wodann:
This resolves a bug where the Hardhat fork state could be created with a block number that forked 1 block later than the Hardhat blockchain. Since we were reading the fork block number from the state in EDR, this was creating a discrepancy.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
@@ -174,6 +174,8 @@ export class HardhatNode extends EventEmitter {
         forkNetworkId
       );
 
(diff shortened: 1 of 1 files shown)


agostbiro:
Hmm I can still repro the timestamp bug locally in this branch:

image

Wodann:
There must be a second part to my fix in `rethnet/feat/dual`, as these tests are succeeding there. Do you mind if I close this PR and wait until `rethnet/feat/dual` is ready to be merged with `rethnet/main` until we fix this, @agostbiro?

agostbiro:
Sure makes sense @Wodann !

PR #4376, opened 2023-09-11
ci: add test provider with EDR RPC cache job

agostbiro:
Add a CI job for running `hardhat-core` tests in `rethnet` mode that retains the Rethnet RPC cache.

Resolves https://github.com/NomicFoundation/rethnet/issues/87

https://github.com/NomicFoundation/hardhat/pull/4375 should be merged first.

### DIFF
--- a/.github/workflows/hardhat-core-ci.yml
+++ b/.github/workflows/hardhat-core-ci.yml
@@ -201,7 +201,7 @@ jobs:
       - name: Install package
         run: yarn --frozen-lockfile
 
-      - name: Cache stack trace artifacts
+      - name: Cache EDR RPC cache
         uses: actions/cache@v2
         with:
           path: |
--- a/.github/workflows/hardhat-core-ci.yml
+++ b/.github/workflows/hardhat-core-ci.yml
@@ -119,7 +119,7 @@ jobs:
         run: yarn test:except-provider
 
   test-provider:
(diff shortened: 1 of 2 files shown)


.github/workflows/hardhat-core-ci.yml
```diff
@@ -169,6 +169,57 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    needs: check-rethnet
+    strategy:
+      fail-fast: false
+      matrix:
+        node: [18.15]
+        # disable until actions/virtual-environments#4896 is fixed
+        # os: ["macos-latest", "ubuntu-latest", "windows-latest"]
+        os: ["ubuntu-latest", "windows-latest"]
+        vm: ["rethnet"]
```
Wodann, line 183: I'm wondering now whether we could avoid the code duplication by adding a `cache: [true, false]` entry in the matrix, which will consequently be used to determine whether to run the cache step.

Would you be able to try that out?

Wodann (changes requested):

.github/workflows/hardhat-core-ci.yml
agostbiro: Good it idea, done in [3362561](https://github.com/NomicFoundation/hardhat/pull/4376/commits/33625615c469012c666081c2893f8369dc4c1e21)

PR #4379, opened 2023-09-12
refactor: lazily fetch remote receipts

Wodann:
I accidentally incorporated part of my next change as well: the addition of `StateError` to `trait Blockchain`. It doesn't have any side-effects, nor introduces compiler issues, so shouldn't pose a problem, but is tricky to remove from the commit. If you're okay with it, @agostbiro, I'd like to keep it in the PR.


crates/rethnet_evm/src/block/remote.rs
```diff
@@ -0,0 +1,118 @@
...
+    block::Header,
+    remote::eth::{self, TransactionConversionError},
+    transaction::SignedTransaction,
+    withdrawal::Withdrawal,
+    Address, B256, B64,
+};
+
+use crate::{blockchain::BlockchainError, Block, SyncBlock};
+
+/// A remote block, which lazily loads additional information when requested.
```
agostbiro, line 13: I found this a bit confusing. As far as I can tell the lazy loading is happening in `RemoteBlockchain::receipt_by_transaction_hash`?

agostbiro (approved):
Thanks, left an informational comment. Otherwise LGTM

crates/rethnet_evm/src/block/remote.rs
Wodann: Resolved by https://github.com/NomicFoundation/hardhat/pull/4379/commits/3ccbd57e6a4a6c166e588dad85ef7a59fb8bbbb1

PR #4380, opened 2023-09-12
ci: use cargo hack instead of cargo check

agostbiro:
Use `cargo hack` instead of `cargo check` to let us verify that all feature combinations compile. Resolves https://github.com/NomicFoundation/rethnet/issues/168

Note that this in itself doesn't guarantee that our code will work with `no_std` if the `std` feature is not specified. If we want to check for that too, we'll have to integrate [`cargo nono`](https://github.com/hobofan/cargo-nono) tool.

I've tested the updated MacOS automated setup, but not the Linux/devcontainer flow.


.cargo-husky/hooks/pre-commit
```diff
@@ -14,7 +14,7 @@ fi
 set -xe
 
 # Setting RUSTFLAGS env for clippy makes it not include custom rules
-RUSTFLAGS=-Dwarnings cargo check --workspace --all-targets --profile bench ${ALL_FEATURES}
+RUSTFLAGS=-Dwarnings cargo hack check --feature-powerset --no-dev-deps
```
Wodann, line 17: This doesn't need to be run in the pre-commit hook. The idea is to make CI more expansive and correct, but keep local checks minimal

Brewfile
```diff
@@ -1,5 +1,9 @@
+tap "taiki-e/tap"
```
Wodann, line 1: This might also be unnecessary in that case

book/src/01_getting_started/01_setup.md
```diff
@@ -25,5 +25,6 @@ If you're using the [Homebrew](https://brew.sh/) package manager, run `brew bund
 If you would like to set up the environment manually, you will need to install the following dependencies:
 
 - Install Rust using [rustup](https://www.rust-lang.org/tools/install)
+- Install `cargo hack`: `cargo install cargo-hack`
```
Wodann, line 28: Might be unnecessary

scripts/setup.sh
```diff
@@ -6,6 +6,9 @@ rust_version=$(<rust-toolchain)
 # rustup
 curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain $rust_version
 
+# cargo hack CLI
```
Wodann, line 9: Probably also unnecessary

Wodann (changes requested):

.cargo-husky/hooks/pre-commit
agostbiro: Hmm are you sure? I find it really beneficial to run the same checks that the CI does (at least for Rust) before opening a PR. I find that it saves me time in the long run due to fewer adjustments and context switching later on. It can also save time for reviewers.

Performance-wise this change leads to 15% slower pre commit hook execution time for a clean build on my machine.

agostbiro: As discussed with @Wodann we'll support running full CI checks in the release CLI (https://github.com/NomicFoundation/rethnet/issues/63)

Reverted to `cargo check` in pre-commit in [0cb0dc8](https://github.com/NomicFoundation/hardhat/pull/4380/commits/0cb0dc81d17d19c1f20b00fc0918d2451b727e20)

Brewfile
agostbiro: Removed in [0cb0dc8](https://github.com/NomicFoundation/hardhat/pull/4380/commits/0cb0dc81d17d19c1f20b00fc0918d2451b727e20)

book/src/01_getting_started/01_setup.md
agostbiro: Removed in [0cb0dc8](https://github.com/NomicFoundation/hardhat/pull/4380/commits/0cb0dc81d17d19c1f20b00fc0918d2451b727e20)

scripts/setup.sh
agostbiro: Removed in [0cb0dc8](https://github.com/NomicFoundation/hardhat/pull/4380/commits/0cb0dc81d17d19c1f20b00fc0918d2451b727e20)

Wodann (approved):

Issue #2594, opened 2022-04-11
Add `hh init` command


fvictorio:
A new hardhat project can be initialized by running `npx hardhat`. But running the same command inside a Hardhat project does something different (it prints the help). This results in a lot of users being confused when they try to initialize a project without realizing that some parent directory has a Hardhat config, because then they get the help info instead of the project initialization prompt. This happens a lot.

Instead of that behavior, we should have an explicit `npx hardhat init` command (it can be called some other way, but `init` is the obvious choice). If you **are not** inside a Hardhat project, then the usual project initialization stuff is shown. If you **are** inside a project, an error message would be shown, indicating the location of the detected Hardhat config.

`npx hardhat` will continue to have the same behavior when you are inside a project. If you **are not** inside a project, then a proper error message should be shown, explaining this and suggesting the user to run `npx hardhat init` if their intention was to initialize a project. This is very important because there is a lot of documentation and educational material out there that has the plain `npx hardhat` step.

Of course, as part of this we should also update all the relevant parts of our documentation, and let people know about this change so that they can update their docs or articles.

- [x] Merge a PR to add `init` command
- [ ] Merge a PR to update the docs. This PR should be merged after that the code is released (Docs PRs are instantly released)
- [ ] Reach out to the main tutorials/docs that are using hardhat to inform about the deprecation

AbhinavMir:
I highly suggest this! 

Today, I had a root folder which (somehow) had a dummy `hardhat.config.js` file lying around. Because of this, in the subfolders, the help was being printed. It is probably better dev-experience to support `npx hardhat init` - happy to be assigned this if needed.

fvictorio:
Not stale.

PR #4384, opened 2023-09-15
Add hh init command

ChristopherDedominici:
Add new CLI command `init`. It will be used to initialize new Hardhat projects.
Example:
```
npx hardhat init
```

[Link to issue](https://github.com/NomicFoundation/hardhat/issues/2594)

[Link to docs PR](https://github.com/NomicFoundation/hardhat/pull/4394)


packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -151,25 +154,47 @@ async function main() {
...
-          undefined
+        parsedTaskName === undefined &&
+        hardhatArguments.config === undefined &&
+        !isCwdInsideProject()
       ) {
-        await createProject();
+        await createNewProject();
+
+        // Warning for Hardhat V3 deprecation
+        console.warn(
```
ChristopherDedominici, line 176: Warning message preview:

image

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -151,25 +154,45 @@ async function main() {
...
 
-      // Many terminal emulators in windows fail to run the createProject()
-      // workflow, and don't present themselves as TTYs. If we are in this
-      // situation we throw a special error instructing the user to use WSL or
-      // powershell to initialize the project.
-      if (process.platform === "win32") {
-        throw new HardhatError(ERRORS.GENERAL.NOT_INSIDE_PROJECT_ON_WINDOWS);
-      }
+    // Tasks are only allowed inside a Hardhat project (except the init task)
+    if (!isCwdInsideProject()) {
```
fvictorio, line 194: I think we should do this here:

```suggestion
    if (hardhatArguments.config === undefined && !isCwdInsideProject()) {
```

Otherwise you can't do something like:

1. Install Hardhat (but don't create a project)
2. Run `npx hardhat node --config /path/to/some/config/in/another/directory`

This works in the current version of Hardhat, but it doesn't in this PR. It wouldn't surprise me if people are relying on this behavior for some use cases (like starting a node anywhere without having to cd into a project).

ChristopherDedominici: done

.changeset/friendly-moles-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Added new CLI command `init`. Purpose: initialize a new Hardhat project. Usage: `npx hardhat init`.
```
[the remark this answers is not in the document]
ChristopherDedominici: done

schaable (approved):

fvictorio (approved):

PR #4389, opened 2023-09-16
Adding the ability to pass arguments to scripts called in RUN tasks.

ianmaddox:
Uses standard end-of-options double-dash syntax:
`npx hardhat run myScript.js --network fork -- --myparam=NootNoot`

- [x] I didn't do anything of this; My Discord question went unanswered. This PR is my attempt at discussing the issue with the HH team. I'm happy to discuss and make changes as needed.

### DIFF
--- a/packages/hardhat-core/src/internal/cli/cli.ts
+++ b/packages/hardhat-core/src/internal/cli/cli.ts
@@ -7,7 +7,7 @@ import {
   TASK_HELP,
   TASK_TEST,
 } from "../../builtin-tasks/task-names";
-import { HardhatConfig, TaskArguments } from "../../types";
+import { HardhatConfig, TaskArguments, ScriptArgs } from "../../types";
 import { HARDHAT_NAME } from "../constants";
 import { HardhatContext } from "../context";
 import {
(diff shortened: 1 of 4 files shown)


ncjones:
This is exactly what I am looking for. My workaround is to define a "script" task that does the same thing:

```js
const { task } = require('hardhat/config');
const { runScriptWithHardhat } = require('hardhat/internal/util/scripts-runner');
task('script')
  .addVariadicPositionalParam('args')
  .setAction(async (taskArgs, hre) => {
    await runScriptWithHardhat(hre.hardhatArguments, taskArgs.args[0], taskArgs.args.slice(1));
  });
```

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -277,7 +278,7 @@ async function main() {
     try {
       const timestampBeforeRun = new Date().getTime();
 
-      await env.run(taskName, taskArguments);
+      await env.run(taskName, taskArguments, scriptArgs);
```
fvictorio, line 281: I don't understand this. Isn't `scriptArgs` always just `undefined` here?

alcuadrado:
Hey, sorry this never got merged. We are currently working on a v3 of Hardhat which will include some similar functionally.

PR #4398, opened 2023-09-18
feat: debug trace transaction

agostbiro:
WIP to add `debug_traceTransaction` support for EDR.

I verified that the simple tracer API and the EIP-3155 API are mutually exclusive, so we don’t have support them both for the same transaction execution and we don’t have to change the `with_trace` logic in the run , `dry_run` and `guaranteed_dry_run` methods.

We initially thought the quickest way to add `debug_traceTransaction` was to keep most of the `debug_traceTransaction` implementation in JS for now and add a `rethnet` variant for `VMDebugTracer` of Hardhat Network. This has proven difficult, because `VMDebugTracer` is tightly coupled with the `ethereum-js` vm.

After studying the code and experimenting a bit, I decided instead to expose a `debug_trace_transaction` function from `rethnet_evm_napi` that delegates to a corresponding function in `rethnet_evm`. This function makes a clone of the state, and executes and commits all the prior transactions in the block to the cloned state and then executes the transaction with the matching hash with `TracerEip3155` and returns the debug trace output. The transactions from the block are taken as `Vec<rethnet_evm_napi::TransactionRequest>` as a conversion method for this already exists on the JS side.  For the other `napi` arguments I follow the pattern in `runtime.rs`. For executing the transactions in `rethnet_evm::debug_trace_transaction` I follow the pattern in `BlockBuilder::add_transaction`.

I decided to adapt the implementation of `revm::TracerEip3155` as it has an awkward interface to get the logs out and it doesn't follow the EIP-3155 spec. The biggest todo left is adding the fields that are obligatory in [EIP-3155](https://eips.ethereum.org/EIPS/eip-3155), but missing from `revm::TracerEip3155` and adding the optional fields that Hardhat currently supports.


crates/rethnet_evm/src/debug_trace.rs
```diff
@@ -0,0 +1,314 @@
...
+        });
+    }
+
+    if cfg.spec_id > SpecId::MERGE && block_env.prevrandao.is_none() {
+        return Err(TransactionError::MissingPrevrandao.into());
+    }
+
+    // Clone the state so that we can apply throw-away modifications on it
+    // TODO figure out why this doesn't work
+    // let mut state = state.clone();
```
Wodann, line 45: You need to pass in the `Box` itself, to be able to clone, which means the clone needs to happen external to this function

crates/rethnet_evm/src/debug_trace.rs
```diff
@@ -0,0 +1,314 @@
...
+) -> Result<DebugTraceResult, DebugTraceError<BlockchainErrorT, StateErrorT>>
+where
+    BlockchainErrorT: Debug + Send + 'static,
+    StateErrorT: Debug + Send + 'static,
+{
+    if cfg.spec_id < SpecId::SPURIOUS_DRAGON {
+        // Matching Hardhat Network behaviour: https://github.com/NomicFoundation/hardhat/blob/af7e4ce6a18601ec9cd6d4aa335fa7e24450e638/packages/hardhat-core/src/internal/hardhat-network/provider/vm/ethereumjs.ts#L427
+        return Err(DebugTraceError::InvalidSpecId {
+            spec_id: cfg.spec_id,
+            message: "`debug_traceTransaction` is not supported prior to Spurious Dragon"
```
Wodann, line 34: Can't this error message be part of the actual `thiserror`?

agostbiro: Done in [f072615](https://github.com/NomicFoundation/hardhat/pull/4398/commits/f0726152c200a20ba6ddf1697179d52ef16c26ed)

crates/rethnet_evm/src/debug_trace.rs
agostbiro: Will need the changes from https://github.com/NomicFoundation/hardhat/pull/4254

crates/rethnet_evm_napi/src/transaction/signed/eip1559.rs
```diff
@@ -30,21 +25,9 @@ pub struct EIP1559SignedTransaction {
...
-            env.create_buffer_with_borrowed_data(
-                input.as_ptr(),
-                input.len(),
-                input,
-                |input: Bytes, _env| {
-                    mem::drop(input);
-                },
-            )
-        }
-        .map(JsBufferValue::into_raw)?;
```
Wodann, line 47: This code was intentional. Using a borrow, it avoids a copy of the memory, since `Bytes` is able to share memory

agostbiro: I had to switch the `JsBuffer` to `Buffer` to make the `rethnet_evm_napi::SignedTransaction` type `Sync` so that the async `rethnet_evm_napi::debug_trace_transaction` function can take `rethnet_evm_napi::SignedTransaction` as argument. 

Alternative would be to have two types: `rethnet_evm_napi::SignedTransactionToJs` with `JsBuffer` fields and `rethnet_evm_napi::SignedTransactionFromJs` with `Buffer` fields. Wdyt?

Wodann: Is the problem resolved by using a `PendingTransaction`?

packages/hardhat-core/test/internal/hardhat-network/provider/utils/assertEqualTraces.ts
```diff
@@ -14,18 +14,51 @@ export function assertEqualTraces(
...
 
-  for (const [i, log] of expected.structLogs.entries()) {
+  for (const [i, expectedLog] of expected.structLogs.entries()) {
+    const actualLog = actual.structLogs[i];
     // we ignore the gasCost of the last step because
     // we don't guarantee that it's correct
     if (i === expected.structLogs.length - 1) {
-      actual.structLogs[i].gasCost = 0;
-      log.gasCost = 0;
+      actualLog.gasCost = 0;
```
Wodann, line 22: I don't remember this code, but I see that we were previously also modifying the expected result.

I'm not sure this is a good idea, as we're introducing a side-effect to a function that's merely supposed to be asserting equality.

Instead, I believe that we should either be modifying Hardhat to comply with the EIP or accepting a wider range of values for properties that we know EDR doesn't support yet (like `STATICCALL` gas costs), similar to what Franco pointed out in our Slack discussion

Wodann (changes requested):

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
```diff
@@ -190,6 +190,7 @@ export class DualModeAdapter implements VMAdapter {
     block: Block,
     config: RpcDebugTracingConfig
   ): Promise<RpcDebugTraceOutput> {
+    // TODO add rethnet adapter
```
agostbiro, line 193: @Wodann I'm assuming we don't want dual mode tests due to the expected differences in output?

crates/rethnet_evm_napi/src/transaction/signed/eip1559.rs
agostbiro: Yes, resolved by using `PendingTransaction`.

The original code looked like this:

```rust
#[napi]
pub async fn foo(buffer: napi::JsBuffer) {}
```

And it failed with this error:

```
error: future cannot be sent between threads safely
  --> crates/rethnet_evm_napi/src/debug_trace.rs:14:1
   |
14 | #[napi]
   | ^^^^^^^ future created by async block is not `Send`
   |
   = help: within `[async block@crates/rethnet_evm_napi/src/debug_trace.rs:14:1: 14:8]`, the trait `Send` is not implemented for `*mut napi_env__`
note: captured value is not `Send`
  --> crates/rethnet_evm_napi/src/debug_trace.rs:14:1
   |
14 | #[napi]
   | ^^^^^^^ has type `JsBuffer` which is not `Send`
note: required by a bound in `execute_tokio_future`
  --> /Users/abiro/.cargo/registry/src/index.crates.io-6f17d22bba15001f/napi-2.13.3/src/tokio_runtime.rs:89:18
   |
89 |   Fut: 'static + Send + Future<Output = Result<Data>>,
   |                  ^^^^ required by this bound in `execute_tokio_future`
   = note: this error originates in the attribute macro `napi` (in Nightly builds, run with -Z macro-backtrace for more info)
```

crates/rethnet_evm/src/debug_trace.rs
```diff
@@ -0,0 +1,421 @@
...
+        return Err(DebugTraceError::InvalidSpecId {
+            spec_id: evm_config.spec_id,
+        });
+    }
+
+    if evm_config.spec_id > SpecId::MERGE && block_env.prevrandao.is_none() {
+        return Err(TransactionError::MissingPrevrandao.into());
+    }
+
+    for tx in transactions {
```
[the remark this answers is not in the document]
agostbiro: Hmm I changed this in ab6da884b3690a345aef011ad11a9584c17918eb and I agree in principle about using more explicit variable names, but I'm not sure that it makes the code more readable here. 

`transaction` is easily confused with `transactions` and I find that using common abbreviations like `i` or `tx` for loop variables communicates well that they're loop variables which is helpful when scanning code. 

In general, I like this quote: "variable names should be proportional in length to their scope".

crates/rethnet_evm/src/debug_trace.rs
```diff
@@ -0,0 +1,421 @@
...
+use revm::primitives::{BlockEnv, Bytes, CfgEnv, ExecutionResult, ResultAndState, SpecId};
+use revm::{EVMData, Inspector, JournalEntry};
+use std::collections::HashMap;
+use std::fmt::Debug;
+
+/// Get trace output for `debug_traceTransaction`
+#[cfg_attr(feature = "tracing", tracing::instrument)]
+pub fn debug_trace_transaction<BlockchainErrorT, StateErrorT>(
+    blockchain: &dyn SyncBlockchain<BlockchainErrorT, StateErrorT>,
+    // Take ownership of the state so that we can apply throw-away modifications on it
```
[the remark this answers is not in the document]
agostbiro: Done in 41f70b18d637fbe87f1b32fa021e5b60a669dbf4

crates/rethnet_evm/src/debug_trace.rs
```diff
@@ -0,0 +1,421 @@
...
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum DebugTraceError<BlockchainErrorT, StateErrorT> {
+    /// Invalid hardfork spec argument.
+    #[error("Invalid spec id: {spec_id:?}. `debug_traceTransaction` is not supported prior to Spurious Dragon")]
+    InvalidSpecId { spec_id: SpecId },
+    /// Invalid transaction hash argument.
+    #[error("Transaction hash {tx_hash} not found in block {block_number}")]
+    InvalidTransactionHash { tx_hash: B256, block_number: U256 },
```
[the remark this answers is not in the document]
agostbiro: Done in eb72c05cda12e43c4bc87b7f2f417052830909df

crates/rethnet_evm/src/debug_trace.rs
```diff
@@ -0,0 +1,421 @@
...
+    blockchain: &dyn SyncBlockchain<BlockchainErrorT, StateErrorT>,
+    // Take ownership of the state so that we can apply throw-away modifications on it
+    // TODO depends on https://github.com/NomicFoundation/hardhat/pull/4254
+    // mut state: Box<dyn SyncState<StateErrorT>>,
+    state: &mut dyn SyncState<StateErrorT>,
+    evm_config: CfgEnv,
+    trace_config: DebugTraceConfig,
+    block_env: BlockEnv,
+    transactions: Vec<SignedTransaction>,
+    transaction_hash: B256,
```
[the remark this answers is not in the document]
agostbiro: Done in 2fd7402b620870663677d5ecc2b494dbfb2702b8

crates/rethnet_evm/src/debug_trace.rs
```diff
@@ -0,0 +1,421 @@
...
+// Based on https://github.com/bluealloy/revm/blob/70cf969a25a45e3bb4e503926297d61a90c7eec5/crates/revm/src/inspector/tracer_eip3155.rs
+// Original licensed under the MIT license.
+struct TracerEip3155 {
+    config: DebugTraceConfig,
+
+    logs: Vec<DebugTraceLogItem>,
+
+    gas_inspector: GasInspector,
+
+    contract: B160,
```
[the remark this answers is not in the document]
agostbiro: Done in 398e4ed3095d8b8bf22499b8a1359d8844f27895

packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToRethnet.ts
```diff
@@ -476,3 +480,64 @@ function isEip2930SignedTransaction(
...
+    return result;
+  });
+
+  // REVM trace adds initial STOP that Hardhat doesn't expect
+  if (structLogs.length > 0 && structLogs[0].op === "STOP") {
+    structLogs.shift();
+  }
+
+  return {
+    failed: !rpcDebugTrace.pass,
```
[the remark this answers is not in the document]
agostbiro: It's just boolean negation. `pass` is true if the transaction succeeded and false otherwise, and we want to negate that to make `failed` true if the transaction didn't succeed. I'm using `pass` in the Rust implementation, because that's what EIP-3155 uses as well.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/rethnet.ts
```diff
@@ -372,9 +379,46 @@ export class RethnetAdapter implements VMAdapter {
   public async traceTransaction(
     hash: Buffer,
     block: Block,
-    config: RpcDebugTracingConfig
+    config: RpcDebugTracingConfig,
+    transactionsWithFakeSender?: Array<
```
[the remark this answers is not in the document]
agostbiro: That was just an attempt to fix the problem with impersonated transactions which is fixed now ea3938a972cb2ab64e2753f828b56dacf384869a. I removed this code in f092061c25fd5f4a964b66fd9086b5cc418a456d

packages/hardhat-core/test/internal/hardhat-network/provider/modules/debug.ts
```diff
@@ -203,8 +202,8 @@ describe.skip("Debug module", function () {
 
       const hardhatNetworkProvider = new HardhatNetworkProvider(
         {
-          hardfork: DEFAULT_HARDFORK,
-          chainId: DEFAULT_CHAIN_ID,
+          hardfork: "muirGlacier",
```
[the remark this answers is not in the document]
agostbiro: The tests are for pre-Berlin and the default hardfork is Shanghai. I think it was a mistake in the original.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/debug.ts
```diff
@@ -328,7 +327,7 @@ describe.skip("Debug module", function () {
       const hardhatNetworkProvider = new HardhatNetworkProvider(
         {
           hardfork: DEFAULT_HARDFORK,
-          chainId: DEFAULT_CHAIN_ID,
+          chainId: 1,
```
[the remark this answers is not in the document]
agostbiro: The tests are failing with `Error: Invalid transaction: InvalidChainId` when using the `DEFAULT_CHAIN_ID` which is 123.

packages/hardhat-core/test/internal/hardhat-network/provider/utils/assertEqualTraces.ts
agostbiro: Good call on not modifying the input. I fixed it by making copies in 373e99b4b73aa8b7b6013e10f20aa474533b6d51. 

This is testing against fixtures, so the alternative would be to modify the fixtures which is very cumbersome.

crates/rethnet_evm_napi/src/debug_trace.rs
```diff
@@ -0,0 +1,139 @@
...
+/// Get trace output for `debug_traceTransaction`
+#[napi]
+#[cfg_attr(feature = "tracing", tracing::instrument(skip_all))]
+pub async fn debug_trace_transaction(
+    blockchain: &Blockchain,
+    state_manager: &StateManager,
+    evm_config: ConfigOptions,
+    trace_config: DebugTraceConfig,
+    block_config: BlockConfig,
+    transactions: Vec<SignedTransaction>,
```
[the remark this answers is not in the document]
agostbiro: Yep, fixed by using `Vec<&PendingTransaction>`

crates/rethnet_evm_napi/src/lib.rs
```diff
@@ -9,6 +9,7 @@ mod blockchain;
 mod cast;
 mod config;
 mod context;
+mod debug_trace;
```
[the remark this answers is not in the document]
agostbiro: Sure, would be good to merge the two anyway as discussed. I'd do this later in a separate refactoring if ok with you.

packages/hardhat-core/test/internal/hardhat-network/provider/modules/debug.ts
```diff
@@ -353,14 +352,12 @@ describe.skip("Debug module", function () {
     });
 
     // see https://github.com/NomicFoundation/hardhat/issues/3519
-    it.skip("Should return the right values for a successful tx", async function () {
+    it("Should return the right values for a successful tx", async function () {
```
agostbiro, line 355: Please note that this test was disabled before, but I assume due to a different error than what I’m seeing now.

The test fails with this error in `rethnet` mode:

`Error: Post-merge transaction is missing prevrandao`

And it fails due to a timeout in `ethereumjs` mode.

Wodann: The mentioned failure is resolved in https://github.com/NomicFoundation/hardhat/pull/4398/commits/60179e7b673d5f0c951ff9ec794ba786a8fef134

Now `rethnet` mode also has a timeout

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
agostbiro: @Wodann after our discussion I tried to make comparisons with `ethereumjs` output, but turns out the `ethereumjs` output doesn't match the fixtures (this is probably why a bunch of debug trace tests were disabled), so I gave up.

crates/rethnet_evm/src/debug_trace.rs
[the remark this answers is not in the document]
Wodann: I'm happy to use that rule of thumb. Will need to keep the quote for the style guide of EDR 😁

packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToRethnet.ts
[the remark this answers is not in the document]
Wodann: 🤦

crates/rethnet_evm/src/debug_trace.rs
```diff
@@ -0,0 +1,432 @@
...
+        });
+    }
+
+    if evm_config.spec_id > SpecId::MERGE && block_env.prevrandao.is_none() {
+        return Err(TransactionError::MissingPrevrandao.into());
+    }
+
+    for transaction in transactions {
+        if transaction.hash() == transaction_hash {
+            let evm = build_evm(
```
Wodann, line 46: Could we build the `evm` outside of the look and just mutate the transaction?

crates/rethnet_evm/src/debug_trace.rs
```diff
@@ -0,0 +1,432 @@
...
+    pub error: Option<String>,
+    /// Array of all allocated values as hex strings.
+    pub memory: Option<Vec<String>>,
+    /// Map of all stored values with keys and values encoded as hex strings.
+    pub storage: Option<HashMap<String, String>>,
+}
+
+// Based on https://github.com/bluealloy/revm/blob/70cf969a25a45e3bb4e503926297d61a90c7eec5/crates/revm/src/inspector/tracer_eip3155.rs
+// Original licensed under the MIT license.
+struct TracerEip3155 {
```
Wodann, line 177: Could you split the tracer into a separate file? We'll need to use this for `traceCall` too

crates/rethnet_evm/src/debug_trace.rs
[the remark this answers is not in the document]
Wodann: The type is still a `B160` and not an `Address`

crates/rethnet_evm_napi/src/lib.rs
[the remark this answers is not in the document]
Wodann: Sure, if that's easier

packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToRethnet.ts
```diff
@@ -476,3 +480,64 @@ function isEip2930SignedTransaction(
...
+
+export function rethnetRpcDebugTraceToHardhat(
+  rpcDebugTrace: DebugTraceResult
+): RpcDebugTraceOutput {
+  const structLogs = rpcDebugTrace.structLogs.map((log) => {
+    const result: RpcStructLog = {
+      depth: Number(log.depth),
+      gas: Number(log.gas),
+      gasCost: Number(log.gasCost),
+      op: log.opName!,
```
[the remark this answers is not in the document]
Wodann: @agostbiro

Wodann (changes requested):

Wodann:
Were you able to get the dual mode adapter working properly, @agostbiro?

I'd also appreciate if you could have a look at the TS code, @fvictorio 🙏

agostbiro:
> Were you able to get the dual mode adapter working properly, @agostbiro?

No, unfortunately not. Looks like my comments got lost: https://github.com/NomicFoundation/hardhat/pull/4398#discussion_r1337185062

packages/hardhat-core/src/internal/hardhat-network/provider/vm/dual.ts
Wodann: @fvictorio is there a reason that ethereumjs output wouldn't match the fixtures? Is it post-processed?

crates/rethnet_evm/src/debug_trace.rs
agostbiro: I tried, but haven't managed to figure this out in reasonable time. It's difficult, because of the need to `commit` the preceding transactions. In order to commit a transaction, we first need to give them `evm` a reference to the state:

```rust
evm.database(DatabaseComponents {
    state: &state,
    block_hash: blockchain,
});
```

And then commit the changes

```rust
// Needs `&mut state`
state.commit(changes)
```

And if the `evm` needs to live past performing a transaction, we won't be able to borrow the `state` as mutable to commit the result, because we gave the `evm` a shared reference to it in `evm.database(...)`.

crates/rethnet_evm/src/debug_trace.rs
agostbiro: Sure, ok if I do that as part of https://github.com/NomicFoundation/edr/issues/192? Wanna think more about the best approach there.

crates/rethnet_evm/src/debug_trace.rs
Wodann: That makes sense. Thanks!

crates/rethnet_evm/src/debug_trace.rs
Wodann: Sure

Wodann (approved):

Wodann:
Merging this as it all tests are succeeding and I don't see any issues with the TS.

When you have time to look at the TS, could you please let us know if there are any issues, @fvictorio ?

PR #4401, opened 2023-09-19
fix: use TryCast for B256 in napi to avoid panics

agostbiro:
Use `TryCast` for `B256` in `napi` to avoid panics in case invalid data is passed.


crates/rethnet_evm_napi/src/block.rs
```diff
@@ -261,7 +262,10 @@ impl TryFrom<BlockHeader> for rethnet_eth::block::Header {
             base_fee_per_gas: value
                 .base_fee_per_gas
                 .map_or(Ok(None), |fee| fee.try_cast().map(Some))?,
-            withdrawals_root: value.withdrawals_root.map(|root| B256::from_slice(&root)),
+            withdrawals_root: value
+                .withdrawals_root
+                .map(TryCast::<B256>::try_cast)
+                .transpose()?,
```
Wodann, line 268: What does the `transpose` do here?

crates/rethnet_evm_napi/src/block.rs
```diff
@@ -136,7 +137,7 @@ impl TryFrom<BlockOptions> for rethnet_eth::block::BlockOptions {
             extra_data: value
                 .extra_data
                 .map(|extra_data| Bytes::copy_from_slice(&extra_data)),
-            mix_hash: value.mix_hash.map(|mix_hash| B256::from_slice(&mix_hash)),
+            mix_hash: value.mix_hash.map(TryCast::<B256>::try_cast).transpose()?,
```
Wodann, line 140: What does the `transpose` do here?

crates/rethnet_evm_napi/src/miner.rs
```diff
@@ -37,7 +37,7 @@ pub async fn mine_block(
     let reward: U256 = BigInt::try_cast(reward)?;
     let base_fee: Option<U256> =
         base_fee.map_or(Ok(None), |base_fee| BigInt::try_cast(base_fee).map(Some))?;
-    let prevrandao: Option<B256> = prevrandao.map(|prevrandao| B256::from_slice(&prevrandao));
+    let prevrandao: Option<B256> = prevrandao.map(TryCast::<B256>::try_cast).transpose()?;
```
Wodann, line 40: `transpose`?

crates/rethnet_evm_napi/src/block.rs
agostbiro: It turns the `Option<Result<T, E>>` into a `Result<Option<T>, E>>`

Wodann (approved):

PR #4403, opened 2023-09-19
Hardhat viem contracts

schaable:


packages/hardhat-viem/README.md
```diff
@@ -114,6 +114,70 @@ For a complete list of options, see:
...
+
+```typescript
+const contract = await this.hre.viem.deployContract("contractName", [
+  "arg1",
+  50,
+  "arg3",
+]);
+```
+
+By default, the first wallet client retrieve by `hre.viem.getWalletClients()` is used to deploy the contract. You can also specify a different wallet client by passing a third parameter, along with other properties such as `gas`, and `value`:
```
ChristopherDedominici, line 133: Small grammar error, retrieve -> retrieved

ChristopherDedominici, line 133: retrieve -> retrieved

packages/hardhat-viem/src/internal/errors.ts
```diff
@@ -44,3 +44,21 @@ You can find a list of supported networks here: https://viem.sh/docs/clients/cha
...
+      `Default wallet client not found. You can set the wallet client by passing it as a parameter in the deployContract function:
+
+const walletClient = await hre.viem.getWalletClient(address);
+const contractA = await hre.viem.deployContract("A", [], { walletClient });
+const contractB = await hre.viem.getContractAt("B", address, { walletClient });`
+    );
+  }
+}
+
+export class DeployContractError extends HardhatViemError {
```
fvictorio, line 60: Let's use a more specific error maybe? With a message like:

```
The deployment transaction '${txHash}` was mined in block '${blockNumber}' but its receipt doesn't contain a contract address
```

Otherwise, if someone runs into this (unlikely) scenario, it will be super hard to figure out what's going on.

Wdyt?

packages/hardhat-viem/src/internal/errors.ts
```diff
@@ -44,3 +44,21 @@ You can find a list of supported networks here: https://viem.sh/docs/clients/cha
     );
   }
 }
+
+export class DefaultWalletClientNotFoundError extends HardhatViemError {
+  constructor() {
+    super(
+      `Default wallet client not found. You can set the wallet client by passing it as a parameter in the deployContract function:
```
fvictorio, line 51: Maybe something like:

```suggestion
      `Default wallet client not found. This can happen if no accounts were configured for this network (network: '${networkName}').
      
Alternatively, you can set a custom wallet client by passing it as a parameter in the deployContract function:
```

Of course this means receiving the network name in the constructor.

fvictorio (approved):

Issue #4404, opened 2023-09-20
Print the evmVersion used when compiling


fvictorio:
See discussion in https://github.com/NomicFoundation/hardhat/issues/4391

SKYBITDev3:
Include `solc.version()` in the output.

PR #4414, opened 2023-09-25
Print the evmVersion used when compiling

ChristopherDedominici:
[See issue here](https://github.com/NomicFoundation/hardhat/issues/4404)

Log the evm versions used during compilation.


.changeset/spotty-tomatoes-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Added logic to print the evm version(s) used when compiling.
```
fvictorio, line 5:
```suggestion
Updated the compilation logs to include the targeted EVM versions.
```

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -1299,7 +1326,10 @@ subtask(TASK_COMPILE_SOLIDITY_LOG_COMPILATION_RESULT)
...
       if (count > 0) {
         console.log(
-          `Compiled ${count} Solidity ${pluralize(count, "file")} successfully`
+          `Compiled ${count} Solidity ${pluralize(
+            count,
+            "file"
+          )} successfully with evmVersions: ${evmVersions.join(", ")}.`
         );
       }
     }
```
fvictorio, line 1335: I think you can accomplish the same thing you did but modifying just this subtask:

```js
subtask(TASK_COMPILE_SOLIDITY_LOG_COMPILATION_RESULT)
  .addParam("compilationJobs", undefined, undefined, types.any)
  .addParam("quiet", undefined, undefined, types.boolean)
  .setAction(
    async ({ compilationJobs }: { compilationJobs: CompilationJob[] }) => {
      let count = 0;
      const evmVersionsSet = new Set<string>();

      for (const job of compilationJobs) {
        const version = job.getSolcConfig().version;
        const evmVersion =
          job.getSolcConfig().settings.evmVersion ?? `default for ${version}`;
        evmVersionsSet.add(evmVersion);

        count += job
          .getResolvedFiles()
          .filter((file) => job.emitsArtifacts(file)).length;
      }

      const evmVersions = Array.from(evmVersionsSet);

      if (count > 0) {
        console.log(
          `Compiled ${count} Solidity ${pluralize(
            count,
            "file"
          )} successfully with evmVersions: ${evmVersions.join(", ")}.`
        );
      }
    }
  );
```

Or am I missing something here?

ChristopherDedominici: Yes, not sure why I missed that. I moved the code.
But I kept the 2 sets to print the versions in order. I think that it looks nicer from a user prospective. What do you think?

schaable (approved):
Looks good! The only thing that I'd change are the variable names. I try to avoid using "Set", "Function" or anything related to the type, because it can cause unnecessary updates later if the type changes. But feel free to do what you think is best.

fvictorio (approved):

PR #4415, opened 2023-09-26
ci: edr evm napi release

agostbiro:


agostbiro:
The PR is ready for review now. The builds and tests succeed for the targets below. There is also a publish step added, but I haven't actually tested uploading a package to NPM yet. Let me know please if the setup looks good and I'll publish a `0.1.0-rc-1` package to the `@ignored` org.

The "EDR NPM release" Github Action workflow is currently configured to run on pull requests and pushes to `main`. The workflow will execute the build and test steps by default. It will only execute the release step if the commit message is a version number (e.g. `0.1.2`) or a version number with a dash suffix (e.g. `0.1.0-rc-2`). Bumping the package version number needs to be done manually.

The setup is generated with the `napi-rs` cli and adjusted where needed to account for differences in repo setup and non-Rust dependencies. 

### Targets

Release is supported for the following targets:

- `x86_64-apple-darwin`
- `aarch64-apple-darwin`
- `x86_64-pc-windows-msvc`
- `i686-pc-windows-msvc`
- `x86_64-unknown-linux-gnu`
- `x86_64-unknown-linux-musl`
- `aarch64-unknown-linux-musl`

It's not supported for these targets:

- `aarch64-unknown-linux-gnu`
- `aarch64-pc-windows-msvc`

Adding support for `aarch64-unknown-linux-gnu` is probably doable short term. It's currently not supported due to an OpenSSL build error that I [reported](https://github.com/napi-rs/napi-rs/issues/1741) to the `napi-rs` team. I tried to resolve it myself, but it's difficult, because I can't build the build container locally on my Mac. Hopefully a response/fix from the `napi-rs` team can clear this up.

Adding support for `aarch64-pc-windows-msvc` is more involved. It's currently not supported, because the [`secp256k1`](https://docs.rs/secp256k1/latest/secp256k1/) crate doesn't build on that target. We could try to patch it, but the project's CI is [not configured](https://github.com/rust-bitcoin/rust-secp256k1/blob/da4f67b274c3061717723a568cfb182e4e2e7cce/.github/workflows/cross.yml#L25) for that target. A more future proof solution would be probably to switch to the pure Rust [`k256`](https://docs.rs/k256/latest/k256/) crate.

.github/workflows/edr-npm-release.yml
```diff
@@ -0,0 +1,329 @@
+name: EDR NPM release
```
Wodann, line 1: This file has a lot of code duplication and thus is prone to error. Another downside of the workflow being encoded in the GitHub CI instead of a Rust CLI app is that it cannot be reproduced locally.

With the knowledge you've gained from doing this, can you easily avoid duplication and configure it in a Rust CLI app?

crates/rethnet_evm_napi/index.d.ts
```diff
@@ -0,0 +1,617 @@
+/* tslint:disable */
```
Wodann, line 1: Similar to Slang, we should move auto-generated files to a dedicated folder, e.g. `generated`

crates/rethnet_evm_napi/npm/darwin-arm64/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "cpu": [
+    "arm64"
+  ],
+  "main": "edr.darwin-arm64.node",
+  "files": [
+    "edr.darwin-arm64.node"
+  ],
+  "license": "MIT",
+  "engines": {
+    "node": ">= 12"
```
Wodann, line 16:
```suggestion
    "node": ">= 18"
```

crates/rethnet_evm_napi/package.json
```diff
@@ -1,25 +1,39 @@
...
     "devDependencies": {
-        "@napi-rs/cli": "^2.11.4",
+        "@napi-rs/cli": "^2.16.3",
         "chai": "^4.3.6",
         "@nomicfoundation/ethereumjs-util": "^8.0.0",
         "mocha": "^10.0.0",
         "typescript": "~4.5.2"
+    },
+    "engines": {
+        "node": ">= 12"
```
Wodann, line 26:
```suggestion
        "node": ">= 18"
```

crates/rethnet_evm_napi/test/evm/StateManager.ts
```diff
@@ -1,7 +1,7 @@
 import { expect } from "chai";
 import { Address, KECCAK256_NULL } from "@nomicfoundation/ethereumjs-util";
 
-import { Account, Bytecode, RethnetContext, State } from "../..";
+import { Account, Bytecode, RethnetContext, State } from "../../index.js";
```
Wodann, line 4: Why does it no longer recognise the `*.d.ts`?

Wodann (changes requested):

.github/workflows/edr-npm-release.yml
```diff
@@ -0,0 +1,329 @@
...
+      - main
+    tags-ignore:
+      - "**"
+    paths-ignore:
+      - "**/*.md"
+      - LICENSE
+      - "**/*.gitignore"
+      - .editorconfig
+      - docs/**
+  pull_request: null
```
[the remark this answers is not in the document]
agostbiro: `tags-ignore` means it's not running when a new tag is pushed to `main`, `paths-ignore` means that it's not running when only the matched file paths are changed on a push to `main`. `pull_request: null` means run on pull requests without further configuration if I'm not mistaken.

crates/rethnet_evm_napi/test/evm/StateManager.ts
agostbiro: Not sure tbh but it was failing in the CI without that change so I would rather not touch it 😅

.github/workflows/edr-npm-release.yml
```diff
@@ -0,0 +1,329 @@
...
+      matrix:
+        settings:
+          - host: macos-latest
+            target: x86_64-apple-darwin
+          - host: windows-latest
+            target: x86_64-pc-windows-msvc
+        node:
+          - "14"
+          - "16"
+          - "18"
```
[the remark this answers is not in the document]
agostbiro: Done in [de59d8f](https://github.com/NomicFoundation/hardhat/pull/4415/commits/de59d8f772e3ec7d3f904e871e7b44ecc28083bb)

crates/rethnet_evm_napi/package.json
agostbiro: Done in [508bd82](https://github.com/NomicFoundation/hardhat/pull/4415/commits/508bd820d3d579ca155328d25d24142b7238644a)

.github/workflows/edr-npm-release.yml
agostbiro: Yes I agree that this needs to be improved. It's not trivial, but definitely doable. I created https://github.com/NomicFoundation/edr/issues/198 as a follow up.

crates/rethnet_evm_napi/index.d.ts
agostbiro: While I agree with this change, the current publish workflow assumes that the files are there and, as noted, it's pretty fragile, so I would rather not touch this and take care of it in https://github.com/NomicFoundation/edr/issues/198 if you don't mind.

.github/workflows/edr-npm-release.yml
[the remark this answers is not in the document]
Wodann: Thanks!

Wodann (approved):
CI is failing because node version 16 is still part of the matrix. We don't need to and can't support that (due to a bug in that version). Could you remove that and add version 20 before merging?

.github/workflows/edr-npm-release.yml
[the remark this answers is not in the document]
agostbiro: Sorry, I overlooked this. Removed Node 16 and added Node 20 to matrix in [fe57696](https://github.com/NomicFoundation/hardhat/pull/4415/commits/fe57696483e634e0fdf2aae4b340f9de43c5b272)

PR #4416, opened 2023-09-26
docs: add benchmark & profiling book chapters

Wodann:


book/src/01_getting_started/05_benchmark.md
```diff
@@ -0,0 +1,63 @@
...
+The expected output will look similar to this:
+
+```bash
+node_modules/hardhat: symbolic link to ../../hardhat/packages/hardhat-core
+```
+
+To prevent the benchmark from being tainted by smart contract compilation, we first run:
+
+```bash
+hardhat compile
```
Wodann, line 49: Might need `npx hardhat compile` and `npx hardhat test`

book/src/01_getting_started/05_benchmark.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Automated
+
+## Manual
+
+To measure real-world performance, we use a build of [Hardhat](https://github.com/NomicFoundation/hardhat) with EDR in third-party projects.
+To make a local build of Hardhat available for linking in other packages, run:
+
+```bash
+cd packages/hardhat-core &&
```
agostbiro, line 16: I think we need to add `yarn build` step here as well.

book/src/01_getting_started/05_benchmark.md
agostbiro: Yes, confirmed that

PR #4417, opened 2023-09-26
fix: revm bug when balance check disabled

Wodann:


agostbiro (approved):
There is a JS lint failure, otherwise LGTM. Approving to expedite merging once the JS lint is fixed.

PR #4419, opened 2023-09-26
improvement: add more tracing markers

Wodann:


crates/rethnet_evm/src/state/fork.rs
```diff
@@ -86,13 +86,13 @@ impl ForkState {
...
 
 impl Clone for ForkState {
+    #[cfg_attr(feature = "tracing", tracing::instrument(skip_all))]
     fn clone(&self) -> Self {
         Self {
             local_state: self.local_state.clone(),
             remote_state: self.remote_state.clone(),
             removed_storage_slots: self.removed_storage_slots.clone(),
             fork_block_number: self.fork_block_number,
-            state_root_to_state: RwLock::new(self.state_root_to_state.read().clone()),
```
agostbiro, line 95: Missing field here I think

agostbiro (approved):
LGTM other than the missing field. Approving to expedite merging.

crates/rethnet_evm/src/state/fork.rs
Wodann: That field was deleted

PR #4421, opened 2023-09-27
refactor: use u64 instead of U256

Wodann:
While working on this, I also had to fix some errors in the JSON-RPC client APIs


crates/edr_eth/src/remote/client.rs
```diff
@@ -636,24 +637,26 @@ impl RpcClient {
...
     /// Calls `eth_chainId` and returns the chain ID.
-    pub async fn chain_id(&self) -> Result<U256, RpcClientError> {
+    pub async fn chain_id(&self) -> Result<u64, RpcClientError> {
         let chain_id = *self
             .chain_id
             .get_or_try_init(|| async {
-                self.call_without_cache(MethodInvocation::ChainId()).await
+                self.call_without_cache::<U64>(MethodInvocation::ChainId())
+                    .await
+                    .map(|chain_id| chain_id.as_limbs()[0])
```
agostbiro, line 659: I think `try_into` should work here.

crates/edr_eth/src/remote/client.rs
```diff
@@ -1753,13 +1745,13 @@ mod tests {
                 .expect("failed to parse address");
 
             let error = TestRpcClient::new(&alchemy_url)
```
agostbiro, line 1747: I feel like we could just delete this test. As far as I understand, it's testing for a specific error response from Alchemy, which is liable to change.

agostbiro (changes requested):

crates/edr_eth/src/remote/client.rs
Wodann: It's infallible though. So we'd be unnecessarily wrapping it in a result and then calling expect

crates/edr_eth/src/remote/client.rs
Wodann: Do you mean to remove the whole test or the specific error?

I feel that the tests did provide value in this case, as some APIs changed from returning an error to returning an option. Having the tests, helped us catch that instead.

agostbiro: Ah ok I didn't catch that 👍 I'd probably remove the check for a specific error code or message, but I'll leave it up to you.

agostbiro (approved):

PR #4426, opened 2023-09-28
Hardhat viem type generation

schaable:
Adds type generation for contracts. You can use the following snippets for testing:

Type errors: https://gist.github.com/schaable/40316c9d1f86e4121c0a2119bd3803bc
Working deploy: https://gist.github.com/schaable/2cd1938b191d9d07eb4f41f163f75a6d

### DIFF
--- a/.github/workflows/compile-with-typescript-v4.yml
+++ b/.github/workflows/compile-with-typescript-v4.yml
@@ -21,7 +21,9 @@ jobs:
         run: |
           sed -i 's/"typescript": "~5.0.0"/"typescript": "^4.0.0"/' package.json packages/*/package.json && yarn
       # hardhat-viem is the only package that requires TypeScript v5
-      - name: Remove hardhat-viem
+      - name: Remove hardhat-viem directory
         run: rm -fr packages/hardhat-viem
+      - name: Remove hardhat-viem from the build script
+        run: sed -i 's/packages\/hardhat-viem//' package.json
       - name: Build
         run: yarn build


packages/hardhat-viem/src/internal/errors.ts
```diff
@@ -57,6 +57,16 @@ const contractB = await hre.viem.getContractAt("B", address, { walletClient });`
   }
 }
 
+export class InvalidConfirmationsError extends HardhatViemError {
+  constructor(confirmations: number) {
+    super(
+      confirmations < 0
```
ChristopherDedominici, line 63: Can the variable confirmations be > 0 ?

packages/hardhat-viem/src/internal/type-extensions.ts
```diff
@@ -30,16 +31,24 @@ declare module "hardhat/types/runtime" {
...
+declare module "hardhat/types/artifacts" {
+  // eslint-disable-next-line @typescript-eslint/no-empty-interface
+  interface ArtifactsMap {}
+
+  interface Artifacts {
+    readArtifact<ArgT extends keyof ArtifactsMap>(
+      contractNameOrFullyQualifiedName: ArgT
+    ): ArtifactsMap[ArgT];
+
+    readArtifactSync<ArgT extends keyof ArtifactsMap>(
```
ChristopherDedominici, line 50: Should this one return a Promise?

schaable: I guess you refer to `readArtifact` and not `readArtifactSync` and it makes a valid point. I see that the generic type definition for `readArtifact` indeed returns a promise: `public async readArtifact(name: string): Promise<Artifact>`. I'm going to change it and test that everything still works.

packages/hardhat-viem/src/internal/errors.ts
schaable: What do you mean? Confirmations are expected to be higher than 0. If it is below 0, we want to throw an error. I don't expect this error to be invoked with confirmations > 0 (see [this](https://github.com/NomicFoundation/hardhat/pull/4426/files#diff-bd2b6167ff354b394e83027a2bb3d5f28b236d392dc745de9c637e0aa72bf5e7R68)). Do you think we need to double check?

packages/hardhat-viem/src/internal/type-extensions.ts
schaable: You were right, good catch! 🚀

packages/hardhat-viem/README.md
```diff
@@ -142,10 +142,37 @@ const contractA = await this.hre.viem.deployContract(
...
 );
 ```
 
+##### Send deployment transaction
+
+By default, the `deployContract` method sends a deployment transaction to the blockchain and waits for the transaction to be mined. If you want to send the transaction without waiting for it to be mined, you can do so by using `sendDeploymentTransaction`:
+
+```typescript
+const { contract: contractName, deploymentTransaction } =
+  await this.hre.viem.sendDeploymentTransaction(
```
fvictorio, line 156: Where does this `this` come from?

packages/hardhat-viem/src/internal/errors.ts
fvictorio: I was also confused by this code. I think I would split this into two errors: one for the case where someone uses 0 and another one for a negative number. The latter doesn't even need to be a custom error, it can be the generic `HardhatViemError` with a proper message.

packages/hardhat-viem/src/internal/tasks.ts
```diff
@@ -0,0 +1,271 @@
...
+          })
+        );
+
+        const fp: Array<Promise<void>> = [];
+        for (const { contractName, declaration } of contractTypeData) {
+          fp.push(writeFile(join(srcDir, `${contractName}.ts`), declaration));
+        }
+
+        const indexTs = generateIndexFile(contractTypeData);
+        fp.push(writeFile(join(srcDir, "index.ts"), indexTs));
```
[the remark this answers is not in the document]
schaable: I think that was because we had ts code in some of them. But now it's all type declarations, so it makes sense to change the extension to d.ts. Additionally, I've renamed some of the file names for clarity.

packages/hardhat-viem/src/internal/tasks.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  const constructorArgs =
+    inputs.length > 0
+      ? `constructorArgs: [${inputs
+          .map(
+            ({ name, type }) =>
+              `AbiParameterToPrimitiveType<${JSON.stringify({ name, type })}>`
+          )
+          .join(", ")}]`
+      : `constructorArgs?: []`;
```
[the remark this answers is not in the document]
schaable: Looking good!

image

packages/hardhat-viem/src/internal/tasks.ts
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    return superRes;
+  }
+);
+
+/**
+ * This override generates a .ts file per contract, and a file.d.ts
+ * per solidity file, which is used in conjunction to artifacts.d.ts
+ * to type hre.artifacts.
```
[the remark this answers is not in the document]
schaable: I've updated all the comments.

packages/hardhat-viem/src/internal/tasks.ts
```diff
@@ -0,0 +1,302 @@
...
+ * - If the 'name' is empty:
+ *   "AbiParameterToPrimitiveType<{ name: string; type: string; }>"
+ */
+function getArgType(name: string, type: string) {
+  const argType = `AbiParameterToPrimitiveType<${JSON.stringify({
+    name,
+    type,
+  })}>`;
+
+  return name !== "" ? `${name}: ${argType}` : argType;
```
fvictorio, line 278: How sure are we that `name` cannot be `undefined` here? It comes from casting an `any` read from a json file, so I'm not 100% sure that it's impossible, but maybe the ABI output is specified in a way that makes `name` always a string.

fvictorio: `abitype` at least seems to think it can be `undefined`: https://github.com/frangio/abitype/blob/3879b593580d3fecc0d585d82f4876c7e6bd8d7b/src/abi.ts#L101

fvictorio: Solc v0.5.1, which is the earliest version [we support](https://hardhat.org/hardhat-runner/docs/reference/solidity-support), generates empty strings for anonymous parameters. So idk. I think I would be defensive anyway just in case, but feel free to leave it as it is if you think it's unnecessary.

packages/hardhat-viem/src/internal/tasks.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    const duplicateContractNames = await findDuplicateContractNames(artifacts);
+
+    const duplicateArtifactsDTs = generateDuplicateArtifactsDefinition(
+      duplicateContractNames
+    );
+
+    try {
+      await writeFile(
+        join(config.paths.artifacts, "duplicate-artifacts.d.ts"),
```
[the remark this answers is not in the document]
schaable: I get your point here. The main reason why this file was renamed, was to not overcomplicate the logic of deleting obsolete types. As I renamed `file.d.ts` to `artifacts.d.ts` the logic to get the existing artifacts started to match the root `artifacts.d.ts`, and the easiest solution was to rename it. I'll work on a better approach 👍

fvictorio (approved):

fvictorio (approved):

PR #4427, opened 2023-09-28
add plugin hardhat-logger

kruglay:
add plugin "hardhat-logger"


fvictorio (approved):

fvictorio:
Thanks, that's a cool way to use `extendProvider`!

PR #4428, opened 2023-09-28
Fix cache in remote client

fvictorio:
Opening this PR mainly to show the problem, but feel free to change the approach.

The problem was that the temp files were created in `/tmp`, but in my machine (and many others, I'm sure), `/` and `/home` are different mount points, causing the `rename` call to fail.

The solution is to create the temporary file "close to" the cache dir. Alternatives to my approach that come to mind:

1. Create a `tmp` directory under the fork cache dir, instead of putting the files in the root of that dir
2. Create the file in the directory where it will end up being, but with an initial `.json.tmp` extension, and then remove the `.tmp` part

### DIFF
--- a/crates/rethnet_eth/src/remote/client.rs
+++ b/crates/rethnet_eth/src/remote/client.rs
@@ -393,7 +393,8 @@ impl RpcClient {
         );
 
         // 1. Write to a random temporary file first to avoid race conditions.
-        let tempfile = match tokio::task::spawn_blocking(NamedTempFile::new).await? {
+        let cache_dir = self.rpc_cache_dir.clone();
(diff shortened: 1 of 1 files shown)


agostbiro:
Thanks @fvictorio, good catch and goo suggestions! I'd take a slightly different approach detailed in this [issue](https://github.com/NomicFoundation/edr/issues/196) and there will be some lifetime business that needs to be resolved for the dir names to avoid a clone, so I think it's easiest if I take care of this next week, but feel free to take a stab at it if you feel like.

agostbiro:
Closing in favor of https://github.com/NomicFoundation/hardhat/pull/4434

PR #4434, opened 2023-10-04
fix: fix RPC cache when tmp is on different filesystem

agostbiro:
Fix RPC cache when `tmp` is on different filesystem by using a temporary directory under the RPC cache dir instead of the system default `tmp` directory. The downside of this approach is that left-over files won't be cleared up in case we crash between creating a temporary file and moving it into its place. This is an acceptable trade-off imo.

Closes https://github.com/NomicFoundation/edr/issues/196

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2492,6 +2492,7 @@ dependencies = [
  "thiserror",
  "tokio",
  "triehash",
+ "uuid",
  "walkdir",
 ]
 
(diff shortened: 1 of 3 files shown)


Wodann:
After merging to `rethnet/main`, could you please also cherry pick the change to the edr merge branch?

agostbiro:
> After merging to `rethnet/main`, could you please also cherry pick the change to the edr merge branch?

Yep, merged it and cherry picked to https://github.com/NomicFoundation/hardhat/pull/4410

cc @fvictorio

PR #4438, opened 2023-10-04
fix: support for both input or data field in rpcCallRequest

lsheva:
Issue: 
go-ethereum from version 13.0 stopped working with local hardhat node. They did a change replacing `data` to `input` field in `eth_call` and other methods https://github.com/ethereum/go-ethereum/pull/28078 referencing [Ethereum specification](https://github.com/ethereum/execution-apis/blob/main/src/schemas/transaction.yaml#L52-L54)

Update:
replaces rpcCallRequest to use `input` field instead of `data`. Fallbacks to `data` if input not set for backward compatibility.

### DIFF
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/input/callRequest.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/input/callRequest.ts
@@ -19,7 +19,8 @@ export const rpcCallRequest = t.type(
     gas: optionalOrNullable(rpcQuantity),
     gasPrice: optionalOrNullable(rpcQuantity),
     value: optionalOrNullable(rpcQuantity),
-    data: optionalOrNullable(rpcData),
+    data: optionalOrNullable(rpcData), // deprecated, use input instead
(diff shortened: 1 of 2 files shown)


rzmahmood (approved):

marlon-wiprud:
keeping an eye on this as its blocking usage of geth v1.13+ with hardhat

compojoom:
What needs to be done for this to get merged? 
I spent half a day figuring out why my eth_calls were working fine against goerli nodes, but not against hardhat. Iit turns out that web3js by default uses input and I was ending with unrecognized selectors. 

@fvictorio

fvictorio (changes requested):
Hi everyone, and sorry for not responding before.

As it is right now, we cannot merge this PR for several reasons:

- There are no tests
- There doesn't seem to be an equivalent change for `eth_sendTransaction` (and I'm not sure if this change applies to `eth_estimateGas`; I think it does, but since there are no tests it's hard to know)
- If we accept both `input` and `data`, then a call/gas estimate/tx using both should throw an invalid input error.

PR #4443, opened 2023-10-05
feat: Cancun support in EDR

Wodann:


crates/rethnet_eth/src/transaction/request/eip4844.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub to: Address,
+    pub value: U256,
+    pub input: Bytes,
+    pub access_list: Vec<AccessListItem>,
+    pub blob_hashes: Vec<B256>,
+}
+
+impl Eip4844TransactionRequest {
+    /// Computes the hash of the transaction.
+    pub fn hash(&self) -> B256 {
```
agostbiro, line 33: It'd be good to add a test vector for this

agostbiro (approved):
Thanks @Wodann, Rust part looks good with an optional suggestion for improvement.

crates/rethnet_eth/src/transaction/request/eip4844.rs
Wodann: I think this might specifically have to do with blobs as the raw encoded transaction does not match the RLP format that we're validating. I'll skip this for now until we find something that matches our use case more accurately

Wodann:
Merged, but if there are any concerns about the HH changes, I'll correct them in a follow-up PR for you, @fvictorio

PR #4446, opened 2023-10-05
Add a better mechanism for variables management

ChristopherDedominici:


packages/hardhat-core/test/internal/core/secrets/secrets-manager.ts
```diff
@@ -0,0 +1,126 @@
+import fs from "fs-extra";
+import { expect } from "chai";
+import { SecretsManager } from "../../../../src/internal/core/secrets/secrets-manager";
+
+const TMP_FILE_PATH = `${__dirname}/test-secrets.json`;
```
fvictorio, line 5: This is better done with [`os.tmpdir`](https://nodejs.org/api/os.html#ostmpdir) in a `before` or `beforeEach`.

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -56,6 +57,8 @@ const log = debug("hardhat:core:cli");
 const ANALYTICS_SLOW_TASK_THRESHOLD = 300;
 const SHOULD_SHOW_STACK_TRACES_BY_DEFAULT = isRunningOnCiServer();
 
+const secretsManager = new SecretsManager(getSecretsFilePath());
```
fvictorio, line 60: We should do this lazily, not in each run of `hh`.

packages/hardhat-core/src/internal/core/config/config-env.ts
```diff
@@ -173,3 +177,25 @@ export function experimentalAddHardhatNetworkMessageTraceHook(
...
+function getSecret(key: string, defaultValue?: string): string {
+  const secretsManager = new SecretsManager(getSecretsFilePath());
+
+  const value = secretsManager.get(key);
+
+  if (value !== undefined) return value;
+
+  if (defaultValue !== undefined) return defaultValue;
+
+  throw new HardhatError(ERRORS.ARGUMENTS.VALUE_NOT_FOUND_FOR_ARGUMENT, {
```
fvictorio, line 197: I think this error is too generic and will be a bit confusing.

Maybe we should create a new error category for secrets (like we have now for arguments, artifacts, plugins, etc.)? Since we already added three, and I can imagine us adding more.

packages/hardhat-core/src/internal/cli/cli.ts
```diff
@@ -445,6 +452,91 @@ async function createNewProject() {
   throw new HardhatError(ERRORS.GENERAL.NOT_IN_INTERACTIVE_SHELL);
 }
 
+async function handleSecrets(args: string[]) {
```
fvictorio, line 455: Let's move this to another file in the same directory, something like `internal/cli/secrets` or whatever.

packages/hardhat-core/test/internal/core/config/secrets-manager.ts
```diff
@@ -0,0 +1,31 @@
+import { expect } from "chai";
+import sinon from "sinon";
+import { secrets } from "../../../../src/internal/core/config/config-env";
+import { SecretsManager } from "../../../../src/internal/core/secrets/secrets-manager";
+
+const spyFunctionSecretManagerGet = sinon.stub(SecretsManager.prototype, "get");
```
schaable: This is usually done in a `before`/`beforeEach` block.

packages/hardhat-core/src/internal/core/secrets/secrets-manager.ts
```diff
@@ -0,0 +1,83 @@
...
+    return Object.keys(secrets);
+  }
+
+  public delete(key: string): boolean {
+    const secrets = this._readSecrets();
+
+    if (secrets[key] === undefined) return false;
+
+    delete secrets[key];
+    this._writeSecrets(secrets);
```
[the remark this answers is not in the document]
ChristopherDedominici: Do you think that is better to catch the error here or have a hard fail? 
I'd prefer a hard fail instead of returning false considering that if the write fails it means that something is seriously wrong. False would be too general.
I could catch and re-throw the error with more info though.

packages/hardhat-core/src/internal/core/secrets/secrets-manager.ts
```diff
@@ -0,0 +1,83 @@
...
+          secrets: {},
+        },
+        { spaces: 2 }
+      );
+    }
+  }
+
+  private _writeSecrets(secrets: Record<string, Secret>) {
+    this._cache.secrets = secrets;
+    fs.writeJSONSync(this._secretsFilePath, this._cache, { spaces: 2 });
```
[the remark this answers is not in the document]
ChristopherDedominici: True, but if the code fails the cache is gonna be lost anyway because the application will stop. 
I set the cache before so it's easier to write to the JSON file because all the JSON structure is stored in the cache. 
If I have to move the cache update after the writing on file I'd have to create a tmp variable that is a hardcopy of the cache, update this tmp variable and store it.
To avoid this step I decided to move the cache update before.
Does it make sense to you or there is a potential problem that I'm missing?

packages/hardhat-core/src/internal/core/config/config-env.ts
```diff
@@ -173,3 +177,25 @@ export function experimentalAddHardhatNetworkMessageTraceHook(
...
+};
+
+function getSecret(key: string, defaultValue?: string): string {
+  const secretsManager = new SecretsManager(getSecretsFilePath());
+
+  const value = secretsManager.get(key);
+
+  if (value !== undefined) return value;
+
+  if (defaultValue !== undefined) return defaultValue;
```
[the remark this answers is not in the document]
ChristopherDedominici: I would keep it here, I would prefer to have the class doing only operations with the value stored in the file. 
@fvictorio what do you think?

packages/hardhat-core/src/internal/core/secrets/secrets-manager.ts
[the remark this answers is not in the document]
schaable: Sounds good! Thanks for the update 👍

packages/hardhat-core/src/internal/core/secrets/secrets-manager.ts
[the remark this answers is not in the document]
schaable: Catch and re-throw with better messaging sounds good!

packages/hardhat-core/src/internal/core/secrets/secrets-manager.ts
```diff
@@ -0,0 +1,83 @@
+import fs from "fs-extra";
```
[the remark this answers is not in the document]
fvictorio: I'd say we don't, it's going to be a distraction. When we remove `fs-extra` we'll have to do it everywhere anyway, so doing it once more here later it's negligible extra effort.

packages/hardhat-core/src/builtin-tasks/vars.ts
```diff
@@ -0,0 +1,46 @@
...
+
+varsScope
+  .task("path", "Show the file path where all the key-value pairs are stored")
+  .setAction(async () => {
+    throw new HardhatError(ERRORS.VARS.ONLY_MANAGED_IN_CLI);
+  });
+
+varsScope
+  .task("setup", "Display a list of key-value pairs that need to be set up")
+  .setAction(async () => {});
```
schaable, line 46: Why is `setup` not throwing an error?

packages/hardhat-core/src/internal/cli/vars.ts
```diff
@@ -0,0 +1,223 @@
...
+
+async function set(key: string, value?: string): Promise<number> {
+  const varsManager = HardhatContext.getHardhatContext().varsManager;
+
+  varsManager.validateKey(key);
+
+  varsManager.set(key, value ?? (await getVarValue()));
+
+  console.warn(
+    `Key-value pair stored at the following path: ${HardhatContext.getHardhatContext().varsManager.getStoragePath()}`
```
schaable, line 48: can we use `varsManager.getStoragePath()`?

packages/hardhat-core/src/internal/core/vars/vars-manager.ts
```diff
@@ -0,0 +1,145 @@
...
+import debug from "debug";
+import { HardhatError } from "../errors";
+import { ERRORS } from "../errors-list";
+
+interface Var {
+  value: string;
+}
+
+interface VarsFile {
+  _format: string; // Version of the json vars file
```
schaable, line 11: why not just `_version`?

packages/hardhat-core/src/internal/core/vars/vars-manager.ts
```diff
@@ -0,0 +1,145 @@
...
+  private _loadVarsFromEnv() {
+    log("Loading ENV variables if any");
+
+    for (const key in process.env) {
+      if (key.startsWith(this._ENV_VAR_PREFIX)) {
+        if (
+          process.env[key] === undefined ||
+          process.env[key]!.replace(/[\s\t]/g, "").length === 0
+        ) {
+          throw new HardhatError(ERRORS.ARGUMENTS.INVALID_ENV_VAR_VALUE, {
```
schaable, line 119: Is this ok?

image

packages/hardhat-core/src/internal/core/vars/vars-manager.ts
```diff
@@ -0,0 +1,145 @@
...
+            varName: key,
+            value: process.env[key]!,
+          });
+        }
+
+        const envKey = key.replace(this._ENV_VAR_PREFIX, "");
+        this.validateKey(envKey);
+
+        // Store only in cache, not in a file, as the vars are sourced from environment variables
+        this._cacheEnv.vars[envKey] = { value: process.env[key]! };
```
schaable, line 129: if you do `const envVar = process.env[key];` before the conditional at line 115, you can drop the `!` here and in the conditional, which I think is clearer.

packages/hardhat-core/src/internal/core/vars/vars-manager.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    vars[key] = { value };
+    this._writeVars(vars);
+  }
+
+  public has(key: string): boolean {
+    return key in this._readVars();
+  }
+
+  public hasWithEnvVars(key: string): boolean {
```
schaable, line 50: why not using a second optional parameter in `has` called `includeEnvVars` or something like that?

packages/hardhat-core/src/internal/core/vars/vars-manager.ts
```diff
@@ -0,0 +1,145 @@
...
+  public get(key: string, defaultValue?: string): string | undefined {
+    return this._readVars()[key]?.value ?? defaultValue;
+  }
+
+  public getWithEnvVars(
+    key: string,
+    defaultValue?: string
+  ): string | undefined {
+    return this._readVars(true)[key]?.value ?? defaultValue;
+  }
```
schaable, line 63: why not using a second optional parameter in `has` /  `get` called `includeEnvVars` or something like that?

packages/hardhat-core/src/internal/cli/vars.ts
```diff
@@ -0,0 +1,223 @@
...
+}
+
+async function set(key: string, value?: string): Promise<number> {
+  const varsManager = HardhatContext.getHardhatContext().varsManager;
+
+  varsManager.validateKey(key);
+
+  varsManager.set(key, value ?? (await getVarValue()));
+
+  console.warn(
```
schaable, line 47: IMO this should be a log, not a warn.

packages/hardhat-core/src/internal/cli/vars.ts
```diff
@@ -0,0 +1,223 @@
...
+  return 1;
+}
+
+function list(): number {
+  const keys = HardhatContext.getHardhatContext().varsManager.list();
+
+  if (keys.length > 0) {
+    keys.forEach((k) => console.log(k));
+
+    console.warn(
```
schaable, line 74: same here

packages/hardhat-core/src/internal/cli/vars.ts
```diff
@@ -0,0 +1,223 @@
...
+  } else {
+    console.warn(chalk.yellow(`There are no key-value pairs stored`));
+  }
+
+  return 0;
+}
+
+function del(key: string): number {
+  if (HardhatContext.getHardhatContext().varsManager.delete(key)) {
+    console.warn(
```
schaable, line 86: and here

packages/hardhat-core/src/internal/core/vars/vars-manager-setup.ts
```diff
@@ -0,0 +1,74 @@
...
+    return this.has(key);
+  }
+
+  // During setup we ignore env variables, so this function is the same as 'get'
+  public getWithEnvVars(
+    key: string,
+    defaultValue?: string | undefined
+  ): string | undefined {
+    return this.get(key, defaultValue);
+  }
```
schaable, line 30: If you make the change about the 2nd parameter in `has` / `get`, these methods wouldn't be needed.

packages/hardhat-core/src/internal/core/vars/vars-manager-setup.ts
```diff
@@ -0,0 +1,74 @@
...
+  // Required vars have priority over optional vars.
+  public get(key: string, defaultValue?: string): string {
+    log(`function 'get' called with key '${key}'`);
+
+    const value = super.get(key, defaultValue);
+
+    if (value === undefined) {
+      // The key is not present and there is no default value.
+      // Remove from optional vars (if it exists) and add to required vars.
+      if (this._optionalVars.has(key)) this._optionalVars.delete(key);
```
schaable, line 54: Personal preference: I'm not a fan of ignoring the `{}`

packages/hardhat-core/src/internal/core/vars/vars-manager-setup.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  // Checks if the key exists, and updates optional sets accordingly.
+  // Required vars have priority over optional vars.
+  public has(key: string): boolean {
+    log(`function 'has' called with key '${key}'`);
+
+    if (super.has(key)) return true;
+
+    if (!this._requiredVars.has(key)) this._optionalVars.add(key);
```
schaable, line 39: Personal preference: I'm not a fan of ignoring the curly braces `{ }` in code blocks.

packages/hardhat-core/src/internal/cli/vars.ts
```diff
@@ -0,0 +1,223 @@
...
+  return 0;
+}
+
+function loadConfigFile(configPath: string | undefined) {
+  const configEnv = require(`../core/config/config-env`);
+
+  const globalAsAny: any = global;
+  Object.entries(configEnv).forEach(
+    ([key, value]) => (globalAsAny[key] = value)
+  );
```
schaable, line 133: what is this doing?

schaable (commented):
Left some comments, mostly minor things and questions: Nice work!

packages/hardhat-core/src/builtin-tasks/vars.ts
```diff
@@ -0,0 +1,46 @@
+import { HardhatError } from "../internal/core/errors";
+import { scope } from "../internal/core/config/config-env";
+import { ERRORS } from "../internal/core/errors-list";
+
+const varsScope = scope("vars", "Manage your key-value pairs");
+
+varsScope
+  .task("set", "Set a new key-value pair")
+  .addPositionalParam("key", "The key associated with the value")
+  .addOptionalPositionalParam("value", "The value to store")
```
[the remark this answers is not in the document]
ChristopherDedominici: Yes, because if the value is not set, a prompt will appear, allowing the user to type it in

packages/hardhat-core/src/builtin-tasks/vars.ts
ChristopherDedominici: Because this task is being executed via js code when loading the config file, so it should not fail

packages/hardhat-core/src/internal/cli/vars.ts
ChristopherDedominici: I used "warn" for all the additional info to avoid polluting the stdout. So if these commands are used via script only the useful values (e.g.: the value associated to a key) will be shown

packages/hardhat-core/src/internal/cli/vars.ts
ChristopherDedominici: See explanation above

packages/hardhat-core/src/internal/cli/vars.ts
ChristopherDedominici: See explanation above

packages/hardhat-core/src/internal/cli/vars.ts
ChristopherDedominici: Loading all the global variables/functions/etc which are defined inside the config-env.ts file.
So you can access them inside your configuration file

packages/hardhat-core/src/internal/core/vars/vars-manager.ts
ChristopherDedominici: I was thinking the same when writing the code. It is a bit ugly but I think it's ok because I do not have to add a new specific error message just for this scenario + it will show the empty string so it will be quite clear that the variable is empty.
Would you do it differently?

packages/hardhat-core/src/internal/core/vars/vars-manager.ts
ChristopherDedominici: In several places in our code I found _format instead of _version.
@fvictorio which one is the correct one?

packages/hardhat-core/src/internal/core/vars/vars-manager.ts
ChristopherDedominici: Yeah good point. I thought I had to handle more complex scenarios but I can actually do it

packages/hardhat-core/src/internal/core/vars/vars-manager.ts
schaable: Yes, I think it's fine. We can add a specific error later if we need it 👍

packages/hardhat-core/src/internal/cli/vars.ts
schaable: Got it! Can we add a small comment above to clarify it?

packages/hardhat-core/src/internal/core/vars/vars-manager-setup.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    // defaultValue is defined
+    if (super.has(key)) {
+      this._getVarsWithDefaultValueAlreadySet.add(key);
+    } else {
+      this._getVarsWithDefaultValueToSet.add(key);
+    }
+
+    return value;
+  }
```
fvictorio, line 83: I found this function really hard to understand. I think an implementation like this would be clearer:

```typescript
  public get(key: string, defaultValue?: string): string {
    log(`function 'get' called with key '${key}'`);

    const varAlreadySet = super.has(key);

    if (varAlreadySet) {
      if (defaultValue !== undefined) {
        this._getVarsWithDefaultValueAlreadySet.add(key);
      } else {
        this._getVarsAlreadySet.add(key);
      }
    } else {
      if (defaultValue !== undefined) {
        this._getVarsWithDefaultValueToSet.add(key);
      } else {
        this._getVarsToSet.add(key);
      }
    }

    const value = super.get(key, defaultValue);

    // Do not return undefined to avoid throwing an error
    return value ?? "";
  }
```

packages/hardhat-core/src/internal/core/vars/vars-manager-setup.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  public getOptionalVarsToSet(): string[] {
+    return this._getOptionals(
+      this._getVarsToSet,
+      this._hasVarsToSet,
+      this._getVarsWithDefaultValueToSet
+    );
+  }
+
+  private _getRequired(getVars: Set<string>, hasVars: Set<string>): string[] {
```
fvictorio, line 109: Let's add some comments explaining this method and the next one. I understand the reasoning now, but I'm pretty sure I'll forget in one month.

packages/hardhat-core/src/internal/cli/vars.ts
```diff
@@ -0,0 +1,305 @@
...
+    case "list":
+      return list();
+    case "delete":
+      return del(taskArguments.var);
+    case "path":
+      return path();
+    case "setup":
+      return setup(configPath);
+    default:
+      throw new HardhatError(ERRORS.VARS.INVALID_TASK_DEFINITION_NAME, {
```
fvictorio, line 38: Sorry if my previous comment was too ambiguous. I just meant adding a `console.error` with an explanation. I don't think it's necessary to add a new error type for this.

fvictorio (approved):

PR #4449, opened 2023-10-06
Support Addressable in .withArgs and .equals

### LINKED ISSUE #4097: Use ethersjs v6 AddressLike type in hardhat-chai-matchers
### Describe the feature

ethersjs v6 implements a AddressLike type that improoves the UX by allowing to pass an signer or contract object where an address is expected

hardhat-chai-matchers still expects a string for an address so I suggest using the same treatment,

I thought this was implemented in the ethers.io project so I fist opened this issue: https://github.com/ethers-io/ethers.js/issues/4191

### Search terms

_No response_

Amxx:
Resolves #4097

---

When calling a contract fuction that takes an address argument, `ethers` support AddressLike object. This allows passing `Contract` or `Signer` object without resolving the address.

Before this PR, hardhat-chai-matchers did not support Addressable object (which are part of AddressLike). This meant that you needed to resolve the address youself, and then pass the address into the chai-matcher when you expect an event or a custom error.

If one of the args is an Addressable object, this PR will resolve its address internally. "Rewritting" the expected arguments should not be an issue because the arguments emitted by events or custom errors are simple type that should not contain a `getAddress()` function. Therefore, I don't expect any usecase where the user would pass an `Addressable` for anything but its address.

Support Addressable objects in `expect(...).to.be.equal(...)` tests is more complexe, because the resolution of Addressable object is async. The proposed implementations looks for immediatly available addresses in common locations (for signers and contracts).


ernestognw (commented):
This is looking extremely good!

packages/hardhat-chai-matchers/src/internal/addressable.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// ethers's Addressable have a .getAddress() that returns a Promise<string>. We don't want to deal with async here,
+// so we are looking for a sync way of getting the address. If an address was recovered, it is returned as a string,
+// otherwise undefined is returned.
+function tryGetAddress(value: any): string | undefined {
+  if (isAddress(value)) {
+    return value;
+  } else if (isAddressable(value)) {
+    return tryGetAddress((value as any).address ?? (value as any).target);
```
ernestognw, line 32: Just realized [`.target` can also be an ENS name](https://docs.ethers.org/v6/api/contract/#BaseContract-target), we should test also with those cases

Amxx: If target is an ENS name, doing `tryGetAddress(value.target)` will return `undefined` (because `isAddress()` returns false on ENS).

AFAIK, this is the best we can do without an async resolution.

packages/hardhat-chai-matchers/src/internal/addressable.ts
```diff
@@ -0,0 +1,71 @@
...
+  chaiUtils: Chai.ChaiUtils
+) {
+  return (_super: (...args: any[]) => any) =>
+    overwriteAddressableFunction(method, name, negativeName, _super, chaiUtils);
+}
+
+// ethers's Addressable have a .getAddress() that returns a Promise<string>. We don't want to deal with async here,
+// so we are looking for a sync way of getting the address. If an address was recovered, it is returned as a string,
+// otherwise undefined is returned.
+function tryGetAddress(value: any): string | undefined {
```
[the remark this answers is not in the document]
ernestognw: Sure, let's keep `tryGetAddressSync` but comment about the return

packages/hardhat-chai-matchers/src/internal/addressable.ts
ernestognw: Just double checked and you're right. Agree this seems to be the best we can do

Amxx:
@fvictorio is it possible to get that review, and possibly shipped? We have had no news in 2 months :/

fvictorio:
@Amxx yes, I'll try to review this today/next week. Sorry for the delay!

fvictorio (approved):

fvictorio:
This looks great, thanks a lot @Amxx and sorry again for making you wait so much! This will be included in the today's release.

PR #4461, opened 2023-10-10
feat: solidity coverage support in EDR mode

agostbiro:
Add [`solidity-coverage`](https://github.com/sc-forks/solidity-coverage/tree/master) support for Hardhat network in EDR mode. Resolves https://github.com/NomicFoundation/edr/issues/44

I had to make [changes](https://github.com/sc-forks/solidity-coverage/pull/817) in `solidity-coverage` as the property layout of the EthereumJS VM changed after the dual mode changes from

```
node._vm.evm.events
```

to 

```
node._context._vm._vm.evm.events
```

I tested the changes in this PR by running the `solidity-coverage` test suite in [my fork](https://github.com/agostbiro/solidity-coverage/tree/agostbiro/hardhat-dual-vm-mode-compat) against a local version of `hardhat` with `HARDHAT_EXPERIMENTAL_VM_MODE=rethnet` and `HARDHAT_EXPERIMENTAL_VM_MODE=ethereumjs`.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/context/rethnet.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/context/rethnet.ts
@@ -1,6 +1,5 @@
 import { Blockchain, RethnetContext, SpecId } from "@ignored/edr";
 import { AsyncEventEmitter } from "@nomicfoundation/ethereumjs-util";
-import { EVMEvents } from "@nomicfoundation/ethereumjs-evm/src/types";
 
 import { BlockchainAdapter } from "../blockchain";
 import { RethnetBlockchain } from "../blockchain/rethnet";
@@ -9,7 +8,7 @@ import { MemPoolAdapter } from "../mem-pool";
 import { BlockMinerAdapter } from "../miner";
 import { VMAdapter } from "../vm/vm-adapter";
 import { RethnetMiner } from "../miner/rethnet";
-import { RethnetAdapter } from "../vm/rethnet";
+import { RethnetAdapter, VMStub } from "../vm/rethnet";
 import { NodeConfig, isForkedNodeConfig } from "../node-types";
 import {
   ethereumjsHeaderDataToRethnetBlockOptions,
@@ -136,9 +135,9 @@ export class RethnetEthContext implements EthContextAdapter {
     const limitContractCodeSize =
       config.allowUnlimitedContractSize === true ? 2n ** 64n - 1n : undefined;
 
-    const vmStub = {
+    const vmStub: VMStub = {
       evm: {
-        events: new AsyncEventEmitter<EVMEvents>(),
+        events: new AsyncEventEmitter(),
       },
     };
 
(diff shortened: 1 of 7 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/context/rethnet.ts
```diff
@@ -133,11 +136,18 @@ export class RethnetEthContext implements EthContextAdapter {
     const limitContractCodeSize =
       config.allowUnlimitedContractSize === true ? 2n ** 64n - 1n : undefined;
 
+    const vmStub = {
```
agostbiro, line 139: This has to be injected to both the Rethnet VM adapter and the Rethnet mempool, as some `solidty-coverage` functionality mines blocks.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/rethnet.ts
```diff
@@ -51,7 +53,9 @@ export class RethnetAdapter implements VMAdapter {
     private _blockchain: Blockchain,
     private _state: RethnetStateManager,
     private readonly _common: Common,
-    private readonly _limitContractCodeSize: bigint | undefined
+    private readonly _limitContractCodeSize: bigint | undefined,
+    // For solidity-coverage compatibility. Name cannot be changed.
+    private _vm: VMStub
```
agostbiro, line 58: Property layout matches the EthereumJS VM adapter.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/rethnet.ts
```diff
@@ -140,6 +155,18 @@ export class RethnetAdapter implements VMAdapter {
       }
     }
 
+    // For solidity-coverage compatibility
+    for (const step of this._vmTracer.tracingSteps) {
+      console.log("emitting", step);
```
Wodann, line 160:
```suggestion
```

packages/hardhat-core/src/internal/hardhat-network/provider/vm/rethnet.ts
```diff
@@ -520,3 +558,12 @@ export class RethnetAdapter implements VMAdapter {
     return undefined;
   }
 }
+
+// For compatibility with solidity-coverage that attaches a listener to the step event.
```
Wodann, line 562: I wonder whether the VMAdapter interface can extend this interface to avoid duplication? That way we can add all of the logic for supporting solidity coverage in one place and call that from the rethnet and ethereumjs implementations

Wodann (changes requested):

packages/hardhat-core/src/internal/hardhat-network/provider/vm/rethnet.ts
fvictorio: I also thought something similar: abstracting this at the context level somehow (e.g., a `context.onStep`) function. That also has the advantage of making the need for this functionality more explicit.

But on the other hand, I like that the current approach doesn't need any changes to the ethereumjs adapter, so I slightly prefer to just do it this way.

fvictorio (approved):

Wodann (approved):

PR #4463, opened 2023-10-10
feat: state overrides

Wodann:


crates/rethnet_evm/src/state/overrides.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// Account code override.
+    pub code: Option<Bytecode>,
+    /// Account storage override.
+    pub storage: Option<StorageOverride>,
+}
+
+impl AccountOverride {
+    /// Overrides the provided original account information.
+    pub fn override_info(&self, original: Option<AccountInfo>) -> Option<AccountInfo> {
+        let has_override = self.balance.is_some() | self.nonce.is_some() | self.code.is_some();
```
agostbiro, line 32: Is `self.storage.is_some()` missing from here?

crates/rethnet_evm/src/state/overrides.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub nonce: Option<u64>,
+    /// Account code override.
+    pub code: Option<Bytecode>,
+    /// Account storage override.
+    pub storage: Option<StorageOverride>,
+}
+
+impl AccountOverride {
+    /// Overrides the provided original account information.
+    pub fn override_info(&self, original: Option<AccountInfo>) -> Option<AccountInfo> {
```
agostbiro, line 31: I would try to change the implementation of this method to make it a compile error if a new property is added to `AccountInfo` to prevent potential future bugs. One option would be to construct the `AccountInfo` explicitly in the return or to use destructuring to access the fields of the `AccountInfo` earlier.

crates/rethnet_evm/src/state/overrides.rs
```diff
@@ -0,0 +1,178 @@
...
+#[derive(Clone, Debug, Default)]
+pub struct StateOverrides {
+    account_overrides: HashMap<Address, AccountOverride>,
+    code_by_hash_overrides: HashMap<B256, Bytecode>,
+}
+
+impl StateOverrides {
+    /// Constructs a new set of state overrides.
+    pub fn new(mut account_overrides: HashMap<Address, AccountOverride>) -> Self {
+        let code_by_hash_overrides = account_overrides
```
agostbiro, line 72: I'm wondering if it'd make sense compute this lazily.

crates/rethnet_evm/src/state/overrides.rs
```diff
@@ -0,0 +1,178 @@
...
+            .values_mut()
+            .filter_map(|account_override| {
+                if let Some(code) = &mut account_override.code {
+                    let code_hash = code.hash_slow();
+
+                    Some((code_hash, code.clone()))
```
agostbiro, line 78: We could potentially store a ref to the code to avoid [cloning it twice.](https://github.com/NomicFoundation/hardhat/pull/4463/files#diff-1e8df9b3c72556c829cc01444e78dae3bb39c57699dd0528f9e3e6bd5c474cbaR140)

agostbiro (changes requested):

packages/hardhat-core/test/internal/hardhat-network/provider/modules/eth/methods/call.ts
```diff
@@ -857,7 +857,8 @@ describe("Eth module", function () {
                   "latest",
                   {
                     [deployerAddress]: {
-                      nonce: numberToRpcQuantity(maxNonce),
+                      // The maximum value that can be stored in a u64 is not allowed for a nonce
+                      nonce: numberToRpcQuantity(maxNonce - 1n),
```
fvictorio, line 861: Good catch. I double-checked this in geth too.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2134,8 +2138,20 @@ export class HardhatNode extends EventEmitter {
...
+    stateOverrideSet: StateOverrideSet = {}
   ): Promise<bigint> {
+    const overridenAccount = Object.entries(stateOverrideSet).find(([key]) =>
+      Address.fromString(key).equals(address)
+    )?.[1];
+
+    if (
+      overridenAccount !== undefined &&
+      overridenAccount.nonce !== undefined
+    ) {
```
fvictorio, line 2151:
```suggestion
    if (overridenAccount?.nonce !== undefined) {
```

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -491,7 +491,11 @@ export class HardhatNode extends EventEmitter {
     blockNumberOrPending: bigint | "pending",
     stateOverrideSet: StateOverrideSet = {}
   ): Promise<RunCallResult> {
-    const tx = await this._getTransactionForCall(call, blockNumberOrPending);
+    const tx = await this._getTransactionForCall(
+      call,
+      blockNumberOrPending,
+      stateOverrideSet
+    );
```
fvictorio, line 498: I'm curious why you added this, because I'm not really sure it's necessary. The description of state overrides is a bit ambiguous about this.

fvictorio: To expand on this:

I don't know if `eth_call({from: a, nonce: X})` is the same as `eth_call({from: a}, {[a]: { nonce: X }})`.

My understanding here is that state overrides mean "make these changes before the execution", so in most scenarios these would be equivalent. The only scenario where _maybe_ there's a difference is during deployments. But observing differences in `eth_call`s that deploy something is really hard. And to make everything worse, geth seems to completely ignore the nonces for `eth_call` "deployments". Which might be a bug or not, I don't know.

The bottom line for me here is: if we are not really sure that this is necessary, I would drop it. This is not a very specified issue, so if we have to guess, let's guess whatever results in simpler code.

crates/rethnet_evm/src/state/overrides.rs
Wodann: I don't know how we could, since we need the hash as a key, but I'm open to suggestions.

The old implementation of `ByteCode` used to cache the hash. I'd like to return to that, but it requires changes to `revm` which are out of scope for this PR.

crates/rethnet_evm/src/state/overrides.rs
Wodann: Under the hood, the `Bytes` class will already do something like that, whenever possible.

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
Wodann: AFAIK, the nonce is extracted from the state. When we are running the transaction in `revm` it validates that the nonce of the transaction is not less than the nonce of the state, throwing an error.

Thus, if we don't use the overriden nonce when determining the transaction nonce, the transaction is invalid, I think.

crates/rethnet_evm/src/state/overrides.rs
agostbiro: I was thinking we could construct `code_by_hash_overrides` the first time `code_by_hash` is called, but it'd only make sense if `code_by_hash` isn't called every time a `StateOverrides` created and I'm not sure about the access patterns. Feel free ignore in any case, this is easy to optimize later if it turns out to be a problem.

crates/rethnet_evm/src/state/overrides.rs
Wodann: Good idea. Thanks!

Addressed by https://github.com/NomicFoundation/hardhat/pull/4463/commits/f5165bdcd8f9085c877aa74c45aa045b0e995826

agostbiro (approved):
Thanks for the change! LGTM

packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
fvictorio: I see, in that case it makes sense to do it :+1:

fvictorio (approved):

PR #4470, opened 2023-10-12
docs: update book to switch from rethnet to edr mode

agostbiro:
@Wodann I'm wondering if we need to update anywhere else?


Wodann (approved):
This is all I recall. If there is more, we should find it in the renaming task

PR #4481, opened 2023-10-16
refactor: replace vm stub in adapters with vm proxy in node

fvictorio:
WIP. I implemented the step handler for both ethereumjs and edr modes, and verified that it works. I also implemented the `_vm.stateManager` part of this, which smock needs.

Pending here is the `beforeMessage`/`afterMessage` handlers, for both ethereumjs and edr modes.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/miner/edr.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/miner/edr.ts
@@ -1,3 +1,5 @@
+import type { MinimalInterpreterStep } from "../vm/proxy-vm";
+
 import { Common } from "@nomicfoundation/ethereumjs-common";
 import { Address } from "@nomicfoundation/ethereumjs-util";
 import { ConfigOptions, MineOrdering, SpecId, mineBlock } from "@ignored/edr";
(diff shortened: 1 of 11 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/miner/edr.ts
```diff
@@ -77,13 +81,15 @@ export class EdrMiner implements BlockMinerAdapter {
...
 
-          // For solidity-coverage compatibility
-          this._vmStub.evm.events.emit("step", {
-            pc: Number(traceItem.pc),
-            depth: traceItem.depth,
-            opcode: { name: traceItem.opcode },
-            stackTop: traceItem.stackTop,
-          });
+          for (const listener of this._stepListeners) {
+            await listener({
```
agostbiro, line 85: Can the listeners throw? I'm also wondering if we should add some warning for slow listeners or alternatively call them outside of the call stack with `setTimeout(listener, 0)` to avoid blocking here.

agostbiro:
Thanks for this PR @fvictorio, looks good overall with one question + in case you haven't seen it, there is [one lint failing.](https://github.com/NomicFoundation/hardhat/actions/runs/6536296334/job/17747679560?pr=4481)

packages/hardhat-core/src/internal/hardhat-network/provider/miner/edr.ts
fvictorio: Hmm, yes, in theory they can throw. But I think we can ignore that for now; I don't think the behavior is fundamentally different to what was happening before (and this should be temporary anyway).

fvictorio:
Fixed!

Wodann (approved):

PR #4499, opened 2023-10-19
fix: static state change error

Wodann:


packages/hardhat-core/src/internal/hardhat-network/provider/vm/exit.ts
```diff
@@ -47,7 +48,7 @@ export class Exit {
       default: {
         // TODO temporary, should be removed in production
```
agostbiro, line 49: Should we remove the TODO?

agostbiro (approved):
Thanks @Wodann ! Added one comment, otherwise I'd defer to @fvictorio on the review of this.

agostbiro:
I tried to resolve the merge conflict, but it's related to https://github.com/NomicFoundation/hardhat/pull/4487 and it's not obvious what to do about it, so I'd leave it to you @Wodann

packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-tracer.ts
```diff
@@ -206,10 +198,15 @@ export class VMTracer {
           ).address;
         }
       } else if (isHaltResult(executionResult)) {
-        trace.exit = Exit.fromEdrExceptionalHalt(executionResult.reason);
+        trace.exit =
+          haltOverride === undefined
+            ? Exit.fromEdrExceptionalHalt(executionResult.reason)
+            : haltOverride;
```
fvictorio, line 204:
```suggestion
          haltOverride ?? Exit.fromEdrExceptionalHalt(executionResult.reason);
```

packages/hardhat-core/src/internal/hardhat-network/provider/vm/ethereumjs.ts
```diff
@@ -764,48 +765,61 @@ export class EthereumJSAdapter implements VMAdapter {
...
-          gasUsed,
-          output: result.execResult.returnValue,
+          result: {
+            gasUsed,
+            output: result.execResult.returnValue,
+          },
         };
       } else {
-        const vmError = Exit.fromEthereumJSEvmError(
+        // Override the exceptional halt reason with the one from the EVM result, to avoid lossy casting
```
fvictorio, line 799: I'm not sure I understand this. The idea is to avoid converting it twice? Is that the lossy part?

packages/hardhat-core/src/internal/hardhat-network/provider/vm/exit.ts
Wodann: Addressed by https://github.com/NomicFoundation/hardhat/pull/4499/commits/f983666a5336bc78cef9dbd70a685d0410b2e81d

PR #4512, opened 2023-10-20
fix: add missing type defs to EDR NPM package

agostbiro:


.github/workflows/edr-npm-release.yml
```diff
@@ -10,6 +10,7 @@ permissions:
   push:
     branches:
       - main
+      - edr/ci/fix-missing-type-def-in-edr-package
```
Wodann, line 13:
```suggestion
      - edr/main
```

agostbiro: Fixed in [bf4e7f2](https://github.com/NomicFoundation/hardhat/pull/4512/commits/bf4e7f20c6691262160e41e78ab064072ed753c8)

Wodann (approved):
LGTM, but there are two things to consider before merging:

- Could you please also upgrade the hardhat-core/package.json  to reflect the EDR version?
- We will likely want to be able to publish from `edr/main`

agostbiro:
We decided to keep the carret version in Hardhat core.

PR #4513, opened 2023-10-21
Add basic support for verification on Sourcify

schaable:
Some examples:

sourcify-1
sourcify-2
sourcify-3


packages/hardhat-verify/src/internal/tasks/sourcify.ts
```diff
@@ -0,0 +1,223 @@
...
+        verificationInterface: sourcify,
+        contractInformation,
+      }
+    );
+
+    if (verificationSuccess) {
+      return;
+    }
+
+    throw new ContractVerificationFailedError(verificationMessage, []);
```
fvictorio, line 126: Should we pass `contractInformation.undetectableLibraries` here?

fvictorio (approved):

PR #4522, opened 2023-10-23
Use relative paths in the recommended .gitignore

alcuadrado:
The reason I'm sending this PR is that ignition creates an `artifacts` folder that you may want to commit, and our `.gitignore` was making that tricky.

---

### DIFF
--- a/packages/hardhat-core/recommended-gitignore.txt
+++ b/packages/hardhat-core/recommended-gitignore.txt
@@ -1,11 +1,14 @@
 node_modules
 .env
-coverage
-coverage.json
-typechain
-typechain-types
 
 # Hardhat files
-cache
-artifacts
+/cache
+/artifacts
 
+# TypeChain files
+/typechain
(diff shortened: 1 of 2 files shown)


packages/hardhat-core/recommended-gitignore.txt
```diff
@@ -1,11 +1,14 @@
...
 .env
-coverage
-coverage.json
-typechain
-typechain-types
 
 # Hardhat files
-cache
-artifacts
+/cache
```
schaable, line 5: this would only ignore the root `cache` and `artifacts` folders, right? what about the ones inside `test/fixture-projects`?

fvictorio: This PR is about changing the `.gitignore` generated by Hardhat's project initialization, not our own `.gitignore` (which is where fixture projects are relevant)

schaable (approved):

PR #4529, opened 2023-10-25
Normalize input addresses to lowercase for comparison

schaable:
Normalized user-input addresses from CLI and libraries file for correct comparison.


ChristopherDedominici (approved):
All good, just a minor: is there a test to cover this scenario?

fvictorio (approved):

fvictorio:
I'm ok with merging this without a test, since doing a proper test for this is going to be super involved, and these are very low-risk changes.

PR #4535, opened 2023-10-26
feat: implement blockchain- and transaction-related methods in the Node

alcuadrado:
Fixes https://github.com/NomicFoundation/edr/issues/151
Fixes https://github.com/NomicFoundation/edr/issues/152

### DIFF
--- a/crates/edr_rpc_server/src/node.rs
+++ b/crates/edr_rpc_server/src/node.rs
@@ -4,6 +4,7 @@ mod node_error;
 use std::{mem, sync::Arc};
 
 use edr_eth::{
+    receipt::BlockReceipt,
     remote::{
         filter::{FilteredEvents, LogOutput},
         BlockSpec, BlockTag,
@@ -525,6 +526,15 @@ impl Node {
 
         Ok(transaction)
     }
+
+    pub async fn transaction_receipt(
+        &self,
+        hash: &B256,
+    ) -> Result<Option<Arc<BlockReceipt>>, BlockchainError> {
+        let node_data = self.lock_data().await;
+
+        node_data.blockchain.receipt_by_transaction_hash(hash).await
+    }
 }
 
 /// An account in this node.
@@ -1021,6 +1031,21 @@ mod tests {
             .unwrap();
 
         assert_eq!(non_existing_tx, None);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn transaction_receipt() -> Result<()> {
+        let fixture = NodeTestFixture::new().await?;
+
+        let non_existing_receipt = fixture
+            .node
+            .transaction_receipt(&B256::zero())
+            .await
+            .unwrap();
+
+        assert_eq!(non_existing_receipt, None);
 
         Ok(())
     }
--- a/crates/edr_rpc_server/src/node.rs
+++ b/crates/edr_rpc_server/src/node.rs
@@ -390,6 +390,11 @@ impl Node {
 
         node_data.impersonated_accounts.remove(&address)
     }
+
+    pub async fn last_block_number(&self) -> u64 {
+        let node_data = self.lock_data().await;
+        node_data.blockchain.last_block_number().await
+    }
 }
 
 /// An account in this node.
@@ -434,6 +439,20 @@ mod tests {
 
         let chain_id = fixture.node.chain_id().await;
         assert_eq!(chain_id, U64::from(fixture.config.chain_id));
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn last_block_number() -> Result<()> {
+        let fixture = NodeTestFixture::new().await?;
+
+        let last_block_number = fixture.node.last_block_number().await;
+        assert_eq!(last_block_number, 0);
+
+        fixture.node.lock_data().await.mine_block(None).await?;
+        let last_block_number = fixture.node.last_block_number().await;
+        assert_eq!(last_block_number, 1);
 
         Ok(())
     }
--- a/crates/edr_rpc_server/src/node.rs
+++ b/crates/edr_rpc_server/src/node.rs
@@ -25,6 +25,12 @@ use tokio::sync::Mutex;
 pub use self::node_error::NodeError;
 use crate::{filter::Filter, node::node_data::NodeData, Config};
 
+#[derive(Debug)]
+pub struct BlockAndTotalDifficulty {
+    pub block: Arc<dyn SyncBlock<Error = BlockchainError>>,
+    pub total_difficulty: U256,
+}
+
 pub struct Node {
     data: Mutex<NodeData>,
 }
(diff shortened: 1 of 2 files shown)


crates/edr_rpc_server/src/node.rs
```diff
@@ -390,6 +398,169 @@ impl Node {
...
+        let total_difficulty = node_data
+            .blockchain
+            .total_difficulty_by_hash(block.hash())
+            .await?
+            .expect("The block exists, so it must have a total difficulty");
+
+        self.workaround_block_by_spec(&mut *node_data.blockchain, block_spec)
+            .await?;
+
+        Ok(BlockAndTotalDifficulty {
```
alcuadrado, line 419: Is this notation idiomatic? I mean `Ok(LiteralConstructor { ... })`, with that level of indentation.

crates/edr_rpc_server/src/node.rs
```diff
@@ -1,28 +1,36 @@
...
-    AccountInfo, Block, Bytecode, MineBlockResult, KECCAK_EMPTY,
+    AccountInfo, Block, Bytecode, MineBlockResult, SyncBlock, KECCAK_EMPTY,
 };
 use k256::SecretKey;
 use tokio::sync::Mutex;
 
 pub use self::node_error::NodeError;
 use crate::{filter::Filter, node::node_data::NodeData, Config};
 
+#[derive(Debug)]
```
alcuadrado, line 28: Should I do anything extra apart from deriving debug in a struct that's as simple as this one? It's only used to return two values from a single method.

crates/edr_rpc_server/src/node.rs
```diff
@@ -390,6 +398,169 @@ impl Node {
...
+    ) -> Result<BlockAndTotalDifficulty, NodeError> {
+        let mut node_data = self.lock_data().await;
+
+        let block = node_data.block_by_block_spec(block_spec).await?;
+
+        let total_difficulty = node_data
+            .blockchain
+            .total_difficulty_by_hash(block.hash())
+            .await?
+            .expect("The block exists, so it must have a total difficulty");
```
alcuadrado, line 414: Is this ok? I know this invariant should hold, but should I use `expect` like this?

crates/edr_rpc_server/src/node.rs
```diff
@@ -390,6 +398,169 @@ impl Node {
...
+
+    pub async fn block_and_total_difficulty_by_hash(
+        &self,
+        block_hash: &B256,
+    ) -> Result<Option<BlockAndTotalDifficulty>, BlockchainError> {
+        let node_data = self.lock_data().await;
+
+        let block = node_data.blockchain.block_by_hash(block_hash).await?;
+
+        match block {
```
alcuadrado, line 433: Is there a better pattern than doing this? Maybe return if the block is None instead? This is pretty ugly.

I actually tried `block.map(...)`, but I need to call an async function.

crates/edr_rpc_server/src/node.rs
```diff
@@ -437,4 +618,548 @@ mod tests {
...
+        fixture.node.lock_data().await.mine_block(None).await?;
+        fixture.node.lock_data().await.mine_block(None).await?;
+
+        let block_and_total_difficulty0 = fixture
+            .node
+            .block_and_total_difficulty_by_block_spec(&BlockSpec::Number(0))
+            .await
+            .unwrap();
+
+        let block1_hash = *fixture
```
alcuadrado, line 869: Is using `*` to generate a `copy` common?

alcuadrado:
@Wodann can you do an initial review of this? I left some comments for you.

The JSON-RPC layer changes are larger than expected (even larger than we thought they were after our call), so I think it makes sense to get this reviewed first.

crates/edr_rpc_server/src/node.rs
Wodann: Just add a documentation comment above the derive `///`

crates/edr_rpc_server/src/node.rs
Wodann: This is correct, yes. `expect(..)` is more descriptive than `.unwrap()`

crates/edr_rpc_server/src/node.rs
Wodann: Yes 👍

crates/edr_rpc_server/src/node.rs
Wodann: I would use `if let Some(block) = block` as it's shorter

crates/edr_rpc_server/src/node.rs
```diff
@@ -437,4 +618,548 @@ mod tests {
...
+            block_tag: BlockTag,
+            expected_block_number: u64,
+        ) {
+            let block = fixture
+                .node
+                .block_and_total_difficulty_by_block_spec(&BlockSpec::Tag(block_tag))
+                .await;
+
+            assert_eq!(block.unwrap().block.header().number, expected_block_number);
+        }
```
Wodann, line 649: Since multiple functions use this, it'd be better to avoid code duplication by moving it one layer up (inside the `mod tests`)

crates/edr_rpc_server/src/node.rs
Wodann: Yes, when dereferencing. Clippy will complain if you use clone, IIRC

crates/edr_rpc_server/src/node.rs
```diff
@@ -437,4 +618,548 @@ mod tests {
...
+            .await
+            .unwrap();
+
+        assert!(non_existing_block.is_none());
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn block_and_total_difficulty_by_hash_total_difficulty() -> Result<()> {
```
Wodann, line 857: This feels excessive as the unit tests for the blockchain implementations already test this.

Is there a more minimal way of testing the code for this function?

crates/edr_rpc_server/src/node.rs
```diff
@@ -437,4 +618,548 @@ mod tests {
...
+        let block_and_total_difficulty = fixture
+            .node
+            .block_and_total_difficulty_by_block_spec(&BlockSpec::Tag(BlockTag::Earliest))
+            .await
+            .unwrap();
+
+        assert_block_transaction_count(&fixture, BlockSpec::Tag(BlockTag::Earliest), 0).await;
+        assert_block_transaction_count(&fixture, BlockSpec::Tag(BlockTag::Latest), 0).await;
+        assert_block_transaction_count(&fixture, BlockSpec::Tag(BlockTag::Pending), 0).await;
+        assert_block_transaction_count(&fixture, BlockSpec::Number(0), 0).await;
```
Wodann, line 958: I would create a unit test for the `block_by_block_spec` function that validates it works. That avoids having to create tests for all variants for every API that accepts a `BlockSpec`

crates/edr_rpc_server/src/node.rs
```diff
@@ -437,4 +618,548 @@ mod tests {
...
+            .transaction_by_block_hash_and_index(
+                &block_and_total_difficulty.block.header().hash(),
+                0,
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(tx, None);
+
+        let non_existing_block_tx = fixture
```
Wodann, line 1026: This seems like a separate test. I'd make two tests, named to indicate what they're testing. When a test fails, it makes it easier to understand which feature/functionality is broken. It also allows for parallel execution

crates/edr_rpc_server/src/node.rs
```diff
@@ -437,4 +618,548 @@ mod tests {
...
+
+        let tx = fixture
+            .node
+            .transaction_by_block_spec_and_index(&BlockSpec::Tag(BlockTag::Latest), 0)
+            .await
+            .unwrap();
+
+        assert_eq!(tx, None);
+
+        let non_existing_block_tx = fixture
```
Wodann, line 1049: I recommend splitting it into separate tests for the same reasons as above

crates/edr_rpc_server/src/node.rs
```diff
@@ -437,4 +618,548 @@ mod tests {
...
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn get_transaction_count() -> Result<()> {
+        let fixture = NodeTestFixture::new().await?;
+
+        let count = fixture
+            .node
+            .get_transaction_count(Address::zero(), Some(&BlockSpec::Tag(BlockTag::Earliest)))
```
Wodann, line 1087: No need to test all variants if we have unit tests for `block_by_block_spec`

Wodann (changes requested):

crates/edr_rpc_server/src/node.rs
alcuadrado: Looks nicer! Thanks

crates/edr_rpc_server/src/node.rs
alcuadrado: They are all slightly different.

PR #4541, opened 2023-10-29
Introduce a rustfmt.toml file with nightly features

alcuadrado:
This PR introduces a `rustfmt.toml` that uses some nightly features and makes the necessary setup changes for it to work.

The first 5 commits implement the necessary setup to use nightly exclusively for rustfmt, while still compiling with stable. This includes the devcontainer (you need to rebuild it), the ci, pre-commit hook, and vscode settings.

Then, each of the following commits enable a rustfmt feature, and automatically reformat the code.

The features enabled are:

1. [`import_granularity`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#imports_granularity): Controls how the import statements are merged. In this case, it automatically merges every import from the same crate, so that there's only 1 import per crate.
2. [`wrap_comments`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#wrap_comments): Automatically wrap comments.
3. [`group_imports`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#group_imports): Groups imports in three groups (i.e. `std`, `core`, and `alloc`; external crates; and `self`, `super`, and `crate`).

Features 1 and 2 were discussed with @Wodann. Feature 3 looked interesting to me (we use an equivalent in TS), but I have no strong opinions about it.

Other interesting features:

* [`comment_width`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#comment_width): So that there's less wrapping.
* [`use_small_heuristics`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#use_small_heuristics): This controls how often `rustfmt` introduces newlines. I saw multiple projects set it to `"Max"`, so that lines are longer. For example, `rust-analyzer` uses it.


alcuadrado:
> * [`use_small_heuristics`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#use_small_heuristics): This controls how often `rustfmt` introduces newlines. I saw multiple projects set it to `"Max"`, so that lines are longer. For example, `rust-analyzer` uses it.

I recommend you enable it and look at some diffs. I don't particularly like it, but it may be my lack of experience.

alcuadrado:
> * [`comment_width`](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#comment_width): So that there's less wrapping.

I played with this for a bit, and setting it to `100` would preserve most comments unchanged, so I believe that was the intended width. 

If we decide to set it, I can do it and rebase this PR.

Wodann (approved):

Wodann:
Let's not change the `use_small_heuristics` for the reasons mentioned above.

For, `comment_width`, the width of comments in the PR matches the max length of code lines, which looks good to me.

alcuadrado:
Rebased against `edr/main`. I'll merge once the CI passes.

PR #4544, opened 2023-10-30
build: bump edr versions

Wodann:


agostbiro (commented):
I'm wondering if we should bump the NAPI JS package as well @Wodann ?

agostbiro (approved):
LGTM, thanks!

PR #4545, opened 2023-10-30
add plugin hardhat-contract-signatures

AxelLoupias:
add plugin "hardhat-contract-signatures"


fvictorio:
Hi @AxelLoupias, thanks for sending this. I skimmed your plugin and I have some tips in case you are interested:

- Your plugin relies on `@nomicfoundation/hardhat-ethers` but doesn't have it listed as a peer dependency; it probably should.
- Your readme says "The tasks throw error if the SC are not compiled.". You can run `hre.run("compile")` at the beginning of each task definition to make sure that the project is compiled first.
- The `signature` scope has "test" as its definition, which is shown when you run `npx hardhat`

fvictorio (approved):

AxelLoupias:
Thank you very much for your comments. I have already applied the tips for the new version.

PR #4548, opened 2023-10-30
feat: EDR RPC send transaction 

agostbiro:
Draft PR to add `eth_sendTransaction` and `eth_sendRawTransaction` support to the RPC crate.

I also changed the argument type in `MethodInvocation::SendTransaction` from `TransactionInput` to `EthTransactionRequest` as only the latter supports conversion into a typed transaction, which is then necessary for signing. 

TODOs left: 

- ~[ ] support for fake signatures for impersonated accounts.~
- [x] some open questions (see comments please)


crates/edr_eth/src/transaction.rs
```diff
@@ -19,6 +19,7 @@ pub use self::{detailed::DetailedTransaction, kind::TransactionKind, request::*,
 #[cfg_attr(feature = "serde", serde(deny_unknown_fields))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct EthTransactionRequest {
+    // TODO review question: should this be non-optional? In the `eth_sendTransaction` docs and in the Hardhat `rpcTransactionRequest` type, `from` is marked as compulsory.
```
agostbiro, line 22: Review question

crates/edr_rpc_server/src/node.rs
```diff
@@ -258,6 +246,28 @@ impl Node {
...
+        &self,
+        transaction_request: EthTransactionRequest,
+    ) -> Result<B256, NodeError> {
+        let mut node_data = self.lock_data().await;
+
+        let signed_transaction = node_data.get_signed_transaction(transaction_request)?;
+
+        // TODO discuss in review: this can block the async executor if the state is remote or
+        // forked. Do we want to do something about in this PR or should we add a follow up issue?
+        node_data.add_pending_transaction(signed_transaction)
```
agostbiro, line 259: Review question

crates/edr_eth/src/transaction.rs
Wodann: I think you are right. Feel free to remove the option

crates/edr_rpc_server/src/node.rs
```diff
@@ -13,15 +13,16 @@ use edr_eth::{
...
 use edr_evm::{
     blockchain::BlockchainError,
     state::{AccountModifierFn, StateError},
     AccountInfo, Block, Bytecode, MineBlockResult, KECCAK_EMPTY,
 };
 
-use crate::{filter::Filter, node::node_data::NodeData, Config};
+use crate::{node::node_data::NodeData, Config};
 
 pub use self::node_error::NodeError;
```
Wodann, line 28: As the node is now a wrapper around `Mutex<NodeData>`, can't we remove one layer here? A lot of the functions seem like boilerplate and doing this would resolve it

crates/edr_rpc_server/src/node.rs
Wodann: Feel free to create an issue for this. I think there are more places than just here, where we could employ the approach of [`spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) to prevent stalling.

Wodann (commented):
Overall LGTM. Just a question

crates/edr_rpc_server/src/node.rs
agostbiro: Yeah making the state `Arc<Mutex<NodeData>>` instead of `Arc<Node>` for the server would work, but I think we still have to validate if a single `Mutex` is ok performance-wise, so I'd rather keep all the locking related logic in one place, abstracted behind `Node`, in case we have to refactor later.

The pattern I'm trying to follow with what to put into `NodeData` is this: if the method is a helper method, then put it in `NodeData` since it'd try to acquire the lock twice resulting in a deadlock if it was in `Node` . But then there will be some methods like `new_pending_transaction_filter` in `NodeData` that are not helper methods, but are needed to test stuff in `NodeData`, so yeah that brings boilerplate.

crates/edr_rpc_server/src/node.rs
agostbiro: Thanks, in the meantime I looked into `tokio::task::block_in_place` once more and learned that it actually turns a worker thread into a blocking thread if it's called from a worker thread which is pretty cool and makes my concerns moot. Sorry for the false alarm!

crates/edr_eth/src/transaction.rs
agostbiro: Done in [Make from in EthTransactionRequest mandatory](https://github.com/NomicFoundation/hardhat/pull/4548/commits/818542e5a42893c2f6235218cc58de79897246e1)

agostbiro:
This is now ready for final review and merging. I created a follow up task for impersonated accounts: https://github.com/NomicFoundation/edr/issues/222

Wodann (approved):

Issue #4490, opened 2023-10-17
Bug in compilation when multiple versions are used with many contracts


fvictorio:
See original discussion here: https://github.com/NomicFoundation/hardhat/issues/3722

Reproduction:

1. Clone the [chainlink repo](https://github.com/smartcontractkit/chainlink) (I used commit `f74c15a717`)
2. `cd contracts && pnpm i`
3. Remove the TypeChain plugin from the config because it messes with `hh clean --global`
4. Run `while true; do hh clean && hh clean --global && hh compile || break; done`. This will remove the downloaded compilers and re-compile the project until it fails.

This seems to be fixed by https://github.com/NomicFoundation/hardhat/pull/3818

PR #4550, opened 2023-10-31
Bug in compilation when multiple versions are used with many contracts 

ChristopherDedominici:
See issue: https://github.com/NomicFoundation/hardhat/issues/4490

PROBLEM
If multiple requests to download the same version of a compiler are sent, the download occurs multiple times.

SOLUTION
Modify the existing mutex logic to check if a compiler of a specific version has already been downloaded or is currently being downloaded.

### DIFF
--- a/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
+++ b/packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
@@ -155,6 +155,10 @@ export class CompilerDownloader implements ICompilerDownloader {
     downloadStartedCb: (isCompilerDownloaded: boolean) => Promise<any>,
     downloadEndedCb: (isCompilerDownloaded: boolean) => Promise<any>
   ): Promise<void> {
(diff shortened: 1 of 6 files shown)


packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
```diff
@@ -155,6 +155,10 @@ export class CompilerDownloader implements ICompilerDownloader {
     downloadStartedCb: (isCompilerDownloaded: boolean) => Promise<any>,
     downloadEndedCb: (isCompilerDownloaded: boolean) => Promise<any>
   ): Promise<void> {
+    // Since only one process at a time can acquire the mutex, we avoid the risk of downloading the same compiler multiple times.
+    // This is because the mutex blocks access until a compiler has been fully downloaded, preventing any new process
+    // from checking whether that version of the compiler exists. Without mutex it might incorrectly
+    // return false, indicating that the compiler isn't present, even though it is currently being downloaded.
```
schaable, line 161: thanks for this!

schaable (approved):

PR #4552, opened 2023-11-01
Fix vyper downloader

fvictorio:
The latest releases of Vyper include tag names like "v0.3.10rc5" which are not valid semver strings. This was causing the downloder to fail when calling semver.eq. This fixes this by previously checking that the tag_name is a valid semver string.

EDIT: that kind of version string is valid for [PEP 440](https://peps.python.org/pep-0440/) but it's not for semver.

[PEP 440](https://peps.python.org/pep-0440/#pre-release-separators):

> Pre-releases should allow a ., -, or _ separator between the release segment and the pre-release segment. The normal form for this is without a separator.

[Semver](https://semver.org/#spec-item-9):

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version.

### DIFF
--- a/.changeset/tough-geese-tie.md
+++ b/.changeset/tough-geese-tie.md
@@ -1,0 +1,5 @@
+---
+"@nomiclabs/hardhat-vyper": patch
+---
+
+Fixed a problem with the Vyper compilers downloader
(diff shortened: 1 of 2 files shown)


ChristopherDedominici (approved):

alcuadrado (approved):
LGTM

schaable (approved):

PR #4553, opened 2023-11-02
refactor: do not insert pending block in blockchain

Wodann:


agostbiro (approved):
Thanks, LGTM. The state root failure is known and the [Vyper failure](https://github.com/NomicFoundation/hardhat/actions/runs/6728124285/job/18294392642?pr=4553) is a regression fixed in `main`. Created a PR to cherry pick the fix: https://github.com/NomicFoundation/hardhat/pull/4555

PR #4558, opened 2023-11-02
refactor: combine Node and NodeData

Wodann:
I left the rename of `Node` to `NodeData` for a separate PR, to simplify review


agostbiro (approved):
Thanks, LGTM!

PR #4563, opened 2023-11-03
fix: update mem pool on irregular state change

Wodann:


crates/edr_evm/src/mempool.rs
```diff
@@ -337,15 +337,18 @@ impl MemPool {
         ) -> bool {
             transaction.gas_limit() <= block_gas_limit
                 && transaction.upfront_cost() <= sender.balance
+                // Remove all finalized transactions
```
agostbiro, line 340: Is this comment correct?

agostbiro (approved):
Thanks, LGTM overall. Approving to expedite merging, but there is one comment that I think needs to be rephrased/removed before merging.

crates/edr_evm/src/mempool.rs
Wodann: Maybe the word `finalized` is overloaded?

What I'm trying to convey is that any pending blocks that have been mined should be removed. I'll use the word `mined` instead, as that's more accurate. Let me know if you have a better alternative; thanks!

PR #4567, opened 2023-11-07
feat: add scaffolding for provider

Wodann:


agostbiro (approved):
Thanks, LGTM!

crates/edr_provider/src/lib.rs
```diff
@@ -0,0 +1,279 @@
...
+    request: Request,
+) -> Result<serde_json::Value, ProviderError> {
+    match request {
+        Request::Eth(request) => handle_eth_request(data, request).await,
+        Request::Hardhat(request) => handle_hardhat_request(data, request).await,
+    }
+}
+
+/// Handles a batch of JSON requests for an execution provider.
+pub async fn handle_batch_request(
```
fvictorio, line 92: I'm curious why this and the previous function are public.

fvictorio (approved):

crates/edr_provider/src/lib.rs
Wodann: Good catch. Thanks!

PR #4571, opened 2023-11-08
chore(chain-config): deprecate xDai && update chiado urls

4rgon4ut:

---

This PR deprecates `xDai` chain mention as the name changed to `Gnosis` and new chain name already supported with corresponding valid config.
Also Chiado testnet urls were updated.


schaable (approved):

4rgon4ut:
Hey @schaable, thanks for the explanation. Changed the test and added changeset as you said, please have a view.

schaable:
@4rgon4ut, I've reconsidered removing xDai because, as you said, it would be a breaking change with no real benefit. I've updated the PR to only include the URL rename for the Chiado explorer.

4rgon4ut:
> @4rgon4ut, I've reconsidered removing xDai because, as you said, it would be a breaking change with no real benefit. I've updated the PR to only include the URL rename for the Chiado explorer.

Make sense, thank you

PR #4574, opened 2023-11-09
Avoid using a maxPriorityFeePerGas of 0

### LINKED ISSUE #4190: HardHat AutoGasManager's fee estimation breaks on low-traffic chains
### Version of Hardhat

2.17.0

### What happened?

Hardhat's `AutomaticGasProvider` seems to use its own algorithm based on eth_FeeHistory to compute maxPriorityFeePerGas and MaxFeePerGas. When the most recent block is empty, the code ends up returning "0" for maxPriorityFeePerGas, producing a transaction that will be immediately rejected by almost all node transaction pools with "transaction underpriced" errors.

We looked into this because of several reports of this error from people who are following the Base contract deployment guide [here](https://docs.base.org/guides/deploy-smart-contracts). I imagine this issue also arises on other development / testnet chains where traffic is low.

Obviously one can work around this by specifying your own gas price in the hardhat user config (we're updating or docs right now to do this), but better for auto-estimation to always work.

To that end, one option would be to leverage the eth_maxPriorityFeePerGas RPC method to get the node-computed estimate. These algorithms are already well tuned and more likely to do the right thing than any roll-your-own approach.  A simpler option would be to just hard-code a non-trivial minimum on the value always suggested (e.g. .1 gwei).

Use hardhat to submit a transaction to a low-traffic chain (e.g. a devnet) when configured to use the "auto" gas price mode. If the current block is empty the transaction will fail with "transaction underpriced".  Some example output containing additional instrumentation demonstrating the failure:

```
fee history: {
  baseFeePerGas: [ '0x32', '0x32' ],
  gasUsedRatio: [ 0.0015637666666666666 ],
  oldestBlock: '0x17d47a',
  reward: [ [ '0x0' ] ]
}
suggesteEip1559Values { maxFeePerGas: 63n, maxPriorityFeePerGas: 0n }

Result: 

ProviderError: transaction underpriced
    at HttpProvider.request (/Users/bayardo/tmp/node_modules/hardhat/src/internal/core/providers/http.ts:88:21)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async HardhatEthersSigner.sendTransaction (/Users/bayardo/tmp/node_modules/@nomicfoundation/hardhat-ethers/src/signers.ts:125:18)
    at async ContractFactory.deploy (/Users/bayardo/tmp/node_modules/ethers/src.ts/contract/factory.ts:112:22)
    at async main (/Users/bayardo/tmp/scripts/deploy.ts:4:15)
```

### Search terms

_No response_

fvictorio:
Closes #4190.

If the computation over the `eth_feeHistory` result results in a max priority fee of 0, we try to use the result of `eth_maxPriorityFeePerGas`. If that method doesn't exist, we use 1.

I should mention that I _couldn't_ reproduce this issue, because right now Base goerli has enough activity to not trigger this. But this change is low-risk, since the new code is only executed when the inferred max priority fee is 0.

Another way in which this is not ideal is that we are only using `eth_maxPriorityFeePerGas` when the estimation is 0. Perhaps ideally we should always use it? But this is a way bigger change that I can't investigate right now. I'm going to create follow-up issues for this.

### DIFF
--- a/packages/hardhat-core/test/internal/core/providers/gas-providers.ts
+++ b/packages/hardhat-core/test/internal/core/providers/gas-providers.ts
@@ -256,6 +256,98 @@ describe("AutomaticGasPriceProvider", () => {
       });
     });
 
+    describe("When the eth_feeHistory result causes maxPriorityFeePerGas to be 0 and eth_maxPriorityFeePerGas doesn't exist", function () {
+      const latestBaseFeeInMockedProvider = 80;
+
+      beforeEach(function () {
+        mockedProvider.setReturnValue("eth_feeHistory", {
+          baseFeePerGas: [
+            numberToRpcQuantity(latestBaseFeeInMockedProvider),
+            numberToRpcQuantity(
+              Math.floor((latestBaseFeeInMockedProvider * 9) / 8)
+            ),
+          ],
+          reward: [["0x0"]],
+        });
+
+        mockedProvider.setReturnValue("eth_getBlockByNumber", {
+          baseFeePerGas: "0x1",
+        });
+      });
+
+      it("should use the reward return value as default maxPriorityFeePerGas", async function () {
+        await provider.request({
+          method: "eth_sendTransaction",
+          params: [
+            {
+              from: "0x0000000000000000000000000000000000000011",
+              to: "0x0000000000000000000000000000000000000011",
+              value: 1,
+            },
+          ],
+        });
+
+        const expectedBaseFee = Math.floor(
+          latestBaseFeeInMockedProvider *
+            (9 / 8) **
+              Number(
+                AutomaticGasPriceProvider.EIP1559_BASE_FEE_MAX_FULL_BLOCKS_PREFERENCE
+              )
+        );
+
+        const [tx] = mockedProvider.getLatestParams("eth_sendTransaction");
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/test/internal/core/providers/gas-providers.ts
```diff
@@ -256,6 +256,98 @@ describe("AutomaticGasPriceProvider", () => {
...
+          ],
+          reward: [["0x0"]],
+        });
+
+        mockedProvider.setReturnValue("eth_getBlockByNumber", {
+          baseFeePerGas: "0x1",
+        });
+      });
+
+      it("should use the reward return value as default maxPriorityFeePerGas", async function () {
```
ChristopherDedominici, line 278: is this description correct? Shouldn't it be the default value 0x1?

ChristopherDedominici (approved):
I just left a comment in the tests but a part from that LGTM

schaable (approved):

packages/hardhat-core/test/internal/core/providers/gas-providers.ts
fvictorio: Good catch, fixed.

Issue #4138, opened 2023-07-12
Add explanation about Hardhat's default gas behavior


fvictorio:
Hardhat by default uses the full block gas limit, which is useful in many scenarios but can be confusing in others. We should explain this better in the docs.

See this comment: https://github.com/NomicFoundation/hardhat/issues/4090#issuecomment-1622155314

PR #4576, opened 2023-11-09
docs: Mining multiple transactions in 1 block on Hardhat Network (#4138)

dwardu:


alcuadrado (approved):
I really like this addition! Thanks!

@fvictorio if you agree with this, please merge it.

dwardu:
That’s great @alcuadrado!

One way it could be improved further is that the default remains `"auto"` for normal networks, but for Hardhat Network, rather than implicitly defaulting to the block gas-limit, `networks.hardhat.gas` changes from optional to required (i.e. no default), and `"blockGasLimit"` is introduced as a new enum value (like `"auto"`.) Then the sample projects would have their `networks.hardhat.gas` _explicitly_ preconfigured to `"blockGasLimit" `. In this way any old projects upgrading to a newer Hardhat would fail with the correct message.

fvictorio:
Thank you @dwardu!

> One way it could be improved further is that the default remains "auto" for normal networks, but for Hardhat Network, rather than implicitly defaulting to the block gas-limit, networks.hardhat.gas changes from optional to required (i.e. no default), and "blockGasLimit" is introduced as a new enum value (like "auto".) 

This is an interesting idea but it would be a breaking change, and I feel it would be kind of annoying to always have to specify the gas.

dwardu:
Thanks @fvictorio!

> This is an interesting idea but it would be a breaking change, and I feel it would be kind of annoying to always have to specify the gas.

Fair enough. It _might_ seem like a secondary issue, maybe not worthy of a breaking change, that is true. Maybe it’s because the solution was so counterintuitive (to me, at least) that I wanted to spare everyone else the pain :sweat_smile: But hopefully the docs change will do the trick :wink:

Thanks again.

PR #4578, opened 2023-11-09
feat: add provider to edr_napi

Wodann:


crates/edr_napi/src/provider.rs
```diff
@@ -0,0 +1,58 @@
...
+    provider: Arc<edr_provider::Provider>,
+}
+
+#[napi]
+impl Provider {
+    #[doc = "Constructs a new provider with the provided configuration."]
+    #[napi(ts_return_type = "Promise<Provider>")]
+    pub fn with_config(env: Env, config: ProviderConfig) -> napi::Result<JsObject> {
+        let config = edr_provider::ProviderConfig::try_from(config)?;
+        let runtime = runtime::Handle::current();
```
agostbiro, line 22: This will panic if called outside a tokio runtime. I'm wondering if we should use `Handle::try_current` instead and return an error if the function is called outside a runtime.

crates/edr_napi/src/provider.rs
```diff
@@ -0,0 +1,58 @@
...
+
+        let (deferred, promise) = env.create_deferred()?;
+        runtime.clone().spawn(async move {
```
agostbiro, line 25: Just for my understanding: what's the reason for cloning the runtime here?

crates/edr_napi/src/provider.rs
```diff
@@ -0,0 +1,58 @@
...
+        });
+
+        Ok(promise)
+    }
+
+    #[doc = "Handles a JSON-RPC request and returns a JSON-RPC response."]
+    #[napi]
+    pub async fn handle_request(&self, json_request: String) -> napi::Result<String> {
+        let request = serde_json::from_str(&json_request)
+            .map_err(|e| napi::Error::new(Status::InvalidArg, e.to_string()))?;
```
agostbiro, line 47: Should we include the request in the error or log it? IIRC `serde_json::Error` doesn't include the source string which will make debugging from JS hard.

agostbiro (changes requested):

crates/edr_napi/src/provider.rs
Wodann: This can never happen, as `napi` is guaranteed to provide the runtime for us.

crates/edr_napi/src/provider.rs
Wodann: The `runtime` is moved into the the `async` callback

crates/edr_napi/src/provider.rs
Wodann: Addressed in https://github.com/NomicFoundation/hardhat/pull/4578/commits/0f392206571870cb24275b3c15236ddf39c5b99f

fvictorio (approved):

PR #4581, opened 2023-11-13
Allow using EDR provider in Hardhat

fvictorio:
This PR lets us run Hardhat tests (and Hardhat in general) with the EDR provider:

```
HARDHAT_EXPERIMENTAL_VM_MODE=edr yarn mocha test/internal/hardhat-network/provider/modules/eth/methods/index.ts
```

Right now it's defaulting to `ethereumjs`, but feel free to change it to default to `edr` if that makes your life easier.

Notes:
- Some tests rely on the provider being a `HardhatNetworkProvider`, which is no longer the case. I fixed those build errors by just casting the provider to `any` and I added some `EDR-TODO` comments. I think we can ignore those for now, although we will need to handle them soon.
- I also made a small change in the `edr_napi` code to make the `cacheDir` field optional, since that made my changes simpler. I think the proper fix is to make the cache dir fully optional, down to the RPC client. I tried to do that, but hilarity ensued and I reverted everything.

Closes https://github.com/NomicFoundation/edr/issues/232


packages/hardhat-core/test/internal/hardhat-network/helpers/retrieveForkBlockNumber.ts
```diff
@@ -1,9 +1,8 @@
-import { HardhatNetworkProvider } from "../../../../src/internal/hardhat-network/provider/provider";
-
 /* eslint-disable @typescript-eslint/dot-notation */
 
+// EDR-TODO: this should be adapted or removed
 export async function retrieveLatestBlockNumber(
-  provider: HardhatNetworkProvider
+  provider: any
```
Wodann, line 5: Can't we just use the `eth_blockNumber` API for this?

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -393,3 +393,82 @@ export class HardhatNetworkProvider
...
+): Promise<EIP1193Provider> {
+  let eip1193Provider: EIP1193Provider;
+
+  const vmModeEnvVar = process.env.HARDHAT_EXPERIMENTAL_VM_MODE ?? "ethereumjs";
+
+  if (vmModeEnvVar === "edr") {
+    eip1193Provider = await EdrProviderWrapper.create(
+      hardhatNetworkProviderConfig
+    );
+  } else if (vmModeEnvVar === "ethereumjs") {
```
Wodann, line 460: I think we should also add the dual mode adapter to use `HardhatNetworkProvider`. That way we'll be running the `DualModeAdapter` at the component-level still, maintaining testing support for any changes that happen at that level.

Wodann (changes requested):
`rustfmt` is currently failing. Could you please fix that?

I have one more question:

How do we want to integrate this PR into `edr/main` while keeping CI functional? Once we merge this, most tasks on the CI will start failing, which is not ideal. I suggested that we keep dual-mode in a separate comment, which should keep those CI tests succeeding. Then the question becomes whether we are okay with `edr` tests failing until we finish all JSON-RPC methods?

agostbiro:
> `rustfmt` is currently failing. Could you please fix that?
> 
> I have one more question:
> 
> How do we want to integrate this PR into `edr/main` while keeping CI functional? Once we merge this, most tasks on the CI will start failing, which is not ideal. I suggested that we keep dual-mode in a separate comment, which should keep those CI tests succeeding. Then the question becomes whether we are okay with `edr` tests failing until we finish all JSON-RPC methods?

Thanks @fvictorio the PR looks good overall, but I agree with Remco that it'd be good to keep the EDR CI green to make sure that we don't introduce regressions by accident. One way we could achieve that is by making EDR tests opt-in either by having some config file with a grep configuration for passing tests or by disabling them dynamically in the source. Then we can enable the ones in the PRs that we added support for. Wdyt? Is this feasible?

packages/hardhat-core/test/internal/hardhat-network/helpers/retrieveForkBlockNumber.ts
fvictorio: Maybe? To be honest, I didn't want to spend any time on this since we'll have to do it properly later anyway, and I want to have the provider working asap since it's a big blocker for all of us.

fvictorio:
> One way we could achieve that is by making EDR tests opt-in either by having some config file with a grep configuration for passing tests or by disabling them dynamically in the source. Then we can enable the ones in the PRs that we added support for. Wdyt? Is this feasible?

Good point, I hadn't thought about this. I'll see if I can come up with something simple for it.

PR #4584, opened 2023-11-14
Merge main into edr/main

fvictorio:
We recently switched to `pnpm` in the `main` branch. This PR merges that branch into `edr/main` and adapts everything to work with `pnpm`.

By far the most difficult part to make it work was the `edr-npm-release.yml' workflow. The main reason is that most docker images don't have `pnpm` installed, so I had to update several jobs to install it inside the container before running the commands. We could've also used a custom docker image that extends the ones we were using and adds this step, but I didn't consider it was worth it.


crates/edr_napi/package.json
```diff
@@ -23,8 +23,10 @@
   "license": "MIT",
   "devDependencies": {
     "@napi-rs/cli": "^2.16.3",
+    "@types/chai": "^4.2.0",
+    "@types/mocha": ">=9.1.0",
```
fvictorio, line 27: The way packages are structured in `pnpm` is more explicit, which means that some things that were working by coincidence no longer work (this is a good thing). One of those was that we were missing these two packages here.

packages/hardhat-core/package.json
```diff
@@ -23,21 +23,22 @@
...
-    "build:tracing": "tsc --build --force --incremental .",
-    "pretest": "cd ../.. && yarn build",
+    "pretest": "cd ../.. && pnpm build",
     "test": "mocha --recursive \"test/**/*.ts\" --exit",
     "test:except-provider": "mocha --recursive \"test/**/*.ts\" --exclude \"test/internal/hardhat-network/provider/**/*.ts\" --exit",
     "test:except-tracing": "mocha --recursive \"test/**/*.ts\" --exclude \"test/internal/hardhat-network/stack-traces/**/*.ts\" --exit",
     "test:provider": "mocha --recursive \"test/internal/hardhat-network/provider/**/*.ts\" --exit",
     "test:tracing": "mocha --recursive \"test/internal/hardhat-network/stack-traces/**/*.ts\" --exit",
-    "prepublishOnly": "yarn build",
+    "test:forking": "mocha --recursive \"test/internal/hardhat-network/{helpers,jsonrpc,provider}/**/*.ts\"",
```
Wodann, line 36:
```suggestion
```

This was removed in `edr/main`, in favour of `test:provider`, `test:tracing`, test:except-provider`, and `test:except-tracing`

Wodann (approved):
I just spotted one merge error in the files that we've changed in `edr/main`. Once that's solved, feel free to merge.

packages/hardhat-core/package.json
fvictorio: Done, thanks!

.github/workflows/edr-npm-release.yml
```diff
@@ -29,35 +29,37 @@ jobs:
...
+              pnpm build --target i686-pc-windows-msvc
+              pnpm test
             target: i686-pc-windows-msvc
           - host: ubuntu-latest
             target: x86_64-unknown-linux-gnu
             docker: ghcr.io/napi-rs/napi-rs/nodejs-rust:lts-debian
             build: |-
               set -e &&
-              yarn build --target x86_64-unknown-linux-gnu &&
+              (wget -qO- 'https://unpkg.com/@pnpm/self-installer' | node) &&
```
agostbiro, line 47: Did the command fail without installing pnpm? I'm asking, because the source indicates that the [alpine](https://github.com/napi-rs/napi-rs/blob/9d62db47e89f78e94a713ead6978637f1158e3d8/alpine.Dockerfile#L19) and [debian](https://github.com/napi-rs/napi-rs/blob/9d62db47e89f78e94a713ead6978637f1158e3d8/debian.Dockerfile#L37) images have pnpm installed. Maybe it's just a path issue?

fvictorio: You are right, fixed!

agostbiro: Cheers!

agostbiro (approved):
I reviewed the EDR NPM release workflow changes and LGTM!

PR #4585, opened 2023-11-15
Fixes related to the pnpm migration

fvictorio:
Some things we missed when we migrated to pnpm.

Some comments:
- I removed the `only-allow` script because it adds 6 seconds to each installation. This can be improved by adding `only-allow` as a dev dependency itself, so that this cost is only paid the first time... but I'd rather just remove it; I don't think it's worth it.
- Dependabot supports pnpm, but (as far as I understand) it auto-detects it; the setting has to be `npm` anyway because the setting is about the ecosystem, not about the package manager.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,6 @@
     "typescript": "~5.0.0"
   },
   "scripts": {
-    "preinstall": "npx -y only-allow pnpm",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -12,7 +12,6 @@
     "typescript": "~5.0.0"
   },
   "scripts": {
-    "preinstall": "npx -y only-allow pnpm",
```
ChristopherDedominici, line 15: why this line has been removed?

ChristopherDedominici (approved):

PR #4588, opened 2023-11-15
feat: add `eth_getTransactionReceipt` support

agostbiro:
Adds `eth_getTransactionReceipt` support:

- Most of the changes are related to changing the log type in the receipt to match Hardhat output.
- I noticed that some log fields are optional in Hardhat, but compulsory in EDR. I marked these with a TODO comment. Should we keep it as is or make it optional to match Hardhat? I think it's ok as is, because worst case we return more data than Hardhat does.
- Testing strategy is that I generated a transaction receipt output from a Hardhat test and then checked that fields of the EDR return type from `handle_get_transaction_receipt` match this.


agostbiro:
Converted this to a draft as a I noticed some Hardhat tests are failing.

agostbiro:
This is now ready for review. This PR revealed a regression in the RPC client cache for which I created a follow up issue: https://github.com/NomicFoundation/edr/issues/238

crates/edr_eth/src/log/block.rs
```diff
@@ -24,15 +24,21 @@ pub struct FullBlockLog {
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub inner: ReceiptLog,
     /// block hash
+    // TODO review: this is optional in Hardhat. Should we keep it as is or make it optional to
```
Wodann, line 27: All of those are either always present or none of them are present. That's why we have a `ReceiptLog` and a `FullBlockLog`. In the former case, use `FullBlockLog`. In the latter case, use `ReceiptLog`

crates/edr_eth/src/log/receipt.rs
```diff
@@ -13,6 +13,7 @@ pub struct ReceiptLog {
     #[cfg_attr(feature = "serde", serde(flatten))]
     pub inner: Log,
     /// transaction hash
+    // TODO review: this is optional in Hardhat
```
Wodann, line 16: Same as the other one. We encoded optionality in static types: `Log` vs `ReceiptLog`

Wodann (approved):
I answered your questions. We can have a discussion about the design choice.

If you have no concerns, you'd just need to remove the `TODO`s before merging

crates/edr_eth/src/log/block.rs
agostbiro: Got it thanks!

agostbiro:
Ok cheers @Wodann removed the TODO comments merging. Thanks for resolving conflicts @fvictorio !

Issue #4386, opened 2023-09-16
Add support for scopes in `hh` autocomplete


fvictorio:
`hh` autocomplete should support scopes and scoped tasks.

search terms: scoped tasks

PR #4593, opened 2023-11-16
Add support for scopes in hh autocomplete

ChristopherDedominici:
[See issue here](https://github.com/NomicFoundation/hardhat/issues/4386)


packages/hardhat-core/test/internal/cli/autocomplete.ts
```diff
@@ -402,7 +407,7 @@ describe("autocomplete", function () {
     });
   });
 
-  describe("overriden task", () => {
+  describe("override task", () => {
```
fvictorio, line 410: I think the previous description was correct. This is a suite of tests around an overridden task (well, it had a typo, it was missing a d).

```suggestion
  describe("overridden task", () => {
```

packages/hardhat-core/src/internal/cli/autocomplete.ts
```diff
@@ -38,6 +38,16 @@ interface CompletionData {
       };
     };
   };
+  scopes: {
+    [taskName: string]: {
```
fvictorio, line 42:
```suggestion
    [scopeName: string]: {
```

fvictorio (changes requested):
This doesn't seem to working for the flags/params of the scoped tasks. For example, given this:

```js
scope("my-scope").task("my-task").addFlag("foo")
```

then doing `hh my-scope my-task --<tab>` should suggest `--foo`.

packages/hardhat-core/src/internal/cli/autocomplete.ts
```diff
@@ -125,39 +176,67 @@ export async function complete({
...
       return HARDHAT_COMPLETE_FILES;
     }
   }
 
-  // if there's no task, we complete either tasks or params
-  if (task === undefined || tasks[task] === undefined) {
+  // If there's no task or scope, we complete either tasks and scopes or params
+  if (
+    (taskName === undefined || tasks[taskName] === undefined) &&
+    (scopeName === undefined || scopes[scopeName] === undefined)
```
fvictorio, line 189: Can this condition be simplified to `taskDefinition === undefined && scopeDefinition === undefined)`?

packages/hardhat-core/src/internal/cli/autocomplete.ts
```diff
@@ -125,39 +176,67 @@ export async function complete({
...
+    return taskSuggestions
+      .concat(scopeSuggestions)
+      .filter((x) => startsWithLast(x.name));
+  }
+
+  // If there's a scope but not a task, we complete with the scopes'tasks
+  if (
+    taskName === undefined &&
+    scopeName !== undefined &&
+    scopes[scopeName] !== undefined
```
fvictorio, line 216: Similar comment here: can we use `scopeDefinition` and `taskDefinition`?

fvictorio (approved):
Minor cosmetic comments; feel free to merge after applying (or not) them, assuming tests pass.

PR #4596, opened 2023-11-17
feat: add hardhat config RPC methods

Wodann:
As part of this change, I made the `network_id` a `u64` as there was no need for it to support up to `U256`. It's represented as a string decimal in the JSON-RPC. The `U64` allows deserialization of these values too, making it a convenient helper.

### DIFF
--- a/crates/edr_provider/src/requests/hardhat/config.rs
+++ b/crates/edr_provider/src/requests/hardhat/config.rs
@@ -1,6 +1,18 @@
 use edr_eth::{Address, SpecId, B256, U256};
+use rpc_hardhat::Metadata;
 
-use crate::{data::ProviderData, ProviderError};
+use crate::{data::ProviderData, requests::eth::client_version, ProviderError};
+
+pub fn handle_metadata_request(data: &ProviderData) -> Result<Metadata, ProviderError> {
+    Ok(Metadata {
+        client_version: client_version(),
+        chain_id: data.chain_id(),
+        instance_id: *data.instance_id(),
+        latest_block_number: data.last_block_number(),
+        latest_block_hash: *data.last_block()?.hash(),
+        forked_network: data.fork_metadata().cloned(),
+    })
+}
 
 pub fn handle_set_coinbase_request(
     data: &mut ProviderData,
--- a/crates/edr_provider/src/requests/hardhat/config.rs
+++ b/crates/edr_provider/src/requests/hardhat/config.rs
@@ -1,4 +1,4 @@
-use edr_eth::{Address, SpecId, U256};
+use edr_eth::{Address, SpecId, B256, U256};
 
 use crate::{data::ProviderData, ProviderError};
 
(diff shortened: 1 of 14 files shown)


crates/edr_rpc_hardhat/src/lib.rs
```diff
@@ -118,7 +119,7 @@ pub enum Request {
         serialize_with = "single_to_sequence",
         deserialize_with = "sequence_to_single"
     )]
-    SetPrevRandao(ZeroXPrefixedBytes),
+    SetPrevRandao(B256),
```
agostbiro, line 122: Just double checking: does `B256` serializes into `0x...` string?

crates/edr_provider/src/requests/hardhat/config.rs
```diff
@@ -0,0 +1,58 @@
...
+    Ok(true)
+}
+
+pub fn handle_set_next_block_base_fee_per_gas_request(
+    data: &mut ProviderData,
+    base_fee_per_gas: U256,
+) -> Result<bool, ProviderError> {
+    let spec_id = data.spec_id();
+    if spec_id < SpecId::LONDON {
+        return Err(ProviderError::UnmetHardfork {
```
agostbiro, line 32: Wondering if this is going to match error returned by Hardhat. Is that a concern at all @alcuadrado @fvictorio ?

crates/edr_rpc_hardhat/src/metadata.rs
```diff
@@ -0,0 +1,37 @@
...
+/// Metadata about the forked network.
+#[derive(Clone, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ForkMetadata {
+    /// The chainId of the network that is being forked
+    pub chain_id: u64,
+    /// The number of the block that the network forked from.
+    pub fork_block_number: u64,
+    /// The hash of the block that the network forked from.
+    pub fork_block_hash: B256,
```
agostbiro, line 36: Noticed that the Hardhat type doesn't explicit say that this is a 0x-prefixed string. Are we sure it is?

```
    // The hash of the block that the network forked from.
    forkBlockHash: string;
```

agostbiro (approved):
Thanks LGTM overall. I left some questions to double check some things, approving assuming none of those need changes.

crates/edr_rpc_hardhat/src/lib.rs
Wodann: Yes, it does.

crates/edr_rpc_hardhat/src/metadata.rs
Wodann:
```ts
    const latestBlockHashHex = latestBlock.header.hash().toString("hex");
    const latestBlockHash = `0x${latestBlockHashHex}`;
```
https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/hardhat-network/provider/node.ts#L2719

PR #4608, opened 2023-11-22
feat: add mining config RPC methods

Wodann:


agostbiro (approved):
Thanks, code LGTM

packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToEdr.ts
```diff
@@ -258,6 +260,19 @@ export function ethereumjsHeaderDataToEdrBlockOptions(
   };
 }
 
+export function ethereumjsIntervalMiningConfigToEdr(
+  config: IntervalMiningConfig
+): number | IntervalRange {
+  if (isNumber(config)) {
```
fvictorio, line 266: I try to avoid using lodash if possible, since we'll probably remove it as a dependency eventually. You can just do this:

```suggestion
  if (typeof config === "number") {
```

crates/edr_provider/src/config.rs
```diff
@@ -28,3 +47,21 @@ pub struct AccountConfig {
...
+            order: MineOrdering::Priority,
+        }
+    }
+}
+
+impl Default for MiningConfig {
+    fn default() -> Self {
+        Self {
+            auto_mine: true,
+            interval: IntervalConfig::Fixed(0),
```
fvictorio, line 63: I regret modelling a disabled interval mining as an interval of 0. If it isn't a lot of work, what about adding a `Disabled` variant to the enum?

packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToEdr.ts
Wodann: Addressed in https://github.com/NomicFoundation/hardhat/pull/4619

crates/edr_provider/src/config.rs
Wodann: Addressed in https://github.com/NomicFoundation/hardhat/pull/4619

PR #4621, opened 2023-11-24
fix: delete cache entry for invalid type entries

agostbiro:
Follow up from https://github.com/NomicFoundation/hardhat/pull/4588 to prevent cache entries with an invalid type permanently crashing the application. A cache item can have an invalid type for example if we add a new field to a cached type. 

The PR fixes this by deleting cache items with an invalid type upon deserialization. Single calls make a remote query if this happens. Batch calls fail first, but in subsequent batch calls, it'll be a cache miss and work on the second try. We cannot do better for batch calls without major refactoring due to the logical separation between fetching batch items and deserializing them (which is done in order to let us support heterogeneous calls in a fairly straightforward way).

I also removed some client tests that seemed superfluous in order to save on the rate limiting budget. I removed calls that were either duplicate tests or calls where the expected result is that nothing is returned.

### DIFF
--- a/crates/edr_eth/src/remote/client.rs
+++ b/crates/edr_eth/src/remote/client.rs
@@ -3,9 +3,11 @@ use std::{
     io,
     path::{Path, PathBuf},
     sync::atomic::{AtomicU64, Ordering},
+    thread::available_parallelism,
     time::{Duration, Instant},
 };
 
+use futures::stream::StreamExt;
 use itertools::{izip, Itertools};
 use reqwest::Client as HttpClient;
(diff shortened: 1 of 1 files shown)


crates/edr_eth/src/remote/client.rs
```diff
@@ -1162,24 +1208,6 @@ mod tests {
             assert!(client.is_cacheable_block_number(16220843).await.unwrap());
         }
 
-        #[tokio::test]
```
Wodann, line 1165: Could you please explain the reasoning behind the removal of these tests?

Wodann (approved):

PR #4631, opened 2023-11-30
Added a guide for verifying contracts with hardhat ignition

zoeyTM:
Adding a guide to the Hardhat Ignition docs explaining how to verify contracts. 

Supports [this PR adding the verify task](https://github.com/NomicFoundation/hardhat-ignition/pull/632).


alcuadrado:
I think this guide is looking good, but given that the configuration variables feature is pretty new, maybe we should also explain how to set them, or at least link to their docs.

zoeyTM:
> I think this guide is looking good, but given that the configuration variables feature is pretty new, maybe we should also explain how to set them, or at least link to their docs.

We do link to the `hardhat-verify` docs at the end. Is that what you mean?

alcuadrado (approved):
Approved, but do not merge until the new version of ignition gets released

PR #4646, opened 2023-12-04
build(clippy): Don't lint against deny-by-default Clippy lints

Xanewok:
Something I noticed when adding more Clippy lints to Slang as part of https://github.com/NomicFoundation/slang/issues/155; figured we might as well remove them, since they're in the "correctness" lint group and so probably won't be moved to warn-by-default or even allow-by-default sometime soon.


Wodann (approved):

Wodann:
Thank you, @Xanewok! 🙏

PR #4649, opened 2023-12-05
fix: overwrite genesis account nonces and code in fork mode

agostbiro:
Fix overwriting genesis account nonces and code in fork mode by reinstating the previous approach of fetching the account info from the remote: https://github.com/NomicFoundation/hardhat/blob/e822bd0e01b1b14653f2ac58ebc28743264ef3d7/crates/rethnet_evm/src/state/fork.rs#L48

Note that in addition to nonces we also reinstate code now for completeness as we only want to allow overwriting the balance.

### DIFF
--- a/crates/edr_eth/src/remote/client.rs
+++ b/crates/edr_eth/src/remote/client.rs
@@ -669,33 +669,53 @@ impl RpcClient {
         address: &Address,
         block: Option<BlockSpec>,
     ) -> Result<AccountInfo, RpcClientError> {
-        let inputs = &[
-            MethodInvocation::GetBalance(*address, block.clone()),
-            MethodInvocation::GetTransactionCount(*address, block.clone()),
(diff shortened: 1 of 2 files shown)


crates/edr_eth/src/remote/client.rs
```diff
@@ -669,33 +669,53 @@ impl RpcClient {
...
+            .flat_map(|address| {
+                [
+                    MethodInvocation::GetBalance(*address, block.clone()),
+                    MethodInvocation::GetTransactionCount(*address, block.clone()),
+                    MethodInvocation::GetCode(*address, block.clone()),
+                ]
+            })
+            .collect();
+
+        let responses = self.batch_call(inputs.as_slice()).await?;
```
Wodann, line 696: Out of curiosity, why `_.as_slice()` and not `&_`?

Wodann (approved):

PR #4651, opened 2023-12-06
fix: balance override for local accounts in forked blockchain is not applied

agostbiro:


Wodann (approved):

crates/edr_provider/src/test_utils.rs
```diff
@@ -38,12 +41,22 @@ pub fn create_test_config_with_impersonated_accounts(
...
 
+    let mut fork = None;
+    if forked {
+        fork = Some(ForkConfig {
+            json_rpc_url: get_alchemy_url(),
+            // Random recent block for better cache consistency
+            block_number: Some(18_725_000),
+            http_headers: None,
+        });
+    }
```
Wodann, line 52:
```suggestion
    let fork = if forked {
        Some(ForkConfig {
            json_rpc_url: get_alchemy_url(),
            // Random recent block for better cache consistency
            block_number: Some(18_725_000),
            http_headers: None,
        });
    } else {
        None
    };
```

Wodann: I prefer this pattern, as it avoids the variable from being mutable.

agostbiro: Good call, thanks!

PR #4657, opened 2023-12-07
Port ContractsIdentifier to EDR

fvictorio:
Migration of the `ContractsIdentifier` class to EDR. This is based on the [`optimize-contract-identifier`](https://github.com/NomicFoundation/hardhat/tree/optimize-contract-identifier) branch, in which @alcuadrado worked on but that we never merged. The relevant files there are mainly:

- The [radix tree implementation](https://github.com/NomicFoundation/hardhat/blob/optimize-contract-identifier/packages/hardhat-core/src/internal/hardhat-network/stack-traces/radix-tree.ts)
- The [contracts identifier implementation](https://github.com/NomicFoundation/hardhat/blob/optimize-contract-identifier/packages/hardhat-core/src/internal/hardhat-network/stack-traces/contracts-identifier.ts)
- The [radix tree tests](https://github.com/NomicFoundation/hardhat/blob/optimize-contract-identifier/packages/hardhat-core/test/internal/hardhat-network/stack-traces/radix-tree.ts)

---

`ContractsIdentifier` is a structure that is meant to contain all the "known" bytecodes in a project (both deployment and runtime bytecodes) and then use them to identify which contract is used in a given execution (because the "Bytecode" structure has a (possible) pointer to an actual contract; this is not currently reflected in the stubs I'm using).

The API is simple.

- `add_bytecode` adds a known bytecode to it
- `get_bytecode_from_message_trace` receives a trace and tries to obtain a bytecode that corresponds to it

As an example of the kind of heuristics this uses: two bytecodes can match but have different metadata hashes (because, for example, one of them has a whitespace modification or a change in a comment). The ContractsIdentifier checks if their difference starts after the metadata marker (FDFE) and in that case it considers them equal.

The code has comments for each of these things, but if anything is not clear please let me know and I will add/update the comments accordingly.


crates/edr_solidity/src/contracts_identifier.rs
```diff
@@ -0,0 +1,107 @@
...
+    bytecodes: HashMap<u64, &'a Bytecode>,
+}
+
+impl<'a> ContractsIdentifier<'a> {
+    pub fn new() -> Self {
+        Self {
+            tree: RadixTree::new(),
+            bytecodes: HashMap::new(),
+        }
+    }
```
Wodann, line 48: Clippy should have complained about this (are you running it?), but the `Default` implement should be used instead of an empty constructor

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,416 @@
...
+    fn new(content: Vec<u8>, is_present: bool, bytes_matched_before: usize) -> RadixNode {
+        RadixNode {
+            content,
+            is_present,
+            bytes_matched_before,
+            child_nodes: HashMap::new(),
+        }
+    }
+
+    pub fn get_content(&self) -> &Vec<u8> {
```
Wodann, line 21:
```suggestion
    pub fn content(&self) -> &Vec<u8> {
```

In Rust getters don't prepend `get_` as it's redundant and complicates suggestions of the LSP (as all getters start with the same 3 letters)

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,416 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct RadixTree {
+    root: RadixNode,
+}
+
+impl RadixTree {
+    pub fn new() -> RadixTree {
```
Wodann, line 158: Use `Default`

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        match next_node {
+            None => (false, matched, &self),
+            Some(next_node) => next_node.get_max_match(&word[prefix_length..]),
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct RadixTree {
```
Wodann, line 153: Is something about this implementation of a radix tree different compared to existing crates; like https://crates.io/crates/radix-tree?

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,416 @@
...
+
+                let node = RadixNode::new(word, true, bytes_matched_before);
+
+                self.child_nodes.insert(b, node);
+            }
+            Some(mut next_node) => {
+                let prefix_length = get_shared_prefix_length(&word, &next_node.content);
+
+                // We know it's at least 1
+                assert!(prefix_length > 0);
```
agostbiro, line 47: I'm wondering how de we know that it's at least 1? Is it because we successfully we removed it?

crates/edr_solidity/src/contracts_identifier.rs
```diff
@@ -0,0 +1,107 @@
+mod radix_tree;
+
+use std::collections::hash_map::DefaultHasher;
+use std::collections::HashMap;
```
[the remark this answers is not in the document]
fvictorio: No, sorry, I did ran the formatter, then I saw that the CI was failing and realized I forgot to add the nightly flag.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: To be honest, I didn't even consider using a crate for this. But I think having our own implementation makes sense, since this is a relatively simple data structure and a very core part of the tracing logic, so having flexibility seems important. Anyway, I'm not against using an external crate if you think it's a better idea, but in that case we should:

- Make sure that it does the same that our implementation does
- Implement the `getDescendantSuffixes` for it (it doesn't seem to have an equivalent function)

I'd say we complete the implementation of `ContractsIdentifier` with this implementation and then decide if we want to swap the radix tree with an external crate.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: The way this is implemented, the key of an element in `child_nodes` is the first byte of its value, so if `.get(word[0])` exists, then its value must start with `word[0]` and the shared prefix is at least 1.

I'm pretty sure there is a smarter way to implement this that removes this assumption, but I tried to keep the existing implementation as much as possible to get this done sooner.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,416 @@
...
+                if prefix_length == next_node.content.len() {
+                    // Check if the next node matches the word exactly
+                    if prefix_length == word.len() {
+                        next_node.is_present = true;
+                        self.child_nodes.insert(b, next_node);
+                        return;
+                    }
+
+                    // TODO can this (and all the other to_vec's) be replaced with .drain()?
+                    next_node.add_word(word[prefix_length..].to_vec());
```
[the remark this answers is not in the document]
fvictorio: That sounds like a good idea, but I think this conflicts with @Wodann suggestion of using `Bytes` instead of `Vec<u8>`.

crates/edr_solidity/src/contracts_identifier.rs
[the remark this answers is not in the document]
fvictorio: Fixed.

crates/edr_solidity/src/contracts_identifier.rs
fvictorio: Done.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: Done.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,416 @@
...
+            bytes_matched_before,
+            child_nodes: HashMap::new(),
+        }
+    }
+
+    pub fn get_content(&self) -> &Vec<u8> {
+        &self.content
+    }
+
+    fn add_word(&mut self, word: Vec<u8>) {
```
Wodann, line 25: Should we not signal that we (un)succesfully added the word?

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,416 @@
...
+        if word.is_empty() {
+            return;
+        }
+
+        let b = word[0];
```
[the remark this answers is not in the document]
Wodann: This might also allow you to skip the `assert`?

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        // we temporarily remove the next node and then insert it back, possibly mutated and/or in a different position
+        let next_node = self.child_nodes.remove(&b);
+
+        match next_node {
+            None => {
+                let bytes_matched_before = self.bytes_matched_before + self.content.len();
+
+                let node = RadixNode::new(word, true, bytes_matched_before);
```
Wodann, line 40: There's no need to use a `match` for option, as you can do `if let Some(mut next_node) = next_node { .. } else { .. }`

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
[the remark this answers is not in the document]
Wodann: With `Bytes` you can use this method:

https://docs.rs/bytes/1.5.0/bytes/struct.Bytes.html#method.split_off

win-win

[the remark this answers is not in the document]
fvictorio: Will do!

crates/edr_solidity/src/contracts_identifier.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::hash::{Hash, Hasher};
+
+use radix_tree::RadixTree;
+
+// bytecode and trace stubs
+pub struct Bytecode {
+    normalized_code: Vec<u8>,
```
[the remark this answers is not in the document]
fvictorio: Done.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: I don't think there's a concept of "failure" here. The only thing we could signal is if the word was already there or not. But since (currently) no callee needs that, I wouldn't add it for now.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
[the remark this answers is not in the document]
fvictorio: Hmm, I don't think so. The assert is related to the result of `get_shared_prefix_length` and the fact that an "edge" in a node has as its value the first byte of the child node. It's unrelated to the word being or not being empty.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: Done, thanks

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
Wodann: If the word is empty, we don't do anything. That seems like it's breaking the user's expectation to me:

1. I call `foo.add_word(word)`
2. The function returns without panicking. As the function signature doesn't communicate that there is a possibility my word won't actually be added to the `radix_tree` I assume it has been added. In fact, it has not.
3. If I call `get_max_match`, I think the result will be a `false` boolean as `entire_word_matched == true` but `entire_content_matched == false`

I would consider this discrepancy between expectation and reality to be a failure mode

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,540 @@
+use std::collections::BTreeMap;
+
+use revm_primitives::Bytes;
+
+#[derive(Debug)]
+pub struct RadixNode {
+    content: Bytes,
+    is_present: bool,
+    bytes_matched_before: usize,
```
Wodann, line 9: Is this an optimisation? It seems redundant as you can accumulate the content length (already pre-calculated) as you traverse the tree

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,540 @@
+use std::collections::BTreeMap;
+
+use revm_primitives::Bytes;
+
+#[derive(Debug)]
+pub struct RadixNode {
+    content: Bytes,
+    is_present: bool,
```
Wodann, line 8: Why not make this an `Option<Bytes>`?

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,540 @@
...
+impl Default for RadixTree {
+    fn default() -> Self {
+        RadixTree {
+            root: RadixNode::new(Bytes::new(), false, 0),
+        }
+    }
+}
+
+fn get_shared_prefix_length(a: &[u8], b: &[u8]) -> usize {
+    let max_index = std::cmp::min(a.len(), b.len());
```
Wodann, line 198:
```rust
a.iter().zip(b.iter()).enumerate().find(|(_, (a, b))| a != b).map_or_else(|| a.len().min(b.len()), |(idx, _)| idx)
```

`zip` limits to the shortest length

crates/edr_solidity/src/contracts_identifier.rs
```diff
@@ -0,0 +1,703 @@
...
+    s.finish()
+}
+
+fn is_matching_metadata(code: &[u8], last_byte: usize) -> bool {
+    let mut byte = 0;
+    while byte < last_byte {
+        let opcode = code[byte];
+
+        // Solidity always emits REVERT INVALID right before the metadata
+        if opcode == opcode::REVERT && code[byte + 1] == opcode::INVALID {
```
Wodann, line 270: Indexing using `[..]` can panic and should thus be avoided.

You can rewrite this using https://doc.rust-lang.org/std/primitive.slice.html#method.windows

Wodann (changes requested):
I need to go over the code another time, because there was quite a lot to review, but here's my first pass

crates/edr_solidity/src/contracts_identifier.rs
```diff
@@ -0,0 +1,703 @@
...
+            EvmMessageTrace::Call(call_message_trace) => &call_message_trace.code,
+        }
+    }
+}
+
+// TODO add cache
+#[derive(Default)]
+pub struct ContractsIdentifier<'a> {
+    tree: RadixTree,
+    bytecodes: HashMap<u64, &'a Bytecode>,
```
agostbiro, line 68: I'm wondering if the `u64` space is large enough to avoid collisions? 

A [news article](https://cryptopotato.com/over-44-million-contracts-deployed-to-ethereum-since-genesis-research/) from 2022 indicates that 44 million smart contracts have been deployed on mainnet, so I think an estimate of 100m smart contracts existing across all chains is a reasonable estimate. With ~100m smart contracts we get roughly a 1 in 2000 probability for a collision using a 64-bit hash space, which is I think too high (see [here](https://en.wikipedia.org/wiki/Birthday_problem#Square_approximation) for formula). 

If a collision can have security implications, I'd increase the hash space to 192-bit, as that'd result in a 2^-129 probability of collision for 2^32 smart contracts. Otherwise, I'd increase the hash space to 128-bit, as that'd result in a 2^-65 probability of collision which seems acceptable as it's comparable to correctly guessing the first time an 11-character alphanumeric password.

For the hash function I'd use [blake3](https://github.com/BLAKE3-team/BLAKE3) in both cases as it's a cryptographic hash function that's faster than md5.

agostbiro (changes requested):

crates/edr_solidity/src/contracts_identifier.rs
agostbiro: On a second reading of the code, I'm wondering if we could use the `Bytes` as keys directly? Then we could let the `HashMap` resolve collisions.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: In that case I would prefer to just panic when an empty word is passed. 

If we return a `bool`, then we should also analyze what happens when a word that already exists is added; for example, `HashSet` returns `false` in that case. This means that there would be three cases here: the word was not added because it was empty, the word was not added because it already existed, and the word was added. Since I don't think it's worth spending time on that analysis, I would just panic for empty words.

crates/edr_solidity/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "edr_solidity"
+version = "0.2.0-dev"
+edition = "2021"
+
+[dependencies]
+revm = { git = "https://github.com/Wodann/revm", rev = "93d8ebe", version = "3.3", default-features = false, features = ["dev", "serde", "std"] }
+revm-primitives = { git = "https://github.com/Wodann/revm", rev = "93d8ebe", version = "1.1", default-features = false }
```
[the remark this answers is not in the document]
fvictorio: Done in 59776f0a650061decb6262e3ca072a78493df3b7, but I have two concerns:

1. I only did it for `revm_primitives`, not for `revm`, because `edr_eth` doesn't use that crate.
2. I had to re-export `hex_literal` from `edr_eth` too, which feels weird because the current re-exports are just types. My impression is that this is fine, but I might be wrong.

crates/edr_solidity/src/opcodes.rs
```diff
@@ -0,0 +1,17 @@
+use revm::interpreter::opcode;
+
+fn is_push(opcode: u8) -> bool {
+    (opcode::PUSH1..=opcode::PUSH32).contains(&opcode)
+}
+
+fn get_push_length(opcode: u8) -> usize {
```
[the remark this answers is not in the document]
fvictorio: Sorry, I keep doing this. I fixed the ones you mentioned in 040048592ac85d46898978f8298e88370cdadb30. There are still two more: `get_max_match` and `get_shared_prefix_length`. Should those be renamed too, or are those fine because they are not "straightforward" getters?

crates/edr_solidity/src/contracts_identifier.rs
```diff
@@ -0,0 +1,703 @@
...
+}
+
+impl<'a> ContractsIdentifier<'a> {
+    pub fn add_bytecode(&mut self, bytecode: &'a Bytecode) {
+        self.tree.add_word(bytecode.normalized_code.clone());
+        self.bytecodes
+            .insert(calculate_hash(&bytecode.normalized_code), bytecode);
+    }
+
+    pub fn get_bytecode_from_message_trace(&self, trace: EvmMessageTrace) -> Option<&Bytecode> {
```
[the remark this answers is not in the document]
fvictorio: 040048592ac85d46898978f8298e88370cdadb30

crates/edr_solidity/src/opcodes.rs
```diff
@@ -0,0 +1,17 @@
...
+    (opcode - opcode::PUSH1 + 1) as usize
+}
+
+pub fn get_opcode_length(opcode: u8) -> usize {
```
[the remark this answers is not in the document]
fvictorio: 040048592ac85d46898978f8298e88370cdadb30

crates/edr_solidity/src/opcodes.rs
```diff
@@ -0,0 +1,17 @@
...
+    if !is_push(opcode) {
+        return 1;
+    }
+
+    1 + get_push_length(opcode)
```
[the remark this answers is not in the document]
fvictorio: 040048592ac85d46898978f8298e88370cdadb30

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
```diff
@@ -0,0 +1,540 @@
...
+
+        // we temporarily remove the next node and then insert it back, possibly mutated
+        // and/or in a different position
+        let next_node = self.child_nodes.remove(&b);
+
+        if let Some(mut next_node) = next_node {
+            let prefix_length = get_shared_prefix_length(&word, &next_node.content);
+
+            // We know it's at least 1
+            assert!(prefix_length > 0);
```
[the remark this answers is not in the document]
fvictorio: 60ec224e6139b967f9a902df06ff25785f6433ed

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: It's an optimization, yes. I tried to keep the original approach as much as possible to get this done faster, so I'm not 100% sure if it's a valid optimization or not.

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: Bytes is never missing, `is_present` is about that node corresponding to an actual word or not.

For example, let's say you add the words `ab` and `ac`. The tree would look like this:

```
 root
  |
  a
 / \
b   c
```

In this case the node that has `a` is not present, because it doesn't correspond to an actual word. If you then add the word `a`, then the only change to the tree is that the `is_present` field of that node becomes `true`. Does that make sense?

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: Thanks, done in 01080aaca06e499c7f1dff15a4d64aedfa01ac2b

crates/edr_solidity/src/contracts_identifier.rs
fvictorio: Thanks for this, it was a dumb oversight on my part (the same code is "safe" in javascript, with big quotes). I tried to add a test that would crash because of this but that made me realize that right now it's actually impossible. But proving that (and making sure that it never becomes possible) is way harder than just avoiding indexing with `[]` :sweat_smile:

That being said, I'm not sure this method can easily be re-written with `.windows` because I also need to advance the window by an arbitrary number (explanation of why this is necessary below). What do you think about this?

```rust
fn is_matching_metadata(code: &[u8]) -> bool {
    let mut byte = 0;
    while let (Some(opcode), Some(next_opcode)) = (code.get(byte), code.get(byte + 1)) {
        if *opcode == opcode::REVERT && *next_opcode == opcode::INVALID {
            return true;
        }

        byte += opcode_length(*opcode);
    }

    false
}
```

and the call becomes `is_matching_metadata(&code[..matched_bytes])`.

---

The reason we use `opcode_length` to skip forward is not (just) an optimization. You could have, for example, a `PUSH20` whose argument, by chance, contains the `REVERT INVALID` sequence.

crates/edr_solidity/src/contracts_identifier.rs
fvictorio: Thanks, using the code as the key seems to work fine. Done in b2f692b193d14af6fff6683751888b6152a47769

crates/edr_solidity/src/contracts_identifier.rs
```diff
@@ -0,0 +1,703 @@
...
+impl EvmMessageTrace {
+    fn code(&self) -> &Bytes {
+        match self {
+            EvmMessageTrace::Create(create_message_trace) => &create_message_trace.code,
+            EvmMessageTrace::Call(call_message_trace) => &call_message_trace.code,
+        }
+    }
+}
+
+// TODO add cache
```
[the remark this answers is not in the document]
fvictorio: Done in f39a9014edd71a20c2e2d86a01015920eb486306, also for the other TODO in that file.

crates/edr_solidity/src/opcodes.rs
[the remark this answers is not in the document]
Wodann: For getters there is no requirement that they return instantly. If you want to be explicit about there being different APIs, one for return a cached value vs another calculating the value; you can use another word compared to `get`, such as `calculate_` or `determine_`

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
Wodann: And the `content` would then be:

a: a
b: ab
c: ac

?

crates/edr_solidity/src/contracts_identifier.rs
Wodann: That would work. Alternatively, you can use `.get(..).expect("Some reason")` to signal that you'd never expect this to fail and why.

That's also valid and safe. It will result in a reason for the panic at runtime, but at least it's explicitly documented and can be tested for

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
Wodann: You can return a `SomeError` and `Result<(), SomeError>` from that function. At leat that'll allow people to recover from the error

crates/edr_solidity/src/contracts_identifier/radix_tree.rs
fvictorio: No, the content is not complete, it would simply be:

a: a
b: b
c: c

The image in [this article](https://en.wikipedia.org/wiki/Radix_tree) explains it better, although it doesn't include the `is_present` flag.

crates/edr_solidity/src/contracts_identifier.rs
fvictorio: Refactored in 2af60e1d241973a23911fcd5201b6f72d0c88e7a

Wodann (approved):

PR #4661, opened 2023-12-08
docs: add a `using viem` guide to Ignition

kanej:
Add a guide to installing and using the `@nomicfoundation/hardhat-ignition-viem` package and its support for returning Viem contract instances from the Ignition Helper.

### DIFF
--- a/docs/src/content/ignition/docs/getting-started/index.md
+++ b/docs/src/content/ignition/docs/getting-started/index.md
@@ -53,7 +53,7 @@ npm install --save-dev @nomicfoundation/hardhat-ignition-ethers
 :::tab{value="npm 6"}
 
 ```sh
-npm install --save-dev @nomicfoundation/hardhat-ignition-ethers
+npm install --save-dev @nomicfoundation/hardhat-ignition-ethers @nomicfoundation/hardhat-ethers @nomicfoundation/hardhat-ignition @nomicfoundation/hardhat-verify @nomicfoundation/ignition-core ethers
 ```
 
 :::
@@ -61,7 +61,7 @@ npm install --save-dev @nomicfoundation/hardhat-ignition-ethers
 :::tab{value=yarn}
 
 ```sh
-yarn add --dev @nomicfoundation/hardhat-ignition-ethers
+yarn add --dev @nomicfoundation/hardhat-ignition-ethers @nomicfoundation/hardhat-ethers @nomicfoundation/hardhat-ignition @nomicfoundation/hardhat-verify @nomicfoundation/ignition-core ethers
 ```
 
 :::
--- a/docs/src/content/ignition/docs/getting-started/index.md
+++ b/docs/src/content/ignition/docs/getting-started/index.md
@@ -75,7 +75,7 @@ Finally, add this to your config file to [enable the plugin](../../../hardhat-ru
 :::tab{value="TypeScript"}
 
 ```typescript
-import "@nomicfoundation/hardhat-ignition";
+import "@nomicfoundation/hardhat-ignition-ethers";
 ```
 
 :::
@@ -83,7 +83,7 @@ import "@nomicfoundation/hardhat-ignition";
 :::tab{value="JavaScript"}
 
 ```javascript
-require("@nomicfoundation/hardhat-ignition");
+require("@nomicfoundation/hardhat-ignition-ethers");
 ```
 
 :::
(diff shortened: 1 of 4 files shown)


docs/src/content/ignition/docs/getting-started/index.md
```diff
@@ -32,30 +32,36 @@ To install Hardhat Ignition in an existing Hardhat project, you will need:
...
-npm install --save-dev @nomicfoundation/hardhat-ignition
+npm install --save-dev @nomicfoundation/hardhat-ignition-ethers
 ```
 
 :::
 
 :::tab{value="npm 6"}
 
 ```sh
-npm install --save-dev @nomicfoundation/hardhat-ignition
```
alcuadrado, line 50: This should include the peer dependencies (transitively all of them).

I think this may be present in the other guides too.

docs/src/content/ignition/docs/guides/viem.md
```diff
@@ -0,0 +1,149 @@
...
+
+Only **one** Hardhat Ignition package should be imported within your Hardhat config file. Viem and ethers support cannot both be enabled at the same time.
+
+:::
+
+## The Ignition object
+
+The `@nomicfoundation/hardhat-plugin-viem` plugin adds an `ignition` object to the [Hardhat Runtime Environment](../../../hardhat-runner/docs/advanced/hardhat-runtime-environment.md).
+
+The `ignition` object exposes a `deploy` method, which takes an Ignition module, deploys it against the current network (in tests this will be the in-memory Hardhat network), and returns the results of the module as typed Viem contract instances.
```
alcuadrado, line 78:
```suggestion
The `ignition` object exposes a `deploy` method, which takes an Ignition module, deploys it against the current network, like Hardhat Network, and returns the results of the module as typed Viem contract instances.
```

I corrected this as it wasn't always right.

docs/src/content/ignition/docs/guides/viem.md
```diff
@@ -0,0 +1,149 @@
...
+
+```typescript
+const [first, second] = await hre.viem.getWalletClients();
+
+const result = await hre.ignition.deploy(ApolloModule, {
+  defaultSender: second.account.address,
+});
+```
+
+---
```
alcuadrado, line 145: Let's add a title instead of `---` here.

docs/src/content/ignition/docs/guides/viem.md
```diff
@@ -0,0 +1,149 @@
...
+
+:::tip
+
+Typescript doesn't support importing JSON as const, it will instead automatically broaden the type, breaking Viem's inference. The artifact must be defined inline within Typescript.
+
+:::
+
+See [Viem's type inference requirements](https://viem.sh/docs/typescript.html#type-inference) for more details.
+
+## Sending transactions with a different account
```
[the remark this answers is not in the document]
alcuadrado: Let's add a :::tip in the general testing guide, and create an issue to revisit the testing docs so that it is clearer that there are two "flavors"

docs/src/content/ignition/docs/guides/viem.md
kanej: I switched it to `Hardhat Ignition and Hardhat Viem`.

docs/src/content/ignition/docs/getting-started/index.md
kanej: I have updated this npm 6/yarn call to include the transitive peer dependencies of `@nomicfoundation/hardhat-ignition-ethers` and done the same for `@nomicfoundation/hardhat-ignition-viem` in the viem guide.

I will do a double check of the manual install for them in the morning as part manually testing the release branch.

docs/src/content/ignition/docs/guides/viem.md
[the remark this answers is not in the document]
kanej: To clarify, sending transactions with a different account is the same in both the ethers and viem variants (the input parameters are the same across both versions, it is only the result that varies). I was suggesting moving the "Sending transactions with a different account" to the main testing section. Which on reflection, I am more convinced is the right move.

Can I check that I am following your comment correctly:

1. that we open an issue saying we should restructure the guides so that the current testing guide (assuming ethers) + viem stuck on as an extra, becomes, ethers testing guide + viem testing guide
2. But for the moment, add a tooltip to the general testing guide indicating there is a Viem option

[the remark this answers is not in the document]
alcuadrado: > that we open an issue saying we should restructure the guides so that the current testing guide (assuming ethers) + viem stuck on as an extra, becomes, ethers testing guide + viem testing guide

Yes. Or maybe have two entirely different testing guides. We have to think about it.

> But for the moment, add a tooltip to the general testing guide indicating there is a Viem option

correct.

And given that situation, I don't think the duplication of the "sending transactions with a different account" is too bad.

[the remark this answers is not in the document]
kanej: I have added the tooltip. I also duplicated the "Sending transactions with a different account" in the general test guide and made it work with ethers.

alcuadrado (approved):

PR #4667, opened 2023-12-11
ci: enable `eth_getTransactionByHash` tests

agostbiro:
- Enable `eth_getTransactionByHash` Hardhat integration tests for the EDR provider.
- Had to expose the EthereumJS "common" config from the EDR provider as well as tests expect it.
- Surfaced an error where the EDR forked provider wasn't using the chain id from the config, but instead defaulted to the remote chain id. This is fixed now.


crates/edr_eth/src/transaction/request.rs
```diff
@@ -113,6 +113,7 @@ impl TransactionRequest {
 }
 
 /// A transaction request and the sender's address.
+#[derive(Debug, Clone)]
```
Wodann, line 116:
```suggestion
#[derive(Clone, Debug)]
```

crates/edr_evm/src/blockchain/forked.rs
```diff
@@ -57,7 +57,8 @@ pub struct ForkedBlockchain {
     remote: RemoteBlockchain<Arc<dyn SyncBlock<Error = BlockchainError>>, true>,
     state_root_generator: Arc<Mutex<RandomHashGenerator>>,
     fork_block_number: u64,
-    chain_id: u64,
+    chain_id_override: Option<u64>,
```
Wodann, line 60: This change is being done because in Hardhat tests we are comparing the returned value of `chainId` opcode with the provided chain ID for the test providers, instead of the actual chain ID of a remote provider (for forks).

I considered this to be a bug in the Hardhat tests, as the `chain_id` opcode  in `revm` should be working with the actual chain ID as reported by the remote provider, for forks.

I fixed it as follows: https://github.com/NomicFoundation/hardhat/pull/4622/files#r1416539300

Feedback from @alcuadrado & @fvictorio would be appreciated.

Wodann (changes requested):

crates/edr_evm/src/blockchain/forked.rs
Wodann: I discussed with Pato and the approach you took is correct. In HH v2 it should be possible to override the `chain_id`, originally designed for security purposes. This behaviour is subject to change in HH v3.

Wodann: One improvement that you can still do is to only store the result of `chain_id_override.unwrap_or(chain_id)`. There's no need to keep both.

crates/edr_eth/src/transaction/request.rs
agostbiro: The other struct in this file follows the pattern `Debug, Clone`.

Wodann: I only left one comment. I thought that was your work too 😅

Wodann (approved):

PR #4669, opened 2023-12-12
Update Standard JSON Input Link

pcaversaccio:
As title. I wanted to understand Hardat's default values for the Solidity settings and realised it's still pointing to an old version that doesn't entail attributes like `viaIR` etc.


alcuadrado:
Thank you sr 🫡

alcuadrado (approved):

PR #4681, opened 2023-12-14
Add support for solc 0.8.23

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/4587)


schaable (commented):
are we still using 0.8.19 in the docs?

packages/hardhat-core/test/builtin-tasks/compile.ts
```diff
@@ -46,6 +48,41 @@ describe("compile task", function () {
...
+  describe("compile with latest solc version", function () {
+    // The 'hardhat.config.js' and 'A.sol' files need to be updated each time a new solc version is released
+
+    useFixtureProject("compilation-latest-solc-version");
+    useEnvironment();
+
+    it("should have the last version of solc in the 'hardhat.config.js' and 'A.sol' files", async function () {
+      // Test to check that the last version of solc is being tested
+      const userConfigSolcVersion =
+        HardhatContext.getHardhatContext().environment?.userConfig.solidity;
```
fvictorio, line 60: You should be able to replace `HardhatContext.getHardhatContext().environment` with just `this.env`

schaable (approved):

PR #4683, opened 2023-12-14
hardhat-verify: add apiUrl and browserUrl in sourcify options

marcocastignoli:

---

Allow to pass `apiUrl` and `browserUrl` to override the default sourcify server and repo urls

### DIFF
--- a/packages/hardhat-verify/src/internal/sourcify.ts
+++ b/packages/hardhat-verify/src/internal/sourcify.ts
@@ -14,10 +14,10 @@ import { ValidationResponse } from "./utilities";
 export class Sourcify {
   public apiUrl: string = "https://sourcify.dev/server";
   public browserUrl: string = "https://repo.sourcify.dev";
-  private _chainId: number;
+  public chainId: number;
 
   constructor(chainId: number, apiUrl?: string, browserUrl?: string) {
-    this._chainId = chainId;
+    this.chainId = chainId;
     if (apiUrl !== undefined) {
       this.apiUrl = apiUrl;
     }
@@ -30,7 +30,7 @@ export class Sourcify {
   public async isVerified(address: string) {
     const parameters = new URLSearchParams({
       addresses: address,
-      chainIds: `${this._chainId}`,
+      chainIds: `${this.chainId}`,
     });
 
     const url = new URL(`${this.apiUrl}/check-all-by-addresses`);
@@ -94,7 +94,7 @@ export class Sourcify {
     const parameters: any = {
       address,
       files,
-      chain: `${this._chainId}`,
+      chain: `${this.chainId}`,
     };
 
     if (chosenContract !== undefined) {
@@ -139,7 +139,7 @@ export class Sourcify {
       contractStatus === ContractStatus.PERFECT
         ? "full_match"
         : "partial_match";
-    return `${this.browserUrl}/contracts/${matchType}/${this._chainId}/${address}/`;
+    return `${this.browserUrl}/contracts/${matchType}/${this.chainId}/${address}/`;
   }
 }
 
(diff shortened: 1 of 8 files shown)


packages/hardhat-verify/src/internal/sourcify.ts
```diff
@@ -14,14 +14,23 @@ import { ValidationResponse } from "./utilities";
 export class Sourcify {
   public apiUrl: string = "https://sourcify.dev/server";
   public browserUrl: string = "https://repo.sourcify.dev";
+  private _chainId: number;
```
schaable, line 17: this change doesn't seem necessary for this PR. Please revert it.

marcocastignoli: The linter forced me to do it

schaable: I just tested it locally, and you should be able to keep `chainId` public without linting errors. Maybe it's a problem in your config?

image
image

marcocastignoli: No ok sorry, I thought you were referring to the underscore part. I reverted the change

packages/hardhat-verify/src/internal/sourcify.ts
```diff
@@ -14,8 +14,17 @@ import { ValidationResponse } from "./utilities";
...
-  constructor(public chainId: number) {}
+  constructor(chainId: number, apiUrl?: string, browserUrl?: string) {
+    this.chainId = chainId;
+    if (apiUrl !== undefined) {
+      this.apiUrl = apiUrl;
+    }
+    if (browserUrl !== undefined) {
+      this.browserUrl = browserUrl;
+    }
+  }
```
schaable, line 27: As `apiUrl` and `browserUrl` are passed as parameters to the constructor and have default values in the config, the initialization of the variables in lines 15 and 16, and the checks in 21 and 24 are not needed. And with these changes, the parameters in the constructor should be required, not optional. The entire section could be just:

```

export class Sourcify {
  constructor(
    public chainId: number,
    public apiUrl: string,
    public browserUrl: string
  ) {}
```

marcocastignoli: In the types I set them optional parameters

```
export interface SourcifyConfig {
  enabled: boolean;
  apiUrl?: string;
  browserUrl?: string;
}
```

Making them mandatory will force the user to set them in the hardhat config?

schaable: That is fine. You are going to make them required for the `constructor`, but you'll leave them optional in the types. If the user doesn't set them, we use the default value specified in the `defaultSourcifyConfig`:
```
config.sourcify = { ...defaultSourcifyConfig, ...userSourcifyConfig };
```

marcocastignoli: ok, done. I passed them using the non-null assertion operator, if the user will pass an empty string a network error will be presented.

schaable:
@marcocastignoli it seems like you have some tests failing. And you need to add a changeset. A patch with a message like "Add apiUrl and browserUrl to Sourcify configuration." should be enough. And finally, you should update the readme to let the users know how to set them in the config. Once we get all that done, we should be ready to merge this.

marcocastignoli:
Hey @schaable thanks for all the support for this PR. Now all the tests are passing 🙂

schaable (approved):

fvictorio:
Closing in favor of https://github.com/NomicFoundation/hardhat/pull/4693

I can't push to this branch for some reason. @marcocastignoli did you enable the  ["Allow edits from maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) option?

PR #4688, opened 2023-12-15
refactor: move `MethodInvocation` from `edr_eth` to `edr_provider`

agostbiro:
Move `MethodInvocation` from `edr_eth` to `edr_provider` as the second step of https://github.com/NomicFoundation/edr/issues/256


crates/edr_provider/src/requests.rs
```diff
@@ -2,9 +2,12 @@
 pub mod eth;
 /// Hardhat RPC request types
 pub mod hardhat;
+mod methods;
 mod validation;
 
-pub use edr_eth::remote::{client::Request as RpcRequest, methods::MethodInvocation as EthRequest};
+pub use edr_eth::remote::client::Request as RpcRequest;
+
+pub use crate::requests::methods::{MethodInvocation, OneUsizeOrTwo, U64OrUsize};
```
agostbiro, line 10: I removed the aliasing `MethodInvocation` to `EthRequest` as the public name will be `MethodInvocation` once its merged with `rpc_hardhat::Request` and the integration tests (the only external users) are using the `MethodInvocation` name.

crates/edr_provider/src/requests/methods.rs
```diff
@@ -29,10 +27,10 @@ mod optional_block_spec {
 #[serde(tag = "method", content = "params")]
 pub enum MethodInvocation {
     /// eth_accounts
-    #[serde(rename = "eth_accounts", with = "crate::serde::empty_params")]
+    #[serde(rename = "eth_accounts", with = "edr_eth::serde::empty_params")]
```
alcuadrado, line 30: This coming from `edr_eth` may be a bit weird, but I think it's fine for now. Maybe we could have a separate crate for base utilities like this, but not worth the effort now.

alcuadrado (approved):

agostbiro:
This failure is probably due to a new version of `types/node `:

https://github.com/NomicFoundation/hardhat/actions/runs/7245426373/job/19735432853?pr=4688

Going to ignore it here after discussion with @fvictorio

PR #4689, opened 2023-12-16
build: upgrade to revm 3.5.0

Wodann:


crates/edr_eth/src/receipt.rs
```diff
@@ -235,182 +256,234 @@ where
...
-                        if let Ok(status) = rlp.val_at(0) {
-                            TypedReceiptData::PostEip658Legacy {
-                                status: normalize_status(status),
-                            }
-                        } else {
-                            let state_root = rlp.val_at::<Vec<u8>>(0)?;
-                            TypedReceiptData::PreEip658Legacy {
-                                state_root: B256::from_slice(&state_root),
-                            }
+            println!("1");
```
agostbiro, line 275: Just marking this so we don't forget

crates/edr_eth/src/receipt.rs
```diff
@@ -235,182 +256,234 @@ where
...
+
+            let alloy_rlp::Header {
+                list,
+                payload_length,
+            } = alloy_rlp::Header::decode(buf)?;
+
+            println!("2");
```
agostbiro, line 282: Just marking this so we don't forget

crates/edr_eth/src/remote/block_spec.rs
```diff
@@ -320,11 +320,11 @@ mod tests {
         help_test_block_spec_serde(BlockSpec::Tag(BlockTag::Pending));
         help_test_block_spec_serde(BlockSpec::Tag(BlockTag::Safe));
         help_test_block_spec_serde(BlockSpec::Eip1898(Eip1898BlockSpec::Hash {
-            block_hash: B256::from_low_u64_ne(1),
+            block_hash: B256::random(),
```
agostbiro, line 323: What's the reason for changing to random here?

Wodann: There is no longer an API available for constructing from a `u64` directly, which means you first have to create a `U256` or `U160` (for `Address`) and then cast to the matching hash/byte type. I realised having a specific value here doesn't matter. We're merely testing the round trip, so any arbitrary value suffices. As such, I decided to use the `random` function.

crates/edr_napi/src/config.rs
```diff
@@ -193,7 +189,8 @@ impl TryFrom<ConfigOptions> for CfgEnv {
         cfg.chain_id = chain_id;
         cfg.spec_id = spec_id;
         cfg.limit_contract_code_size = limit_contract_code_size;
-        cfg.limit_initcode_size = limit_initcode_size;
+        // Question(agost): Did this end up being needed?
```
Wodann, line 192: @agostbiro

fvictorio (approved):

PR #4691, opened 2023-12-18
Added a guide for using upgradeable proxies with hardhat ignition

zoeyTM:
Follows along with the example added in [here](https://github.com/NomicFoundation/hardhat-ignition/pull/655)

### DIFF
--- a/docs/src/content/ignition/docs/guides/upgradeable-proxies.md
+++ b/docs/src/content/ignition/docs/guides/upgradeable-proxies.md
@@ -38,13 +38,19 @@ pragma solidity ^0.8.9;
 
 // A contrived example of a contract that can be upgraded
 contract DemoV2 {
+  string public name;
+
   function version() public pure returns (string memory) {
     return "2.0.0";
+  }
+
+  function setName(string memory _name) public {
+    name = _name;
   }
 }
 ```
 
-This contract is identical to the first one, except that it returns an updated version string.
+In addition to updating the version string, this contract also adds a `name` state variable and a `setName` function that allows us to set the value of `name`. We'll use this function later when we upgrade our proxy.
 
 Finally, we'll create a file called `Proxies.sol` to import our proxy contracts. This file will look a little different than the others:
 
@@ -184,7 +190,7 @@ module.exports = demoModule;
 
 ::::
 
-### Part 3: Upgrading our proxy
+### Part 3: Upgrading our proxy with an initialization function
 
 Next it's time to upgrade our proxy to a new version. To do this, we'll create a new file within our `ignition/modules` directory called `UpgradeModule.js` (or `UpgradeModule.ts` if you're using TypeScript). Inside this file, we'll again break up our module into two parts. To start, we'll write our `UpgradeModule`:
 
@@ -196,7 +202,11 @@ const upgradeModule = buildModule("UpgradeModule", (m) => {
 
   const demoV2 = m.contract("DemoV2");
 
-  m.call(proxyAdmin, "upgradeAndCall", [proxy, demoV2, "0x"], {
+  const encodedFunctionCall = m.encodeFunctionCall(demoV2, "setName", [
+    "Example Name",
+  ]);
+
+  m.call(proxyAdmin, "upgradeAndCall", [proxy, demoV2, encodedFunctionCall], {
     from: proxyAdminOwner,
   });
 
@@ -210,7 +220,9 @@ Next, we use the `m.useModule(...)` method to get the `ProxyAdmin` and proxy con
 
 Then, we deploy our `DemoV2` contract. This will be the contract that we'll upgrade our proxy to.
 
-Finally, we call the `upgradeAndCall` method on the `ProxyAdmin` contract. This method takes three arguments: the proxy contract, the new implementation contract, and a data parameter that can be used to call an additional function. We don't need it right now, so we'll leave it blank by setting it to an empty hex string (`"0x"`). We also provide the `from` option to ensure that the upgrade is called from the owner of the `ProxyAdmin` contract.
+Next, we encode a call to the `setName` function in the `DemoV2` contract. This function takes a single argument, a string, which we'll set to `"Example Name"`. This encoded function call will be used to call the `setName` function on the `DemoV2` contract when we upgrade the proxy.
+
+Finally, we call the `upgradeAndCall` method on the `ProxyAdmin` contract. This method takes three arguments: the proxy contract, the new implementation contract, and a data parameter that can be used to call an additional function on the target contract. In this case, we're calling the `setName` function on the `DemoV2` contract with the encoded function call we created earlier. We also provide the `from` option to ensure that the upgrade is called from the owner of the `ProxyAdmin` contract.
 
 Lastly, we again return the `ProxyAdmin` and proxy contracts so that we can use them in our next module.
 
@@ -289,6 +301,14 @@ describe("Demo Proxy", function () {
 
       expect(await demo.connect(otherAccount).version()).to.equal("2.0.0");
     });
+
+    it("Should have set the name during upgrade", async function () {
+      const [owner, otherAccount] = await ethers.getSigners();
+
+      const { demo } = await ignition.deploy(UpgradeModule);
+
+      expect(await demo.connect(otherAccount).name()).to.equal("Example Name");
+    });
   });
 });
 ```
@@ -322,6 +342,14 @@ describe("Demo Proxy", function () {
 
       expect(await demo.connect(otherAccount).version()).to.equal("2.0.0");
     });
+
+    it("Should have set the name during upgrade", async function () {
+      const [owner, otherAccount] = await ethers.getSigners();
+
+      const { demo } = await ignition.deploy(UpgradeModule);
+
+      expect(await demo.connect(otherAccount).name()).to.equal("Example Name");
+    });
   });
 });
 ```
@@ -330,7 +358,7 @@ describe("Demo Proxy", function () {
 
 ::::
 
-Here we use Hardhat Ignition to deploy our imported modules. First, we deploy our base `ProxyModule` that returns the first version of our `Demo` contract and test it to ensure the proxy worked and retrieves the appropriate version string. Then, we deploy our `UpgradeModule` that returns the second version of our `Demo` contract and test it to ensure the proxy now returns the updated version string.
+Here we use Hardhat Ignition to deploy our imported modules. First, we deploy our base `ProxyModule` that returns the first version of our `Demo` contract and test it to ensure the proxy worked and retrieves the appropriate version string. Then, we deploy our `UpgradeModule` that returns the second version of our `Demo` contract and test it to ensure the proxy now returns the updated version string. We also test that our initialization function was called and set the `name` state variable to `"Example Name"`.
 
 ## Further reading
 
--- a/docs/src/content/ignition/docs/guides/upgradeable-proxies.md
+++ b/docs/src/content/ignition/docs/guides/upgradeable-proxies.md
@@ -1,0 +1,282 @@
+# Using Hardhat Ignition with upgradeable proxies
+
+When developing smart contracts, you may decide to use an upgradeable proxy pattern to allow for future upgrades to your contracts. This guide will explain how to create Ignition modules to deploy and interact with your upgradeable proxy contracts.
+
+While there are several different proxy patterns, each with their own tradeoffs, this guide will focus on the [TransparentUpgradeableProxy](https://docs.openzeppelin.com/contracts/5.x/api/proxy#TransparentUpgradeableProxy) pattern. You can read more about upgradeable proxy patterns [on OpenZeppelin's blog](https://blog.openzeppelin.com/proxy-patterns).
+
+:::tip
+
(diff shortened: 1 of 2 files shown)


kanej:
This is a really good first pass. Thinking about the structure, I think we should move away from a straight description of the code towards:

* write initial module
* hook up to test
* modify deployment to upgrade the contract
* look the test code keeps the same structure (but the version changed)!

What do you think?

kanej:
Part of the structure of this is what do we think our advice would be best practice for upgrading open zeppelin contracts using Ignition?

The pattern in the code is nested modules where upgradeV2 uses deployV1; there is an extent to which I miss being able to express passing parameters:

```ts
const {proxy: proxyV1, proxyAdmin} = m.useModule(DeployV1, {})
const {proxy: proxyV2} = m.useModule(UpgradeToV2, {proxyAdmin})
...
const {proxy: proxyVX} = m.useMOdule(UpgradeToVX, {proxyAdmin})

return {proxy: proxyVX}
```

Maybe we can achieve the same thing internally within one module with plain functions, so something like:

```ts
const upgrade = (m: ModuleBuilder, proxyAdmin, proxyAdminOwner, contractName: string) => {
  const instance = m.contract(contractName);

  m.call(proxyAdmin, "upgradeAndCall", [proxy, demoV2, "0x"], {
    from: proxyAdminOwner,
  });

  const updatedProxy = m.contractAt(contractName, proxy);

  return updatedProxy;
}

cont module = buildModule((m) => {
  const proxyAdminOwner = m.getAccount(1);
  const {proxy: proxyV1, proxyAdmin} = m.useModule(DeployV1, {});

  upgrade(m, poxyAdmin, proxyAdminOwner, "DemoV2");
  // ...
  const proxyVX = upgrade(m, poxyAdmin, proxyAdminOwner, "DemoVX");

  return {proxy: proxyVX}
})
```

The key point being to make the "I need to upgrade the contract to the next version" problem as concise to express as possible.

zoeyTM:
@kanej 

> Part of the structure of this is what do we think our advice would be best practice for upgrading open zeppelin contracts using Ignition?

My opinion is that, long or even medium term, none of this will be our recommended best practice for upgradeable proxy patterns in Ignition. The reason for that is because I think we can do better, whether that comes in the form of an `UpgradeableProxyStrategy` or some other abstraction, similar to or inspired by OpenZeppelin's [hardhat-upgrades plugin](https://docs.openzeppelin.com/upgrades-plugins/1.x/hardhat-upgrades). 

My understanding was that this guide and the accompanying sample project were only meant to show how a user could leverage Ignition's current API to support their upgradeable proxy needs (mostly demonstrating how to use `m.contractAt(...)`, to be entirely honest) since a more complete solution is not in our immediate roadmap. 

Additionally, there are other upgradeable contract patterns out there, and individual implementations can vary widely. My goal was to demonstrate the specific parts of the API that allow for upgradeable proxies to work in Ignition today and to explain why they are used how they are to enable users to apply those tools and logic to their own upgradeable contract system, whatever that may look like.

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,282 @@
...
+First, we're getting our account that will own the ProxyAdmin contract. This account will not be able to interact with the proxy, but it will be able to upgrade it. In this case, we'll use the first account in our Hardhat accounts array.
+
+Next, we deploy our `Demo` contract. This will be the contract that we'll upgrade.
+
+Then, we deploy our `TransparentUpgradeableProxy` contract. This contract will be deployed with the `Demo` contract as its implementation, and the `proxyAdminOwner` account as its owner. The third argument is the initialization code, which we'll leave blank for now by setting to an empty hex string (`"0x"`).
+
+When we deploy the proxy, it will automatically create a new `ProxyAdmin` contract within its constructor. We'll need to get the address of this contract so that we can interact with it later. To do this, we'll use the `m.readEventArgument(...)` method to read the `newAdmin` argument from the `AdminChanged` event that is emitted when the proxy is deployed.
+
+Finally, we'll use the `m.contractAt(...)` method to get the `ProxyAdmin` contract at the address we retrieved from the event and return it along with the proxy.
+
```
alcuadrado, line 144: Let's explain the cast of proxy to demo here, so that this example is self-contained. That means, for example, that users can stop here and start testing their upgradeable demo

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,282 @@
...
+Next, we use the `m.useModule(...)` method to get the `ProxyAdmin` and proxy contracts from the previous module. This will ensure that the proxy is deployed before we try to upgrade it.
+
+Then, we deploy our `DemoV2` contract. This will be the contract that we'll upgrade our proxy to.
+
+Finally, we call the `upgradeAndCall` method on the `ProxyAdmin` contract. This method takes three arguments: the proxy contract, the new implementation contract, and a data parameter that can be used to call an additional function. We don't need it right now, so we'll leave it blank by setting it to an empty hex string (`"0x"`). We also provide the `from` option to ensure that the upgrade is called from the owner of the `ProxyAdmin` contract.
+
+Lastly, we again return the `ProxyAdmin` and proxy contracts so that we can use them in our next module.
+
+### Part 3: Interacting with our proxy
```
alcuadrado, line 175: Let's make this part 2

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,372 @@
...
+
+```solidity
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.9;
+
+import "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
+import "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol";
+```
+
+Because we're using the OpenZeppelin proxy contracts, we need to import them here to make sure Hardhat knows to compile them. This will ensure that their artifacts are available for Hardhat Ignition to use later when we're writing our Ignition modules.
```
kanej, line 65: Do we need to add a section here suggestion the installation of `@openzeppelin/contracts`?

```sh
npm install --save-dev @openzeppelin/contracts
```

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,372 @@
...
+
+Finally, we'll use the `m.contractAt(...)` method to tell Ignition to use the `ProxyAdmin` ABI for the contract at the address we just retrieved. This will allow us to interact with the `ProxyAdmin` contract when we upgrade our proxy.
+
+### Part 2: Interacting with our proxy
+
+Now that we have our proxy deployed, we're ready to interact with it. To do this, we'll create a new module called `DemoModule`:
+
+```js
+const demoModule = buildModule("DemoModule", (m) => {
+  const { proxy, proxyAdmin } = m.useModule(upgradeModule);
```
kanej, line 157:
```suggestion
  const { proxy, proxyAdmin } = m.useModule(proxyModule);
```

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,372 @@
...
+
+::::
+
+Let's break down what's happening here.
+
+First, we're getting our account that will own the ProxyAdmin contract. This account will not be able to interact with the proxy, but it will be able to upgrade it. In this case, we'll use the first account in our Hardhat accounts array.
+
+Next, we deploy our `Demo` contract. This will be the contract that we'll upgrade.
+
+Then, we deploy our `TransparentUpgradeableProxy` contract. This contract will be deployed with the `Demo` contract as its implementation, and the `proxyAdminOwner` account as its owner. The third argument is the initialization code, which we'll leave blank for now by setting to an empty hex string (`"0x"`).
```
kanej, line 145:
```suggestion
Then we deploy our `TransparentUpgradeableProxy` contract. This contract will be deployed with the `Demo` contract as its implementation, and the `proxyAdminOwner` account as its owner. The third argument is the initialization code, which we'll leave blank for now by setting to an empty hex string (`"0x"`).
```

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,372 @@
...
+
+When we deploy the proxy, it will automatically create a new `ProxyAdmin` contract within its constructor. We'll need to get the address of this contract so that we can interact with it later. To do this, we'll use the `m.readEventArgument(...)` method to read the `newAdmin` argument from the `AdminChanged` event that is emitted when the proxy is deployed.
+
+Finally, we'll use the `m.contractAt(...)` method to tell Ignition to use the `ProxyAdmin` ABI for the contract at the address we just retrieved. This will allow us to interact with the `ProxyAdmin` contract when we upgrade our proxy.
+
+### Part 2: Interacting with our proxy
+
+Now that we have our proxy deployed, we're ready to interact with it. To do this, we'll create a new module called `DemoModule`:
```
kanej, line 153:
```suggestion
Now that we have a module for deploying our proxy, we're ready to interact with it. To do this, we'll create a new module called `DemoModule`:
```

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,372 @@
...
+
+      expect(await demo.connect(otherAccount).version()).to.equal("1.0.0");
+    });
+  });
+
+  describe("Upgrading", function () {
+    it("Should have upgraded the proxy to DemoV2", async function () {
+      const [, otherAccount] = await ethers.getSigners();
+
+      const { demo } = await ignition.deploy(ProxyModule);
```
kanej, line 300:
```suggestion
      const { demo } = await ignition.deploy(UpgradeModule);
```

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,372 @@
...
+    });
+  });
+});
+```
+
+:::
+
+::::
+
+Here we use Hardhat Ignition to deploy our imported modules. First, we deploy our base `ProxyModule` that returns the first version of our `Demo` contract and test it to ensure the proxy worked and retrieves the appropriate version string. Then, we deploy our `UpgradeModule` that returns the second version of our `Demo` contract and test it to ensure the proxy now returns the updated version string. We also test that our initialization function was called and set the `name` state variable to `"Example Name"`.
```
kanej, line 361:
```suggestion
Here we use Hardhat Ignition to deploy our imported modules. First, we deploy our base `ProxyModule` that returns the first version of our `Demo` contract and tests it to ensure the proxy worked and retrieves the appropriate version string. Then, we deploy our `UpgradeModule` that returns an upgraded version of our `Demo` contract and tests it to ensure the proxy returns the updated version string. We also test that our initialization function was called, setting the `name` state variable to `"Example Name"`.
```

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,372 @@
+# Using Hardhat Ignition with upgradeable proxies
```
kanej, line 1:
```suggestion
# Upgradeable proxies
```
To avoid going over two lines on the navbar. What do we think?

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,372 @@
...
+
+Inside our `test` directory, we'll create a file called `ProxyDemo.js` (or `ProxyDemo.ts` if you're using TypeScript):
+
+::::tabsgroup{options="TypeScript,JavaScript"}
+
+:::tab{value="TypeScript"}
+
+```typescript
+import { expect } from "chai";
+
```
kanej, line 281:
```suggestion
import { ignition, ethers } from "hardhat";

```

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
zoeyTM: Hmm I was thinking not, because it's mostly meant as a guide to show how it'll work with whatever proxy you choose, but on second thought, might as well add it

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,397 @@
+# Upgradeable proxies
```
kanej, line 1:
```suggestion
# Upgradeable Contracts
```
What do we think? I know I proposed "Upgradeable Proxies", but I think contracts is more comprehensible than proxies.

docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -0,0 +1,397 @@
...
+yarn add @openzeppelin/contracts
+```
+
+:::
+
+::::
+
+## Getting to know our contracts
+
+Before we start writing our Ignition modules, let's take a look at the contracts we'll be deploying and interacting with.
```
kanej, line 39:
```suggestion
Let's take a look at the contracts we'll be deploying and interacting with.
```

The "Before we get started", is used in the previous section.

kanej (approved):
Pre-approving. My only remaining question is whether we prefer the title "Upgradeable Contracts" to "Upgradeable Proxies"

PR #4696, opened 2023-12-18
refactor: merge `edr_rpc_hardhat` into `edr_provider`

agostbiro:
Accomplishes the following steps for https://github.com/NomicFoundation/edr/issues/256:

-  Merge `edr_rpc_hardhat` into `edr_provider` and the Hardhat request type into MethodInvocation
- Use `MethodInvocation` directly instead of `Request` in `edr_provider`


crates/edr_test_utils/src/lib.rs
```diff
@@ -1,47 +1 @@
 pub mod env;
-
```
agostbiro, line 2: I move these into the `edr_provider` integration tests as they're only used there now.

crates/edr_provider/src/requests.rs
```diff
@@ -14,18 +14,7 @@ pub use crate::requests::methods::{MethodInvocation, OneUsizeOrTwo, U64OrUsize};
 #[serde(untagged)]
 pub enum ProviderRequest {
     /// A single JSON-RPC request
-    Single(Request),
+    Single(MethodInvocation),
```
Wodann, line 17: I thought this had been renamed to `..Request`?

crates/edr_provider/src/requests/hardhat/rpc_types.rs
```diff
@@ -0,0 +1,7 @@
+mod compiler;
```
Wodann, line 1: Would just `rpc` not suffice for the module name?

Wodann (commented):
A couple of questions. Otherwise LGTM

crates/edr_provider/src/requests.rs
agostbiro: I thought `Request` in itself might be a bit confusing, because it doesn't contain all the fields of a JSON-RPC request. Happy to rename it though

crates/edr_provider/src/requests/hardhat/rpc_types.rs
agostbiro: `rpc_types` is clearer to me, because `rpc` would be for me the entire `requests` module. Happy to rename it though

Wodann: Feel free to leave it

Wodann (approved):

crates/edr_provider/src/requests.rs
agostbiro: Ah ok. Yeah for the RPC client I named it `RequestMethod`, but for the RPC "server" `MethodInvocation` makes more sense to me. Happy to change it to `RequestMethod` though

PR #4706, opened 2023-12-20
test: fix and enable personal sign integration tests

agostbiro:
Fix and enable personal sign integration tests.


crates/edr_provider/src/requests/methods.rs
```diff
@@ -206,7 +206,7 @@ pub enum MethodInvocation {
     SendTransaction(EthTransactionRequest),
     /// eth_sign
     #[serde(rename = "eth_sign", alias = "personal_sign")]
-    Sign(Address, ZeroXPrefixedBytes),
+    Sign(ZeroXPrefixedBytes, Address),
```
agostbiro, line 209: The order of arguments was wrong

Wodann (approved):

Issue #4648, opened 2023-12-05
eth_call doesn't work with geth 1.13


nuliknol:
### Version of Hardhat

2.19.1

### What happened?

When sending eth_call to a contract for accessing read-only function with golang of Ethereum geth 1.13, the response from Hardhat is "unknown selector" error. 

The reason for this error is that Hardhat expect contract input provided in the field `data` , but ethereum geth 1.13 is sending it in field name `input`. 

This is RPC message sent with golang of geth 1.13.5

```
{"jsonrpc":"2.0","id":1,"method":"eth_call","params":[
    {   
        "from":"0x0000000000000000000000000000000000000000",
        "input":"0x8b1329e0",
        "to":"0x5fbdb2315678afecb367f032d93f642f64180aa3"
    },  
    "latest"
]}
```

and this is RPC message sent with hardhat `run` command:

```
{"jsonrpc":"2.0","method":"eth_call","params":[
    {   
        "from":"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266",
        "to":"0x5fbdb2315678afecb367f032d93f642f64180aa3",
        "data":"0x8b1329e0"
    },  
    "latest"
],"id":4}
```

As you can see Hardhat encodes the RPC data with `data` field, while geth uses `input` field. For this reason I am getting `unknown selector` error because Hardhat simply doesn't read contract input when decoding JSON text sent over HTTP

How to reproduce: 
1. Create a contract with read-only function
2. Download geth 1.13 and generate contract wrappers using `abigen` tool
3. Create a program in golang to send eth_call to the read-only function

The result is "unknow selector" error and execution of a fallback function

### Minimal reproduction steps

Hardhat expects contract input in field named `data` while Geth 1.12 or 1.13 uses field name `input` for sending contract's input.

this is RPC message sent with hardhat using `run` command:
```
{"jsonrpc":"2.0","method":"eth_call","params":[
    {   
        "from":"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266",
        "to":"0x5fbdb2315678afecb367f032d93f642f64180aa3",
        "data":"0x8b1329e0"
    },  
    "latest"
],"id":4}
```

and this is RPC message sent with golang of geth 1.12 or 1.13 ;

```
{"jsonrpc":"2.0","id":1,"method":"eth_call","params":[
    {   
        "from":"0x0000000000000000000000000000000000000000",
        "input":"0x8b1329e0",
[...]

PR #4718, opened 2023-12-26
Allow for input instead of data in eth_call and eth_sendTransaction

krzysztof-cywinski:

A hardhat node only accepts a `data` param in inbound `eth_call` and `eth_sendTransaction`. This behavior is not according to [spec defined here](https://ethereum.github.io/execution-apis/api-documentation/), so I'd file it as a bug.
Most RPCs I've tested so far support both `input` and `data`.

This behavior creates a problem when one tries to use a hardhat node together with go-ethereum, because the latter sends data as `input`.

This PR adds support for an `input` param while retaining support for the `data` param. If both `input` and `data` are provided `input` takes precedence, because it's aligned with the spec.

### DIFF
--- a/packages/hardhat-core/src/internal/core/jsonrpc/types/input/callRequest.ts
+++ b/packages/hardhat-core/src/internal/core/jsonrpc/types/input/callRequest.ts
@@ -20,6 +20,7 @@ export const rpcCallRequest = t.type(
     gasPrice: optionalOrNullable(rpcQuantity),
     value: optionalOrNullable(rpcQuantity),
     data: optionalOrNullable(rpcData),
+    input: optionalOrNullable(rpcData),
     accessList: optionalOrNullable(rpcAccessList),
     maxFeePerGas: optionalOrNullable(rpcQuantity),
     maxPriorityFeePerGas: optionalOrNullable(rpcQuantity),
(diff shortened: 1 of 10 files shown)


krzysztof-cywinski:
This PR should fix issue https://github.com/NomicFoundation/hardhat/issues/4648

xjhweb:
same p.b +1

xjhweb (approved):

xjhweb (approved):

krzysztof-cywinski:
@xjhweb any idea how to push this forward? what's the next step?

DarrenBangsund:
Any updates on this?

adasauce:
Applied these patches to my hardhat and it makes calls from modern ethclient code work again. 

Confirmed this fixes the issue.

IMO it makes sense to just handle it as input internally everywhere rather than having `input` and `data` optional fields be settable and putting the value into `data`. Then if you want to support passing the data parameter via jsonrpc still just have 1 transformation at the front end into the standardized format.

jchappelow:
Looks like this PR supersedes https://github.com/NomicFoundation/hardhat/pull/4438. @fvictorio does it address the issues in https://github.com/NomicFoundation/hardhat/pull/4438#pullrequestreview-1788349955?
The incompatibility with current go-ethereum client is a blocker and I don't know how to work around it without using a fork of one or the other.  Thanks.

jchappelow:
Oh, this was before the EDR refactor.  I guess the provider bits don't apply anymore and there are changes needed in rs code now.

PR #4744, opened 2024-01-08
Update final-thoughts.md

pucedoteth:

---

when you click on the toolbox link it takes you to not found page so i updated links.


fvictorio:
Hi @pucedoteth. I don't think this change is necessary, is it? The link seems to work fine in https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-toolbox Maybe it didn't work for you because you were reading that page in GitHub, but these docs are meant to be consumed from the website.

PR #4747, opened 2024-01-09
feat: port filters to EDR provider

Wodann:


crates/edr_eth/src/remote/cacheable_method_invocation.rs
```diff
@@ -339,6 +397,16 @@ impl WriteCacheKey {
         Self::Resolved(hasher.finalize())
     }
 
+    fn needs_range_check(hasher: Hasher, range: CacheableLogFilterRange<'_>) -> Option<Self> {
+        match range {
+            CacheableLogFilterRange::Hash(_) => Some(Self::finalize(hasher)),
+            CacheableLogFilterRange::Range { to_block, .. } => {
+                // TODO should we check that to < from?
+                Self::needs_safety_check(hasher, to_block)
```
agostbiro, line 405: I'm not sure what's the expected behavior if `to < from`, but to be safe, we could potentially check the safety of the max of `to` and `from`. Wdyt?

agostbiro (approved):
Thanks, Rust part LGTM! There is just one small thing that we could potentially improve where I left a comment, but feel free to merge without it as it was there prior.

agostbiro:
Sorry, one more thing: I think the `websocket` tests should be passing now as well, could you try to enable them and fix any potential issues?

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -438,6 +445,10 @@ export class EdrProviderWrapper
       };
     }
 
+    // To accomodate construction ordering, we need an adapter to listen to
+    // events from the provider's subscriber callback
```
fvictorio, line 449: I'm not sure I understand this comment.

fvictorio: Ok, I think I get it now. Maybe I would say something like "we need an adapter to forward events from the EdrProvider callback to the wrapper's listener". I'm not sure if that would be an improvement though. Anyway, feel free to ignore this.

fvictorio (approved):

crates/edr_eth/src/remote/cacheable_method_invocation.rs
Wodann: For Alchemy, if `to < from`, you get this error:

```rust
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"One of the blocks specified in filter (fromBlock, toBlock or blockHash) cannot be found."}}
```

so using the maximum of those two wouldn't resolve the issue, I think. We might just have to prohibit caching if this occurs?

Wodann:
> Sorry, one more thing: I think the `websocket` tests should be passing now as well, could you try to enable them and fix any potential issues?

The `"plain websocket"` tests indeed work now, but the `ethers.WebSocketProvider` tests don't yet. I'll leave this for a separate task, if you don't mind.

@fvictorio can you shine some light on how we'd want to implement `ethers.WebSocketProvider` in this case?

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
Wodann: Incorporated in https://github.com/NomicFoundation/hardhat/pull/4747/commits/10606d8809530f7b8f6023d595ffef07d13197c9

PR #4756, opened 2024-01-11
test: fix and enable hardhat network options tests

agostbiro:
Fix and enable Hardhat network options integration tests.


crates/edr_provider/src/error.rs
```diff
@@ -164,59 +164,73 @@ pub enum ProviderError {
...
+            ProviderError::TimestampEqualsPrevious { .. } => -32000,
+            ProviderError::TransactionFailed(_) => -32000,
+            ProviderError::TransactionCreationError(_) => -32000,
+            ProviderError::TryFromIntError(_) => -32000,
+            ProviderError::Unimplemented(_) => -32000,
+            ProviderError::UnknownAddress { .. } => -32000,
+            ProviderError::UnmetHardfork { .. } => -32602,
+            ProviderError::UnsupportedAccessListParameter { .. } => -32602,
+            ProviderError::UnsupportedEIP1559Parameters { .. } => -32602,
+        };
```
agostbiro, line 210: Didn't change any numbers, just factored out the `data` as it's only used in one variant.

Wodann (approved):

PR #4757, opened 2024-01-12
refactor: remove EDR_BINARY

Wodann:


agostbiro:
The code looks good, but I reran 5 times a test that looks like a temporary error and it still fails: https://github.com/NomicFoundation/hardhat/actions/runs/7497624491/job/20428604379?pr=4757

Could you have a look at that?

Wodann:
> The code looks good, but I reran 5 times a test that looks like a temporary error and it still fails: https://github.com/NomicFoundation/hardhat/actions/runs/7497624491/job/20428604379?pr=4757
> 
> Could you have a look at that?

None of the logic I changed should affect that particular test. I've seen the failure once before after we re-enabled the selfdestruct tests. It seems like it's an error in initialization code of the Typescript dual mode context, so I'd be inclined to finish the port of the node without fixing that particular issue. Once the port is done, we will be dropping all of that Typescript code anyway.

WDYT, @agostbiro ?

agostbiro:
> > The code looks good, but I reran 5 times a test that looks like a temporary error and it still fails: https://github.com/NomicFoundation/hardhat/actions/runs/7497624491/job/20428604379?pr=4757
> > Could you have a look at that?
> 
> None of the logic I changed should affect that particular test. I've seen the failure once before after we re-enabled the selfdestruct tests. It seems like it's an error in initialization code of the Typescript dual mode context, so I'd be inclined to finish the port of the node without fixing that particular issue. Once the port is done, we will be dropping all of that Typescript code anyway.
> 
> WDYT, @agostbiro ?

I looked into a bit and I can't seem to be able to reproduce it locally on MacOS, but I reran the CI again now and the same errors (both state root and time out) consistently occurs on Windows. This is a bit concerning, as it might indicate an actual problem on Windows. Could you try to reproduce it on Windows?

Alternatively, if you don't think it's a problem, I'd just disable these dual mode tests so the CI stays green.

Issue #4759, opened 2024-01-12
Fix e2e issues


fvictorio:
Our e2e tests have a several issues:

- The shebangs are wrong
- The temporary tests should be created outside the repository (for example, using `mktemp -d`, although I think `/tmp` is valid in the three runners used by the CI)
- `npx` is always used, but it should only be used for `npm`. `yarn` and `pnpm` should use their own versions (which is just the same command; that is: `npx hardhat`, `yarn hardhat` and `pnpm hardhat`)
- After fixing the shebangs, `bash` will be always used. This means that `echo` might not interpret `\n` (it doesn't do it in the CI nor on my machine), so we should use `printf` instead.
- We should add an assertion that no spurious, empty files are created in the root of the project. This function should work I think:

```bash
no_empty_files() {
  for file in $(find . -maxdepth 1 -type f); do
    if [ ! -s $file ]; then
      echo "File $file is empty"
      exit 1
    fi
  done
}
```
- Spend some time investigating why Windows e2e tests take way longer than Ubuntu e2e tests

When I fixed some of these things (not sure which one), the CI started failing for windows+pnpm, so we should look into that.

PR #4764, opened 2024-01-16
Merge main into edr/main

fvictorio:
I'm doing this before adding support for Cancun in `main`. I think syncing main and edr/main now should make things easier when we have to sync them again after finishing the Cancun effort.

No major merge conflicts here, so I think we should just merge if the CI passes.


agostbiro (approved):
Thanks!

fvictorio:
e2e tests are failing for some reason. For now I just disabled them for `edr/main`, because they have several problems (see https://github.com/NomicFoundation/hardhat/issues/4759) so I don't think it makes sense to spend time making them work for this branch right now.

PR #4768, opened 2024-01-17
added @dgma/hardhat-sol-bundler plugin

killroy192:
A new @dgma/hardhat-sol-bundler plugin has been added to the community plugins list. We use this plugin internally in Dogma Labs and would like to open-source it.

### DIFF
--- a/docs/src/content/hardhat-runner/plugins/plugins.ts
+++ b/docs/src/content/hardhat-runner/plugins/plugins.ts
@@ -867,10 +867,9 @@ const communityPlugins: IPlugin[] = [
     name: "@dgma/hardhat-sol-bundler",
     author: "Dogma Labs",
     authorUrl: "https://github.com/dgma",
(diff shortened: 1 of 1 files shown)


fvictorio:
Hi @killroy192, thanks for this!

Just in case, have you checked our [Ignition](https://hardhat.org/ignition/docs/getting-started) project?

killroy192:
> Hi @killroy192, thanks for this!
> 
> Just in case, have you checked our [Ignition](https://hardhat.org/ignition/docs/getting-started) project?

Hey, yeah I did, cool stuff! This solution differs from the "modular" Ignition approach since it avoids deployment script creation for each contract. The plugin requires the creation of the JS object declaratively describing what to deploy, that's it :) 
It also automatically manages redeployments, deploying only modified and new contracts. This is very helpful when you develop using a proxy.

PR #4769, opened 2024-01-17
feat: `eth_feeHistory`

agostbiro:
Ports `eth_feeHistory` to EDR provider and enables tests. 

I originally set out to implement this in a more idiomatic way, but it turns out that Hardhat includes the pending block in the results even if only the last mined block is requested. This prevents implementing fee history on the blockchain, so I just ended up doing a verbatim port with some hackery to expose the RPC client of the remote blockchain to the `ProviderData`.


agostbiro:
I realized there is still a todo left, so I set it to draft. Will fix tomorrow

Wodann:
Have you considered using `BlockchainWithPending` to model this?

agostbiro:
> Have you considered using `BlockchainWithPending` to model this?

Yeah my was original approach to add a `fee_history` method to the `Blockchain` trait and then implement `fee_history` in terms of local + remote for `ForkedBlockchain` and in terms of the wrapped one + the pending block for `BlockchainWithPending`. It turned out to be difficult to replicate the Hardhat logic around what blocks to include this way, so I reverted to the verbatim port.

PR is ready for review now 🙏

crates/edr_eth/src/remote/cacheable_method_invocation.rs
```diff
@@ -719,6 +751,21 @@ impl Hasher {
         Ok(this)
     }
 
+    fn hash_reward_percentile(self, value: &RewardPercentile) -> Self {
+        const RESOLUTION: f64 = 100.0;
```
Wodann, line 755: Does this mean we treat everything smaller than `0.01` as negligible? Can this introduce issues?

crates/edr_provider/src/data/gas.rs
```diff
@@ -144,3 +145,68 @@ fn min_difference(lower_bound: u64) -> u64 {
...
+                    gas_price.saturating_sub(base_fee_per_gas)
+                };
+
+            (gas_used, effective_reward)
+        })
+        .sorted_by(|(_, reward_first), (_, reward_second)| reward_first.cmp(reward_second))
+        .collect::<Vec<(_, _)>>();
+
+    // Ethereum block gas limit is 30 million, so it's safe to cast to f64.
+    let gas_limit = block.header().gas_limit as f64;
```
Wodann, line 184: Why not use `try_into()` with the comment in the `expect("")` such that this at least panics if the assumption is somehow invalidated?

Wodann (approved):
Overall looks good, but there are 2 small things worth looking at. Approving to expedite merging

crates/edr_eth/src/remote/cacheable_method_invocation.rs
agostbiro: Yeah, this means that we allow specifying reward percentiles with 0.01% granularity. I think this is fine, because Foundry caches with 0.5% granularity.

crates/edr_provider/src/data/gas.rs
agostbiro: Good idea, but `f64` doesn't implement `TryFrom<u64>` unfortunately.

agostbiro:
I'm merging this following the reason stated in this PR: https://github.com/NomicFoundation/hardhat/pull/4770#issuecomment-1900673641

PR #4770, opened 2024-01-18
feat: add `eth_signTypedData_v4` support

agostbiro:
Add `eth_signTypedData_v4` support and enable relevant Hardhat integration tests.


Wodann (approved):
The code looks good to me; nice and simple. However, there are failing tests due to panics in EDR. Could you solve those before merging please?

agostbiro:
I dug into this and the transaction failures are due to `maxFeePerBlobGas` is missing from a type `0x3` transaction. It's either a [Geth problem](https://github.com/ethereum/go-ethereum/issues/28841) or an Alchemy problem (reached out via Discord support), so I think this is safe to merge.

Wodann:
> I dug into this and the transaction failures are due to `maxFeePerBlobGas` is missing from a type `0x3` transaction. It's either a [Geth problem](https://github.com/ethereum/go-ethereum/issues/28841) or an Alchemy problem (reached out via Discord support), so I think this is safe to merge.

To avoid panics in the future, is there a way we can bubble up the error, such that the background thread doesn't crash?

agostbiro:
> > I dug into this and the transaction failures are due to `maxFeePerBlobGas` is missing from a type `0x3` transaction. It's either a [Geth problem](https://github.com/ethereum/go-ethereum/issues/28841) or an Alchemy problem (reached out via Discord support), so I think this is safe to merge.
> 
> To avoid panics in the future, is there a way we can bubble up the error, such that the background thread doesn't crash?

Done here: https://github.com/NomicFoundation/hardhat/pull/4775

Issue #4491, opened 2023-10-17
Race condition in compiler download when more than one Hardhat process is compiling


fvictorio:
Original discussions [here](https://github.com/NomicFoundation/hardhat/issues/3722) and [here](https://github.com/NomicFoundation/hardhat/issues/4382).

Reproduction: https://github.com/robercano/hardhat-concurrency-issue

This can happen in repos with workspaces that run a `build` script in several workspaces in parallel.

PR #4771, opened 2024-01-18
Race condition in compiler download when more than one Hardhat process is compiling

ChristopherDedominici:
[See issue here](https://github.com/NomicFoundation/hardhat/issues/4491)


packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
```diff
@@ -8,7 +8,7 @@ import { promisify } from "util";
 import { download } from "../../util/download";
 import { assertHardhatInvariant, HardhatError } from "../../core/errors";
 import { ERRORS } from "../../core/errors-list";
-import { Mutex } from "../../vendor/await-semaphore";
+import { MultiProcessMutex } from "../../vendor/await-semaphore";
```
fvictorio, line 11: We should rename this file since it's no longer a vendored implementation.

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
```diff
@@ -126,7 +126,10 @@ export class CompilerDownloader implements ICompilerDownloader {
   }
 
   public static defaultCompilerListCachePeriod = 3_600_00;
-  private readonly _mutex = new Mutex();
+  private readonly _compilerDownloadMultiProcessMutex = new MultiProcessMutex(
```
fvictorio, line 129: I think `_mutex` was fine. The `MultiProcess` part is a implementation information, it doesn't need to be in the variable name. And since this is the only mutex in the compiler downloader module, then `compilerDownload` can go too.

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
fvictorio: Ah, we are not removing the vendored code. In that case, we should create a separate file for the multi-process mutex.

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
```diff
@@ -2,6 +2,9 @@
 // which is distributed under this license: https://github.com/notenoughneon/await-semaphore/blob/f117a6b59324038c9e8ee04c70c328215a727812/LICENSE
 
 /* eslint-disable @nomicfoundation/hardhat-internal-rules/only-hardhat-error */
+import debug from "debug";
+import fs from "node:fs";
+import os from "os";
```
fvictorio, line 7: Why is `fs` imported with `node:` but `os` is not?

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
```diff
@@ -64,3 +67,154 @@ export class Mutex extends Semaphore {
     super(1);
   }
 }
+
+export class MultiProcessMutex {
+  // Logic explanation: the fs.writeFile function, when used with the wx+ flag, performs an atomic operation to create a file.
```
fvictorio, line 72: Why `wx+` and not just `wx`? Also, can we include a reference here? I don't see this mentioned in node docs.

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
```diff
@@ -64,3 +67,154 @@ export class Mutex extends Semaphore {
...
+  // If multiple processes try to create the same file simultaneously, only one will succeed.
+  // This logic can be utilized to implement a mutex.
+  // For more info check the Nomic Notion page (internal link).
+  // Timeout logic explanation:
+  // When the function passed to the mutex is executed, it must not take longer than _defaultTimeoutInMs.
+  // If it exceeds this duration, the function is terminated, and the mutex is released.
```
fvictorio, line 78: I'm not sure we should do this. If a download is taking a long time for a valid reason (say, many compilers are being downloaded in a slow network), then this will interrupt the process unnecessarily. The only reason to do this is to handle the case where a download hangs, but this results in the same scenario than a process crashing, which should be handled anyway.

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
```diff
@@ -64,3 +67,154 @@ export class Mutex extends Semaphore {
...
+  // In case the process running the function crashes, other processes waiting for the mutex can detect this
+  // by checking the elapsed time since the mutex was acquired. This approach helps to avoid deadlock scenarios.
```
fvictorio, line 80: Pedantic comment, but: AFAIK that's not a deadlock. It's just a stale lock file.

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
```diff
@@ -64,3 +67,154 @@ export class Mutex extends Semaphore {
...
+  // The _safeMarginInMs is employed here to ensure that waiting processes do not prematurely terminate the current mutex holder
+  // before it reports an error for exceeding the allowed execution time.
+  private _log: debug.Debugger;
+  private _mutexFilePath: string;
+  private _timeoutInMs: number;
+  private _defaultTimeoutInMs = 20000;
```
fvictorio, line 86: I think we should remove this (see comment about this above), but a personal-preference comment anyway: default values like this shouldn't be instance fields IMO; they are not part of the state of the class. We normally use file-level constants for this, or perhaps static fields.

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
```diff
@@ -64,3 +67,154 @@ export class Mutex extends Semaphore {
...
+  private _safeMarginInMs = 2000;
+  private _mutexReleaseWaitingTimeInMs = 500;
+
+  constructor(mutexName: string, timeoutInMs?: number) {
+    this._log = debug("hardhat:await-semaphore:multi-process-mutex");
+
+    this._log(`Creating mutex with name '${mutexName}'`);
+
+    this._mutexFilePath = `${os.tmpdir()}/${mutexName}.txt`;
```
fvictorio, line 95: Don't use `/` to concatenate files; that can fail on Windows. Always use `path.join` or `path.resolve`. We should probably have a linter rule for this.

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
```diff
@@ -64,3 +67,154 @@ export class Mutex extends Semaphore {
...
+        );
+      }, this._defaultTimeoutInMs);
+    });
+
+    try {
+      // Execute the function passed to the mutex, no other function will be able to run at the same time.
+      // If the function will take more than the timeout, it will be terminated with an error and the mutex will be released.
+      // Promise.race is used to race the timeoutChecker against the execution of fn().
+      // Whichever completes first will be the outcome of Promise.race.
+      return await Promise.race([f(), timeoutChecker]).then((result) => {
```
fvictorio, line 161: Is there a reason to use `.then` here instead of just doing `const result = await Promise.race(...)`?

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
```diff
@@ -64,3 +67,154 @@ export class Mutex extends Semaphore {
...
+        clearTimeout(timeoutHandle);
+
+        // Release the mutex
+        this._deleteMutexFile();
+
+        this._log(`Mutex released at path '${this._mutexFilePath}'`);
+
+        return result;
+      });
+    } catch (error: any) {
```
fvictorio, line 171: Perhaps this catch and the `deleteMutexFile` call in line 165 could be replaced just with a `finally { this._deleteMutexFile() }`? I'm not 100% sure, but I think it's equivalent and it would better convey the intended behavior.

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
```diff
@@ -64,3 +67,154 @@ export class Mutex extends Semaphore {
...
+      if (error.code === "ENOENT") {
+        // The file might have been deleted by another process while this function was trying to access it.
+        return false;
+      }
+
+      throw error;
+    }
+
+    const fileDate = new Date(fileStat.mtime);
```
fvictorio, line 194: Although for this particular case is not super relevant, in general you want to use `ctime` instead of `mtime`, because `mtime` is only about the file contents being changed, while `ctime` also includes metadata changes.

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
ChristopherDedominici: I moved the code for the multi-process-mutex in a different file

packages/hardhat-core/src/internal/solidity/compiler/downloader.ts
ChristopherDedominici: Done

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
ChristopherDedominici: It's quite easy to find ion the fs nodeJS docs: https://nodejs.org/api/fs.html

image

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
ChristopherDedominici: I'll remove this logic for now, as we agreed via private messages

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
ChristopherDedominici: Done

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
ChristopherDedominici: This logic has been removed, see previous comments

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
ChristopherDedominici: This logic has been removed, see previous comments

packages/hardhat-core/src/internal/vendor/await-semaphore/index.ts
ChristopherDedominici: Done

packages/hardhat-core/test/internal/multi-process-mutex.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const arr = [false];
+    await mutex.use(async () => {
+      arr[0] = true;
+    });
+
+    const end = performance.now();
+    const duration = end - start;
+
+    expect(arr[0] === true);
```
schaable, line 54:
```suggestion
    let res = false;
    await mutex.use(async () => {
      res = true;
    });

    const end = performance.now();
    const duration = end - start;

    expect(res === true);
```

schaable (approved):

packages/hardhat-core/test/internal/multi-process-mutex.ts
```diff
@@ -0,0 +1,86 @@
...
+      }),
+    ]);
+
+    const end = performance.now();
+    const duration = end - start;
+
+    expect(duration).to.be.greaterThan(ms[0] + ms[1] + ms[2]);
+  });
+
+  it("should overwrite an old mutex file", async () => {
```
fvictorio, line 36: This test is checking an implementation detail, and the important part of this (that an attempt to acquire the mutex won't have to wait until the function finished) is already covered by the next test. I think we should remove it.

packages/hardhat-core/test/internal/multi-process-mutex.ts
```diff
@@ -0,0 +1,86 @@
...
+            res.push(2);
+          });
+          resolve(true);
+        }, 200)
+      ),
+    ]);
+
+    expect(res).to.deep.equal([2, 1]);
+  });
+});
```
fvictorio, line 82: We should add a test in which the executed function fails.

ChristopherDedominici: Good catch 👍

fvictorio (approved):

PR #4775, opened 2024-01-19
fix: change panics to errors in forked blockchain

agostbiro:
I went over `expect`s in the forked blockchain implementation to change panics to errors to avoid panicking if the remote responds with unexpected data. 

I introduced a new `ForkedBlockchainError` error type as I didn't want to add errors that can only occur for a forked blockchain to the `BlockchainError` type. I also moved errors from `BlockchainError` that can only occur for a forked blockchain to `ForkedBlockchainError`.

### DIFF
--- a/crates/edr_evm/src/block.rs
+++ b/crates/edr_evm/src/block.rs
@@ -13,7 +13,7 @@ use edr_eth::{
 pub use self::{
     builder::{BlockBuilder, BlockBuilderCreationError, BlockTransactionError, BuildBlockResult},
     local::LocalBlock,
-    remote::RemoteBlock,
+    remote::{CreationError as RemoteBlockCreationError, RemoteBlock},
 };
 
 /// Trait for implementations of an Ethereum block.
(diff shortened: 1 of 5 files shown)


Wodann (approved):
Code looks good to me and it's easier to detect what's going wrong in the CI now; thanks!

A general note:

I noticed that there are several failures (also on Ubuntu) due to the mismatched state root in the Dual mode adapter.

I wonder whether it only occurs when we are using the back-off strategy due to a high CI workload.

agostbiro:
> Code looks good to me and it's easier to detect what's going wrong in the CI now; thanks!
> 
> A general note:
> 
> I noticed that there are several failures (also on Ubuntu) due to the mismatched state root in the Dual mode adapter.
> 
> I wonder whether it only occurs when we are using the back-off strategy due to a high CI workload.

Yeah, the timeout is probably caused by the back-off strategy I think which in turn causes the mismatched state root in dual mode.

Issue #4606, opened 2023-11-21
Found an error at the docs


Genesis3800:
### Version of Hardhat

2.19.1

### What happened?

Found what I believe to be an error at the docs:

The [Hardhat-Foundry](https://hardhat.org/hardhat-runner/docs/advanced/hardhat-and-foundry#integrating-with-foundry) has two main sections:
1. [Adding Foundry to a Hardhat project](https://hardhat.org/hardhat-runner/docs/advanced/hardhat-and-foundry#adding-foundry-to-a-hardhat-project)
2. [#Adding Hardhat to a Foundry project](https://hardhat.org/hardhat-runner/docs/advanced/hardhat-and-foundry#adding-hardhat-to-a-foundry-project)

The first section is fine, but the second step misses the steps to install a crucial dependency.
See a basic hardhat project initialized normally will have [hardhat-toolbox](https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-toolbox) installed by default.
But when you follow the steps in section number 2, the steps only include installing the hardhat-foundry plugin, not the Hardhat-toolbox plugin.
There needs to be an extra step documented that tells folks to install the toolbox plugin so that the most basic Hardhat test functionalities are supported by default.

I would love to make a PR if possible.

### Minimal reproduction steps

Follow the steps at this exact link:

[https://hardhat.org/hardhat-runner/docs/advanced/hardhat-and-foundry#adding-hardhat-to-a-foundry-project](https://hardhat.org/hardhat-runner/docs/advanced/hardhat-and-foundry#adding-hardhat-to-a-foundry-project)

fvictorio:
Makes sense, feel free to send a PR

Genesis3800:
Ty ser, doing it right now

Genesis3800:
Hey @fvictorio it seems like I can't set up a local environment, so I guess I can't push a PR. Sucks for me.

I did this:
1. Forked, cloned the fork.
2. Did not install anything in the parent directory and straight up just installed dependencies inside docs folder with `pnpm install`. Ran this with `pnpm dev`.
3. The homepage loads correctly, but when I try to navigate anywhere I get this image.

![image](https://github.com/NomicFoundation/hardhat/assets/56264430/5da8d4f3-b300-44a0-b116-b31a72f59056)

Genesis3800:
Hey @fvictorio , this is what I am using:

node -v: 18.17.1
npm -v: 10.2.0
[...]

PR #4782, opened 2024-01-23
Fixing hardhat-foundry combined setup page

Genesis3800:
Hey @fvictorio apologies for being so delayed, but I am pushing a PR to fix an error at the docs as acknowledged in issue #4606 .

I came across this issue while writing this article for the official foundry book, which is now merged:
https://book.getfoundry.sh/config/hardhat

The second section of the article inside Hardhat's docs should now work out of the box.


fvictorio:
Thanks!

Issue #4290, opened 2023-08-11
Make sure that solc default import callback is not being used


fvictorio:
The way we are using solc right now means that the default import callback is used, and this checks the file system if a file is not available in the virtual file system.

We don't want this behavior because it can lead to irreproducible builds.

The only way to disable this right now is to create an empty directory and use `--base-path ${tmpDir}`. In some future versions we might be able to use the flag suggested [here](https://github.com/ethereum/solidity/issues/14484).

`--base-path` is not available in all the solc versions we support though. We need to find the first version that supports it, and only add this parameter in those cases. This means that (really old) solc versions can run into the reproducibility problem, but it doesn't seem worth it to look for a workaround for those versions.

---

1) From versions >= 0.8.22: we can use the `--no-import-callback` flag

2) The first solc version to support _--base-path_ is 0.6.9.
It can be verified via this 2 commands:
- 0.6.9 (supported): 
`rm -rf solc-static-linux && wget https://github.com/ethereum/solidity/releases/download/v0.6.9/solc-static-linux && chmod +x solc-static-linux && ./solc-static-linux --help | grep base`
- 0.6.8 (not supported):
`rm -rf solc-static-linux && wget https://github.com/ethereum/solidity/releases/download/v0.6.8/solc-static-linux && chmod +x solc-static-linux && ./solc-static-linux --help | grep base`

3) For versions <= 0.6.8: do nothing, no option or flag is added

PR #4783, opened 2024-01-23
Make sure that solc default import callback is not being used

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/4290)

### DIFF
--- a/.changeset/chatty-ravens-decide.md
+++ b/.changeset/chatty-ravens-decide.md
@@ -1,5 +1,0 @@
----
-"hardhat": patch
----
-
-Added logic to make sure that solc default import callback is not being used
--- a/packages/hardhat-core/src/builtin-tasks/compile.ts
+++ b/packages/hardhat-core/src/builtin-tasks/compile.ts
@@ -685,7 +685,7 @@ subtask(TASK_COMPILE_SOLIDITY_RUN_SOLC)
       solcPath: string;
       solcVersion?: string;
     }) => {
-      if (solcVersion !== undefined && semver.valid(solcVersion)) {
+      if (solcVersion !== undefined && semver.valid(solcVersion) === null) {
         throw new HardhatError(ERRORS.ARGUMENTS.INVALID_VALUE_FOR_TYPE, {
           value: solcVersion,
           name: "solcVersion",
--- a/packages/hardhat-core/src/builtin-tasks/compile.ts
+++ b/packages/hardhat-core/src/builtin-tasks/compile.ts
@@ -745,8 +745,6 @@ subtask(TASK_COMPILE_SOLIDITY_COMPILE_SOLC)
           solcJsPath: solcBuild.compilerPath,
         });
       } else {
-        console.log(`---------------------solcVersion: ${solcVersion}`);
-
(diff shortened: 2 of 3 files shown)


schaable (approved):

.changeset/chatty-ravens-decide.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Added logic to make sure that solc default import callback is not being used
```
fvictorio, line 5: I don't think we need to include a changeset for this; there's no user-facing change, it's more of an internal quality improvement.

packages/hardhat-core/src/internal/solidity/compiler/index.ts
```diff
@@ -68,14 +71,39 @@ export class Compiler implements ICompiler {
...
+  constructor(private _pathToSolc: string, private _solcVersion?: string) {}
 
   public async compile(input: CompilerInput) {
+    const args = ["--standard-json"];
+
+    // Logic to make sure that solc default import callback is not being used.
+    // If solcVersion is not defined or <= 0.6.8, do not add extra args.
+    if (this._solcVersion !== undefined) {
+      if (semver.parse(this._solcVersion) === null) {
+        throw new HardhatError(ERRORS.ARGUMENTS.INVALID_VALUE_FOR_TYPE, {
```
fvictorio, line 83: This error belongs to a different layer, because it's about _tasks_ arguments. While right now the `_solcVersion` value comes directly from a task argument, this class shouldn't be aware of that. I see two possible solutions:

1. Replace this with an assertion error (perhaps using `assertHardhatInvariant`, even if this is not exactly an invariant).
2. Move this check to `TASK_COMPILE_SOLIDITY_RUN_SOLC` and just assume that the version is valid here.

I don't have a preference.

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -675,8 +675,16 @@ subtask(TASK_COMPILE_SOLIDITY_RUN_SOLC)
   .addParam("input", undefined, undefined, types.any)
   .addParam("solcPath", undefined, undefined, types.string)
   .setAction(
```
fvictorio, line 677: This is missing the new param definition in the task. I think it's something like:

```
.addOptionalParam("solcVersion", undefined, undefined, types.string)
```

packages/hardhat-core/src/builtin-tasks/compile.ts
```diff
@@ -684,6 +685,14 @@ subtask(TASK_COMPILE_SOLIDITY_RUN_SOLC)
       solcPath: string;
       solcVersion?: string;
     }) => {
+      if (solcVersion !== undefined && semver.parse(solcVersion) === null) {
```
fvictorio, line 688: Sorry, I should've commented this before too, but it's a minor readability change:

```suggestion
      if (solcVersion !== undefined && semver.valid(solcVersion)) {
```

fvictorio (approved):
Very minor comment, feel free to merge

PR #4785, opened 2024-01-23
Fix e2e issues

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/4759)

### DIFF
--- a/e2e/test-project-initialization.sh
+++ b/e2e/test-project-initialization.sh
@@ -38,7 +38,7 @@ mkdir $TESTS_DIR
 # store the path to hardhat-core so it can be used in the tmp folder
 HARDHAT_CORE_FOLDER_PATH="$(pwd)/../packages/hardhat-core"
 
-echo "[e2e] Starting e2e initialization tests in $TESTS_DIR\n\n"
+echo -e "[e2e] Starting e2e initialization tests in $TESTS_DIR\n\n"
 
 pkg_managers="npm pnpm yarn"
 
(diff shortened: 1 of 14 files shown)


e2e/test-project-initialization.sh
```diff
@@ -45,17 +58,28 @@ for pkg_manager in $pkg_managers; do
...
   # pkg_manager, javascript, cjs
   echo "[e2e] Testing: $pkg_manager, javascript, cjs"
   mkdir ${TESTS_DIR}/${pkg_manager}-javascript-cjs
   cd ${TESTS_DIR}/${pkg_manager}-javascript-cjs
   create_package_json
-  $pkg_manager add ../../../packages/hardhat-core/$HARDHAT_TGZ_FILE >/dev/null 2>&1
-  HARDHAT_CREATE_JAVASCRIPT_PROJECT_WITH_DEFAULTS=true npx hardhat init
-  npx hardhat compile
-  npx hardhat test
+  $pkg_manager add $HARDHAT_CORE_FOLDER_PATH/$HARDHAT_TGZ_FILE >/dev/null 2>&1
```
schaable, line 78: These lines are duped several times. Maybe it's worth it to move them to a fn?

fvictorio (approved):

e2e/test-project-initialization.sh
fvictorio: I'm ok with this duplication (for now), because bash is so bad for abstracting things, and I'd rather have more straightforward code. But I'm fine either way.

PR #4786, opened 2024-01-24
Fix slow imports workflow

schaable:

---


.github/workflows/test-slow-imports-rule.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Install
         run: pnpm install --frozen-lockfile --prefer-offline
       - name: Add slow import
-        run: echo 'import "lodash";' >> src/internal/constants.ts
+        run: echo 'import "lodash";' >> src/internal/core/config/config-resolution.ts
```
fvictorio, line 35: Let's keep the previous one too. The more files we use, the better.

packages/eslint-plugin-slow-imports/lib/rules/no-top-level-external-import.js
```diff
@@ -80,9 +82,20 @@ module.exports = {
...
+          modulePath,
+          filename,
+          context.settings
+        );
+
+        if (!absoluteModulePath) {
+          context.report({
+            node: originalNode,
+            messageId: "CANNOT_RESOLVE_MODULE",
+          });
```
fvictorio, line 95: Shouldn't the function return after this?

packages/hardhat-chai-matchers/src/internal/addressable.ts
```diff
@@ -1,4 +1,4 @@
-import { isAddress, isAddressable } from "ethers";
+import type EthersT from "ethers";
```
fvictorio, line 1: This has a significant impact in performance, I think we should add a changeset for it, at least for the chai matchers (the change in core also improves the loading performance, but not that much).

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -68,6 +68,8 @@ export function loadConfigAndTasks(
     showSolidityConfigWarnings: false,
   }
 ): { resolvedConfig: HardhatConfig; userConfig: HardhatUserConfig } {
+  const { validateConfig, validateResolvedConfig } =
+    require("./config-validation") as ConfigValidationT;
```
fvictorio, line 72: You can also do this and drop the top-level type import:

```ts
  const { validateConfig, validateResolvedConfig } =
    require("./config-validation") as typeof import("./config-validation");
```

I don't know why the same trick doesn't work for `ordinal` though.

(I don't have a strong preference, use whichever form you like.)

fvictorio: Ah, doing this also means that you don't need to define and export that `ConfigValidationT` interface, so yeah, let's do this change.

.github/workflows/test-slow-imports-rule.yml
schaable: Hmm, but wouldn't the first fail hide the rest of the violations?

packages/eslint-plugin-slow-imports/lib/rules/no-top-level-external-import.js
schaable: I think so. Good catch.

packages/hardhat-core/src/internal/core/config/config-loading.ts
schaable: I'm not a fan of that syntax but 👍

.changeset/curvy-cherries-beg.md
```diff
@@ -0,0 +1,8 @@
+---
+"@nomicfoundation/eslint-plugin-slow-imports": patch
```
fvictorio, line 2: This one shouldn't be included; we are not publishing it.

fvictorio (approved):
Minor comment, pre-approving

Issue #4543, opened 2023-10-30
Support Sourcify Verification via Programmatic Access


pcaversaccio:
### Describe the feature

For simple repro, see my verification script [here](https://github.com/pcaversaccio/createx/blob/4305ac2125f8b55e58793bf36f12d5f99e213f43/scripts/deploy.ts#L7-L20):

```ts
async function main() {
  const contract = await ethers.deployContract("CreateX");

  await contract.waitForDeployment();
  const contractAddress = await contract.getAddress();

  console.log("CreateX deployed to:", contractAddress);

  await delay(30000); // Wait for 30 seconds before verifying the contract

  await hre.run("verify:verify", {
    address: contractAddress,
  });
}
```

and Sourcify is enabled [here](https://github.com/pcaversaccio/createx/blob/4305ac2125f8b55e58793bf36f12d5f99e213f43/hardhat.config.ts#L459-L462):
```ts
  sourcify: {
    // Enable Sourcify verification by default
    enabled: true,
  },
```

If you would manually verify a contract, it looks as intended:

![image](https://github.com/NomicFoundation/hardhat/assets/25297591/25b27a47-6256-4339-8e79-d666ea078208)

However, using it programmatically doesn't verify it on Sourcify:

![image](https://github.com/NomicFoundation/hardhat/assets/25297591/b0730189-d88e-4fd3-9919-993523ab8a57)

Let me know if you can repro as well.

### Search terms

sourcify

pcaversaccio:
Any update on this? I would like to use this feature for my upcoming [`CreateX` deployments](https://github.com/pcaversaccio/createx/blob/main/scripts/deploy.ts#L29).

PR #4791, opened 2024-01-25
Add support for programmatic verification in Sourcify

schaable:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/4543)

### DIFF
--- a/packages/hardhat-verify/src/index.ts
+++ b/packages/hardhat-verify/src/index.ts
@@ -271,7 +271,7 @@ subtask(TASK_VERIFY_VERIFY)
   .setAction(
     async (
       { address, constructorArguments, libraries, contract }: VerifySubtaskArgs,
-      { run }
+      { run, config }
     ) => {
       // This can only happen if the subtask is invoked from within Hardhat by a user script or another task.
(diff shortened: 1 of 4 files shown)


packages/hardhat-verify/src/index.ts
```diff
@@ -282,11 +282,21 @@ subtask(TASK_VERIFY_VERIFY)
         throw new InvalidLibrariesError();
       }
 
-      await run(TASK_VERIFY_ETHERSCAN, {
-        address,
-        constructorArgsParams: constructorArguments,
-        libraries,
-        contract,
-      });
+      if (config.etherscan.enabled) {
```
schaable, line 285: If the Etherscan verification fails, it will prevent the Sourcify verification from happening. @fvictorio, do you think we should mimic the main task's behavior and throw an error at the end?

fvictorio: No, let's keep it this way. It's simpler, and I think that failing as soon as one of the providers fails makes more sense for programmatic verification.

fvictorio (approved):

PR #4795, opened 2024-01-26
Fix hardhat-ledger to work with hardhat console

area:

---

I am attempting to use `npx hardhat console` in conjunction with `hardhat-ledger`. When doing so, after sending any transaction, `hardhat console` exits. This behaviour is related, either unintentionally or not, to the [default value of `discardStdin`](https://github.com/sindresorhus/ora/issues/220). This PR changes the value used.

There is a slight behavioral change in the console, which is that, as you might expect, keyboard input while the spinner is running is not discarded. This feels more acceptable than exiting after every transaction.

It does not seem to me that additional tests are required here, but open to guidance on that front.

### DIFF
--- a/.changeset/poor-nails-walk.md
+++ b/.changeset/poor-nails-walk.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ledger": patch
+---
+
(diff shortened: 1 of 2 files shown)


fvictorio (approved):

fvictorio:
Great catch, thanks! I had no idea that `ora` did that.

PR #4805, opened 2024-01-30
test: enable remaining Hardhat tests

Wodann:
This removes `Sinon` from the interval mining tests, which was used to control the clock. As we don't offer this functionality in EDR, we're sleeping the thread instead.


agostbiro (approved):
Thanks, LGTM!

PR #4806, opened 2024-01-30
fix: transport layer RPC errors

agostbiro:
This PR fixes transport layer RPC errors when running `pnpm test:provider` locally by reducing the number of requests sent when initializing fork mode. The `reqwest` client won't reuse connections for parallel requests which caused Alchemy refusing connections. We get around this by sending fork mode initialization requests in batch mode and caching aggressively.

In addition to the main fix, this PR contains some smaller fixes:

- Makes chain id requests uncachable as intended.
- Determines the chain id from the RPC url if possible to avoid sending requests.
- Retries with backoff on connection refused errors.
- Changes the cache path to include the host to avoid storing invalid data in case the remote is forked chain which can happen with remotes running locally.
- Add tracing to the RPC client.

The cache will be invalidated so expect slower CI the first time.

When running locally with warm cache, `edr` mode `pnpm test:provider` completes in 20 minutes and `ethereumjs` mode completes in 30 minutes.


crates/edr_napi/src/context.rs
```diff
@@ -50,46 +49,17 @@ impl EdrContext {
...
-        let (flame_layer, guard) = {
-            let (flame_layer, guard) =
-                tracing_flame::FlameLayer::with_file("tracing.folded").unwrap();
-
-            let flame_layer = flame_layer.with_empty_samples(false);
-            (flame_layer, guard)
-        };
-
-        #[cfg(feature = "tracing")]
-        let subscriber = subscriber.with(flame_layer);
```
agostbiro, line 80: I removed the flame layer, because we don't have a singleton object to store it on currently and I didn't manage to make it work with as a static variable tonight.

Wodann: If I'm not mistaken, this singleton still exists on the Hardhat side. We might just have to move its construction to `provider.ts`, alongside the `EdrProviderWrapper`.

In my work removing all component-based implementations this actually keeps existing, as enabling of tracing needs to happen in a binary (i.e. a shared library or executable). I'd like to avoid using `lazy_static` for this as construction/deconstruction ordering is not guaranteed for static variables. By having the context, it's possible for users to incorporate it in their memory management as it should be done safely.

crates/edr_napi/src/provider.rs
```diff
@@ -30,6 +33,22 @@ impl Provider {
         logger_config: LoggerConfig,
         #[napi(ts_arg_type = "(event: SubscriptionEvent) => void")] subscriber_callback: JsFunction,
     ) -> napi::Result<JsObject> {
+        START.call_once(|| {
```
Wodann, line 36: Does this approach work? I'm confused based on your previous comment. It currently at least still doesn't disable when the tracing feature is off

Wodann (changes requested):

crates/edr_napi/src/context.rs
agostbiro: Ok makes sense. Are you ok with re-adding the flame tracer once the code structure on the JS side crystalized?

crates/edr_napi/src/provider.rs
agostbiro: Yeah this works to initialize the tracing config only once, we just can't store the flame drop guard as static. Based on my reading of the code, previously only the flame guard depended on the `tracing` feature in `edr_napi` and basic tracing was always enabled. Always enabling basic tracing is replicated here, but the flame guard is removed.

crates/edr_eth/src/remote/client.rs
```diff
@@ -679,8 +705,79 @@ impl RpcClient {
         .await
     }
 
+    /// Fetch the latest block number, chain id and network id in a batch call.
+    #[cfg_attr(feature = "tracing", tracing::instrument(level = "trace", skip(self)))]
+    pub async fn fetch_fork_metadata(&self) -> Result<ForkMetadata, RpcClientError> {
+        let mut inputs = vec![
+            RequestMethod::NetVersion(()),
+            RequestMethod::BlockNumber(()),
```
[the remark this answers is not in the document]
agostbiro: Yes that's a mistake thanks! I added caching for the block number later and forgot to remove it. The tests were passing, because we're always getting the chain id from the Alchemy url, so the third return value was never utilized.

Fixed in [96602d7](https://github.com/NomicFoundation/hardhat/pull/4806/commits/96602d7138b4e8c22e67eeb05e3ea79aedffc188)

Wodann (approved):

jwbda:
@Wodann @agostbiro 
Thank you guys for create the `edr`. It is really good in `npx hardhat test`.

And I'm trying to sort out the logic of `npx hardhat test` command. And met some question, I hope you guys would help if you have some time. Here is the things: 
- 1 . Run the command `npx hardhat test` 
- 2 . And code execution to [const provider = await Provider.withConfig(](https://github.com/NomicFoundation/hardhat/blob/a1a385c45c7d9b92b38ca599b0d096197833feff/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts#L232) of `packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts` file

And here is the question: 
1. How the code  [const provider = await Provider.withConfig(](https://github.com/NomicFoundation/hardhat/blob/a1a385c45c7d9b92b38ca599b0d096197833feff/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts#L232)  of `packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts` file run to code [static withConfig(context: EdrContext, config: ProviderConfig, loggerConfig: LoggerConfig, subscriberCallback: (event: SubscriptionEvent) => void): Promise<Provider>](https://github.com/NomicFoundation/hardhat/blob/a1a385c45c7d9b92b38ca599b0d096197833feff/crates/edr_napi/index.d.ts#L389) of `crates/edr_napi/index.d.ts` file ?
2. Where is the implementation of  `withConfig` method?
3. How does the file `index.js` and `index.d.ts` of `crates/edr_napi` generated? How can I reproduce them again if I delete them?
4. Last, How to let the breakpoint of  `Javascript Debug Terminal` of vscode come into `crates/edr_napi` to follow the logic of rust code?

And any reply would be appreciated, thanks in advance.

Issue #4776, opened 2024-01-19
Unrelated Etherscan verification message appears when verifying with a custom Sourcify instance


acuarica:
### Version of Hardhat

2.19.4

### What happened?

> It's not a bug in the sense that's something is not working properly. It is an unrelated error message which might be misleading to the user.

I'm working on the verification service for the Hedera network, which uses a custom instance of Sourcify. I wanted to validate that the `hardhat-verify` plugin works with our instance. It works well, but the `verify` task shows an unrelated _Etherscan_ error message. 

I created a new project, configure it with Hedera testnet, and deploy the sample contract. When I want to verify it with the `hardhat-verify` plugin, I got the following output
```console
$ npx hardhat verify --network testnet <CONTRACT_ADDR>
Successfully verified contract Lock on Sourcify.
https://repository-verify.hashscan.io/contracts/full_match/296/<CONTRACT_ADDR>/
hardhat-verify found one or more errors during the verification process:

Etherscan:
Trying to verify a contract in a network with chain id 296, but the plugin doesn't recognize it as a supported chain.

You can manually add support for it by following these instructions: https://hardhat.org/verify-custom-networks

To see the list of supported networks, run this command:

  npx hardhat verify --list-networks
```

I expected the first part of the output, but the Etherscan message seems unrelated.

### Minimal reproduction steps

Create a new Hardhat project, install its dependencies and use the following `hardhat.config.js`

```js
require("@nomicfoundation/hardhat-toolbox");

/** @type import('hardhat/config').HardhatUserConfig */
module.exports = {
  solidity: "0.8.19",
  networks: {
    testnet: {
      url: 'https://testnet.hashio.io/api',
      accounts: ["<your private key, go to https://portal.hedera.com/ to setup one>"],
      chainId: 296,
    },
  },
  sourcify: {
    enabled: true,
    apiUrl: "https://server-verify.hashscan.io",
    browserUrl: "https://repository-verify.hashscan.io",
  }
};
```

Then run

```console
npx hardhat run --network testnet scripts/deploy.js
npx hardhat verify --network testnet <CONTRACT_ADDR>
```

Verification in the custom Sourcify instance works well, but an unrelated Etherscan error message is displayed.

### Search terms

verify hardhat-verify hedera sourcify etherscan

fvictorio:
[...]

PR #4829, opened 2024-02-07
docs: minor - highlight "verify:sourcify" subtask, based on feedback …

bguiz:
…from @schaable

- Ref: https://github.com/NomicFoundation/hardhat/issues/4776#issuecomment-1917272500

---

### DIFF
--- a/packages/hardhat-verify/README.md
+++ b/packages/hardhat-verify/README.md
@@ -230,6 +230,16 @@ await hre.run("verify:verify", {
 
 If the verification is not successful, an error will be thrown.
 
+Note that the `"verify:verify"` subtask will *always* run the etherscan verifier, even when you have the following in your hardhat config file:
+
+```js
+etherscan: {
+  enabled: false,
+}
+```
+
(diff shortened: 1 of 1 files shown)


packages/hardhat-verify/README.md
```diff
@@ -230,6 +230,16 @@ await hre.run("verify:verify", {
 
 If the verification is not successful, an error will be thrown.
 
+Note that the `"verify:verify"` subtask will *always* run the etherscan verifier, even when you have the following in your hardhat config file:
```
schaable, line 233: As the last published version of `hardhat-verify` this would be incorrect, see [here](https://github.com/NomicFoundation/hardhat/blob/%40nomicfoundation/hardhat-verify%402.0.4/packages/hardhat-verify/src/index.ts#L264). You can verify both providers with the same task, and control which one is executed through the config.

bguiz: Ah ok, if so this PR is redundant. Closing!

PR #4837, opened 2024-02-08
Restore ARITHMETIC_UNDER_OR_OVERFLOW and mark as deprecated

tamtamchik:

---

The silent deletion of ARITHMETIC_UNDER_OR_OVERFLOW in https://github.com/NomicFoundation/hardhat/pull/4755 is causing a breakdown in test logic. As per the [documentation](https://hardhat.org/hardhat-chai-matchers/docs/overview), these codes can be used with the `.revertedWithPanic` matcher. However, this matcher is also designed to accept undefined values, which silently take any panic code. This leads to the test case expectation being spoiled.

So as not to spoil tests that users might already rely on, it's safer to keep the `ARITHMETIC_UNDER_OR_OVERFLOW` option until the next major release.


tamtamchik:
cc @fvictorio can you please review this?

PR #4854, opened 2024-02-13
Better different hardfork forking tests

fvictorio:
Depends on https://github.com/NomicFoundation/hardhat/pull/4843


packages/hardhat-core/src/internal/hardhat-network/provider/node.ts
```diff
@@ -2594,6 +2594,31 @@ Hardhat Network's forking functionality only works with blocks from at least spu
         );
       }
 
+      // If this VM is running without cancun, but the block has cancun fields,
+      // we remove them from the block
+      if (
+        !this.isCancunBlock(blockNumberOrPending) &&
+        blockContext.header.blobGasUsed !== undefined
```
schaable, line 2601: is `blobGasUsed` always available in blocks? shouldn't we use `parentBeaconBlockRoot`?

fvictorio: My understanding is that either three three fields are present, or none is.

ChristopherDedominici (approved):

PR #4855, opened 2024-02-13
Add support for solc 0.8.24

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/4839)


packages/hardhat-core/src/internal/solidity/compiler/solc-info.ts
```diff
@@ -60,6 +60,7 @@ const defaultEvmTargets: { [key: string]: string } = {
   "0.8.21": "shanghai",
   "0.8.22": "shanghai",
   "0.8.23": "shanghai",
+  "0.8.24": "shanghai",
```
schaable, line 63: shouldn't be `cancun`?

schaable: nvm, I thought they'd upgrade the default evm version but it's still `shanghai`:

image

schaable (approved):

PR #4864, opened 2024-02-15
Fix setStorageAt in untouched addresses

fvictorio:
The latest version of ethereumjs fails if you modify the storage of an account that doesn't exist yet. This only happens for `setStorageAt` but I added tests for the other setters too (nonce, balance and code).


alcuadrado:
Can we report this and any other issue we find to ethereumjs?

schaable (approved):

PR #4869, opened 2024-02-15
fix: remove `deny_unknown_fields` from RPC types

agostbiro:
Remove `deny_unknown_fields` from RPC types to match Hardhat behavior of accepting additional fields.


fvictorio (approved):

Wodann:
Does this resolve a bug or was it an observation?

agostbiro:
> Does this resolve a bug or was it an observation?

It'd resolve the bug, but it's not strictly necessary after: https://github.com/NomicFoundation/hardhat/pull/4868

So it's more of an observation to make sure we follow Hardhat behavior.

Wodann (approved):

Issue #4566, opened 2023-11-07
Support `Settings` (a.k.a. Input JSON Description) in `@nomiclabs/hardhat-vyper`


pcaversaccio:
### Describe the feature

The `@nomiclabs/hardhat-vyper` package only supports currently the setting of Vyper version(s): https://github.com/NomicFoundation/hardhat/blob/ef4a94dfc49585b031faf75d5ef49bc438d43091/packages/hardhat-vyper/src/types.ts#L3-L9

It would be very convenient to have an option to at least set the `evmVersion`, as since Vyper [`0.3.8`](https://github.com/vyperlang/vyper/releases/tag/v0.3.8) (latest version is [`0.3.10`](https://github.com/vyperlang/vyper/releases/tag/v0.3.10)) the `evmVersion` is set to `shanghai` and people might be interested to deploy on e.g. Optimism with these versions but can natively since `PUSH0` is not yet supported.

The full Input JSON Description can be found [here](https://docs.vyperlang.org/en/stable/compiling-a-contract.html#input-json-description):

```bash
{
    // Required: Source code language. Must be set to "Vyper".
    "language": "Vyper",
    // Required
    // Source codes given here will be compiled.
    "sources": {
        "contracts/foo.vy": {
            // Optional: keccak256 hash of the source file
            "keccak256": "0x234...",
            // Required: literal contents of the source file
            "content": "@external\ndef foo() -> bool:\n    return True"
        }
    },
    // Optional
    // Interfaces given here are made available for import by the sources
    // that are compiled. If the suffix is ".vy", the compiler will expect
    // a contract-as-interface using proper Vyper syntax. If the suffix is
    // "abi" the compiler will expect an ABI object.
    "interfaces": {
        "contracts/bar.vy": {
            "content": ""
        },
        "contracts/baz.json": {
            "abi": []
        }
    },
    // Optional
    "settings": {
        "evmVersion": "shanghai",  // EVM version to compile for. Can be istanbul, berlin, paris, shanghai (default) or cancun (experimental!).
        // optional, optimization mode
        // defaults to "gas". can be one of "gas", "codesize", "none",
        // false  and true (the last two are for backwards compatibility).
        "optimize": "gas",
        // optional, whether or not the bytecode should include Vyper's signature
        // defaults to true
        "bytecodeMetadata": true,
        // The following is used to select desired outputs based on file names.
        // File names are given as keys, a star as a file name matches all files.
[...]

PR #4872, opened 2024-02-15
Support Settings (a.k.a. Input JSON Description) in @nomiclabs/hardhat-vyper

ChristopherDedominici:
[See issue here](https://github.com/NomicFoundation/hardhat/issues/4566)

Added properties evmVersion and optimize.
Follow [this issue](https://github.com/NomicFoundation/hardhat/issues/4873) to track additional settings properties that we might consider adding.


packages/hardhat-vyper/README.md
```diff
@@ -58,12 +58,23 @@ module.exports = {
 };
 ```
 
-You can also configure multiple versions of the Vyper compiler:
+You can also configure multiple versions of the Vyper compiler together with the compiler settings properties evmVersion (supported starting from Vyper 0.3.8) and optimize (supported starting from Vyper 0.3.10).
```
schaable, line 61:
```suggestion
You can also configure multiple versions of the Vyper compiler, including the compiler settings `evmVersion` (available from Vyper 0.3.8) and `optimize` (available from Vyper 0.3.10).
```

packages/hardhat-vyper/test/tests.ts
```diff
@@ -44,6 +45,86 @@ describe("Vyper plugin", function () {
...
+
+    describe("compilation with wrong settings", function () {
+      useFixtureProject("compilation-with-vyper-settings-failure");
+      useEnvironment();
+
+      it("should fail the compilation, invalid settings", async function () {
+        // compiler version is set to 0.3.9, which does not support the setting 'optimize'
+        await expect(this.env.run(TASK_COMPILE)).to.be.rejectedWith(
+          Error,
+          /unrecognized arguments: --optimize/
```
schaable, line 69: can we also test evmVersion?

packages/hardhat-vyper/README.md
ChristopherDedominici: I changed it, but with a different phrasing

packages/hardhat-vyper/src/compiler.ts
```diff
@@ -7,10 +8,21 @@ export class Compiler {
    *
    * @param inputPaths array of paths to contracts to be compiled
    */
-  public async compile(inputPaths: string[]) {
+  public async compile(inputPaths: string[], settings?: VyperSettings) {
```
schaable, line 11: defaulting the `settings` parameter to `{}` would allow us to avoid the optional chaining operator (`?.`) after every usage of `settings` in the function.

packages/hardhat-vyper/src/index.ts
```diff
@@ -249,14 +254,26 @@ subtask(TASK_COMPILE_VYPER)
...
+          version,
+          settings,
+        ])
+      );
+
       const versionGroups: Record<string, ResolvedFile[]> = {};
       const unmatchedFiles: ResolvedFile[] = [];
 
       for (const file of resolvedFiles) {
+        const settings = versionsToSettings[file.content.versionPragma];
```
schaable, line 268: I'd change this to `const settings = versionsToSettings[file.content.versionPragma] || {};` to avoid this check: `...(settings !== undefined ? { settings } : {})`

packages/hardhat-vyper/src/index.ts
```diff
@@ -298,6 +314,8 @@ ${list}`
       }
 
       for (const [vyperVersion, files] of Object.entries(versionGroups)) {
+        const settings = versionsToSettings[vyperVersion];
```
schaable, line 317: same here

packages/hardhat-vyper/test/tests.ts
ChristopherDedominici: Done. 
FYI: this test is a bit of a hack. Basically is very tricky to see if the evmVersion and optimize settings are actually used during the compilation. So I use this error to check that the command called includes the evmVersion and optimize params.

schaable (approved):
LGTM

packages/hardhat-vyper/src/compiler.ts
```diff
@@ -7,10 +8,21 @@ export class Compiler {
...
-  public async compile(inputPaths: string[]) {
+  public async compile(inputPaths: string[], settings: VyperSettings = {}) {
     const output: string = await new Promise((resolve, reject) => {
+      let settingsStr =
+        settings.evmVersion !== undefined
+          ? `--evm-version ${settings.evmVersion} `
+          : "";
+      settingsStr +=
+        settings.optimize !== undefined
+          ? `--optimize ${String(settings.optimize)} `
```
fvictorio, line 19: Is this casting needed?

packages/hardhat-vyper/src/index.ts
```diff
@@ -249,14 +254,24 @@ subtask(TASK_COMPILE_VYPER)
...
       const unmatchedFiles: ResolvedFile[] = [];
 
       for (const file of resolvedFiles) {
         const hasChanged = vyperFilesCache.hasFileChanged(
           file.absolutePath,
           file.contentHash,
-          { version: file.content.versionPragma }
+          {
+            version: file.content.versionPragma,
+            settings: versionsToSettings[file.content.versionPragma] ?? {},
```
fvictorio, line 273: I might be missing something here, but this seems conceptually wrong. The keys in `versionsToSettings` are versions, but `versionPragma` is a version _range_.

In terms of observable behavior, what you want is this:

1. Some file has version pragma `^0.3.0`
2. That file gets compiled with version X (which satisfies that range)
3. When compiling again, if a) the version that will be used to compile that file changes, or b) the same version is used but the settings changed, then you re-compile.

packages/hardhat-vyper/src/types.ts
```diff
@@ -1,7 +1,13 @@
 export type VyperUserConfig = string | VyperConfig | MultiVyperConfig;
 
+export interface VyperSettings {
+  evmVersion?: string;
+  optimize?: "gas" | "codesize" | "none";
```
fvictorio, line 5: Perhaps this should be just a `string`. Otherwise, if vyper adds a new `optimize` mode, we would need to release a new version adding it. But OTOH  new modes are not added all the time, and using `string` would mean that wrong values would result in a (vyper) execution error instead of a typing error. So I don't know. What would you prefer @pcaversaccio?

fvictorio (changes requested):

packages/hardhat-vyper/src/types.ts
pcaversaccio: Hmm, so there is an issue here. For Vyper versions `<0.3.10`, the `optimize` property was a Boolean: https://docs.vyperlang.org/en/v0.3.9/compiling-a-contract.html#setting-the-target-evm-version (see the PR https://github.com/vyperlang/vyper/pull/3493)). So the type changes depending on the Vyper version (some pseudocode lol):

```
(if evmVersion < 0.3.10) ? (typeof(optimize) == bool) : (typeof(optimize) == string)
```

fvictorio: From what I see in the help output of vyper `0.3.9`, there is a `--no-optimize` flag, but there's no `--optimize` one. Does this mean that versions older than 0.3.10 use the optimizer by default? In that case, maybe the behavior should be something like this:

```js
if (typeof optimize === "boolean") {
  if (optimize) {
    if (version >= "0.3.10") {
      // throw? or perhaps we should interpret "true" as `--optimize gas`
    }
    // for older versions there's nothing to do, because the optimizer is enabled by default
  } else {
    if (version < "0.3.10") cmd += "--no-optimize"
    else cmd += "--optimize none"
  }
} else if (typeof optimize === "string") {
  cmd += `--optimize ${optimize}`
}
```

pcaversaccio: Right, the `--no-optimize` flag was added in version `0.3.1` and also the attribute `"optimize": true,` in the "Input JSON File" (see [here](https://github.com/vyperlang/vyper/pull/2522)). Before that Vyper had always enabled the optimiser and the same holds true until today, even though the default is `gas` now in the latest version. So taking your snippet I above, I would do the following:

```ts
if (typeof optimize === "boolean") {
  if (optimize) {
    if (version >= "0.3.10" || version < "0.3.1") {
      // throw? or perhaps we should interpret "true" as `--optimize gas`
    }
    // for older versions there's nothing to do, because the optimizer is enabled by default
  } else {
    if (version < "0.3.10") cmd += "--no-optimize"
    else cmd += "--optimize none"
  }
} else if (typeof optimize === "string") {
  cmd += `--optimize ${optimize}`
}
```

fvictorio:
I pushed a commit using `VyperPluginError` instead of plain errors. This is necessary to get proper error output in the CLI.

Before:

image

After:

image

fvictorio:
I'm trying several combinations of versions and settings. This is what I got:

- `0.3.0`
  - `optimize: true`: throws a proper error ✅
  - `optimize: false`: compiler error ❌
  - `optimize: "gas"`: throws a proper error ✅
- `0.3.1`
  - `optimize: true`: works ✅
  - `optimize: false`: works ✅
  - `optimize: "gas"`: throws a proper error ✅
- `0.3.10`
  - `optimize: true`: throws a proper error ✅
  - `optimize: false`: works ✅
  - `optimize: "gas"`: works ✅

So mostly good. We should just fix that combination that produces a compiler error.

fvictorio (commented):
LGTM. Just some minor comments. I'll let @kanej decide if he wants to review it or if @schaable's previous code review is enough (my guess is that it is).

packages/hardhat-vyper/src/compiler.ts
```diff
@@ -28,3 +39,72 @@ export class Compiler {
...
+): string {
+  if (compilerVersion === "" && optimize !== undefined) {
+    throw new VyperPluginError(
+      "The 'compilerVersion' parameter must be set when the setting 'optimize' is set."
+    );
+  }
+
+  if (optimize === undefined) {
+    return "";
+  }
```
schaable, line 69:
```suggestion
  if (optimize === undefined) {
    return "";
  }
  
  if (compilerVersion === "") {
    throw new VyperPluginError(
      "The 'compilerVersion' parameter must be set when the setting 'optimize' is set."
    );
  }
```

packages/hardhat-vyper/src/compiler.ts
```diff
@@ -28,3 +39,72 @@ export class Compiler {
...
+
+  if (typeof optimize === "boolean") {
+    if (optimize) {
+      if (
+        semver.gte(compilerVersion, "0.3.10") ||
+        semver.lte(compilerVersion, "0.3.0")
+      ) {
+        throw new VyperPluginError(
+          `The 'optimize' setting with value 'true' is not supported for versions of the Vyper compiler older than or equal to 0.3.0 or newer than or equal to 0.3.10. You are currently using version ${compilerVersion}.`
```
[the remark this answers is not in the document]
schaable: (same for the rest of the error messages)

packages/hardhat-vyper/src/compiler.ts
```diff
@@ -28,3 +39,72 @@ export class Compiler {
...
+      return `--optimize ${optimize}`;
+    }
+
+    throw new VyperPluginError(
+      `The 'optimize' setting, when specified as a string value, is available only starting from the Vyper compiler version 0.3.10. You are currently using version ${compilerVersion}.`
+    );
+  }
+
+  throw new VyperPluginError(
+    `The 'optimize' setting has an invalid type value. Type is: ${typeof optimize}.`
```
schaable, line 108:
```suggestion
    `The 'optimize' setting has an invalid type value: ${typeof optimize}. Type should be either string or boolean.`
```

schaable (approved):
Included a few minor suggestions, but none are merge blockers.

fvictorio (approved):

packages/hardhat-vyper/src/compiler.ts
```diff
@@ -7,10 +10,18 @@ export class Compiler {
    *
    * @param inputPaths array of paths to contracts to be compiled
```
kanej, line 11: We should add the new parameters.

kanej (approved):
I had one small comment on some missing param docs, but pre-approving. Looks good!

PR #4875, opened 2024-02-15
Upgrade hardhat-viem to support viem@2

schaable:
This PR represents the first part of adding support for `viem@2`. A subsequent PR will update `hardhat-toolbox-viem` and the dependencies in `packages/hardhat-core/src/internal/cli/project-creation.ts`.

[Link to second PR](https://github.com/NomicFoundation/hardhat/pull/4876).

### DIFF
--- a/packages/hardhat-viem/src/types.ts
+++ b/packages/hardhat-viem/src/types.ts
@@ -42,8 +42,10 @@ export type GetContractReturnType<
   TAbi extends viemT.Abi | readonly unknown[] = viemT.Abi
 > = viemT.GetContractReturnType<
   TAbi,
-  PublicClient,
-  WalletClient,
+  {
+    public: PublicClient;
+    wallet: WalletClient;
+  },
   viemT.Address
 >;
 
--- a/packages/hardhat-viem/src/types.ts
+++ b/packages/hardhat-viem/src/types.ts
@@ -13,13 +13,22 @@ export type TestClient = viemT.TestClient<
   viemT.Chain
 >;
 
+export type KeyedClient =
+  | {
+      public?: PublicClient;
+      wallet: WalletClient;
+    }
+  | {
+      public: PublicClient;
+      wallet?: WalletClient;
+    };
+
 export type TestClientMode = Parameters<
   typeof viemT.createTestClient
 >[0]["mode"];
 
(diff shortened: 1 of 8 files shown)


kanej (approved):

ChristopherDedominici (approved):

fvictorio:
I think our own `hre.viem.getContractAt` should ideally follow the API of `viem.getContract`. That is, instead of:

```typescript
await hre.viem.getContract("Foo", "0x1234567890123456789012345678901234567890", {
  walletClient: secondWalletClient
};
```

it should become:

```typescript
await hre.viem.getContract("Foo", "0x1234567890123456789012345678901234567890", {
  client: {
    wallet: secondWalletClient
  }
}
```

Viem also accepts this:

```
await hre.viem.getContract("Foo", "0x1234567890123456789012345678901234567890", {
  client: extendedClient
}
```

But I tried to do it, and making the type-inference work was not straightforward. So I think we could start by only supporting the `client: { public, wallet }` version. Then adding support for `client: extendedClient` can be done without breaking changes.

packages/hardhat-viem/src/types.ts
```diff
@@ -34,18 +43,12 @@ export interface DeployContractConfig extends SendTransactionConfig {
...
 
 export interface GetContractAtConfig {
-  walletClient?: WalletClient;
-  publicClient?: PublicClient;
+  client?: KeyedClient;
 }
 
 export type GetContractReturnType<
   TAbi extends viemT.Abi | readonly unknown[] = viemT.Abi
-> = viemT.GetContractReturnType<
```
kanej, line 43: My understanding of this line is, previously we needed to re-export the `GetContractReturnType` to allow some abi shenanigans (which makes sense as we encode abi's as `any[]`), and to pass explicit `PublicClient` and `WalletClient` over the `viem` type. We did this because the type we get back should have both the public and wallet actions as we create default instances of both.
With `viem2` we change that to match the new viem KeyedType but with both populated. `Required<KeyedClient>` being the equivalent of `{public: PublicClient, wallet: WalletClient}`.

kanej (approved):
LGTM!

fvictorio (approved):

PR #4890, opened 2024-02-19
fix: use custom address recovery for fake signature

agostbiro:
Make sure we're not trying to recover the address from fake signatures. Fixes https://github.com/NomicFoundation/edr/issues/294

### DIFF
--- a/crates/edr_eth/src/transaction.rs
+++ b/crates/edr_eth/src/transaction.rs
@@ -4,6 +4,7 @@
 
 //! transaction related data
 
+mod fake_signature;
 mod kind;
 mod request;
 mod signed;
--- a/crates/edr_eth/src/transaction/fake_signature.rs
+++ b/crates/edr_eth/src/transaction/fake_signature.rs
@@ -26,6 +26,11 @@ pub(super) fn make_fake_signature<const V: usize>(sender: &Address) -> Signature
     Signature { r, s, v }
 }
 
+pub(super) fn recover_fake_signature(signature: &Signature) -> Address {
+    let address: [u8; 32] = signature.r.to_be_bytes();
+    Address::from_slice(&address.as_slice()[12..])
+}
+
 #[cfg(test)]
 pub(super) mod tests {
     macro_rules! test_fake_sign_properties {
@@ -59,6 +64,22 @@ pub(super) mod tests {
                 let hash_two = signed_transaction_two.hash();
 
                 assert_ne!(hash_one, hash_two);
+            }
+
+            #[test]
+            fn recovers_fake_sender() {
+                let transaction_request = dummy_request();
+
+                // Fails to recover with signature error if tried to ecrocver a fake signature
+                let sender: Address = "0x67091a7dd65bf4f1e95af0a479fbc782b61c129a"
+                    .parse()
+                    .expect("valid address");
+
+                let signed_transaction = transaction_request.fake_sign(&sender);
+
+                let recovered = signed_transaction.recover().expect("valid signature");
+
+                assert_eq!(recovered, sender);
             }
         };
     }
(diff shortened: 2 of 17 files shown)


Wodann (approved):
I think I would have solved this differently, but this fixes the issue for now.

The function `transaction_from_block` can return the `ExecutableTransaction` which includes the from-address, voiding the need for recovery.

Something we should wonder is whether the concept of faking a transaction should be incorporated in the raw transaction type? This PR does so now, which I'm not a fan of.
A limiting factor atm is that we don't have a way of expressing that an `Address` is a valid Ethereum address.

E.g. it's currently still possible that someone uses a JSON-RPC client with fake transactions for testing, which would result in the recovery of an invalid Ethereum address.

I'll approve to expedite the merging process, but if you agree with the above, I'd prefer a minimal change to use the `ExecutionTransaction` (which already contains the correct address) and to revisit this discussion when we deal with https://github.com/NomicFoundation/edr/issues/260

fvictorio:
> Something we should wonder is whether the concept of faking a transaction should be incorporated in the raw transaction type?

I don't think it should.

agostbiro:
Thanks for thinking deeply about this  @Wodann! I'm answering between the lines below.

> Something we should wonder is whether the concept of faking a transaction should be incorporated in the raw transaction type? This PR does so now, which I'm not a fan of.

Faking a transaction is already incorporated into the raw transaction type: https://github.com/NomicFoundation/hardhat/blob/f6eb9365fdc23033f41a8ff7744c9398c8e2459f/crates/edr_eth/src/transaction/request.rs#L103

The source of the bug is that this is not taken into account when doing recovery which this PR fixes.

> A limiting factor atm is that we don't have a way of expressing that an Address is a valid Ethereum address.

> E.g. it's currently still possible that someone uses a JSON-RPC client with fake transactions for testing, which would result in the recovery of an invalid Ethereum address.

I don't follow either, sorry. Please elaborate :)

> The function `transaction_from_block` can return the `ExecutableTransaction` which includes the from-address, voiding the need for recovery. https://github.com/NomicFoundation/hardhat/pull/4911 shows my intended solution.

That's a good idea, but it still leaves `SignedTransaction::recover` in a broken state, so this bug is bound to reoccur in the future unless we fix that.


> I'll approve to expedite the merging process, but if you agree with the above, I'd prefer a minimal change to use the ExecutionTransaction (which already contains the correct address) and to revisit this discussion when we deal with https://github.com/NomicFoundation/edr/issues/260

The right way to handle this imo would be to have two signed transaction types: `SignedTransaction` and `FakeSignedTransaction` (or possibly a `SignedTransaction<SignatureType>`), as that'd ensure at compile time (instead of the runtime check solution in this PR) that the recovery logic is appropriate for the signature method. 

But moving the check to compile time is a bigger change, hence the approach in this PR.

All in all, I'd prefer the solution in this PR over https://github.com/NomicFoundation/hardhat/pull/4911 as while the latter fixes the immediate problem, it doesn't prevent it from reoccurring.

Wodann:
> The right way to handle this imo would be to have two signed transaction types: SignedTransaction and FakeSignedTransaction (or possibly a SignedTransaction<SignatureType>), as that'd ensure at compile time (instead of the runtime check solution in this PR) that the recovery logic is appropriate for the signature method.

I agree with this. The `SignedTransaction<SignatureType>` would allow us to re-use as much of the data as possible.

> All in all, I'd prefer the solution in this PR over https://github.com/NomicFoundation/hardhat/pull/4911 as while the latter fixes the immediate problem, it doesn't prevent it from reoccurring.

It's unnecessary to use `recover()` given that we already know the `Address` when we call it. So at the very least we should merge both solutions; given that they solve different things.

PR #4911, opened 2024-02-21
fix: avoid recovering address from fake transactions

Wodann:


agostbiro (approved):
Thanks, as we discussed we want both: https://github.com/NomicFoundation/hardhat/pull/4890#issuecomment-1959713195

PR #4912, opened 2024-02-21
Viem 2 on ethereumjs main

kanej:
A PR to recreate the viem@2 changes on `ethereumjs/main`.

### DIFF
--- a/packages/hardhat-viem/package.json
+++ b/packages/hardhat-viem/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@nomicfoundation/hardhat-viem",
-  "version": "1.0.4",
+  "version": "2.0.0",
   "description": "Hardhat plugin for viem",
   "homepage": "https://github.com/nomicfoundation/hardhat/tree/main/packages/hardhat-viem",
   "repository": "github:nomicfoundation/hardhat",
--- a/packages/hardhat-viem/package.json
+++ b/packages/hardhat-viem/package.json
@@ -66,12 +66,12 @@
     "sinon": "^9.0.0",
     "ts-node": "^10.8.0",
     "typescript": "~5.0.0",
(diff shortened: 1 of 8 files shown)


packages/hardhat-viem/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nomicfoundation/hardhat-viem",
-  "version": "1.0.4",
+  "version": "2.0.0",
```
schaable, line 3: This is fine as long as it is for testing, but shouldn't be merged with the version bump, as `changeset` should take care of that.

schaable (approved):
Looks good but remember to revert the hardhat-viem version bump.

kanej:
> Looks good but remember to revert the hardhat-viem version bump.

This is being merged into `ethereumjs/main` with the intent of recreating the equivalent of the commits that have updated `hardhat-viem` on `main`.

On `main` the version of `hardhat-viem` is `2.0.0`:
https://github.com/NomicFoundation/hardhat/blob/5aab75287fe2a96d6d9fedf7c1db0df8dd7e4a5a/packages/hardhat-viem/package.json#L3

On `ethereumjs/main` it is `1.0.4`:
https://github.com/NomicFoundation/hardhat/blob/2a280a7aa88aa2882e1512a8ce8febda6e7fe819/packages/hardhat-viem/package.json#L3

Merging this will update `ethereumjs/main` to the same, without re-releasing. We will only use changeset for future releases, here we use the commit that changset created to recreate the effect i.e. https://github.com/NomicFoundation/hardhat/pull/4912/commits/6ed36543da829bf324a61f7f8a0a3b4f19aed61a

kanej:
This PR is no longer needed as Toolbox will ship based on `main` rather than `ethereumjs/main`.

PR #4928, opened 2024-02-22
feat: save and replay scenarios

agostbiro:
This PR adds support for saving and replaying scenarios. I tried to make it more efficient by saving the scenarios to disk gzipped, but this turned out to be impossible due to NAPI limitations.


crates/tools/src/scenario.rs
```diff
@@ -0,0 +1,174 @@
...
+    logger_enabled: bool,
+}
+
+pub async fn execute(scenario_path: &Path, max_count: Option<usize>) -> anyhow::Result<()> {
+    let (config, requests) = load_requests(scenario_path).await?;
+
+    if config.logger_enabled {
+        // anyhow::bail!("This scenario expects logging, but logging is not yet
+        // implemented")
+    }
```
Wodann, line 29: Is this old code?

Wodann (changes requested):
Some minor requests for changes but otherwise looks good

crates/tools/src/scenario.rs
agostbiro: Good catch thanks, that shouldn't be commented out. The scenario runner doesn't support logging yet (or maybe ever, idk how to do it relatively consistently), so I reenabled that line in [484e428](https://github.com/NomicFoundation/hardhat/pull/4928/commits/484e4288249abe91dcba4b213adb8b02762fc123)

Wodann: Why would there need to be special support for logging? It still executes the same RPC methods; it just writes logs to the CLI.

The scenario recorder doesn't have any guarantee about logging, right?

agostbiro: Not sure how comparable that'll be across runs. Probably a solvable issue just not in this PR :)

Wodann (approved):
Approving so we can merge. The last remaining issue can be tackled in a follow-up PR

Issue #4925, opened 2024-02-22
Support multiple test runners as part of HHv3


kanej:
After HH + EDR, fork `main` and split out the mocha test runner.
We want to add an abstraction that will allow plugins to swap in an alternative test runner.
A minimal alternative runner should be built based on the in-built `node` test runner (available since v20).
A plugin that encapsulates the current mocha test runner should also be created.

It is not a goal to resolve which test runner should be default in v3. It is not a goal to determine which assertion libraries or network helpers should be used.

## TODO

- [x] Pull together info on the most popular test runners, to aid our decision of default test runner
- [x] Create a test runner abstraction that is pluggable
- [x] Move the exiting mocha runner into that abstraction
- [x] Add a simple test runner based on the most popular test runner

kanej:
In particular we need to understand ESM and typescript better in the HH context.

PR #4938, opened 2024-02-27
Support multiple test runners as part of HHv3

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/4925)

This is a proof of concept. Only used to validate several concepts


alcuadrado:
Closing for book keeping. We'll get back to this later.

PR #4940, opened 2024-02-27
fix: use block builder to replay blocks

Wodann:
Previously, we were using the provider but the mem pool logic of the provider is not advanced enough to deal with blocks that transfer funds to an empty account and then subtract those funds, as it only takes the current state into consideration.


fvictorio:
The workflow wasn't working because the `$ALCHEMY_URL` token was being escaped. I tried inlining the secret in the command directly.

agostbiro (approved):

Issue #4844, opened 2024-02-09
Expose hardhat-ledger api to allow down stream plugins to control UI


kanej:
Ignition clashes with the `@nomicfoundation/hardhat-ledger` plugin in that both manipulate the cli UI.

The goal is to support a command line deploy from Ignition using a ledger, without clashing UI displays (Ignition rewrites the terminal).

## Potential Approach

Based on a quick look over the `hardhat-ledger` package, the UI manipulating elements come from a wrapper `withSpinners`. The ledger provider at the heart of the package acts as an event emitter, which `withSpinner` attaches to.

If `withSpinners` could be optionally disabled if operating within the context of an Ignition deploy run, and Ignition can hook into the events of the provider, then Ignition can fully control the UI display of the deploy while also customising for ledger interactions.

Open questions then are:

* How do we model "running in an Ignition deploy context" so that `withSpinners` can be disabled only in that situation?
* Can Ignition determine that it is dealing with a ledger provider (and should alter its UI behaviour appropriately)?

PR #4946, opened 2024-02-29
Add custom rpc method to toggle ledger feedback

schaable:
Note: This RPC method is intended for internal usage and may change without notice.

Usage:

```javascript
try {
  // Disable the default ledger output by calling `hardhat_setLedgerOutputEnabled`
  await hre.network.provider.send("hardhat_setLedgerOutputEnabled", [false]);

  // Subscribe to hardhat-ledger events and manually handle the output
  hre.network.provider.once("connection_start", () => console.log("IGNITION: [hardhat-ledger] Connecting wallet"));
  hre.network.provider.once("connection_success", () => console.log("[🗸] IGNITION: [hardhat-ledger] Connecting wallet"));
  hre.network.provider.once("connection_failure", () => console.log("[x] IGNITION: [hardhat-ledger] Connecting wallet"));
  // confirmation_start
  // confirmation_success
  // confirmation_failure
  // derivation_start
  // derivation_progress
  // derivation_success
  // derivation_failure
  
  // Placeholder for contract deployment logic

  // Re-enable the default ledger output after deployment
  await hre.network.provider.send("hardhat_setLedgerOutputEnabled", [true]);
} catch (error) {
  if (/Unsupported method/.test(error.message)) {
    // The current provider does not support the hardhat-ledger method
    console.error("The method hardhat_setLedgerOutputEnabled is not supported by the current provider.");
  }
}
```

Closes https://github.com/NomicFoundation/hardhat/issues/4844

### DIFF
--- a/packages/hardhat-ledger/src/internal/with-spinners.ts
+++ b/packages/hardhat-ledger/src/internal/with-spinners.ts
@@ -1,7 +1,11 @@
 import ora from "ora";
 import EventEmitter from "events";
 
-export function withSpinners<T extends EventEmitter>(emitter: T): T {
+export type OutputControlledEmitter = EventEmitter & {
+  isOutputEnabled: boolean;
+};
+
+export function withSpinners<T extends OutputControlledEmitter>(emitter: T): T {
   attachSpinner(emitter, {
     startText: "[hardhat-ledger] Connecting wallet",
     eventPrefix: "connection",
@@ -16,17 +20,17 @@ export function withSpinners<T extends EventEmitter>(emitter: T): T {
     startText: "[hardhat-ledger] Finding derivation path",
     eventPrefix: "derivation",
   });
-  emitter.on(
-    "derivation_progress",
-    (path: string, index: number) =>
-      (derivationSpinner.text = `[hardhat-ledger] Deriving address #${index} (path "${path}")`)
+  emitter.on("derivation_progress", (path: string, index: number) =>
+    emitter.isOutputEnabled
+      ? (derivationSpinner.text = `[hardhat-ledger] Deriving address #${index} (path "${path}")`)
+      : undefined
   );
 
   return emitter;
 }
 
 function attachSpinner(
-  emmiter: EventEmitter,
+  emitter: OutputControlledEmitter,
   spinnerOptions: {
     startText: string;
     eventPrefix: string;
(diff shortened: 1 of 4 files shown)


packages/hardhat-ledger/src/internal/with-spinners.ts
```diff
@@ -35,9 +39,15 @@ function attachSpinner(
   const { startText, eventPrefix } = spinnerOptions;
   const spinner = ora({ text: startText, discardStdin: false });
 
-  emmiter.on(`${eventPrefix}_start`, () => spinner.start());
-  emmiter.on(`${eventPrefix}_success`, () => spinner.succeed());
-  emmiter.on(`${eventPrefix}_failure`, () => spinner.fail());
+  emitter.on(`${eventPrefix}_start`, () =>
+    emitter.isOutputEnabled ? spinner.start() : undefined
```
kanej, line 43: No need to change.

Can I check. The use of the conditional operator here is equivalent to:

```ts
if (!emitter.isOutputEnabled) {
  return
}

spinner.start();
```

This is a style question, and not a blocker (nor even necessarily right 😉). When I read a conditional operator I assume it is returning a usable value. I also expect it to have no side-effects.

The functional programmer in me wants to be explicit about state updates and side-effects. So I typically go for obviously procedural `if` statements for side-effects (and for loops rather than maps/folds) and conditional operators and maps/folds for pure/transform code with no side effects.

This may be a convention I entirely made up, but I found the conditional operator here unexpected.

packages/hardhat-ledger/test/internal/with-spinners.ts
```diff
@@ -1,25 +1,27 @@
+import type { OutputControlledEmitter } from "../../src/internal/with-spinners";
+
 import { assert } from "chai";
 import sinon from "sinon";
 import EventEmitter from "events";
-import * as spinners from "../../src/internal/with-spinners";
+import { withSpinners } from "../../src/internal/with-spinners";
 
 describe("withSpinners", () => {
```
kanej, line 8: Is it hard to add tests for turning off and on events here?

packages/hardhat-ledger/src/internal/with-spinners.ts
schaable: Thanks for the insight. AFAIK, we don't have rules about ternary operators and side-effects. I see your point about being clear with state updates and effects. If it makes things clearer, I'm all for changing it.

packages/hardhat-ledger/test/internal/with-spinners.ts
schaable: I thought about it and I think it is. I'm not sure how can we test the spinner states.

schaable:
@kanej are we good to merge this?

kanej:
> @kanej are we good to merge this?

Sorry, I haven't got to this yet.

zoeyTM (approved):

kanej (approved):
I added a changeset and tested this manually:

1. Apply changeset
2. Install into local verdaccio
3. In a fresh HH project install the update version of `hardhat-ledger`
4. Ran through the usage example from the docs: https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-ledger#usage

PR #4949, opened 2024-03-01
Fix json-rpc provider error object

fvictorio:
Closes https://github.com/NomicFoundation/edr/issues/308.


crates/edr_provider/src/error.rs
```diff
@@ -314,31 +314,31 @@ pub struct TransactionFailure {
...
-        transaction_hash: &B256,
+        transaction_hash: Option<&B256>,
         solidity_trace: &Trace,
     ) -> Option<Self> {
         match execution_result {
             ExecutionResult::Success { .. } => None,
             ExecutionResult::Revert { output, .. } => Some(Self::revert(
                 output.clone(),
-                *transaction_hash,
+                transaction_hash.cloned(),
```
agostbiro, line 330: Small nit: `cloned` could be changed to `copied` here and on line 335 to make it more expressive, but it's not a blocker to merge for me, since the clone will copy anyway under the hood.

agostbiro (approved):
Thanks lgtm overall!

PR #4952, opened 2024-03-01
ci: add scenarios

agostbiro:
Adds the collected scenarios to the repo. The scenario files are [JSON Lines](https://jsonlines.org/) files compressed with GZIP at the highest compression level. The first line in each scenario file contains the provider config and subsequent lines contain RPC calls.

Tested that all the scenarios successfully completed with the Rust runner. The longest (`syntethix`) takes 15 minutes on my machine.

Also contains a fix to serialize storage values with padding for `hardhat_setStorageAt`. Otherwise they wouldn't be accepted by the deserializer implementation. This caused a failure originally in `openzeppelin-contracts`.


crates/edr_provider/src/requests/methods.rs
```diff
@@ -386,7 +386,9 @@ pub enum MethodInvocation {
     SetStorageAt(
         #[serde(deserialize_with = "crate::requests::serde::deserialize_address")] Address,
         #[serde(deserialize_with = "crate::requests::serde::deserialize_storage_key")] U256,
-        #[serde(deserialize_with = "crate::requests::serde::deserialize_storage_value")] U256,
+        #[serde(deserialize_with = "crate::requests::serde::deserialize_storage_value")]
```
Wodann, line 389: Are you aware of the `with` attribute that allows you to specify the deserialization and serialization functions in one go, using a `mod`?

agostbiro: Sure, I can change it if you want

agostbiro: Circling back to this: I'm fine with either option. There are no `mod`s yet in `edr_provider::requests::serde`, so for me this version is more idiomatic in this repo, but `mod` is shorter. Let me know if I should change it please @Wodann !

Wodann: My preference is the one I left in the comment above, yes, so if you could change it that would be great.

If I'm not mistaken, this is a result of your recent additions of the alternative (serialisation) methods and not a general pattern, as can be seen elsewhere in the repo (and the serde docs).

agostbiro: Ok cheers done in [dc94cd3](https://github.com/NomicFoundation/hardhat/pull/4952/commits/dc94cd32eaec8d6f53b278065ff66a186d717dfd)

Wodann (approved):

Issue #4790, opened 2024-01-25
Hardhat-ledger should support legacy derivation paths


area:
### Describe the feature

As part of the curse of the early adopter, users may have created addresses with Ledger prior to them fully implementing BIP44, when they used [a different derivation path](https://github.com/MyCryptoHQ/MyCrypto/issues/2070). Such users should be able to use `hardhat-ledger`.

One possibility to capture most of these users (myself included) would be to add a `ledgerLegacyDerivationPath` option as a boolean in the network configs, alongside `ledgerAccounts` - or possibly combining them in to a `ledgerOptions`.

A more general solution that would capture all possible issues would be to allow users to specify the derivation path themselves - e.g. `m/44'/60'/x'/0/0`. Again, either as a separate option or as part of a combined `ledgerOptions` property.

Of course, the default behaviour should remain unchanged - most users are not impacted by this.

I need this feature, and it is small, so would be keen to work on a PR myself. I am opening the discussion here to begin with, as requested in `CONTRIBUTING.md`, with the hope that someone can indicate which of these possibilities is most likely to get accepted.

fvictorio:
Thanks for opening this @area. I'm tempted to go with a function, since that would cover other use cases (although tbh I don't know how likely it is that there are use cases besides legacy paths):

```js
ledgerOptions: {
  derivationFunction = (accountNumber) => `44'/60'/${account}'/0/0`
}
```

What do you think?

cc @alcuadrado

area:
Yep, that would work for me. Making the option a function also makes it easier to adjust messages that output derivation paths than it would be if it were a string.

area:
I would still like to work on this if the desired implementation could be confirmed by whoever needs to :smile:

kanej:
@area you can assume that the function approach is the one you should take.

schaable:
The feature has been added and it will be released on the next `hardhat-ledger` version. You can read how it works [here](https://github.com/NomicFoundation/hardhat/tree/main/packages/hardhat-ledger#usage). Thanks @area!

kanej:
Support for derivation paths was released in `@nomicfoundation/hardhat-ledger@1.0.3`.

PR #4953, opened 2024-03-01
Legacy ledger derivation path support in hardhat-ledger

area:
- [x ] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.

---

First pass at feature discussed in #4790

### DIFF
--- a/packages/hardhat-ledger/src/type-extensions.ts
+++ b/packages/hardhat-ledger/src/type-extensions.ts
@@ -2,8 +2,8 @@ import "hardhat/types/config";
 
 declare module "hardhat/types/config" {
   interface HardhatNetworkUserConfig {
-    ledgerAccounts: string[];
-    ledgerOptions: {
+    ledgerAccounts?: string[];
+    ledgerOptions?: {
       derivationFunction?: (accountNumber: number) => string;
     };
   }
--- a/packages/hardhat-ledger/src/type-extensions.ts
+++ b/packages/hardhat-ledger/src/type-extensions.ts
@@ -2,27 +2,27 @@ import "hardhat/types/config";
 
 declare module "hardhat/types/config" {
   interface HardhatNetworkUserConfig {
+    ledgerAccounts: string[];
     ledgerOptions: {
-      accounts: string[];
       derivationFunction?: (accountNumber: number) => string;
     };
   }
   interface HardhatNetworkConfig {
-    ledgerOptions: {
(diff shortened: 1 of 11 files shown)


schaable:
Sorry @area. It seems there was a misunderstanding regarding the `ledgerOptions`. Moving `ledgerAccounts` into `ledgerOptions` would be a breaking change, which we should avoid. Instead, we should add a `ledgerOptions.derivationFunction` while keeping `ledgerAccounts` within the `networkConfig`. Could you update the PR reverting that change?

area:
Ah, I see. Makes sense, and was a misunderstanding on my part.

packages/hardhat-ledger/src/type-extensions.ts
```diff
@@ -2,16 +2,28 @@ import "hardhat/types/config";
 
 declare module "hardhat/types/config" {
   interface HardhatNetworkUserConfig {
-    ledgerAccounts?: string[];
```
schaable, line 5: this should still be optional

packages/hardhat-ledger/test/provider.ts
```diff
@@ -144,6 +144,7 @@ describe("LedgerProvider", () => {
         accounts,
         openTimeout: 1000,
         connectionTimeout: 5432,
+        derivationFunction: (index: number) => `m/44'/60'/${index}'/0/`,
```
schaable, line 147: Could you clarify if there's a need to specify a `derivationFunction` here? It seems like it might not be necessary, but I might be missing something.

schaable:
I've added some minor comments. Also, could you consider updating the readme explaining the new option?

packages/hardhat-ledger/test/provider.ts
area: I agree that it is not necessary, and I have removed it.

packages/hardhat-ledger/src/type-extensions.ts
area: Agreed. I've also made `ledgerOptions` optional here.

kanej (approved):
I did a round of manual testing on a new ledger, I was able to:
- run through the sign example from the `hardhat-ledger` docs
- update it with a derivation function using the default derivation path
- update with a failing derivation path and get a reported error
- run through an Ignition deploy

I don't have an old account for testing the legacy derivation path.

PR #4655, opened 2023-12-07
Add Hardhat Ignition to hardhat-toolbox

zoeyTM:
Updates the `@nomicfoundation/hardhat-toolbox` and `@nomicfoundation/hardhat-toolbox-viem` packages to include Hardhat Ignition.

This includes

- Updated sample projects with Hardhat Ignition files
- Added Hardhat Ignition to `hardhat-toolbox` and `hardhat-toolbox-viem`
- Updated relevant docs to use Hardhat Ignition where relevant

A sibling PR to update [hardhat-boilerplate](https://github.com/NomicFoundation/hardhat-boilerplate/pull/102) has also been opened.

resolves https://github.com/NomicFoundation/hardhat-ignition/issues/614
resolves https://github.com/NomicFoundation/hardhat-ignition/issues/613

## Manual Testing Instructions

To manually create HH projects with the new toolbox:

1. Pull the branch locally
2. Install the new deps: `pnpm install`
3. Apply the changeset versioning locally: `npx changeset version`
4. Run the build `pnpm build`
5. Start a new verdaccio instance and login (see https://www.notion.so/nomicfoundation/Using-verdaccio-e15e379b59a64ef8a82fe0c1a6b93dd8?pvs=4)
6. publish `hardhat` to verdaccio:
```shell
cd packages/hardhat-core
pnpm publish --registry=http://localhost:4873/ --no-git-check
```
7. publish `@nomicfoundation/hardhat-toolbox` to verdaccio:
```shell
cd packages/hardhat-toolbox
pnpm publish --registry=http://localhost:4873/ --no-git-check
```
8. publish `@nomicfoundation/hardhat-toolbox-viem` to verdaccio:
```shell
cd packages/hardhat-toolbox-viem
pnpm publish --registry=http://localhost:4873/ --no-git-check
```
9. Install the new hardhat globally from verdaccio (is there a better approach? Also you might want to do this outside a volta controller directory):
```shell
npm install -g hardhat@2.22.1 --registry=http://localhost:4873/
```
10. Create your new sample project with the globally installed hardhat:
```shell
hardhat --version
# 2.20.2
echo "registry=http://localhost:4873" > .npmrc
hardhat init
```
11. Run the docs locally:
```shell
cd docs
pnpm install
pnpm dev
```

## Questions and decisions

- [x] Are we happy with the failing CI Test - yes a local manual run with updated versions counts as a pass
- [x] How do we approach scripting in the docs given that it is currently about deploymeent
[...]

PR #4955, opened 2024-03-03
chore(hardhat-viem,hardhat-toolbox-viem): updated package.json with loosen version constraints for typescript

Chlebamaticon:
### Problem

Once adding `hardhat-toolbox-viem` to my local project I've stumbled upon following error while running `npm install`.

image

I have following packages installed in my repo;
```json
  "@nomicfoundation/hardhat-viem": "^2.0.0",
  "@nomicfoundation/hardhat-toolbox-viem": "^2.0.0",
``` 

### Solution

I've took a look at two `package.json` files for `hardhat-viem` and `hardhat-toolbox-viem`. What I found out is that both of them do contain conflicting peerDependencies.

The first is `hardhat-toolbox-viem` package and I believe it supposed to already rely on latest `hardhat-viem` version. Where now it depends on `^1.0.0`, so what I did I bumped it by one major version, so now it's `hardhat-viem@workspace:^2.0.0`.

The second was `typescript@~5.0.0` as a peer dependency of `hardhat-viem`. As you can already it capped up to only latest patch updates, where in my project I already have version `5.3.2`. To avoid such conflicts I've changed typescript version constraint to accept any minor version as well by setting `typescript@^5.0.0` in both; `hardhat-viem` & `hardhat-toolkit-viem`.

### DIFF
--- a/packages/hardhat-toolbox-viem/package.json
+++ b/packages/hardhat-toolbox-viem/package.json
@@ -73,7 +73,7 @@
   "peerDependencies": {
     "@nomicfoundation/hardhat-network-helpers": "workspace:^1.0.0",
     "@nomicfoundation/hardhat-verify": "workspace:^2.0.0",
-    "@nomicfoundation/hardhat-viem": "workspace:^1.0.0",
+    "@nomicfoundation/hardhat-viem": "workspace:^2.0.0",
     "@types/chai": "^4.2.0",
     "@types/chai-as-promised": "^7.1.6",
     "@types/mocha": ">=9.1.0",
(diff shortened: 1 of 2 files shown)


kanej:
Thanks for this, I am going to close in favour of https://github.com/NomicFoundation/hardhat/pull/4655.

Our toolbox doesn't contain `hardhat-viem`, as we are intending to package `hardhat-viem` + `hardhat-ignition-viem` in the latest version of the toolbox.

We will take a look at the typescript constraint, thanks for bringing our attention to it.

PR #4961, opened 2024-03-05
Run gas report and coverage in e2e

fvictorio:
The e2e workflow should fail with this change. For some reason, `npx hardhat coverage` doesn't work with the viem-based sample project. As part of this PR, we should fix that.

### DIFF
--- a/e2e/test-project-initialization.sh
+++ b/e2e/test-project-initialization.sh
@@ -140,7 +140,7 @@ for pkg_manager in $pkg_managers; do
   assert_no_empty_files
   $pkg_runner hardhat compile
   $pkg_runner hardhat test
-  $pkg_runner hardhat coverage
+  SOLIDITY_COVERAGE=true $pkg_runner hardhat coverage
   REPORT_GAS=true $pkg_runner hardhat test
   cd -
 
--- a/e2e/test-project-initialization.sh
+++ b/e2e/test-project-initialization.sh
@@ -80,6 +80,8 @@ for pkg_manager in $pkg_managers; do
   assert_no_empty_files
   $pkg_runner hardhat compile
   $pkg_runner hardhat test
+  $pkg_runner hardhat coverage
+  REPORT_GAS=true $pkg_runner hardhat test
   cd -
 
   # pkg_manager, javascript, esm
@@ -94,6 +96,8 @@ for pkg_manager in $pkg_managers; do
   assert_no_empty_files
   $pkg_runner hardhat compile
   $pkg_runner hardhat test
+  $pkg_runner hardhat coverage
+  REPORT_GAS=true $pkg_runner hardhat test
   cd -
 
   # pkg_manager, typescript, cjs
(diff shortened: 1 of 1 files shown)


cgewecke:
Hi @schaable,
 
Just leaving a few notes b/c I looked at this recently. The gas reporter is also [broken for viem][1] but that problem will be resolved in [the next major version][3] which is adding necessary logic and migrating to the new `extendProvider` hook.  (Should be published soon).

I'm not certain but for coverage I think the issue might come from a conflict caused by: 

+ coverage [re-defining the provider at the task level][2]
+ the viem plugin [passing the provider by reference when it initializes][5]

Coverage needs to [initialize the VM to certain values][4] - would like to do this in an `extendConfig` hook to minimize conflicts but that should only happen for the coverage task and it doesn't look like there's a way to know what command has been called from there. Would it be possible to pass TASK context info to the hook? e.g

```typescript
extendConfig((config: HardhatConfig, userConfig: Readonly<HardhatUserConfig>, [...here...])
```

cf: [solidity-coverage issue 819][6] 

[1]: https://github.com/cgewecke/hardhat-gas-reporter/issues/157
[2]: https://github.com/sc-forks/solidity-coverage/blob/6b576cae5a1072f7f282b82762f86c4153d0b795/plugins/resources/nomiclabs.utils.js#L128-L143
[3]: https://github.com/cgewecke/hardhat-gas-reporter/pull/170
[4]: https://github.com/sc-forks/solidity-coverage/blob/6b576cae5a1072f7f282b82762f86c4153d0b795/plugins/resources/nomiclabs.utils.js#L120-L126
[5]: https://github.com/NomicFoundation/hardhat/blob/ef367a4006c2c1f398c75fa6dc06d28d05c97494/packages/hardhat-viem/src/index.ts#L18
[6]: https://github.com/sc-forks/solidity-coverage/issues/819

cgewecke:
Update: The gas reporter issues with viem are fixed in the latest release there: 

https://github.com/cgewecke/hardhat-gas-reporter/releases/tag/v2.0.0

The problems with solidity-coverage will require some kind of fix here I think....or possibly both places. Just lmk.

cgewecke:
Another note...I think if the viem plugin followed the same strategy as the ethers plugin and passed `hre` instead of `provider` the conflict with solidity-coverage would be resolved

See...
https://github.com/NomicFoundation/hardhat/blob/ef367a4006c2c1f398c75fa6dc06d28d05c97494/packages/hardhat-ethers/src/internal/index.ts#L31-L50

cgewecke:
Another update...

In the latest version of solidity-coverage (v0.8.12) there is a work-around for viem. Coverage will configure the provider via `extendConfig` (side-stepping the conflict describe above) if the environment variable SOLIDITY_COVERAGE is set to true, e.g
```
SOLIDITY_COVERAGE=true npx hardhat coverage
```

If coverage detects `viem` on the hardhat environment without this env variable set, it now throws the advisory error below:

   
Screen Shot 2024-04-05 at 1 41 43 PM

schaable (approved):

kanej (approved):

PR #4964, opened 2024-03-05
research: add data analysis ipython notebook

agostbiro:
This is now ready to review.

The per scenario pattern visualization is not great, but making it better would take at least one more day, so I'm wrapping it up here.

You can view the rendered notebook here: https://github.com/NomicFoundation/hardhat/blob/418d44dd38797b4d35afa5f0a034044952557f11/crates/tools/python/scenarios.ipynb

Since you last saw it @fvictorio @Wodann, I have added transaction type and transaction kind analysis, patter analysis and extended the conclusion. 

@Wodann I didn't manage to improve the RPC call types graph in reasonable time.


Wodann:
Given that the patterns are detected per scenario, isn't there a natural bias towards scenarios that have more RPC calls?

crates/tools/python/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Requirements
+
+- [Miniconda](https://docs.anaconda.com/free/miniconda/index.html)
+
+## Running
+
+From this directory:
+
+```bash
+conda env create -f env.yml
```
Wodann, line 12: If you use `conda`, you might be interested in https://pixi.sh/latest/

Wodann (approved):
Overall this LGTM!

The only two things we could improve in the future are:

- flipping the axes such that text is alongside the vertical axis and numbers are below the horizontal axis, for easier readability
- the methodology used to determine the patterns isn't clear to me. Is there a minimum pattern length? Are certain patterns excluded? Is the overview for all scenarios weighted?

crates/tools/python/README.md
agostbiro: Thanks, I just use `conda` for reproducible ipython notebooks.

agostbiro:
> Overall this LGTM!
> 
> The only two things we could improve in the future are:
> 
> * flipping the axes such that text is alongside the vertical axis and numbers are below the horizontal axis, for easier readability

Agreed!

> * the methodology used to determine the patterns isn't clear to me. Is there a minimum pattern length? Are certain patterns excluded? Is the overview for all scenarios weighted?

Yes, minimum pattern length is 2. 

Patterns containing `{'eth_accounts', 'eth_chainId', 'eth_getBlockByNumber', 'net_version'}` as they added a lot of noise. If a pattern is a prefix of longer pattern it's also excluded to favor longer patterns.

The overview is scored on two factors:

1. The sum of the percentages (this makes sure longer scenarios are not favored) the pattern takes up in each scenario in its ngram class where it's included.
2. The number of scenarios the pattern appears in (min 2 two be included in the overview).

Shorter patterns are more likely to appear in multiple scenarios, this is why the overview has shorter patterns while the per scenario patterns have longer patterns.

PR #4967, opened 2024-03-06
Ci fixes

fvictorio:
See individual commits for the details.

### DIFF
--- a/.github/workflows/edr-npm-release.yml
+++ b/.github/workflows/edr-npm-release.yml
@@ -390,7 +390,7 @@ jobs:
       - test-linux-aarch64-gnu-binding
       - test-linux-aarch64-musl-binding
     # Only run workflow if the PR is merged to main and the commit message is a release commit.
-    if: ${{ needs.check_commit.outputs.match == 'true' && github.ref == 'refs/heads/main' }}
+    if: ${{ needs.check_commit.outputs.match == 'true' && (github.ref == 'refs/heads/main' || github.ref == 'refs/heads/edr/release') }}
     defaults:
       run:
         working-directory: ./crates/edr_napi
--- a/.github/workflows/edr-npm-release.yml
+++ b/.github/workflows/edr-npm-release.yml
@@ -20,7 +20,6 @@ on:
       - "**/*.gitignore"
       - .editorconfig
       - docs/**
-  pull_request: null
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -76,4 +76,4 @@ jobs:
           NODE_OPTIONS: "--max-old-space-size=4096"
         run: |
           pnpm build
-          pnpm test:forking
+          pnpm test:except-tracing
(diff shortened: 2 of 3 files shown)


.github/workflows/edr-npm-release.yml
```diff
@@ -20,7 +20,6 @@ on:
       - "**/*.gitignore"
       - .editorconfig
       - docs/**
-  pull_request: null
```
agostbiro, line 23: We've been running it on PRs to catch if a change introduces a regression that causes the build to fail on one of the target platforms (the publish step is not ran in PRs). Not sure if you intentionally wanted to remove this functionality @fvictorio ? Imo we should keep it.

agostbiro (changes requested):

.github/workflows/edr-npm-release.yml
fvictorio: The problem we have right now is that the "Check commit" job fails in PRs from external contributors, because `github.event.pull_request.head.ref` cannot be checked out (I guess because the ref is from a fork). We have several options:

1. Ignore it for now. After migrating to our own repo, there are going to be fewer external PRs (at least at the beginning), and so this won't be that important.
2. Skip that job for external PRs. I don't know how to do this.
3. Make that job work for external PRs. I also don't know how to do this.
4. Don't run the full workflow for external PRs (this is what I did). I assumed that the chances of this catching a bug are so low than catching it after merging into `main` is not that bad.

I might be underestimating the chances of a regression in the build though. In that case, I can undo this change for now, and we can re-visit the issue later.

agostbiro: If there is a regression, it probably comes from a new dependency not supporting one of the platforms (we had to switch crypto dependencies because of this), and that's quite difficult to fix after a PR has been merged, so imo we should keep this on. 

I think we could skip the `check_commit` and `publish` steps for forks by adding an `if: ${{ github.repository_owner == 'NomicFoundation' }}` check to the `check_commit` step. Haven't tested this though so not 100%.

fvictorio: That makes sense. I'll try that.

fvictorio: No, that doesn't work, it runs anyway for external users (I checked [here](https://github.com/fvictorio/test-repo/actions/runs/8186063562/job/22383690910)).

I'll give it another try later.

.github/workflows/edr-npm-release.yml
```diff
@@ -360,6 +360,7 @@ jobs:
   check_commit:
     name: Check commit
     runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request' || github.event.pull_request.author_association == 'OWNER' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'COLLABORATOR'
```
fvictorio, line 363: @agostbiro this is the best I could do. I don't love it, but I tested it a lot on a test repository and it seems to work fine.

agostbiro: Looks good cheers!

PR #4973, opened 2024-03-07
fix: avoid snapshotting when automining with empty mempool

agostbiro:
This is a draft PR to avoid snapshotting when automining with empty mempool which is the default case when automining.

It makes the `uniswap` scenario 3x faster and `openzeppelin` 2x faster on my machine, but it introduces new failures, so not ready to merge yet.


agostbiro:
Closing this as it doesn't improve performance since we just clone the state cache ids in snapshots now which makes them cheap.

PR #4977, opened 2024-03-08
Use explicit content-type header in RPC requests

fvictorio:
Closes https://github.com/NomicFoundation/edr/issues/323

EDIT: turning this into a draft because, while this lets the node start with that network, it can't mine a block after doing it.

The problem is that a `BlockchainError::InvalidParentHash` is thrown. I couldn't figure out why that happens with this network but not with mainnet. It might be related to faster mining times in that network. My guess is that there is a race condition where the "latest" block is fetched twice at different points, with different results. Not sure though.

### DIFF
--- a/crates/edr_eth/src/remote/client.rs
+++ b/crates/edr_eth/src/remote/client.rs
@@ -11,7 +11,7 @@ use std::{
 use futures::stream::StreamExt;
 pub use hyper::{http::Error as HttpError, HeaderMap};
 use itertools::{izip, Itertools};
-use reqwest::Client as HttpClient;
+use reqwest::{self, Client as HttpClient};
 use reqwest_middleware::{ClientBuilder as HttpClientBuilder, ClientWithMiddleware};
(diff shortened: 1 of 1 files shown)


Wodann:
One possible explanation is a reorg

fvictorio:
I don't think that chain has reorgs, and it happened too often to be caused by that.

Wodann:
I had a look at this and the reason for the failure is that when we're filling in the parent hash for the next block, we calculate the hash from the parent block's header. In this case that results in a different hash compared to the actual, as the chain has additional data in the header.

Theoretically, it's possible to support mining of local Ethereum blocks for chains like this by just blindly copying the parent block's hash, instead of calculating it. That would however require a refactor. Should we do that before focusing on proper multi-chain support?

cc: @fvictorio

Wodann:
Superseded by https://github.com/NomicFoundation/hardhat/pull/4992

PR #4978, opened 2024-03-08
ci: add JS benchmark runner

agostbiro:
Adds JS benchmark runner for scenarios.

Scenarios are patched to include 0x-prefixed secret keys instead of PEM.


book/src/02_development/01_tools.md
```diff
@@ -27,7 +27,8 @@ Create a provider test execution report for the candidate branch:
 yarn build && yarn test:provider --reporter json | tee candidate-test-provider-logs.json
 ```
 
-Generate a comparison report that will list slower tests in the candidate branch:
+Generate a comparison report that will list slower tests in the candidate
```
Wodann, line 30: It seems like some form of formatter was triggered. To simplify merge requests, each sentence was on a separate line. That's no longer the case.

crates/edr_provider/src/config.rs
```diff
@@ -58,10 +58,10 @@ pub struct MiningConfig {
 /// Configuration for the provider
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct ProviderConfig {
-    pub allow_blocks_with_same_timestamp: bool,
-    pub allow_unlimited_contract_size: bool,
+    pub allow_blocks_with_same_timestamp: bool, // x
+    pub allow_unlimited_contract_size: bool,    // x
     pub accounts: Vec<AccountConfig>,
-    /// Whether to return an `Err` when `eth_call` fails
+    /// Whether to return an `Err` when `eth_calll fails
```
Wodann, line 64: Seemingly erroneously introduced changes

crates/tools/js/benchmark/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+To run:
+
+```shell
+pnpm install
+pnpm run benchmark
+```
+
+The measurements will be printed stdout as machine-readable json and to stderr
+as human-readable output.
```
Wodann, line 13: Formatting

Wodann, line 13:
```suggestion
The measurements will be printed to stdout as machine-readable json and to stderr
as human-readable output.
```

crates/tools/js/benchmark/index.js
```diff
@@ -0,0 +1,216 @@
...
+} = require("hardhat/internal/hardhat-network/provider/provider");
+
+const SCENARIOS_DIR = "../../scenarios/";
+
+async function main() {
+  const numArgs = process.argv.length;
+  let grep = undefined;
+  if (
+    (numArgs > 2 && process.argv[numArgs - 2] === "--grep") ||
+    process.argv[numArgs - 2] === "-g"
```
Wodann, line 19:
```suggestion
    numArgs > 2 && (process.argv[numArgs - 2] === "--grep" ||
    process.argv[numArgs - 2] === "-g")
```

No?

crates/tools/js/benchmark/index.js
```diff
@@ -0,0 +1,216 @@
...
+    );
+    totalTime += scenarioResult.result.timeMs;
+    totalFailures += scenarioResult.result.failures.length;
+    result[scenarioResult.name] = scenarioResult.result;
+  }
+
+  console.log(JSON.stringify(result));
+
+  // Log info to stderr so that it doesn't pollute stdout where we write the result
+  console.error(
```
Wodann, line 48: Should this be an error? Is there something like info or debug?

crates/tools/js/benchmark/index.js
```diff
@@ -0,0 +1,216 @@
...
+    } seconds with ${totalFailures} failures.`
+  );
+
+  process.exit(0);
+}
+
+async function runScenario(scenarioPath) {
+  const { config, requests } = await loadScenario(scenarioPath);
+  const name = path.basename(scenarioPath).split(".")[0];
+  console.error(`Running ${name} scenario`);
```
Wodann, line 60: Same question as above around usage of `console.error`

crates/tools/js/benchmark/index.js
```diff
@@ -0,0 +1,216 @@
...
+    const hardforkHistory = new Map();
+    const hardforks = config.providerConfig.chains[key].hardforks;
+    for (let [blockNumber, hardfork] of hardforks) {
+      hardforkHistory.set(normalizeHardfork(hardfork), blockNumber);
+    }
+    chains.set(Number(key), { hardforkHistory });
+  }
+  config.providerConfig.chains = chains;
+
+  if (!_.isUndefined(config.providerConfig.fork)) {
```
Wodann, line 177: The `_` is very confusing to me. Can't we just import `isUndefined` from lodash?

crates/tools/js/benchmark/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "benchmark",
+  "version": "1.0.0",
+  "private": true,
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "benchmark": "node index.js",
+    "prebenchmark": "cd ../../../edr_napi/ && pnpm build && cd ../../packages/hardhat-core/ && pnpm build"
```
Wodann, line 9: How will we do this when we move to a separate repo? Should we already follow that process?

cc: @fvictorio

Wodann (changes requested):

crates/tools/js/benchmark/index.js
agostbiro: Yes, thanks!

crates/tools/js/benchmark/index.js
agostbiro: Info is logged to stderr so that you can do `npm run benchmark > results.json` and still see human readable logs as it progresses.

crates/tools/js/benchmark/index.js
agostbiro: https://github.com/NomicFoundation/hardhat/pull/4978#discussion_r1521278713

crates/tools/js/benchmark/index.js
agostbiro: It's standard to use `lodash` like this

crates/tools/js/benchmark/package.json
agostbiro: It'd need `verdaccio` I think if we treated `hardhat-core` as a separate repo to be able to use custom version of `edr` with it. `verdaccio` means we'd need a Docker runner for consistent results locally, so I'd rather cross that bridge once we get there.

crates/tools/js/benchmark/README.md
agostbiro: Thanks fixed in [daa0615](https://github.com/NomicFoundation/hardhat/pull/4978/commits/daa061553a1a33cecfe7e55f4fe978fc1e8980fc)

crates/edr_provider/src/config.rs
agostbiro: Thanks fixed in [daa0615](https://github.com/NomicFoundation/hardhat/pull/4978/commits/daa061553a1a33cecfe7e55f4fe978fc1e8980fc)

book/src/02_development/01_tools.md
agostbiro: Thanks fixed in [daa0615](https://github.com/NomicFoundation/hardhat/pull/4978/commits/daa061553a1a33cecfe7e55f4fe978fc1e8980fc)

crates/tools/js/benchmark/package.json
fvictorio: My feeling is that this should live on the Hardhat side. Linking EDR on it is something we'll do often, so it's better to always do it in one direction. The problem is how this would play with the automated benchmark runs, but that doesn't seem like an impossible problem to figure out.

crates/tools/js/benchmark/index.js
```diff
@@ -0,0 +1,217 @@
...
+  config.providerConfig.initialDate = new Date(
+    config.providerConfig.initialDate.secsSinceEpoch * 1000
+  );
+
+  config.providerConfig.hardfork = normalizeHardfork(
+    config.providerConfig.hardfork
+  );
+
+  // "accounts" in EDR are "genesisAccounts" in Hardhat
+  if (Object.keys(config.providerConfig.genesisAccounts).length !== 0) {
```
fvictorio, line 143: The way I'm understanding this from looking at the code is that EDR has a concept of `genesis_accounts`, which is an initial list of impersonated accounts, and it also has an `accounts` field which is closer to what Hardhat calls `genesisAccounts`. Is this correct?

fvictorio (approved):
Pushed some minor changes, but otherwise LGTM

crates/tools/js/benchmark/index.js
agostbiro: Yes, the difference in EDR is that you need a secret key for accounts, but not for genesis accounts.

crates/tools/js/benchmark/README.md
Wodann: Just one line was fixed. The whole file was reformatted

Wodann (approved):
LGTM!

Issue #4970, opened 2024-03-07
Overriding a storage slot of a fork does not work anymore


julien-devatom:
### Version of Hardhat

2.21.0

### What happened?

When you override a storage slot, it does not work anymore.
If I do a query that should use this storage slot in the same block, it returns "could not decode result data".
If I mine a block, it returns the initial value (without the slot override) 

### Minimal reproduction steps

Example with the override of the USDC balance of a given user
```typescript
    const userAddress = "0x...";
    const amount = toBeHex("1000000", 32);
    const usdc = "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48";

    const erc20 = new Contract(usdc, ["function balanceOf(uint256) public view returns (uint256)"], provider);
 
    const hexUserStorageSlot = AbiCoder.defaultAbiCoder().encode(
      ["address", "uint256"],
      [userAddress, 9]
    );

    const userStorageSlot = ethers.stripZerosLeft(ethers.keccak256(hexUserStorageSlot));
    await hardhat.provider.send("hardhat_setStorageAt", [usdc, 9, amount]);

    const balance = await erc20.balanceOf(userAddress); // Before 2.21.0, it was returning "1000000", after 2.21.0, it throws

    await hardhat.provider.send("evm_mine", []);
    const balanceAfterMine = await erc20.balanceOf(userAddress); // Before 2.21.0: "1000000", after 2.21.0: "0"
```

Wodann:
Thank you for reporting the bug! 😃

A fix has been released in https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.22.1

PR #4985, opened 2024-03-12
fix: use remote account when storage is manually modified

Wodann:
Fixes https://github.com/NomicFoundation/hardhat/issues/4970 and https://github.com/NomicFoundation/edr/issues/324

TODO: I would like to add a test based on the example in https://github.com/NomicFoundation/hardhat/issues/4970, but wasn't able to figure out the storage slot index. @fvictorio, could you please attempt to fix that Typescript test - if the solution is obvious to you?


packages/hardhat-ethers/test/hardhat-ethers-provider.ts
```diff
@@ -483,6 +484,44 @@ describe("hardhat ethers provider", function () {
...
+        ["address", "uint256"],
+        [userAddress, balanceStorageSlot]
+      );
+
+      const userStorageSlot = stripZerosLeft(keccak256(hexUserStorageSlot));
+      console.log("slot:", userStorageSlot);
+
+      await this.env.network.provider.send("hardhat_setStorageAt", [
+        usdc,
+        userStorageSlot,
```
Wodann, line 511: @fvictorio this is the part that I'm unfamiliar with. How can you calculate the storage index for the value returned by `balanceOf`?

fvictorio:
@Wodann I replaced the `hardhat-ethers` test with a test in `hardhat-core`, because AFAIK `hardhat-ethers` tests don't run with forked providers at the moment. While ideally the test would be in `hardhat-ethers`, because it's closer to the reported issue, I don't want to spend time changing the setup to add forking. I think the `hardhat-core` test is a good approximation (I double-checked that it fails in `main` and that it passes with your change.)

packages/hardhat-ethers/test/hardhat-ethers-provider.ts
fvictorio: You need to either use the original code to figure it out (manually or with the compiler), or trace a call to the function and manually inspect which storage slots are accessed. Neither option is too easy, so for my test I just guessed and verified that the storage value matched the value returned by the `totalSupply` function.

.changeset/big-ghosts-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
+Fixed #4970
```
fvictorio, line 5: The way we write changesets is like entries in the `CHANGELOG.md` file.

```suggestion
Fixed a bug in `hardhat_setStorageAt` that occured when the storage of a remote contract was modified during forking (#4970)
```

PR #4986, opened 2024-03-12
EDR book improvements

fvictorio:
See commit by commit (except the "Run prettier" commit, which can be ignored).


book/src/00_introduction.md
```diff
@@ -1,8 +1,5 @@
 # Introduction
 
-> Note: EDR and this book are currently under active development.
```
Wodann, line 3: To make reviews of markdown simpler, we're purposefully putting one sentence per line. When entire paragraphs are combined into a single line, it blows up the diffs (as happened here).

Could you please revert the formatter commit?

package.json
```diff
@@ -20,7 +20,7 @@
     "test": "node scripts/run-tests.js",
     "lint": "pnpm run --recursive lint && pnpm prettier --check",
     "lint:fix": "pnpm run --recursive lint:fix && pnpm prettier --write",
-    "prettier": "prettier *.md \"{docs,.github}/**/*.{md,yml,ts,js}\" \"scripts/**/*.js\""
+    "prettier": "prettier *.md \"{book,docs,.github}/**/*.{md,yml,ts,js}\" \"scripts/**/*.js\""
```
Wodann, line 23: We specifically don't want to do this to improve readability and minimise diff sizes (we write one sentence per line)

book/src/00_introduction.md
fvictorio: Done.

Wodann (approved):

PR #4987, opened 2024-03-12
fix: remove empty pending and future transaction lists

Wodann:
Fixes https://github.com/NomicFoundation/edr/issues/325


agostbiro (approved):
Thanks, LGTM!

PR #4988, opened 2024-03-13
fix: don't check nonce in calls

Wodann:
[This change](https://github.com/bluealloy/revm/commit/b25b44aac8132bd7086faf3bc3e5f79cbf0ded49) was required to our fork of `revm`. We'll need to port this to the latest version of `revm`


crates/edr_evm/src/runtime.rs
```diff
@@ -66,6 +66,7 @@ where
 {
     cfg.disable_balance_check = true;
     cfg.disable_block_gas_limit = true;
+    cfg.disable_nonce_check = true;
```
Wodann, line 69: This disables nonce checks for all calls; i.e `eth_call`, `debug_traceCall`, and `eth_estimateGas`. Is this desired @fvictorio ?

agostbiro (approved):
Thanks, LGTM, but I'd wait for @fvictorio 's review as well before merging.

crates/edr_evm/src/runtime.rs
fvictorio: I think it's desired, yes. I also checked and this matches geth's behavior.

crates/edr_evm/src/mempool.rs
```diff
@@ -335,6 +335,10 @@ impl MemPool {
                     let mut invalidated_transactions = pending_transactions.split_off(idx + 1);
                     let removed = pending_transactions.remove(idx);
 
+                    if pending_transactions.is_empty() {
+                        self.pending_transactions.remove(caller);
+                    }
```
fvictorio, line 340: I'm curious why this was needed. I would've guessed that having an entry in the pending/future transactions map whose value is an empty list wouldn't be a problem.

fvictorio (approved):

PR #4990, opened 2024-03-13
feat: set cancun as default hardfork

kanej:
Set cancun as the default config for the hardfork used in Hardhat network.

### DIFF
--- a/.changeset/gold-comics-do.md
+++ b/.changeset/gold-comics-do.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Set cancun as the default hardfork in Hardhat network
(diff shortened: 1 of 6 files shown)


.changeset/gold-comics-do.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
```
fvictorio, line 2: We use minor bumps when changing the default hardfork.

```suggestion
"hardhat": minor
```

Wodann:
This PR will also be of interest if we're rolling this out in a release: https://github.com/NomicFoundation/hardhat/pull/4991

PR #4991, opened 2024-03-13
feat: add Cancun activation block number

Wodann:
This adds detection of `Cancun` blocks on mainnet, when using the default hardfork activation detection.


fvictorio (approved):

fvictorio (changes requested):
Some tests are failing, I'm looking into that now.

fvictorio (approved):

Issue #4974, opened 2024-03-07
Forking a network locally doesn't work anymore


kuco23:
### Version of Hardhat

2.21.0

### What happened?

After updating hardhat from `2.20.1` to `2.21.0` forking a test node fails with `Error HH604: Error running JSON-RPC server: The Http server returned error status code: HTTP status client error (415 Unsupported Media Type) for url (https://flare-api.flare.network/ext/C/rpc)`.

### Minimal reproduction steps

Run `yarn hardhat node --fork https://coston-api.flare.network/ext/bc/C/rpc --fork-block-number 12508443`.

fvictorio:
Hey @kuco23, kick question, is that network Avalanche's C-chain?

kuco23:
yes, it is

FournyP:
Hi, I faced the same issue using polygon mainnet :
With this configuration :
```
hardhat: {
  chainId: 137,
  forking: {
    url: "https://polygon.llamarpc.com",
    blockNumber: 54461490,
  },
},
```
```
Error: The Http server returned error status code: HTTP status client error (400 Bad Request) for url (https://polygon.llamarpc.com/)
```

And with a `infura.io` node, I got the following error :
```
Error: The Http server returned error status code: HTTP status client error (429 Too Many Requests) for url (https://polygon-mainnet.infura.io/v3/xxx)
```

Doc-Pixel:
Getting the same issue with shibarium.

Rolled back to 2.20.0 and it works again

fullkomnun:
I have an internal Ethereum network that I fork by targeting a Geth node running remotely (used to work with V2.20.0).
I get the same issue for it with V2.21.0 (forking public networks with Alchemy endpoints works though).

Here are the logs:

```
  hardhat:core:hre Running task node +5ms
  hardhat:core:hre Running node's super +0ms
  hardhat:core:hre Running task node:get-provider +0ms
  hardhat:core:hre Running node:get-provider's super +0ms
  hardhat:core:hre Creating provider for network hardhat +0ms
Error HH604: Error running JSON-RPC server: The Http server returned error status code: HTTP status client error (415 Unsupported Media Type) for url (http://geth-testnet:8545/)

HardhatError: HH604: Error running JSON-RPC server: The Http server returned error status code: HTTP status client error (415 Unsupported Media Type) for url (http://geth-testnet:8545/)
[...]

PR #4992, opened 2024-03-13
fix: set content-type and user-agent in RPC client

Wodann:
Fixes https://github.com/NomicFoundation/hardhat/issues/4974 & https://github.com/NomicFoundation/edr/issues/323


agostbiro (approved):
Thanks LGTM! There was a CI failure, but it turned out to be a temporary problem with Alchemy.

PR #4998, opened 2024-03-14
ci: add performance benchmarks

agostbiro:
Adds automated performance regression checks for PRs and historic progress tracking in `main`. The benchmark replays RPC calls captured in third-party test suites.

## Runner

We're using a self-hosted Github Action runner to run benchmarks in a reproducible environment on a bare metal instance with an `Intel(R) Xeon(R) E-2276G CPU @ 3.80GHz` cpu. The runner is executed with an unprivileged user and standard security hardening was applied to the instance. The runner is ran as a `systemd` service with automatic restarts and reboot if many restarts fail.

I spent some time trying find a managed CI provider with a deterministic execution environment, but couldn't, so it seems we're stuck with self-hosted runners at least mid term.

## Reports

We are using [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) to track historic progress and raise regression alerts on PRs.

### Historic

Benchmark data is to be stored in a separate repo. The Github pages in this repo will show historic benchmark data for `main` like [this.](https://agostbiro-hardhat-benchmark.github.io/hardhat-fork-benchmark-results/bench/) (The example only includes the `seaport` scenario as the entire benchmark run is slow).

### Regressions in PRs

When a performance regression is detected in a PR based on a preconfigured threshold and compared to the latest commit in `main`, the following error is displayed in the PR:

Screenshot 2024-03-14 at 19 13 57

Clicking into the workflow, the following table is displayed giving information about the magnitude of the regression (the Markdown isn't rendered properly, this is a bug in [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark)):

Screenshot 2024-03-14 at 19 14 11

It is a failure if there is a regression on any of the metrics.

#### Snapshot testing

The benchmark workflow checks that the same RPC calls succeed and fail as when the scenario was collected. If the snapshot doesn't match, the workflow fails and the difference is reported.

## TODOs for this PR

- [x] Update expected failures snapshot in the repo to include all scenarios
- [x] Calibrate regression alert threshold percentage
- [x] Create benchmark data repo under `NomicFoundation` and configure permissions and self-hosted runner in  `NomicFoundation/hardhat`

## Limitations/things to improve in future

- Full run takes an hour and there is only one runner so tasks might queue up. We could improve things by making the benchmark run faster and by introducing more runners.
- Only runs for team members’ and collaborators’ PRs as it's unsafe to allow third-parties to run code on self-hosted runners [according to Github.](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security)
- Currently we rerun the benchmark in `main` even if it was already ran in a PR that was merged. This could be probably avoided.
- Can't use `Swatinem/rust-cache@v2` in the workflow as it needs sudo on the runner which is too dangerous. We should look into alternatives to prevent recompiling from scratch on every run.
- The regression error message doesn’t render Markdown properly. I’ll open an issue about this in [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark).
- [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark) has a feature to comment on the PR which would allow seeing the regression results easier, but this is blocked by the action using the same Github token to push to the benchmark result repo and to comment. We’d need to open a PR in the action repo to support it.


.github/workflows/edr-benchmark.yml
```diff
@@ -0,0 +1,86 @@
...
+      - "crates/**"
+  pull_request:
+    branches:
+      - "**"
+    paths:
+      - ".github/workflows/edr-benchmark.yml"
+      - "rust-toolchain"
+      - "Cargo.lock"
+      - "Cargo.toml"
+      - "crates/**"
```
agostbiro, line 21: Only running for changes to EDR since it's very slow and there is only one runner + we'll do this anyway once EDR is a separate repo

crates/tools/js/benchmark/index.js
```diff
@@ -3,35 +3,116 @@ const fs = require("fs");
...
+  });
+  parser.add_argument("-g", "--grep", {
+    type: "str",
+    help: "Only execute the scenarios that contain the given string",
+  });
+  parser.add_argument("-o", "--benchmark-output", {
+    type: "str",
+    default: "./benchmark-output.json",
+    help: "Where to save the benchmark output file",
+  });
```
agostbiro, line 32: Save this to disk instead of stdout since multiple comments operate on it

.github/workflows/edr-benchmark.yml
```diff
@@ -0,0 +1,86 @@
...
+  group: ${{github.workflow}}-${{github.ref}}
+  cancel-in-progress: true
+
+jobs:
+  js-benchmark:
+    name: Run JS scenario runner benchmark
+    environment: github-action-benchmark
+    runs-on: self-hosted
+    # Only run for trusted collaborators since third-parties could run malicious code on the self-hosted benchmark runner.
+    if: github.event_name != 'pull_request' || github.event.pull_request.author_association == 'OWNER' || github.event.pull_request.author_association == 'MEMBER' || github.event.pull_request.author_association == 'COLLABORATOR'
```
Wodann, line 38: My understanding is that we have a single workflow that has two paths:

1. a path that runs for PRs (of members) to notify about regressions
2. a path for `main` that records historic performance

Is that understanding correct?

Wodann (approved):
LGTM! Thank you for the clear explanation in the PR description. That was very helpful!

I just have one question to clarify that my understanding of the GitHub workflow is correct.

.github/workflows/edr-benchmark.yml
agostbiro: Yes that's correct! The difference is that in `main` the results are recorded, but there are no errors for regressions while on PRs (by members) the results aren't recorded, but the workflow fails if there is a regression.

agostbiro:
Follow up: https://github.com/NomicFoundation/edr/issues/337

Issue #4984, opened 2024-03-11
`hardhat_setBalance` on contracts is reverting with  "Contract with code hash `0x...` does not exist"


bugduino:
### Version of Hardhat

2.21.0

### What happened?

Hi, I have an old repo which was recently updated to hardhat >= 2.17.3 as I needed to have shanghai evm version (currently at 2.21.0) and I'm running a script in eth mainnet fork.
Everything seems to be working fine except for `hardhat_setBalance` which is now constantly halting script execution.
I've tried many different versions >= 2.17.3 but it always reverts on the hardhat_setBalance

This is the trace using `loggingEnabled: true`:

```
hardhat_setBalance [THIS IS IN RED]

  Contract with code hash `0xaea7d4252f6245f301e540cfbee27d3a88de543af8e49c5c62405d5499fab7e5` does not exist.

An unexpected error occurred:

ProviderError: Contract with code hash `0xaea7d4252f6245f301e540cfbee27d3a88de543af8e49c5c62405d5499fab7e5` does not exist.
    at EdrProviderWrapper.request (/idle/idle-proposals/node_modules/hardhat/src/internal/hardhat-network/provider/provider.ts:424:19)
    at SimpleTaskDefinition.action (/idle/idle-proposals/scripts/execute-proposal-or-simulate.ts:19:7)
    at Environment._runTaskDefinition (/idle/idle-proposals/node_modules/hardhat/src/internal/core/runtime-environment.ts:359:14)
    at Environment.run (/idle/idle-proposals/node_modules/hardhat/src/internal/core/runtime-environment.ts:192:14)
    at SimpleTaskDefinition.action (/idle/idle-proposals/scripts/iip-39.ts:113:3)
    at Environment._runTaskDefinition (/idle/idle-proposals/node_modules/hardhat/src/internal/core/runtime-environment.ts:359:14)
    at Environment.run (/idle/idle-proposals/node_modules/hardhat/src/internal/core/runtime-environment.ts:192:14)
    at main (/idle/idle-proposals/node_modules/hardhat/src/internal/cli/cli.ts:323:7)
```

I'm not sure why ```ProviderError: Contract with code hash `0xaea7d4252f6245f301e540cfbee27d3a88de543af8e49c5c62405d5499fab7e5` does not exist.``` is even thrown honestly. If I use other methods to set the balance (eg transfer funds from another address) it works as expected.

### Minimal reproduction steps

[...]

PR #5000, opened 2024-03-14
fix: retrieve code for forked accounts when modifying state

Wodann:
Fixes https://github.com/NomicFoundation/hardhat/issues/4984


agostbiro (approved):
Thanks, LGTM

PR #5008, opened 2024-03-19
ci: disable snapshot tests in benchmarks

agostbiro:
Had to disable snapshot tests for `neptune-mutual` as they were unreliable and replaced `Set.difference` with custom implementation as the former is not universally supported.


Wodann (approved):

agostbiro:
Now `open-zeppelin` snapshot tests failed in the CI that worked for me locally. Going to disable all snapshot tests to unblock performance regression checks. Follow up: https://github.com/NomicFoundation/edr/issues/338

PR #5019, opened 2024-03-25
Relax EDR dependency range

fvictorio:
The `^0.3.0` range won't install versions like `0.4.0`. This, plus the fact that we won't be introducing breaking changes in new minor 0.x versions, makes me thing that it's simpler for Hardhat to just accept any 0.x version.

We might also want to include a `>=` part in this range, if we add something to the EDR API that's then used by Hardhat. But we can think about that when it happens.

### DIFF
--- a/packages/hardhat-core/package.json
+++ b/packages/hardhat-core/package.json
@@ -107,7 +107,7 @@
   "dependencies": {
     "@ethersproject/abi": "^5.1.2",
     "@metamask/eth-sig-util": "^4.0.0",
-    "@nomicfoundation/edr": "workspace:^0.3.1",
+    "@nomicfoundation/edr": "workspace:<1.0.0",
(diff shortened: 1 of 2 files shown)


kanej (approved):

Wodann:
> plus the fact that we won't be introducing breaking changes in new minor 0.x versions, makes me thing that it's simpler for Hardhat to just accept any 0.x version.

I think it's important to note that we might be introducing breaking changes but that we're in control of fixing them on the Hardhat side. I'm not sure how this might affect external users if we have released a new EDR package but not a new Hardhat package yet

fvictorio:
@Wodann is right. If we do this, we are committing to not introduce breaking changes in `0.x`, and I don't think we want to do that.

The downside is that the latest EDR versions will take longer to reach users, but I think that's fine.

Issue #4997, opened 2024-03-14
Run a script using the hardhat network, makes the script to never end if the provider and some promise was used from =>v2.21.0


invocamanman:
### Version of Hardhat

It happens from v2.21.0

### What happened?

Once i run the following script:
```
import {ethers} from "hardhat";

async function main() {
    // Load provider
    const provider = ethers.provider;
    await provider.getBalance(ethers.ZeroAddress);
}

main().catch((e) => {
    console.error(e);
    process.exit(1);
});
```
Using `npx hardhat run script.ts` and having the default network on hardhat.

### Minimal reproduction steps

Open a harhdat repository with  v2.21.0 or higher

- Cerate a script.ts: 
```
import {ethers} from "hardhat";

async function main() {
    // Load provider
    const provider = ethers.provider;
    await provider.getBalance(ethers.ZeroAddress);
}

main().catch((e) => {
    console.error(e);
    process.exit(1);
});
```
- execute it using: `npx hardhat run script.ts`

invocamanman:
Added info: it happens as well with other networks

doerfli:
I can confirm the same issue. Reproduceable on latest dependabot updates of our repo, e.g. https://github.com/etherisc/gif-next/pull/283

Wodann:
A fix to EDR was pushed and we'll publish a new release ASAP

PR #5025, opened 2024-03-26
fix: use weak references for stored threadsafe functions

Wodann:
Fixes https://github.com/NomicFoundation/edr/issues/339 & #4997.

In the process of making this fix, I also bumped `napi` to v2.16.0. As it doesn't affect the generated `index.d.ts` of `edr_napi`, this can still be considered a patch version bump for EDR.


packages/hardhat-core/scripts/issue-339.js
```diff
@@ -0,0 +1,41 @@
+const edr = require("@nomicfoundation/edr")
```
Wodann, line 1: @fvictorio should this file be kept? Feel free to modify this to be automatically testable

fvictorio (approved):
Ran some manual tests and it seems to work fine:

- Running a script with `hh run`
- Running a script with `node --require hardhat/register`
- Running a script with `node script.js`, where the script explicitly requires the HRE
- Running `hh test`
- Running `mocha --require hardhat/register`
- Running scripts and tests in a typescript project

PR #5033, opened 2024-03-27
fix: support hex string in `eth_signTypedData_v4` salt

agostbiro:
Adds workaround for https://github.com/NomicFoundation/edr/issues/346

### DIFF
--- a/.changeset/neat-apes-mix.md
+++ b/.changeset/neat-apes-mix.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/edr": patch
+---
+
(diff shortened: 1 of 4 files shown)


Wodann (approved):

agostbiro:
Merging despite the performance regression alert since this PR only changed `eth_signTypedData_v4` and the reported regression was for the `synthetix` scenario which doesn't include this RPC call.

agostbiro:
Hardhat e2ee tests will continue to hang until latest version of EDR is released, so it's safe to ignore.

Issue #5036, opened 2024-03-26
Sporadic test failures when running `hardhat test` due to incorrect usage of chai matchers when hardhat >= 2.21.0


gzliudan:
### Version of Hardhat

>=2.21.0

### What happened?

- pass all tests when hardhat < 2.21.0
- fail some tests when hardhat >= 2.21.0

### Minimal reproduction steps

```bash
git clone https://github.com/gzliudan/MatrixProtocol.git
cd MatrixProtocol
# setup hardhat's version in package.json 
yarn
yarn make
```

The test will cost about 10 minutes.

gzliudan:
I test in 3 VMs with your PR, but still has some tests failed each time. My test environment:

- Ubuntu: 22.04.04
- node: v18.20.0 and v18.19.1

gzliudan:
I fixed this problem according to your suggestions. Thanks for your kindly help again.

PR #5037, opened 2024-03-27
ci: enable snapshot testing

agostbiro:
This PR enables snapshot testing in the CI by checking that the same requests succeed and fail between benchmark runs. 

The `openzeppelin-contracts` and `neptune-mutual-blue-protocol` scenarios are excluded from snapshot testing, because there is some variance in which calls and transactions succeed or fail. A time-limited investigation found that this variance is related to block timestamps. Since it's difficult to know the expected behavior based on the snapshots, as a follow up we added https://github.com/NomicFoundation/edr/issues/240

This PR also updates the `openzeppelin-contracts` scenario to a version that was collected after some sporadically failing tests were fixed in the `openzeppelin-contracts` repo.


crates/tools/js/benchmark/index.js
```diff
@@ -89,6 +89,14 @@ async function verify(benchmarkResultPath) {
   ));
 
   for (let scenarioName in snapshotResult) {
+    // Snapshot testing is unreliable for these scenarios
+    if (
+      scenarioName.includes("openzeppelin") ||
```
Wodann, line 94: Could this be related to https://github.com/NomicFoundation/hardhat/issues/5036#issuecomment-2023343177?

Wodann (approved):
One discussion but otherwise LGTM

crates/tools/js/benchmark/index.js
Wodann: Just an FYI, when I enabled logging for a repo - that suffers from the described problem - and compared the generated logs, the order of calls wasn't deterministic. This was a sign to me that tests were being executed in parallel.

This might be a relatively straightforward way of testing the hypothesis for other repos.

PR #5056, opened 2024-04-02
Throw CORRUPTED_LOCKFILE error when requiring EDR fails

fvictorio:
I created a `requireNapiRsModule` wrapper function to use both when requiring `solidity-analyzer` and EDR. The rest of the changes should be self-explanatory.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToEdr.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToEdr.ts
@@ -1,4 +1,4 @@
-import {
+import type {
   SpecId,
   MineOrdering,
   IntervalRange,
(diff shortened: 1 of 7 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToEdr.ts
```diff
@@ -128,6 +137,10 @@ export function ethereumjsIntervalMiningConfigToEdr(
 export function ethereumjsMempoolOrderToEdrMineOrdering(
   mempoolOrder: MempoolOrder
 ): MineOrdering {
+  const { MineOrdering } = requireNapiRsModule(
```
kanej, line 140: Can I double check the pattern in play here.

The `MineOrdering` static import at the top is switched to `type` so that only the type information is pulled in.

Then on actual invocation we pull in th implementation of `MineOrdering` via a dynamic require (wrapped in the `requireNapiRsModule` helper for the corrupt lockfile check).

So we get to work with the right type, but gain more control over our dependency loading (and its potential failure).

In effect there is now a rule that every import of `@nomicfoundation/edr` or `@nomicfoundation/solidity-analyzer` should follow this pattern?

fvictorio: Yes to everything. Ideally we would have a new custom eslint rule for that (it shouldn't be too hard to add), but for now I think this is enough.

kanej (approved):

PR #5061, opened 2024-04-02
POC - Viem Assertion Helpers

ChristopherDedominici:


packages/hardhat-viem-assert/src/index.ts
```diff
@@ -0,0 +1,31 @@
...
+      async changeEtherBalance(
+        publicClient: any,
+        address: any,
+        amount: bigint
+      ) {
+        const balanceBefore = await publicClient.getBalance({
+          address,
+        });
+
+        await fn();
```
schaable, line 15: we should wrap the fn execution in a try/catch

packages/hardhat-viem-assert/src/index.ts
```diff
@@ -0,0 +1,31 @@
...
+  // TODO: function or value
+
+  return {
+    to: {
+      async changeEtherBalance(
+        publicClient: any,
+        address: any,
+        amount: bigint
+      ) {
+        const balanceBefore = await publicClient.getBalance({
```
schaable, line 11: I believe we can get the public client from `hre.viem` instead of passing it as a parameter

alcuadrado:
Closing to clean up the list of PRs. We'll get back to this later.

PR #5062, opened 2024-04-02
Remove EDR from `v-next`

alcuadrado:
This PR removes EDR from the repository in the `v-next` branch, as EDR will be moved to its own repository before `v-next` gets merged.

Note that I also removed every Hardhat Network related CI jobs, as those should also be run by EDR, and not Hardhat.


fvictorio:
LGTM, but can you expand on commit cff2badc3ef9c95f88504cf2e476cde4af27b235? Why was that job broken?

fvictorio (approved):

alcuadrado:
> LGTM, but can you expand on commit [cff2bad](https://github.com/NomicFoundation/hardhat/commit/cff2badc3ef9c95f88504cf2e476cde4af27b235)? Why was that job broken?

It's the setup that's broken. It wouldn't pass, because a PR was merged that didn't modify core so it didn't trigger it and a "broken" version passed. In `v-next` we'll only use node 20+, so we don't need it anymore.

PR #5067, opened 2024-04-03
fix: remove api keys from RPC error messages

agostbiro:
Only log the host instead of the full URL in RPC errors to avoid leaking api keys in CI.

### DIFF
--- a/.changeset/brave-scissors-train.md
+++ b/.changeset/brave-scissors-train.md
@@ -1,5 +1,5 @@
 ---
-
+"@nomicfoundation/edr": patch
 ---
 
 fix: remove api keys from RPC error messages
--- a/.changeset/brave-scissors-train.md
+++ b/.changeset/brave-scissors-train.md
@@ -1,0 +1,5 @@
+---
+
+---
+
+fix: remove api keys from RPC error messages
--- a/crates/edr_eth/src/remote/client.rs
+++ b/crates/edr_eth/src/remote/client.rs
@@ -1206,7 +1206,7 @@ mod tests {
 
         if let RpcClientError::HttpStatus(error) = error {
             assert_eq!(
-                <ReqwestError as Into<reqwest::Error>>::into(error).status(),
+                reqwest::Error::from(error).status(),
                 Some(StatusCode::from_u16(STATUS_CODE).unwrap())
             );
         } else {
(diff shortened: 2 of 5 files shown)


crates/edr_eth/src/remote/client.rs
```diff
@@ -1203,7 +1206,7 @@ mod tests {
 
         if let RpcClientError::HttpStatus(error) = error {
             assert_eq!(
-                error.status(),
+                <ReqwestError as Into<reqwest::Error>>::into(error).status(),
```
Wodann, line 1209: Why did `reqwest::Error::from(..)` not work here?

crates/edr_eth/src/remote/client.rs
```diff
@@ -1245,20 +1248,26 @@ mod tests {
...
                 .call::<Option<eth::Transaction>>(RequestMethod::GetTransactionByHash(hash))
                 .await
                 .expect_err("should have failed to interpret response as a Transaction");
 
+            assert!(!error.to_string().contains(api_key));
+
             if let RpcClientError::HttpStatus(error) = error {
-                assert_eq!(error.status(), Some(StatusCode::from_u16(401).unwrap()));
+                assert_eq!(
+                    <ReqwestError as Into<reqwest::Error>>::into(error).status(),
```
Wodann, line 1268: Same question w.r.t `From::from`

.changeset/brave-scissors-train.md
```diff
@@ -0,0 +1,5 @@
+---
+
```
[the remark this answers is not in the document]
agostbiro: thanks!

crates/edr_eth/src/remote/client.rs
agostbiro: Yeah that's cleaner thanks, fixed in [Simplify error conversion](https://github.com/NomicFoundation/hardhat/pull/5067/commits/fa0abb8e7574ecdeae2c4a4a6d6761d27ac52023)

crates/edr_eth/src/remote/client.rs
agostbiro: fixed in [Simplify error conversion](https://github.com/NomicFoundation/hardhat/pull/5067/commits/fa0abb8e7574ecdeae2c4a4a6d6761d27ac52023)

fvictorio (approved):

PR #5074, opened 2024-04-04
Fix cargo doc warnings

fvictorio:
I got tired of seeing this in PRs:

image

### DIFF
--- a/crates/edr_evm/src/blockchain/forked.rs
+++ b/crates/edr_evm/src/blockchain/forked.rs
@@ -55,7 +55,7 @@ pub enum CreationError {
     },
 }
 
-/// Error type for [`RemoteBlockchain`].
+/// Error type for `RemoteBlockchain`.
 #[derive(Debug, thiserror::Error)]
 pub enum ForkedBlockchainError {
     /// Remote block creation error
--- a/crates/edr_evm/src/blockchain/forked.rs
+++ b/crates/edr_evm/src/blockchain/forked.rs
@@ -55,7 +55,7 @@ pub enum CreationError {
     },
 }
 
-/// Error type for `RemoteBlockchain`.
+/// Error type for [`ForkedBlockchain`].
 #[derive(Debug, thiserror::Error)]
 pub enum ForkedBlockchainError {
     /// Remote block creation error
--- a/crates/edr_provider/src/error.rs
+++ b/crates/edr_provider/src/error.rs
@@ -66,7 +66,7 @@ pub enum ProviderError<LoggerErrorT> {
         latest_block_number: u64,
     },
     /// The block tag is not allowed in pre-merge hardforks.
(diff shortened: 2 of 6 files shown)


crates/edr_evm/src/blockchain/forked.rs
```diff
@@ -55,7 +55,7 @@ pub enum CreationError {
     },
 }
 
-/// Error type for [`RemoteBlockchain`].
+/// Error type for `RemoteBlockchain`.
```
fvictorio, line 58: To properly fix this, I had to make the `remote` module public, which in turn meant documenting it and its public things... so I just removed the link here.

crates/edr_provider/src/error.rs
```diff
@@ -305,7 +305,7 @@ impl std::fmt::Display for TransactionFailureWithTraces {
     }
 }
 
-/// Wrapper around [`revm_primitives::Halt`] to convert error messages to match
+/// Wrapper around [`ExecutionResult::Halt`] to convert error messages to match
```
fvictorio, line 308: I'm not 100% sure about this change.

crates/edr_evm/src/blockchain/forked.rs
Wodann, line 58:
```suggestion
/// Error type for [`ForkedBlockchain`].
```

crates/edr_provider/src/error.rs
Wodann, line 308:
```suggestion
/// Wrapper around [`revm_primitives::HaltReason`] to convert error messages to match
```

Wodann (changes requested):

crates/edr_evm/src/blockchain/forked.rs
Wodann: See https://github.com/NomicFoundation/hardhat/pull/5074#discussion_r1551178330

crates/edr_provider/src/error.rs
Wodann: See https://github.com/NomicFoundation/hardhat/pull/5074#discussion_r1551185667

fvictorio: Did this work for you locally? I think it didn't work for me because `revm_primitives` is not a dependency in `edr_provider`.

Wodann: I didn't test it locally. We re-export `revm_primitives::HaltReason` in `edr_evm`, so the local use statement is `edr_evm::HaltReason`. I didn't realise that doctest validates with local includes.

fvictorio: ```[`edr_evm::HaltReason`]``` does work.

Wodann: With that change, feel free to merge :)

Wodann (approved):
One pending change, but other than that it looks good to merge.

PR #5076, opened 2024-04-04
Check that error object doesn't have unexpected fields

fvictorio:
This PR just updates some tests with extra checks. I was expecting to reproduce a problem that @alcuadrado noticed in the Ignition tests, but I couldn't do it. But since these extra assertions are good to have anyway, I think we should include them.


kanej (approved):
This looks good - by which I mean that I am understanding it as a clarification of the spec for what is already happening.

PR #5083, opened 2024-04-04
Introduce a template package

alcuadrado:
This PR introduces a new template package with updated setups for node, typescript, `package.json`, `eslint`, etc.

It includes a few example files and tests, using `node:test`. It doesn't have any assertion library installed though, except for `expect-type` for type-level tests. We still need to pick one, or use `node:assert`, which we should use in the meantime.

Something that's important to note is that now imports, even ts imports, have to end in `.js`. TBH it doesn't look good, but it's the currently recommended approach to use TS with native esm.

This PR doesn't change the global config files, but introduces new ones instead, to avoid breaking all other packages.

### DIFF
--- a/packages/template-package/.eslintrc.cjs
+++ b/packages/template-package/.eslintrc.cjs
@@ -1,0 +1,3 @@
+const { createConfig } = require("../../config/v-next/eslint.cjs");
+
+module.exports = createConfig(__filename);
--- a/packages/template-package/.eslintrc.cjs
+++ b/packages/template-package/.eslintrc.cjs
@@ -1,3 +1,6 @@
 const { createConfig } = require("../../config/v-next/eslint.cjs");
 
-module.exports = createConfig(__filename);
+module.exports = createConfig(__filename, [
+  "src/index.ts",
+  "src/other-entry-point.ts",
+]);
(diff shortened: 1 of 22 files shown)


packages/template-package/.eslintrc.cjs
```diff
@@ -0,0 +1,3 @@
+const { createConfig } = require("../../config/v-next/eslint.cjs");
+
+module.exports = createConfig(__filename);
```
schaable, line 3: shouldn't the entry points under `exports` in the `package.json` be passed here as a second arg?

alcuadrado: done

schaable (approved):

packages/template-package/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "README.md"
+  ],
+  "devDependencies": {
+    "@nomicfoundation/eslint-plugin-hardhat-internal-rules": "workspace:^",
+    "@nomicfoundation/eslint-plugin-slow-imports": "workspace:^",
+    "@types/node": "^20.0.0",
+    "eslint": "8.57.0",
+    "eslint-config-prettier": "9.1.0",
+    "eslint-plugin-import": "2.29.1",
+    "eslint-plugin-no-only-tests": "3.1.0",
```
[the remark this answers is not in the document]
alcuadrado: No, as mocha is not used in that package.

PR #5084, opened 2024-04-05
CI fix

fvictorio:
The other failures seem to be related to Goerli, which will be fixed by #5068


agostbiro (approved):
Thanks @fvictorio , I fixed this in https://github.com/NomicFoundation/hardhat/pull/5068 as well

fvictorio:
Ah, in that case let's just merge that one, since it will fix the other errors too. Thanks!

PR #5085, opened 2024-04-05
refactor: retrieve genesis account info in smaller batches

Wodann:


Cargo.toml
```diff
@@ -9,3 +9,6 @@ rpath = true
 
 [profile.release]
 rpath = true
+
+[profile.test]
```
Wodann, line 13: This ensures that tests, in particular the full block tests, run faster.

PR #5086, opened 2024-04-05
fix: tests with async fn in describe

Wodann:
Every now and then, Hardhat tests with fail on one system for non-obvious reasons and I can't reproduce the problem (e.g. [this](https://github.com/NomicFoundation/hardhat/actions/runs/8568834063/job/23483577220?pr=5085#step:11:3020)).

I'm hoping that removing these erroneous usages of `async function`s in tests will avoid any issues due to race conditions


alcuadrado:
Great catch! I can confirm that all of these async describes have been removed from hardhat-core

alcuadrado (approved):

alcuadrado:
PS: I don't know if that would fix that issue, but those callbacks being async is weird, to say the least.

PR #5096, opened 2024-04-08
Add eslint rule to forbid describes with async functions

fvictorio:
Continuation of https://github.com/NomicFoundation/hardhat/pull/5086

Using this plugin also lets us remove the `eslint-plugin-no-only-tests` package and replace it with a rule from `eslint-plugin-mocha`.


kanej (approved):
LGTM. I have added a comment to our template package for v3 to updated the eslint rules to switch to `eslint-plugin-mocha` and add the async in describe rule.

Issue #4871, opened 2024-02-15
Clarify in the documentation that hardhat-chai-matchers is not compatible with hardhat-viem


schaable:
Currently, our `hardhat-chai-matchers` plugin is designed to work with `hardhat-ethers`. Attempting to use it in conjunction with `hardhat-viem` results in compatibility issues. To prevent confusion, we need to explicitly state this limitation in the documentation: [hardhat-chai-matchers Overview](https://hardhat.org/hardhat-chai-matchers/docs/overview).

anthonygood:
This bit me.

PR #5102, opened 2024-04-10
Clarify in the documentation that hardhat-chai-matchers is not compatible with hardhat-viem

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/4871)


ChristopherDedominici:
This is how it looks:
image

Do you think it should be moved in the "Overview" section? @schaable @kanej

schaable:
> Do you think it should be moved in the "Overview" section? @schaable @kanej

Yes, I'd put it on the overview section. A lot of users won't continue reading past the installation step 😂

schaable (approved):

kanej (approved):
LGTM!

Issue #4834, opened 2024-02-08
Re-enable macos CI tests


fvictorio:
We should temporarily re-enable the macos CI tests and check if the [timeouts problem](https://github.com/actions/runner-images/issues/4896) is still there.

PR #5103, opened 2024-04-10
Re-enable macos CI tests

ChristopherDedominici:
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/4834)


Wodann:
👀 

The MacOS failure with EDR seems to be the result of a combination of:
1. the machine having too little RAM; and
2. (potentially) GC holding on to objects from previous tests

schaable (approved):

fvictorio (approved):

Issue #5060, opened 2024-04-02
Refactor to utility package


kanej:
Create a utility package that abstracts some of our use of Node APIs in Hardhat.

We have not explicit goal to support alternate javascript runtimes (i.e. bun, deno), but as we move through HHv3 next work it would be useful to be explicit in our usage of Node APIs.

This refactor should be pragmatic - lets remove the low hanging fruit, and take the rest case by case.

Isolating Node APIs or depenedencies that leverage should help us be more intentional in our dependencies.

This work should be based off of the `Hardhat v3` branch.

## TODO

- [x] Create a new `utils` package meeting the Hardhat v3 conventions
- [x] Analyze usage of `fs` across the codebase, attempt to pull into utils package
- [x] Analyze usage of `fs-extra`, attempt to drop dependency, pull into utils package otherwise
- [x] Analyze usage of `glob`, attempt to drop dependency, pull into utils package otherwise
- [x] Analyze usage of `undici`, pull into utils package
- [x] Determine other potential dependencies that this approach would work for

alcuadrado:
A good way to think about this is: hardhat core should not install `@types/node`.

PR #5110, opened 2024-04-11
Hardhat utils

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/5060


packages/hardhat-utils/src/internal/request.ts
```diff
@@ -0,0 +1,129 @@
...
+  return new Agent(options);
+}
+
+export function getBaseDispatcherOptions(
+  timeout: number = DEFAULT_TIMEOUT_IN_MILLISECONDS,
+  isTestDispatcher: boolean = false,
+): UndiciT.Client.Options {
+  // These have good defaults for production, but need to be tweaked to avoid hanging tests.
+  // https://github.com/nodejs/undici/blob/961b76ad7cac17d23580d172702e11a080974f5d/docs/docs/best-practices/writing-tests.md
+  const keepAliveTimeouts = isTestDispatcher
```
schaable, line 113: I'm not sure about this approach. I think we should use a mock dispatcher in our tests, similar to what I've done in `packages/hardhat-utils/test/request.ts`. Maybe we could add a `getMockDispatcher` function in `packages/hardhat-utils/src/request.ts` to make sure the `MockAgent` is initialized with this options. @alcuadrado, wdyt?

packages/hardhat-utils/src/fs.ts
```diff
@@ -0,0 +1,546 @@
...
+
+/**
+ * Looks for a file in the current directory and its parents.
+ *
+ * @param fileName The name of the file to look for.
+ * @param from The directory to start the search from. Defaults to the current working directory.
+ * @returns The absolute path to the file, or `undefined` if it wasn't found.
+ */
+export async function findUp(
+  fileName: string,
```
schaable, line 529: I realized this will not fail if `fileName` is a path, but it may not behave as expected. I would add a check at the start and throw if a path is used.

packages/hardhat-utils/src/bigint.ts
```diff
@@ -0,0 +1,88 @@
...
+ */
+export async function toBigInt(
+  x: number | string | bigint | object,
+): Promise<bigint> {
+  switch (typeof x) {
+    case "number":
+      if (!Number.isInteger(x)) {
+        throw new BigIntError(`${x} is not an integer`);
+      }
+      if (!Number.isSafeInteger(x)) {
```
schaable, line 50: This doesn't make sense, as the way to handle unsafe integer is casting them to BigInt, which is what this function does.

packages/hardhat-utils/src/bigint.ts
```diff
@@ -0,0 +1,88 @@
...
+        throw new BigIntError(
+          `Integer ${x} is unsafe. Consider using ${x}n instead. For more details, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger`,
+        );
+      }
```
schaable, line 54: We should check that the number is not a floating point number.

PR #5113, opened 2024-04-12
Fix build with latest viem dependency

fvictorio:
In the latest version of viem, the `contractAddress` field of the receipt can be both `null` or `undefined`, which causes our build to fail (and the latest dependencies workflow to fail). This fixes that.

I don't know in which cases this `undefined` can show up though.


kanej (approved):
LGTM. I added a changeset on the grounds we should cut a release.

schaable (approved):

PR #5119, opened 2024-04-12
Remove deprecated packages and no-hoist @types/

alcuadrado:
This PR cleans up the `v-next` in two ways:

1. Deletes the deprecated plugins and their workflows.
    1.1. `common` can't be removed just yet, but at least I removed `ganache-cli` from its dependencies. 
2. Adds a setting so that pnpm doesn't hoist @types/ packages.


alcuadrado:
The reasoning behind (2) is that typescript automatically loads all the `@types/` it finds in a `node_modules`, so if they are hoisted all of them get to be loaded in every package.

kanej (approved):
LGTM.

PR #5127, opened 2024-04-16
fix: mining blocks when forking non-eth chains

Wodann:
Fixes https://github.com/NomicFoundation/edr/issues/384

### DIFF
--- a/crates/edr_provider/tests/issue_384.rs
+++ b/crates/edr_provider/tests/issue_384.rs
@@ -1,0 +1,27 @@
+#![cfg(feature = "test-utils")]
+
+use edr_provider::{
+    hardhat_rpc_types::ForkConfig, test_utils::create_test_config_with_fork, MethodInvocation,
+    NoopLogger, Provider, ProviderRequest,
+};
+use tokio::runtime;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn flare_network_mine_local_block() -> anyhow::Result<()> {
+    const BLOCK_NUMBER: u64 = 22_587_773;
+
+    let logger = Box::new(NoopLogger);
+    let subscriber = Box::new(|_event| {});
+
+    let config = create_test_config_with_fork(Some(ForkConfig {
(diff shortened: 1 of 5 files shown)


crates/edr_provider/tests/issue_384.rs
```diff
@@ -0,0 +1,27 @@
...
+
+#[tokio::test(flavor = "multi_thread")]
+async fn flare_network_mine_local_block() -> anyhow::Result<()> {
+    const BLOCK_NUMBER: u64 = 22_587_773;
+
+    let logger = Box::new(NoopLogger);
+    let subscriber = Box::new(|_event| {});
+
+    let config = create_test_config_with_fork(Some(ForkConfig {
+        json_rpc_url: "https://flare-api.flare.network/ext/bc/C/rpc".to_string(),
```
fvictorio, line 17: I think it would be better to do this, assuming our Infura endpoint support Avalanche:

```
        json_rpc_url: get_infura_url().replace("mainnet", "avalanche-mainnet"),
```

Public endpoints are more likely to fail, and that would generate false negatives for us.

fvictorio: I don't have access to the Infura token used by the CI, so the simplest thing here is to just try it out and see if the CI passes :man_shrugging:

Wodann: Thanks! Addressed by https://github.com/NomicFoundation/hardhat/pull/5127/commits/3f2390478c1004095673e9e2e658cee486f0e64b

TIL that Flare and Avalanche have common properties 😁

fvictorio (approved):

gianniskokTr:
after the 2.22.3 update I keep on getting the same error on forking avalanche c-chain mainnet, 

I use the infura RPC: `https://avalanche-mainnet.infura.io/v3/apiKey`

> Invalid parent hash: 0x39f65747b5fa30af4f2f4848a0c267adf90618dbf6cc8dc482345a87052d4c83. Expected: 0x6aa58036b7669d0b635c42800fcfd18aa37525cb59013ce086580e99b8a0a154.

PR #5139, opened 2024-04-17
fix: avoid transferring traces to JS

agostbiro:
This PR brings a fix to avoid transferring transaction traces to JS if there are no subscribers to consume this. With the performance improvements in the base branches this is where we spend now the bulk of the execution time, so it's a very significant improvement. Still, if there are errors, the transaction traces have to be transferred to JS to generate Solidity stack traces which remains very costly.

Due to the performance improvements the iteration count of some scenarios had to be updated to get consistent results.

The base branch should be merged first.


agostbiro:
Rebased on base branch to fix CI, but as a result the diff got slightly borked, please ignore the Rust parts, those come from the base branch.

Wodann (approved):
This LGTM, but I'll cede to @fvictorio for a second opinion as he's more experienced with the TS part

PR #5145, opened 2024-04-18
Don't ues batch requests in forking client.

kowalski:
The motivation is that Alchemy returns 200 response when batch requests hit rate limit. These errors are not parsed correctly so they end up failing entire process instead of being retried

---

This is a naive fix to issue  https://github.com/NomicFoundation/edr/issues/381
I'm sorry if it's too naive. This is literally the first code I've written in Rust ever. 
I wish I knew how to write a test, but I haven't figured out how to run them yet.
Hopefully this work is of help to you. If it's not, please disregard

### DIFF
--- a/crates/edr_eth/src/remote/client.rs
+++ b/crates/edr_eth/src/remote/client.rs
@@ -110,8 +110,6 @@ pub enum RpcClientError {
         method: Box<RequestMethod>,
         /// The id of the request iwth the missing response
         id: Id,
-        /// The response text
-        response: String,
     },
 
     /// The JSON-RPC returned an error.
@@ -645,14 +643,15 @@ impl RpcClient {
                 })
                 .collect())
         } else {
-            let request_body = SerializedRequest(
-                serde_json::to_value(&requests).map_err(RpcClientError::InvalidJsonRequest)?,
-            );
-            let remote_response = self.send_request_body(&request_body).await?;
-            let remote_responses: Vec<jsonrpc::Response<serde_json::Value>> =
-                Self::parse_response_str(&remote_response)?;
+            for request in requests {
+                let request_body = SerializedRequest(
+                    serde_json::to_value(&request).map_err(RpcClientError::InvalidJsonRequest)?,
+                );
 
-            for response in remote_responses {
+                let remote_response = self.send_request_body(&request_body).await?;
+                let response: jsonrpc::Response<serde_json::Value> =
+                    Self::parse_response_str(&remote_response)?;
+
                 let index = id_to_index
                     // Remove to make sure no duplicate ids in response
                     .remove(&response.id)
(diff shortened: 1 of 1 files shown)


fvictorio:
Hey @kowalski, thanks for this. I think there are some issues with this approach:

1. The transactions are sent sequentially, which can affect performance during forking.
2. Instead of changing the implementation of the `batch_call` function, I think it would be better to remove this function and to handle this on a case-by-case basis in each usage. The reason for this is that the genesis accounts initialization will make a lot of requests that we probably want to limit somehow.

We were planning to do this, but if you want to give it a try, be my guest! Happy to help with any questions you might have.

kowalski:
Hi @fvictorio,

Thank you for your comment. I definitely agree that my diff is more a hack than an actual fix. I figured this is what I can do with my very limited knowledge of Rust. Scratch that, I have no knowledge, just an intuition of sorts.

I only found the following usages of batch_request:
* In `fetch_fork_metadata`
* In `get_account_info`
* In `get_transaction_receipts`

What I'm not certain of is how napi works and if maybe this method is exposed to nodejs layer and getting called from there. 

In any case, are you sure you want to remove it? You've mentioned you don't want these calls to made in a sequential way. If that's the case you do want to keep a method that receive a vector of requests, returns a vector of results and inside uses `join!` to make requests in parallel? Because if that's it, I could
* change implementation to use `join!`
* change name from `batch_call` to something like `call_in_parallel`, so that it doesn't suggest that it's actually using jsonrpc batch

> Happy to help with any questions you might have.

I actually have two questions:
* what command do you use to run tests?
* I can see you using tracing. How can see outputs of traces and maybe add my own logs for debugging (`println`?)

Also please, feel free to ignore these. I don't want to use your resources. I think there is a high chance that this is less time for you to make changes yourself than to spend time explaining stuff to me and reviewing my changes.

fvictorio:
> what command do you use to run tests?

`cargo test --workspace --all-targets --all-features` from the root is what we run in the CI.

> I can see you using tracing. How can see outputs of traces and maybe add my own logs for debugging (println?)

I don't know how to see the outputs of traces tbh. @Wodann should we add a section to the book about this?

With respect to adding logs: just adding `println!` statements should be enough.

> I think there is a high chance that this is less time for you to make changes yourself than to spend time explaining stuff to me and reviewing my changes.

I think this is the case for this change, mainly because there are some non-obvious things to decide. Also, we are in the middle of moving the code to the `NomicFoundation/edr` repo, so ideally this would wait until we do that anyway :sweat_smile:

kowalski:
:(
Honestly I've hoped so muuuch you guys would fix it soon.
We are using forking chain as a source of data in EVM Debugger (https://www.rumblefish.dev/evm-debugger/). We use it to generate trace logs which are then subject to complex analysis and a bit of magic.

Since Cancun upgrade our tool doesn't work for some new transactions. It's clear that we need to upgrade hardhat, however we run into the issue with batch requests. Historically we depended on patching hardhat to detect Alchemy's 429 responses, but now, that it's rewritten to Rust it's a bit more complicated to get my patched implementation in.

At 10th of May I'm giving a public presentation of EVM Debugger in ETHBratislava (https://www.ethbratislava.com/). Understandingly I'm slightly anxious about our ability to get debugger working again for new transactions.

fvictorio:
Ugh, sorry about that. This issue is very high priority for us though, so we should have a release fixing it out there soon

PR #5152, opened 2024-04-19
use absolute path for foundry sources

### LINKED ISSUE #5043: Format of `hre.config.paths.sources` changes when `hardhat-foundry` is in use
### Version of Hardhat

2.22.2

### What happened?

The `hardhat-foundry` package changes `hre.config.paths.sources` from an absolute path to a relative path.  For example, without the plugin the path is "/home/user/hardhat-project/contracts", but with the plugin it's just "contracts".  This has caused some issues with one of my plugins (https://github.com/ItsNickBarry/hardhat-dependency-compiler/issues/16).

Check the value of `hre.config.paths.sources`.  Install `@nomicfoundation/hardhat-foundry`.  Check the value of `hre.config.paths.sources` again.

### Search terms

foundry, hardhat-foundry, sources, path, absolute, relative

ItsNickBarry:
closes #5043

Not sure how to test this because the plugin directory is not a Hardhat project.

### DIFF
--- a/.changeset/cyan-cheetahs-add.md
+++ b/.changeset/cyan-cheetahs-add.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-foundry": patch
+---
+
+Use absolute rather than relative path when overriding sources configuration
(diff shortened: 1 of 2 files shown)


alcuadrado (approved):
Thanks nick!

PR #5153, opened 2024-04-19
Add `BigInt` task argument type

### LINKED ISSUE #5089: HH301: Invalid value 1 for argument value of type int
### Version of Hardhat

2.22.2

### What happened?

`BigInt` types cannot be passed to tasks.

Maybe this is more of a feature request, but at the very least the error message itself is a bug, because `1` is clearly an `int`.

```
hh console
```

```
const { task, types } = require('hardhat/config');
task('err').addOptionalParam('value', '', 0n, types.int).setAction(async () => {});
await hre.run('err', { value: 1n });
```

### Search terms

_No response_

ItsNickBarry:
closes #5089

### DIFF
--- a/packages/hardhat-core/src/internal/core/params/argumentTypes.ts
+++ b/packages/hardhat-core/src/internal/core/params/argumentTypes.ts
@@ -122,6 +122,50 @@ export const int: CLIArgumentType<number> = {
 };
 
 /**
+ * BigInt type.
+ * Accepts either a decimal string integer or hexadecimal string integer.
+ * @throws HH301
+ */
+export const bigint: CLIArgumentType<bigint> = {
+  name: "bigint",
+  parse: (argName, strValue) => {
+    const decimalPattern = /^\d+(?:n)?$/;
+    const hexPattern = /^0[xX][\dABCDEabcde]+$/;
+
+    if (
+      strValue.match(decimalPattern) === null &&
+      strValue.match(hexPattern) === null
+    ) {
+      throw new HardhatError(ERRORS.ARGUMENTS.INVALID_VALUE_FOR_TYPE, {
+        value: strValue,
+        name: argName,
+        type: bigint.name,
+      });
+    }
+
+    return BigInt(strValue.replace("n", ""));
+  },
+  /**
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/src/internal/core/params/argumentTypes.ts
```diff
@@ -121,6 +121,50 @@ export const int: CLIArgumentType<number> = {
...
+  /**
+   * Check if argument value is of type "bigint".
+   *
+   * @param argName {string} argument's name - used for context in case of error.
+   * @param value {any} argument's value to validate.
+   *
+   * @throws HH301 if value is not of type "bigint"
+   */
+  validate: (argName: string, value: any): void => {
+    const isBigInt = typeof value === "bigint";
```
ItsNickBarry, line 157: Should we allow inputs of type `number` as long as they're integers and less than `MAX_SAFE_INTEGER`?

ItsNickBarry: I suppose not.  That behavior would be consistent with some interfaces, but is not that intuitive and might cause problems.

ItsNickBarry: Do we also need a `number` type?  There's a great deal of inconsistency in how numbers are handled across libraries.  I'd like to be able to call tasks from scripts without worrying about what type of number I"m passing in (and for `validate` to check that regular JS numbers are `< MAX_SAFE_INTEGER`).

alcuadrado: I'm not sure if I understand the `number` type idea.

alcuadrado: I see. I think I'd personally use `bigint`, but I wouldn't be against adding a new type.

alcuadrado (approved):

alcuadrado:
Thanks again, @ItsNickBarry !

I'm merging despite the EDR test failures, as those are just because of a CI setting issue.

PR #5157, opened 2024-04-22
fix: use `eth_maxPriorityFeePerGas` for determing gas

### LINKED ISSUE #5093: provider.getFeeData().maxPriorityFeePerGas vs eth_maxPriorityFeePerGas different values
### Version of Hardhat

2.20

### What happened?

Why the value of maxPriorityFeePerGas are different by using these two methods?
I tested it on the Ethereum mainnet

1. config hardhat network:
```
  networks: {
    mainnet: {
      url: "https://eth.llamarpc.com",
    },
  }
```

2. goto hardhat console: `npx hardhat console --network mainnet`

3. using following codes:
```typescript
const p = new ethers.JsonRpcProvider("https://eth.llamarpc.com")

console.log([(await ethers.provider.getFeeData()), Number(await ethers.provider.send('eth_maxPriorityFeePerGas', []))]);
console.log([(await p.getFeeData()), Number(await p.send('eth_maxPriorityFeePerGas', []))]);
```
output:
```
> [
  FeeData {
    gasPrice: 16951633560n,
    maxFeePerGas: 34883267120n,
    maxPriorityFeePerGas: 1000000000n
  },
  10000000
]


> [
  FeeData {
    gasPrice: 16951633560n,
    maxFeePerGas: 33893267120n,
    maxPriorityFeePerGas: 10000000n
  },
  10000000
]
```

You can see 'maxPriorityFeePerGas' in 'getFeeData' is different from the value of 'eth_maxPriorityFeePerGas' by using ethers.provider.

### Search terms

_No response_

kanej:
Update `getFeeData` to match the `ethers` algorithm, specifically don't just assume `1_000_000_000n` as the `maxPriorityFeePerGas`. Instead call `eth_maxPriorityFeePerGas`, and only fall back to `1_000_000_000n`, if there is no response.

Fixes #5093

### DIFF
--- a/packages/hardhat-ethers/test/hardhat-ethers-provider.ts
+++ b/packages/hardhat-ethers/test/hardhat-ethers-provider.ts
@@ -1,6 +1,8 @@
 import { assert, use } from "chai";
 import chaiAsPromised from "chai-as-promised";
 
+import { EthereumProvider } from "hardhat/types/provider";
+import { HardhatEthersProvider } from "../src/internal/hardhat-ethers-provider";
 import { ExampleContract, EXAMPLE_CONTRACT } from "./example-contracts";
 import { usePersistentEnvironment } from "./environment";
 import { assertIsNotNull, assertWithin } from "./helpers";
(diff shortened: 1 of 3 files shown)


packages/hardhat-ethers/test/hardhat-ethers-provider.ts
```diff
@@ -117,12 +119,83 @@ describe("hardhat ethers provider", function () {
...
+      assert.typeOf(feeData.maxPriorityFeePerGas, "bigint");
+    });
+
+    // This helper overrides the send method of an EthereumProvider to allow
+    // altering the default Hardhat node's reported results.
+    function overrideSendOn(
+      provider: EthereumProvider,
+      sendOveride: (method: string, params?: any[] | undefined) => Promise<any>
+    ) {
+      return new Proxy(provider, {
```
kanej, line 137: My knowledge of providers is not high enough to know if there is a better pattern for overriding the RPC mechanism of the underlying Ethereum provider.
If there is a cleaner approach, let me know.

alcuadrado: I think we have a mock provider object somewhere, but let's move fwd with this.

alcuadrado (approved):

PR #5160, opened 2024-04-23
docs: ignition installation, add missing dependency

zk1st:


kanej:
An install of `npm install --save-dev @nomicfoundation/hardhat-ignition-ethers` on npm 7+ should install `@nomicfoundation/hardhat-ignition` automatically as it is a peer dependency:

https://github.com/NomicFoundation/hardhat-ignition/blob/2ece9520446adc495f9ad9c87ef5605e39fb62f5/packages/hardhat-plugin-ethers/package.json#L44

Is there something we are missing?

kanej:
Does this show up as a inline error/warning on the js import line? Is that what you mean by `vscode` cannot recognize the import?

zk1st:
> Does this show up as a inline error/warning on the js import line? Is that what you mean by `vscode` cannot recognize the import?

Yes, like this
image

I'm using `tsconfig.json`:
```
{
  "compilerOptions": {
    "target": "es2020",
    "module": "commonjs",
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "skipLibCheck": true,
    "resolveJsonModule": true
  }
}
```

kanej:
> > Does this show up as a inline error/warning on the js import line? Is that what you mean by `vscode` cannot recognize the import?
> 
> Yes, like this image
> 
> I'm using `tsconfig.json`:
> 
> ```
> {
>   "compilerOptions": {
>     "target": "es2020",
>     "module": "commonjs",
>     "esModuleInterop": true,
>     "forceConsistentCasingInFileNames": true,
>     "strict": true,
>     "skipLibCheck": true,
>     "resolveJsonModule": true
>   }
> }
> ```

I have tried reproducing this locally by:

1. create a new HH project with just a hardhat config
2. `npm install --save-dev hardhat`
3. `npm install --save-dev @nomicfoundation/hardhat-ignition-ethers`
4. `npm install --save-dev typescript`
5. replicating the tsconfig above
6. creating a simple Ignition module under `./ignition/modules/Simple.ts`

I don't see the vscode error.

Is it possible you are on a an older version of `npm`? So v6 or bellow?

Issue #5122, opened 2024-04-15
Enforce contract verification


rimrakhimov:
### Description

On contract verification, blockscout differentiates source codes on partial and full matches. In case of partially verified 
contracts re-verification of such contracts is allowed. But it is not possible to re-verify such contracts via Hardhat rn, as it always returns "The contract {contract_address} has already been verified on Etherscan."

### Suggestion

We suggest to add a new flag on `verify:verify` subtask - `--skip-is-verified-check`. That should disable the [`if(isVerified)`](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-verify/src/internal/tasks/etherscan.ts#L102) condition to prevent verification attempt.

A kind of similar approach has been adopted by foundry - https://github.com/foundry-rs/foundry/pull/6461.

I may try to create a PR with corresponding changes. Just would like to ensure that you are fine with suggested approach.

### Minimal Reproducible Steps
1. Create a new template project and update `Lock.sol` contract to have a contract named `ModifiedMetadataHashLock`.
2. config hardhat network:
```
const config: HardhatUserConfig = {
  solidity: "0.8.24",
  networks: {
    sepolia: {
      url: 'https://endpoints.omniatech.io/v1/eth/sepolia/public',
      accounts: [PRIVATE_KEY],
    },
  },
  etherscan: {
    apiKey: 'ETHERSCAN_API_KEY',
    customChains: [
      {
        network: "sepolia",
        chainId: 11155111,
        urls: {
          apiURL: "https://eth-sepolia.blockscout.com/api",
          browserURL: "https://eth-sepolia.blockscout.com/",
          // apiURL: "https://api-sepolia.etherscan.io/api",
          // browserURL: "https://sepolia.etherscan.io/",
        }
      }
    ]
  }
};
```
3. `npx hardhat verify 0xcA30DBB0Cb79cef0194f6e212B49fa1EB7246b61 --network sepolia 123`
4. Output (even though the contract has been verified only partially and can be re-verified):
```
The contract 0xcA30DBB0Cb79cef0194f6e212B49fa1EB7246b61 has already been verified on Etherscan.
https://eth-sepolia.blockscout.com/address/0xcA30DBB0Cb79cef0194f6e212B49fa1EB7246b61#code
```

kanej:
This is a feature we want to see, though we would suggest `--force` instead of `--skip-is-verified-check` as the flag.

PR #5164, opened 2024-04-24
Enforce contract verification

rimrakhimov:

---

That PR adds the `--force` verification flag described in https://github.com/NomicFoundation/hardhat/issues/5122, which allows the re-verification of already verified contracts (for blockscout-based explorers).

Notice that sometimes, an underlying explorer may return an error saying that the contract has already been verified. For blockscout-based explorers that happens if the contract is already verified with a [full match](https://docs.sourcify.dev/docs/full-vs-partial-match/), while for etherscan that is the only possible option for verified contracts. The PR considers those cases as errors and fails with a corresponding error message. I don't know if that is what is expected, though.

Also I plan to add some test cases, but would be glad to get some suggestions to start with

### DIFF
--- a/.changeset/pink-goats-jam.md
+++ b/.changeset/pink-goats-jam.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
+Added 'force' flag to allow verification of partially verified contracts (thanks @rimrakhimov!)
--- a/packages/hardhat-verify/src/index.ts
+++ b/packages/hardhat-verify/src/index.ts
@@ -52,6 +52,7 @@ export interface VerifyTaskArgs {
   constructorArgs?: string;
   libraries?: string;
   contract?: string;
+  force: boolean;
   listNetworks: boolean;
 }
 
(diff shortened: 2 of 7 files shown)


schaable (approved):

schaable:
Hi @rimrakhimov, the code looks good. I've updated the error message to include the contract FQN, as other error messages also do this. I'll do some manual testing tomorrow. Regarding adding tests, you could add some integration tests by mocking the block explorer's response in `packages/hardhat-verify/test/integration/index.ts`. I suggest adding a test where `etherscan.isVerified` is mocked to return `true` and includes the `--force` flag, and another test that throws the `ContractAlreadyVerifiedError` exception.

rimrakhimov:
Hi, @schaable, thank you for the feedback. I added some test cases as you'd suggested. Also, found a bug in the implementation related to etherscan returning different "already verified" messages depending on whether the contract has been actually verified by some user (e.g., [0x7b45A6d08F26ecAD9422fb85FbCAb1a0fec806CB](https://sepolia.etherscan.io/address/0x7b45A6d08F26ecAD9422fb85FbCAb1a0fec806CB#code)), or has been automatically found verified contract with the same deployed bytecode (e.g., [0x9498322CB2193Bc906d2dF94D7C04A17A59790C8](https://sepolia.etherscan.io/address/0x9498322CB2193Bc906d2dF94D7C04A17A59790C8#code))

schaable:
Manual testing:

- [x] Unverified contract: 
   - Without `--force`: contract verified.
   - With `--force`: contract verified.
- [x] Verified contract (full match):
   - Without `--force`: return with message `The contract ${address} has already been verified on the block explorer`.
   - With `--force`: throw `ContractAlreadyVerifiedError` error.
- [x] Verified contract (partial match - only Blockscout):
   - Without `--force`: return with message `The contract ${address} has already been verified on the block explorer`.
   - With `--force`: contract verified.

PR #5171, opened 2024-04-27
fix(ts): import type adjustments

sambacha:
these are only used as types

---


packages/hardhat-core/sample-projects/typescript-viem/hardhat.config.ts
```diff
@@ -1,4 +1,5 @@
-import { HardhatUserConfig } from "hardhat/config";
+// @ts-check
```
kanej, line 1: Hey @sambacha, can you walk me through the advantage of adding `// @ts-check` here?

alcuadrado, line 1:
```suggestion
```

PR #5174, opened 2024-04-30
Guide for using ignition with hardhat scripts

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat-ignition/issues/737

### DIFF
--- a/docs/src/content/ignition/docs/guides/_dirinfo.yaml
+++ b/docs/src/content/ignition/docs/guides/_dirinfo.yaml
@@ -10,3 +10,4 @@ order:
   - /verify
   - /viem
   - /create2
+  - /scripts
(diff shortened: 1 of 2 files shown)


schaable (approved):
LGTM

docs/src/content/ignition/docs/guides/_dirinfo.yaml
```diff
@@ -10,3 +10,4 @@ order:
   - /verify
   - /viem
   - /create2
+  - /scripts
```
kanej, line 13: Lets move this one under the advanced section. Or maybe a rename that excludes the word advanced?

docs/src/content/ignition/docs/guides/scripts.md
```diff
@@ -0,0 +1,241 @@
+# Advanced deployment with Hardhat scripts
```
kanej, line 1: If we move to the advanced section maybe a name like "Deploying within Hardhat scripts"

kanej (changes requested):
I really like this, it will help massively as a reference for helping users with support requests.
I think it is arguable whether the conditional example is needed, but I suspect you are thinking in terms of having an exact example to quote when someone says "How do I do a condinitional in a module?".

kanej (approved):
LGTM!

PR #5177, opened 2024-05-03
Remove EDR changesets

fvictorio:
These were added before the repo migration, and they were already released.

The one that is modified comes from an optimization on the Hardhat side, which hasn't been released yet.


kanej (approved):
LGTM!

PR #5184, opened 2024-05-07
Update dependencies

alcuadrado:
This PR updates the lockfile to make and workarounds some superfluous typescript types. I suspect they are instances of https://github.com/microsoft/TypeScript/issues/51434


alcuadrado:
Merging without approval as this is blocking progress on `v-next`.

PR #5193, opened 2024-05-09
Esmify compilation against vnext

kanej:
This is a test branch to trigger compilation.


v-next/hardhat-build-system/src/internal/builtin-tasks/utils/solidity-files-cache.ts
```diff
@@ -0,0 +1,135 @@
+import type { SolcConfig } from "../../types/index.js";
+import debug from "debug";
+import fsExtra from "fs-extra";
+import * as t from "io-ts";
+import { deepEqual } from "fast-equals";
```
kanej, line 5: To pull this in directly from utils would mean making the chain of function calls above the cache `async`, is that what we intend with an async deepEqual?

v-next/hardhat-build-system/src/internal/solidity/compiler/adm-zip.d.ts
```diff
@@ -0,0 +1 @@
+declare module "adm-zip";
```
kanej, line 1: This appeared to be necessary once we moved to `v-next`. I will look again at why we are not picking up the types.

v-next/hardhat-build-system/src/internal/builtin-tasks/utils/solidity-files-cache.ts
schaable: This is probably a question for @alcuadrado, but I believe we should be fine. The asynchronicity comes from dynamically importing the `fast-equals` package inside `deepEqual`. However, after the initial load, it is cached, so subsequent loads should be fast. The only potential disadvantage is the extra boilerplate required by all the `async`/`await` statements.

v-next/hardhat-build-system/src/internal/errors/errors.ts
```diff
@@ -0,0 +1,193 @@
...
+      );
+    }
+
+    template = replaceAll(template, variableTag, value);
+  }
+
+  return template;
+}
+
+export function assertHardhatInvariant(
```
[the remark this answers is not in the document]
kanej: It is now in the Hardhat Errors package, we should pull it from there.

v-next/hardhat-build-system/src/internal/builtin-tasks/utils/solidity-files-cache.ts
alcuadrado: We discussed this and decided to defer the decision

PR #5199, opened 2024-05-12
feat: check check constructor args type

fwx5618177:

---

1. Added detection of parameters when passing arguments to the constructor. Currently, the detection is done on the backend, which makes it difficult to distinguish between strings and numbers in the tool. Therefore, a new feature has been submitted to address this issue.

This is the contract code:
```solidity
pragma solidity ^0.8.20;

contract MyContract {

    string public url;
    uint256 public add;

    constructor(string memory _url, uint256 _add) {
        url = _url;
        add = _add;
    }

    function setUrl(string memory _url) public {
        url = _url;
    }
}
```

Parameter passing file:
```js
module.exports = [
    123,
    123
];
```

Using the command:
```bash
npx hardhat verify --contract contracts/MyToken.sol:MyContract [ADDRESS] --constructor-args scripts/arguments.js
```

### DIFF
--- a/.changeset/thick-beers-tie.md
+++ b/.changeset/thick-beers-tie.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
+Improved validation of contructor arguments
--- a/.changeset/thick-beers-tie.md
+++ b/.changeset/thick-beers-tie.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/hardhat-verify": patch
 ---
 
-Improved validation of contructor arguments
+Improved validation of contructor arguments (thanks @fwx5618177!)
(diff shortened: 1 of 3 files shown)


fwx5618177:
@kanej  Maybe this pr not feature, it's fix?

schaable (approved):
This is a nice check when the constructor arguments are provided via a file or when the task is called programmatically. However, it doesn't have any effect when the arguments are passed through the CLI, as they are interpreted as strings. I've cleaned up the code, added a test, and included the changeset.

image

fwx5618177:
> This is a nice check when the constructor arguments are provided via a file or when the task is called programmatically. However, it doesn't have any effect when the arguments are passed through the CLI, as they are interpreted as strings. I've cleaned up the code, added a test, and included the changeset.
> 
> image

Got it. Maybe We need to transform string to a number and check if it's NaN or not.

schaable:
> Got it. Maybe We need to transform string to a number and check if it's NaN or not.

I don't think we need to, as the verification succeeds anyway.

Issue #5166, opened 2024-04-24
Remove "please report to the Hardhat team" from `hardhat-verify`


kanej:
We see lots of issues with `hardhat-verify` that are really temporary issues with the block explorer or an odd setup, or just a block explorer that doesn't meet the expected api. However due to our error message, it is reported as an issue:

```
Etherscan:
An unexpected error occurred during the verification process.
Please report this issue to the Hardhat team.
Error Details: Unexpected token '<', "

<!DOCTYPE "... is not valid JSON
```

To give an example:
https://github.com/NomicFoundation/hardhat/issues/5111

We should alter this error message to:

1. Remove the request to report the issue
2. If we can _pass through_ any more details from the underlying error/block explorer, that we do that

In particular, if we could display more of the failed http response (particularly if it is an HTML file) we are likely to get less support requests - maybe include the entirety of the failed http response in debug for instance.

Our goal here is to try and minimize the support burden of issues in the block explorers taking up Hardhat investigation time.

kanej:
Talking it through some more. Etherscan is providing only that the verification failed, no further details. Other block explorers might throw an http error and return an html error page.

So lets look at
* Change the error to indicate:
  - verification worked locally, this is not a HH issue
  - your next steps are to work with the block explorer directly, as they are giving no more details
  - if it is an http error rather than a 200 but failed verification we should investigate what printing the entire output would look like.

PR #5203, opened 2024-05-13
Hardhat verify fixes

schaable:
**Changes in this PR:**

  - Removed `UnexpectedError` from `hardhat-verify`.
  - Functions that make API calls are no longer enclosed in a broad try/catch block. Instead, calls to `undici` are specifically wrapped, and now throw a new `NetworkRequestError`.
  - Replaced generic `Error` instances with `VerificationAPIUnexpectedMessageError` to better indicate API-related issues. The error message no longer instruct users to report issues to the Hardhat team.
  - Added a correction in `HardhatVerifyError` to address an problem introduced by `NomicLabsHardhatPluginError`. This fix resolves an issue causing the `instanceof` operator to work incorrectly.

Closes https://github.com/NomicFoundation/hardhat/issues/5166

### DIFF
--- a/packages/hardhat-verify/src/internal/errors.ts
+++ b/packages/hardhat-verify/src/internal/errors.ts
@@ -9,6 +9,7 @@ import { TASK_VERIFY_VERIFY } from "./task-names";
 export class HardhatVerifyError extends NomicLabsHardhatPluginError {
   constructor(message: string, parent?: Error) {
     super("@nomicfoundation/hardhat-verify", message, parent);
+    Object.setPrototypeOf(this, this.constructor.prototype);
   }
 }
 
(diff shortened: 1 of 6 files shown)


packages/hardhat-verify/src/internal/errors.ts
```diff
@@ -9,6 +9,7 @@ import { TASK_VERIFY_VERIFY } from "./task-names";
 export class HardhatVerifyError extends NomicLabsHardhatPluginError {
   constructor(message: string, parent?: Error) {
     super("@nomicfoundation/hardhat-verify", message, parent);
+    Object.setPrototypeOf(this, this.constructor.prototype);
```
kanej, line 12: This is to enable `instanceOf` for code that is catching the error, or is there something else this enables?

packages/hardhat-verify/src/internal/errors.ts
```diff
@@ -421,25 +422,23 @@ Encoder error reason: ${reason} fault in ${operation}`,
...
-      e instanceof Error
-        ? e.message ?? defaultErrorDetails
-        : defaultErrorDetails;
-    super(`An unexpected error occurred during the verification process.
-Please report this issue to the Hardhat team.
-Error Details: ${errorDetails}`);
+export class NetworkRequestError extends HardhatVerifyError {
+  constructor(e: Error) {
+    super(
+      `A network request failed. This error is not related to Hardhat. Original error: ${e.message}`
```
kanej, line 440:
```suggestion
      `A network request failed. This is an error from the block explorer, not Hardhat. Error: ${e.message}`
```
Is block explorer the right term here? I think `Original error` makes more sense from our point of view, but less so from a user's perspective.

packages/hardhat-verify/src/internal/errors.ts
schaable: Yes, without this line, the `instanceof` operator doesn't work correctly for classes that extend `NomicLabsHardhatPluginError`

schaable: This is because the `NomicLabsHardhatPluginError` constructor is messing with the prototype chain: `Object.setPrototypeOf(this, NomicLabsHardhatPluginError.prototype);`.

kanej (approved):

PR #5204, opened 2024-05-13
[v-next] writeUtf8File: throw an error when a path is a directory and not a file

ChristopherDedominici:


schaable (approved):
Good catch! LGTM!

PR #5215, opened 2024-05-15
Remove EDR-related code

fvictorio:
This PR is of course impossible to review line by line. What I recommend doing is:
- Read the commit messages to get an idea of what I did.
- Check the files that were just modified. There aren't that many: 
  - `Brewfile`
  - `.gitignore`
  - `.github/workflows/hardhat-core-ci.yml`
  - `.github/workflows/add-issue-to-project.yml`
  - `.devcontainer/devcontainer.json`
  - `.changeset/config.json`
- Please pay special attention to the GitHub workflows change, in case I removed something I should've not, or that I forgot something.


fvictorio:
I forgot to mention: I considered removing the stack traces tests, but then I decided not to, because much of the logic for them still lives on the Hardhat side. If we remove them, then we are effectively freezing the related code, and I'm not sure we want to do that yet.

kanej (approved):
I took the commit at a time approach, and the moves made sense. Thanks.

schaable (approved):
All the changes make sense. I'm not sure if we have anything else to remove, as I'm not fully aware of all the ERD-related code. If we missed something, we can remove it later. The workflow changes look good to me. I have no idea about the dev container 😅

alcuadrado (approved):

PR #5218, opened 2024-05-15
Bump EDR and remove dependabot

fvictorio:
@kanej and I decided that we are going to bump EDR manually instead of relying on dependabot, and will use the changeset in those PRs to summarize the fixes/features of the new version.


kanej (approved):
Looks good.

PR #5220, opened 2024-05-15
chore/fix npm scripts escaping

kanej:
The escaping was not always working as intended on windows. Use a consistent quote escaping throughout the npm scripts.


schaable (approved):
LGTM

PR #5249, opened 2024-05-17
Better `node:test` reporter and scripts

alcuadrado:
This PR introduces a few changes to make working with `node:test` in `v-next` easier:

1. It changed the reporter from the builtin `spec` to our own test reporter.
2. It adds a `test:only` npm script, to run the tests with `.only`. Just run `pnpm test:only`.
3. It adds a CI job to ensure that all the npm scripts in `v-next/` are consistent with each other.
4. Removes `glob` from our peer dependencies, as `node:test` has native support for globs in node 22


scripts/check-v-next-npm-scripts.js
```diff
@@ -0,0 +1,56 @@
...
+  const packageJson = require(packageJsonPath);
+
+  for (const scriptName in templatePackageJson.scripts) {
+    if (scriptName === "clean") {
+      if (
+        !packageJson.scripts[scriptName].startsWith(
+          templatePackageJson.scripts[scriptName]
+        )
+      ) {
+        console.error(`Mismatch in script ${scriptName} in ${dir.name}`);
```
schaable, line 27: do you think we could use node assert here?

v-next/core/package.json
```diff
@@ -55,14 +56,14 @@
...
     "@typescript-eslint/parser": "^7.7.1",
+    "@voxpelli/node-test-pretty-reporter": "^1.1.1",
     "c8": "^9.1.0",
     "eslint": "8.57.0",
     "eslint-config-prettier": "9.1.0",
     "eslint-import-resolver-typescript": "^3.6.1",
     "eslint-plugin-import": "2.29.1",
     "eslint-plugin-no-only-tests": "3.1.0",
     "expect-type": "^0.19.0",
-    "glob": "^10.3.12",
```
schaable, line 65: so glob was not needed?

schaable:
we could add a linter rule to enforce [strict assertion mode](https://nodejs.org/api/assert.html#strict-assertion-mode) by importing `import assert from 'node:assert/strict';` instead of `import assert from 'node:assert';`

schaable (approved):

scripts/check-v-next-npm-scripts.js
alcuadrado: that leads to a single error being reported instead of all of them

v-next/core/package.json
alcuadrado: it was in node 20

alcuadrado:
Let's wait before merging this. See: https://github.com/voxpelli/node-test-pretty-reporter/issues/16

alcuadrado:
> we could add a linter rule to enforce [strict assertion mode](https://nodejs.org/api/assert.html#strict-assertion-mode) by importing `import assert from 'node:assert/strict';` instead of `import assert from 'node:assert';`

This was implemented in another PR

v-next/hardhat-build-system/package.json
```diff
@@ -19,9 +19,9 @@
     "lint:fix": "pnpm prettier --write && pnpm eslint --fix",
     "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
-    "test": "glob --cmd=\"node --import tsx/esm --test\" \"test/**/*.ts\"",
-    "test:github": "glob --cmd=\"node --import tsx/esm --test --test-reporter=@reporters/github --test-reporter-destination=stdout --test-reporter=spec --test-reporter-destination=stdout\" \"test/**/*.ts\"",
-    "test:coverage": "c8 --reporter html --reporter text --all --src src glob --cmd=\"node --import tsx/esm --test\" \"test/**/*.ts\"",
+    "test": "node --import tsx/esm --test --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/**/*.ts\"",
+    "test:only": "node --import tsx/esm --test --test-only --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/**/*.ts\"",
+    "test:github": "node --import tsx/esm --test --test-reporter=@reporters/github --test-reporter-destination=stdout --test-reporter=@nomicfoundation/hardhat-node-test-reporter --test-reporter-destination=stdout \"test/**/*.ts\"",
```
schaable, line 24: is this line correct?

v-next/hardhat-node-test-reporter/integration-tests/index.ts
```diff
@@ -1,24 +1,22 @@
+import { readFileSync, readdirSync, statSync } from "node:fs";
```
schaable, line 1: can you explain why we're using the sync versions?

schaable (commented):
Left a couple questions

v-next/hardhat-node-test-reporter/integration-tests/index.ts
alcuadrado: This doesn't run anything in parallel, so there's no value in using async versions, and it was just complicating a `filter()` call. If this were a test run by the actual test runner, I'd use async versions, but it's a top-level single-purpose script.

v-next/hardhat-build-system/package.json
alcuadrado: Not sure what exactly you mean by "this". Can you expand a bit?

v-next/hardhat-node-test-reporter/integration-tests/index.ts
schaable: I understand the call, but I'm uneasy about starting to use `node:fs` directly. If we add exceptions, this practice might spread, and we could miss out on good error handling and other checks. Do you think it's worth adding the sync versions to the utils?

v-next/hardhat-build-system/package.json
schaable: yes, sorry. The test:github has `--test-reporter` and `--test-reporter-destination` set twice. It looks like an error.

v-next/hardhat-node-test-reporter/integration-tests/index.ts
alcuadrado: Oh, I see. Your question is not about `sync`/`async`, but I guess you want to know why not the utils package. 

That's not possible because it would introduce a circular dependency between projects, and the typescript will complain. 

The reporter used to use the utils package, but as soon as I started using the reporter in the utils package, that error appeared.

I think it's better to have a nicer reporter in -utils than to avoid importing fs directly here.

v-next/hardhat-build-system/package.json
schaable: 🤯

v-next/hardhat-node-test-reporter/integration-tests/index.ts
schaable: Got it. Makes sense under that scenario 👍

schaable (approved):

Issue #4874, opened 2024-02-15
Make hardhat-chai-matchers compatible with hardhat-viem / library-agnostic


schaable:
The hardhat-chai-matchers plugin is currently only compatible with hardhat-ethers. This issue aims to see if there's interest in making it work with hardhat-viem and other Ethereum libraries. Please upvote or comment if you think this would be an useful feature.

dennis-gonzales:
viem is great imo, hopefully we can support this when able.

andreujuanc:
Would love to have this! much more cleaner with the matchers!

mwawrusch:
To add to that, hardhat-chain-matchers now uses the chaiAsPromised library which is not maintained anymore and is incompatible with chai 5. So it would make sense to upgrade / replace that as well while getting viem going.

petdomaa100:
+1

johnpm-12:
I put up a PR to add a viem specific package for chai matchers: #5252 

In case anyone finds it useful, in the meantime, I deployed it here: https://www.npmjs.com/package/hardhat-chai-matchers-viem

arkaydeus:
Really hope this is accepted soon. Very annoying not to be able to use viem

kanej:
Hey, we have released a new `viem` assertion plugin as part of the Hardhat 3 Alpha: https://hardhat.org/hardhat3-alpha

A minimal README.md for the new `@nomicfoundation/hardhat-viem-assertions` plugin can be found here: https://github.com/NomicFoundation/hardhat/tree/v-next/v-next/hardhat-viem-assertions

We are keen to get feedback and iterate on the assertion's plugin over the course of the Alpha.

kanej:
I am closing this issue, we have added a new `viem` assertion plugin as part of Hardhat 3: https://www.npmjs.com/package/@nomicfoundation/hardhat-viem-assertions

PR #5252, opened 2024-05-20
Feature: add chai matchers for viem

johnpm-12:

---

This adds a new package hardhat-chai-matchers-viem which is forked from hardhat-chai-matchers but supports viem instead of ethers.

I originally planned to update the existing ethers-based package to support both viem and ethers, but that proved to be difficult. So I made it a separate package, similar to how there is hardhat-toolbox and hardhat-toolbox-viem.

I understand that issue #4874 was to make the existing package compatible, so if that's still the plan, no skin off my back if ya'll decide not to merge this. I wrote this because I needed it internally. It was very time consuming. I pushed it to npm if anyone else finds it useful in the meantime: https://www.npmjs.com/package/hardhat-chai-matchers-viem

All existing tests have been updated to use viem instead of ethers, and all tests pass. However, viem with an external hardhat node does not return contract error data in some cases, so the tests to use an external hardhat node are skipped. All tests pass with the in-process hardhat node. I was going to parse the viem error to piece together what I can about the contract error, but I figured it would be better if the hardhat node or viem or hardhat-viem were updated to include the contract error data instead.

Linting succeeds.

I don't know what your process is for publishing to npm, so I'm not sure if anything needs to be done to add a new package there.

If you want to compare this package to the ethers one, you can use `git diff --no-index packages/hardhat-chai-matchers/src/index.ts packages/hardhat-chai-matchers-viem/src/index.ts` in the root folder, for example. VS Code also has a command to compare 2 local files.


alcuadrado:
Hey, thanks for submitting this pr! I'll review it asap and we can discuss how to move forward

johnpm-12:
Sounds good.

After using this for a few days, I think it might be worth updating the addressable stuff to lowercase (or convert to checksum) any address it find in a .equal or .not.equal matcher before comparison.
Since the original ethers version doesn't do that, I won't add it here, but worth discussing adding it to that package too.

It's annoying to have to use viem's getAddress() constantly (or .toLowerCase()) because RPC calls typically return lowercase addresses (contract addresses, wallet address, etc), while viem contract calls return checksummed addresses (contract.read.someAddress()).

TateB:
hey, i also wrote something very similar [here](https://github.com/ensdomains/hardhat-chai-matchers-viem) just last week!
my rewrite is significantly more opinionated than this so i think this is a much more suitable match for the existing matchers package, but it also includes significantly more advanced types so maybe worth looking at for ideas. i also added a `.toEqualAddress` matcher after running into the same checksummed address issue you mentioned.

johnpm-12:
Hah, wish I knew that before I started this!

At first glance, definitely some useful things in your more opinionated version.

Could always bring the good stuff from your version into the ethers one.
Keeping the ethers and viem ones in parity could get annoying.
I might take another crack at combining them, if I have the time

vrde:
👋 for a new project I switched from ethers to viem and I was expecting a better developer experience. So far it's good, but today I discovered that I cannot use the chai matchers I'm used to, specifically the ones to check if a tx reverted or if it emitted events.

Once it's ready, it would be great to see this PR merged, as it would help devs to better test their contracts (plus I don't want to go back to ethers only for this).

johnpm-12:
> 👋 for a new project I switched from ethers to viem and I was expecting a better developer experience. So far it's good, but today I discovered that I cannot use the chai matchers I'm used to, specifically the ones to check if a tx reverted or if it emitted events.
> 
> Once it's ready, it would be great to see this PR merged, as it would help devs to better test their contracts (plus I don't want to go back to ethers only for this).

You can use this for now: https://www.npmjs.com/package/hardhat-chai-matchers-viem

johnpm-12:
@alcuadrado any updates on this?

kanej:
Hey @johnpm-12, we are currently in depths of our Hardhat v3 work, a major overhaul.
How we approach viem test assertions, either taking this PR as the basis or a different approach is something we will have more visibility on once we are working through our new test subsystem.
Sorry for the delay in getting you a full response.

kanej:
Hey @johnpm-12, thanks for this PR - we have gone back and forth on how to best support viem assertions in Hardhat. The route we opted for was a new `viem` assertion plugin unconnected with chai matchers. So I am going to close this PR.

The new `@nomicfoundation/hardhat-viem-assertions` is now available as part of the Hardhat 3 Alpha: https://hardhat.org/hardhat3-alpha

A minimal README.md plugin can be found here: https://github.com/NomicFoundation/hardhat/tree/v-next/v-next/hardhat-viem-assertions

PR #5255, opened 2024-05-20
test: add resolve-plugin-list tests

### LINKED ISSUE #5254: Polish and test the plugin resolution/ordering logic
Provide tests coverage for `./v-next/core/src/plugins/resolve-plugin-list.ts`.

kanej:
Add tests on the dependency resolution of plugins.

Resolve #5254.

### DIFF
--- a/v-next/core/test/plugins/resolve-plugin-list.ts
+++ b/v-next/core/test/plugins/resolve-plugin-list.ts
@@ -1,6 +1,8 @@
 import assert from "node:assert";
 import { describe, it } from "node:test";
 
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
+
 import { resolvePluginList } from "../../src/internal/plugins/resolve-plugin-list.js";
 import { HardhatPlugin } from "../../src/types/plugins.js";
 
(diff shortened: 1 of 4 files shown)


v-next/core/test/plugins/resolve-plugin-list.ts
```diff
@@ -0,0 +1,119 @@
...
+      b,
+    ]);
+  });
+
+  it("should error on different plugins with the same id", () => {
+    const a = { id: "dup" };
+    const copy = { id: "dup" };
+
+    assert.throws(
+      () => resolvePluginList([a, copy]),
```
alcuadrado, line 114: It should throw a `HardhatError` and test that instead of the string.

alcuadrado:
This looks good. I left a minor comment, but otherwise looks good.

v-next/core/test/plugins/resolve-plugin-list.ts
kanej: I added the Hardhat error for duplicate plugins under `GENERAL` in the error descriptors. Let me know if that is not the category you were expecting.

alcuadrado (approved):

PR #5261, opened 2024-05-22
feat: plugin install validation

### LINKED ISSUE #5256: Implement and test the plugin validation logic
Provide an implementation of plugin validation in: `./v-next/core/src/internal/plugins/plugin-validation.ts`

If the plugin is from an npm package, validate their peer dependencies, if any are missing we throw throw.

kanej:
Validates the plugins provided through the initial config object.

If the plugin comes from an npm package check that the package has been installed.

If the plugin is present further confirm that peer dependencies have been installed. It then checks that the installed version falls within the range of the originating package.

This leverages the node require mechanism based on `process.cwd()` as the starting directory.

Resolves #5256

## Review Notes

While 21 files are changed, most of these are fixture files for the small number of tests. The core change is in:

* `v-next/core/src/internal/plugins/plugin-validation.ts`
* `v-next/core/test/plugins/plugin-validation.ts`

### DIFF
--- a/v-next/core/package.json
+++ b/v-next/core/package.json
@@ -53,6 +53,7 @@
     "@nomicfoundation/eslint-plugin-slow-imports": "workspace:^",
     "@reporters/github": "^1.7.0",
     "@types/node": "^20.0.0",
+    "@types/semver": "^6.2.7",
     "@typescript-eslint/eslint-plugin": "^7.7.1",
     "@typescript-eslint/parser": "^7.7.1",
     "c8": "^9.1.0",
@@ -73,6 +74,7 @@
   "dependencies": {
     "@nomicfoundation/hardhat-errors": "workspace:^3.0.0",
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.0",
-    "chalk": "^5.3.0"
+    "chalk": "^5.3.0",
+    "semver": "^6.3.0"
   }
 }
--- a/v-next/core/package.json
+++ b/v-next/core/package.json
@@ -53,7 +53,7 @@
     "@nomicfoundation/eslint-plugin-slow-imports": "workspace:^",
     "@reporters/github": "^1.7.0",
     "@types/node": "^20.0.0",
-    "@types/semver": "^6.2.7",
+    "@types/semver": "^7.5.8",
     "@typescript-eslint/eslint-plugin": "^7.7.1",
     "@typescript-eslint/parser": "^7.7.1",
     "c8": "^9.1.0",
@@ -75,6 +75,6 @@
     "@nomicfoundation/hardhat-errors": "workspace:^3.0.0",
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.0",
     "chalk": "^5.3.0",
-    "semver": "^6.3.0"
+    "semver": "^7.6.2"
   }
 }
--- a/v-next/core/src/internal/plugins/plugin-validation.ts
+++ b/v-next/core/src/internal/plugins/plugin-validation.ts
@@ -3,19 +3,10 @@ import path from "node:path";
 import process from "node:process";
 
 import { HardhatError } from "@nomicfoundation/hardhat-errors";
+import { PackageJson } from "@nomicfoundation/hardhat-utils/package";
 import semver from "semver";
 
 import { HardhatPlugin } from "../../types/plugins.js";
-
-/**
- * A simplified version of the package.json object
- */
-interface PackageJson {
-  version: string;
-  peerDependencies: {
-    [name: string]: string;
-  };
-}
 
 /**
  * Validate that a plugin is installed and that its peer dependencies are installed and satisfy the version constraints.
--- a/v-next/core/src/internal/plugins/plugin-validation.ts
+++ b/v-next/core/src/internal/plugins/plugin-validation.ts
@@ -26,7 +26,7 @@ export async function validatePluginNpmDependencies(
     return;
   }
 
-  const pluginPackageJson = readPackageJson(
+  const pluginPackageJson = readPackageJsonViaNodeRequire(
     plugin.npmPackage,
     basePathForNpmResolution,
   );
@@ -44,7 +44,7 @@ export async function validatePluginNpmDependencies(
   for (const [dependencyName, versionSpec] of Object.entries(
     pluginPackageJson.peerDependencies,
   )) {
-    const dependencyPackageJson = readPackageJson(
+    const dependencyPackageJson = readPackageJsonViaNodeRequire(
       dependencyName,
       basePathForNpmResolution,
     );
@@ -83,7 +83,7 @@ export async function validatePluginNpmDependencies(
  * @param baseRequirePath - the directory path to use for resolution, defaults to `process.cwd()`
  * @returns the package.json object or undefined if the package is not found
  */
-export function readPackageJson(
+function readPackageJsonViaNodeRequire(
   packageName: string,
   baseRequirePath?: string,
 ): PackageJson | undefined {
--- a/v-next/core/src/internal/plugins/plugin-validation.ts
+++ b/v-next/core/src/internal/plugins/plugin-validation.ts
@@ -13,6 +13,10 @@ import { HardhatPlugin } from "../../types/plugins.js";
  *
  * @param plugin - the plugin to be validated
  * @param basePathForNpmResolution - the directory path to use for node module resolution, defaulting to `process.cwd()`
+ * @throws {HardhatError} with descriptor:
+ * - {@link ERRORS.ARTIFACTS.PLUGIN_NOT_INSTALLED} if the plugin is not installed as an npm package
+ * - {@link ERRORS.ARTIFACTS.PLUGIN_MISSING_DEPENDENCY} if the plugin's package peer dependency is not installed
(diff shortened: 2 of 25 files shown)


v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -69,6 +69,11 @@ export const ERROR_CATEGORIES: {
     max: 1199,
     websiteTitle: "Contract name errors",
   },
+  PLUGINS: {
```
kanej, line 72: I added separate plugins section. I initially put these in general. Let me know if I should move it back.

v-next/core/src/internal/plugins/plugin-validation.ts
```diff
@@ -1,6 +1,110 @@
...
+ * @param plugin - the plugin to be validated
+ * @param basePathForNpmResolution - the directory path to use for node module resolution, defaulting to `process.cwd()`
+ */
+export async function validatePluginNpmDependencies(
+  plugin: HardhatPlugin,
+  // TODO: When all loads where based on a hardhat config file
+  // on disk we used the config file's directory as the base path
+  // We need to decide if the current working directory is the right
+  // option for module resolution now
+  basePathForNpmResolution?: string,
```
kanej, line 32: This is for testing currently. Though we need to consider passing in the npm resolution folder when we are loading the config file from disk.

v-next/core/package.json
```diff
@@ -73,6 +74,7 @@
   "dependencies": {
     "@nomicfoundation/hardhat-errors": "workspace:^3.0.0",
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.0",
-    "chalk": "^5.3.0"
+    "chalk": "^5.3.0",
+    "semver": "^6.3.0"
```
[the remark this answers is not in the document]
alcuadrado: We don't need to match does. `v-next` packages have their own set of versions.

v-next/hardhat-errors/src/descriptors.ts
alcuadrado: looks good

v-next/core/package.json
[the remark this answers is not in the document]
kanej: I have bumped the the `semver` and `@types/semver` to the latest in both `./v-next/core` and `./v-next/hardhat-build-system`.

ChristopherDedominici (approved):

ChristopherDedominici:
LGTM 😎

v-next/core/src/internal/plugins/plugin-validation.ts
```diff
@@ -1,6 +1,110 @@
...
+
 import { HardhatPlugin } from "../../types/plugins.js";
 
-export async function validatePluginNpmDependencies(_plugin: HardhatPlugin) {
-  // TODO: If it has an npm package, validate their peer dependencies
-  // If any is missing, throw
+/**
+ * A simplified version of the package.json object
+ */
+interface PackageJson {
```
schaable, line 13: you have this interface available in `hardhat-utils/package`

v-next/core/src/internal/plugins/plugin-validation.ts
```diff
@@ -1,6 +1,110 @@
...
+
+/**
+ * Read the package.json of a named package resolved through the node
+ * require system.
+ *
+ * @param packageName - the package name i.e. "@nomiclabs/hardhat-waffle"
+ * @param baseRequirePath - the directory path to use for resolution, defaults to `process.cwd()`
+ * @returns the package.json object or undefined if the package is not found
+ */
+export function readPackageJson(
```
schaable, line 95: how dare you! 😂 see `hardhat-utils/package`

v-next/core/test/plugins/plugin-validation.ts
```diff
@@ -0,0 +1,115 @@
...
+    );
+  });
+
+  describe("when the plugin has no peer deps", () => {
+    it("should pass validation if the npm package has been installed", async () => {
+      const installedPackageProjectFixture = import.meta.resolve(
+        "./fixture-projects/installed-package",
+      );
+
+      await assert.doesNotReject(async () =>
```
schaable, line 30: The Node docs actually recommend against using `doesNotReject` / `doesNotThrow`:

> Using assert.doesNotReject() is actually not useful because there is little benefit in catching a rejection and then rejecting it again. Instead, consider adding a comment next to the specific code path that should not reject and keep error messages as expressive as possible.

v-next/core/test/plugins/plugin-validation.ts
```diff
@@ -0,0 +1,115 @@
...
+            err.message,
+            'HHE1200: Plugin "example-plugin" is not installed.',
+          );
+
+          return true;
+        },
+        "Expected a plugin not installed error",
+      );
+    });
+  });
```
schaable, line 58:
```suggestion
    it("should fail validation if the npm package has not been installed", async () => {
      const nonInstalledPackageProjectFixture = import.meta.resolve(
        "./fixture-projects/not-installed-package",
      );

      await assert.rejects(
        async () =>
          validatePluginNpmDependencies(
            plugin,
            nonInstalledPackageProjectFixture,
          ),
        (err) => {
          assert(HardhatError.isHardhatError(err), "Expected a HardhatError");
          assert.equal(
            err.message,
            'HHE1200: Plugin "example-plugin" is not installed.',
          );

          return true;
        },
        "Expected a plugin not installed error",
      );
    });
  });
```

v-next/core/test/plugins/plugin-validation.ts
```diff
@@ -0,0 +1,115 @@
...
+          assert(HardhatError.isHardhatError(err), "Expected a HardhatError");
+          assert.equal(
+            err.message,
+            'HHE1201: Plugin "example-plugin" is missing a peer dependency "peer2".',
+          );
+
+          return true;
+        },
+        "Expected a missing peer dependency error",
+      );
```
schaable, line 89: You can simplify this:
```suggestion
      await assert.rejects(
        validatePluginNpmDependencies(plugin, notInstalledPeerDepFixture),
        {
          name: "HardhatError"
          message: 'HHE1201: Plugin "example-plugin" is missing a peer dependency "peer2".',
        }
      );
```

v-next/core/test/plugins/plugin-validation.ts
kanej: That is interesting. I was trying to find a pattern that allowed me to assert that an async function does not throw. What do we think the pattern should be in this case? Just a direct await on the function under test?

I think I prefer `assert.doesNotReject(...)` as its more explicit, but happy to hear if there are better patterns.

v-next/core/test/plugins/plugin-validation.ts
kanej: Oh that is nice.

v-next/core/src/internal/plugins/plugin-validation.ts
kanej: I have pulled it in directly.

v-next/core/src/internal/plugins/plugin-validation.ts
```diff
@@ -1,6 +1,101 @@
...
+  if (pluginPackageJson.peerDependencies === undefined) {
+    return;
+  }
+
+  for (const [dependencyName, versionSpec] of Object.entries(
+    pluginPackageJson.peerDependencies,
+  )) {
+    const dependencyPackageJson = readPackageJsonViaNodeRequire(
+      dependencyName,
+      basePathForNpmResolution,
```
alcuadrado, line 49: I think the base path here should be the directory of the plugin's package.json

v-next/core/src/internal/plugins/plugin-validation.ts
kanej: I have renamed this function `readPackageJsonViaNodeRequire` (and removed the export to make it private to validation).

Looking at the `hardhat-utils/package` versions they appear to be about finding a package.json up the directory tree from a particular file. Here the intent is to use the node require resolution logic to find a package in `node_modules` and read its `package.json`.

I suspect this use case is only ever going to be used in this validation, so would be tempted to leave it here until a second use case appears.

v-next/core/src/internal/plugins/plugin-validation.ts
alcuadrado: The reason for this is that it's the plugin's dependency. Normally it would also resolve from CWD, but to be correct, it should be resolvable from the plugin itself.

alcuadrado (approved):
I left a comment, the rest looks good

v-next/core/test/plugins/plugin-validation.ts
schaable: In the `hardhat-utils`, I think I left the code that shouldn't throw at the top level of the test with a comment. The message is pretty similar, to be honest. But I get your point about being explicit. Additionally, using the assertion could be better if we provide a custom `message` along with the assertion.

kanej: I am going to merge. For the moment, I have a mild preference for the explicit assertion style, but lets revisit if we find a better pattern or come to the conclusion we shouldn't be using it after some more experience.

v-next/core/src/internal/plugins/plugin-validation.ts
kanej: I have made this change and applied a test case with `./node_modules/example/node_modules/peer1` that fails if the original version is kept in place.

PR #5264, opened 2024-05-23
Add and test a singleton hre instance

### LINKED ISSUE #5236: Enhance HRE initialization within Hardhat
This task is to cover HRE enhancements and testing from within `./v-next/hardhat`.

## TODO

- [ ] Test that the builtin plugins are always included when initializing
- [ ] Implement and test a mechanism so that the index.ts (i.e.import {} from "hardhat") and the CLI share the same HRE instance (i.e. use a file as a singleton)
- [x] #5263

zoeyTM:
resolves #5236 

- added a singleton hre instance for both CLI and programmatic hardhat API
- added a test for the above
- added a test to ensure builtin plugins are included in initialization

### DIFF
--- a/v-next/hardhat/src/internal/hre-singleton.ts
+++ b/v-next/hardhat/src/internal/hre-singleton.ts
@@ -8,8 +8,10 @@ let hre: HardhatRuntimeEnvironment;
  * This function returns a singleton instance of the Hardhat Runtime Environment.
  *
  * It exists so that the CLI and the programmatic API are always using the same HRE instance.
+ *
+ * Note: Only the params of the first call are used.
  */
-export async function getHRE(
+export async function getHardhatRuntimeEnvironmentSingleton(
   ...params: Parameters<typeof createHardhatRuntimeEnvironment>
 ): Promise<HardhatRuntimeEnvironment> {
   if (hre === undefined) {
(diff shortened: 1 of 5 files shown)


v-next/hardhat/src/internal/hre-singleton.ts
```diff
@@ -0,0 +1,20 @@
...
+import { createHardhatRuntimeEnvironment } from "@nomicfoundation/hardhat-core";
+
+let hre: HardhatRuntimeEnvironment;
+
+/**
+ * This function returns a singleton instance of the Hardhat Runtime Environment.
+ *
+ * It exists so that the CLI and the programmatic API are always using the same HRE instance.
+ */
+export async function getHRE(
```
schaable, line 12: the all uppercase `HRE` bugs me a bit 😅

v-next/hardhat/src/internal/hre-singleton.ts
```diff
@@ -0,0 +1,20 @@
...
+  ...params: Parameters<typeof createHardhatRuntimeEnvironment>
+): Promise<HardhatRuntimeEnvironment> {
+  if (hre === undefined) {
```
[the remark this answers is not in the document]
alcuadrado: While that's a good point, this is intended behavior. We can add a comment, but this is a private API that should only be used in two different places.

v-next/hardhat/test/hre/index.ts
```diff
@@ -0,0 +1,33 @@
+import assert from "node:assert";
```
schaable, line 1: We should use `node:assert/strict` and remove "strict" from the assertions (`deepEqual` instead of `deepStrictEqual`).

alcuadrado: There's now a new eslint rule for this. So merging this will make the linter break in `v-next`. Do you mind changing this before merging, @zoeyTM ?

alcuadrado (approved):

PR #5273, opened 2024-05-24
v-next/lazy-plugin-loading

kanej:
Alter `HardhatPlugin` to use lazily loaded dependencies. This guards against failed installs as deps can have the form:

```ts
{
  id: "example",
  npmPackage: "my-plugin",
  dependencies: [
    () => import("another-plugin")
  ]
}
```

Several moves were made to support this:

1. The `HardhatPlugin` type was updated
2. Validation of plugin installs was moved into the plugin resolution logic. The dependency load is tried first, on exception a validation of the plugin installation is attempted. If there are no installation issue a generic dependency loading failed error is thrown.
3. The base path for npm resolution (the directory of the hardhat config file in v2) is now exposed through the `resolvePluginList` signature.

### DIFF
--- a/v-next/core/src/internal/hook-manager.ts
+++ b/v-next/core/src/internal/hook-manager.ts
@@ -9,11 +9,7 @@ import {
 import { HardhatPlugin } from "../types/plugins.js";
 import { LastParameter, Return } from "../types/utils.js";
 
-import { validatePluginNpmDependencies } from "./plugins/plugin-validation.js";
-
 export class HookManagerImplementation implements HookManager {
-  readonly #validatedPlugins = new Set<string>();
-
   readonly #pluginsInReverseOrder: HardhatPlugin[];
 
   /**
@@ -242,11 +238,6 @@ export class HookManagerImplementation implements HookManager {
 
         if (hookHandlerCategoryFactory === undefined) {
           return;
-        }
-
-        if (!this.#validatedPlugins.has(plugin.id)) {
-          await validatePluginNpmDependencies(plugin);
-          this.#validatedPlugins.add(plugin.id);
         }
 
         let hookCategory: Partial<HardhatHooks[HookCategoryNameT]>;
--- a/v-next/core/src/internal/hre.ts
+++ b/v-next/core/src/internal/hre.ts
@@ -39,7 +39,7 @@ export class HardhatRuntimeEnvironmentImplementation
 
     const resolvedPlugins =
       unsafeOptions?.resolvedPlugins ??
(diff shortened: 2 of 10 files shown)


v-next/core/src/internal/hook-manager.ts
```diff
@@ -9,11 +9,7 @@ import {
 import { HardhatPlugin } from "../types/plugins.js";
 import { LastParameter, Return } from "../types/utils.js";
 
-import { validatePluginNpmDependencies } from "./plugins/plugin-validation.js";
-
 export class HookManagerImplementation implements HookManager {
-  readonly #validatedPlugins = new Set<string>();
```
kanej, line 15: This does not appear to have a usage now that validation happens before the `HookManager` is constructed.

v-next/core/src/internal/hre.ts
```diff
@@ -37,9 +37,18 @@ export class HardhatRuntimeEnvironmentImplementation
     // TODO: Or maybe don't clone at all
     const clonedUserConfig = inputUserConfig;
 
+    // TODO: When all loads where based on a hardhat config file
+    // on disk we used the config file's directory as the base path
+    // We need to decide if the current working directory is the right
+    // option for module resolution now
+    const basePathForNpmResolution = process.cwd();
```
kanej, line 44: I pulled up the point at which the npm resolution path is set. Currently this is set to `process.cwd()`, but that should probably change to `process.cwd()` for programmatically loaded config, and the directory of the `Hardhat.config.{js,ts}` if loaded from the command line.

v-next/core/src/internal/plugins/resolve-plugin-list.ts
```diff
@@ -42,17 +47,54 @@ export function reverseTopologicalSort(
...
+  basePathForNpmResolution: string,
+): Promise<HardhatPlugin> {
+  try {
+    return await loadFn();
+  } catch (error) {
+    ensureError(error);
+
+    await validatePluginNpmDependencies(plugin, basePathForNpmResolution);
+
+    throw new HardhatError(
```
kanej, line 92: This is the fallback if validation of plugin install did not throw an error.

v-next/core/src/internal/hre.ts
alcuadrado: agreed

alcuadrado (approved):
I think this looks good.

In succesive PRs we should:
1. Improve the error messages in the validation
2. Validate when loading a hook category
3. Validate when loading a task action
4. Rename the validation process to something that implies that it's detecting an error condition, not validating

Issue #5229, opened 2024-05-16
Core tasks missing features


kanej:
The tasks part of `./v-next/core` needs functionality filled out around the building up of task definitions and running the tasks.

## TODO

- [x] Complete the implementation of and test the TaskDefinition builders
- [x] Implement and test the ResolvedTask#run method, which runs the tasks
- [x] Polish and test the TaskManager
- [x] #5302
- [x] #5303
- [x] #5304
- [x] #5333
- [x] #5335
- [x] Check for missing dependencies using `detectPluginNpmDependencyProblems` in the `ResolvedTask#run` method when failing to import a file action.

PR #5274, opened 2024-05-24
Tasks builders implementation and testing

schaable:

This should be easy to review if you go commit by commit instead of looking at the files.

### DIFF
--- a/v-next/core/src/internal/tasks/builders.ts
+++ b/v-next/core/src/internal/tasks/builders.ts
@@ -10,6 +10,8 @@ import type {
   TaskOverrideDefinition,
 } from "../../types/tasks.js";
 
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
+
 import { ParameterType, isParameterValueValid } from "../../types/common.js";
 import { TaskDefinitionType } from "../../types/tasks.js";
 import {
@@ -17,7 +19,7 @@ import {
   isValidParamNameCasing,
 } from "../parameters.js";
 
-import { formatValue, isValidActionUrl } from "./utils.js";
+import { formatTaskId, formatValue, isValidActionUrl } from "./utils.js";
 
 export class NewTaskDefinitionBuilderImplementation
   implements NewTaskDefinitionBuilder
@@ -44,7 +46,12 @@ export class NewTaskDefinitionBuilderImplementation
 
   public setAction(action: NewTaskActionFunction | string): this {
     if (typeof action === "string" && !isValidActionUrl(action)) {
-      throw new Error("Invalid action file URL");
+      throw new HardhatError(
+        HardhatError.ERRORS.TASK_DEFINITIONS.INVALID_FILE_ACTION,
+        {
+          action,
+        },
+      );
     }
 
     this.#action = action;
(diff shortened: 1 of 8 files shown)


v-next/core/src/types/common.ts
```diff
@@ -9,3 +9,24 @@ export enum ParameterType {
   FLOAT = "FLOAT",
   FILE = "FILE",
 }
+
+interface ParameterToValueTypeMap {
```
alcuadrado, line 13: I think this should be exported.

v-next/core/src/internal/tasks/utils.ts
```diff
@@ -11,3 +16,11 @@ const FILE_PROTOCOL_PATTERN = /^file:\/\/.+/;
 export function isValidActionUrl(action: string): boolean {
   return FILE_PROTOCOL_PATTERN.test(action);
 }
+
+export function formatValue<T extends ParameterType>(
```
alcuadrado, line 20: this should be handled by HardhtError, so that we are consistent.

v-next/core/src/internal/tasks/builders.ts
```diff
@@ -86,6 +86,10 @@ export class NewTaskDefinitionBuilderImplementation
       );
     }
 
+    if (type === ParameterType.BOOLEAN && defaultValue === undefined) {
+      defaultValue = false as ParameterTypeToValueType<T>;
+    }
```
alcuadrado, line 109: This is implicitly turning a required parameter into an optional one

v-next/core/src/internal/tasks/builders.ts
```diff
@@ -222,32 +300,74 @@ export class TaskOverrideDefinitionBuilderImplementation
...
+        HardhatError.ERRORS.ARGUMENTS.INVALID_VALUE_FOR_TYPE,
+        {
+          value: formatValue(defaultValue),
+          name: "defaultValue",
+          type: parameterType,
+        },
+      );
+    }
+
+    if (type === ParameterType.BOOLEAN && defaultValue === undefined) {
```
alcuadrado, line 363: same here

v-next/core/test/internal/tasks/builders.ts
```diff
@@ -0,0 +1,1157 @@
...
+describe("Task builders", () => {
+  describe("NewTaskDefinitionBuilderImplementation", () => {
+    it("should create a new task definition builder", () => {
+      const builder = new NewTaskDefinitionBuilderImplementation("task-id");
+      const taskAction = () => {};
+      const taskDefinition = builder.setAction(taskAction).build();
+
+      expectTypeOf(
+        builder,
+      ).toEqualTypeOf<NewTaskDefinitionBuilderImplementation>();
```
alcuadrado, line 26: I don't think I'd do `expectTypeOf` test unless there's some advanced typing like the one you did with the default values. The reason is that the type system is implicitly doing the same test. e.g. the deepEqual below woudln't type.

v-next/core/test/internal/tasks/builders.ts
```diff
@@ -0,0 +1,1157 @@
...
+              description: "",
+              parameterType: ParameterType.INT,
+              defaultValue: undefined,
+            },
+          },
+          positionalParameters: [],
+        });
+      });
+
+      it("should set the default defaultValue to false for when the type is BOOLEAN", () => {
```
alcuadrado, line 290: I think this is worth clarifying, as my TODOs could have been misleading.

Boolean name parameters do have a special treatment, but in the argument parser, not here. 

That special treatment is this:
* If a boolean named parameter has a default value (not all of them have), and if that default value is false, it is consider a "flag".
* Flags can be used in the CLI either like this `--flag true` or just `--flag`.

v-next/core/test/internal/tasks/builders.ts
```diff
@@ -0,0 +1,1157 @@
...
+              message: `HHE302: Argument name ${name} is already in use`,
+            });
+          });
+        });
+      });
+
+      it("should throw if the parameter name is reserved", () => {
+        const builder = new NewTaskDefinitionBuilderImplementation("task-id");
+
+        const reservedNames = ["config", "help", "version"];
```
alcuadrado, line 653: I think we should import the constant, as I'm 100% sure this will change soon

alcuadrado (changes requested):
I left a few comments, but everything else looks good

v-next/core/src/internal/tasks/utils.ts
```diff
@@ -1,7 +1,28 @@
+import type {
+  ParameterType,
+  ParameterTypeToValueType,
+} from "../../types/common.js";
```
alcuadrado, line 4: Testing github actions

PR #5277, opened 2024-05-27
Config loading

### LINKED ISSUE #5263: Implement and test the lookup and loading of the hardhat config file in the CLI and index.ts

zoeyTM:
resolves #5263

### DIFF
--- a/v-next/hardhat/src/internal/helpers/config-loading.ts
+++ b/v-next/hardhat/src/internal/helpers/config-loading.ts
@@ -1,21 +1,25 @@
 import { isAbsolute, resolve } from "node:path";
+import { pathToFileURL } from "node:url";
 
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
 import { findUp } from "@nomicfoundation/hardhat-utils/fs";
 
+import { ERRORS } from "../../../../hardhat-errors/src/descriptors.js";
+
 async function findClosestHardhatConfig(): Promise<string> {
-  let hardhatConfigPath = await findUp("hardhat.config.js");
+  let hardhatConfigPath = await findUp("hardhat.config.ts");
 
   if (hardhatConfigPath !== undefined) {
     return hardhatConfigPath;
   }
 
-  hardhatConfigPath = await findUp("hardhat.config.ts");
+  hardhatConfigPath = await findUp("hardhat.config.js");
 
   if (hardhatConfigPath !== undefined) {
     return hardhatConfigPath;
   }
 
-  throw new Error("No Hardhat config file found");
+  throw new HardhatError(ERRORS.GENERAL.NO_CONFIG_FILE_FOUND);
 }
 
 export async function resolveConfigPath(): Promise<string> {
@@ -36,19 +40,22 @@ export async function importUserConfig(configPath: string) {
   const { exists } = await import("@nomicfoundation/hardhat-utils/fs");
 
   if (!(await exists(normalizedPath))) {
-    throw new Error(`Config file ${configPath} not found`);
+    throw new HardhatError(ERRORS.GENERAL.INVALID_CONFIG_PATH, { configPath });
   }
 
-  const imported = await import(normalizedPath);
+  console.log("also here", pathToFileURL(normalizedPath).href);
+  const imported = await import(pathToFileURL(normalizedPath).href);
 
   if (!("default" in imported)) {
-    throw new Error(`No config exported in ${configPath}`);
+    throw new HardhatError(ERRORS.GENERAL.NO_CONFIG_EXPORTED, { configPath });
   }
 
   const config = imported.default;
 
   if (typeof config !== "object" || config === null) {
-    throw new Error(`Invalid config exported in ${configPath}`);
+    throw new HardhatError(ERRORS.GENERAL.INVALID_CONFIG_OBJECT, {
+      configPath,
+    });
   }
 
   return config;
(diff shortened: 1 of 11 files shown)


v-next/hardhat/test/helpers/project.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  return getRealPath(projectPath);
+}
+
+/**
+ * Returns the real path of absolutePath, resolving symlinks.
+ *
+ * @throws Error if absolutePath doesn't exist.
+ */
+async function getRealPath(absolutePath: string): Promise<string> {
```
schaable, line 59: you should use `getRealPath` from `hardhat-utils`

v-next/hardhat/test/helpers/project.ts
```diff
@@ -0,0 +1,67 @@
...
+import { exists } from "@nomicfoundation/hardhat-utils/fs";
+
+/**
+ * This helper adds mocha hooks to run the tests inside one of the projects
+ * from test/fixture-projects.
+ *
+ * @param projectName The base name of the folder with the project to use.
+ * @param changeDirTo If provided, the working directory will be changed to this. Must be a child of the project folder.
+ */
+export function useFixtureProject(projectName: string, changeDirTo?: string) {
```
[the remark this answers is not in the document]
alcuadrado: I think it should be a separate package, as it should be a devDependency. I'm planning to add it asap. Before we start repeating ourselves.

v-next/hardhat/src/internal/helpers/config-loading.ts
```diff
@@ -0,0 +1,55 @@
...
+import { findUp } from "@nomicfoundation/hardhat-utils/fs";
+
+async function findClosestHardhatConfig(): Promise<string> {
+  let hardhatConfigPath = await findUp("hardhat.config.js");
+
+  if (hardhatConfigPath !== undefined) {
+    return hardhatConfigPath;
+  }
+
+  hardhatConfigPath = await findUp("hardhat.config.ts");
```
alcuadrado, line 12: I think `.ts` should be before `.js`. So that if a user accidentally compiles his config once, we still use the most updated `.ts` version.

v-next/hardhat/src/internal/helpers/config-loading.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  if (hardhatConfigPath !== undefined) {
+    return hardhatConfigPath;
+  }
+
+  throw new Error("No Hardhat config file found");
```
alcuadrado, line 18: This should be a HardhatError

alcuadrado: same for the rest

v-next/hardhat/test/hre/index.ts
```diff
@@ -3,7 +3,9 @@ import { describe, it } from "node:test";
 
 import { createHardhatRuntimeEnvironment } from "../../src/hre.js";
 import { builtinPlugins } from "../../src/internal/builtin-plugins/index.js";
+import { resolveConfigPath } from "../../src/internal/helpers/config-loading.js";
```
alcuadrado, line 6: I think the `cacelledByParent` errors are a failure in the `before` or `after`.

(I'm commenting in the closest line that github allowed me to, they are above)

alcuadrado (approved):

v-next/hardhat/src/internal/helpers/config-loading.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  return findClosestHardhatConfig();
+}
+
+export async function importUserConfig(configPath: string) {
+  const normalizedPath = isAbsolute(configPath)
+    ? configPath
+    : resolve(process.cwd(), configPath);
+
+  const { exists } = await import("@nomicfoundation/hardhat-utils/fs");
```
schaable, line 40: We can probably use regular imports for the `hardhat-utils` package, as the package only imports things from node and should be fast enough to avoid using dynamic imports.

v-next/hardhat/src/internal/helpers/config-loading.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  const imported = await import(pathToFileURL(normalizedPath).href);
+
+  if (!("default" in imported)) {
+    throw new HardhatError(ERRORS.GENERAL.NO_CONFIG_EXPORTED, { configPath });
+  }
+
+  const config = imported.default;
+
+  if (typeof config !== "object" || config === null) {
```
schaable, line 54: @alcuadrado This would allow `Array` configs to be passed, is that intended? By the way, this is the second time I've seen an `isObject` check (which is also missing the `Array` check). I believe we should add this as a util.

schaable: @zoeyTM you can ignore this, I'll put a separate PR adding `isObject` and replacing the occurrences with the helper.

v-next/hardhat/src/internal/helpers/config-loading.ts
schaable: Ignore this comment for now, we'll need more data to decide if and when it is worth using dynamic imports, as most things now will be lazy and wont benefit from using this the pattern.

zoeyTM: This code in particular already existed in the `v-next` branch, I just moved it from `internal/cli/main.ts` into here. Definitely something to think about moving forward though.

PR #5282, opened 2024-05-28
Add parsing for tasks, subtasks and tasks' parameters

ChristopherDedominici:
This PR implements:
- the logic to parse tasks and subtasks
- the logic to parse named parameters, positional parameters and variadic parameters
- tests to check these functionalities

The PR is organized as such:
- in the 'src/internal/cli/main.ts' file there is all the code logic to handle the parsing
- I added a few ERROR messages, they are in the descriptors.ts file
- all the tests are in the 'hardhat/test/internal/cli/main.ts' file

The tests are quite a lot, but they are also covering very small specific scenarios so it should be quite easy to understand what they do based on the description in the "describe" and "it" functions

The parsing of the GLOBAL PARAMETERS will be in a separate PR


v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -0,0 +1,647 @@
...
+interface TaskRes {
+  task: Task;
+  taskArguments: Record<string, any>;
+}
+
+describe("main", function () {
+  let hre: HardhatRuntimeEnvironment;
+
+  // TASKS and SUBTASKS are defined in the "before()" hooks before every "functionality test groups".
+  let TASK: NewTaskDefinition;
```
schaable, line 24: I think I'd keep these as camelCase as they're not constants.

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -0,0 +1,647 @@
...
+          usedCliArguments,
+          hre,
+        ) as TaskRes;
+
+        assert.equal(res.task.id, TASK.id);
+        assert.deepStrictEqual(usedCliArguments, [true, true]);
+        assert.deepStrictEqual(res.taskArguments, { flag: true });
+      });
+
+      it("should throw because the parameter when set as bool (flag behavior) should not be followed by a bool value", function () {
```
schaable, line 140: but this is not throwing, right? we should update the description

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -0,0 +1,647 @@
...
+
+      it("should throw because the parameter is not defined", async function () {
+        const command = "npx hardhat task --undefinedParam <value>";
+
+        const cliArguments = command.split(" ").slice(2);
+        const usedCliArguments = [false, false, false];
+
+        assert.throws(
+          () => parseTaskAndArguments(cliArguments, usedCliArguments, hre),
+          new HardhatError(
```
schaable, line 183: I think an easier syntax may be using an object with the error and the message: 
```
{
  name: "HardhatError",
  message: 'Invalid parameter "--undefinedParam". It is neither a valid global parameter nor associated with any task. Did you forget to add the task first, or did you misspell it?',
}
```

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -0,0 +1,647 @@
...
+        const usedCliArguments = [false, true, true, false];
+
+        const res = parseTaskAndArguments(
+          cliArguments,
+          usedCliArguments,
+          hre,
+        ) as TaskRes;
+
+        assert.equal(res.task.id, SUBTASK.id);
+        assert.deepStrictEqual(usedCliArguments, [true, true, true, true]);
```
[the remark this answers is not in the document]
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5282/commits/17c0d94f88790c6ab94f7907988671c5fc4fa32f)

v-next/hardhat/test/internal/cli/main.ts
ChristopherDedominici: [Done in  this commit](https://github.com/NomicFoundation/hardhat/pull/5282/commits/75bcb877e7701e0a746b8d147c18693414309f68)

v-next/hardhat/test/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5282/commits/d2e2f9877dfbba79626a4417708dbd2591f9217a)

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +272,214 @@ function getTaskFromCliArguments(
...
+    }
+
+    // The value immediately following a named parameter (if the parameter does not behave as a flag)
+    // is the parameter's value; otherwise, it's an error
+    if (
+      cliArguments[i + 1] !== undefined &&
+      usedCliArguments[i + 1] === false
+    ) {
+      taskArguments[paramOriginalName] = formatParameterValue(
+        cliArguments[++i],
```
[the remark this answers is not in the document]
ChristopherDedominici: On this comment I'm a bit confused. Which flag is missing?

But, I did simplified the if condition and I moved the "i" variable increment before performing the operations to increase readability.
[See here](https://github.com/NomicFoundation/hardhat/pull/5282/commits/5d008d0d0ac7ea655a2ac538ca9f49d093450e03)

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +272,214 @@ function getTaskFromCliArguments(
 
   return task;
 }
+
+function parseTaskArguments(
+  cliArguments: string[],
+  usedCliArguments: boolean[],
+  task: Task,
+): Record<string, any> {
+  const taskArguments: Record<string, any> = {};
```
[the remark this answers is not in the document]
ChristopherDedominici: Why would you go with unknow over any?

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +272,214 @@ function getTaskFromCliArguments(
...
+    }
+
+    if (cliArguments[i] === "--") {
+      // A standalone '--' is ok because it is used to separate CLI tool arguments from script arguments,
+      // ensuring the tool passes subsequent options directly to the script.
+      usedCliArguments[i] = true;
+      continue;
+    }
+
+    if (task.positionalParameters[paramI] === undefined) {
```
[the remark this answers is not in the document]
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5282/commits/78afa898226ac81d34e68a3038b943643bcdf807)

v-next/hardhat/test/internal/cli/main.ts
ChristopherDedominici: This will show the message in the test, true, but then all the test should be modified if the error message is changed. 
Wdyt?

v-next/hardhat/src/internal/cli/main.ts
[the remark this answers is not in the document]
schaable: Using `any` is usually considered a bad practice. It's just a way to bypass type checking and it tends to spread when you use it in type expressions, making the codebase less type safe. `unknown` is better suited for the type above as it contains every type.

schaable (approved):

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+  parseNamedParameters(cliArguments, usedCliArguments, task, taskArguments);
+
+  parsePositionalAndVariadicParameters(
+    cliArguments,
+    usedCliArguments,
+    task,
+    taskArguments,
+  );
+
+  usedCliArguments.forEach((arg, index) => {
```
alcuadrado, line 289: `.forEach` just adds overhead without any obvious benefit. A pattern like this would be better:

```js
      const asd = [false, false, false];
      const unsedIndex = asd.indexOf(false);
      if (unsedIndex !== -1) {
        throw new Error('asd');
      }
```

Please don't read this as "this is too slow". My point is about not using something that has a cost for now benefit.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+) {
+  for (let i = 0; i < cliArguments.length; i++) {
+    if (usedCliArguments[i]) {
+      continue;
+    }
+
+    if (cliArguments[i] === "--") {
+      // A standalone '--' is ok because it is used to separate CLI tool arguments from script arguments,
+      // ensuring the tool passes subsequent options directly to the script.
+      continue;
```
alcuadrado, line 314: Shouldn't this be a `break`? 

I'll try to validate it with a test.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+        },
+      );
+    }
+
+    usedCliArguments[i] = true;
+
+    if (paramInfo.parameterType === ParameterType.BOOLEAN) {
+      if (
+        usedCliArguments[i + 1] === false &&
+        ["true", "false"].includes(cliArguments[i + 1])
```
alcuadrado, line 340: This pattern introduces unnecessary overhead, as you need to allocate that array and call a function rather than running two comparisons.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+          parameter: arg,
+        },
+      );
+    }
+
+    usedCliArguments[i] = true;
+
+    if (paramInfo.parameterType === ParameterType.BOOLEAN) {
+      if (
+        usedCliArguments[i + 1] === false &&
```
alcuadrado, line 339: Can't `i + 1` be out of bounds?

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+        ["true", "false"].includes(cliArguments[i + 1])
+      ) {
+        // The flag could be follow by the boolean value
+        taskArguments[paramName] = formatParameterValue(
```
alcuadrado, line 343: This isn't a format operation, but a parsing operation instead. We have a textual representation of the parameter value, and we want to parse it.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+          ParameterType.BOOLEAN,
+          paramName,
+        );
+
+        usedCliArguments[i + 1] = true;
+        continue;
+      }
+
+      // If the flag is not followed by a boolean value, then the flag itself means that the value is true
+      taskArguments[paramName] = true;
```
alcuadrado, line 354: This should only happen when its default value is `false`. If it's either `true` or `undefined`, it shouldn't.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+): any {
+  switch (type) {
+    case ParameterType.STRING:
+    case ParameterType.FILE:
+      return value;
+    case ParameterType.INT:
+    case ParameterType.FLOAT:
+      return Number(value);
+    case ParameterType.BIGINT:
+      return BigInt(value);
```
alcuadrado, line 445: These should also handle errors.

Also, in v2 we were accepting these strings:

* int     
```js
const decimalPattern = /^\d+(?:[eE]\d+)?$/;
const hexPattern = /^0[xX][\dABCDEabcde]+$/;
```

* bigint
```
const decimalPattern = /^\d+(?:n)?$/;
const hexPattern = /^0[xX][\dABCDEabcde]+$/;
```

* float
```
const decimalPattern = /^(?:\d+(?:\.\d*)?|\.\d+)(?:[eE]\d+)?$/;
const hexPattern = /^0[xX][\dABCDEabcde]+$/;
```

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+      }
+
+      // If the flag is not followed by a boolean value, then the flag itself means that the value is true
+      taskArguments[paramName] = true;
+      continue;
+    }
+
+    // The value immediately following a named parameter (if the parameter does not behave as a flag)
+    // is the parameter's value; otherwise, it's an error
+    if (usedCliArguments[i + 1] === false) {
```
alcuadrado, line 360: Also here, I think this can be out of bounds

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+      // A standalone '--' is ok because it is used to separate CLI tool arguments from script arguments,
+      // ensuring the tool passes subsequent options directly to the script.
+      usedCliArguments[i] = true;
+      continue;
+    }
+
+    const paramInfo = task.positionalParameters[paramI];
+
+    if (paramInfo === undefined) {
+      continue;
```
alcuadrado, line 406: I think this should be a break

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+    usedCliArguments[i] = true;
+
+    const formattedValue = formatParameterValue(
+      cliArguments[i],
+      paramInfo.parameterType,
+      paramInfo.name,
+    );
+
+    if (paramInfo.isVariadic === false) {
+      taskArguments[task.positionalParameters[paramI++].name] = formattedValue;
```
alcuadrado, line 418: This line is a bit confusing. In particular the `++`, which is not really related to assigning the value.


I think that increment would be better to understand on a separate statement. This got me thinking that it could be wrong, as it's not present on the other side of the branch (which is correct though).

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: I just checked, and it can indeed be out of bounds. If the intention is for this same if to handle the array boundary, let's make that explicit. Otherwise, it looks like a mistake. 

Note that this comment is also related: https://github.com/NomicFoundation/hardhat/pull/5282/files#r1620824303

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: This actually works correctly, but it looks like unintended behavior.

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: Yes, it should be a break. For example, this parsers correctly, when it should be an error:

```ts
const hre = await createHardhatRuntimeEnvironment({
      tasks: [
        task(["task"])
          .addNamedParameter({ name: "param" })
          .addNamedParameter({
            name: "flag",
            type: ParameterType.BOOLEAN,
          })
          .addNamedParameter({
            name: "optionFlag",
          })
          .setAction(() => {})
          .build(),
      ],
    });

    const command = "npx hardhat task -- --param asd";
    const cliArguments = command.split(" ").slice(2);
    const unusedCliArguments = new Array(cliArguments.length).fill(false);
```

It prints:
```
{
  task: ResolvedTask {
    id: [ 'task' ],
    description: '',
    actions: [ [Object] ],
    namedParameters: Map(3) {
      'param' => [Object],
      'flag' => [Object],
      'optionFlag' => [Object]
    },
    positionalParameters: [],
    pluginId: undefined,
    subtasks: Map(0) {}
  },
  taskArguments: { param: 'asd' }
}
```

But everything after `--` should be considered positional params, and fail, as this task doesn't have any.

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -0,0 +1,647 @@
...
+            HardhatError.ERRORS.ARGUMENTS.UNRECOGNIZED_NAMED_PARAM,
+            {
+              parameter: "--param",
+            },
+          ),
+        );
+      });
+
+      it("should throw because the task parameter is declared but no value is associated to it", async function () {
+        const command = "npx hardhat task --param";
```
alcuadrado, line 210: I think there are more interesting cases to test that follow under this situation. For example

`npx hardhat task --param --global-param global-param-value`

alcuadrado: the difference is that there's no value because the next cli arg has been used, and not because there isn't any

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -0,0 +1,647 @@
...
+        ) as TaskRes;
+
+        assert.equal(res.task.id, newTaskDefinition.id);
+        assert.deepEqual(usedCliArguments, [true, true, true]);
+        assert.deepEqual(res.taskArguments, {
+          param: "subtask",
+        });
+      });
+
+      it("should get the tasks and its required parameter (the positional argument has the same value as a subtask name)", function () {
```
alcuadrado, line 291: this is a cool test case

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -0,0 +1,647 @@
...
+        ) as TaskRes;
+
+        assert.equal(res.task.id, newTaskDefinition.id);
+        assert.deepEqual(usedCliArguments, [true, true, true, true]);
+        assert.deepEqual(res.taskArguments, {
+          param: ["<val1>", "<val2>", "<val3>"],
+        });
+      });
+
+      it("should not throw when the parameters are not passed", function () {
```
alcuadrado, line 383: If there's no default value it should throw

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -0,0 +1,647 @@
...
+          })
+          .setAction(() => {})
+          .build();
+
+        hre = await createHardhatRuntimeEnvironment({
+          tasks: [newTaskDefinition, newSubtaskDefinition],
+        });
+      });
+
+      it("should get the tasks and its required parameter", function () {
```
alcuadrado, line 252: what if it's missing?

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+
+function kebabToCamelCase(str: string) {
+  return str.replace(/-./g, (match) => match.charAt(1).toUpperCase());
+}
+
+function formatParameterValue(
+  value: string,
+  type: ParameterType,
+  name: string,
+): any {
```
alcuadrado, line 436: Instead of any, we can properly type this. It can be done in a follow-up PR. @schaable already worked on something similar for the task builders.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -233,11 +228,20 @@ function getTaskFromCliArguments(
...
+      // ensuring the tool passes subsequent options directly to the script.
       if (arg.length === 2 || task !== undefined) {
         break;
       }
 
-      throw new Error(`Found task argument ${arg} before the task name`);
+      // At this point in the code, the global parameters have already been parsed, so the remaining parameters starting with '--' are task named parameters.
+      // Hence, if no task is defined, it means that the parameter is not assigned to any task, and it's an error.
+      throw new HardhatError(
+        HardhatError.ERRORS.ARGUMENTS.UNRECOGNIZED_NAMED_PARAM,
```
alcuadrado, line 240: I think this error is too general. In this case we are almost sure that the user accidentally added a non-global param before the task name, so it's better to throw a more specialized error.

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -232,6 +232,32 @@ Please double check your arguments.`,
 
 Please double check your arguments.`,
     },
+    UNRECOGNIZED_NAMED_PARAM: {
+      number: 304,
+      messageTemplate:
+        "Invalid parameter {parameter}. It is neither a valid global parameter nor associated with any task. Did you forget to add the task first, or did you misspell it?",
```
alcuadrado, line 238: This error is slightly confusing. It's not associated with the current task, not with any task.

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: Sorry, this message wasn't self contained. See this one for more context: https://github.com/NomicFoundation/hardhat/pull/5282/files#r1620815115 — The last paragraph in particular.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -265,3 +269,192 @@ function getTaskFromCliArguments(
...
+    if (paramInfo.isVariadic === false) {
+      taskArguments[task.positionalParameters[paramI++].name] = formattedValue;
+      continue;
+    }
+
+    // Handle variadic parameters
+    taskArguments[paramInfo.name] = taskArguments[paramInfo.name] ?? [];
+    taskArguments[paramInfo.name].push(formattedValue);
+  }
+}
```
alcuadrado, line 426: I guess [this](https://github.com/NomicFoundation/hardhat/pull/5282/files#r1621117929) should be handled here.

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: Before I had a out of bound check, but then I realized that, as you mention, this work correctly. So I decided to remove code that was "useless". Do you think that being more explicit would be better?

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -233,11 +228,20 @@ function getTaskFromCliArguments(
     const arg = cliArguments[i];
 
     if (arg.startsWith("--")) {
+      // A standalone '--' is ok because it is used to separate CLI tool arguments from script arguments,
+      // ensuring the tool passes subsequent options directly to the script.
```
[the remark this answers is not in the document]
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/c52e575bc7721414822ade410e7888bab666e1bc)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/eb56b333ad3c3cbcc0f3904d73ef797f89689971)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/49e9e2c630dd244de22da5ad5ec843827ee4769e)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/a72d4dbbfeec10d93c457e9620686c5903d9ab0c)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/e3e958a3ed540e89742f3fa6af3a3e6e62f3e351)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/e3e958a3ed540e89742f3fa6af3a3e6e62f3e351)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/74bbbcb8afd72b8efacaf0a80fc1ec7e122a6d30)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/74bbbcb8afd72b8efacaf0a80fc1ec7e122a6d30)

I added a comment explaining why on the other side of the branch "paramI" is not incremented. Let me know if it's still not clear

v-next/hardhat/test/internal/cli/main.ts
ChristopherDedominici: But should something like this work?

`npx hardhat task --param --global-param global-param-value <value>`

Where `<value>` will be associated to `--param`?
My understanding was that the next argument after a named parameter should be its value

v-next/hardhat/test/internal/cli/main.ts
ChristopherDedominici: This comment made me realize that I forgot to add a function to check the required parameters, I implemented it and added the tests in this commit: https://github.com/NomicFoundation/hardhat/pull/5300/commits/2b9eb55b0404b9a700c86981063a5223c6de88d4

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: Same fix as this comment: https://github.com/NomicFoundation/hardhat/pull/5282#discussion_r1622383160Same fix as this comment: https://github.com/NomicFoundation/hardhat/pull/5282#discussion_r1622383160

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: I thought that a bool flag (--flag) had the meaning of setting the value to true automatically. What is the expected behavior if the default value is true or undefined?

v-next/hardhat-errors/src/descriptors.ts
ChristopherDedominici: Related to this comment: https://github.com/NomicFoundation/hardhat/pull/5282#discussion_r1622408255

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/627e32e197b7342b38d1205b2ef3f3cc12bf5666)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: Should we consider to create a function for these operations in hh-utils? I could add the functions there and then use them here. WDYT?

alcuadrado: I wouldn't, as these have user-facing behavior that can accidentally be changed when/if we change the utils package. I think, in this case, locality is more important than reusability.

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: In those cases you have to provide the value explictly

v-next/hardhat/test/internal/cli/main.ts
alcuadrado: > My understanding was that the next argument after a named parameter should be its value

That's correct, it should fail. I just think it's worth testing it, as it's pretty different.

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: Great point!

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: Done it [this commit](https://github.com/NomicFoundation/hardhat/pull/5300/commits/bab2ad3654f68b050182e8e424254774d9800078) but follow up question: for the FILE type should we return the path passed by the user or the absolute path? Currently I'm returning the absolute path.

Also, the type validation modified a function from sync to async, so in this commit you'll see quite a lot of code changed because a lot of functions had to be converted into async to wait for promises

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: So should we keep the current error message?

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: After our offline sync we decided to handle FILE types as normal strings. So I will have to change the validation for the FILE type and undo the changes related to the async functions because the async function was the one used to read the folder but now it's no longer needed

PR #5283, opened 2024-05-28
Type-safe `HardhatError`

alcuadrado:
This PR modifies `HardhatError` so that its constructor and its `messageArguments` property are type-safe.

This introduces two extra changes:

1. The error message template system now uses `{variable}` instead of `%variable%`, as that dramatically simplified this PR.
2. The constructor only takes two parameters now if the message template doesn't use any variable. Those are the descriptor and the optional cause.

### DIFF
--- a/v-next/hardhat-errors/src/errors.ts
+++ b/v-next/hardhat-errors/src/errors.ts
@@ -9,6 +9,7 @@ export type ErrorMessageTemplateValue =
   | bigint
   | undefined
   | null
+  | ErrorMessageTemplateValue[]
   | { toString(): string };
 
 export type MessagetTemplateArguments<MessageTemplateT extends string> =
@@ -193,8 +194,17 @@ export function applyErrorMessageTemplate(
     if (rawValue === undefined) {
       return "undefined";
     }
+
     if (rawValue === null) {
       return "null";
+    }
+
+    if (typeof rawValue === "bigint") {
+      return `${rawValue}n`;
+    }
+
+    if (Array.isArray(rawValue)) {
+      return JSON.stringify(rawValue);
     }
 
     return rawValue.toString();
--- a/v-next/hardhat-errors/src/errors.ts
+++ b/v-next/hardhat-errors/src/errors.ts
@@ -2,10 +2,6 @@ import { CustomError } from "@nomicfoundation/hardhat-utils/error";
 
 import { ERRORS, ErrorDescriptor } from "./descriptors.js";
 
-export const ERROR_PREFIX = "HHE";
-
-const IS_HARDHAT_ERROR_PROPERTY_NAME = "_isHardhatError";
-
 export type ErrorMessageTemplateValue =
   | string
   | number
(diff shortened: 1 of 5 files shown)


v-next/hardhat-errors/src/errors.ts
```diff
@@ -130,68 +186,27 @@ function getErrorCode(errorDescriptor: ErrorDescriptor): string {
...
-      !/%([a-zA-Z][a-zA-Z0-9]*)?%/.test(value),
-      `Trying to apply an error template but variable "${variableName}" has a value that contains a variable tag.`,
-    );
+    if (typeof rawValue === "bigint") {
+      return `${rawValue}n`;
+    }
 
-    template = template.replaceAll(variableTag, value);
-  }
+    if (Array.isArray(rawValue)) {
```
alcuadrado, line 206: @schaable when merging this conflict I realized that the `hasOwnProperty` method didn't allow for classes to define their `toString`, so I switched it to explicitly handling arrays.

alcuadrado:
It should be easiear to review this one commit at the time. The 4th and 6th commits are just automatic string replacements.

v-next/hardhat-errors/test/errors.ts
```diff
@@ -325,57 +275,155 @@ describe("applyErrorMessageTemplate", () => {
...
+          ...mockErrorDescriptor,
+          messageTemplate: "{asd}",
+        } as const,
+        { asd: 123 },
+        new Error(),
+      );
+    });
+  });
+
+  describe("messageArguments propery types", () => {
```
kanej, line 412:
```suggestion
  describe("messageArguments propery types", () => {
```
```suggestion
  describe("messageArguments property types", () => {
```

v-next/hardhat-errors/src/errors.ts
```diff
@@ -2,45 +2,92 @@ import { CustomError } from "@nomicfoundation/hardhat-utils/error";
...
-const IS_HARDHAT_ERROR_PROPERTY_NAME = "_isHardhatError";
-
 export type ErrorMessageTemplateValue =
   | string
   | number
   | boolean
   | bigint
   | undefined
   | null
+  | ErrorMessageTemplateValue[]
```
kanej, line 12: I am not following this move. Is this simply that in an error message you can include an array so:

```ts
 const error = new HardhatError(
        {
          number: 12,
          messageTemplate: "{a} {b} {c}",
          websiteTitle: "Title",
          websiteDescription: "Description",
        } as const,
        { a: [1,2,3] },
        new Error(),
      );
```

Was this not supported before and it is now, or did something else prompt the move?

v-next/hardhat-errors/src/errors.ts
```diff
@@ -2,45 +2,92 @@ import { CustomError } from "@nomicfoundation/hardhat-utils/error";
...
 
-    super(prefix + formattedMessage, parentError);
-
     this.#descriptor = errorDescriptor;
 
+    if (
+      messageArgumentsOrParentError === undefined ||
+      messageArgumentsOrParentError instanceof Error
+    ) {
+      this.#arguments = {} as MessagetTemplateArguments<
```
kanej, line 84: Was this the problematic line you where talking about? 😄

kanej (approved):
I left a few questions for my own understanding, but pre-approving.

v-next/hardhat-errors/src/errors.ts
alcuadrado: it was, because arrays have a `toString`, but I made it more explicit.

v-next/hardhat-errors/src/errors.ts
alcuadrado: This one was also problematic, but the one I was confused about was that if you had no variables in your template this would work `new HardhatError(descriptor, {extraField: 123}, cause)`.

The solution was to change the constructor so that now, if you don't have any variable, it is called like this `new HardhatError(descriptor, cause)`

v-next/hardhat-errors/src/errors.ts
```diff
@@ -130,68 +186,27 @@ function getErrorCode(errorDescriptor: ErrorDescriptor): string {
...
-    return template
-      .split("%%")
-      .map((part) => _applyErrorMessageTemplate(part, values, true))
-      .join("%");
-  }
-
-  for (const variableName of Object.keys(values)) {
-    let value: string;
-
+  return template.replaceAll(/{(.*?)}/g, (_match, variableName) => {
```
schaable, line 191: should this match the empty group `{}`?

schaable (approved):
LGTM, just a question regarding the regex

v-next/hardhat-errors/src/errors.ts
alcuadrado: Good catch! But after giving it a few minutes, I think it's better if it matches the behavior of the types. To avoid matching the empty group, the types will be significantly more complex. 

I'll add a test for this case.

alcuadrado: Added the tests in [409dc51](https://github.com/NomicFoundation/hardhat/pull/5283/commits/409dc510353906bf50140abbec1319723fd7be8c)

PR #5284, opened 2024-05-28
Introduce a custom `node:test` reporter

alcuadrado:
This PR introduces a new package with a custom `node:test` reporter, but doesn't modify the rest of the packages to use it.

The new test reporter has a few objectives:

1. It tries to look like mocha.
2. If an error comes from a library using `actual` and `expected` (like chai/earl/etc), it diffs those values, printing the diff.
3. It prints its output as soon as possible but preserves the tests' definition order.

This PR is based on @ChristopherDedominici's work and https://github.com/voxpelli/node-test-pretty-reporter


v-next/hardhat-node-test-reporter/.eslintrc.cjs
```diff
@@ -0,0 +1,3 @@
+const { createConfig } = require("../../config-v-next/eslint.cjs");
```
alcuadrado, line 1: Github actions test

alcuadrado (commented):
Github actions test

schaable (approved):
LGTM

v-next/hardhat-node-test-reporter/integration-tests/index.ts
```diff
@@ -0,0 +1,61 @@
...
+  getAllFilesMatching,
+  readUtf8File,
+} from "@nomicfoundation/hardhat-utils/fs";
+import { run } from "node:test";
+
+import reporter from "../src/reporter.js";
+import { diff } from "jest-diff";
+const SHOW_OUTPUT = process.argv.includes("--show-output");
+
+for (let entry of await readdir(import.meta.dirname + "/fixture-tests")) {
```
ChristopherDedominici, line 13: I think the variable "entry" can be a const instead of let because it's not modified

v-next/hardhat-node-test-reporter/src/diagnostics.ts
```diff
@@ -0,0 +1,54 @@
...
+    const [name, numberString] = diagnostic.message.split(" ");
+    if (!(name in globalDiagnostics) || numberString === undefined) {
+      unusedDiagnostics.push(diagnostic);
+      continue;
+    }
+
+    try {
+      const value = parseFloat(numberString);
+      globalDiagnostics[name as keyof GlobalDiagnostics] = value;
+    } catch {
```
ChristopherDedominici, line 48: Can this "catch" block ever be triggered?

ChristopherDedominici (approved):
I just left 2 small comments but a part from them all good!

v-next/hardhat-node-test-reporter/integration-tests/index.ts
alcuadrado: oh, good catch! the linter wasn't enabled for these tests.

PR #5285, opened 2024-05-29
TaskManagerImplementation improvements

schaable:
This PR may be hard to read as there are a bunch of changes to the same file and the diff doesn't look good. Here's a summary of the most important changes:

- Replaced all Errors with HardhatErrors. Added error descriptors for each error.
- Added validation to the task builders to account for empty IDs (either empty string or empty array), along with tests.
- Added comments to make the code more legible.
- Since most errors can be thrown by a task added by the user or by a plugin, I've added a `getActorFragment` helper to avoid some of the branching and to prevent having different descriptors for the same error type.
- Used `assertHardhatInvariant` to avoid the non-null assertion in `getTask`.
- Updated `#insertTask` by moving the successful case to the end and handling errors first if there's an `existingTask`.
- Removed the check for an `undefined` task in `#processTaskOverride` as the call to `getTask` will throw an error if the task can't be found.
- Used `Object.entries` in `#processTaskOverride` to avoid looping over the named params twice.

### DIFF
--- a/v-next/core/src/internal/tasks/task-manager.ts
+++ b/v-next/core/src/internal/tasks/task-manager.ts
@@ -1,3 +1,8 @@
+import {
+  HardhatError,
+  assertHardhatInvariant,
+} from "@nomicfoundation/hardhat-errors";
+
 import { GlobalParameterMap } from "../../types/global-parameters.js";
 import { HardhatRuntimeEnvironment } from "../../types/hre.js";
 import {
@@ -10,7 +15,7 @@ import {
 } from "../../types/tasks.js";
 
 import { ResolvedTask } from "./resolved-task.js";
-import { formatTaskId } from "./utils.js";
+import { formatTaskId, getActorFragment } from "./utils.js";
 
 export class TaskManagerImplementation implements TaskManager {
   readonly #hre: HardhatRuntimeEnvironment;
@@ -22,6 +27,7 @@ export class TaskManagerImplementation implements TaskManager {
   ) {
     this.#hre = hre;
 
+    // reduce plugin tasks
     for (const plugin of this.#hre.config.plugins) {
       if (plugin.tasks === undefined) {
         continue;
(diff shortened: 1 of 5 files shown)


v-next/core/src/internal/tasks/task-manager.ts
```diff
@@ -36,82 +42,90 @@ export class TaskManagerImplementation implements TaskManager {
...
       }
 
       task = currentTask;
       tasks = task.subtasks;
     }
 
-    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion -- Can't be undefined as we set it in the non-empty loop.
-    return task!;
+    // task can't be undefined as we set it in the non-empty loop
+    assertHardhatInvariant(task !== undefined, "Task not found");
```
alcuadrado, line 81: This comment should be part of the assertion, so that it's self-descriptive, and more helpful it it ever fails:

```suggestion
    assertHardhatInvariant(task !== undefined, "Task is undefined despite it being always set by a non-empty loop");
```

v-next/core/src/internal/tasks/task-manager.ts
```diff
@@ -190,34 +204,21 @@ export class TaskManagerImplementation implements TaskManager {
...
-        throw new Error(
-          `Plugin ${pluginId} is trying to override the task "${formatTaskId(taskDefinition.id)}" but it hasn't been defined`,
-        );
-      } else {
-        throw new Error(
-          `Trying to override the task "${formatTaskId(taskDefinition.id)}" but it hasn't been defined`,
+    for (const [namedParamName, namedParamValue] of Object.entries(
+      taskDefinition.namedParameters,
+    )) {
+      if (task.namedParameters.has(namedParamName)) {
```
alcuadrado, line 210: If a positional parameter with the same name exists, it should also fail.

alcuadrado (commented):
Looks good, I only left a few comments

alcuadrado (approved):

PR #5286, opened 2024-05-29
Add tests for user interruptions

### LINKED ISSUE #5233: Test the user interruptions manager
Test the user interruptions manager, including its default and customized behavior.

### LINKED ISSUE #5235: Test the `AsyncMutex`
Test the AsyncMutex under `./v-next/core`. The mutex is used by the user interruptions manager.

> [!NOTE]  
> time box this, it may be too challenging or even impossible.

zoeyTM:
resolves #5233 
resolves #5235 

added tests for `UserInterruptionsManagerImplementation` and `AsyncMutex`

### DIFF
--- a/v-next/core/package.json
+++ b/v-next/core/package.json
@@ -33,6 +33,7 @@
     "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
     "test": "glob --cmd=\"node --import tsx/esm --test\" \"test/**/*.ts\"",
+    "test:only": "glob --cmd=\"node --import tsx/esm --test --test-only\" \"test/**/*.ts\"",
     "test:github": "glob --cmd=\"node --import tsx/esm --test --test-reporter=@reporters/github --test-reporter-destination=stdout --test-reporter=spec --test-reporter-destination=stdout\" \"test/**/*.ts\"",
     "test:coverage": "c8 --reporter html --reporter text --all --src src glob --cmd=\"node --import tsx/esm --test\" \"test/**/*.ts\"",
     "pretest": "pnpm build",
--- a/v-next/core/package.json
+++ b/v-next/core/package.json
@@ -33,7 +33,6 @@
     "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
     "test": "glob --cmd=\"node --import tsx/esm --test\" \"test/**/*.ts\"",
-    "test:only": "glob --cmd=\"node --import tsx/esm --test --test-only\" \"test/**/*.ts\"",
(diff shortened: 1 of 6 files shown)


v-next/core/test/internal/user-interruptions/async-mutex.ts
```diff
@@ -0,0 +1,28 @@
...
+      running--;
+    }
+
+    await Promise.all([
+      mutex.excluiveRun(run),
+      mutex.excluiveRun(run),
+      mutex.excluiveRun(run),
+    ]);
+
+    assert.strictEqual(maxRunning, 1);
```
kanej, line 26: Lets assert running is 0 as well.

v-next/core/test/internal/user-interruptions/async-mutex.ts
```diff
@@ -0,0 +1,28 @@
...
+import { AsyncMutex } from "../../../src/internal/async-mutex.js";
+
+describe("AsyncMutex", () => {
+  it("should run a function exclusively", async () => {
+    const mutex = new AsyncMutex();
+
+    let running = 0;
+    let maxRunning = 0;
+
+    async function run() {
```
schaable, line 13: nice!

v-next/core/test/internal/user-interruptions/user-interruptions-manager.ts
```diff
@@ -0,0 +1,111 @@
...
+      };
+
+      hookManager.registerHandlers("userInterruptions", handlers);
+
+      await userInterruptionManager.displayMessage(
+        "test-interruptor",
+        "test-message",
+      );
+
+      assert(called);
```
schaable, line 36: It's better if we add a `message` to `assert`, as it's an alias of `assert.ok` and there's a bug that makes `assert.ok` really slow unless you provide a `message`. See: https://github.com/nodejs/node/issues/52962

schaable (approved):
Just a minor comment, but LGTM

v-next/core/test/internal/user-interruptions/async-mutex.ts
kanej: Agreed.

v-next/core/package.json
```diff
@@ -33,6 +33,7 @@
     "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
     "test": "glob --cmd=\"node --import tsx/esm --test\" \"test/**/*.ts\"",
+    "test:only": "glob --cmd=\"node --import tsx/esm --test --test-only\" \"test/**/*.ts\"",
```
kanej, line 36: I would drop this line and let the later PR add it.

kanej (approved):
Looks good. Well done on the mutex test.

There are two points I am interested in:

- is there value in testing around `uninterrupted`?
- what do we think about tests of the default implementations?

I think we merge this PR and depending on the discussion, follow up on these two points.

zoeyTM:
@kanej 

- I don't really think there's value in testing `uninterrupted` since it would be identical to the mutex test
- During the sync with Pato, we agreed that testing the default implementations would not really be worth the effort and that testing the dynamic handlers would be enough

PR #5296, opened 2024-05-30
[v-next] eslint config update rules

alcuadrado:
This PR introduces two changes to our `eslint` config:

1) It enables the rule `@eslint-community/eslint-comments/require-description`. It forces you to explain why you used an `eslint-disable` directive in the same comment that uses it.
2) It changes the rule `@typescript-eslint/consistent-type-assertions` so type casts are now a linter error.

As this PR fixes a ton of new linter errors, it is better to review it one commit at the time. The first and third commits are the actual config changes, and the second and fourth commits just fix the new linter errors.

### DIFF
--- a/v-next/hardhat-errors/src/errors.ts
+++ b/v-next/hardhat-errors/src/errors.ts
@@ -82,7 +82,7 @@ export class HardhatError<
       messageArgumentsOrParentError instanceof Error
     ) {
       /* eslint-disable @typescript-eslint/consistent-type-assertions -- 
-      Typescript infercente get's lost here, but we know that if we didn't get
+      Typescript inference get's lost here, but we know that if we didn't get
       arguments, it's because the error doesn't have any. */
       this.#arguments = {} as MessagetTemplateArguments<
         ErrorDescriptorT["messageTemplate"]
--- a/v-next/hardhat-errors/src/errors.ts
+++ b/v-next/hardhat-errors/src/errors.ts
@@ -81,6 +81,9 @@ export class HardhatError<
       messageArgumentsOrParentError === undefined ||
       messageArgumentsOrParentError instanceof Error
     ) {
+      /* eslint-disable @typescript-eslint/consistent-type-assertions -- 
+      Typescript infercente get's lost here, but we know that if we didn't get
+      arguments, it's because the error doesn't have any. */
       this.#arguments = {} as MessagetTemplateArguments<
         ErrorDescriptorT["messageTemplate"]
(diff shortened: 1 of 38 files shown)


v-next/hardhat-errors/src/errors.ts
```diff
@@ -81,6 +81,9 @@ export class HardhatError<
       messageArgumentsOrParentError === undefined ||
       messageArgumentsOrParentError instanceof Error
     ) {
+      /* eslint-disable @typescript-eslint/consistent-type-assertions -- 
+      Typescript infercente get's lost here, but we know that if we didn't get
```
schaable, line 85: typo: _infercente_, I don't know why it doesn't let me submit a code suggestion.

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -46,12 +38,9 @@ describe("main", function () {
...
         const usedCliArguments = [false, true, true, false];
 
-        const res = parseTaskAndArguments(
-          cliArguments,
-          usedCliArguments,
-          hre,
-        ) as TaskRes;
+        const res = parseTaskAndArguments(cliArguments, usedCliArguments, hre);
 
+        assert.ok(!Array.isArray(res));
```
schaable, line 43: Add a message to avoid the bug in `assert.ok`

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -88,12 +77,9 @@ describe("main", function () {
...
         const usedCliArguments = [false, false, false];
 
-        const res = parseTaskAndArguments(
-          cliArguments,
-          usedCliArguments,
-          hre,
-        ) as TaskRes;
+        const res = parseTaskAndArguments(cliArguments, usedCliArguments, hre);
 
+        assert.ok(!Array.isArray(res));
```
schaable, line 82: same here

schaable: and in the other `assert.ok`s of the file

schaable (approved):
Just a few comments. Looks good!

v-next/hardhat/test/internal/cli/main.ts
alcuadrado: done

v-next/hardhat-errors/src/errors.ts
alcuadrado: Thanks!

PR #5299, opened 2024-05-31
Make the build-system work with solcjs

alcuadrado:
I discovered that the hardhat-build-systems weren't working anymore locally (maybe after recreating my devcontainer?), because they don't work on ARM+linux. The reason is that on ARM you need to use solcjs, and that wasn't ported to ESM.

Note: This just copied the logic from v2 and ESM-ified it.

### DIFF
--- a/v-next/hardhat-build-system/package.json
+++ b/v-next/hardhat-build-system/package.json
@@ -52,6 +52,7 @@
     "p-map": "^4.0.0",
     "resolve": "1.17.0",
     "semver": "^7.6.2",
+    "solc": "0.7.3",
     "undici": "^6.16.1"
   },
   "devDependencies": {
--- a/v-next/hardhat-build-system/src/internal/solidity/compiler/index.ts
+++ b/v-next/hardhat-build-system/src/internal/solidity/compiler/index.ts
@@ -2,6 +2,7 @@ import { execFile } from "node:child_process";
 import * as fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
+import { fileURLToPath } from "node:url";
 
 import {
   HardhatError,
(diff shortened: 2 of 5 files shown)


v-next/hardhat-build-system/package.json
```diff
@@ -52,6 +52,7 @@
     "p-map": "^4.0.0",
     "resolve": "1.17.0",
     "semver": "^7.6.2",
+    "solc": "0.7.3",
```
kanej, line 55: How should I think about the solc version we would use here. We are porting the `hardhat-core` version here, but shouldn't we be pulling in the latest version (and keeping it up to date)?

v-next/hardhat-build-system/src/internal/solidity/compiler/solcjs-runner.ts
```diff
@@ -0,0 +1,39 @@
...
+
+async function main() {
+  const input = await readStream(process.stdin);
+
+  const solcjsPath = process.argv[2];
+  const solc = await getSolcJs(solcjsPath);
+
+  const output = solc.compile(input);
+
+  console.log(output);
```
kanej, line 33: Can I check, this is how the output is being passed between this child process and the parent process - that console.logging here writes to the stdout?

kanej (approved):

v-next/hardhat-build-system/src/internal/solidity/compiler/index.ts
```diff
@@ -17,66 +18,55 @@ export interface ICompiler {
...
+      : [];
 
-    const previousHook = Module._extensions[".js"];
+    const output: string = await new Promise((resolve, reject) => {
+      try {
+        const subprocess = execFile(
+          process.execPath,
+          [...nodeOptions, scriptPath, this.#pathToSolcJs],
+          {
+            maxBuffer: 1024 * 1024 * 500,
```
ChristopherDedominici, line 42: Could we add a short comment explaining why this specific memory size was chosen?

ChristopherDedominici (approved):
Just a small minor comment, but other than that all good

v-next/hardhat-build-system/package.json
alcuadrado: I just deferred the decision tbh. IIRC in v2 we pull the one that's part of the default config, but that doesn't add a ton of value, as you are gonna want to download the native compiler anyway.

v-next/hardhat-build-system/src/internal/solidity/compiler/index.ts
alcuadrado: TBH I copied that from the other compiler, so I don't really know 😅

PR #5300, opened 2024-05-31
Implement feedback on task, subtask and parameter cli parsing

ChristopherDedominici:
See the original PR with the comments to implement: https://github.com/NomicFoundation/hardhat/pull/5282.

Every comment is addressing a specific issue raised in the PR linked above. 

There are also additional commits to improve a few smaller code parts, but also in this case everything should be quite easy to understand from the commit message


alcuadrado (approved):
I think this looks good.

Two comments, which can be addressed on separate PRs:

1) As discussed, `file` type should be just string, without any validation.
2) Flags can have explicit values too, for consistency. e.g. `--flag` and `--flag true` are both valid an equivalent.

I think we can merge this PR and address these things on separate PRs.

PR #5308, opened 2024-05-31
Added tests to task manager 

schaable:

NOTES:
 - Read note at the top on why I'm testing this through `createHardhatRuntimeEnvironment` and not directly.
 - There is some duplication in the tests, but I avoided extracting it to keep things simple. 
 - Technically, we could repeat many of the tests with config tasks, but that seemed like overkill.

### DIFF
--- a/v-next/core/test/internal/tasks/task-manager.ts
+++ b/v-next/core/test/internal/tasks/task-manager.ts
@@ -1,7 +1,3 @@
-import type { HardhatUserConfig } from "../../../src/types/config.js";
-import type { GlobalParameterMap } from "../../../src/types/global-parameters.js";
-import type { HardhatPlugin } from "../../../src/types/plugins.js";
-
 import assert from "node:assert/strict";
 import { describe, it } from "node:test";
 
@@ -10,7 +6,6 @@ import { HardhatError } from "@nomicfoundation/hardhat-errors";
 import { ParameterType } from "../../../src/config.js";
 import { createHardhatRuntimeEnvironment } from "../../../src/index.js";
 import { buildGlobalParameterDefinition } from "../../../src/internal/global-parameters.js";
-import { resolvePluginList } from "../../../src/internal/plugins/resolve-plugin-list.js";
 import {
   NewTaskDefinitionBuilderImplementation,
   TaskOverrideDefinitionBuilderImplementation,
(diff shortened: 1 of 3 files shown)


v-next/core/test/internal/tasks/task-manager.ts
```diff
@@ -0,0 +1,908 @@
...
+ * TaskManagerImplementation.
+ */
+async function createHre({
+  config = {},
+  plugins = [],
+}: {
+  config?: HardhatUserConfig;
+  plugins?: HardhatPlugin[];
+  globalParameterMap?: GlobalParameterMap;
+} = {}) {
```
alcuadrado, line 37: This function is not needed. Calling `createHardhatRuntimeEnvironment` has the same effect, or should have it. The third parameter shouldn't be used anywhere except for the CLI.

v-next/core/test/internal/tasks/task-manager.ts
```diff
@@ -0,0 +1,908 @@
...
+    {},
+    {
+      resolvedPlugins,
+    },
+  );
+}
+
+describe("TaskManagerImplementation", () => {
+  it("should initialize the task manager with an empty set of tasks if no plugins or tasks are provided", async () => {
+    await assert.doesNotReject(createHre());
```
alcuadrado, line 51: This test isn't really testing what its description does. At the same time, I don't think we have a way to actually test it, which makes me think that the `TaskManager` interface is incomplete. Maybe it should be able to either list all the tasks, or at least all the root tasks.

A similar problem will appear when printing the help messages.

alcuadrado: Maybe add a `getTopLevelTasks()` in a separate PR?

v-next/core/test/internal/tasks/task-manager.ts
```diff
@@ -0,0 +1,908 @@
...
+    const subsubtask1 = hre.tasks.getTask(["task1", "subtask1", "subsubtask1"]);
+    assert.deepEqual(subsubtask1.id, ["task1", "subtask1", "subsubtask1"]);
+    assert.equal(subsubtask1.pluginId, "plugin2");
+  });
+
+  /**
+   * These are all tested with plugin tasks, but the same logic applies to config tasks
+   */
+  describe("errors", () => {
+    it("should throw if there's a global parameter with the same name as a task named parameter", async () => {
```
alcuadrado, line 386: Also positional

v-next/core/test/internal/tasks/task-manager.ts
```diff
@@ -0,0 +1,908 @@
...
+            },
+          ],
+        }),
+        new HardhatError(HardhatError.ERRORS.TASK_DEFINITIONS.TASK_NOT_FOUND, {
+          task: "task1",
+        }),
+      );
+    });
+
+    it("should throw if trying to override a task and there is a name clash with an exising named parameter", async () => {
```
alcuadrado, line 556: also with positional params

v-next/core/test/internal/tasks/task-manager.ts
```diff
@@ -0,0 +1,908 @@
...
+          {
+            actorFragment: "Plugin plugin2 is",
+            namedParamName: "flag1",
+            task: "task1",
+          },
+        ),
+      );
+    });
+
+    it("should throw if a plugins tries to override a task defined in the config", async () => {
```
alcuadrado, line 836: Nice test case

alcuadrado (approved):
LGTM. I left 3 comments, one of which should be addressed on a separate PR.

v-next/core/test/internal/tasks/task-manager.ts
schaable: Currently, it's only checking against named params, see https://github.com/NomicFoundation/hardhat/blob/6413a98470ba3c8671b41f5d90342ada56ea6cce/v-next/core/src/internal/tasks/task-manager.ts#L188

Does it also need to validate against positional parameters? If so, I can add the validation & tests in a follow-up PR.

alcuadrado: Yes, i think it's better to be conservative. A follow up pr looks better.

PR #5329, opened 2024-06-01
Custom reporter GitHub integration and better formatting of special cases

alcuadrado:
This PR removes `@reporters/github` in favor of integrating directly ourselves from our reporter. 

While testing it I discovered two small formatting improvements that could be made, so I also included them here.

Note that there's a commit that includes an intentional test failure to test the integration. I'll remove it before merging.


v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -21,6 +23,14 @@ export function formatError(error: Error): string {
     );
   }
 
+  if (isTestFileExecutionFailureError(error)) {
```
schaable, line 26: 💯

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -57,10 +68,9 @@ ${stack}`;
 
 // TODO: Do this in a more robust way and that works well with windows
```
schaable, line 69: is the TODO still needed?

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -57,10 +68,9 @@ ${stack}`;
...
 function replaceFileUrlsWithRelativePaths(stack: string): string {
-  return stack.replaceAll(
-    "(" + pathToFileURL(process.cwd() + "/").toString(),
-    "(",
-  );
+  return stack
+    .replaceAll("(" + pathToFileURL(process.cwd() + path.sep).toString(), "(")
+    .replaceAll("(" + process.cwd() + path.sep, "(");
```
schaable, line 73: what's the difference with the previous replaceAll?

v-next/hardhat-node-test-reporter/src/github-actions.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  const error = event.details.error;
+
+  const location = getLocation(error);
+  if (location === undefined) {
+    return;
+  }
+
+  const { default: core } = await import("@actions/core");
```
schaable, line 25: do we need to import packages dynamically in the reporter?

v-next/hardhat-node-test-reporter/src/error-formatting.ts
alcuadrado: Yes, this is still a bit of a hack IMO. For example, see your later comment. Sometimes stack traces use file-urls, sometimes paths. Unclear when, and if we are properly handling everything. Eventually, we'll want to polish this, properly test it, etc.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
alcuadrado: See my previous comment

v-next/hardhat-node-test-reporter/src/github-actions.ts
alcuadrado: TBH it's unclear. Not very well documented by github if this package is safe outside of github actions. 

Also, locally it won't be used it, so it makes sense not to import it.

schaable (approved):

PR #5334, opened 2024-06-04
Implement cli global parameters parsing

ChristopherDedominici:
Implements:
- parsing of user global parameters: modify the function to parse the named parameters to also parse global parameters
- parsing of initial hh params (configPath, showStackTraces, help and version). The logic was already implemented, I just moved it into a function to improve readability and added a few lines of codes


v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -166,29 +135,91 @@ export async function main(cliArguments: string[]) {
...
-  // - Mark all the used entries in usedCliArguments
-  // - Do not resolve defaults here, the HRE does that.
-  // - Return the values
-  return {};
+export async function parseInitialHardhatParameters(
+  cliArguments: string[],
+  usedCliArguments: boolean[],
+) {
+  let configPath: string | undefined;
+  let showStackTraces: boolean = false; // true if ci
```
ChristopherDedominici, line 150: @alcuadrado should we use the hh-core function [isRunningOnCiServer](https://github.com/NomicFoundation/hardhat/blob/f579adca83f0a4c47bf29cd5df9ebc9347443039/packages/hardhat-core/src/internal/util/ci-detection.ts#L20)

alcuadrado: That's a good point. I added a new issue for this: https://github.com/NomicFoundation/hardhat/issues/5337

I think we may reuse it from plugins/libs, so let's place it in `hardhat-utils`.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -166,29 +135,91 @@ export async function main(cliArguments: string[]) {
...
+      usedCliArguments[i] = true;
+
+      if (configPath !== undefined) {
+        throw new HardhatError(HardhatError.ERRORS.ARGUMENTS.DUPLICATED_NAME, {
+          name: "--config",
+        });
+      }
+
+      if (
+        usedCliArguments[i + 1] === undefined ||
```
schaable, line 167: can we simplify this expression to `usedCliArguments[i + 1] !== false`?

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -28,63 +29,31 @@ import { getHardhatRuntimeEnvironmentSingleton } from "../hre-singleton.js";
 
 export async function main(cliArguments: string[]) {
   const hreInitStart = performance.now();
-  let configPath: string | undefined;
-  let showStackTraces: boolean = false; // true if ci
-  let help: boolean = false;
-  let version: boolean = false;
 
   const usedCliArguments = new Array(cliArguments.length).fill(false);
```
schaable, line 33:
```suggestion
  const usedCliArguments: boolean[] = new Array(cliArguments.length).fill(false);
```

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -166,29 +135,91 @@ export async function main(cliArguments: string[]) {
...
+export async function parseGlobalArguments(
+  globalParamsIndex: GlobalParameterMap,
+  cliArguments: string[],
+  usedCliArguments: boolean[],
+): Promise<Partial<GlobalArguments>> {
+  const globalArguments: Partial<GlobalArguments> = {};
+
+  const parameters = new Map(
+    Array.from(globalParamsIndex, ([key, value]) => [key, value.param]),
+  );
```
schaable, line 213:
```suggestion
  const parameters = new Map(
    [...globalParamsIndex].map(([key, value]) => [key, value.param]),
  );
```

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -358,7 +390,7 @@ async function parseNamedParameters(
         (cliArguments[i + 1] === "true" || cliArguments[i + 1] === "false")
```
schaable, line 390: it doesn't let me comment on the right line, but the expression could be simplified as `usedCliArguments[i + 1] === false && (cliArguments[i + 1] === "true" || cliArguments[i + 1] === "false")`

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -379,7 +411,7 @@ async function parseNamedParameters(
     ) {
```
schaable, line 411: same here, you don't need the `usedCliArguments[i + 1] !== undefined` check

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -399,10 +431,7 @@ async function parseNamedParameters(
   }
 
   // Check if all the required parameters have been used
-  validateRequiredParameters(
-    Array.from(task.namedParameters.values()),
-    taskArguments,
-  );
+  validateRequiredParameters(Array.from(parametersMap.values()), argumentsMap);
```
schaable, line 434:
```suggestion
  validateRequiredParameters([...parametersMap.values()], argumentsMap);
```

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -28,63 +29,31 @@ import { getHardhatRuntimeEnvironmentSingleton } from "../hre-singleton.js";
...
-      continue;
-    }
-
-    if (arg === "--version") {
-      usedCliArguments[i] = true;
-      version = true;
-      continue;
-    }
-  }
+  const hhGlobalParams = await parseInitialHardhatParameters(
```
schaable, line 35: The function and variable names are a bit confusing, as we have another function with a similar name (`parseGlobalArguments`). What do you think about `parseSpecialCliArguments`? For the variable, I'd consider destructuring it, but another option could be `hardhatSpecialArgs`.

schaable (changes requested):
I've left a couple of suggestions. Let me know if anything doesn't make sense :)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: Yes, it would work, but in a previous PR Pato mentioned that we should be explicit when checking for out of bounds checks

schaable: ohh got it. Then you can ignore it!

v-next/hardhat/src/internal/cli/main.ts
schaable: N/A, see https://github.com/NomicFoundation/hardhat/pull/5334#discussion_r1628104331

v-next/hardhat/src/internal/cli/main.ts
schaable: N/A, see https://github.com/NomicFoundation/hardhat/pull/5334#discussion_r1628104331

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: This would be a bit less confusing if it were `i < usedCliArguments.length &&`

schaable: I'd prefer that. The current check doesn't make the intention explicit.

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -47,16 +60,160 @@ async function getTasksAndHreEnvironment(
...
+        true,
+        false,
+      ]);
+      assert.equal(configPath, "./path-to-config");
+      assert.equal(showStackTraces, true);
+      assert.equal(help, true);
+      assert.equal(version, true);
+    });
+
+    it("should not set any hardhat initial global parameters except the path to the config", async function () {
```
alcuadrado, line 91: I think this title is incorrect

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: Agreed. I mostly find it confusing that the names are not consistent. I'm ok with "initial", but here they are both all initial and global, and there's another global.

alcuadrado (commented):
Left some minor comments. Otherwise looks good

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5334/commits/52d98a27918466f91da88df82eccaba54e02eac3)

v-next/hardhat/test/internal/cli/main.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5334/commits/f45d1f1e4442fad9871c2ced4fe4bc53899dbdca)

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -47,16 +60,160 @@ async function getTasksAndHreEnvironment(
...
+        usedCliArguments,
+        new Array(cliArguments.length).fill(false),
+      );
+      assert.equal(configPath, undefined);
+      assert.equal(showStackTraces, false);
+      assert.equal(help, false);
+      assert.equal(version, false);
+    });
+
+    it("should throw an error because the config param is passed twice", async function () {
```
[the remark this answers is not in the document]
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5334/commits/f45d1f1e4442fad9871c2ced4fe4bc53899dbdca)

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: I created [this PR](https://github.com/NomicFoundation/hardhat/pull/5340) to address this, it should be merged before this one.

ChristopherDedominici: Done, the CI function is now included

schaable (approved):
Seal of approval 👍

alcuadrado (approved):

Issue #5247, opened 2024-05-17
Re-enable the e2e tests on Windows


alcuadrado:
We disabled the e2e tests on windows because they were taking >5x more time (>20min) than on Windows or Mac. 

This is probably due to the dependencies not being cached in those GA workflows.

michalbrabec:
No additional CI work planned for HH2

PR #5339, opened 2024-06-05
Optimize and Enable E2E tests on Windows

sisco0:

---

This pull request introduces several optimizations to the existing `e2e-tests.yml` workflow to improve the overall execution time and efficiency of our end-to-end (e2e) tests on GitHub Actions as required in #5247.

**Changes Introduced**:
- [x] **Re-enable Windows Testing**:
The original workflow had Windows tests disabled due to performance issues and a bug with pnpm. We have re-enabled Windows testing with specific conditions to handle Windows-specific issues, ensuring that pnpm operations are executed correctly.
- [x] **Matrix Strategy Optimization**:
Ensured that the jobs are run in parallel across ubuntu-latest, macos-latest, and windows-latest, reducing the overall execution time.
- [x] **Caching Dependencies**:
Added caching for pnpm modules using actions/cache@v3 to avoid re-downloading and installing dependencies in every run. This change significantly reduces setup time for the jobs.
- [x] **Improved Script Handling**:
Introduced conditional execution of scripts based on the operating system. For non-Windows systems, the scripts are executed as before. For Windows, equivalent PowerShell scripts are used to ensure compatibility.

No testing has been performed over the current solution.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -45,7 +45,6 @@ jobs:
           chmod +x run-fixture-projects.sh
           ./run-fixture-projects.sh
         shell: bash
-        if: runner.os != 'Windows'
 
       - name: Run test-project-initialization script
         run: |
@@ -53,13 +52,4 @@ jobs:
           chmod +x test-project-initialization.sh
           ./test-project-initialization.sh
         shell: bash
-        if: runner.os != 'Windows'
         
-      - name: Run Windows-specific scripts
-        run: |
-          cd e2e
-          .\run-fixture-projects.ps1
-          .\test-project-initialization.ps1
-        shell: pwsh
-        if: runner.os == 'Windows'
-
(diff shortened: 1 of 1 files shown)


alcuadrado:
I'm a bit confused about this PR, it fails because you are trying to run a file that doesn't exist.

sisco0:
Reason behind was PowerShell scripts references, which are not currently present in the project.
Instead of commiting the new PowerShell (.ps1) files, references about these have been removed.
Re-factored the current source-code and pushed the commit for it.

alcuadrado:
Isn't this removing a comment about a bug without any reason? I don't think this has been fixed, is it?

sisco0:
> Isn't this removing a comment about a bug without any reason? I don't think this has been fixed, is it?

Comment was restored as requested.
Agree on the given fact. Ideally, a reference to a Github Issue should be added for properly handling the case actively.
This could be debugged in the future via https://github.com/nektos/act

alcuadrado:
Thanks once again for submitting this PR, @sisco0. Unfortunately, we won't be able to prioritize this just yet. Our CI needs are pretty different in our main development branch, so we can't put much effort into this right now.

PR #5340, opened 2024-06-05
Add function in hh-utils to check if process is running in the CI

ChristopherDedominici:
Implement a function to check if the process is running in GitHub CI. 
The logic combines the code from [pnpm](https://pnpm.io/cli/install#--frozen-lockfile) and the [hardhat code](https://github.com/NomicFoundation/hardhat/blob/bdd04d37f2ee41f68943f60bf0fd3ac5546a05ad/packages/hardhat-core/src/internal/util/ci-detection.ts#L20) in hh-core

### DIFF
--- a/v-next/hardhat-utils/src/ci.ts
+++ b/v-next/hardhat-utils/src/ci.ts
@@ -1,0 +1,6 @@
+/**
+ * Returns 'true' if the current process is running in the GitHub CI environment.
+ */
+export function isCI() {
+  return process.env.CI === "true";
+}
--- a/v-next/hardhat-utils/src/ci.ts
+++ b/v-next/hardhat-utils/src/ci.ts
@@ -1,6 +1,33 @@
 /**
- * Returns 'true' if the current process is running in the GitHub CI environment.
+ * Checks whether the current process is running in a CI environment.
+ *
+ * @returns True if the current process is running in a CI environment.
  */
-export function isCI() {
-  return process.env.CI === "true";
+export function isCi() {
+  const env = process.env;
+
+  return !!(
+    isGithubActions() ||
+    isNow() ||
+    isAwsCodeBuild() ||
+    env.CI !== undefined || // Travis CI, CircleCI, Cirrus CI, GitLab CI, Appveyor, CodeShip, dsari
+    env.CONTINUOUS_INTEGRATION !== undefined || // Travis CI, Cirrus CI
+    env.BUILD_NUMBER !== undefined || // Jenkins, TeamCity
+    env.RUN_ID !== undefined || // TaskCluster, dsari
+    false
+  );
+}
+
+function isGithubActions(): boolean {
+  return process.env.GITHUB_ACTIONS !== undefined;
+}
+
+function isNow() {
+  return (
+    process.env.NOW !== undefined || process.env.DEPLOYMENT_ID !== undefined
+  );
+}
+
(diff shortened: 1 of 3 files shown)


v-next/hardhat-utils/src/ci.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Returns 'true' if the current process is running in the GitHub CI environment.
+ */
+export function isCI() {
```
ChristopherDedominici, line 4: This can be either a const or a function, what do you prefer?

Note: if we go for a const, how should we handle the tests? 
Because the module "ci" should be dynamically imported for every test because first the ENV variables need to be modified. And we might have to use the ugly solution to force the reload of the module:

 
image

alcuadrado: let's go for a function

v-next/hardhat-utils/src/ci.ts
```diff
@@ -0,0 +1,6 @@
...
+  return process.env.CI === "true";
```
alcuadrado, line 5: As we discussed, let's use the same ones than PNPM uses, and let's double check that we cover the same cases than v2, but without any dependency. 

See packages/hardhat-core/src/internal/util/ci-detection.ts

Except for the `isLinuxWithoutDisplayServer()` function that's too restrictive. e.g. it flags devcontainer as a ci.

v-next/hardhat-utils/src/ci.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  return !!(
+    isGithubActions() ||
+    isNow() ||
+    isAwsCodeBuild() ||
+    env.CI !== undefined || // Travis CI, CircleCI, Cirrus CI, GitLab CI, Appveyor, CodeShip, dsari
+    env.CONTINUOUS_INTEGRATION !== undefined || // Travis CI, Cirrus CI
+    env.BUILD_NUMBER !== undefined || // Jenkins, TeamCity
+    env.RUN_ID !== undefined || // TaskCluster, dsari
+    false
```
alcuadrado, line 17: `|| false` doesn't add anything, does it?

v-next/hardhat-utils/src/ci.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Checks whether the current process is running in a CI environment.
+ *
+ * @returns True if the current process is running in a CI environment.
+ */
+export function isCi() {
+  const env = process.env;
+
+  return !!(
```
alcuadrado, line 9: `!!` of boolean doesn't add anything

alcuadrado (approved):
I left two minor comments, but looks good

v-next/hardhat-utils/src/ci.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5340/commits/d5ce13f963d5f7450bb946cb3c0933e48f895730)

v-next/hardhat-utils/src/ci.ts
ChristopherDedominici: [Done in this commit](https://github.com/NomicFoundation/hardhat/pull/5340/commits/d5ce13f963d5f7450bb946cb3c0933e48f895730)

Issue #5304, opened 2024-05-31
move formatParameterValue/parseParameterValue and isParameterValueValid to a util


kanej:

schaable:
Since `parseParameterValue` is currently only used by the CLI, we will leave it in `v-next/hardhat`. Therefore, this issue only refers to `isParameterValueValid`.

alcuadrado:
See #5379

PR #5341, opened 2024-06-05
Improve task types & validation

schaable:

 - Improves typings.
 - Makes `isParameterValueValid` work with arrays (for variadic parameters). This will be used for validation in the `run` function.

### DIFF
--- a/v-next/core/src/types/common.ts
+++ b/v-next/core/src/types/common.ts
@@ -23,6 +23,12 @@ export interface ParameterToValueTypeMap {
 }
 
 /**
+ * All the possible values for a parameter.
+ */
+export type ParameterValue =
+  ParameterToValueTypeMap[keyof ParameterToValueTypeMap];
+
+/**
  * Maps a `ParameterType` to its corresponding value type.
  *
  * This type takes a `ParameterType` as a type parameter and returns the type
@@ -57,11 +63,19 @@ const parameterTypeValidators: Record<
  *
  * This function uses a map of validators, where each validator is a function
  * that checks if a value is valid for a specific parameter type.
+ * If the parameter type is variadic, the value is considered valid if it is an
+ * array and all its elements are valid for the parameter type.
  */
 export function isParameterValueValid(
   type: ParameterType,
   value: unknown,
+  isVariadic: boolean = false,
 ): boolean {
   const validator = parameterTypeValidators[type];
+
+  if (isVariadic) {
+    return Array.isArray(value) && value.every(validator);
+  }
+
   return validator(value);
 }
--- a/v-next/hardhat/src/internal/cli/main.ts
+++ b/v-next/hardhat/src/internal/cli/main.ts
@@ -1,21 +1,30 @@
-import {
-  buildGlobalParameterMap,
-  resolvePluginList,
-} from "@nomicfoundation/hardhat-core";
-import { ParameterType } from "@nomicfoundation/hardhat-core/types/common";
-import {
+import type {
   GlobalArguments,
   GlobalParameter,
   GlobalParameterMap,
 } from "@nomicfoundation/hardhat-core/types/global-parameters";
-import { HardhatRuntimeEnvironment } from "@nomicfoundation/hardhat-core/types/hre";
-import {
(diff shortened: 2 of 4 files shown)


v-next/core/src/types/common.ts
```diff
@@ -57,11 +63,19 @@ const parameterTypeValidators: Record<
  *
  * This function uses a map of validators, where each validator is a function
  * that checks if a value is valid for a specific parameter type.
+ * If the parameter type is variadic, the value is considered valid if it is an
```
alcuadrado, line 66: I think we already discussed this, but `src/types` should only have types, not functions nor classes.

alcuadrado: Same for `parameterTypeValidators`, that's a value, not a type

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -26,6 +33,14 @@ import {
 } from "../helpers/config-loading.js";
 import { getHardhatRuntimeEnvironmentSingleton } from "../hre-singleton.js";
 
+/**
+ * The parsed arguments of a task. This is the result of parsing the raw
+ * arguments passed to a task.
+ */
+interface ParsedTaskArguments extends TaskArguments {
```
alcuadrado, line 40: I'm confused about this again, how is it different from TaskArguments?

v-next/core/src/types/common.ts
schaable: Yes, it's under a different issue. I'll do it once I've finished with the `run` method. See https://github.com/NomicFoundation/hardhat/issues/5304

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: Maybe the thing that needs to change is TaskArguemtns to have `ParameterValue | ParameterValue[]` instead of `any`.

schaable: These are the task arguments after they've been parsed by `parseTaskAndArguments`, so they can have better typings for the values. The interface It's only used here. If it's too confusing we could remove it and use `Record<string, ParameterValue | ParameterValue[]>` directly in the function return type instead.

alcuadrado: But shouldn't we also restrict `TaskArguments` to be `Record<string, ParameterValue | ParameterValue[]>`? that's what I meant.

schaable: The issue is that we have no guarantees that the arguments have been parsed by the CLI beforehand when the `run` method is called. It could be called programmatically with unknown values. We could restrict it anyway, as for the parameter to be valid, it needs to be a `ParameterValue`.

schaable: Just to leave the resolution documented: we've decided to use `TaskArguments` over `ParsedTaskArguments` with the improved typings of `Record<string, ParameterValue | ParameterValue[]>`, as any value that does not satisfy this constraint will not be a valid parameter value anyway.

schaable:
I'm going to hold off on merging this until https://github.com/NomicFoundation/hardhat/pull/5334 is merged.

schaable:
@alcuadrado @ChristopherDedominici Can you double-check that I didn't break anything with the rebase?

v-next/core/src/types/common.ts
```diff
@@ -57,11 +63,19 @@ const parameterTypeValidators: Record<
...
+ * array and all its elements are valid for the parameter type.
  */
 export function isParameterValueValid(
   type: ParameterType,
   value: unknown,
+  isVariadic: boolean = false,
 ): boolean {
   const validator = parameterTypeValidators[type];
```
ChristopherDedominici, line 74: Should we use the same function used [here ](https://github.com/NomicFoundation/hardhat/blob/c9df7e6799b151643dce0bbcc94f3a79d0130544/v-next/hardhat/src/internal/cli/main.ts#L506C10-L506C29) to validate the different types?

ChristopherDedominici (approved):

alcuadrado (approved):

PR #5344, opened 2024-06-05
Task run implementation

schaable:

### DIFF
--- a/v-next/core/src/internal/tasks/resolved-task.ts
+++ b/v-next/core/src/internal/tasks/resolved-task.ts
@@ -101,7 +101,7 @@ export class ResolvedTask implements Task {
       ...this.positionalParameters,
     ];
 
-    const providedParameterNames = new Set(Object.keys(taskArguments));
+    const providedArgumentNames = new Set(Object.keys(taskArguments));
     for (const parameter of allParameters) {
       const value = taskArguments[parameter.name];
 
@@ -113,13 +113,13 @@ export class ResolvedTask implements Task {
         taskArguments[parameter.name] = parameter.defaultValue;
       }
 
-      // Remove processed parameter from the set as it has been processed
-      providedParameterNames.delete(parameter.name);
+      // Remove processed parameter from the set
+      providedArgumentNames.delete(parameter.name);
     }
 
     // At this point, the set should be empty as all the task parameters have
     // been processed. If there are any extra parameters, an error is thrown
-    this.#validateExtraParameters(providedParameterNames);
+    this.#validateExtraArguments(providedArgumentNames);
 
     const next = async (
       nextTaskArguments: TaskArguments,
@@ -210,17 +210,17 @@ export class ResolvedTask implements Task {
   }
 
   /**
-   * Validates that no extra parameters were provided in the task arguments.
+   * Validates that no extra arguments were provided in the task arguments.
    *
-   * @throws HardhatError if extra parameters were provided. The error message
-   * includes the name of the first extra parameter.
+   * @throws HardhatError if extra arguments were provided. The error message
+   * includes the name of the first extra argument.
    */
-  #validateExtraParameters(providedParameterNames: Set<string>) {
-    if (providedParameterNames.size > 0) {
+  #validateExtraArguments(providedArgumentNames: Set<string>) {
+    if (providedArgumentNames.size > 0) {
       throw new HardhatError(
         HardhatError.ERRORS.ARGUMENTS.UNRECOGNIZED_NAMED_PARAM,
         {
-          parameter: [...providedParameterNames][0],
+          parameter: [...providedArgumentNames][0],
         },
       );
     }
--- a/v-next/core/src/internal/tasks/resolved-task.ts
+++ b/v-next/core/src/internal/tasks/resolved-task.ts
@@ -165,9 +165,10 @@ export class ResolvedTask implements Task {
   ) {
     if (parameter.defaultValue === undefined && value === undefined) {
       throw new HardhatError(
-        HardhatError.ERRORS.ARGUMENTS.MISSING_VALUE_FOR_PARAMETER,
+        HardhatError.ERRORS.TASK_DEFINITIONS.MISSING_VALUE_FOR_PARAMETER,
         {
-          paramName: parameter.name,
+          parameter: parameter.name,
+          task: formatTaskId(this.id),
         },
       );
     }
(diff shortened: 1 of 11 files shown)


v-next/core/src/internal/tasks/resolved-task.ts
```diff
@@ -69,24 +77,190 @@ export class ResolvedTask implements Task {
...
+    ): param is PositionalTaskParameter => "isVariadic" in param;
+    const isVariadic = isPositionalParameter(parameter) && parameter.isVariadic;
+
+    // check if the value is valid for the parameter type
+    if (!isParameterValueValid(parameter.parameterType, value, isVariadic)) {
+      throw new HardhatError(
+        HardhatError.ERRORS.ARGUMENTS.INVALID_VALUE_FOR_TYPE,
+        {
+          value,
+          name: parameter.name,
```
alcuadrado, line 204: Again, i think adding the task name would be valuable

v-next/core/src/internal/tasks/resolved-task.ts
```diff
@@ -69,24 +77,190 @@ export class ResolvedTask implements Task {
...
+   * @throws HardhatError if the parameter is required and doesn't have a value.
+   */
+  #validateRequiredParameter(
+    parameter: TaskParameter,
+    value: ParameterValue | ParameterValue[],
+  ) {
+    if (parameter.defaultValue === undefined && value === undefined) {
+      throw new HardhatError(
+        HardhatError.ERRORS.ARGUMENTS.MISSING_VALUE_FOR_PARAMETER,
+        {
```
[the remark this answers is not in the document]
alcuadrado: The reason is that tasks can call other tasks, making this harder to debug otherwise

v-next/core/src/internal/tasks/resolved-task.ts
```diff
@@ -69,24 +77,190 @@ export class ResolvedTask implements Task {
...
+   *
+   * @throws HardhatError if extra arguments were provided. The error message
+   * includes the name of the first extra argument.
+   */
+  #validateExtraArguments(providedArgumentNames: Set<string>) {
+    if (providedArgumentNames.size > 0) {
+      throw new HardhatError(
+        HardhatError.ERRORS.ARGUMENTS.UNRECOGNIZED_NAMED_PARAM,
+        {
+          parameter: [...providedArgumentNames][0],
```
alcuadrado, line 222: Same

v-next/core/src/internal/tasks/resolved-task.ts
```diff
@@ -69,24 +77,190 @@ export class ResolvedTask implements Task {
...
+    } catch (error) {
+      ensureError(error);
+      throw new HardhatError(
+        HardhatError.ERRORS.TASK_DEFINITIONS.INVALID_ACTION_URL,
+        {
+          action: actionFileUrl,
+          task: formatTaskId(taskId),
+        },
+        error,
+      );
```
alcuadrado, line 251: To be addressed in a separate PR, but here is where you need to "validate" the plugin's installation.

v-next/core/src/internal/tasks/resolved-task.ts
```diff
@@ -69,24 +77,190 @@ export class ResolvedTask implements Task {
...
+    // At this point, the set should be empty as all the task parameters have
+    // been processed. If there are any extra parameters, an error is thrown
+    this.#validateExtraArguments(providedArgumentNames);
+
+    const next = async (
+      nextTaskArguments: TaskArguments,
+      currentIndex = this.actions.length - 1,
+    ): Promise<any> => {
+      // The first action may be empty if the task was originally an empty task
+      const currentAction = this.actions[currentIndex].action ?? (() => {});
```
alcuadrado, line 128: Isn't this handled by `isEmpty` at the top? It think so.

If that's the case, I think it's better to have an `assertHardhatInvariant` and let it crash if the invariant doesn't hold.

v-next/core/src/internal/tasks/resolved-task.ts
```diff
@@ -69,24 +77,190 @@ export class ResolvedTask implements Task {
...
+    // skip type validation for optional parameters with undefined value
+    if (value === undefined && parameter.defaultValue !== undefined) {
+      return;
+    }
+
+    // check if the parameter is variadic
+    const isPositionalParameter = (
+      param: TaskParameter,
+    ): param is PositionalTaskParameter => "isVariadic" in param;
+    const isVariadic = isPositionalParameter(parameter) && parameter.isVariadic;
```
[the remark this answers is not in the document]
schaable: if you do that, `isVariadic` resolves to `unknown` :(

v-next/core/src/internal/tasks/resolved-task.ts
schaable: Yes, I've added it to the main task issue description.

v-next/core/src/internal/tasks/resolved-task.ts
schaable: This is a different scenario. The `isEmpty` check at the top addresses empty tasks with just one action (the one from the empty task). However, you can override the empty task and add an action to it. In this case, you need to define an empty function to avoid errors when the override action calls `runSuper`. Does that make sense? There's a test that may clarify the scenario.

v-next/core/src/internal/tasks/resolved-task.ts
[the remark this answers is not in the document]
schaable: I'll have to duplicate the descriptor because this one it's used in `v-next/hardhat/src/internal/cli/main.ts`, where there's no `taskId` to provide.

v-next/core/src/internal/tasks/resolved-task.ts
schaable: Same here

v-next/core/src/internal/tasks/resolved-task.ts
schaable: And same here 😅

v-next/core/src/internal/tasks/resolved-task.ts
alcuadrado: Makes sense

v-next/core/test/internal/tasks/task-manager.ts
```diff
@@ -875,4 +877,661 @@ describe("TaskManagerImplementation", () => {
...
+      assert.equal(overrideTaskRun, true);
+    });
+
+    it("should run a task with an action url", async () => {
+      const actionUrl = pathToFileURL(
+        path.resolve(
+          import.meta.dirname,
+          "./fixture-projects/file-actions/action-fn.js",
+        ),
+      ).toString();
```
alcuadrado, line 1173: FYI:

`import.meta.resolve("./fixture-projects/file-actions/action-fn.js")` is equivalent

alcuadrado (approved):
I left a few comments, but LGTM

v-next/core/test/internal/tasks/task-manager.ts
schaable: Nice, I'll change it just to have it as a reference for the future.

PR #5356, opened 2024-06-10
Ethers import missing in documentation

bloodybit:

---


bloodybit:
Notionally, it would also be nice to explain why importing from "hardhat" and not from "ethers" package direcly.

kanej:
Later on in the tutorial we explain that `ethers` is available in the global scope but can be imported to be explicit:

image

Our convention in the docs is to use the implicit versions in the global scope, and in typescript to explicitly import it.

I am going to close this as the current version is in conformance with our approach here. The wider question as to whether we should load `ethers` and others into the global scope we will revisit in Hardhat v3.

PR #5371, opened 2024-06-11
Check for missing dependencies when failing to import a file action

schaable:

Also added the default specified in the JSDoc for `basePathForNpmResolution` in `detectPluginNpmDependencyProblems` (double checked with Pato).

### DIFF
--- a/v-next/core/src/internal/tasks/resolved-task.ts
+++ b/v-next/core/src/internal/tasks/resolved-task.ts
@@ -130,11 +130,13 @@ export class ResolvedTask implements Task {
     ): Promise<any> => {
       // The first action may be empty if the task was originally an empty task
       const currentAction = this.actions[currentIndex].action ?? (() => {});
-
       const actionFn =
         typeof currentAction === "function"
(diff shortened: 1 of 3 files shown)


v-next/core/src/internal/tasks/resolved-task.ts
```diff
@@ -237,18 +241,31 @@ export class ResolvedTask implements Task {
...
     actionFileUrl: string,
-    taskId: string[],
   ): Promise<NewTaskActionFunction | TaskOverrideActionFunction> {
     let resolvedActionFn;
     try {
       resolvedActionFn = await import(actionFileUrl);
     } catch (error) {
       ensureError(error);
+
+      if (this.pluginId !== undefined) {
```
alcuadrado, line 251: I believe this logic is incorrect. Because it always validates the first plugin that defined this task. We should validate the one defining the current action instead.

alcuadrado (changes requested):
See the comment

alcuadrado (approved):

PR #5377, opened 2024-06-11
Improve task not found logic

ChristopherDedominici:
Currently, when a task is not found, an error is thrown. 
However, we need a more graceful approach so that we can print a help message in the CLI.

### DIFF
--- a/v-next/hardhat/test/internal/cli/main.ts
+++ b/v-next/hardhat/test/internal/cli/main.ts
@@ -415,6 +415,17 @@ describe("main", function () {
         });
       });
 
+      it("should throw because the task parameter is not defined", function () {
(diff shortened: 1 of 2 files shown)


v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -415,6 +415,17 @@ describe("main", function () {
         });
       });
 
+      it("should throw because the task parameter is not defined", function () {
```
alcuadrado, line 418: I think this title is probably a copy-paste accident.
```suggestion
      it("should return the task id if not found", function () {
```

alcuadrado (approved):
I just left a small comment, otherwise LGTM

schaable (approved):

PR #5380, opened 2024-06-11
Create an integration test to combine all the parsing steps (global parameters, named parameters, positional, etc.)

ChristopherDedominici:
Add "integration" tests for the CLI parsing function called "main".

Some of the tests are marked as "TODO" because the code logic to handle these scenarios is not implemented yet.
However, these tests are already covering the right branch that will be explored by the code logic. Only the messages displayed in the CLI will need to be changed (plus, maybe, minor adjustments).
Example: [link to a TODO](https://github.com/NomicFoundation/hardhat/pull/5380/files#:~:text=//%20TODO%3A%20update%20with%20a%20real%20task%20as%20soon%20as%20they%20are%20implemented)

If, during the PR review, you agree with the TODOs, I'll create work items to track them.


v-next/hardhat/test/fixture-projects/cli/type-validation/test.txt
ChristopherDedominici, line 1: This file in the fixture projects is no longer used

v-next/hardhat/test/fixture-projects/cli/parsing/subtask-help/hardhat.config.ts
```diff
@@ -0,0 +1,10 @@
+import {
+  HardhatUserConfig,
+  emptyTask,
+} from "@nomicfoundation/hardhat-core/config";
+
+const customTask = emptyTask("empty-task", "empty task description").build();
```
ChristopherDedominici, line 6: @alcuadrado  this is how I build the empty task.
The code logic, I think, should go in this if branch:

image

But it does not, [this test is failing](https://github.com/NomicFoundation/hardhat/pull/5380/files#diff-0c92acf051033f1d9388e08f2cf7363ce89dc689512eb4c57ae00847c1e4bd20:~:text=it(%22should%20print%20an%20help%20message%20for%20the%20task%27s%20subtask%22%2C%20async%20function%20()%20%7B). Can you double check if I'm misinterpreting how empty task should work or if I should open an issue?

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -57,7 +62,196 @@ async function getTasksAndHreEnvironment(
...
+        const cliArguments = command.split(" ").slice(2);
+
+        await main(cliArguments);
+
+        const results = await getTasksAndSubtaskResults();
+        assert.deepEqual(results, [true, false, false]);
+      });
+    });
+
+    // TODO: as soon as the 'help task' is done, this test should be updated
```
ChristopherDedominici, line 203: @zoeyTM I know that you're currently working on the help task. I created this test and the following ones as placeholder so you just need to modify them later (or copy the logic)

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -57,7 +62,196 @@ async function getTasksAndHreEnvironment(
...
+        const m = mock.method(console, "log", () => {});
+        const spy = mock.fn(importUserConfig);
+
+        const command = "npx hardhat --version";
+        const cliArguments = command.split(" ").slice(2);
+
+        await main(cliArguments);
+
+        assert.equal(m.mock.calls.length, 1);
+        assert.equal(m.mock.calls[0].arguments[0], "3.0.0");
```
schaable, line 98: can't we get the hardhat version from the `package.json`? Otherwise, every time hardhat is published, this test will fail.

ChristopherDedominici: Yes, this is a TODO, it is already implemented in this PR: https://github.com/NomicFoundation/hardhat/pull/5384

schaable (approved):
Looks good in general. As a note: I'm fine with the TODOs, but I'd like some of them to be more explicit about what needs to be done. Also we can mark the `it` as TODO with `it.todo(...`.

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -57,7 +62,199 @@ async function getTasksAndHreEnvironment(
...
+  // This function imports that array, allowing the tests to verify if the tasks have been executed.
+  // If a boolean flag is true, it indicates that the corresponding task (or a specific part of it) has been executed.
+  // The array is set in the hardhat.config.ts file of the fixture project.
+  return (
+    await import(pathToFileURL(`${process.cwd()}/${configFileName}`).toString())
+  ).results;
+}
+
 describe("main", function () {
+  describe("main", function () {
```
kanej, line 78: This `describe("main", ...` is duplicated

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -57,7 +62,199 @@ async function getTasksAndHreEnvironment(
...
+
+    describe("version", function () {
+      useFixtureProject("cli/parsing/base-project");
+
+      // TODO: as soon as the 'version task' is done, this test should be updated
+      it.todo(
+        "should print the version and instantly return",
+        async function () {
+          const m = mock.method(console, "log", () => {});
+          const spy = mock.fn(importUserConfig);
```
kanej, line 92: I am not sure about this level of mocking. I don't think we should do it at the level of `console.log`.

I think I would prefer an approach where `main` takes additional optional parameters for logging and for importing a user config.

In tests we inject functions or mock spys. In `cli.ts` we pass in the actual implementation.

What do we think?

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -57,7 +62,199 @@ async function getTasksAndHreEnvironment(
...
+        assert.equal(m.mock.calls[1].arguments[0], "Global help");
+      });
+    });
+
+    // TODO: as soon as the 'help task' is done, this test should be updated
+    describe("subtask help", function () {
+      useFixtureProject("cli/parsing/subtask-help");
+
+      it.todo(
+        "should print an help message for the task's subtask",
```
kanej, line 225: I think most of the time you can get away with either leaving out the todos (the follow task will have a testing requirement) or just putting in the string description.

v-next/hardhat/test/internal/cli/main.ts
ChristopherDedominici: It's ugly, but it's expected. Because the name of the file that I'm testing is "main" and the function that I'm testing is "main" too

v-next/hardhat/test/internal/cli/main.ts
ChristopherDedominici: @alcuadrado what do you think? I'm asking about this specific scenario, but also in general. How should we approach spying and mocking in the new v3 code?

alcuadrado: I think I prefer @kanej's approach, as this level of mocking is assuming a certain implementation.

I think `main` can take an optional `print = console.sol` param, or something along those lines.

The change here would be small.

alcuadrado (approved):
Looks good, but I think @kanej's comment should be addressed. Maybe in a follow-up PR?

A small nitpicky comment: the tests would be easier to read if they set their values in an object with descriptive fields, or something like that, instead of having to check what each index in the array means.

v-next/hardhat/test/internal/cli/main.ts
ChristopherDedominici: Done in this commit with readability improvement for tests too (like mentioned in your last commit @alcuadrado): https://github.com/NomicFoundation/hardhat/pull/5380/commits/213f009b955f89a9f5e3565d5c0fb02b266196e3

alcuadrado (approved):

Issue #5360, opened 2024-06-10
`version` task


kanej:
A parameter to show the current version of Hardhat:

```bash
hardhat --version
# 3.0.0
```

Version is a global parameter. Its presence should short circuit all other commands

alcuadrado:
Small clarification, this isn't technically a task

PR #5384, opened 2024-06-12
Implement logic for the --version cli global argument

ChristopherDedominici:
[Task description](https://github.com/NomicFoundation/hardhat/issues/5360)

[This PR](https://github.com/NomicFoundation/hardhat/pull/5380) should be merged first.

CHANGES:
- add a function to console.log the Hardhat version
- add a function to get the Hardhat version in the 'hardhat package'  at the path src/internal/utils/package.ts. For this function no test has been added because not very useful: a test would repeat the same logic in the test file as the original function because the hh package version cannot be hardcoded, since it constantly changes

### DIFF
--- a/v-next/hardhat/src/internal/cli/version.ts
+++ b/v-next/hardhat/src/internal/cli/version.ts
@@ -1,0 +1,5 @@
+import { getHardhatVersion } from "../utils/package.js";
+
+export async function printVersionMessage(print = console.log) {
+  print(await getHardhatVersion());
+}
--- a/v-next/hardhat/src/internal/utils/package.ts
+++ b/v-next/hardhat/src/internal/utils/package.ts
@@ -1,0 +1,22 @@
+import type { PackageJson } from "@nomicfoundation/hardhat-utils/package";
+
+import { fileURLToPath } from "node:url";
+
+import { assertHardhatInvariant } from "@nomicfoundation/hardhat-errors";
+import { readJsonFile } from "@nomicfoundation/hardhat-utils/fs";
+import { findClosestPackageJson } from "@nomicfoundation/hardhat-utils/package";
+
+export async function getHardhatVersion(): Promise<string> {
+  const packageJsonPath = await findClosestPackageJson(
+    fileURLToPath(import.meta.url),
+  );
+
+  assertHardhatInvariant(
+    packageJsonPath !== null,
+    "There should be a package.json in hardhat's root directory",
+  );
+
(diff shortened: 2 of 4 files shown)


ChristopherDedominici:
@alcuadrado @schaable should we export the function "getHardhatVersion" outside of the hh-utils package so I can use it here too? Do you think that this function could be used in multiple places a part from here?

schaable:
> @alcuadrado @schaable should we export the function "getHardhatVersion" outside of the hh-utils package so I can use it here too? Do you think that this function could be used in multiple places a part from here?

I thought about this when I was working on the utils. It is also useful for the `User-Agent` header in the `request` module, but I didn't add it because the implementation was not trivial. Here, you can read the closest `package.json` relative to `import.meta.url` to get the version. However, if you run the same code within `hardhat-utils`, you'll get the version of that package. You'd need to navigate up the `node_modules` structure, find Hardhat, and return its version. Alternatively, you could accept the file URL relative to the `package.json` as a parameter, but at that point, it is not that different from what you have in this PR. We may want to add this as a utility within the Hardhat package itself.

v-next/hardhat/src/internal/utils/package.ts
```diff
@@ -0,0 +1,22 @@
...
+  const packageJsonPath = await findClosestPackageJson(
+    fileURLToPath(import.meta.url),
+  );
+
+  assertHardhatInvariant(
+    packageJsonPath !== null,
+    "There should be a package.json in hardhat's root directory",
+  );
+
+  const packageJson: PackageJson = await readJsonFile(packageJsonPath);
```
schaable, line 19: there a util that does this already in the utils:
```typescript
const packageJson = await readClosestPackageJson(import.meta.url);
```

ChristopherDedominici: Good call! Done it 😎

alcuadrado (approved):

v-next/hardhat/src/internal/cli/version.ts
```diff
@@ -0,0 +1,5 @@
+import { getHardhatVersion } from "../utils/package.js";
```
ChristopherDedominici, line 1: This file is very short so it doesn't really make sense BUT it is for consistency. The logic for these "special hh commands" should be moved in separate files to reduce the amount of code in the main.ts file

alcuadrado (approved):

PR #5386, opened 2024-06-12
update node versions supported to include node 22

kanej:
This PR does three things:

1. Adds Node.js v22 support
2. Updates the stability guarantees text in our docs to clarify our version support policy
3. Updates the logic for displaying the unsupported version warning

## Node.js v22 support

The [node releases repo](https://github.com/nodejs/Release) indicate that node 22 is in the _Current_ release status. So we should support it.

## Improved stability guarantee text

We are updating our stability guarantees text to indicate that we will support **even numbered major versions** with a status of either:

* Current
* Active LTS
* Maintentance LTS

This is based on the node releases: https://nodejs.org/en/about/previous-releases and https://github.com/nodejs/Release

## Updated Warning Rule

Previously, Hardhat showed a version warning for any unsupported version:

1. All odd numbered major version numbers
4. All versions before the min version
5. All version after the max version

That rule has now been updated so that warnings are only shown for 1 and 2. Newer even versions don't display the unsupported versions warning message.

## PR Review Approach

Commit at a time.

### DIFF
--- a/.github/workflows/hardhat-core-ci.yml
+++ b/.github/workflows/hardhat-core-ci.yml
@@ -118,6 +118,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        # TODO: Currently there is no @types/node for Node 22,
+        # include it when it is available.
         node: [18, 20]
     steps:
       - uses: actions/checkout@v4
--- a/.changeset/shiny-wasps-invite.md
+++ b/.changeset/shiny-wasps-invite.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Add support for Node v22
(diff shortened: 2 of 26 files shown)


schaable (approved):

.github/workflows/hardhat-core-ci.yml
```diff
@@ -118,6 +118,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
+        # TODO: Currently there is no @types/node for Node 22,
+        # include it when it is available.
         node: [18, 20]
```
kanej, line 123: I believe the types for node v22 will only become available once it is in Active LTS

alcuadrado (approved):

Issue #5379, opened 2024-06-11
Complete the implementation and testing of global paramters


alcuadrado:
This issue is to track the global parameter's functionality implementation in core.

They are implemented [here](https://github.com/NomicFoundation/hardhat/blob/3a3de6b20ca9a566e6715e87c7081a07b5c60323/v-next/core/src/internal/global-parameters.ts), but there are some missing parts of its functionality, and it's not tested.

## Missing functionality
- [x] `buildGlobalParameterDefinition`, which is a function to create the config values used to define a new global parameter, should validate its options. 
- [x] `buildGlobalParameterMap`, which reduces those definitions into the actual map into the internal representation of the global params, needs to validate them again (see the comments).
- [x] `resolveGlobalArguments` should try to fetch every missing global parameter value from the env variables. See below.

### Getting global parameters from env variables

You can pass global parameters to hardhat as `--my-global-parameter value` in the CLI, or as an env variable with `HARDHAT_MY_GLOBAL_PARAMETER=value`. The CLI version has presedence over the env variables.

schaable:
@alcuadrado 

The `resolveGlobalArguments` function receives the `globalParametersMap`. I assume this is to validate that the global argument names do not clash with any global parameter names, correct?

> The CLI version has presedence over the env variables.

Should we throw an error if a global argument is passed through the CLI and is also defined as an environment variable?

PR #5390, opened 2024-06-13
buildGlobalParameterDefinition implementation & testing

schaable:

This should be easy to review by going through it commit by commit.

### DIFF
--- a/v-next/core/src/config.ts
+++ b/v-next/core/src/config.ts
@@ -1,3 +1,4 @@
+import type { ParameterTypeToValueType } from "./types/common.js";
 import type { ConfigurationVariable } from "./types/config.js";
 import type { GlobalParameter } from "./types/global-parameters.js";
 import type {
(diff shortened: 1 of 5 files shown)


v-next/core/src/config.ts
```diff
@@ -57,11 +58,11 @@ export function overrideTask(
...
  * Defines a global parameter.
  */
-export function globalParameter(options: {
+export function globalParameter<T extends ParameterType>(options: {
   name: string;
   description: string;
-  parameterType: ParameterType;
-  defaultValue: any;
+  parameterType?: T;
+  defaultValue: ParameterTypeToValueType<T>;
```
schaable, line 65: @alcuadrado I've added the type validation we use in the task parameters, I think it makes sense to have it here as well.

v-next/core/src/internal/global-parameters.ts
```diff
@@ -48,21 +58,47 @@ export function buildGlobalParameterMap(
...
+  name,
+  description,
+  parameterType,
+  defaultValue,
+}: {
   name: string;
   description: string;
-  parameterType: ParameterType;
-  defaultValue: any;
+  parameterType?: T;
```
schaable, line 69: @alcuadrado I think `parameterType` should be just `type` to be consistent with task parameters. wdyt? I'll change it on a follow up PR if you agree.

alcuadrado (approved):
LGTM. And agreed with both of your comments

v-next/core/src/internal/global-parameters.ts
schaable: Added an [issue](https://github.com/NomicFoundation/hardhat/issues/5415) to track it.

PR #5393, opened 2024-06-14
hook manager tests

### LINKED ISSUE #5231: Hook Manager testing
The hook manager is a key part of the new `./v-next/core`, but is implemented based on happy paths and without tests, that needs to be remedied.

## TODO

- [x] Determine further cleanup required
- [x] Add unit tests to cover core cases

kanej:
Add unit tests around the `HookManager` in `./v-next/core/src/internal/Hook-Manager.ts`.

It applies some small refactoring to the `HookManager` swapping in assertion instead of conditionals throwing errors, and replacing a plain `Error` with a specific `HardhatError`.

Resolves #5231.

## TODO

- [x] runHandlerChain
- [x] runSequentialHandlers
- [x] runParallelHandlers
- [x] unregisterHandlers

### DIFF
--- a/v-next/core/src/internal/hook-manager.ts
+++ b/v-next/core/src/internal/hook-manager.ts
@@ -12,6 +12,8 @@ import type {
 import type { HardhatPlugin } from "../types/plugins.js";
 import type { LastParameter, Return } from "../types/utils.js";
 
+import { assertHardhatInvariant } from "@nomicfoundation/hardhat-errors";
+
 export class HookManagerImplementation implements HookManager {
   readonly #pluginsInReverseOrder: HardhatPlugin[];
 
@@ -108,10 +110,10 @@ export class HookManagerImplementation implements HookManager {
 
     let handlerParams: Parameters<typeof defaultImplementation>;
     if (hookCategoryName !== "config") {
-      // TODO: assert that this.#context is not undefinded
-      if (this.#context === undefined) {
-        throw new Error(`Context must be set before running non-config hooks`);
-      }
+      assertHardhatInvariant(
+        this.#context !== undefined,
+        "Context must be set before running non-config hooks",
+      );
 
       handlerParams = [this.#context, ...params] as any;
     } else {
--- a/v-next/core/src/internal/hook-manager.ts
+++ b/v-next/core/src/internal/hook-manager.ts
@@ -182,10 +182,10 @@ export class HookManagerImplementation implements HookManager {
 
     let handlerParams: any;
     if (hookCategoryName !== "config") {
-      // TODO: assert that this.#context is not undefinded
-      if (this.#context === undefined) {
-        throw new Error(`Context must be set before running non-config hooks`);
-      }
+      assertHardhatInvariant(
+        this.#context !== undefined,
+        "Context must be set before running non-config hooks",
+      );
 
       handlerParams = [this.#context, ...params];
     } else {
--- a/v-next/core/src/internal/hook-manager.ts
+++ b/v-next/core/src/internal/hook-manager.ts
@@ -299,21 +299,17 @@ export class HookManagerImplementation implements HookManager {
 
     const factory = mod.default;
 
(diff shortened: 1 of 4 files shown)


v-next/core/test/internal/fixture-plugins/config-plugin.ts
```diff
@@ -0,0 +1,22 @@
+import type { HardhatUserConfig } from "../../../src/types/config.js";
```
ChristopherDedominici, line 1: Just to double check: this import and the following one are both coming from the "core" package.
They are imported in different ways on purpose? One via relative path and one via "@nomicfoundation/hardhat-core"

v-next/core/src/internal/hook-manager.ts
```diff
@@ -288,30 +293,31 @@ export class HookManagerImplementation implements HookManager {
     path: string,
   ): Promise<Partial<HardhatHooks[HookCategoryNameT]>> {
     if (!path.startsWith("file://")) {
-      throw new Error(
-        `Plugin ${pluginId} hook factory for ${hookCategoryName} is not a valid file:// URL: ${path}`,
+      throw new HardhatError(
+        HardhatError.ERRORS.HOOKS.INVALID_HOOK_FACTORY_PATH,
```
ChristopherDedominici, line 297: For my curiosity: why this is the only error coming from HardhatError.ERRORS?

v-next/core/test/internal/hook-manager.ts
```diff
@@ -0,0 +1,938 @@
...
+          "validateUserConfig",
+          [
+            {
+              plugins: [],
+              tasks: [],
+            },
+          ],
+        );
+
+        // Assert
```
ChristopherDedominici, line 870: I think this comment can be removed?

v-next/core/test/internal/fixture-plugins/config-plugin.ts
kanej: Good catch. I have opted to make both imports relative, rather than mixing.

v-next/core/src/internal/hook-manager.ts
kanej: That is the only error left in `HookManager`, the other errors have been converted to assertions. I limited the conversion to `HardhatError` to only those in the code under test.

v-next/core/test/internal/hook-manager.ts
```diff
@@ -0,0 +1,938 @@
...
+
+// This allows us to test the sequential return of handlers
+// Currently `hre.created` is the only hook that is returned sequentially
+// but it returns void. `testExample` takes and returns a string to
+// ease testing.
+declare module "../../src/types/hooks.js" {
+  interface HardhatRuntimeEnvironmentHooks {
+    testExample: (context: HookContext, input: string) => Promise<string>;
+  }
+}
```
alcuadrado, line 32: I wouldn't do this, as adding the declaration here is equivalent to adding it to the whole codebase. Maybe some alternative to test that they are run in a sequence could be pushing to some well known or global-ish array?

alcuadrado: after reading more, i get why my proposed alternative doesn't work. This is trickier than I originally understood.

Maybe we can do some casting within the test instead?

alcuadrado (approved):
I think this looks good, but I'm hesitant about merging it with the type augmentation issue I pointed out.

v-next/core/test/internal/hook-manager.ts
kanej: I have removed the type augmentation, and replaced it with a small amount of casting:

https://github.com/NomicFoundation/hardhat/pull/5393/commits/f9fa705a4e2fe81c724dead0764f19f0a1fc32cf

This strikes me as a fair balance between `any` usage and clarity; but I am happy to push in a different direction.

alcuadrado (approved):

Issue #5375, opened 2024-06-11
Add an Easier Way to Import Contract Types for hardhat-viem


beepidibop:
### Describe the feature

Currently, in order to import contract types with hardhat-viem users need to import `GetContractReturnType<ContractName$Type>` from two sources (`@nomicfoundation/hardhat-viem/types` and the artifact path for the contract), the amount of imports grow as well if you need multiple contracts since you need one for each artifact, or you can use the `ArtifactsMap` but it's a handful to type `GetContractReturnType<ArtifactsMap["ContractName"]>` each time.

Adding a `ContractTypesMap` that returns `GetContractReturnType<${ctd.typeName}["abi"]>` to `artifacts.d.ts` by modifying `generateArtifactsDefinition()` and `generateDuplicateArtifactsDefinition()` can solve this problem. Users will only need to do `import { ContractTypesMap } from "hardhat/types/artifacts";` and the contract type can be used like this

` const contract: ContractTypesMap["ContractName"];`

I have a [fork](https://github.com/NomicFoundation/hardhat/compare/main...beepidibop:hardhat:Add-ContractTypesMap) ready for this as well, I've run the tests and updated the snapshots. Let me know if anything else needs to be done before I submit a PR.

### Search terms

viem contract type

kanej:
Hey @beepidibop, we would be open to a PR along these lines. Feel free to submit it.

beepidibop:
Awesome, I've submitted the PR [here](https://github.com/NomicFoundation/hardhat/pull/5398#issue-2354819991).

PR #5398, opened 2024-06-15
Add contract types map to hardhat-viem

beepidibop:

---

Adding a `ContractTypesMap` that returns `GetContractReturnType<${ctd.typeName}["abi"]>` to `artifacts.d.ts` by modifying `generateArtifactsDefinition()` and `generateDuplicateArtifactsDefinition()` can improve UX. Users will only need to do `import { ContractTypesMap } from "hardhat/types";` and the contract type can be used like this

` const contract: ContractTypesMap["ContractName"];`

This PR was previously discussed [here](https://github.com/NomicFoundation/hardhat/issues/5375#issue-2346940178).


schaable (approved):
LGTM. Tested in a sample project by importing the map and verifying that the types for the `Lock` contract match what is returned from `deployContract` and `getContractAt`.

PR #5399, opened 2024-06-15
buildGlobalParametersMap implementation & testing

schaable:

This should be easy to review by going through it commit by commit.

Merge this after https://github.com/NomicFoundation/hardhat/pull/5390

### DIFF
--- a/v-next/core/src/index.ts
+++ b/v-next/core/src/index.ts
@@ -28,4 +28,4 @@ export async function createHardhatRuntimeEnvironment(
 }
 
 export { resolvePluginList } from "./internal/plugins/resolve-plugin-list.js";
-export { buildGlobalParameterMap } from "./internal/global-parameters.js";
+export { buildGlobalParametersMap } from "./internal/global-parameters.js";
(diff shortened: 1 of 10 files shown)


v-next/core/src/index.ts
```diff
@@ -28,4 +28,4 @@ export async function createHardhatRuntimeEnvironment(
 }
 
 export { resolvePluginList } from "./internal/plugins/resolve-plugin-list.js";
-export { buildGlobalParameterMap } from "./internal/global-parameters.js";
+export { buildGlobalParametersMap } from "./internal/global-parameters.js";
```
alcuadrado, line 31: @schaable  We already discussed changing this name in a future PR, but just FYI, this was correct. I It can be tricky for us, native Spanish speakers, but it's ParameterMap, not ParametsMap. You have to keep the first word in singular, even though the name represents a collection of those things. Pretty much the contrary to what you'd do in Spanish.

alcuadrado (approved):
LGTM, but let's not forget about the naming issue

schaable:
> LGTM, but let's not forget about the naming issue

I've added an [issue](https://github.com/NomicFoundation/hardhat/issues/5416) to track it.

PR #5400, opened 2024-06-17
Remove leftover `_eventAdapter` from `EdrProviderWrapper`

Xanewok:

---

Seems to be a leftover from #4747 and it looks to be unused now as we create a new instance of `EdrProviderEventAdapter` explicitly in `create()` and move it into lambda.

Feel free to disregard if that's not true and/or we want to keep it.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
@@ -182,7 +182,6 @@ export class EdrProviderWrapper
     private readonly _node: {
       _vm: MinimalEthereumJsVm;
     },
-    private readonly _eventAdapter: EdrProviderEventAdapter,
     private readonly _vmTraceDecoder: VmTraceDecoder,
     private readonly _rawTraceCallbacks: RawTraceCallbacks,
     // The common configuration for EthereumJS VM is not used by EDR, but tests expect it as part of the provider.
@@ -323,7 +322,6 @@ export class EdrProviderWrapper
     const wrapper = new EdrProviderWrapper(
       provider,
       minimalEthereumJsNode,
-      eventAdapter,
       vmTraceDecoder,
       rawTraceCallbacks,
       common,


kanej (approved):
LGTM.

fvictorio:
@Wodann do you remember if that's the reason you added this to `EdrProviderWrapper`?

Wodann:
> @Wodann do you remember if that's the reason you added this to `EdrProviderWrapper`?

There is a comment [here](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts#L227) explaining things. Maybe the garbage collector prevents destroying the `_eventAdapter` because it's part of a callback, even if not stored on the EDR provider. I'm not 100% sure though.

Xanewok:
In theory it should always be traced (in the GC sense) through the lambda and thus kept alive but I see now that the subscriber callback is moved to the Rust side, so it depends on the fact how it's implemented exactly.

From the cursory glance it seems that:
1. it's moved into `ProviderData` 
2. as `SubscriberCallback` with a `create_threadsafe_function` napi-rs call
3. which internally holds `Arc<ThreadsafeFunctionHandle>`
4. only that implements `Drop`, which calls `napi_release_threadsafe_function`

which leads me to believe that it should be reachable by v8 and thus kept alive; I'd be really surprised if the func argument to `napi_create_threadsafe_function` would have to be kept alive separately and not added to the 'alive' roots by napi/v8 itself.

Going a bit deeper into the node internals:
1. https://github.com/nodejs/node/blob/0062d5a07632c1333f35e509af002c3b2f81cf18/src/node_api.cc#L1315 creates a new `ThreadSafeFunction` class
2. https://github.com/nodejs/node/blob/0062d5a07632c1333f35e509af002c3b2f81cf18/src/node_api.cc#L228 During construction, it resets the persistent handle to the v8 function via the `v8::Persistent` class, which the docs state is alive for the engine's lifetime

The last part is a bit hand-wave-y as I don't want to get so deep into how exactly environment/resource management is implemented (there's a lot of ground and minutiae to cover!)

All in all, this looks safe to do. WDYT?

fvictorio:
Yes, overall it seems like a safe bet. I'm not sure how this would manifest if our assumption is wrong, but I think it would mean that filters would stop working "after a while". Worth keeping in mind in case we see that behavior in the future, although it does seem highly unlikely.

fvictorio (approved):

PR #5411, opened 2024-06-19
Streamline `ConsoleLogger` logic and state

Xanewok:

---

Part of https://github.com/NomicFoundation/edr/issues/246

The starting point for this PR was analyzing how we create the Provider and how different pieces interact with each other in JS and Rust<>JS in general around the stack traces logic and tests.

This removes some of the dead or redundant code and refreshes the console.log signature lookup table generation script. If we decide to move it to Rust, we won't have to call JS anymore and move buffers around just to decode and execute the console.log formatting logic.

After moving the auxiliary script to TS I re-ran it to verify that it outputs the same code (incl. the changes from this PR).


agostbiro:
Hey thanks for this PR @Xanewok ! I had a look, but I don't feel qualified to review it.

packages/hardhat-core/src/internal/hardhat-network/stack-traces/consoleLogger.ts
```diff
@@ -42,113 +42,61 @@ import {
...
-        if (log !== undefined) {
-          logs.push(log);
-        }
-
-        continue;
-      }
-
-      this._collectExecutionLogs(messageTrace, logs);
-    }
-  }
```
Wodann, line 116: Have these been moved to EDR already?

Wodann (commented):
The changes to `logger.ts` look good to me.

I can't comment on anything else, as I'm not familiar enough with Hardhat's coding standard to validate it.

fvictorio:
Checking my understanding here. This PR has two parts:

- The most important part is turning `ConsoleLogger` into a bag of static methods, because (I guess) that will make porting it easier. Is this correct?
- The other part is migrating `scripts/console-library-generator.js` to typescript. I guess you did this as part of checking how the code works, and it doesn't hurt to commit it.

If this is correct, then LGTM.

Xanewok:
That is correct:
1. The first part isolates the logic in `ConsoleLogger` - the class did not in fact depend on some initial state but rather just used the selectors that are static/const and have to be generated by the script.
  After the refactor we know that we only need to depend on a single function to correctly format the logs in the EDR and it will be easier to port and use it directly rather than having to roundtrip via FFI only to format the console logs.
2. The script that was originally used to generate the data used by the `ConsoleLogger` was discovered to be broken in the meantime, so I ported it to TS to ensure we catch any such regressions in the future (even if a short one, as we want to port it anyways) and made sure it still works as expected.

fvictorio:
Makes sense! Feel free to merge this if you want.

PR #5414, opened 2024-06-19
resolveGlobalArguments implementation & testing

schaable:

This should be easier to review by reviewing each commit individually.

Merge this after https://github.com/NomicFoundation/hardhat/pull/5399

### DIFF
--- a/v-next/core/src/internal/global-options.ts
+++ b/v-next/core/src/internal/global-options.ts
@@ -145,15 +145,21 @@ export function resolveGlobalOptions(
       -- GlobalOptions is empty for user extension, so we need to cast it to
       assign the value. */
       (userProvidedGlobalOptions as Record<string, string | undefined>)[name];
-    if (value === undefined) {
-      value = process.env[`HARDHAT_${camelToSnakeCase(name).toUpperCase()}`];
-    }
 
     let parsedValue: ParameterValue;
+    // if the value is provided in the user options, it's already parsed
+    // and it takes precedence over env vars
     if (value !== undefined) {
-      parsedValue = parseParameterValue(value, option.parameterType, name);
+      parsedValue = value;
     } else {
-      parsedValue = option.defaultValue;
+      value = process.env[`HARDHAT_${camelToSnakeCase(name).toUpperCase()}`];
+      if (value !== undefined) {
+        // if the value is provided via an env var, it needs to be parsed
+        parsedValue = parseParameterValue(value, option.parameterType, name);
+      } else {
+        // if the value is not provided by the user or env var, use the default
+        parsedValue = option.defaultValue;
+      }
     }
 
     /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
--- a/v-next/core/src/internal/global-options.ts
+++ b/v-next/core/src/internal/global-options.ts
@@ -156,7 +156,10 @@ export function resolveGlobalOptions(
       parsedValue = option.defaultValue;
     }
 
-    globalOptions[name] = parsedValue;
+    /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      -- GlobalOptions is empty for user extension, so we need to cast it to
+      assign the value. */
+    (globalOptions as Record<string, ParameterValue>)[name] = parsedValue;
   }
 
   return globalOptions;
--- a/v-next/core/src/internal/global-options.ts
+++ b/v-next/core/src/internal/global-options.ts
@@ -1,4 +1,7 @@
-import type { ParameterTypeToValueType } from "../types/common.js";
+import type {
+  ParameterTypeToValueType,
+  ParameterValue,
+} from "../types/common.js";
 import type {
   GlobalOptions,
   GlobalOption,
@@ -7,6 +10,7 @@ import type {
 import type { HardhatPlugin } from "../types/plugins.js";
 
 import { HardhatError } from "@nomicfoundation/hardhat-errors";
+import { camelToSnakeCase } from "@nomicfoundation/hardhat-utils/string";
 
 import { ParameterType } from "../types/common.js";
 
@@ -14,14 +18,16 @@ import {
   RESERVED_PARAMETER_NAMES,
   isParameterValueValid,
   isValidParamNameCasing,
+  parseParameterValue,
 } from "./parameters.js";
 
 /**
- * Builds a map of the global options, validating them.
+ * Builds a map of the global option definitions by going through all the
+ * plugins and validating the global options they define.
  *
  * Note: this function can be used before initializing the HRE, so the plugins
  * shouldn't be consider validated. Hence, we should validate the global
- * parameters.
+ * options.
  */
 export function buildGlobalOptionsMap(
   resolvedPlugins: HardhatPlugin[],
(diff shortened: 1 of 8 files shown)


v-next/core/src/internal/global-parameters.ts
```diff
@@ -104,13 +114,43 @@ export function buildGlobalParameterDefinition<T extends ParameterType>({
...
+ * are not provided. Only the arguments defined in the global parameters map
+ * are resolved.
+ *
+ * @param userProvidedGlobalArguments The arguments provided by the user. These
+ * take precedence over environment variables.
+ * @param globalParametersMap The map of global parameter definitions to
+ * validate the arguments.
+ */
 export function resolveGlobalArguments(
   userProvidedGlobalArguments: Partial<GlobalArguments>,
```
schaable, line 130: I believe the type for `userProvidedGlobalArguments` is incorrect, as the arg values are always strings and might differ from what the user specifies in the type. I suggest using `Record<string, string | undefined>` instead. wdyt, @alcuadrado?

v-next/core/src/types/global-parameters.ts
```diff
@@ -25,7 +29,7 @@ export interface GlobalParameter<T extends ParameterType = ParameterType> {
  * Runtime Environment.
  */
 // eslint-disable-next-line @typescript-eslint/no-empty-interface -- To be used through module augmentation
-export interface GlobalArguments {}
+export interface GlobalArguments extends Record<string, ParameterValue> {}
```
schaable, line 32: @alcuadrado I think this is fine, but please double check.

zoeyTM:
@schaable now that https://github.com/NomicFoundation/hardhat/pull/5417 has been merged, this PR will need rebasing and will need to also follow the new name scheme. i.e. `resolveGlobalArguments` becomes `resolveGlobalOptions`, etc.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -378,7 +378,7 @@ function parseDoubleDashArgs(
   cliArguments: string[],
   usedCliArguments: boolean[],
   optionsMap: Map<string, TaskOption | GlobalOption>,
-  argumentsMap: TaskArguments,
+  argumentsMap: TaskArguments | Partial<GlobalOptions>,
```
schaable, line 381: I've been out the last few days, so I might have missed some context. However, the naming here seems a bit odd now with the rename to "options".

schaable:
> @schaable now that #5417 has been merged, this PR will need rebasing and will need to also follow the new name scheme. i.e. `resolveGlobalArguments` becomes `resolveGlobalOptions`, etc.

done!

v-next/core/src/internal/global-options.ts
```diff
@@ -104,13 +114,50 @@ export function buildGlobalOptionDefinition<T extends ParameterType>({
...
+      -- GlobalOptions is empty for user extension, so we need to cast it to
+      assign the value. */
+      (userProvidedGlobalOptions as Record<string, string | undefined>)[name];
+    if (value === undefined) {
+      value = process.env[`HARDHAT_${camelToSnakeCase(name).toUpperCase()}`];
+    }
+
+    let parsedValue: ParameterValue;
+    if (value !== undefined) {
+      parsedValue = parseParameterValue(value, option.parameterType, name);
```
alcuadrado, line 154: I believe this is trying to parse the `userProvidedGlobalOptions`'s values.

alcuadrado: if those are provided, they should already be parsed

v-next/core/src/types/global-options.ts
```diff
@@ -25,7 +29,7 @@ export interface GlobalOption<T extends ParameterType = ParameterType> {
  * Runtime Environment.
  */
 // eslint-disable-next-line @typescript-eslint/no-empty-interface -- To be used through module augmentation
-export interface GlobalOptions {}
+export interface GlobalOptions extends Record<string, ParameterValue> {}
```
alcuadrado, line 32: I think this will be a bit annoying, as you can do `globalOptions.thisDoesntExist` and you wouldn't get a type error

alcuadrado: What motivated it? We may find an alternative

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: the rename is still incomplete, I added a few missing things in #5421

v-next/core/test/utils.ts
```diff
@@ -0,0 +1,30 @@
+import { afterEach } from "node:test";
+
+export function createTestEnvManager() {
```
alcuadrado, line 3: this is pretty nice

v-next/core/test/internal/global-options.ts
```diff
@@ -240,7 +242,202 @@ describe("Global Options", () => {
...
+            }),
+          ],
+        },
+      ]);
+
+      setEnvVar("HARDHAT_PARAM3", "5n");
+
+      const globalOptions = resolveGlobalOptions(
+        {
+          param1: "false",
```
alcuadrado, line 306: this is an example of a param that should already be parsed

alcuadrado: I think a good way to think about this is: first the cli parses the cli arguments, then the hre is initialized, using those arguments, which requires resolving like this.

v-next/core/src/internal/parameters.ts
```diff
@@ -54,3 +58,99 @@ const parameterTypeValidators: Record<
   [ParameterType.FLOAT]: (value): value is number => typeof value === "number",
   [ParameterType.FILE]: (value): value is string => typeof value === "string",
 };
+
+/**
+ * Parses a parameter value from a string to the corresponding type.
+ */
+// TODO: this code is duplicated in v-next/hardhat/src/internal/cli/main.ts
```
alcuadrado, line 65: I think exposing these things from core makes sense. We already expose functions like this. e.g. `resolvePluginList`.

alcuadrado (commented):
Left a few small-ish comments

v-next/core/src/types/global-options.ts
schaable: If I remember correctly, this was done to avoid an error when assigning the parsed value to the global options. As the interface is empty, this throws a TS error: `globalOptions[name] = parsedValue;`

> Element implicitly has an 'any' type because expression of type 'string' can't be used to index type 'GlobalOptions'.
>   No index signature with a parameter of type 'string' was found on type 'GlobalOptions'.ts(7053)

v-next/core/src/internal/parameters.ts
schaable: Agreed. I'll do it on a separate PR.

v-next/core/src/types/global-options.ts
schaable: Fixed in https://github.com/NomicFoundation/hardhat/pull/5414/commits/22abbd50240a9fcb3ac488c72b64b5dc0fce4ec8

v-next/core/src/internal/parameters.ts
schaable: Issue created: https://github.com/NomicFoundation/hardhat/issues/5431

alcuadrado (approved):

Issue #5358, opened 2024-06-10
`init` task


kanej:
The `init` command will provide a basis for demonstration and provide an extension point when we come to adding compilation and js/ts testing.

The `init` command will also stress our design of user interruptions.

```shell
> hardhat init
888    888                      888 888               888
888    888                      888 888               888
888    888                      888 888               888
8888888888  8888b.  888d888 .d88888 88888b.   8888b.  888888
888    888     "88b 888P"  d88" 888 888 "88b     "88b 888
888    888 .d888888 888    888  888 888  888 .d888888 888
888    888 888  888 888    Y88b 888 888  888 888  888 Y88b.
888    888 "Y888888 888     "Y88888 888  888 "Y888888  "Y888

👷 Welcome to Hardhat v3.0.0 👷‍

? What do you want to do? …
❯ Create an empty hardhat.config.ts
  Quit
```

Only the `Create an empty hardhat.config.ts` and `Quit` options will be select-able initially.

The `Create an empty hardhat.config.ts` will create the minimal `hardhat.config.ts`:

```ts
export default {
  solidity: "0.8.24",
}
```

If there is already a hardhat.config.{js,ts} , then the init task should error:

```sh
Error HHXXX: You are trying to initialize a project inside an existing Hardhat project. The path to the project's configuration file is:  /Users/fred/projects/nomiclabs/temp-sample/hardhat.config.js.

For more info go to https://hardhat.org/HHXXX or run Hardhat with --show-stack-traces
```

kanej:
> Small clarification, this isn't technically a task

What term should we use for these, "special tasks"?

PR #5425, opened 2024-06-24
Create base logic for the 'init' task "

ChristopherDedominici:
[Link to task description](https://github.com/NomicFoundation/hardhat/issues/5358)

PR GOAL
Implement:
- the creation of an empty hardhat.config.ts
-  the "quit" option

CODE CHANGES
- code to implement the base functionalities of the init command
- modified the "parseHardhatSpecialArguments" tests so that the result object is checked altogether. This ensures that if additional properties are added in the future, the test will fail, forcing the testing of the new properties. Also, the init command tests have been added
- added a "todo" test to check that the latest sol version is used in the hardhat.config.ts file. Currently, V3 has no function to check this. [See here](https://github.com/NomicFoundation/hardhat/pull/5425/files#r1651022486)
- tests about the logic are ONLY done using the ENV variables (if specific ENV variables are set, specific logic will be triggered). I wanted to avoid mocking the user input (please provide feedback if you do not agree with this approach). [See here](https://github.com/NomicFoundation/hardhat/pull/5425/files#r1651024511)
Also, for these tests I created several fixture-projects


v-next/hardhat/test/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,8 @@
+import { describe, it } from "node:test";
+
+describe("project-creation", function () {
+  it.todo(
```
ChristopherDedominici, line 4: added a "todo" test to check that the latest sol version is used in the hardhat.config.ts file. Currently, V3 has no function to check this

v-next/hardhat/src/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,238 @@
...
+    chalk.cyan(
+      `👷 Welcome to ${HARDHAT_NAME} v${await getHardhatVersion()} 👷\n`,
+    ),
+  );
+}
+
+async function getAction(): Promise<Action> {
+  // If a matching ENV variable is passed, we do not prompt the user
+  if (
+    process.env.HARDHAT_CREATE_EMPTY_TYPESCRIPT_HARDHAT_CONFIG !== undefined
```
ChristopherDedominici, line 137: The tests are using ENV variables to trigger the logic

v-next/hardhat/src/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,238 @@
...
+      }
+    }
+
+    throw new HardhatError(HardhatError.ERRORS.GENERAL.UNSUPPORTED_OPERATION, {
+      operation: `Responding with "${actionResponse.action}" to the project initialization wizard`,
+    });
+  } catch (e) {
+    if (e === "") {
+      return Action.QUIT;
+    }
```
schaable, line 175: Which scenario does this cover? Can we add a default value to the prompt to avoid it?

schaable (commented):
Left some comments. Sorry if it's too much, I got carried away 😅 I'll check the tests tomorrow

alcuadrado (commented):
Test

v-next/hardhat/src/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,238 @@
...
+  readJsonFile,
+  writeJsonFile,
+  writeUtf8File,
+} from "@nomicfoundation/hardhat-utils/fs";
+import chalk from "chalk";
+
+import { getHardhatVersion } from "../../utils/package.js";
+
+const HARDHAT_NAME = "Hardhat";
+const HARDHAT_PACKAGE_NAME = "hardhat";
```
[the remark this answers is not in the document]
ChristopherDedominici: A const file in the same folder like in [this commit](https://github.com/NomicFoundation/hardhat/pull/5425/commits/c7d84094e52470c986ccf50a9419263519ad8302)?

v-next/hardhat/src/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,238 @@
...
+
+// TODO: test that is testing the last version of solidity
+export const EMPTY_HARDHAT_CONFIG = `/** import type { HardhatUserConfig } from "@nomicfoundation/hardhat-core/config"; */
+
+export default {
+  solidity: "0.8.24",
+};
```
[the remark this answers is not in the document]
ChristopherDedominici: Something like [this](https://github.com/NomicFoundation/hardhat/pull/5425/commits/6f99e6d08438f4ef58555b862ce2c30cc147b8a0)?

v-next/hardhat/src/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,238 @@
...
+      return Action.QUIT;
+    }
+
+    ensureError(e);
+    throw e;
+  }
+}
+
+async function createPackageJson() {
+  await writeJsonFile("package.json", {
```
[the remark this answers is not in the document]
ChristopherDedominici: Good question. This is code will only create a  a basic  starting point for Hardhat. In V2 we were doing the same. @alcuadrado already took a look at the design dock, but I think it's worth to double check. @alcuadrado what do you think?

v-next/hardhat/src/internal/cli/init/project-creation.ts
ChristopherDedominici: When the user interrupts the cli input request (e.g.: ctrl+c) an error is thrown. We do not want to throw the error, so we just quit

v-next/hardhat/src/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,238 @@
...
+    chalk.blue("888    888 .d888888 888    888  888 888  888 .d888888 888"),
+  );
+  console.log(
+    chalk.blue("888    888 888  888 888    Y88b 888 888  888 888  888 Y88b."),
+  );
+  console.log(
+    chalk.blue('888    888 "Y888888 888     "Y88888 888  888 "Y888888  "Y888'),
+  );
+  console.log("");
+}
```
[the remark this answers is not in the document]
ChristopherDedominici: Done [here ](https://github.com/NomicFoundation/hardhat/pull/5425/commits/500b512678c434745821c2061c2a14570b1fc01c)with other smaller improvements

v-next/hardhat/src/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    throw new HardhatError(HardhatError.ERRORS.GENERAL.UNSUPPORTED_OPERATION, {
+      operation: `Responding with "${actionResponse.action}" to the project initialization wizard`,
+    });
+  } catch (e) {
+    if (e === "") {
+      return Action.QUIT;
+    }
+
+    ensureError(e);
```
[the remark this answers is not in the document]
ChristopherDedominici: Done [here ](https://github.com/NomicFoundation/hardhat/pull/5425/commits/500b512678c434745821c2061c2a14570b1fc01c)with other smaller improvements

v-next/hardhat/src/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,238 @@
...
+    await createPackageJson();
+  }
+
+  if (action === Action.CREATE_EMPTY_TYPESCRIPT_HARDHAT_CONFIG) {
+    return createEmptyTypescriptHardhatConfig();
+  }
+}
+
+async function getProjectPackageJson(): Promise<PackageJson | undefined> {
+  const pathToJsonPackage = path.join(process.cwd(), "package.json");
```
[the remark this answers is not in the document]
ChristopherDedominici: Done [here ](https://github.com/NomicFoundation/hardhat/pull/5425/commits/500b512678c434745821c2061c2a14570b1fc01c)with other smaller improvements

v-next/hardhat/src/internal/cli/init/project-creation.ts
```diff
@@ -0,0 +1,238 @@
...
+    });
+
+    console.log("");
+    console.log(cmd.join(" "));
+    console.log("");
+  }
+
+  console.log();
+
+  showReminderMessages();
```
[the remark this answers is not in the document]
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/5425/commits/d5b388f50f3a8ff4b2a18d2d13ffba04126dae79)

v-next/hardhat/src/internal/cli/init/project-creation.ts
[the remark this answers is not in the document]
schaable: I have the feeling that we'll use these outside the init command. But it's good for now!

v-next/hardhat/src/internal/cli/init/project-creation.ts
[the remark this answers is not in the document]
schaable: Hmm, no, that's not what I meant. You moved the logic to a different file, but it's still a variable with the Hardhat config code within a string. I meant having an actual file with the code, so we can easily update it, ensure it's well formatted, and maintain it more efficiently. Then you can read that file here and copy it.

v-next/hardhat/src/internal/cli/init/project-creation.ts
schaable: Interesting. It's a shame enquirer doesn't provide a more elegant way to handle this. I've seen this pattern a lot in v2. Maybe in the future, we can use a wrapper to abstract it.

v-next/hardhat/src/internal/cli/init/project-creation.ts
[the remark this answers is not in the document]
ChristopherDedominici: I see what you mean. The only downside with the suggested approach is that the solidity version will have to be updated manually. While with the current approach the version update will happen automatically. 
Since it's a very short example file I'd keep this approach, wdyt?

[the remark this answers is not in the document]
schaable: That's a good point. However, we'll need to manually update the Solidity version for the JS/TS sample projects anyway, right? This makes me think that we should export a constant with the current version from Hardhat and use it there. But that's something to discuss. I'm good if you want to leave it the way it is for now.

v-next/hardhat/src/internal/cli/init/project-structure.ts
```diff
@@ -0,0 +1,55 @@
...
+    return cjsConfigPath;
+  }
+
+  const jsConfigPath = await findUp(JS_CONFIG_FILENAME);
+  if (jsConfigPath !== null) {
+    return jsConfigPath;
+  }
+
+  throw new HardhatError(HardhatError.ERRORS.GENERAL.NOT_INSIDE_PROJECT);
+}
```
alcuadrado, line 43: We have `findClosestHardhatConfig` that replaces both of these functions.

v-next/hardhat/src/internal/cli/init/project-structure.ts
```diff
@@ -0,0 +1,55 @@
...
+  const packageRoot = await getPackageRoot();
+  const gitIgnorePath = path.join(packageRoot, "recommended-gitignore.txt");
+
+  return readUtf8File(gitIgnorePath);
+}
+
+async function getPackageRoot(): Promise<string> {
+  const packageJsonPath = await findClosestPackageJson(import.meta.url);
+  return path.dirname(packageJsonPath);
+}
```
alcuadrado, line 55: I think it'd be better having the gitignore in a constant

alcuadrado: The original intention to have it as a file was that it was easy to find it in the root of the repo, but that's no longer the case, as we use a monorepo.

v-next/hardhat/src/internal/cli/init/project-creation.ts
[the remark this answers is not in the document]
alcuadrado: Let's defer that decision. This is the minimal package.json that we need.

v-next/hardhat/src/internal/cli/init/init.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  // Many terminal emulators in windows fail to run the createProject()
+  // workflow, and don't present themselves as TTYs. If we are in this
+  // situation we throw a special error instructing the user to use WSL or
+  // powershell to initialize the project.
+  if (process.platform === "win32") {
+    throw new HardhatError(
+      HardhatError.ERRORS.GENERAL.NOT_INSIDE_PROJECT_ON_WINDOWS,
+    );
```
alcuadrado, line 31: I don't think this is the right error to be throwing here. Still, we don't have a good answer to the problem described in the comment. I'll create an issue about this.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -160,10 +164,17 @@ export async function parseHardhatSpecialArguments(
   let showStackTraces: boolean = isCi();
   let help: boolean = false;
   let version: boolean = false;
+  let init: boolean = false;
 
   for (let i = 0; i < cliArguments.length; i++) {
     const arg = cliArguments[i];
 
+    if (arg === "init") {
```
alcuadrado, line 172: Let's switch to `--init` instead. `init` would be too ambiguous for this naive parsing logic.

v-next/hardhat/src/internal/cli/init/project-structure.ts
alcuadrado: Actually, we aren't using these functions, are we? If so, let's remove them. 

Note that combined with the other comment, that means removing this file entirely.

alcuadrado (changes requested):
Looks good. I only left a few comments

v-next/hardhat/src/internal/cli/init/project-structure.ts
ChristopherDedominici: But doesn't 'findClosestHardhatConfig' only search for the hardhat.config.ts file? Here we also have .js, .cjs and .cts files.
Also, 'findClosestHardhatConfig' is not currently exported from hh-utils. Should we export it and maybe change the logic to handle all the possible file extensions?

@schaable looping you in because of hh-utils. Wdyt?

v-next/hardhat/src/internal/cli/main.ts
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/5425/commits/3a04f738ef6c6e8dafd0f1141459824172b058d2)

v-next/hardhat/src/internal/cli/init/init.ts
ChristopherDedominici: So for now we keep it like this?

v-next/hardhat/src/internal/cli/init/project-structure.ts
ChristopherDedominici: I removed these 2 functions [here](https://github.com/NomicFoundation/hardhat/pull/5425/commits/a81b6fda5b98beb0e03fa5d2adfd48d0c8fe3828). For the other comment, I replied with a follow up question

v-next/hardhat/src/internal/cli/init/project-structure.ts
schaable: I wouldn't add it to hh-utils because the package doesn't really know about hardhat; all the utils are meant to be generic.

v-next/hardhat/src/internal/cli/init/init.ts
alcuadrado: yep

v-next/hardhat/src/internal/cli/init/project-structure.ts
alcuadrado: @ChristopherDedominici I think that's actually a strong point for reusing it. If we recreate this logic here, it can easily be out of sync with the other implementation, which is the one used to load the config.

Regarding the actual extensions:

- `.js` is supported there.
- `.cjs` and `.cts` don't seem as important in v3, as it's designed for ESM projects. In v2 the config required CJS because it was imported by hardhat v2, which was cjs.

ChristopherDedominici: As agreed offline on the approach. Implementation done in [this commit](https://github.com/NomicFoundation/hardhat/pull/5425/commits/63daf1f0286aaa0c25a8f3599deead2ee414020f)

alcuadrado (approved):

alcuadrado:
Closing and opening again as github is not working properly here

PR #5429, opened 2024-06-25
Clean up the VMTracer and trace events interface in TS

Xanewok:

---

Rather than defining the callbacks at the provider instantiation time for the entire lifetime of the provider, this allows for a more granular control of injecting the trace event listener (VMTracer), thus simplifying how the information flows. In the test logic, we no longer tightly couple two classes but have to keep them side by side and we can stop observing the EVM if needed, without paying the runtime cost.

While technically it's a standalone improvement, the https://github.com/NomicFoundation/edr/pull/531 can nicely build on top if it as it will also allow us to conveniently observe the trace whenever needed all at once in Rust (with the Rust port of `VmTracer` in that PR); it also removes some dead code.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
@@ -469,11 +469,11 @@ export class EdrProviderWrapper
   }
 
   /**
-   * Injects a `VMTracer` that observes EVM throughout requests.
+   * Sets a `VMTracer` that observes EVM throughout requests.
    *
    * Used for internal stack traces integration tests.
    */
-  public injectVmTracer(vmTracer?: VMTracer) {
+  public setVmTracer(vmTracer?: VMTracer) {
     this._vmTracer = vmTracer;
   }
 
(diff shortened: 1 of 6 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -484,6 +468,15 @@ export class EdrProviderWrapper
     }
   }
 
+  /**
+   * Injects a `VMTracer` that observes EVM throughout requests.
+   *
+   * Used for internal stack traces integration tests.
+   */
+  public injectVmTracer(vmTracer?: VMTracer) {
```
Wodann, line 476: Nit: I'm not a fan of the use of the word "inject" here. It seems to be used both to set and to reset (to `undefined`), so I'd just keep it straightforward with:
```suggestion
  public setVmTracer(vmTracer?: VMTracer) {
```

Wodann (approved):
LGTM!

One nitpick

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
Xanewok: Sure, makes sense - I didn't want to use `set` myself because I didn't want to read it like it's a regular a config setting but rather a special mode, however, I understand the set/unset argument here :+1: 

Changed to `setVmTracer` in https://github.com/NomicFoundation/hardhat/pull/5429/commits/2fe2cff7863bd69c22e2b7669fdf0a88f5c3d77c

Issue #5359, opened 2024-06-10
`run` task


kanej:
The Hardhat run task enables js/ts scripts to be run in a node environment and leverage a HardhatRuntimeEnvironment generated based on the given hardhat.config.{js,ts}:

```sh
# js
hardhat run ./path/to/script.js

# ts
hardhat --config ./Alternate.hardhat.config.ts run ./path/to/script.ts
```

In both JavaScript and Typescript the HardhatRuntimeEnvironment needs to be imported:

JavaScript:
```js
// ./path/to/script.js

const hre = require("hardhat");

console.log(hre.config.solidity);
```

Typescript:
```ts
// ./path/to/script.ts

import hre from "hardhat";

console.log(hre.config.solidity);
```

The `HardhatRuntimeEnvironment` that is initialized will by default be based on the closest `hardhat.config.{js,ts}`, where closest mean looking in the current working directory, then moving to the parent directory recursively.

The `hardhat.config.{js,ts}` that is used can be overridden on the command line with the `--config` flag.

A Typescript script will be executed with `tsx`.

alcuadrado:
> Hardhat can also be used from plain node scripts, passing a config programmatically or relying on implicit file loading.

A small clarification. This can also be done with any script. These other lower-level apis are just for advanced/non-standard things.

> `readHardhatConfig`

I think exporting a function that finds the config file is also helpful, and may even be a better alternative to `readHardhatConfig`.

kanej:
> > Hardhat can also be used from plain node scripts, passing a config programmatically or relying on implicit file loading.
> 
> A small clarification. This can also be done with any script. These other lower-level apis are just for advanced/non-standard things.
> 
> > `readHardhatConfig`
> 
> I think exporting a function that finds the config file is also helpful, and may even be a better alternative to `readHardhatConfig`.

Can you provide example code, I am not sure I have followed what your preference is here?

alcuadrado:
Sure.

You can run a hardhat script with `node script.js`. As in v3 you have to import hardhat, that import will initialize the HRE.

```
/// script.js
import {config} from "hardhat";

console.log(config);
```

Or, if it was about the second part:

```js
import { createHardhatRuntimeEnvironment, findHardhatConfigFile } from "hardhat/hre"

const configPath = await findHardhatConfigFile()  
const {default: config} = await import(configPath); 
const hre = await createHardhatRuntimeEnvironment(config)
console.log(hre.config.solidity)
```

PR #5432, opened 2024-06-26
Add `run` task

zoeyTM:
fixes https://github.com/NomicFoundation/hardhat/issues/5359

- adds the `run` task with tests
- exports the `resolveHardhatConfigPath` function in the public API for use in scripts
- moves all tests from `v-next/hardhat/test` into `v-next/hardhat/test/src`
  - the reason for this change was because running the node test runner on `test/**/*.ts` was causing meta files such as `test/fixture-projects` and `test/helpers` to be included in the "tested" files. This threw an error with new fixture projects testing the `run` task as they were causing the HRE to be imported and initialized in unintended places. With this change, I also updated the npm scripts for running tests to `test/src/**/*.ts`


v-next/hardhat/test/src/internal/builtin-plugins/run/runScriptWIthHardhat.ts
```diff
@@ -0,0 +1,110 @@
...
+  let hre: HardhatRuntimeEnvironment;
+
+  before(async function () {
+    hre = await createHardhatRuntimeEnvironment({});
+  });
+
+  it("should throw if script is not a string", async function () {
+    await assert.rejects(async () => {
+      await runScriptWithHardhat({ script: 123, noCompile: false }, hre);
+    }, /An internal invariant was violated: Expected script to be a string/);
```
schaable, line 22: very minor, but for future reference `assert.rejects` can take a promise, so this can be simplified a bit:

```suggestion
    await assert.rejects(
      runScriptWithHardhat({ script: 123, noCompile: false }, hre),
      /An internal invariant was violated: Expected script to be a string/);
```

schaable (approved):

v-next/hardhat/src/internal/builtin-plugins/run/index.ts
```diff
@@ -0,0 +1,23 @@
...
+      .addPositionalParameter({
+        name: "script",
+        description: "A js or ts file to be run within hardhat's environment",
+        type: ParameterType.STRING,
+      })
+      .addFlag({
+        name: "noCompile",
+        description: "Don't compile before running this task",
+      })
+      .setAction(runScriptWithHardhat)
```
alcuadrado, line 20: I think we should do this instead

```suggestion
      .setAction(import.meta.resolve('./runScriptWithHardhat.js'))
```

and export the function as default within that file.

This way the implementation of the task is only loaded if it's run.

v-next/hardhat/src/internal/builtin-plugins/run/runScriptWithHardhat.ts
```diff
@@ -0,0 +1,57 @@
...
+  HardhatError,
+  assertHardhatInvariant,
+} from "@nomicfoundation/hardhat-errors";
+import { exists } from "@nomicfoundation/hardhat-utils/fs";
+
+export const runScriptWithHardhat: NewTaskActionFunction = async (
+  { script, noCompile },
+  _hre,
+) => {
+  assertHardhatInvariant(
```
alcuadrado, line 16: This will improve with #5434

v-next/hardhat/package.json
```diff
@@ -35,8 +35,8 @@
     "lint:fix": "pnpm prettier --write && pnpm eslint --fix",
     "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
-    "test": "node --import tsx/esm --test --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/**/*.ts\"",
-    "test:only": "node --import tsx/esm --test --test-only --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/**/*.ts\"",
+    "test": "node --import tsx/esm --test --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/src/**/*.ts\"",
+    "test:only": "node --import tsx/esm --test --test-only --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/src/**/*.ts\"",
```
alcuadrado, line 39: I don't understand this change. Can you shed some light here? If we need this, I'll modify the github actions check that validates the npm scripts.

alcuadrado (changes requested):
Just left a few comments, but this looks good

v-next/hardhat/package.json
zoeyTM: Yes, so if you look at the new fixture project added with this PR, you'll see there are script files that import the HRE. The HRE runs initialization steps (specifically config resolution steps) when it gets imported. The current test script uses the glob `"test/**/*.ts"`, which makes node's test runner load all typescript files in the `test` directory. This was fine when we only had tests in the `test` directory, but now it causes both the `fixture-projects` as well as `helpers` to also get loaded by the test runner. This has actually been happening for a while, but is only now becoming a problem because when the test runner loads the new fixture project, it causes the HRE instances to be immediately initialized outside the scope of a test that might be using them. This causes an error since the config loading process expectedly can't find a hardhat config file (since they're being unintentionally run outside the scope of the `useFixtureProject` helpers)

zoeyTM: This change moves the actual test files into `test/src` so we can specifically pass that glob to the node test runner. This means we can put anything else that we want/need to inside `test/*` and not worry about it being loaded unintentionally by the node test runner.

v-next/hardhat/test/src/internal/builtin-plugins/run/runScriptWIthHardhat.ts
zoeyTM: I blame copilot :)

alcuadrado (approved):

PR #5441, opened 2024-06-27
Remove package.json from `hardhat`'s tsconfig.json

alcuadrado:
This PR removes the `package.json` from `hardhat`'s `tsconfig.json`. 

The main reason this is important is that having the pacakge.json as part of the files that `tsc` process means that it's copied over to `dist/`, bringing a few problems `tsc` when importing `hardhat` from other packages. With the `dist/package.json`, it incorrectly thinks that `dist/` is the root of the package.

As this was probably added to support `example.config.ts` in `hardhat/`, I removed that and moved its contents to `example-project/hardhat.config.ts`.


zoeyTM (approved):

alcuadrado:
While testing this @schaable noticed that the example-project was actually broken, as the type extensions of the builtin plugins weren't loaded. I fixed that too.

schaable (approved):
LGTM

PR #5445, opened 2024-06-27
Don't check dependencies of private packages

alcuadrado:
This PR fixes the v-next workflow to enforce consistent dependency versions across packages. It now doesn't take private packages into account.


kanej (approved):
I am going to merge this to test.

Issue #5418, opened 2024-06-20
Unify `RESERVED_PARAMETER_NAMES` and `GLOBAL_OPTIONS`


alcuadrado:
We have two ways to define global builtin options. I think we should unify them, so that most of hardhat can treat the same as user-defined global options.

The one exception would be that some of these built-in global options need to be parsed differently, but it's better to contain that in the parser.

See my [comment](https://github.com/NomicFoundation/hardhat/pull/5409#discussion_r1646657511) here: #5409

schaable:
List of changes:

- [ ] `RESERVED_PARAMETER_NAMES`: update this to be a list of reserved words instead of the current list of built-in options. This list is intended to reserve specific words for future use. The built-in options are already validated by `TaskManager.#validateClashesWithGlobalOptions`.

- [ ] `GLOBAL_OPTIONS`: change this to be an actual array of options, including all their properties, and use `globalOption` to construct them. Also, rename the variable from `GLOBAL_OPTIONS` to `BUILTIN_OPTIONS`, as it only includes the Hardhat built-in options. The logic in `v-next/hardhat/src/internal/cli/helpers/utils.ts` should be updated to reflect this new format.

- [ ] Pass the user-defined global options to `getGlobalHelpString` and print them along with the built-in options.

PR #5453, opened 2024-06-28
update vnext to typescript 5.5

kanej:
Update the `v-next` packages to use typescript v5.5 and [enable isolated declarations](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5/#isolated-declarations).

The type determination has been mechanical based on vscode - I have not created any new types.

### DIFF
--- a/v-next/example-project/hardhat.config.ts
+++ b/v-next/example-project/hardhat.config.ts
@@ -95,7 +95,7 @@ const greeting = task("hello", "Prints a greeting")
   })
   .build();
 
-export default {
+const config: HardhatUserConfig = {
   tasks: [
     exampleTaskOverride,
     testTask,
@@ -130,4 +130,6 @@ export default {
     },
   ],
   privateKey: configVariable("privateKey"),
-} satisfies HardhatUserConfig;
+};
+
+export default config;
--- a/v-next/hardhat-utils/src/fs.ts
+++ b/v-next/hardhat-utils/src/fs.ts
@@ -343,7 +343,7 @@ export async function mkdir(absolutePath: string): Promise<void> {
  * Alias for `mkdir`.
  * @see mkdir
  */
-export const ensureDir = mkdir;
+export const ensureDir: (absolutePath: string) => Promise<void> = mkdir;
 
 /**
  * Retrieves the last change time of a file or directory's properties.
--- a/v-next/hardhat-utils/src/fs.ts
+++ b/v-next/hardhat-utils/src/fs.ts
@@ -343,7 +343,7 @@ export async function mkdir(absolutePath: string): Promise<void> {
  * Alias for `mkdir`.
  * @see mkdir
  */
-export const ensureDir: (absolutePath: string) => Promise<void> = mkdir;
+export const ensureDir: typeof mkdir = mkdir;
 
 /**
  * Retrieves the last change time of a file or directory's properties.
--- a/v-next/hardhat-zod-utils/src/index.ts
+++ b/v-next/hardhat-zod-utils/src/index.ts
@@ -7,7 +7,22 @@ import { z } from "zod";
 /**
  * A Zod type to validate Hardhat's ConfigurationVariable objects.
  */
-export const configurationVariableType = z.object({
+export const configurationVariableType: z.ZodObject<
+  {
+    _type: z.ZodLiteral<"ConfigurationVariable">;
+    name: z.ZodString;
+  },
+  "strip",
+  z.ZodTypeAny,
+  {
+    _type: "ConfigurationVariable";
+    name: string;
+  },
+  {
+    _type: "ConfigurationVariable";
+    name: string;
+  }
+> = z.object({
   _type: z.literal("ConfigurationVariable"),
   name: z.string(),
 });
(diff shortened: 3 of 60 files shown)


v-next/example-project/hardhat.config.ts
```diff
@@ -95,7 +95,7 @@ const greeting = task("hello", "Prints a greeting")
   })
   .build();
 
-export default {
```
schaable, line 98: I believe there was a reason why we used `satisfies` here, but I can't remember why. @alcuadrado do you remember?

v-next/hardhat-utils/src/fs.ts
```diff
@@ -343,7 +343,7 @@ export async function mkdir(absolutePath: string): Promise<void> {
  * Alias for `mkdir`.
  * @see mkdir
  */
-export const ensureDir = mkdir;
+export const ensureDir: (absolutePath: string) => Promise<void> = mkdir;
```
schaable, line 346: can we make this `typeof mkdir` instead?

v-next/hardhat-zod-utils/src/index.ts
```diff
@@ -7,18 +7,50 @@ import { z } from "zod";
 /**
  * A Zod type to validate Hardhat's ConfigurationVariable objects.
  */
-export const configurationVariableType = z.object({
+export const configurationVariableType: z.ZodObject<
```
schaable, line 10: hmm, having to define these types manually looks annoying. Maybe there's a better way, but I wouldn't worry until we work on the zod types.

v-next/hardhat/src/internal/cli/helpers/utils.ts
```diff
@@ -1,7 +1,10 @@
 import type { ParameterType } from "@ignored/hardhat-vnext-core/config";
 import type { Task } from "@ignored/hardhat-vnext-core/types/tasks";
 
-export const GLOBAL_OPTIONS = [
+export const GLOBAL_OPTIONS: Array<{
```
schaable, line 4: how about `Pick<GlobalOption, "name" | "description">`

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -35,7 +35,10 @@ import { getGlobalHelpString } from "./helpers/getGlobalHelpString.js";
 import { getHelpString } from "./helpers/getHelpString.js";
 import { printVersionMessage } from "./version.js";
 
-export async function main(cliArguments: string[], print = console.log) {
+export async function main(
+  cliArguments: string[],
+  print: (message?: any, ...optionalParams: any[]) => void = console.log,
```
schaable, line 40: I'll also need a similar typing for `print` in the error handler. Can we create a type for it?

v-next/hardhat/src/internal/helpers/config-loading.ts
```diff
@@ -21,7 +21,9 @@ export async function findClosestHardhatConfig(): Promise<string> {
   throw new HardhatError(HardhatError.ERRORS.GENERAL.NO_CONFIG_FILE_FOUND);
 }
 
-export async function importUserConfig(configPath: string) {
+export async function importUserConfig(
+  configPath: string,
+): Promise<Record<string | symbol, unknown>> {
```
schaable, line 26: wouldn't this be `HardhatUserConfig`?

schaable (commented):
Looks good! Left a few comments.

v-next/example-project/hardhat.config.ts
alcuadrado: I think this is the right move. We were using `satisfies` instead of `as`, which is much much better, but `: T` is equivalent to `satisfies` in this case.

v-next/hardhat-utils/src/fs.ts
alcuadrado: I believe that would work. I have a slight preference for the explicit version tbh.

v-next/hardhat-zod-utils/src/index.ts
alcuadrado: agreed

v-next/hardhat/src/internal/cli/helpers/utils.ts
alcuadrado: Let's cover this in #5418

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: I think typing it as `(message:string) => void` if possible would be better.

v-next/hardhat/src/internal/helpers/config-loading.ts
alcuadrado: Good catch. Yes.

v-next/hardhat-utils/src/fs.ts
alcuadrado: Addressed in d3cb5d9e3

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: Addressed in 01c2e6f0b

v-next/hardhat/src/internal/helpers/config-loading.ts
alcuadrado: Addressed in bc0d98b94

schaable (approved):

PR #5455, opened 2024-06-28
Error printing for Hardhat v3

schaable:
Errors are formatted and printed based on their category. Errors may belong to one of these categories:

 **1. Hardhat errors:** these are thrown from `core` or `hardhat` packages.
 **2. Hardhat plugin errors:** these are thrown from official Hardhat plugins, such as `hardhat-ignition`. Technically, this category would also include built-in plugins like `run`, but since we’re not adding the `pluginId` property to their descriptors, they fall into category 1 (Hardhat errors).
 **3. Hardhat community plugin errors:** these are thrown from community-developed plugins that use the `HardhatPluginError`, such as `hardhat-deploy`.
 **4. Other errors:** this category includes any other errors that don't fall into the above categories. This includes instances of `Error`, as well as objects or strings.
 
 Doc: https://www.notion.so/nomicfoundation/Design-doc-for-error-messages-2d49562c43de497d8241d8b024fe83b2#851e84f778c04437a5953c5f98a0b6d6

**Screenshots:**
(ignore the _Time to run the task.._ messages)

**Hardhat errors:**
hardhat-core

**Hardhat plugin errors (added `pluginId` to the `run` descriptor, as we don't have any):**
hardhat-plugin

**Hardhat community plugin errors:**
community-plugin

**Other errors (`Error` instance and `object`):**
unexpected-error
unknown-error


v-next/hardhat-errors/src/errors.ts
```diff
@@ -76,7 +80,7 @@ export class HardhatError<
           );
 
     super(
-      prefix + formattedMessage,
+      `${errorCode}: ${formattedMessage}`,
```
schaable, line 83: I'm not sure if it's still worth passing he `prefix` here. Using only the `formattedMessage` as the `message` may achieve a cleaner look.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -56,98 +56,80 @@ export async function main(cliArguments: string[], print = console.log) {
     hardhatSpecialArgs.configPath = await resolveHardhatConfigPath();
   }
 
-  try {
```
schaable, line 59: I actually didn't change too much here, just removed the `try/catch` as it's handled in `v-next/hardhat/src/cli.ts`

v-next/hardhat/src/internal/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const HARDHAT_NAME = "Hardhat";
```
schaable, line 1: This needs to be merged with `v-next/hardhat/src/internal/cli/init/constants.ts`

v-next/hardhat/test/helpers/env-vars.ts
```diff
@@ -0,0 +1,34 @@
+import { afterEach } from "node:test";
+
+export function createTestEnvManager() {
```
schaable, line 3: I've copied this from core. I think we need to move all these to a test utils module. We already discussed this with Pato, but I wanted to leave a note for awareness.

v-next/hardhat-errors/src/errors.ts
alcuadrado: I think keeping it is pretty important because the formatting is only done in hardhat's CLI, but these errors can be thrown in other places (e.g., a script using hardhat).

re test failures: I fixed that in #5460

v-next/hardhat/src/cli.ts
```diff
@@ -2,10 +2,12 @@
...
-// eslint-disable-next-line no-restricted-syntax -- Allow top-level await here
+/* eslint-disable no-restricted-syntax -- Allow top-level await here */
 const { main } = await import("./internal/cli/main.js");
+const { printErrorMessages } = await import("./internal/cli/error-handler.js");
+/* eslint-enable no-restricted-syntax */
 
 main(process.argv.slice(2)).catch((error: unknown) => {
-  console.error(error);
-  process.exitCode = 1;
+  printErrorMessages(error);
```
alcuadrado, line 11: I'd rather do this in `main`, so that that contains all the logic of the cli, and this file is just a bootstrap.

Also, `process.exit()` shouldn't be used, as it kills the process before giving it the opportunity to flush any i/o buffer (e.g. printing the error messages)

v-next/hardhat/src/internal/cli/error-handler.ts
```diff
@@ -0,0 +1,151 @@
...
+ * {@link ErrorCategory}.
+ * @param print the function used to print the error message, defaults to
+ * `console.error`. Useful for testing to capture error messages.
+ */
+export function printErrorMessages(
+  error: unknown,
+  print: (message: string) => void = console.error,
+): void {
+  const showStackTraces =
+    process.argv.includes("--show-stack-traces") ||
```
alcuadrado, line 72: I'd pass this from `main`, so that we keep everything argv parsing there. It's a bit unexpected here.

alcuadrado (changes requested):
I just left a few comments

v-next/hardhat/test/helpers/env-vars.ts
schaable: Removed as it is no longer used.

v-next/hardhat/src/cli.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/5455/commits/a0b79acec496263f4e79f8027f67c913783b808b

v-next/hardhat/src/internal/cli/error-handler.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/5455/commits/8111de5583807c3c6e62fe51030f313c746cf329

alcuadrado (approved):

PR #5457, opened 2024-06-28
Simplify `tsconfig.json` files with ${configDir}

alcuadrado:
TS 5.5 added a `${configDir}` variable that you can use in shared configs to define values relative to the actual config dir. This allows you to simplify the shared configurations. This PR takes advantage of that.


alcuadrado:
I added a second commit using this variable to set the `typesRoot` to each package's own `@types/` packages. Otherwise all the hoisted `@types/` get loaded (e.g. prettier's types).

schaable (approved):

PR #5460, opened 2024-06-29
Isolate the build-system package errors

alcuadrado:
This PR removes the build system's descriptors to avoid copying over multiple unnecessary ones.

I did this by copying the required descriptors into the package itself and then deleting everything that was not needed. After removing many descriptors, I also had to reorder them. Reviewing commit by commit may be simpler. 

Also, while fixing some errors, I ensured that we weren't using `assert.throws` or `assert.rejects` with error messages. Because of that, too many tests were failing after this refactor.

### DIFF
--- a/v-next/hardhat-errors/src/descriptors.ts
+++ b/v-next/hardhat-errors/src/descriptors.ts
@@ -55,20 +55,7 @@ export const ERROR_CATEGORIES: {
     websiteTitle: "Task definition errors",
   },
   ARGUMENTS: { min: 300, max: 399, websiteTitle: "Arguments related errors" },
-  RESOLVER: {
-    min: 400,
-    max: 499,
-    websiteTitle: "Dependencies resolution errors",
-  },
-  SOLC: { min: 500, max: 599, websiteTitle: "Solidity related errors" },
   BUILTIN_TASKS: { min: 600, max: 699, websiteTitle: "Built-in tasks errors" },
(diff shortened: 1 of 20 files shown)


ChristopherDedominici (approved):

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -231,21 +195,13 @@ Please add the property "type" with the value "module" in your package.json to e
       number: 100,
       messageTemplate: "An internal invariant was violated: {message}",
       websiteTitle: "Invariant violation",
-      websiteDescription: `An internal invariant was violated. This is probably caused by a programming error in Hardhat or in one of the used plugins.
-
+      websiteDescription: `An internal invariant was violated. This is probably caused by a programming in Hardhat or in one of the used plugins.
```
schaable, line 198: typo

schaable (approved):
Looks good, left just one small comment.

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -231,123 +195,154 @@ Please add the property "type" with the value "module" in your package.json to e
...
```
alcuadrado, line 198:
```suggestion
      websiteDescription: `An internal invariant was violated. This is probably caused by a programming error in Hardhat or in one of the used plugins.
```

PR #5461, opened 2024-06-29
New only-hardhat-error rule

alcuadrado:
This PR introduces a `no-restricted-syntax` rule that forbids throwing any error except for:

- `throw new HardhatError(...)`
- `throw variableName` within `catch` clauses. — The rationale here is to allow re-throwing of the error that was caught.

We could later improve this with a custom rule to:

- Allow subclasses of `HardhatError`
- Only allow `throw variable` when the variable is the one in the `catch (variable)`.

I believe that `no-restricted-syntax` isn't expressive enough for these things, as they are semantic conditions and not entirely syntactic. Still, I think this is a good trade-off.

PS: This depends on #5460 just for convenience and to avoid merge conflicts.

### DIFF
--- a/config-v-next/eslint.cjs
+++ b/config-v-next/eslint.cjs
@@ -10,9 +10,10 @@ const path = require("path");
  * The only packages that should not use this config are our own eslint plugins/rules.
  *
  * @param {string} configFilePath The path to the config file that is using this function.
+ * @param {{ onlyHardhatError?: boolean }} options An object with options to enable/disable certain rules.
  * @returns {import("eslint").Linter.Config}
  */
-function createConfig(configFilePath) {
+function createConfig(configFilePath, options = { onlyHardhatError: true }) {
   /**
    * @type {import("eslint").Linter.Config}
    */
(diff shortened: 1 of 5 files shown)


config-v-next/eslint.cjs
```diff
@@ -393,6 +394,25 @@ function createConfig(configFilePath) {
...
+    config.overrides?.push({
+      files: ["src/**/*.ts"],
+      rules: {
+        "no-restricted-syntax": [
+          "error",
+          {
+            // This is a best effor selector that forbids every throw unless it's a `new HardhatError`,
+            // or throwing a variable within a catch clause.
+            selector:
+              "ThrowStatement:not(:has(ThrowStatement > NewExpression[callee.name='HardhatError']), CatchClause ThrowStatement:has(ThrowStatement > Identifier))",
```
schaable, line 407: should we consider HardhatPluginError as well?

schaable (approved):
LGTMT

config-v-next/eslint.cjs
alcuadrado: No, as official plugins shouldn’t throw those

PR #5464, opened 2024-07-01
feat: added etherlink and etherlink testnet

devon-n:
I wanted to add a network that I use to hardhat verify.

### PR adds following changes
- [x] Added Etherlink and etherlink testnet to verify smart contracts

### DIFF
--- a/packages/hardhat-verify/src/internal/chain-config.ts
+++ b/packages/hardhat-verify/src/internal/chain-config.ts
@@ -187,6 +187,22 @@ export const builtinChains: ChainConfig[] = [
     },
   },
   {
+    network: "etherlink",
+    chainId: 42793,
+    urls: {
+      apiURL: "https://explorer.etherlink.com/api/v2",
+      browserURL: "https://explorer.etherlink.com",
+    },
+  },
+  {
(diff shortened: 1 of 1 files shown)


devon-n:
Hi Guys, Is this possible to merge?

schaable:
Hi @devon-n! As stated [here](https://github.com/NomicFoundation/hardhat/blob/0826378460b301ccbc4a3351aa9f34609a48b149/packages/hardhat-verify/src/internal/chain-config.ts#L285), we're not expanding the list of networks supported by `hardhat-verify`. The main reason behind this decision is the difficulty in maintaining an ever-growing list of networks.

However, there's a workaround for users looking to add custom networks. You can do this by configuring them in the [etherscan custom network section](https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify#adding-support-for-other-networks) of your hardhat configuration.

For a better user experience, consider abstracting this into a plugin. Here's an [example plugin](https://github.com/schaable/hardhat-custom-chain-plugin-example) on how to do that. Once set up, you can simply import this plugin into your hardhat configuration, removing the need to manually add networks to the config.

PR #5470, opened 2024-07-02
Updates to RESERVED_PARAMETER_NAMES and GLOBAL_OPTIONS

schaable:
See https://github.com/NomicFoundation/hardhat/issues/5418#issuecomment-2201166033 for details. This can be reviewed commit by commit.


alcuadrado (approved):
Approved, but let's move the builtin options to the construction of the global options map, within hardhat

Issue #5452, opened 2024-06-28
Implement logic in hh-utils to handle sub processes


ChristopherDedominici:
[Parent task](https://github.com/NomicFoundation/hardhat/issues/5450)

PR #5471, opened 2024-07-02
Implement logic in hh-utils to handle sub processes

ChristopherDedominici:
[Link to task description](https://github.com/NomicFoundation/hardhat/issues/5452)

FEATURES
- add a function to spawn a subprocess. The subprocess will not be awaited by the parent.
- test: spawn a subprocess that creates a JSON file with values that are the arguments passed to the subprocess. In the test, check that the values in the file are the expected ones. To wait for the subprocess to finish, an interval is set.


v-next/hardhat-utils/src/sub-process.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export function spawnSubProcess(
+  absolutePathToSubProcessFile: string,
+  args: string[] = [],
+): void {
+  const subprocess = spawn(
+    process.execPath,
+    ["--import", "tsx/esm", absolutePathToSubProcessFile, ...args],
+    {
+      detached: true,
```
ChristopherDedominici, line 21: Do we want to allow to pass ENV variables too via the "env" option?

v-next/hardhat-utils/src/sub-process.ts
```diff
@@ -0,0 +1,27 @@
...
+ * Spawns a detached subprocess to execute a given file with optional arguments.
+ *
+ * @param absolutePathToSubProcessFile - The absolute path to the subprocess file to be executed.
+ * @param [args=[]] - Optional arguments to pass to the subprocess.
+ *
+ * The subprocess runs in a detached mode and its standard input/output is ignored.
+ * This function does not wait for the subprocess to complete and the subprocess is unreferenced
+ * to allow the parent process to exit independently.
+ */
+export function spawnSubProcess(
```
schaable, line 13: should we handle errors?

ChristopherDedominici: Good question. Not sure because if something fails is not related to hardhat errors. So not sure if it makes sense to catch anything. @alcuadrado what do you think?

schaable (approved):

v-next/hardhat-utils/src/sub-process.ts
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * The subprocess runs in a detached mode and its standard input/output is ignored.
+ * This function does not wait for the subprocess to complete and the subprocess is unreferenced
+ * to allow the parent process to exit independently.
+ */
+export function spawnDetachedSubProcess(
+  absolutePathToSubProcessFile: string,
+  args: string[] = [],
+): void {
+  const subprocess = spawn(
```
alcuadrado, line 17: I think we should use `execFile` instead. The reason is that `exec` and `spawn` create a new shell, so if  you have a slow `bashrc`/`zshrc`, this will be slow.

Instead, `execFile` executes a file directly, ignoring the shell.

alcuadrado: We should also try to get some sense of how slow this is. Not awaiting it of course, but launching it. If it's too slow maybe doing this from a worker thread is better.

v-next/hardhat-utils/src/sub-process.ts
```diff
@@ -0,0 +1,27 @@
...
+    process.execPath,
+    ["--import", "tsx/esm", absolutePathToSubProcessFile, ...args],
```
alcuadrado, line 19: Let's only add this import if the file is `.ts`.

v-next/hardhat-utils/src/sub-process.ts
alcuadrado: Yes, I think we'd need it.

v-next/hardhat-utils/src/sub-process.ts
```diff
@@ -0,0 +1,27 @@
...
+    {
+      detached: true,
+      stdio: "ignore",
```
alcuadrado, line 22: I'm not sure if this works on windows, see `https://nomic-labs.sentry.io/issues/5567768827/?referrer=slack&notification_uuid=5783a85c-f818-4109-be8a-5de9dfdf3212&alert_rule_id=1079337&alert_type=issue`

But we can address it later.

v-next/hardhat-utils/test/fixture-projects/sub-process/sub-process.ts
```diff
@@ -0,0 +1,23 @@
...
+  "tmp-file.json",
+);
+
+const subprocessInfo = {
+  executed: false,
+  arg1: "",
+  arg2: "",
+};
+
+(async () => {
```
alcuadrado, line 17: Why async here?

v-next/hardhat-utils/src/sub-process.ts
alcuadrado: No, the subprocess should be responsible to handle its own errors

v-next/hardhat-utils/test/fixture-projects/sub-process/sub-process.ts
alcuadrado: I guess you wanted to introduce a small delay, but I believe this gets immediately executed, because there's nothing else in the event loop

v-next/hardhat-utils/src/sub-process.ts
alcuadrado: Can you add a `console.log` and `console.error` call to your test to make sure they still work?

v-next/hardhat-utils/src/sub-process.ts
alcuadrado: If launching it fails, that'd be a fatal error and hardhat should crash.

v-next/hardhat-utils/src/sub-process.ts
alcuadrado: I was wrong about this. `spawn` is what we need. I doesn't initialize a shell unless you use `shell:true`

v-next/hardhat-utils/package.json
```diff
@@ -26,7 +26,8 @@
     "./number": "./dist/src/number.js",
     "./package": "./dist/src/package.js",
     "./request": "./dist/src/request.js",
-    "./string": "./dist/src/string.js"
+    "./string": "./dist/src/string.js",
+    "./sub-process": "./dist/src/sub-process.js"
```
alcuadrado, line 30: This is very nitpicky, but i think most people consider `subprocess` a single word

ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5471/commits/a1354fca15ac2aa95bc5082f157cc5338bcc29c5)

v-next/hardhat-utils/test/fixture-projects/sub-process/sub-process.ts
ChristopherDedominici: Actually it's just a distraction error, it was not supposed to be there. Removed [here](https://github.com/NomicFoundation/hardhat/pull/5471/commits/a1354fca15ac2aa95bc5082f157cc5338bcc29c5)

v-next/hardhat-utils/src/sub-process.ts
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/5471/commits/870d0e5d59f95aa9a4ff872dd8498604706dc721)

v-next/hardhat-utils/src/sub-process.ts
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/5471/commits/9fa9f2c7ac30b7ffbd5ea007eec221fbf616a73c)

v-next/hardhat-utils/package.json
schaable: you should rename the function name as well

alcuadrado: it's still `sub-process` in the package.json

v-next/hardhat-utils/test/subprocess.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    await checkIfSubprocessWasExecuted();
+
+    const subprocessInfo = await readJsonFile(
+      ABSOLUTE_PATH_TO_TMP_RESULT_SUBPROCESS_FILE,
+    );
+
+    // Checks if the file created by the subprocess contains the expected data.
+    // The subprocess writes its received arguments to a JSON file.
+    assert.deepEqual(subprocessInfo, {
```
alcuadrado, line 66: I think this tests is too strict. You can't assume that the OS won't set any other env variable. I would just check that the ones that you define were passed.

v-next/hardhat-utils/package.json
alcuadrado: resolving and moving the conversation to the new version of this file

v-next/hardhat-utils/package.json
```diff
@@ -26,7 +26,8 @@
     "./number": "./dist/src/number.js",
     "./package": "./dist/src/package.js",
     "./request": "./dist/src/request.js",
-    "./string": "./dist/src/string.js"
+    "./string": "./dist/src/string.js",
+    "./sub-process": "./dist/src/subprocess.js"
```
alcuadrado, line 30: I think you forgot to rename it here, it's still `sub-process`

v-next/hardhat-utils/src/subprocess.ts
```diff
@@ -0,0 +1,32 @@
...
+  const subprocessArgs = [absolutePathToSubProcessFile, ...args];
+
+  if (absolutePathToSubProcessFile.endsWith(".ts")) {
+    subprocessArgs.unshift("--import", "tsx/esm");
+  }
+
+  const subprocess = spawn(process.execPath, subprocessArgs, {
+    detached: true,
+    env,
+    stdio: "ignore",
```
alcuadrado, line 28: At least one of the tests should print to stdout and stderr to validate that this works

alcuadrado (changes requested):
Just left a few comments

v-next/hardhat-utils/test/subprocess.ts
ChristopherDedominici: Done in [this commit](https://github.com/NomicFoundation/hardhat/pull/5471/commits/b3ed594afe2438f7585cb8834b7359a829e973f4) with other improvements

v-next/hardhat-utils/package.json
ChristopherDedominici: Done in [this commit](https://github.com/NomicFoundation/hardhat/pull/5471/commits/b3ed594afe2438f7585cb8834b7359a829e973f4) with other improvements

v-next/hardhat-utils/src/subprocess.ts
ChristopherDedominici: Done in [this commit] (https://github.com/NomicFoundation/hardhat/pull/5471/commits/b3ed594afe2438f7585cb8834b7359a829e973f4) with other improvements and[ this commit](https://github.com/NomicFoundation/hardhat/pull/5471/commits/b7dd913a82ab58d71a5f603fee8a661e14a1bfac)

v-next/hardhat-utils/test/fixture-projects/subprocess/ts-subprocess-stdio.ts
```diff
@@ -0,0 +1,23 @@
...
+const ABSOLUTE_PATH_TO_TMP_RESULT_SUBPROCESS_FILE: string = path.join(
+  path.dirname(fileURLToPath(import.meta.url)),
+  "tmp-file.json",
+);
+
+const subprocessInfo = {
+  executed: false,
+};
+
+(() => {
```
ChristopherDedominici, line 15: Console functions to check windows behavior

v-next/hardhat-utils/src/subprocess.ts
ChristopherDedominici: @alcuadrado windows tests passed with console.log and console.error in the subprocess. I guess that this means that everything is ok (?). 
Link to subprocess code: https://github.com/NomicFoundation/hardhat/pull/5471/files#r1665894434

alcuadrado: great!

alcuadrado (approved):

PR #5491, opened 2024-07-05
Add file in hh-utils to handle hardhat configuration paths

ChristopherDedominici:
[Link to task](https://github.com/NomicFoundation/hardhat/issues/5490)


ChristopherDedominici:
This code will be moved in hh-core

PR #5492, opened 2024-07-05
Add `hardhat-test-utils` package

alcuadrado:
This PR introduces a new private package, `@nomicfoundation/hardhat-test-utils`, and installs it in all the `v-next` packages that are compatible with it (i.e. not dendencies of it).

This PR only introduces helpers for `HardhatError` related assertions, but it doesn't change any existing test.

A follow-up PR will change the existing tests to use the new utilities.

### DIFF
--- a/v-next/hardhat-build-system/package.json
+++ b/v-next/hardhat-build-system/package.json
@@ -38,9 +38,9 @@
     "README.md"
   ],
   "dependencies": {
-    "@nomicfoundation/ethereumjs-util": "9.0.4",
     "@ignored/hardhat-vnext-errors": "workspace:^3.0.0-next.0",
     "@ignored/hardhat-vnext-utils": "workspace:^3.0.0-next.0",
+    "@nomicfoundation/ethereumjs-util": "9.0.4",
     "@nomicfoundation/solidity-analyzer": "^0.1.0",
     "adm-zip": "^0.4.16",
     "aggregate-error": "^3.0.0",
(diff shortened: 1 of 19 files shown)


v-next/hardhat-build-system/package.json
```diff
@@ -38,9 +38,9 @@
     "README.md"
   ],
   "dependencies": {
-    "@nomicfoundation/ethereumjs-util": "9.0.4",
```
alcuadrado, line 41: Jus an automatic reordering that `pnpm` did

v-next/hardhat-build-system/package.json
```diff
@@ -61,6 +61,7 @@
   "devDependencies": {
     "@eslint-community/eslint-plugin-eslint-comments": "^4.3.0",
     "@ignored/hardhat-vnext-node-test-reporter": "workspace:^3.0.0-next.0",
+    "@nomicfoundation/hardhat-test-utils": "workspace:^",
```
alcuadrado, line 64: Contrary to the rest of our packages, we do `workspace:^` here because this is a private package, which we don't intend to publish, so versioning is not an issue.

v-next/hardhat-test-utils/src/hardhat-error.ts
```diff
@@ -0,0 +1,68 @@
...
+    ensureError(error);
+    assertIsHardhatError(error, descritor, messageArguments);
+
+    return;
+  }
+
+  assert.fail("Function did not throw any error");
+}
+
+export async function assertThrowsHardhatErrorAsync<
```
schaable, line 47: how about `assertRejectsWithHardhatError`?

schaable (approved):
LGTM! One minor suggestion regarding naming: I liked how Node assert distinguishes between throwing and rejecting.

PR #5494, opened 2024-07-08
Include the builtin options as part of the global options & augment the types

schaable:
This is a follow-up PR to [#5470](https://github.com/NomicFoundation/hardhat/pull/5470) based on discussions with Pato. In this PR: 

- We pass the built-in global options as part of the global options to `createHardhatRuntimeEnvironment`.
- We augment the `GlobalOption` types with the built-in option types.


v-next/hardhat/test/internal/cli/helpers/getGlobalHelpString.ts
```diff
@@ -7,34 +7,33 @@ import { buildGlobalOptionDefinitions } from "@ignored/hardhat-vnext-core";
...
       const tasks = new Map();
-      const help = await getGlobalHelpString(tasks, new Map());
+      const globalOptionDefinitions = new Map();
+      const help = await getGlobalHelpString(tasks, globalOptionDefinitions);
 
       const expected = `Hardhat version ${packageJson.version}
 
 Usage: hardhat [GLOBAL OPTIONS] <TASK> [SUBTASK] [TASK OPTIONS] [--] [TASK ARGUMENTS]
 
 GLOBAL OPTIONS:
```
schaable, line 26: I believe we should consider this a bug in `getGlobalHelpString`, as it shows the GLOBAL OPTIONS section even when there are no global options. In reality, this will never happen since we always have the built-in global options, so it's probably not a high-priority bug.

v-next/example-project/hardhat.config.ts
```diff
@@ -126,6 +128,15 @@ const pluginExample = {
   ],
 };
 
+// Type includes the builtin global options
```
alcuadrado, line 131: Is this meant to be a type-level tests? If so, let's place it in `hardhat`, as here it may go away accidentally.

v-next/hardhat/src/internal/builtin-global-options.ts
```diff
@@ -0,0 +1,56 @@
+import type { GlobalOptionDefinitions } from "../types/global-options.js";
+
+import { globalOption, ArgumentType } from "../config.js";
+
+export const BUILTIN_GLOBAL_OPTIONS_DEFINITIONS: GlobalOptionDefinitions =
+  new Map([
+    [
+      "config",
+      {
+        pluginId: "hardhat",
```
alcuadrado, line 10: Should we make this optional? Or maybe switch to `builtin` or something more explicit?

v-next/hardhat/test/internal/cli/helpers/getGlobalHelpString.ts
alcuadrado: agreed

alcuadrado: It's mostly a weird artifact of testing the helper function instead of the higher level end-to-end thing though

alcuadrado (changes requested):
I think that these changes won't initialize the HRE with the builtin global options when doing `import "hardhat"`. There was a similar with plugins that I fixed in github.com/nomicfoundation/hardhat/pull/5485

v-next/hardhat/src/internal/builtin-global-options.ts
schaable: I thought about making it optional, but this required less changes. `builtin` also works for me. what do you prefer?

v-next/example-project/hardhat.config.ts
schaable: yeah it was the first thing I did to test the augmentation works. But then I added a type test, so we can probably remove it.

schaable: Done in https://github.com/NomicFoundation/hardhat/pull/5494/commits/1deda23713471da2ef4a85a4ef4c22e42b4d959b

v-next/hardhat/src/internal/builtin-global-options.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/5494/commits/26c1d6c8beb6cca12942f03108e0029d466fefe8

schaable:
> I think that these changes won't initialize the HRE with the builtin global options when doing `import "hardhat"`. There was a similar with plugins that I fixed in github.com/[/pull/5485](https://github.com/NomicFoundation/hardhat/pull/5485)

Done in https://github.com/NomicFoundation/hardhat/pull/5494/commits/88d8e5dfb3ebe1be930ddd34a66c45ef288e4688. I've also noticed that we were missing the `init` builtin option and `config` was incorrectly named `configPath`. It's fixed in https://github.com/NomicFoundation/hardhat/pull/5494/commits/c07d1b5eff58fd8ed54de83fd5a0c46d01f06f22 and https://github.com/NomicFoundation/hardhat/pull/5494/commits/e09c0f472a1b5a647ad2443646592234288b71cb

alcuadrado (approved):

PR #5496, opened 2024-07-08
Use `hardhat-test-utils`' functions

alcuadrado:
This PR adds new `eslint` rules to forbid the error-related assertions that `node:assert/strict` has, and forces us to use our own instead.

It also fixies all the new linter errors, so it's a large and repetitive PR.

### DIFF
--- a/v-next/hardhat-test-utils/src/hardhat-error.ts
+++ b/v-next/hardhat-test-utils/src/hardhat-error.ts
@@ -5,6 +5,14 @@ import assert from "node:assert/strict";
 import { HardhatError } from "@ignored/hardhat-vnext-errors";
 import { ensureError } from "@ignored/hardhat-vnext-utils/error";
 
+/**
+ * Asserts that an error is a HardhatError with a certain descriptor and message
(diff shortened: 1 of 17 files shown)


v-next/hardhat-test-utils/src/hardhat-error.ts
```diff
@@ -44,18 +60,27 @@ export function assertThrowsHardhatError<
...
+ * @param op The async operation. If it's a function, it's called and awaited.
+ * @param descritor The error descriptor.
+ * @param messageArguments The message arguments.
+ */
 export async function assertRejectsWithHardhatError<
   ErrorDescriptorT extends ErrorDescriptor,
 >(
-  f: (() => Promise<any>) | Promise<any>,
+  op: (() => Promise<any>) | Promise<any>,
   descritor: ErrorDescriptorT,
```
schaable, line 76: oops, we missed this in the previous PR. There's a typo here, and it's also present in the other function.

schaable (approved):

Issue #5419, opened 2024-06-20
Consider making all task options (FKA named arguments) optional


alcuadrado:
During the review of https://github.com/NomicFoundation/hardhat/pull/5409, @zoeyTM and I discussed that maybe we should force all the task options to be optional, as their name implies.

It's unclear if we (or our users) ever used an `--required-param <value>` like this.

PR #5499, opened 2024-07-09
Make options optional

schaable:
Merge after https://github.com/NomicFoundation/hardhat/pull/5495
Closes https://github.com/NomicFoundation/hardhat/issues/5419

### DIFF
--- a/v-next/core/src/internal/tasks/resolved-task.ts
+++ b/v-next/core/src/internal/tasks/resolved-task.ts
@@ -1,14 +1,15 @@
-import type { ArgumentValue } from "../../types/arguments.js";
+import type {
+  ArgumentValue,
+  OptionDefinition,
+  PositionalArgumentDefinition,
+} from "../../types/arguments.js";
 import type { HardhatRuntimeEnvironment } from "../../types/hre.js";
 import type {
-  TaskOptionDefinition,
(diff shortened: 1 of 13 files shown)


v-next/core/src/internal/tasks/resolved-task.ts
```diff
@@ -98,25 +99,34 @@ export class ResolvedTask implements Task {
...
+
+    // Validate and resolve the task options
+    for (const option of this.options.values()) {
+      const value = taskArguments[option.name];
+
+      this.#validateArgumentType(option, value);
+
+      // resolve defaults for optional arguments
+      if (value === undefined) {
+        taskArguments[option.name] = option.defaultValue;
```
alcuadrado, line 112: I wouldn't modify this input here.

alcuadrado (changes requested):
I just left two small comments.

Unfortunately, I think all your open PRs need to be rebased because of the new assertion-related eslint rules.

v-next/core/src/internal/tasks/resolved-task.ts
schaable: Will be addressed in a follow-up PR.

alcuadrado (approved):

Issue #5502, opened 2024-07-09
Use the telemetry logic to report if the telemetry consent was given or not


ChristopherDedominici:
Implemented in [this PR](https://github.com/NomicFoundation/hardhat/pull/5545)

PR #5503, opened 2024-07-09
Implement logic to ask for telemetry consent #5501

ChristopherDedominici:
[Link to task](https://github.com/NomicFoundation/hardhat/issues/5501)

Add logic to ask for telemetry consent. 
The consent should be registered but this feature will be implemented in [this work item](https://github.com/NomicFoundation/hardhat/issues/5502), not here.

There are no tests because the telemetry logic cannot be triggered in the CI


v-next/hardhat/src/internal/cli/telemetry/telemetry-consent.ts
```diff
@@ -0,0 +1,82 @@
...
+  const consent = await confirmTelemetryConsent();
+
+  if (consent === undefined) {
+    return false;
+  }
+
+  // Store user's consent choice
+  await writeJsonFile(await getTelemetryConsentFilePath(), { consent });
+
+  // TODO: this will be enabled in a following PR as soon as the function to send telemetry is implemented
```
ChristopherDedominici, line 67: Will be done in [this work item](https://github.com/NomicFoundation/hardhat/issues/5502). These tasks are under the same parent feature

v-next/hardhat/src/internal/cli/prompt/prompt.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (result === undefined) {
+      await prompt.cancel();
+      return undefined;
+    }
+
+    return result;
+  } catch (e) {
+    if (e === "") {
+      // If the user cancels the prompt, we quit
+      process.exit();
```
schaable, line 44: it is correct to terminate the process?

v-next/hardhat/src/internal/cli/prompt/prompt.ts
```diff
@@ -0,0 +1,59 @@
...
+ * @param name Used as the key for the answer on the returned values (answers) object.
+ * @param message The message to display when the prompt is rendered in the terminal.
+ * @param timeoutMilliseconds After how much time the prompt will be cancelled if no answer is given.
+ * @returns True or false based on the user input, or undefined if the prompt times out.
+ */
+export async function confirmationPromptWithTimeout(
+  name: string,
+  message: string,
+  timeoutMilliseconds: number = 10_000,
+): Promise<boolean | undefined> {
```
schaable, line 12: I'm not a fan of returning `| undefined`. If the timeout ends without response, can't we assume the response is `false`?

schaable (approved):
Looks good, just some minor comments

v-next/hardhat/src/internal/cli/prompt/prompt.ts
ChristopherDedominici: Good question!
Let me give you a bit of context: I use undefined because later in the logic we store the consent value on file. And if we hit the timeout it means that  we did not get a true or false value about the consent from the user (which most likely is not a person but a CI if we hit 10 seconds timeout). 
So by returning undefined I can handle this specific scenario and skip saving the consent value on file (if I just receive false I could think that the user denied the telemetry consent).
But we could argue that we could save false for undefined too on the file.
Wdyt?

schaable: Thanks for the explanation! Given that information I think we can leave it as it is.

alcuadrado: I think having three possible values is inherent to this API, as that's the expected behavior. `| undefined` is probably the less annoying third value in typescript.

v-next/hardhat/src/internal/cli/prompt/prompt.ts
alcuadrado: I think we should return undefined here, as it represents the lack of an answer. This should be clear in the function docs.

In general, we should never exit the process from helper functions like this one, as that's unexpected.

v-next/hardhat/src/internal/cli/telemetry/telemetry-consent.ts
```diff
@@ -0,0 +1,82 @@
...
+      .consent;
+  }
+
+  // Telemetry consent not provided yet, ask for it
+  return requestTelemetryConsent();
+}
+
+function canTelemetryBeEnabled(isHelpCommand: boolean): boolean {
+  return (
+    !isHelpCommand &&
```
alcuadrado, line 45: This was here in v2 because telemitry wasn't done using a subprocess and it was slow. We don't need it anymore.

alcuadrado (changes requested):
Just left a few comments

v-next/hardhat/src/internal/cli/prompt/prompt.ts
ChristopherDedominici: Done in [this commit](https://github.com/NomicFoundation/hardhat/pull/5503/commits/0ca57efb65ebadaf11d07d755a750f37bc684bf8)

v-next/hardhat/src/internal/cli/telemetry/telemetry-consent.ts
ChristopherDedominici: Done in [this commit](https://github.com/NomicFoundation/hardhat/pull/5503/commits/6201713b17dc8a944bbc40ffffe8963224412f06)

alcuadrado (approved):

PR #5504, opened 2024-07-09
Adds logic for forcing through verification when is verified check fails

3commascapital:

---

There is a bug in blockscout's latest releases. It has been around for quite a while. Unfortunately it only happens in certain situations. The [issue](https://github.com/blockscout/blockscout/issues/9001) is that an `isVerified` check will fail if the backend is put into a certain state. The easy way around this would simply be to include that set of errors / states to be ignored when provided a force flag. 

I do not think that hardhat verify should cover all of the quirks of the many backends that are bound to emerge into eternity, however, I think that the force flag being more comprehensive (less narrowly applied) than it currently is could be a reasonable middle ground, provided that it does not run afoul of any eip's.

### DIFF
--- a/.changeset/fuzzy-eyes-collect.md
+++ b/.changeset/fuzzy-eyes-collect.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/hardhat-verify": patch
 ---
 
-Makes force flag comprehensive enough to cover failing optional calls
+Make the `--force` flag override the check of any existing verification, even in the presence of errors.
--- a/.changeset/fuzzy-eyes-collect.md
+++ b/.changeset/fuzzy-eyes-collect.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
+Makes force flag comprehensive enough to cover failing optional calls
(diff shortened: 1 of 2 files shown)


.changeset/fuzzy-eyes-collect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
+Makes force flag comprehensive enough to cover failing optional calls
```
alcuadrado, line 5: What about this?
```suggestion
Make the `--force` flag override the check of any existing verification, even in the presence of errors.
```

alcuadrado:
> I do not think that hardhat verify should cover all of the quirks of the many backends that are bound to emerge into eternity, however, I think that the force flag being more comprehensive (less narrowly applied) than it currently is could be a reasonable middle ground, provided that it does not run afoul of any eip's.

Agreed. I think this is a reasonable exception.

Thanks for the PR!

gibsfinancedev:
This is great. Thank you!

3commascapital:
Looks great. Thank you!

alcuadrado (approved):

Issue #5224, opened 2024-05-15
Add Blockscout as verification provider 


rimrakhimov:
### Describe the feature

We suggest adding Blockscout as a verification provider to Hardhat. Technically, it is possible right now to verify contracts on blockscout Explorer via etherscan configuration. But we would like to propose adding an explicit blockscout option along with existing sourcify and etherscan providers.

### Motivation

Using etherscan configuration requires specifying an API key even though blockscout does not need one. Users have to select any non-empty key to avoid that limitation. Having blockscout specified explicitly allows them to have blockscout support without any workarounds.

It also helps to integrate blockscout-specific features into the process in the future. For example, for enforcing smart contracts verification, added in https://github.com/NomicFoundation/hardhat/pull/5164, it does not distinguish between partially and fully verified contracts, returning an error if the contract has already been fully verified. We may add that distinction to the usual verification process to avoid using the `--force` flag. Besides, we are currently working on a separate application that will allow us to verify the contracts on several blockscout chains ([link](https://verification-app.k8s-dev.blockscout.com/)). In the future, this could be used to prevent users from specifying the details for the most common chains at all.

We are witnessing an increasing number of users, in general, asking whether hardhat supports blockscout and how to use it for contract verification. And it would be nice if Hardhat could support blockscout as a separate option.

### Suggestion

Our proposal involves creating a separate Blockscout class that would be integrated into the current Hardhat configuration. This class would share most of its methods with the existing Etherscan implementation, with the key difference being the absence of the `api_key` option requirement. Further enhancements can be discussed in separate issues.

If agreed upon, we may work on the corresponding PR with your guidance on how you see that implemented.

alcuadrado:
Hi @rimrakhimov,

We discussed this internally and it looks like something we'd be open to add to the hardhat-verify plugin. We don't have the bandwidth to do it ourselves right now. Will you be open to send a PR?

If so, I think we can start with you providing a list of changes that you think are needed, both at the product/functional level, and technical level. Then we can go through it, and provide feedback and technical guidance.

rimrakhimov:
Hi @alcuadrado, I have the capacity to work on that PR now.

From a functional point of view, I see it this way:

1. Add a new verification provider inside the `hardhat.config` with fields identical to `Etherscan` except for the `apiKey` value, which can be added but is not required to verify the contracts.
2. For the provider, do not require `constructorArguments` during verification, as they are not used by Blockscout anyway.
[...]

PR #5507, opened 2024-07-10
Feature: blockscout plugin support

rimrakhimov:
The blockscout plugin support suggestion: https://github.com/NomicFoundation/hardhat/issues/5224

---

### DIFF
--- a/packages/hardhat-verify/src/blockscout.ts
+++ b/packages/hardhat-verify/src/blockscout.ts
@@ -1,0 +1,1 @@
+export { Blockscout } from "./internal/blockscout";
--- a/packages/hardhat-verify/src/index.ts
+++ b/packages/hardhat-verify/src/index.ts
@@ -17,10 +17,12 @@ import {
   TASK_VERIFY_SOURCIFY,
   TASK_VERIFY_SOURCIFY_DISABLED_WARNING,
   TASK_VERIFY_GET_CONTRACT_INFORMATION,
+  TASK_VERIFY_BLOCKSCOUT,
 } from "./internal/task-names";
 import {
   etherscanConfigExtender,
   sourcifyConfigExtender,
+  blockscoutConfigExtender,
 } from "./internal/config";
 import {
   InvalidConstructorArgumentsError,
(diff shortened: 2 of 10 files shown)


alcuadrado:
I had a call with @rimrakhimov, and we decided to review this and get it merged. The intention of this PR is not to use a different blockscout specific API, but to be able to configure and use blockscout as a first-class citizen like etherscan and sourcify.

alcuadrado:
@schaable, once you get back, can you review this? You are more familiar with the different providers that the plugin has.

schaable (approved):
Looks good. There are a few things I'd like to improve, such as the chain list duplication, general messaging, and updating the docs, but we can leave it for a future iteration. 

Tests:

1) Each provider individually (Etherscan with constructor args, Sourcify and Blockscout without constructor args).
2) All providers simultaneously with constructor args.
3) All providers simultaneously without constructor args (Etherscan fails, Sourcify and Blockscout succeed).
4) All providers with an already verified contract.
5) All providers, with one throwing an error (contract already verified error for Blockscout). This occurred frequently when verifying with all providers. If you visit the contract address in Blockscout, it shows "This contract has been verified using Blockscout Bytecode Database." We may need to update the message to avoid throwing an error in this case.
6) All providers throwing an error.

rimrakhimov:
Hi @schaable , thanks for the review! I'm interested if you want us to implement those changes or you've just added them to your internal backlog? We may probably at least make a PR with updates into documentation 

Also, as further enhancements we discussed with @alcuadrado the possibility to 
1) Add more networks into blockscout provider available by default - what are the requirements for chains in order to be specified in the default network list?)
2) Make blockscout provider enabled by default along with etherscan in a way, that if both blockscout and etherscan have the chain information, then verification occurs on both of them

Wdyt about those changes and what do we need to do to have them added?

pcaversaccio:
I was quickly testing the plugin. A couple of observations:
1. I faced some weird error for constructor arguments even though it verified everywhere. The command I used was `npx hardhat verify --network sepolia 0xF8bec66138C75B0C068e290820dF59227297D807` in that case

image

2. Why doesn't it verify also on Etherscan in parallel?

2. I use `sourcify` as enabled usually in my config and since Blockscout syncs it with Sourcify I don't see any reason why you should again to a verification attempt (since already verified). It only makes sense for chains where Sourcify is not supported but Blockscout is.

3. The plugin doesn't work programmatically currently as it appears. Put this into your deployment script and you will see:

```ts
await hre.run("verify:verify", {
  address: contractAddress,
  constructorArguments: constructorArgs,
});
```

PR #5512, opened 2024-07-11
Upgrade to EDR v0.5 and add `enableRip7212` option

Wodann:
This PR upgrades EDR to v0.5 which includes support for EIP-7212.

Closes https://github.com/NomicFoundation/edr/issues/523.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/vm/exit.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/vm/exit.ts
@@ -24,6 +24,7 @@ export class Exit {
       case SuccessReason.Stop:
       case SuccessReason.Return:
       case SuccessReason.SelfDestruct:
+      case SuccessReason.EofReturnContract:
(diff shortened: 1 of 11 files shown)


fvictorio:
I added the Hardhat Network flag here, I think it makes sense to review everything in a single PR.

packages/hardhat-core/src/internal/hardhat-network/provider/vm/exit.ts
```diff
@@ -24,6 +24,7 @@ export class Exit {
       case SuccessReason.Stop:
       case SuccessReason.Return:
       case SuccessReason.SelfDestruct:
+      case SuccessReason.EofReturnContract:
```
kanej, line 27: Is this new from the latest EDR and unrelated to Rip-7212?

fvictorio: Yes, this is something I should've updated in the previous version bump but didn't.

kanej (approved):

PR #5517, opened 2024-07-12
[WIP] Project root resolution

### LINKED ISSUE #5436: Project root resolution
Figure out how to resolve the project root and how it interacts with dependencies.

## TODO

- [x] https://github.com/NomicFoundation/hardhat/pull/5517
- [x] https://github.com/NomicFoundation/hardhat/issues/5551

alcuadrado:
This PR revamps how we manage the project root resolution and how it impacts other parts of the system.

The project root in Hardhat v3 will be the root of the npm package containing the Hardhat config file, or the one containing the CWD if an instance of the HRE is initialized without using a config file.

This PR touches several files, so here's a summary of the changes:
- The base HRE constructor in the `core` package now accepts an optional projectRoot.
- There's now a `resolveProjectRoot`  helper exposed in `core`.
- Plugin installation error detection is now run based on the project root, as plugins and their dependencies should all be installed in the npm package containing the project.
- I added that same error detection logic to the hook manager, because loading actions can fail.
- The three ways to initialize the HRE (the cli, importing `hardhat`, and using `createHardhatRuntimeEnvironment`) were updated according to the new project root logic.
- I implemented a small change to the `/package` module in the utils package.

I'm marking this as WIP as I still haven't added tests for the new functionality yet.

Fixes #5436

### DIFF
--- a/v-next/core/src/internal/plugins/detect-plugin-npm-dependency-problems.ts
+++ b/v-next/core/src/internal/plugins/detect-plugin-npm-dependency-problems.ts
@@ -8,26 +8,30 @@ import { HardhatError } from "@ignored/hardhat-vnext-errors";
 import semver from "semver";
 
 /**
- * Validate that a plugin is installed and that its peer dependencies are installed and satisfy the version constraints.
+ * Validate that a plugin is installed and that its peer dependencies are
+ * installed and satisfy the version constraints.
  *
- * @param plugin - the plugin to be validated
- * @param basePathForNpmResolution - the directory path to use for node module resolution, defaulting to `process.cwd()`
+ * @param basePathForNpmResolution the dir path for node module resolution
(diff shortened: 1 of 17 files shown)


zoeyTM (approved):
Pre-approving before tests are added, but I think this looks good!

v-next/core/src/internal/plugins/detect-plugin-npm-dependency-problems.ts
```diff
@@ -19,7 +19,7 @@ import semver from "semver";
  */
 export async function detectPluginNpmDependencyProblems(
```
schaable, line 20: Minor: you revamped the other functions to take the project root as the first parameter, but here it is received as the second parameter

schaable (approved):
LGTM, just left a minor comment about parameter order

Issue #5434, opened 2024-06-26
Improve the types of task arguments


alcuadrado:
When creating the first few tasks we noticed that our types within actions are pretty annoying, as the task arguments are typed as `Record<string, unknown>`.

[...]

PR #5527, opened 2024-07-16
Improve the types of task arguments

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/5434


v-next/core/src/types/tasks.ts
```diff
@@ -123,6 +131,12 @@ export type TaskDefinition =
   | NewTaskDefinition
   | TaskOverrideDefinition;
 
+export type ExtendTaskArguments<
```
alcuadrado, line 134:
```suggestion
/**
 * This helper type adds an argument to an existing TaskArgumentsT.
 **/
export type ExtendTaskArguments<
```

alcuadrado (approved):
Great work! I just left a suggestion to document a helper type so that it's easier to understand.

PR #5528, opened 2024-07-17
User config validation

### LINKED ISSUE #5230: Implement and test the default config validation
Implement and test the default config validation (i.e. validation for plugins and tasks — The only things available in core).

zoeyTM:
This PR adds validation for all the currently supported fields in `HardhatUserConfig` (currently just `tasks` and `plugins`). 

Fixes #5230


v-next/core/src/type-guards.ts
```diff
@@ -0,0 +1,70 @@
...
+
+import {
+  ArgumentType,
+  type OptionDefinition,
+  type PositionalArgumentDefinition,
+} from "./types/arguments.js";
+import { TaskDefinitionType } from "./types/tasks.js";
+
+function isValidEnumValue(
+  theEnum: Record<string, string>,
```
ChristopherDedominici, line 11: Double check: do we also want to allow numbers? Record<string, string|number> ?

zoeyTM: This is the same function we're using in Ignition, I just copied it over. My assumption would be that we probably wouldn't use numbers here as the only time this would be used is to verify one of our internal enums that is always (as far as I know) initialized with strings.

schaable (changes requested):
I left a couple of comments. There are two things I'm not completely sure about:

 - Doing the validations manually vs using Zod. I think this might be achievable with Zod, and it would probably look nicer. However, I haven't used Zod before, so I could be wrong.
 - There are some duplicate validations between this PR and tasks. I'm inclined to say we should remove them either here or in the task manager/builders, unless these validations don't run in some scenarios.

v-next/core/src/internal/config-validation.ts
```diff
@@ -0,0 +1,537 @@
...
+    }
+
+    if (typeof option.description !== "string") {
+      validationErrors.push({
+        path: [...path, name, "description"],
+        message: "option description must be a string",
+      });
+    }
+
+    if (!isOptionDefinition(option)) {
```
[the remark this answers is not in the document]
zoeyTM: I might be mistaken, but I don't think we do know that `option` is an object before this check.

zoeyTM:
pushed up a commit addressing most of the comments. i'm working on a further update to validate via zod instead, but it won't be ready until tomorrow.

zoeyTM:
After discussing pros/cons with Pato, we agreed that manual validation was the preferred choice as opposed to full config validation via zod.

schaable (approved):

PR #5532, opened 2024-07-17
Refactor argument validations

schaable:


v-next/core/src/internal/tasks/resolved-task.ts
```diff
@@ -112,11 +112,16 @@ export class ResolvedTask implements Task {
       if (isPositional) {
         this.#validateRequiredArgument(argumentDefinition, value);
       }
-      this.#validateArgumentType(
-        argumentDefinition,
-        value,
-        isPositional && argumentDefinition.isVariadic,
-      );
+
+      if (value !== undefined) {
```
kanej, line 116: If I am reading correctly, the previous check inside `#validateArgumentType` was:

```js
// skip type validation for optional arguments with undefined value
if (value === undefined && argument.defaultValue !== undefined) {
  return;
}
```

Are we safe in ignored the `argument.defaultValue` here?

schaable: I don't think it's needed. We're validating the required arguments first, meaning at this point the argument is either optional or has a value. If the argument is optional, the default value was already validated in the builder/task-manager. That leaves us with the simplified check `value !== undefined`, meaning we validate all arguments with a defined value.

kanej (approved):
LGTM!

PR #5540, opened 2024-07-23
Clean task

schaable:
This PR requires https://github.com/NomicFoundation/hardhat/pull/5527 ~and https://github.com/NomicFoundation/hardhat/pull/5517.~

The logic to create the `cache` and `artifacts` paths is missing, as it's outside the scope of this issue. Because of this, the tests are somewhat "manual" – we handle the folders manually to simulate different scenarios. However, this should be acceptable.

~The build is currently failing because it needs [PR #5517](https://github.com/NomicFoundation/hardhat/pull/5517) and some additional code (e.g., adding the `cache` and `artifacts` properties in `v-next/core/test/internal/hook-manager.ts`).~

The task might require additional steps, such as calling `artifacts.clearCache?.()`, which depends on how the new build system works.

### DIFF
--- a/v-next/core/src/internal/hre.ts
+++ b/v-next/core/src/internal/hre.ts
@@ -196,6 +196,8 @@ async function resolveUserConfig(
     tasks: config.tasks ?? [],
     paths: {
       root: projectRoot,
+      cache: config.paths?.cache ?? "", // TODO: resolve cache path
+      artifacts: config.paths?.artifacts ?? "", // TODO: resolve artifacts path
     },
   };
 
--- a/v-next/hardhat/src/internal/builtin-plugins/clean/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/clean/task-action.ts
@@ -1,0 +1,26 @@
+import type { NewTaskActionFunction } from "@ignored/hardhat-vnext-core/types/tasks";
+
+import { getCacheDir } from "@ignored/hardhat-vnext-core/global-dir";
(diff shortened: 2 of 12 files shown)


v-next/core/src/internal/hre.ts
```diff
@@ -196,6 +196,8 @@ async function resolveUserConfig(
     tasks: config.tasks ?? [],
     paths: {
       root: projectRoot,
+      cache: config.paths?.cache ?? "", // TODO: resolve cache path
```
alcuadrado, line 199: Do we have a task to properly implement this resolution?

v-next/hardhat/src/internal/builtin-plugins/clean/task-action.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (global) {
+    const globalCacheDir = await getCacheDir();
+    await emptyDir(globalCacheDir);
+  }
+
+  await emptyDir(config.paths.cache);
+  await remove(config.paths.artifacts);
+
+  // TODO: Uncomment this when we have a clearCache method in the artifacts module
+  // artifacts.clearCache?.();
```
alcuadrado, line 23: I think this can go away. We won't have that feature.

alcuadrado (approved):

v-next/core/src/internal/hre.ts
alcuadrado: Maybe these things should be defined in `hardhat`, not core

PR #5541, opened 2024-07-23
Basic debug logging

### LINKED ISSUE #5363: Add `debug` traces
Support the DEBUG environment variable for tasks.

We should document guidelines on the usage of the debug library within the built-in Hardhat tasks.

We should then cover the existing tasks (i.e. `help`, `run`, `console` etc).

```sh
> DEBUG=* hardhat --help
  hardhat:core:vars:varsManager Creating a new instance of VarsManager +0ms
  hardhat:core:vars:varsManager Loading ENV variables if any +1ms
  hardhat:core:config Loading Hardhat config from /Users/fred/projects/nomiclabs/temp-sample/hardhat.config.js +0ms
  hardhat:core:hre Creating HardhatRuntimeEnvironment +0ms
  hardhat:core:global-dir Looking up Client Id at /Users/fred/Library/Application Support/hardhat-nodejs/analytics.json +0ms
  hardhat:core:global-dir Client Id found: 6e29ffc9-9450-4a6d-8a46-4f1cce03d837 +0ms
  hardhat:core:analytics Sending hit for task +0ms
```

zoeyTM:
- added basic debug logging throughout the main CLI function
- added a `withDebugLogs` decorator to core to easily enable debug logging to any class method
  - added this decorator to some class methods that I thought would be helpful, but it can always be added to others in the future as well

fixes #5363 

I would add a screenshot of the output, but there's A Lot. Recommend pulling the branch and testing inside `v-next/example-project`

### DIFF
--- a/v-next/hardhat/src/internal/cli/main.ts
+++ b/v-next/hardhat/src/internal/cli/main.ts
@@ -28,6 +28,7 @@ import {
 import { isCi } from "@ignored/hardhat-vnext-utils/ci";
 import { getRealPath } from "@ignored/hardhat-vnext-utils/fs";
 import { kebabToCamelCase } from "@ignored/hardhat-vnext-utils/string";
+import debug from "debug";
 
 import { resolveHardhatConfigPath } from "../../config.js";
 import { createHardhatRuntimeEnvironment } from "../../hre.js";
@@ -47,7 +48,11 @@ export async function main(
   cliArguments: string[],
   print: (message: string) => void = console.log,
 ): Promise<void> {
+  const log = debug("hardhat:core:cli:main");
+
   let builtinGlobalOptions;
+
+  log("Hardhat CLI started");
 
   try {
     const usedCliArguments: boolean[] = new Array(cliArguments.length).fill(
@@ -59,6 +64,8 @@ export async function main(
       usedCliArguments,
     );
 
+    log("Parsed builtin global options");
+
     if (builtinGlobalOptions.version) {
       return await printVersionMessage(print);
     }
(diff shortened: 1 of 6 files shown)


ChristopherDedominici (approved):
LGTM (I saw the comment about the issue with tsx)

I love the idea of the decorator! 
One suggestion:  what about a class level decorator that will automatically log for all the methods of the class? So you only have to add it at the beginning instead of every method

alcuadrado:
I've been testing this, and I think this is too verbose to be useful. I believe the decorator approach will lead to printing too much data, so I think it's better to remove it and manually log what we find useful.

For example, I think logging these things is useful, but they could be much more succinct:
* running hooks in the hook manager, which one is run and how.
* The HRE being created, maybe not even with the config. But knowing that we are creating an HRE can be helpful to debug potential errors (e.g. creating multiple of them by accident)
* Running a task, but maybe not creating/inserting/fetching them (as that's really verbose)

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -45,18 +46,26 @@ export async function main(
...
   let builtinGlobalOptions;
 
+  log("Hardhat CLI started with arguments:", cliArguments);
+
   try {
     const usedCliArguments: boolean[] = new Array(cliArguments.length).fill(
       false,
     );
 
+    log("Parsing builtin global options");
```
alcuadrado, line 60: I think printing just the results would probably be enough in most cases, and make the output less verbose/more helpful.

alcuadrado: I mean in general. I noticed a pattern of "doing this", "this is its result", which I do a ton for manual debugging, but can be too verbose.

alcuadrado (changes requested):
Just left a few comments

alcuadrado (approved):

Issue #5057, opened 2024-04-02
npx hardhat run check: raise error, exit code stays 0 


codekoriko:
### Version of Hardhat

2.22.2

### What happened?

I'm using the [hardhat-solhint](https://hardhat.org/hardhat-runner/plugins/nomiclabs-hardhat-solhint) plugin, when the litting process raise error, the exit code stays the 0, therefor command chaining with && don't break, which is a problem

### Minimal reproduction steps

Running the linter as mentioned in the doc: 
```shell
npx hardhat check && echo 'Error Code: '$? 
```

An error arise during the linting but _Exit Code_ stays 0
```shell
    8:1  error    Unexpected import of console file
             no-console
    8:1  warning  global import of path hardhat/console.sol is not allowed. Specify names to import individually or bind all exports of the module into a name (import "path" as Name)
             no-global-import
   31:5  warning  Explicitly mark visibility in function (Set ignoreConstructors to true if using solidity >=0.7.0)
             func-visibility
   56:9  warning  Use Custom Errors instead of require statements
             custom-errors
   74:9  warning  Use Custom Errors instead of require statements
             custom-errors
   76:9  warning  Use Custom Errors instead of require statements
             custom-errors
   94:9  warning  Use Custom Errors instead of require statements
             custom-errors
  114:9  warning  Use Custom Errors instead of require statements
             custom-errors

✖ 40 problems (1 error, 39 warnings)

Error Code: 0
```

kanej:
I was able to reproduce this locally, by taking a new typescript sample project, adding the solhint plugin, running `solhint --init` and updating the config to:

```json
{
  "extends": "solhint:recommended"
}
```

On running `npx hardhat check` we don't get a non-zero exit code:

![image](https://github.com/NomicFoundation/hardhat/assets/24030/4c51e713-a6b6-402b-86e3-b03a430eb057)

madlabman:
Because of this I think https://github.com/protofire/solhint/pull/554

madlabman:
Opened the issue https://github.com/protofire/solhint/issues/572

kanej:
Thanks @madlabman, we will keep this open to track the issue coming in on the `hardhat-solhint` plugin.

AlissonRS:
@kanej just a heads up this was fixed in Solhint 5.0.1 :rocket:

kanej:
[...]

PR #5556, opened 2024-07-29
fix: exit check task with exit code 1 when solhint raises errors

galargh:

---

Resolves https://github.com/NomicFoundation/hardhat/issues/5057

This PR modifies the `hardhat-solhint` plugin.

It changes the `check` task behaviour. After the changes, when `solhint` reports **any** errors (note, this does not apply to warnings), the task exists with error code 1.

The updated behaviour is in line with how `solhint` behaves as a standalone CLI tool.

Please note that this is, currently, a breaking change since it's modifying solhint plugin/check task behaviour. I will update the changesets and mark the PR as "ready for review", once we align on the open questions I outline below.

#### Open Questions
1. ~~Should we update `solhint` to the most recent version `v5.0.2`?~~ Yes
2. ~~Should we control the newly added behaviour via configuration? If so, we could also support failing the check task when a threshold of warnings is crossed. In solhint, this is controlled via command line options.~~ No (not yet, at least)

### DIFF
--- a/.changeset/breezy-chairs-relax.md
+++ b/.changeset/breezy-chairs-relax.md
@@ -2,8 +2,4 @@
 "@nomiclabs/hardhat-solhint": major
 ---
 
-Ensured the check task exists with exit code 1 when solhint raises any errors and updated solhint dependency to [v5.0.2](https://github.com/protofire/solhint/releases/tag/v5.0.2)
-
-## ⚠️ Breaking Change ⚠️
-
-This release introduces a breaking change. Previously, the check task would exit with exit code 0 even if solhint raised some errors. Starting with this version, the check task will exit with exit code 1 if any errors are raised and 0 otherwise.
+Ensured the check task exits with exit code 1 when solhint raises any errors; this is a breaking change since the check task would previously always exit with exit code 0
--- a/.changeset/breezy-chairs-relax.md
+++ b/.changeset/breezy-chairs-relax.md
@@ -1,0 +1,9 @@
+---
+"@nomiclabs/hardhat-solhint": major
+---
+
+Ensured the check task exists with exit code 1 when solhint raises any errors and updated solhint dependency to [v5.0.2](https://github.com/protofire/solhint/releases/tag/v5.0.2)
+
+## ⚠️ Breaking Change ⚠️
+
+This release introduces a breaking change. Previously, the check task would exit with exit code 0 even if solhint raised some errors. Starting with this version, the check task will exit with exit code 1 if any errors are raised and 0 otherwise.
--- a/packages/hardhat-solhint/src/index.ts
+++ b/packages/hardhat-solhint/src/index.ts
@@ -127,4 +127,15 @@ task("check", async (_, { run }, runSuper) => {
   const reports = await run("hardhat-solhint:run-solhint");
 
   printReport(reports);
+
+  const errorsCount = reports.reduce(
+    (acc: number, i: { errorCount: number }) => {
+      return acc + i.errorCount;
+    },
+    0
+  );
+
+  if (errorsCount > 0) {
+    process.exit(1);
+  }
 });
--- a/packages/hardhat-solhint/src/index.ts
+++ b/packages/hardhat-solhint/src/index.ts
@@ -136,6 +136,7 @@ task("check", async (_, { run }, runSuper) => {
   );
 
   if (errorsCount > 0) {
-    process.exit(1);
+    process.exitCode = 1;
+    return;
   }
 });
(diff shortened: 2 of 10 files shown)


kanej:
I think we bump to version `v5.0.2`. We are taking the hit of a major update due to the breaking cli behaviour change; so lets move to the latest version at the same time.

I think we leave off of replicating the cli behaviour around warnings. This can be added later if there is demand.

packages/hardhat-solhint/test/fixture-projects/no-errors-project/contracts/Greeter.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity ^0.8.0;
```
kanej, line 1: Do we know why the `0.8.0` version was needed rather than the other examples that use `0.5`?

galargh: This is a default minimal compiler version set by solhint (see: https://github.com/protofire/solhint/blob/v5.0.2/docs/rules/security/compiler-version.md#compiler-version)

```suggestion
// NOTE: This is a default minimal compiler version set by solhint (see: https://github.com/protofire/solhint/blob/v5.0.2/docs/rules/security/compiler-version.md#compiler-version)
pragma solidity ^0.8.0;
```

.changeset/breezy-chairs-relax.md
```diff
@@ -0,0 +1,9 @@
+---
+"@nomiclabs/hardhat-solhint": major
+---
+
+Ensured the check task exists with exit code 1 when solhint raises any errors and updated solhint dependency to [v5.0.2](https://github.com/protofire/solhint/releases/tag/v5.0.2)
```
kanej, line 5: Can we split this into two separate changesets.

So one for `Ensured the check task exists with exit code 1 when solhint raises any errors` and the other for update to v5.

packages/hardhat-solhint/test/tests.ts
```diff
@@ -36,11 +36,28 @@ describe("Solhint plugin", function () {
       );
     });
 
-    it("should run the check task without throwing an error", async function () {
+    it("should run the check task and exit", async function () {
       const consoleLogStub = sinon.stub(console, "log");
+      const processExitStub = sinon.stub(process, "exit");
       await this.env.run("check");
       assert.isTrue(consoleLogStub.calledOnce);
+      assert.isTrue(processExitStub.calledOnceWith(1));
```
kanej, line 44: Does process exit need to be a spy rather than a stub. Or put another way, in this test will the process exit actually be called? Is that relevant?
Or is it the case that we don't expect any code after the error count conditional so we don't need to actually trigger the `process.exit` call.

kanej: Lets not worry about this. It does mean you need to grok the context of the test - `process.exit` isn't being called and that has implications for follow on `check` tasks; but I don't think the trade off here in effort is worth it.

packages/hardhat-solhint/src/index.ts
```diff
@@ -127,4 +127,15 @@ task("check", async (_, { run }, runSuper) => {
...
   printReport(reports);
+
+  const errorsCount = reports.reduce(
+    (acc: number, i: { errorCount: number }) => {
+      return acc + i.errorCount;
+    },
+    0
+  );
+
+  if (errorsCount > 0) {
```
kanej, line 138: I think this is legitimate. My expectation is for a set of linters being run is to fail fast. Our plugin system does not encode patterns for parallelism, so we should assume linear running of check tasks.
Basically I think we should exit out after this task if there are errors.

packages/hardhat-solhint/test/tests.ts
galargh: Would the ask here be to basically split the test case into 2? One that checks whether we print out the report and a separate one that checks if we properly exit when we find errors?

packages/hardhat-solhint/src/index.ts
galargh: Thanks! I was just about to ask on the dev channel. Should I follow up on it there anyway or do you think it's OK for us to proceed?

kanej (changes requested):
This PR is looking good. I left two requests:

1. splitting the changeset into two parts
2. updating the exit technique if there are errors

packages/hardhat-solhint/test/tests.ts
kanej: I think we leave this, but in an ideal world I think having two test cases:

1. two check tasks, confirm the second one runs if there are no solhint errors
2. two check tasks, confirm the second one doesn't run if there are solhint errors

The stubbing of `process.exit` would fail here; the stub would be called but it wouldn't "exit", it would continue on.

However, I don't think this particular case justifies the effort of figuring out how to safely `process.exit` in a test. What do you think?

kanej (approved):
LGTM!

packages/hardhat-solhint/test/tests.ts
galargh: Yes, that makes sense and I do agree in this particular case. We can deal with it the next time we encounter a need for a similar check.

PR #5557, opened 2024-07-29
Refactor telemetry-consent file and add tests

ChristopherDedominici:


v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -67,8 +67,7 @@ export async function main(
       return await initHardhat();
     }
 
-    // TODO: the consent will be enabled in the other PRs related to telemetry
-    const _telemetryConsent = await getTelemetryConsent();
+    await ensureTelemetryConsent();
```
ChristopherDedominici, line 70: Based on the new logic, the telemetry consent will be read directly when needed. 
This function purpose is to ensure that telemetry-consent is registered, so no value needs to be returned

schaable (approved):

PR #5559, opened 2024-07-31
ci(v3): update GitHub actions and review workflow triggers

galargh:

---

This PR is a subset of #5301 and, unlike the original, it targets the `v-next` branch.

In this PR, I update GitHub actions to the latest versions. I reviewed the changelogs to ensure there are no breaking changes announcements that would affect us.

I also reviewed the workflow triggers to ensure we don't use invalid syntax (e.g. `$default-branch`) and that we use them consistently.

I will propose improvements to the caching setup and work deduplication in separate PRs to minimise the diffs.

### DIFF
--- a/.github/workflows/v-next-changesets-release.yml
+++ b/.github/workflows/v-next-changesets-release.yml
@@ -1,4 +1,5 @@
 name: Changesets
+
 on:
   push:
     branches:
--- a/.github/workflows/v-next-ci.yml
+++ b/.github/workflows/v-next-ci.yml
@@ -66,7 +66,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install pnpm
-        uses: pnpm/action-setup@v3
+        uses: pnpm/action-setup@v4
         with:
           version: 9
       - uses: actions/setup-node@v4
@@ -99,7 +99,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Install pnpm
-        uses: pnpm/action-setup@v3
+        uses: pnpm/action-setup@v4
         with:
           version: 9
       - uses: actions/setup-node@v4
--- a/.github/workflows/v-next-ci.yml
+++ b/.github/workflows/v-next-ci.yml
@@ -2,11 +2,15 @@ name: vNext CI
 
 on:
   push:
-    branches: v-next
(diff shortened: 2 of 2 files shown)


alcuadrado:
> 1. I see the E2E tests currently do not pass against the `v-next` branch. Should I restore the filter that runs them only on pull requests to `main`?

Yes, we don't expect the e2e tests to run on `v-next`, as they aren't relevant, and they are quite slow.

> 2. Would the expected flow for changes like this one be to first merge them into `v-next` and then cherry-pick applicable subset onto `main`?

I think for this kind of change, this makes sense. We don't have many of situations like this, luckily.

alcuadrado (approved):

alcuadrado:
> > 2. Would the expected flow for changes like this one be to first merge them into `v-next` and then cherry-pick applicable subset onto `main`?
> 
> I think for this kind of change, this makes sense. We don't have many of situations like this, luckily.

Actually, after looking at your other PRs, I'd say that it depends on what we are targeting. If the change is meant to go to `main`, I'd do a PR into `main`.

We periodically merge `main` into `v-next`.

galargh:
> > > 2. Would the expected flow for changes like this one be to first merge them into `v-next` and then cherry-pick applicable subset onto `main`?
> > 
> > 
> > I think for this kind of change, this makes sense. We don't have many of situations like this, luckily.
> 
> Actually, after looking at your other PRs, I'd say that it depends on what we are targeting. If the change is meant to go to `main`, I'd do a PR into `main`.
> 
> We periodically merge `main` into `v-next`.

That's fantastic! In that case, I split this PR into this one which targets `v-next` and only modifies `v-next-*` workflows and https://github.com/NomicFoundation/hardhat/pull/5566 which does the same job of updating actions versions and triggers but for the v2 packages and targets `main`. I followed the same approach with other ci PRs too.

> > 1. I see the E2E tests currently do not pass against the `v-next` branch. Should I restore the filter that runs them only on pull requests to `main`?
> 
> Yes, we don't expect the e2e tests to run on `v-next`, as they aren't relevant, and they are quite slow.

Thanks! I restored the `branches` filter for the `pull_request` trigger of the e2e tests workflow in https://github.com/NomicFoundation/hardhat/pull/5566

PR #5562, opened 2024-08-01
ci(v2): remove duplicated lint jobs/steps

galargh:

---

This PR is a subset of https://github.com/NomicFoundation/hardhat/pull/5301.

In this PR, I minimise the number of places where we perform linting to limit duplication. In particular:
1. I changed the trigger of the `Lint` workflow to `pull_request_target` and added a second job that comments on a PR when needed.
2. I removed `Comment on lint failure` workflow because it is duplicated by the `Lint` workflow after the changes.
3. I removed the linting step from the `Check Docs Site` workflow because it is duplicated in the `Lint` workflow.


.github/workflows/lint.yml
```diff
@@ -19,8 +21,17 @@ jobs:
   lint:
     name: Lint
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - uses: actions/checkout@v4
+        with:
```
alcuadrado, line 28: Out of curiosity: what's the deal behind this double checkout? 

Is the first one for pushes to `main` and the other one checking out the merge commit of a pr?

Maybe leaving a comment would be good for future readers.

galargh: I missed an `if` here. The first checkout should only be executed if we're not running in the pull request context. Thanks for catching this 🙇

alcuadrado (approved):

.github/workflows/lint.yml
galargh: FYI, after resolving the conflicts I had to restore 2 checkouts when we're running on `pull_request_target`. This is so that we can run lint on the code from the pull request but get the code of the `setup-env` action from the base branch.

I added the following comment above the first checkout so that this bit of context is not lost:
```
      # NOTE: When running on a pull_request_target, we're checking out the repo
      # twice. The first checkout is for the base branch of the PR. The second,
      # for the merge commit. We do this to be able to use the `setup-env`
      # action that's on the base branch.
 ```

PR #5570, opened 2024-08-03
Port sentry logic

ChristopherDedominici:


v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -0,0 +1,379 @@
...
+      anonymizedFilename: parts.slice(nodeModulesIndex).join(path.sep),
+      anonymizeContent: false,
+    };
+  }
+
+  public anonymizeErrorMessage(errorMessage: string): string {
+    errorMessage = this.#anonymizeMnemonic(errorMessage);
+
+    // the \\ before path.sep is necessary for this to work on windows
+    const pathRegex = /\S+[\/\\]\S+/g;
```
alcuadrado, line 126: Note: This is a bit different, it accepts `\` or `/` in any platform. Previously it used `path.sep`.

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -0,0 +1,379 @@
...
+      if (err instanceof PackageJsonNotFoundError) {
+        return null;
+      }
+
+      throw err;
+    }
+  }
+
+  #isHardhatFile(filename: string): boolean {
+    const nomicFoundationPath = path.join("node_modules", "@nomicfoundation");
```
alcuadrado, line 199: Note: this has also changed

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -0,0 +1,379 @@
+import type { Event, Exception, StackFrame, Stacktrace } from "@sentry/node";
+
+import * as path from "node:path";
+
+import {
```
alcuadrado, line 5: Maybe add a comment explaining that this is run in a subprocess so it's fine to load them here, in the top-level

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -0,0 +1,379 @@
...
+import * as czech from "ethereum-cryptography/bip39/wordlists/czech.js";
+import * as english from "ethereum-cryptography/bip39/wordlists/english.js";
+import * as french from "ethereum-cryptography/bip39/wordlists/french.js";
+import * as italian from "ethereum-cryptography/bip39/wordlists/italian.js";
+import * as japanese from "ethereum-cryptography/bip39/wordlists/japanese.js";
+import * as korean from "ethereum-cryptography/bip39/wordlists/korean.js";
+import * as simplifiedChinese from "ethereum-cryptography/bip39/wordlists/simplified-chinese.js";
+import * as SPANISH from "ethereum-cryptography/bip39/wordlists/spanish.js";
+import * as traditionalChinese from "ethereum-cryptography/bip39/wordlists/traditional-chinese.js";
+import { either } from "fp-ts";
```
alcuadrado, line 18: Note to myself: think about how to remove this dependency

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -0,0 +1,120 @@
...
+    this.#telemetryEnabled = telemetryAllowed;
+  }
+
+  public static async getInstance(): Promise<Reporter> {
+    if (this.#instance !== undefined) {
+      return this.#instance;
+    }
+
+    const telemetryAllowed = await isTelemetryAllowed();
+    this.#instance = new Reporter(telemetryAllowed);
```
alcuadrado, line 46: Maybe we can return early here if not allowed, as nothing else is used

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    const { Integrations, init, setExtra } = await import("@sentry/node");
+
+    const linkedErrorsIntegration = new Integrations.LinkedErrors({
+      key: "parent",
```
alcuadrado, line 51: I think this is `cause` now, because in v3 we follow the new js standard

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -0,0 +1,120 @@
...
+    verbose: boolean = false,
+    configPath: string = "",
+  ): Promise<boolean> {
+    if (!(await this.canSendTelemetry(error))) {
+      return false;
+    }
+
+    const { captureException, setExtra } = await import("@sentry/node");
+
+    setExtra("verbose", verbose);
```
alcuadrado, line 84: If not used we can remove it. It was there just to allow the subprocess to `log()` things, for debugging

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    captureException(error);
+
+    // TODO: where to run it?
+    // await this.#sentry.flush(1000);
+
+    return true;
+  }
+
+  private async canSendTelemetry(error: Error): Promise<boolean> {
```
alcuadrado, line 95: Maybe something like `shouldBeReported`?

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -0,0 +1,120 @@
...
+      !error.descriptor.shouldBeReported
+    ) {
+      return false;
+    }
+
+    if (HardhatPluginError.isHardhatPluginError(error)) {
+      // Don't log errors from third-party plugins
+      return false;
+    }
+
```
alcuadrado, line 111: We should probably be a bit smarter here, and NOT send the errors that we know aren't ours (i.e. like a blocklist). For example, errors sent by ethers.js.

This can be a separate task, but we should do it.

alcuadrado: What I mean is checking the stack trace, and if it comes from a well-known file/package that we don't want to report, we just return false.

Example:

```
Error: ERROR processing <user-file>
Error: invalid address or index (argument="addressOrIndex", value="<SignerWithAddress xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx>", code=INVALID_ARGUMENT, version=provide...
  File "node_modules/@ethersproject/logger/src.ts/index.ts", line 269, col 28, in Logger.makeError
    const error: any = new Error(message);
  File "node_modules/@ethersproject/logger/src.ts/index.ts", line 281, col 20, in Logger.throwError
    throw this.makeError(message, code, params);
  File "node_modules/@ethersproject/logger/src.ts/index.ts", line 285, col 21, in Logger.throwArgumentError
    return this.throwError(message, Logger.errors.INVALID_ARGUMENT, {
  File "node_modules/@ethersproject/providers/src.ts/json-rpc-provider.ts", line 183, col 20, in new JsonRpcSigner
    logger.throwArgumentError("invalid address or index", "addressOrIndex", addressOrIndex);
  File "node_modules/@ethersproject/providers/src.ts/json-rpc-provider.ts", line 489, col 16, in EthersProviderWrapper.JsonRpcProvider.getSigner
    return new JsonRpcSigner(_constructorGuard, this, addressOrIndex);
  File "node_modules/@nomiclabs/hardhat-ethers/src/internal/helpers.ts", line 62, col 38, in getSigner
    const signer = hre.ethers.provider.getSigner(address);
  File "<user-file>", line 95, col 5, in processTicksAndRejections
  File "<user-file>", line 275, col 32, in async createAndInitialize_UniV2_PoolForPairs
  File "<user-file>", line 411, col 13, in async processPairsWithParams
```

v-next/hardhat/src/internal/cli/telemetry/sentry/send-to-sentry.ts
```diff
@@ -0,0 +1,80 @@
+import type { Event } from "@sentry/node";
```
alcuadrado, line 1: Maybe moving all the subprocess files to a `/subprocess`  folder can be easier to understand.

v-next/hardhat/src/internal/cli/telemetry/sentry/transport.ts
```diff
@@ -0,0 +1,49 @@
...
+      const { configPath } = extra;
+
+      // Don't send user's full config path for privacy reasons
+      delete event.extra?.configPath;
+      // We don't care about the verbose setting
+      delete event.extra?.verbose;
+
+      const serializedEvent = JSON.stringify(event);
+
+      const fileExt =
```
alcuadrado, line 22: Explaining this would be helpful

v-next/hardhat/src/internal/cli/telemetry/sentry/send-to-sentry.ts
```diff
@@ -0,0 +1,80 @@
...
+    );
+    return;
+  }
+
+  try {
+    const anonymizer = new Anonymizer(configPath);
+    const anonymizedEvent = await anonymizer.anonymize(event);
+
+    if (anonymizedEvent.isRight()) {
+      if (anonymizer.raisedByHardhat(anonymizedEvent.value)) {
```
alcuadrado, line 43: This could be in the reporter, and I'm not convinced that this needs to be part of the anonymizer

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -0,0 +1,379 @@
...
+    const hexRegex = /(0x)?[0-9A-Fa-f]{20,}/g;
+
+    return errorMessage
+      .replace(pathRegex, ANONYMIZED_FILE)
+      .replace(fileRegex, ANONYMIZED_FILE)
+      .replace(hexRegex, (match) => match.replace(/./g, "x"));
+  }
+
+  public raisedByHardhat(event: Event): boolean {
+    const exceptions = event?.exception?.values;
```
alcuadrado, line 141: If we also add a check for the name of the error, and return true if it's 'HardhatError'? it'd simplify testing, so that we don't need to commit a `node_modules/`

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -0,0 +1,120 @@
...
+    const { captureException, setExtra } = await import("@sentry/node");
+
+    setExtra("verbose", verbose);
+    setExtra("configPath", configPath);
+
+    captureException(error);
+
+    // TODO: where to run it?
+    // await this.#sentry.flush(1000);
```
alcuadrado, line 90: can you link the docs regarding this?

v-next/hardhat/src/internal/cli/telemetry/sentry/send-to-sentry.ts
alcuadrado: we also discussed merging them

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5570/commits/480589a11adfeba18855a8be0679e4241aa9f8d8) with also another comment explanation

v-next/hardhat/src/internal/cli/telemetry/sentry/transport.ts
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5570/commits/480589a11adfeba18855a8be0679e4241aa9f8d8) with also another comment explanation

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5570/commits/ce56aa00718d650947cc78323c9eb975c21329bf)

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5570/commits/bdb9042d15e6896415bbc12393c941cd2d136e08)

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5570/commits/d28340efda41933ad4a11b11b14405d01c68a51e)

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5570/commits/4f70a7220b616ed256699ec834567324b5ab0565)

v-next/hardhat/src/internal/cli/telemetry/sentry/send-to-sentry.ts
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5570/commits/dea358cc46629e6f707bc63e371c65ecce3e5bcc), I merged the file

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
ChristopherDedominici: As discussed in the messages, I'll leave it like this because checking for the error name will handle the HardhatErrors but not the "normal JS errors"

v-next/hardhat/src/internal/cli/telemetry/sentry/send-to-sentry.ts
ChristopherDedominici: As discussed in the messages, we will keep it here because the errors are stringified by Sentry in an easy-to-analyze format. Hence, it's easy to check the file paths, etc.

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/5570/commits/bed832cd1836c4fe16d1f965686e6e826884a039). But let's discuss if we want to use a different approach

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
ChristopherDedominici: The Sentry docs here are not very clear on the usage of flush (or I did not find the right ones). Is it possible to send events right away without queuing them? Not sure. 

- [This link](https://stackoverflow.com/questions/67241239/sentry-capturemessage-without-flush) mentions that "flush" should NOT be used after every call (but it's for go).

- from [this doc](https://docs.sentry.io/platforms/go/guides/http/configuration/transports/), also for go, it looks like that if the transporter is sync, then the event will be sent in a sync way. But it's a guess. Also, does it apply to NodeJS too?

- [link to JS custom transporter doc](https://docs.sentry.io/platforms/javascript/configuration/transports/)

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -0,0 +1,401 @@
...
+      if (err instanceof PackageJsonNotFoundError) {
+        return null;
+      }
+
+      throw err;
+    }
+  }
+
+  #errorRaisedByPackageToIgnore(filename: string): boolean {
+    const pkgsToIgnore: string[] = [
```
ChristopherDedominici, line 205: Update with real packages

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -0,0 +1,401 @@
...
+  constructor(configPath?: string) {
+    this.#configPath = configPath;
+  }
+
+  /**
+   * Given a sentry serialized exception
+   * (https://develop.sentry.dev/sdk/event-payloads/exception/), return an
+   * anonymized version of the event.
+   */
+  public async anonymize(event: any): Promise<either.Either<string, Event>> {
```
alcuadrado, line 44: Let's replace this with something like:

```ts
type AnonymizeResult = {success: true, Event} | {success: false, error:string}
```

so that we don't add `fp-ts` as a dependency.

alcuadrado (changes requested):
I left a small comment about fp-ts

v-next/hardhat/package.json
```diff
@@ -78,9 +78,12 @@
     "@ignored/hardhat-vnext-errors": "workspace:^3.0.0-next.2",
     "@ignored/hardhat-vnext-utils": "workspace:^3.0.0-next.2",
     "@ignored/hardhat-vnext-zod-utils": "workspace:^3.0.0-next.2",
+    "@sentry/node": "^5.18.1",
     "chalk": "^5.3.0",
     "debug": "^4.1.1",
     "enquirer": "^2.3.0",
+    "ethereum-cryptography": "1.1.1",
+    "fp-ts": "1.19.3",
```
kanej, line 86: I think we should remove this. We are using it for the `either` result, we should just handle this in native typescript.

kanej:
I was able to get Sentry exceptions through on the test environment with manual testing.

It did throw up that our anonymizing is not setup for the `@ignored` org and namespace, nor will it work with the `./v-next/example-project` (the paths don't meet the anonymizer expectations).

I think we deal with this issue when we add Sentry to the CLI's uncaught exception handler.

ChristopherDedominici:
> I left a small comment about fp-ts

Done [here](https://github.com/NomicFoundation/hardhat/pull/5570/commits/46ef23102f79850c04307cefe5d0cd4dd22668d3)

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -17,13 +17,16 @@ import * as korean from "ethereum-cryptography/bip39/wordlists/korean.js";
...
 import * as SPANISH from "ethereum-cryptography/bip39/wordlists/spanish.js";
 import * as traditionalChinese from "ethereum-cryptography/bip39/wordlists/traditional-chinese.js";
-import { either } from "fp-ts";
 
 interface WordMatch {
   index: number;
   word: string;
 }
 
+type AnonymizeResult =
```
kanej, line 26: We should export this type as its part of the Anonymizer API.

ChristopherDedominici: Done

alcuadrado (approved):

PR #5578, opened 2024-08-06
chore: do not send hardhat_setLedgerOutputEnabled over http

### LINKED ISSUE #5406: "hardhat_setLedgerOutputEnabled - Method not supported" message when deploying a contract with Hardhat Ignition
### Version of Hardhat

2.22.5

### What happened?

There is a "hardhat_setLedgerOutputEnabled - Method not supported" message on the "Hardhat node" console when deploying a contract with Hardhat Ignition.

Create a new contract `contracts/Bank.sol`

```solidity
pragma solidity 0.8.24;

///@title A simple Bank contract which allows users to deposit and then, withdraw ethers
///@author Ben BK 

contract Bank {

    struct Account {
        uint balance;
        uint lastDeposit;
    }

    mapping(address => Account) accounts;

    event etherDeposited(address indexed account, uint amount);
    event etherWithdrawed(address indexed account, uint amount);

    ///@notice Allows the user to withdraw ethers from the smart contract 
    ///@param _amount The amount of ethers the user wants to withdraw
    function withdraw(uint _amount) external {
        require(accounts[msg.sender].balance >= _amount, "Not enough funds");
        accounts[msg.sender].balance -= _amount;
        (bool received, ) = msg.sender.call{value: _amount}("");
        require(received, "An error occured");
        emit etherWithdrawed(msg.sender, _amount);
    }

    ///@notice Allows a user to deposit ethers on the smart contract
    function deposit() external payable {
        require(msg.value > 0, "Not enough funds deposited");
        accounts[msg.sender].balance += msg.value;
        accounts[msg.sender].lastDeposit = block.timestamp;
        emit etherDeposited(msg.sender, msg.value);
    }

    ///@notice Allows to get the amount of ethers the user has on the smart contract 
    ///@return The amount of ethers the user has on the smart contract
    function getBalanceOfUser() external view returns(uint) {
        return accounts[msg.sender].balance;
    }
}
```

Create a new ignition module `ignition/modules/Bank.js`

```javascript
const { buildModule } = require("@nomicfoundation/hardhat-ignition/modules");

module.exports = buildModule("BankModule", (m) => {

  const bank = m.contract("Bank");

  return { bank };
});
```

`hardhat.config.js` file : 
```javascript 
require("@nomicfoundation/hardhat-toolbox");

/** @type import('hardhat/config').HardhatUserConfig */
module.exports = {
  solidity: "0.8.24",
};
```

`package.json` : 
```json
{
  "devDependencies": {
    "@nomicfoundation/hardhat-chai-matchers": "^2.0.0",
    "@nomicfoundation/hardhat-ethers": "^3.0.0",
    "@nomicfoundation/hardhat-ignition": "^0.15.0",
    "@nomicfoundation/hardhat-ignition-ethers": "^0.15.0",
    "@nomicfoundation/hardhat-network-helpers": "^1.0.0",
    "@nomicfoundation/hardhat-toolbox": "^5.0.0",
    "@nomicfoundation/hardhat-verify": "^2.0.0",
    "@typechain/ethers-v6": "^0.5.0",
    "@typechain/hardhat": "^9.0.0",
    "chai": "^4.2.0",
    "ethers": "^6.4.0",
    "hardhat": "^2.22.5",
    "hardhat-gas-reporter": "^1.0.8",
    "solidity-coverage": "^0.8.0",
    "typechain": "^8.3.0"
  }
}

```

**Launch the Hardhat local node with :** `yarn hardhat node`

**Open a new Terminal and launch :** `yarn hardhat ignition deploy .\ignition\modules\Bank.js --network localhost`

**Come back on the Hardhat local node terminal, there is the following message in red :** `hardhat_setLedgerOutputEnabled - Method not supported`

### Search terms

hardhat, ignition, hardhat_setLedgerOutputEnabled - Method not supported, setLedgerOutputEnabled

galargh:

---

Resolves #5406

The goal of this PR is to suppress `hardhat_setLedgerOutputEnabled - Method not supported` warnings from the local hardhat node logs. They appear there when one performs a deployment against a node and does not use `hardhat-ledger` plugin.

The approach presented in this PR is a "quick fix". It assumes that if a message with `hardhat_setLedgerOutputEnabled` method reaches an HTTP provider, it should be treated as an error and not processed further (otherwise, it would have been consumed by the ledger provider before it could reach the http provider). This ensures that if `hardhat-ledger` plugin is not used, the `hardhat_setLedgerOutputEnabled` message will never reach a running node. Hence, a warning will not be printed out by the node.

This is not an ideal solution as it exposes plugin context in the core package further. Some alternatives that I was thinking about in no particular order:
- modify the ignition not to send out `hardhat_setLedgerOutputEnabled` at all if a ledger provider is not present in the chain (the challenge being identifying a ledger provider)
- make `outputEnabled` a configuration option of the ledger plugin (I'm not sure whether completely enabling/disabling output is valid for the plugin; during deployment, we only disable it temporarily)
- suppress the message at a different layer (but the alternatives don't seem much better than the http provider as far as I can tell)

### DIFF
--- a/.changeset/shaggy-bags-work.md
+++ b/.changeset/shaggy-bags-work.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Do not send `http_setLedgerOutputEnabled` messages if they reacht the HTTP Provider to prevent unwanted warnings in the local hardhat node logs
--- a/.changeset/shaggy-bags-work.md
+++ b/.changeset/shaggy-bags-work.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Do not send `http_setLedgerOutputEnabled` messages if they reacht the HTTP Provider to prevent unwanted warnings in the local hardhat node logs
+Do not send `http_setLedgerOutputEnabled` messages beyond the HTTP Provider to prevent unwanted warnings in the logs of the local hardhat node
--- a/packages/hardhat-core/src/internal/core/providers/http.ts
+++ b/packages/hardhat-core/src/internal/core/providers/http.ts
@@ -86,6 +86,15 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
   }
 
   public async request(args: RequestArguments): Promise<unknown> {
+    if (args.method === "hardhat_setLedgerOutputEnabled") {
+      const error = new ProviderError(
+        "hardhat_setLedgerOutputEnabled - Method not supported",
+        -32004
+      );
+      // eslint-disable-next-line @nomicfoundation/hardhat-internal-rules/only-hardhat-error
+      throw error;
+    }
+
     const jsonRpcRequest = this._getJsonRpcRequest(
       args.method,
       args.params as any[]
--- a/packages/hardhat-core/src/internal/core/providers/http.ts
+++ b/packages/hardhat-core/src/internal/core/providers/http.ts
@@ -86,6 +86,8 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
   }
 
   public async request(args: RequestArguments): Promise<unknown> {
+    // This is a temporary fix to an issue with noisy warnings in the logs
+    // of a local node (#5406). This will be fixed in the next major release.
     if (args.method === "hardhat_setLedgerOutputEnabled") {
       const error = new ProviderError(
         "hardhat_setLedgerOutputEnabled - Method not supported",
--- a/packages/hardhat-core/test/internal/core/providers/http.ts
+++ b/packages/hardhat-core/test/internal/core/providers/http.ts
@@ -1,10 +1,11 @@
-import { assert } from "chai";
+import { assert, expect } from "chai";
 import { MockAgent, MockPool } from "undici";
 
 import { HttpProvider } from "../../../../src/internal/core/providers/http";
 import { ERRORS } from "../../../../src/internal/core/errors-list";
 import { SuccessfulJsonRpcResponse } from "../../../../src/internal/util/jsonrpc";
 import { expectHardhatError } from "../../../helpers/errors";
+import { ProviderError } from "../../../../src/internal/core/providers/errors";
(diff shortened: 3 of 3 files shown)


packages/hardhat-core/src/internal/core/providers/http.ts
```diff
@@ -86,6 +86,15 @@ export class HttpProvider extends EventEmitter implements EIP1193Provider {
   }
 
   public async request(args: RequestArguments): Promise<unknown> {
+    if (args.method === "hardhat_setLedgerOutputEnabled") {
```
alcuadrado, line 89: The only thing I'd change is adding a comment explaining why this is here and that it's a temporary fix.

.changeset/shaggy-bags-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Do not send `http_setLedgerOutputEnabled` messages if they reacht the HTTP Provider to prevent unwanted warnings in the local hardhat node logs
```
schaable, line 5: typo 😅

schaable (approved):
LGTM, just a small typo in the changeset

.changeset/shaggy-bags-work.md
galargh: Good catch, thanks 🙇 

```suggestion
Do not send `http_setLedgerOutputEnabled` messages beyond the HTTP Provider to prevent unwanted warnings in the logs of the local hardhat node
```

packages/hardhat-core/src/internal/core/providers/http.ts
galargh: Good point, adding ➕ 

```suggestion
    // This is a temporary fix to an issue with noisy warnings in the logs
    // of a local node (#5406). This will be fixed in the next major release.
    if (args.method === "hardhat_setLedgerOutputEnabled") {
```

Issue #5549, opened 2024-07-25
Port base providers


alcuadrado:
We should port over the base providers that we know we'll use in v3, including their tests and public types.

They are:

- [x] `HttpProvider`
- [x] The backward compatibility provider wrapper
- [ ] `ProviderWrapper`, which I believe @fvictorio discovered it's buggy.
- [ ] The gas-related providers
- [ ] The accounts-related providers

Some things to note:

- They should now use our utils package
- They shouldn't use ethereumjs dependencies, in favor of paul miller's.

kanej:
This has been replaced by #5719.

PR #5580, opened 2024-08-06
feat: port the Http Provider to v3

schaable:
Part of [#5549](https://github.com/NomicFoundation/hardhat/issues/5549)

The PR includes all the changes required for porting the `HttpProvider` and tests for the `request` and `sendBatch` methods. The PR is quite large, but it can be reviewed commit by commit. Each commit includes a description of the changes made.


v-next/hardhat-utils/src/request.ts
```diff
@@ -125,7 +144,20 @@ export async function postJsonRequest(
       body: JSON.stringify(body),
     });
   } catch (e) {
-    ensureError(e);
+    ensureError<NodeJS.ErrnoException>(e);
```
ChristopherDedominici, line 138: The error handling is the same in these functions: postJsonRequest, postFormRequest and download. Could it be moved into a function?

v-next/hardhat-utils/src/lang.ts
```diff
@@ -36,3 +36,13 @@ export function isObject(
...
   return typeof value === "object" && value !== null && !Array.isArray(value);
 }
+
+/**
+ * Pauses the execution for the specified number of seconds.
+ *
+ * @param seconds The number of seconds to delay.
+ * @returns A promise that resolves after the specified delay.
+ */
+export async function delay(seconds: number): Promise<void> {
```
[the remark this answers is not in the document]
schaable: Done!

alcuadrado (approved):
Testing the merge queue, please ignore this review

v-next/core/src/internal/providers/errors.ts
```diff
@@ -0,0 +1,57 @@
...
+const ProviderErrorMessage: ProviderErrorMessages = {
+  [ProviderErrorCode.LIMIT_EXCEEDED]: "Request exceeds defined limit",
+  [ProviderErrorCode.INVALID_PARAMS]: "Invalid method parameters",
+};
+
+export class ProviderError extends CustomError implements ProviderRpcError {
+  public code: number;
+  public data?: unknown;
+
+  constructor(code: ProviderErrorCode, parentError?: Error) {
```
alcuadrado, line 33: I think this is incorrect, as the values in `ProviderErrorCode` are just a subset of all the possible error codes. Same for the messages.

I think it would be better to receive the code and message independently.

For example, the HTTP provider initializes the provider errors with a code returned by a json-rpc server, potentially missing the message if it's the code is neither `-32005` nor `-32602`.

v-next/core/src/internal/providers/http.ts
```diff
@@ -0,0 +1,361 @@
...
+
+    const jsonRpcRequest = getJsonRpcRequest(
+      this.#nextRequestId++,
+      method,
+      params,
+    );
+    const jsonRpcResponse = await this.#fetchJsonRpcResponse(jsonRpcRequest);
+
+    if (isFailedJsonRpcResponse(jsonRpcResponse)) {
+      const error = new ProviderError(jsonRpcResponse.error.code);
```
alcuadrado, line 133: See https://github.com/NomicFoundation/hardhat/pull/5580/files#r1715725985

alcuadrado: this also makes me feel that there's some test missing somewhere

v-next/core/src/internal/providers/errors.ts
```diff
@@ -0,0 +1,57 @@
+import type { ProviderRpcError } from "../../types/providers.js";
```
alcuadrado, line 1: Everything networks related should be in hardhat, not core. Ideally, inside a folder like `v-next/hardhat/src/internal/network`.

v-next/core/src/internal/utils/package.ts
```diff
@@ -0,0 +1,19 @@
+import type { PackageJson } from "@ignored/hardhat-vnext-utils/package";
```
alcuadrado, line 1: This should also live in hardhat, not core

v-next/core/src/internal/providers/http.ts
```diff
@@ -0,0 +1,361 @@
...
+  ) {
+    super();
+
+    this.#url = url;
+    this.#networkName = networkName;
+    this.#extraHeaders = extraHeaders;
+    this.#dispatcher = dispatcher;
+  }
+
+  /**
```
alcuadrado, line 104: I think it'd be better to place this doc in the interface, as that's public, while this is private

v-next/core/src/types/providers.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export interface ProviderRpcError extends Error {
+  code: number;
+  data?: unknown;
+}
+
+export interface EIP1193Provider extends EventEmitter {
+  request(args: RequestArguments): Promise<unknown>;
+}
```
alcuadrado, line 15: We should define the interface that also includes the backwards compatibility methods

v-next/core/src/internal/providers/http.ts
```diff
@@ -0,0 +1,361 @@
...
+          method: jsonRpcRequest.method,
+          params: jsonRpcRequest.params,
+        });
+        jsonRpcResponse = {
+          jsonrpc: "2.0",
+          id: jsonRpcRequest.id,
+          result,
+        };
+      } catch (error) {
+        ensureError<Error & { code: unknown }>(error);
```
alcuadrado, line 200: Note that using `ensureError<Foo>` like this is equivalent to casting, which can be incorrect at runtime.

This just happens to work because the first line after this throws if code is not defined, but this is a risky pattern.

v-next/hardhat-utils/src/request.ts
```diff
@@ -255,6 +280,17 @@ export async function getDispatcher(
   }
 }
 
+export async function getTestDispatcher(
```
alcuadrado, line 283: Shouldn't this be in the helpers instead? I find it confusing that's defined in a test and reexported form a helper.

v-next/hardhat-utils/src/request.ts
```diff
@@ -84,7 +91,10 @@ export async function getRequest(
       ...baseRequestOptions,
     });
   } catch (e) {
-    ensureError(e);
+    ensureError<NodeJS.ErrnoException>(e);
```
alcuadrado, line 94: As mentioned somewhere else in the PR, this is actually unsafe.

v-next/core/test/internal/providers/http.ts
```diff
@@ -0,0 +1,408 @@
...
+
+    it("should throw if the connection is refused", async () => {
+      // We don't have a way to simulate a connection refused error with the
+      // mock agent, so we use a real HttpProvider with localhost to test this
+      // scenario.
+      const provider = await HttpProvider.create({
+        // Using a high-numbered port to ensure connection refused error
+        url: "http://localhost:49152",
+        networkName: "exampleNetwork",
+        timeout: 20_000,
```
alcuadrado, line 115: Won't this test take 20s to run?

v-next/core/src/internal/providers/http.ts
schaable: > this also makes me feel that there's some test missing somewhere
I believe this doesn't apply anymore after the refactor, but let me know if you think otherwise.

v-next/core/src/internal/utils/package.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5580/commits/8f13be53096fc3eea0191fef0b4b28e3312124ce)

v-next/core/src/internal/providers/http.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5580/commits/7068a5fc633606e278496d74379d8276341d11e7)

v-next/core/src/types/providers.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5580/commits/7e133bda7d4aa5894222258027ffc72ead77ab29)

v-next/core/src/internal/providers/http.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5580/commits/7229bd0a281274f58361b6468a5957a5d29ca124)

v-next/hardhat-utils/src/request.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5580/commits/7229bd0a281274f58361b6468a5957a5d29ca124)

v-next/core/src/internal/providers/errors.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5580/commits/8f13be53096fc3eea0191fef0b4b28e3312124ce)

alcuadrado (approved):

Issue #5437, opened 2024-06-27
Enhancements to custom reporter


kanej:
In Hardhat v3 we expect the default experience will include the `./v-next/hardhat-node-test-reporter` as part of the install.

This is a placeholder for improvements to the custom reporter.

## PRs

- [x] #5369
- [x] #5662

## Done

- Add a README.md explaining the project, how to install it, use it, and that it aims to have an output similar to mocha's default reporter. On top of that, explain the custom features that we add.
The general architecture of the project is a bit rough, I think it could use some documentation/explanations. Things about separation of the main generator with the UI are there, but maybe a bit messy.

## In Review

- Testing, mostly integration tests. Most edge cases aren't tested. Just to list a few: running with --test-only, tests failing an a before/after/beforeEach/afterEach, a test file throwing top level, a describe callback throwing, no tests found, --test-only excluding every tests, multiple tests files, tests files importing other tests files, errors with and without cause.

## In Progress

- The error formatting is somewhat fragile, as it tries to do util.inspect, then split apart the stack trace from the message, and maybe add a diff. This leads to a few edge cases, and doesn't properly support error causes. I think we can take more control over it and parse the stack trace manually (ideally without any dependency except if we find a well-maintained ~0-dep one), to make it more robust.
- We should clean up the internal node stack trace frames (related to the above).

## TODO

- We should normalize paths between windows/linux (i.e. path.sep), and relative/absolute paths. Ideally, showing absolute paths, but it is relative in case they are within the CWD.

alcuadrado:
This reporter was hacked together by me, based on a project I found offline and some work by Chris.

It needs general polishing and better testing, and we also need to add some custom functionality.

These are the polishing and testing things we need to improve:

1. Add a `README.md` explaining the project, how to install it, use it, and that it aims to have an output similar to mocha's default reporter. On top of that, explain the custom features that we add.
1. The general architecture of the project is a bit rough, I think it could use some documentation/explanations. Things about separation of the main generator with the UI are there, but maybe a bit messy.
1. Testing, mostly integration tests. Most edge cases aren't tested. Just to list a few: running with `--test-only`, tests failing an a before/after/beforeEach/afterEach, a test file throwing top level, a describe callback throwing, no tests found, `--test-only` excluding every tests, multiple tests files, tests files importing other tests files, errors with and without cause.
[...]

PR #5584, opened 2024-08-08
Node test runner plugin supporting js tests

### LINKED ISSUE #5534: Integration testing with Node Test Runner
Add integration testing support to Hardhat v3 based on Node Test Runner.

Node Test Runner support will be implemented as a separate package and plugin. It will add a `hardhat test` task that will initialize the `hre` and then execute the test suites, injecting the HRE into the context. 

Full details can be find in the [design doc](https://www.notion.so/nomicfoundation/Node-Test-Runner-Design-Doc-0aab6afb06814d2bb2fc5e299c00f77c?pvs=4).

## TODO

- [ ]  Add a new `@ignored/hardhat-node-test-runner` package under `./v-next`
- [ ]  Implement as a plugin task shim code similar to v2 mocha, that runs the designated test folder js/ts files as tests in Node test runner. Before the tests are run the HRE should be initialized and injected into the context.
- [ ]  Include the Hardhat custom test reporter in the default execution
- [ ]  Add `@ignored/hardhat-node-test-runner` to the example project and add some basic integration tests confirming that testing works and that the hre is being initialized
- [ ]  Add support for `--only` to the `npx hardhat test` command

## Out of Scope

- Coverage
- Gas reporting
- Parallel test execution
- Support for console.log in tests

zoeyTM:
closes #5534


alcuadrado:
I think the CI is failing due to the pnpm problem that @galargh mentioned

galargh:
I see you already found the fix in [`a2290a3` (#5584)](https://github.com/NomicFoundation/hardhat/pull/5584/commits/a2290a349b6c1cebd44833213e55c8a5e82462dd) 🥳  I just wanted to confirm that without this I had the same issue as CI reported and that defining the references fixes it for me.

galargh (commented):
Looks great 😸 I left a couple of questions to ensure I understand all the changes being made. I hope you don't mind I'm trying to use the review process as a learning opportunity, too. 

The main items I couldn't find definite answers for myself were:
1. Do we want to cover all the test cases outlined in the [Design Doc](https://www.notion.so/nomicfoundation/Node-Test-Runner-Design-Doc-0aab6afb06814d2bb2fc5e299c00f77c?pvs=4#8d6e3b1e98c8411fb3aa75bceb17df3a) in this iteration?
2. Should `hre` injection into the context be supported after all or not? I thought imports were supposed to be the only path, but I wanted to double-check.

---

## Review Checklist (copied from the issue)

- [x]  Add a new `@ignored/hardhat-node-test-runner` package under `./v-next`
  - It is called `@ignored/hardhat-vnext-node-test-runner` which follows the guidelines for `v-next` packages.
- [x]  Implement as a plugin task shim code similar to v2 mocha, that runs the designated test folder js/ts files as tests in Node test runner. Before the tests are run the HRE should be initialized and injected into the context.
  - HRE can be imported in the tests, it does not need to be injected.
- [x]  Include the Hardhat custom test reporter in the default execution
- [x]  Add `@ignored/hardhat-node-test-runner` to the example project and add some basic integration tests confirming that testing works and that the hre is being initialized
- [x]  Add support for `--only` to the `npx hardhat test` command
- [x] Add cli support for testFiles an optional list of test files

## Out of Scope (copied from the issue)

- Coverage
- Gas reporting
- Parallel test execution
- Support for console.log in tests

alcuadrado:
I'll review this PR tomorrow morning

v-next/example-project/package.json
```diff
@@ -22,8 +22,12 @@
   },
   "devDependencies": {
     "@ignored/hardhat-vnext": "workspace:^3.0.0-next.3",
+    "@ignored/hardhat-vnext-node-test-runner": "workspace:^3.0.0-next.2",
     "@types/node": "^20.14.9",
     "prettier": "3.2.5",
     "typescript": "~5.5.0"
+  },
+  "dependencies": {
+    "tsx": "^4.11.0"
```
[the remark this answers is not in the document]
zoeyTM: I actually think it should probably be a peer dependency now that I think about it. @alcuadrado would know for sure

v-next/example-project/test/other-example-test.js
```diff
@@ -0,0 +1,10 @@
+import assert from "node:assert/strict";
+import { describe, it } from "node:test";
+
+import hre from "@ignored/hardhat-vnext";
```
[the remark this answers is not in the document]
zoeyTM: Correct. The issue was out of date

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+async function getTestFiles(
+  testFiles: string[],
+  config: HardhatConfig,
+): Promise<string[]> {
+  if (testFiles.length !== 0) {
+    const testFilesAbsolutePaths = testFiles.map((x) =>
+      resolve(process.cwd(), x),
+    );
```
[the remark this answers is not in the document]
zoeyTM: I don't have answers to the other questions, but we're already checking for TS files later in the process. What is your reasoning for thinking we might need that here as well?

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,77 @@
...
+      );
+    }
+
+    try {
+      import.meta.resolve("tsx");
+    } catch {
+      throw new HardhatError(HardhatError.ERRORS.GENERAL.TSX_NOT_INSTALLED);
+    }
+
+    process.env.NODE_OPTIONS = "--import tsx";
```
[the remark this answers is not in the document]
zoeyTM: This was the suggested solution, and I think it is okay personally

v-next/hardhat-node-test-runner/test/index.ts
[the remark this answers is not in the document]
zoeyTM: The tests live in `v-next/example-project`

v-next/hardhat-node-test-runner/src/task-action.ts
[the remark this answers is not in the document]
galargh: I was thinking about it in terms of validation - to make sure we only get JavaScript and TypeScript files and not, let's say, some Markdown by mistake.

v-next/example-project/package.json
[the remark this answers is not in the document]
alcuadrado: I think this shouldn't be needed.

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -196,6 +196,26 @@ Please add the property "type" with the value "module" in your package.json to e
       websiteTitle: "Invalid config",
       websiteDescription: `The configuration you provided is invalid. Please check the documentation to learn how to configure Hardhat correctly.`,
     },
+    TYPESCRIPT_NOT_INSTALLED: {
+      number: 17,
+      messageTemplate: `Your Hardhat project uses typescript, but it's not installed.
```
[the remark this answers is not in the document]
alcuadrado: We'll polish all the error messages before the release.

v-next/hardhat-node-test-runner/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "../../config-v-next/tsconfig.json",
+  "references": [
+    {
```
alcuadrado, line 4: I think hardhat is missing here.

v-next/hardhat-node-test-runner/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "eslint-plugin-import": "2.29.1",
+    "eslint-plugin-no-only-tests": "3.1.0",
+    "expect-type": "^0.19.0",
+    "prettier": "3.2.5",
+    "rimraf": "^5.0.5",
+    "typescript": "~5.5.0",
+    "typescript-eslint": "7.7.1"
+  },
+  "dependencies": {
+    "@ignored/hardhat-vnext-core": "workspace:^3.0.0-next.2",
```
alcuadrado, line 59: Only hardhat should install core. The plugiins should install `hardhat`.

alcuadrado: We may even merge those packages to avoid confusion.

v-next/hardhat-node-test-runner/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "@ignored/hardhat-vnext-errors": "workspace:^3.0.0-next.2",
+    "@ignored/hardhat-vnext-node-test-reporter": "workspace:^3.0.0-next.2",
+    "@ignored/hardhat-vnext-utils": "workspace:^3.0.0-next.2",
+    "tsx": "^4.11.0"
```
alcuadrado, line 63: I think that installing hardhat may make this unneeded.

v-next/hardhat-node-test-runner/src/task-action.ts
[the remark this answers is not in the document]
alcuadrado: I think we need to handle absolute paths here.

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,77 @@
...
+      throw new HardhatError(
+        HardhatError.ERRORS.GENERAL.TYPESCRIPT_NOT_INSTALLED,
+      );
+    }
+
+    try {
+      import.meta.resolve("tsx");
+    } catch {
+      throw new HardhatError(HardhatError.ERRORS.GENERAL.TSX_NOT_INSTALLED);
+    }
```
alcuadrado, line 69: Same here, we don't really need this. v3 always supports ts.

v-next/hardhat-node-test-runner/package.json
alcuadrado: See my comment blow

v-next/hardhat-node-test-runner/src/task-action.ts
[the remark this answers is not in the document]
zoeyTM: > I was thinking about it in terms of validation - to make sure we only get JavaScript and TypeScript files and not, let's say, some Markdown by mistake.

@galargh I don't think we care about validating this actually. If a user wants to pass a .md file or something else to the test runner, the file would just get ignored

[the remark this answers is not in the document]
galargh: That's fair, makes sense 🙇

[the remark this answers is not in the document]
alcuadrado: now that I think of, doesn't `node:test` resolve relative paths? Same for globs

alcuadrado (changes requested):
Just left a few comments.

I think adding some of the tests laid out in the design doc would be great.

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,54 @@
...
+const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
+  { testFiles, only },
+  hre,
+) => {
+  const files = await getTestFiles(testFiles, hre.config);
+
+  const tsx = fileURLToPath(import.meta.resolve("tsx/esm"));
+  process.env.NODE_OPTIONS = `--import ${tsx}`;
+
+  run({ files, only }).compose(hardhatTestReporter).pipe(process.stdout);
```
alcuadrado, line 51: Don't we have to await this or something? How can we get to know if the tests failed?

We should set `process.exitCode` if it fails, and also return some error code.

zoeyTM: It's not a promise, it's a stream. Failed tests get piped through to `process.stdout`

v-next/hardhat-node-test-runner/src/index.ts
```diff
@@ -0,0 +1,23 @@
...
+const hardhatPlugin: HardhatPlugin = {
+  id: "test",
+  tasks: [
+    task("test", "Runs tests using the NodeJS test runner")
+      .addVariadicArgument({
+        name: "testFiles",
+        description: "An optional list of files to test",
+        defaultValue: [],
+      })
+      .addFlag({
```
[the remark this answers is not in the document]
zoeyTM: The design doc specifically mentions that these should be added as follow on tasks later. Do we actually want to do them now then?

v-next/hardhat-node-test-runner/src/task-action.ts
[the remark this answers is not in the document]
zoeyTM: it does, I just tested this

v-next/hardhat-node-test-runner/src/index.ts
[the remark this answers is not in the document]
alcuadrado: Follow-up prs are ok, but can you add them to this and the mocha plugins as follow-up prs before moving to a different topic? I think it's better to do these while you have all the context in your mind.

v-next/hardhat-node-test-runner/src/task-action.ts
alcuadrado: I went deep into this rabbit hole, and came out with this example snippet showing how to do these things

```ts
import path from "node:path";
import { finished } from "node:stream/promises";
import { run } from "node:test";
import { tap } from "node:test/reporters";

async function runTests(): Promise<number> {
  let failures = 0;
  // `files:` resolves relative paths, not globs, that's ok.
  const reporterStream = run({ files: [path.resolve("./test.ts")] })
    .on("test:fail", () => failures++)
    .compose(tap);

  reporterStream.pipe(process.stdout);

  // This should use the reporter's stream, not the one piped to stdout, as it
  // never finishes, as stdout is never closed.
  await finished(reporterStream);

  return failures;
}

const failures = await runTests();
console.log("Failures:", failures);

if (failures > 0) {
  process.exitCode = failures;
}
```

Here we:

1. Count the test failures and return it.
2. Set `process.exitCode` if needed.
3. Await for the test run to finish.

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,101 @@
...
+    await finished(reporterStream);
+
+    return failures;
+  }
+
+  const testFailures = await runTests();
+  console.log("Failures: ", testFailures);
+
+  if (testFailures > 0) {
+    process.exitCode = testFailures;
```
galargh, line 95: I'd stick to setting the exit code to 1. That way the users won't "overinterpret" the meaning of the exit code.

```suggestion
    process.exitCode = 1;
```

zoeyTM: This is copying what hardhat v2 is currently doing, @alcuadrado can you give any insight on the intention behind it?

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,101 @@
...
+          // If a suite failed only because a subtest failed, we don't want to
+          // count it as a failure since the subtest failure will be reported as well
+          if (isSubtestFailedError(event.details.error)) {
+            return;
+          }
+        }
+
+        failures++;
+      })
+      .compose(hardhatTestReporter);
```
galargh, line 82: I'm just starting to look into #5437 and it made me think, is the long-term plan to make the reporter for the node:test and mocha plugins replacable? Or would we expect the users to implement their own test task if they want to swap out the reporter?

v-next/hardhat-node-test-runner/src/task-action.ts
alcuadrado: copying Mocha's behavior tbh

v-next/hardhat-node-test-runner/src/task-action.ts
alcuadrado: That's a good point, but we can probably defer that decision until we get more user input

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,101 @@
...
+
+    reporterStream.pipe(process.stdout);
+
+    await finished(reporterStream);
+
+    return failures;
+  }
+
+  const testFailures = await runTests();
+  console.log("Failures: ", testFailures);
```
alcuadrado, line 92:
```suggestion
  const testFailures = await runTests();
```

That output is a responsibility of the reporter.

alcuadrado (approved):
I left a comment to remove a `console.log`, apart from that this looks ready to be merged

v-next/hardhat-node-test-runner/src/task-action.ts
galargh: Oh wow! I didn't know that, just read through https://github.com/mochajs/mocha/issues/3559

galargh: I thought a little bit more about it over the weekend, and I just wanted to say that, personally, I'd prefer (and expect) the exit codes to follow the bash exit code guidelines. I do, however, understand if we want to follow and replicate what mocha is doing.

PR #5574, opened 2024-08-05
Merge branch main into v-next to update GitHub Actions

galargh:
- [ ] Because this PR includes a **bug fix**, relevant tests have been included.
- [ ] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.
- [x] I didn't do anything of this.

---

This PR merges the current `main` into the `v-next` branch. The intent is to get the recent updates of the GitHub Actions workflows on the `v-next` branch too. It is a result of running the following commands (writing them out to ensure I'm following the right procedure):
```bash
git merge origin/main -X theirs

rm .github/workflows/comment-on-linter-error.yml
rm .github/workflows/hardhat-chai-matchers-ci.yml
rm .github/workflows/hardhat-ethers-ci.yml
rm .github/workflows/hardhat-foundry-ci.yml
rm .github/workflows/hardhat-ledger-ci.yml
rm .github/workflows/hardhat-network-helpers-ci.yml
rm .github/workflows/hardhat-network-tracing-all-solc-versions.yml
rm .github/workflows/hardhat-shorthand-ci.yml
rm .github/workflows/hardhat-toolbox-ci.yml
rm .github/workflows/hardhat-verify-ci.yml
rm .github/workflows/hardhat-viem-ci.yml
rm .github/workflows/hardhat-vyper-ci.yml
rm .github/workflows/hardhat-web3-v4-ci.yml

rm -r packages/hardhat-solhint
rm -r packages/hardhat-truffle5

# Update pnpm-lock.yaml(s)
pushd docs && pnpm install && popd
pnpm install

git add . && git merge --continue
```

PR #5585, opened 2024-08-08
fix: run pnpm install to fix the lockfile

galargh:

---

I noticed that the pnpm lockfile is broken on the `v-next` currently. This is likely because I merged https://github.com/NomicFoundation/hardhat/pull/5574 which brought in `main` into `v-next` so it naturally had a lot of changes (including having to rewrite pnpm lock) and then https://github.com/NomicFoundation/hardhat/pull/5541 got merged which likely got tested before and, later, the conflict was too obscure for GitHub to notice.

To fix the lockfile, I just ran `pnpm install` in the root.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1539,7 +1539,7 @@ importers:
         version: 5.3.0
       debug:
         specifier: ^4.1.1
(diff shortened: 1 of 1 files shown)


kanej (approved):
LGTM.

PR #5587, opened 2024-08-08
chore: upgrade typescript to ~5.5.0

galargh:

---

We use `${configDir}` template variable in our `tsconfig.json` on `v-next` (e.g. https://github.com/NomicFoundation/hardhat/blob/bae09d38603cc31b7257c43b9d3abc768384abd4/config-v-next/tsconfig.json#L4) which has only been released in TypeScript 5.5 (https://devblogs.microsoft.com/typescript/announcing-typescript-5-5-beta/#the-configdir-template-variable-for-configuration-files).

I've noticed it only, because I had my VSCode configured to use the workspace version of TypeScript, and when browsing files in `v-next`, it'd fail to resolve types.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "hardhat": "workspace:*",
     "prettier": "2.4.1",
     "shelljs": "^0.8.5",
-    "typescript": "~5.0.0"
+    "typescript": "~5.5.0"
   },
   "scripts": {
     "build": "tsc --build packages/hardhat-core packages/hardhat-ethers packages/hardhat-verify packages/hardhat-vyper packages/hardhat-web3-v4 packages/hardhat-chai-matchers packages/hardhat-network-helpers packages/hardhat-toolbox packages/hardhat-foundry packages/hardhat-ledger packages/hardhat-viem packages/hardhat-toolbox-viem",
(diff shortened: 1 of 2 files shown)


alcuadrado:
> I've noticed it only, because I had my VSCode configured to use the workspace version of TypeScript,

I didn't understand the motivation for this PR at first, but I think this setting is the key reason, isn't it? Because I never had a problem, but I don't have that setting enabled (I probably should).

How would this work if it didn't have any version in the root of the project? I'm wondering because we only use that one in the v2 parts of the repo (i.e. `packages/`), not in the v3 ones. 

Having said that, as this didn't create any compilation error in `packages/`, I don't see any reason not to merge it.

alcuadrado (approved):

galargh:
> > I've noticed it only, because I had my VSCode configured to use the workspace version of TypeScript,
> 
> I didn't understand the motivation for this PR at first, but I think this setting is the key reason, isn't it? Because I never had a problem, but I don't have that setting enabled (I probably should).
> 
> How would this work if it didn't have any version in the root of the project? I'm wondering because we only use that one in the v2 parts of the repo (i.e. `packages/`), not in the v3 ones.
> 
> Having said that, as this didn't create any compilation error in `packages/`, I don't see any reason not to merge it.

Yes, exactly.

By default, VSCode looks for TypeScript in the root `node_modules`. If we don't define `typescript` dependency at the root level, then the only option it has is to use the VSCode version of TypeScript. Unless one opens individual packages, of course.

PR #5592, opened 2024-08-09
feat: add a built-in console task

### LINKED ISSUE #5575: Console task
Start a node console with the `HardhatRuntimeEnvironment` initialized and attached to the `global` scope:

```tsx
hardhat console
> console.log(hre.config.solidity)
// "0.8.24"
```

For inspiration consider the HHv2 version of console:

https://github.com/NomicFoundation/hardhat/blob/83721d9cf23fb8d40eb372fea793377eba89e582/packages/hardhat-core/src/builtin-tasks/console.ts#L13

The v3 version of the `run` task is likely a good comparison for integration testing.

We should consider accepting a --config parameter to allow specifying a path to an alternative `hardhat.config.{js,ts}` file.

galargh:

---

Resolves #5575

This PR adds a `console` task to the next version of `hardhat`. Here's how it looks like:

```
$ # Show console task's help
$ 
$ pnpm hardhat console --help
Opens a hardhat console

Usage: hardhat [GLOBAL OPTIONS] console [--history <STRING>] [--no-compile] [--] [commands]

OPTIONS:

  --history         Path to a history file
  --no-compile      Don't compile before running this task

POSITIONAL ARGUMENTS:

  commands          Commands to run in the console

For global options help run: hardhat --help
$ 
$ # Run the console task and preview the `hre` context variable
$ 
$ pnpm hardhat console
> .help
.break    Sometimes you get stuck, this gets you out
.clear    Break, and also clear the local context
.editor   Enter editor mode
.exit     Exit the REPL
.help     Print this help message
.load     Load JS from a file into the REPL session
.save     Save all evaluated commands in this REPL session to a file

Press Ctrl+C to abort current expression, Ctrl+D to exit the REPL
> hre
HardhatRuntimeEnvironmentImplementation { userConfig: { tasks: [Array], plugins: [Array], privateKey: [Object] }, config: { plugins: [Array], tasks: [Array], path...
```

The task includes the following options:
- `history` - allows specifying the path where REPL history file should be stored. If it's relative, it will be resolved in the context of global hardhat cache dir. If it is explicitly set to an empty string, the history will not be preserved. It defaults to `console-history.txt` in the global hardhat cache dir. This is very helpful during testing, but I thought it might be useful to the users, too.
- `noCompile` - disables compilation task. This is not implemented yet.
- `options` - This is a hidden argument. It is not exposed to the users. It is only used during testing to set custom REPL options. In particular, we override the IO streams during testing because otherwise the test runner hangs indefinitely.
- `commands` - This is a variadic argument. One can provide a list of commands that should be executed in the REPL after the startup. It defaults to running the `.help` command after startup. This is very helpful during testing because we want to run the commands after the entire REPL setup fully complete there, but I thought it might be useful to the users, too. 

#### Testing
- [x] Added tests for the newly created task action; they are passing in CI
- [x] Manually run the newly added task in the `example-project`; `pnpm hardhat console`

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/console/index.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/console/index.ts
@@ -7,6 +7,11 @@ const hardhatPlugin: HardhatPlugin = {
   tasks: [
     task("console", "Opens a hardhat console")
       .setAction(import.meta.resolve("./task-action.js"))
+      .addOption({
+        name: "history",
+        description: "Path to a history file",
+        defaultValue: "console-history.txt",
+      })
       .addVariadicArgument({
         name: "commands",
         description: "Commands to run in the console",
--- a/v-next/hardhat/src/internal/builtin-plugins/console/index.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/console/index.ts
@@ -1,0 +1,19 @@
+import type { HardhatPlugin } from "@ignored/hardhat-vnext-core/types/plugins";
+
+import { task } from "@ignored/hardhat-vnext-core/config";
+
+const hardhatPlugin: HardhatPlugin = {
+  id: "console",
+  tasks: [
+    task("console", "Opens a hardhat console")
+      .setAction(import.meta.resolve("./task-action.js"))
+      .addVariadicArgument({
+        name: "commands",
+        description: "Commands to run in the console",
+        defaultValue: [".help"],
+      })
+      .build(),
+  ],
+};
+
+export default hardhatPlugin;
--- a/v-next/hardhat/src/internal/builtin-plugins/console/index.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/console/index.ts
@@ -12,6 +12,10 @@ const hardhatPlugin: HardhatPlugin = {
         description: "Path to a history file",
         defaultValue: "console-history.txt",
       })
+      .addFlag({
+        name: "noCompile",
+        description: "Don't compile before running this task",
+      })
       .addVariadicArgument({
         name: "commands",
         description: "Commands to run in the console",
--- a/v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
@@ -1,0 +1,46 @@
+import type { NewTaskActionFunction } from "@ignored/hardhat-vnext-core/types/tasks";
+import type { REPLServer } from "node:repl";
+
+import path from "node:path";
+import repl from "node:repl";
+
+import { getCacheDir } from "@ignored/hardhat-vnext-core/global-dir";
+import debug from "debug";
+
+const log = debug("hardhat:core:tasks:console");
+
+interface ConsoleActionArguments {
+  commands: string[];
+}
+
+const consoleAction: NewTaskActionFunction<ConsoleActionArguments> = async (
+  { commands },
+  _hre,
+) => {
+  const globalCacheDir = await getCacheDir();
+
+  return new Promise<REPLServer>((resolve) => {
+    // Start a new REPL server with the default options
+    const replServer = repl.start();
+
+    // Set up the REPL history file in the global cache directory
+    const historyPath = path.join(globalCacheDir, "console-history.txt");
+    replServer.setupHistory(historyPath, (err: Error | null) => {
+      if (err !== null) {
+        log(`Failed to setup REPL history: ${err.message}`);
+      }
+    });
+
+    // Execute each command in the REPL server
+    for (const command of commands) {
+      replServer.write(`${command}\n`);
+    }
+
+    // Resolve the task action promise when the REPL server exits
+    replServer.on("exit", () => {
+      resolve(replServer);
+    });
+  });
+};
+
+export default consoleAction;
--- a/v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
@@ -20,16 +20,16 @@ const consoleAction: NewTaskActionFunction<ConsoleActionArguments> = async (
   { commands, history, options },
   hre,
 ) => {
-  // Resolve the history path if it is not empty
-  let historyPath: string | undefined;
-  if (history !== "") {
-    const globalCacheDir = await getCacheDir();
-    historyPath = path.isAbsolute(history)
-      ? history
-      : path.resolve(globalCacheDir, history);
-  }
+  return new Promise<REPLServer>(async (resolve) => {
+    // Resolve the history path if it is not empty
+    let historyPath: string | undefined;
+    if (history !== "") {
+      const globalCacheDir = await getCacheDir();
+      historyPath = path.isAbsolute(history)
+        ? history
+        : path.resolve(globalCacheDir, history);
+    }
 
-  return new Promise<REPLServer>((resolve) => {
     // Start a new REPL server with the default options
     const replServer = repl.start(options);
 
@@ -43,11 +43,14 @@ const consoleAction: NewTaskActionFunction<ConsoleActionArguments> = async (
 
     // Set up the REPL history file if the historyPath has been set
     if (historyPath !== undefined) {
-      replServer.setupHistory(historyPath, (err: Error | null) => {
-        // Fail silently if the history file cannot be set up
-        if (err !== null) {
-          log(`Failed to setup REPL history: ${err.message}`);
-        }
+      await new Promise<void>((resolveSetupHistory) => {
+        replServer.setupHistory(historyPath, (err: Error | null) => {
+          // Fail silently if the history file cannot be set up
+          if (err !== null) {
+            log(`Failed to setup REPL history: ${err.message}`);
+          }
+          resolveSetupHistory();
+        });
       });
     }
 
--- a/v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
@@ -46,6 +46,11 @@ const consoleAction: NewTaskActionFunction<ConsoleActionArguments> = async (
 
     // Add the Hardhat Runtime Environment to the REPL context
     replServer.context.hre = hre;
+    replServer.context.config = hre.config;
+    replServer.context.tasks = hre.tasks;
+    replServer.context.globalOptions = hre.globalOptions;
+    replServer.context.hooks = hre.hooks;
+    replServer.context.interruptions = hre.interruptions;
 
     // Set up the REPL history file if the historyPath has been set
     if (historyPath !== undefined) {
--- a/v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
@@ -24,10 +24,11 @@ const consoleAction: NewTaskActionFunction<ConsoleActionArguments> = async (
   // Resolve the history path if it is not empty
   let historyPath: string | undefined;
   if (history !== "") {
-    const globalCacheDir = await getCacheDir();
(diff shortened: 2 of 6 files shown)


v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
```diff
@@ -0,0 +1,46 @@
...
+const consoleAction: NewTaskActionFunction<ConsoleActionArguments> = async (
+  { commands },
+  _hre,
+) => {
+  const globalCacheDir = await getCacheDir();
+
+  return new Promise<REPLServer>((resolve) => {
+    // Start a new REPL server with the default options
+    const replServer = repl.start();
+
```
alcuadrado, line 25: I think what's missing here is exporting the `hre`. I believe that's done with `replServer.context.hre = hre`.

alcuadrado: We should also consider exposing each `hre` field in the context, similar to what we do in `v-next/hardhat/src/index.ts`

v-next/hardhat/src/internal/builtin-plugins/console/index.ts
```diff
@@ -0,0 +1,19 @@
+import type { HardhatPlugin } from "@ignored/hardhat-vnext-core/types/plugins";
+
+import { task } from "@ignored/hardhat-vnext-core/config";
+
+const hardhatPlugin: HardhatPlugin = {
+  id: "console",
+  tasks: [
+    task("console", "Opens a hardhat console")
+      .setAction(import.meta.resolve("./task-action.js"))
```
alcuadrado, line 9: In v2 we compile the project before running by default, and have a `--no-compile` flag. We should keep that functionality.

v-next/hardhat/src/internal/builtin-plugins/console/index.ts
```diff
@@ -0,0 +1,19 @@
...
+      .addVariadicArgument({
```
alcuadrado, line 10: I'm not sure what this'd do

alcuadrado:
I did an initial review despite this being a draft. Looking good! I just left some minor comments.

v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
```diff
@@ -0,0 +1,70 @@
...
+    // Start a new REPL server with the default options
+    const replServer = repl.start(options);
+
+    // Resolve the task action promise only when the REPL server exits
+    replServer.on("exit", () => {
+      resolve(replServer);
+    });
+
+    // Add the Hardhat Runtime Environment to the REPL context
+    replServer.context.hre = hre;
```
galargh, line 48: @alcuadrado I'm continuing [the dicsussion about the `hre` in the context](https://github.com/NomicFoundation/hardhat/pull/5592#discussion_r1712244971) here since I now added it.

> I think what's missing here is exporting the hre. I believe that's done with replServer.context.hre = here.
> 
> We should also consider exposing each hre field in the context, similar to what we do in v-next/hardhat/src/index.ts

I didn't expose it initially because I assumed we wanted the users to import it explicitly with `const { default: hre } = await import("@ignored/hardhat-vnext");` for example.

Here's also a suggestion for exposing other hre fields in the context. Let me know if we should proceed with it.   

```suggestion
    // Add the Hardhat Runtime Environment to the REPL context
    replServer.context.hre = hre;
    replServer.context.config = config;
    replServer.context.tasks = tasks;
    replServer.context.globalOptions = globalOptions;
    replServer.context.hooks = hooks;
    replServer.context.interruptions = interruptions;
```

v-next/hardhat/test/internal/builtin-plugins/console/task-action.ts
```diff
@@ -0,0 +1,191 @@
...
+
+    afterEach(async function () {
+      // We try to remove the temporary cache dir after each test, but we don't
+      // fail the test if it fails. For example, we have observed that in GHA
+      // on Windows, the temp dir cannot be removed due to permission issues.
+      try {
+        await remove(cacheDir);
+      } catch (error) {
+        log("Failed to remove temporary cache dir", error);
+      }
```
galargh, line 165: I added a try-catch and a comment about an issue I observed in https://github.com/NomicFoundation/hardhat/actions/runs/10332073748/job/28602920453, for example. Have you seen something similar before? Do you think it's worth ensuring we can always delete the temp dirs in CI? In this particular case I think it might have something to do that the temp dir is on `C:` while the runner is set up to have perms and execute in the context of drive `D:`.

v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  // If noCompile is false, run the compile task first
+  if (!noCompile) {
+    // todo: run compile task
+  }
+
+  return new Promise<REPLServer>(async (resolve) => {
+    // Start a new REPL server with the default options
+    const replServer = repl.start(options);
```
galargh, line 40: Unfortunately, the only option to create an instance of REPL is to start it immediately. This is because the constructor of the REPL class is private. Ideally, we would only start it after the rest of the setup is complete.

galargh (commented):
Thanks for the initial review! I usually tend to use drafts for work-in-progress proposals which is pending my self-review. Here, it turned out not to be too far off from the final "ready for review" state ;)

v-next/hardhat/src/internal/builtin-plugins/console/index.ts
alcuadrado: correct

v-next/hardhat/test/internal/builtin-plugins/console/task-action.ts
```diff
@@ -0,0 +1,182 @@
...
+  });
+
+  describe("history", function () {
+    let cacheDir: string;
+    let history: string;
+
+    beforeEach(async function () {
+      cacheDir = await fsPromises.mkdtemp(
+        path.resolve(os.tmpdir(), "console-action-test-"),
+      );
```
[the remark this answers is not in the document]
alcuadrado: Yes, let's add it to the utils instead. Our intention is to avoid using `fs` anywhere except that one module. We'll soon have a linter rule for that.

But let's do it in a follow up PR. Can you do it/create an issue about it?

v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
alcuadrado: > I didn't expose it initially because I assumed we wanted the users to import it explicitly with const { default: hre } = await import("@ignored/hardhat-vnext"); for example.

We want that for tests/scripts, but I think in the console it makes sense to automatically import them, similarly to what node's own console does.

alcuadrado: Let's commit that change, yes

v-next/hardhat/test/internal/builtin-plugins/console/task-action.ts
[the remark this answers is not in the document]
alcuadrado: @schaable tagging you just for visibility.

v-next/hardhat/src/internal/builtin-plugins/console/index.ts
```diff
@@ -0,0 +1,28 @@
...
+const hardhatPlugin: HardhatPlugin = {
+  id: "console",
+  tasks: [
+    task("console", "Opens a hardhat console")
+      .setAction(import.meta.resolve("./task-action.js"))
+      .addOption({
+        name: "history",
+        description: "Path to a history file",
+        defaultValue: "console-history.txt",
+      })
```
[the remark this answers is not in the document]
alcuadrado: Let's keep them. But as a small note, it can be faster to sync on these things on slack first.

v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
```diff
@@ -0,0 +1,70 @@
...
+import debug from "debug";
+
+const log = debug("hardhat:core:tasks:console");
+
+interface ConsoleActionArguments {
+  commands: string[];
+  history: string;
+  noCompile: boolean;
+  // We accept ReplOptions as an argument to allow tests overriding the IO streams
+  options?: repl.ReplOptions;
```
alcuadrado, line 17: This is really interesting! I never thought about it.

v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const consoleAction: NewTaskActionFunction<ConsoleActionArguments> = async (
+  { commands, history, noCompile, options },
+  hre,
+) => {
+  // Resolve the history path if it is not empty
+  let historyPath: string | undefined;
+  if (history !== "") {
+    const globalCacheDir = await getCacheDir();
```
alcuadrado, line 27: Given that we have a local cache dir per project, I think it's better if we use that one.

If it's not available yet, let's just create an issue to make sure we get back to this.

v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
```diff
@@ -0,0 +1,70 @@
...
+    historyPath = path.isAbsolute(history)
+      ? history
+      : path.resolve(globalCacheDir, history);
+  }
+
+  // If noCompile is false, run the compile task first
+  if (!noCompile) {
+    // todo: run compile task
```
alcuadrado, line 35: Let's create an issue so that we don't forget about it. Do you mind doing it?

v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
alcuadrado: It's crazy, but unfortunately, not surprising.

v-next/hardhat/test/internal/builtin-plugins/console/task-action.ts
alcuadrado: that's actually sirprising, but I think we can live with it.

Maybe we should move it to the test utils package so that we don't hit this edge case again.

alcuadrado (approved):
This looks good! :)

I left a few comments, mostly about things that can be done later. Do you mind creating issues so that we don't loose track of them?

schaable (commented):
Nothing to add to Pato's comments. This looks great, nice work!

v-next/hardhat/src/internal/builtin-plugins/console/index.ts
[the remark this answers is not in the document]
galargh: Got it! Thanks :)

v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
galargh: Of course, the issue created at https://github.com/NomicFoundation/hardhat/issues/5599. I'll also add an inline comment about it.

v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
galargh: Not at all, the issue's here - https://github.com/NomicFoundation/hardhat/issues/5600

v-next/hardhat/test/internal/builtin-plugins/console/task-action.ts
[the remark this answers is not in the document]
galargh: I created the issue here - https://github.com/NomicFoundation/hardhat/issues/5601

v-next/hardhat/test/internal/builtin-plugins/console/task-action.ts
galargh: I created an issue for this - https://github.com/NomicFoundation/hardhat/issues/5602 If I cannot resolve it, I'm going to add the utility function for this.

Issue #5586, opened 2024-08-08
Integration testing with mocha


kanej:
Add integration testing support to Hardhat v3 based on mocha.

Mocha support will be implemented as a separate package and plugin.

We need some design decisions as whether, we should refactor the test task so that plugins hook into it, or whether integration test plugins should detect clashing test plugins.

PR #5598, opened 2024-08-14
Mocha test plugin

zoeyTM:
closes https://github.com/NomicFoundation/hardhat/issues/5586

Branches off of https://github.com/NomicFoundation/hardhat/pull/5584 so that PR should be merged first


v-next/example-project/test/mocha/other-mocha-test.ts
```diff
@@ -0,0 +1,16 @@
...
+//   it("should have the example task", () => {
+//     assert.ok(hre.tasks.getTask("example"));
+//   });
+// });
+
+describe("Other mocha test", () => {
+  it("should have the example task", () => {
+    assert.ok(true);
+  });
+});
```
galargh, line 16: We should uncomment the version that uses `hre`, I think. 

```suggestion
import hre from "@ignored/hardhat-vnext";

describe("Other mocha test", () => {
  it("should have the example task", () => {
    assert.ok(hre.tasks.getTask("example"));
  });
});
```

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -492,5 +492,12 @@ Please double check your script's path.`,
 
 Please check Hardhat's output for more details.`,
     },
+    TEST_TASK_ESM_TESTS_RUN_TWICE: {
+      number: 602,
+      messageTemplate: `Your project uses ESM and you've programmatically run your tests twice. This is not supported yet.`,
+      websiteTitle: "Running tests twice in an ESM project",
+      websiteDescription:
+        'You have run your tests twice programmatically and your project is an ESM project (you have `"type": "module"` in your `package.json`, or some of your files have the `.mjs` extension). This is not supported by Mocha yet.',
```
galargh, line 500: Is there a Mocha issue for that that we could link to?

v-next/hardhat-mocha-test-runner/dist/src/index.d.ts
galargh, line 1: It looks like the `dist` directory, which we intend to gitignore, got added to the branch here.

v-next/hardhat-mocha-test-runner/node_modules/.bin/_mocha
galargh, line 1: So did the `node_modules`, which we should remove, too. We might want to force push that change so that they don't enter git history on our target branch.

v-next/hardhat-mocha-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  const mocha = new Mocha(mochaConfig);
+
+  files.forEach((file) => mocha.addFile(file));
+
+  // if the project is of type "module" or if there's some ESM test file,
+  // we call loadFilesAsync to enable Mocha's ESM support
+  const projectPackageJson = await readClosestPackageJson(import.meta.url);
+  const isTypeModule = projectPackageJson.type === "module";
+  const hasEsmTest = files.some((file) => file.endsWith(".mjs"));
```
galargh, line 86: Should we check `.mts` as well?

v-next/hardhat-mocha-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,109 @@
...
+    testsAlreadyRun = true;
+
+    // This instructs Mocha to use the more verbose file loading infrastructure
+    // which supports both ESM and CJS
+    await mocha.loadFilesAsync();
+  }
+
+  await new Promise<number>((resolve) => {
+    mocha.run(resolve);
+  });
```
galargh, line 106: Is it an exit code that we get back from mocha? Do we have a mechanism to report it back to the user/caller? In particular, do we want the test task to fail if there are errors reported?

galargh (commented):
I left a couple of comments, but not much beyond what we discussed around node test runner :)

v-next/hardhat-errors/src/descriptors.ts
alcuadrado: That'd be cool to add to the websiteDescription

v-next/hardhat-mocha-test-runner/node_modules/.bin/_mocha
alcuadrado: I think you accidentally missed the `.gitignore`, @zoeyTM

v-next/hardhat-mocha-test-runner/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "eslint-plugin-import": "2.29.1",
+    "eslint-plugin-no-only-tests": "3.1.0",
+    "expect-type": "^0.19.0",
+    "prettier": "3.2.5",
+    "rimraf": "^5.0.5",
+    "typescript": "~5.5.0",
+    "typescript-eslint": "7.7.1"
+  },
+  "dependencies": {
+    "@ignored/hardhat-vnext-core": "workspace:^3.0.0-next.2",
```
alcuadrado, line 61: Should be `hardhat` instead. See the review of the other test plugin

v-next/hardhat-mocha-test-runner/src/index.ts
```diff
@@ -0,0 +1,19 @@
+import type { HardhatPlugin } from "@ignored/hardhat-vnext-core/types/plugins";
+
+import { task } from "@ignored/hardhat-vnext-core/config";
+
+const hardhatPlugin: HardhatPlugin = {
+  id: "test",
+  tasks: [
+    task("test", "Runs tests using the Mocha test runner")
+      .addVariadicArgument({
```
alcuadrado, line 9: Like in the other plugin, let's add `--grep` and `--bail`.

v-next/hardhat-mocha-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,109 @@
...
+
+import { HardhatError } from "@ignored/hardhat-vnext-errors";
+import { findUp, getAllFilesMatching } from "@ignored/hardhat-vnext-utils/fs";
+import { readClosestPackageJson } from "@ignored/hardhat-vnext-utils/package";
+
+interface TestActionArguments {
+  testFiles: string[];
+}
+
+function isTypescriptFile(path: string): boolean {
```
alcuadrado, line 15: The same is valid about typescript always being supported here.

v-next/hardhat-mocha-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,109 @@
...
+      );
+    }
+
+    try {
+      import.meta.resolve("tsx");
+    } catch {
+      throw new HardhatError(HardhatError.ERRORS.GENERAL.TSX_NOT_INSTALLED);
+    }
+
+    process.env.NODE_OPTIONS = "--import tsx";
```
alcuadrado, line 71: Same here

v-next/hardhat-mocha-test-runner/src/task-action.ts
alcuadrado: I think we can assume that this is always true, as hardhat runs in `esm` mode.

v-next/hardhat-mocha-test-runner/src/type-extensions.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import type { MochaOptions } from "mocha";
+
+declare module "@ignored/hardhat-vnext-core/types/config" {
+  export interface HardhatUserConfig {
+    mocha?: MochaOptions;
+  }
+
+  export interface HardhatConfig {
+    mocha: MochaOptions;
```
alcuadrado, line 11: We need a hook to resolve this config

v-next/hardhat-mocha-test-runner/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "../../config-v-next/tsconfig.json",
+  "references": [
+    {
+      "path": "../core"
```
alcuadrado, line 5: Hardhat instead of core

alcuadrado (changes requested):
This is looking good. It requires a similar set of changes than the other plugin.

v-next/hardhat-errors/src/descriptors.ts
zoeyTM: This was copied directly from hardhat v2. If an issue exists, there wasn't one linked there

v-next/example-project/test/mocha/mocha-test.ts
```diff
@@ -0,0 +1,8 @@
+import assert from "node:assert/strict";
+import { describe, it } from "mocha";
```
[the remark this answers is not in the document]
zoeyTM: yep!

v-next/hardhat-errors/src/descriptors.ts
galargh: I've just found one! I think it might be it - https://github.com/mochajs/mocha/issues/2706

v-next/hardhat-mocha-test-runner/src/task-action.ts
```diff
@@ -0,0 +1,94 @@
...
+  // This instructs Mocha to use the more verbose file loading infrastructure
+  // which supports both ESM and CJS
+  await mocha.loadFilesAsync();
+
+  const testFailures = await new Promise<number>((resolve) => {
+    mocha.run(resolve);
+  });
+
+  if (testFailures > 0) {
+    process.exitCode = testFailures;
```
galargh, line 88: Same as in the node test runner plugin, I think we might want to use exit code 1 here.

```suggestion
    process.exitCode = 1;
```

v-next/hardhat-mocha-test-runner/src/hookHandlers/config.ts
```diff
@@ -0,0 +1,79 @@
...
+    })
+    .optional(),
+  require: z.array(z.string()).optional(),
+  isWorker: z.boolean().optional(),
+});
+
+export default async (): Promise<Partial<ConfigHooks>> => {
+  const handlers: Partial<ConfigHooks> = {
+    validateUserConfig: async (userConfig) => {
+      return validateUserConfigZodType(userConfig, mochaConfigType);
```
alcuadrado, line 56: I think this doesn't work, as the helper should receive a ZodType like

```
const userConfigType = z.object({
  mocha: z.optional(mochaConfigType)
})
```

But can we add a tiny test that validates one or two of these fields?

v-next/hardhat-mocha-test-runner/src/hookHandlers/config.ts
```diff
@@ -0,0 +1,79 @@
...
+      const resolvedConfig = await next(
+        userConfig,
+        resolveConfigurationVariable,
+      );
+
+      return {
+        ...resolvedConfig,
+        mocha: {
+          ...resolvedConfig.mocha,
+          ...userConfig.mocha,
```
alcuadrado, line 72:
```suggestion
          timeout: 40000,
          ...resolvedConfig.mocha,
          ...userConfig.mocha,
```

I think I'd expect something like this, setting the only default value we set in v2.

v-next/example-project/package.json
```diff
@@ -22,8 +22,11 @@
   },
   "devDependencies": {
     "@ignored/hardhat-vnext": "workspace:^3.0.0-next.3",
+    "@ignored/hardhat-vnext-mocha-test-runner": "workspace:^3.0.0-next.2",
     "@ignored/hardhat-vnext-node-test-runner": "workspace:^3.0.0-next.2",
+    "@types/mocha": ">=9.1.0",
     "@types/node": "^20.14.9",
+    "mocha": "^10.0.0",
```
alcuadrado, line 29: I think this is correct, but the `tsx` dependency shouldn't be needed.

Can you try rebasing this PR and removing `tsx` to make sure that it works and passes the updated dependencies linter?

v-next/hardhat-mocha-test-runner/test/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  assertRejectsWithHardhatError,
+  useFixtureProject,
+} from "@nomicfoundation/hardhat-test-utils";
+
+describe("Hardhat Mocha plugin", () => {
+  describe("Success", () => {
+    useFixtureProject("test-project");
+
+    afterEach(async () => {
+      const { _resetGlobalHardhatRuntimeEnvironment } = await import(
```
alcuadrado, line 15: I think that instead of this, you could use `@ignored/hardhat-vnext/hre` and initialize the HRE explicitly. That way, we don't use the global instance in our tests, and we don't need to export the reset function.

Would you mind doing that? I think we could also add this to the guidelines, as it sounds like something that every plugin would have to deal with.

alcuadrado (commented):
I just left two comments

alcuadrado (approved):

PR #5653, opened 2024-08-18
ci: start tracking hardhat dependencies

### LINKED ISSUE #5366: Dependency tracking metric
We have put significant effort into reducing and clarifying our dependencies in `v-next`. To build on this effort we should establish a metric or mechanism for measuring the number and size of our dependencies that we can use to benchmark the dependencies as the project progresses.

## Possible approach

Create a script that runs a clean install of the `v-next` Hardhat cli package and then counts the number of packages in the dependency tree and the size on disk of those dependencies.

## TODO

- [x] Investigate best practices, does anyone else do this and how do they do it
- [x] Put together a design doc in notion (https://www.notion.so/nomicfoundation/Hardhat-Dependency-Tracking-4da1bbd17a7443b7a2ab9f9da1931903?pvs=4)
- [x] Implement the process
- [x] Include in the `v-next` CI

galargh:

---

Resolves #5366 

This PR implements the solution to dependency tracking initiative proposed in https://www.notion.so/nomicfoundation/Hardhat-Dependency-Tracking-4da1bbd17a7443b7a2ab9f9da1931903?pvs=4

The newly added job posts "sticky" comments on PRs, ensuring that there is no more than 1 comment generated by the job per PR.

An example of a comment - https://github.com/NomicFoundation/hardhat/pull/5653#issuecomment-2298481372

### DIFF
--- a/.github/workflows/v-next-ci.yml
+++ b/.github/workflows/v-next-ci.yml
@@ -50,6 +50,7 @@ jobs:
     runs-on: ubuntu-latest
     outputs:
       packages: ${{ steps.filter.outputs.changes }}
+      files: ${{ toJSON(steps.filter.outputs) }}
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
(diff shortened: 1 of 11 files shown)


galargh:
The reported total size seems too big, looking into it.

alcuadrado (approved):

Issue #5579, opened 2024-08-06
hardhat-keystore


kanej:
Implement the new config variables support, within an initially uncrypted keystore file in the user global Hardhat config directory. This will be provided as a new Hardhat plugin `hardhat-keystore`.

The details design doc can be found here: https://www.notion.so/nomicfoundation/New-plugin-hardhat-keystore-ec80978a6de54098b60350a436b3e666

There will be follow on work to make the keystore fully encrypted.

PR #5654, opened 2024-08-19
Hardhat-keystore

ChristopherDedominici:
[Task description](https://github.com/NomicFoundation/hardhat/issues/5579)

This PR implements the main logic for the the Hardhat keystore. The encryption part will be added in a separate PR.

The technical design doc can be found [here](https://www.notion.so/nomicfoundation/Technical-design-doc-0fc1243d03f14723ba0638714f22f4d9).

## Manual Testing

The keystore plugin has been added to the example project.

You can run the following command from there:

```shell
npx hardhat keystore set mykey
# Enter secret to store: ******
# Key "mykey" set

npx hardhat keystore list
# Keys:
# mykey

npx hardhat keystore get mykey
# secret

DEBUG=* npx hardhat keystore list # debug to see the path to your local keystore file
```

## Implementation considerations

The plugin is modeled as:

- keystore (in memory representation of key/value store)
- keystore-file-loader (responsible for creating an in memory keystore, reading a keystore from file or writing a keystore to file)
- user interactions (an abstraction putting the messages display to the user and the user input requests in one place)

The configuration variable hooks read from a keystore if present, and defer to the next config variable hook otherwise. The configuration variables cache the keystore between runs with shared state in the hook. The loader is shared, which itself has caching capabilities.

The tasks have no task level caching, the intention being to run as isolated tasks.

We are setting the `process.exitCode` within tasks where there is a failure (e.g. getting a non-existant key). This leads to some ugliness in tests were the process exit code has to be reset.

The file read from disk by the loader is validated with zod based on the format specified in the types file.

In CI environments a "empty" keystore plugin is returned from the `./index.ts` file. It has not hooks or tasks, which effectively turns off the plugin in CI. A test has been added that asserts different results between local development and CI to test this case.

## TODO

- [x] Look at coverage on `configuration-variables.ts`, are we missing a test or do we reconsider logic
- [x] Delete the key argument checks - they are done by the task manager
- [x] Switch the filepath test helper Hardhat plugin to use dynamic imports to avoid the error
- [x] Double check we aren't reading/writing to the global hardhat folder during test runs
- [x] Add Zod validation when loading the unencrypted file
- [x] Convert HardhatPlugin errors  into Hardhat errors
- [x] Eliminate the keystore loader and keystore mocking in tests, replacing it with only integration tests and component level tests that mock UserInterruptions and FileManager interactions (@kanej)
- [x] revert the plugin to being inline
- [x] add a force flag on delete - it might need revist after we have more info on zero code exit
- [x] add debug entry in config.ts to show the keystore file path
- [x] Do we need to handle no zero exit codes from tasks - waiting for info - @ChristopherDedominici 
- [x] disable the keystore in CI (@kanej)
- [x] look at cache (@kanej)
- [x] Move `getConfigDir` to hh-utils (@ChristopherDedominici) - depends on [this PR](https://github.com/NomicFoundation/hardhat/pull/5734)
- [x] Update the CI check to be limited to config hook (@kanej)
- [x] Update comment on loader responsibilities (@kanej)
- [x] Ensure error rather than instance of (@ChristopherDedominici)
- [x] Cleanup validate password in `setupPassword` (@ChristopherDedominici)
- [x] Cleanup the user interactions abstractions and clarify interruptions usage (@kanej)


v-next/hardhat-keystore/src/methods.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  if (keystore === undefined) {
+    await setupKeystore();
+  }
+
+  if (!validateKey(key)) {
+    return;
+  }
+
+  if ((await isAuthorized()) === false) {
```
kanej, line 26: What is `isAuthorized` adding here? My expectation is there is no need for authorization, the user either provides a valid password for updating the keystore or doesn't?

v-next/hardhat-keystore/src/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+
+import { io } from "./io.js";
+import { PLUGIN_ID } from "./methods.js";
+import { setUpPassword } from "./password-manager.js";
+
+let keystoreCache: Keystore | undefined;
+let keystoreFilePath: string | undefined;
+
+// ATTENTION: For testing purposes
+export function setKeystoreCache(value: Keystore | undefined): void {
```
kanej, line 23: Lets look at how we remove the need for this. I suspect if we passed keystore state in to these helper functions we would avoid the need for cache global.

v-next/hardhat-keystore/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+      .build(),
+
+    task(["keystore", "get"], "Get a value given a key")
+      .addPositionalArgument({
+        name: "key",
+        type: ArgumentType.STRING,
+        description: "Specifies the key to retrieve the value for",
+      })
+      .setAction(async ({ key }) => {
+        await get(key);
```
kanej, line 47: Should this be returning a the value, so it can be used programmatically in a Hardhat script?

v-next/hardhat-keystore/src/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+function assertFilePath(fileP: string | undefined): asserts fileP is string {
+  if (fileP === undefined) {
+    throw new HardhatPluginError(
+      PLUGIN_ID,
+      "The filePath should be available at this point!",
+    );
+  }
+}
+
+async function getConfigDir(): Promise<string> {
```
kanej, line 137: Is this available in core or utils?

ChristopherDedominici: It is in core but we cannot use core methods in plugins. It will probably be moved in the hardhat package, so I could import it from there when ready. But in the meantime I declared it here

v-next/hardhat-keystore/src/index.ts
ChristopherDedominici: I think that the keystore should not be used programmatically considering its purpose. For example we do not want people to use it in CI because how would they input the password? 
@alcuadrado

v-next/hardhat-keystore/src/hook-handlers/configuration-variables.ts
```diff
@@ -0,0 +1,30 @@
...
+import { getKeystore } from "../utils.js";
+
+export default async (): Promise<Partial<ConfigurationVariableHooks>> => {
+  const handlers: Partial<ConfigurationVariableHooks> = {
+    fetchValue: async (
+      context: HookContext,
+      variable: ConfigurationVariable,
+      next,
+    ) => {
+      const keystore = await getKeystore();
```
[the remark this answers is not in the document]
ChristopherDedominici: Yes, good catch! I will handle the get in the hook in a different way

v-next/hardhat-keystore/src/methods.ts
ChristopherDedominici: It's a placeholder. It will be implemented/modified as soon as we have the specs for the encryption

v-next/hardhat-keystore/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "typescript": "~5.5.0",
+    "typescript-eslint": "7.7.1"
+  },
+  "dependencies": {
+    "@ignored/hardhat-vnext": "workspace:^3.0.0-next.2",
+    "@ignored/hardhat-vnext-errors": "workspace:^3.0.0-next.2",
+    "@ignored/hardhat-vnext-utils": "workspace:^3.0.0-next.2",
+    "@ignored/hardhat-vnext-zod-utils": "workspace:^3.0.0-next.2",
+    "chalk": "^5.3.0",
+    "env-paths": "^2.2.0",
```
kanej, line 68: We can get rid of this when we deal with calculation of the global cache.

v-next/hardhat-keystore/src/internal/utils/get-keystore-file-path.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { ensureDir } from "@ignored/hardhat-vnext-utils/fs";
+import envPaths from "env-paths";
+
+export async function getKeystoreFilePath(): Promise<string> {
+  const configDirPath = await getConfigDir();
+  return path.join(configDirPath, "keystore.json");
+}
+
+async function getConfigDir(): Promise<string> {
```
kanej, line 11: I suspect we shouldn't be handling this in the plugin directly.

Are we expecting Hardhat plugins to be able to write to the Hardhat config global folder?

## Options

1. Allow the direct writing to the folder using envPaths
2. Setup the global config path during hre initialation - should plugins know about this folder?
3. Provide it as a `hardhat-util` helper functions that can be used by Hardhat plugins

v-next/hardhat-keystore/src/index.ts
```diff
@@ -0,0 +1,63 @@
+import type { HardhatPlugin } from "@ignored/hardhat-vnext/types/plugins";
+
+import { task } from "@ignored/hardhat-vnext/config";
+import { ArgumentType } from "@ignored/hardhat-vnext/types/arguments";
+
+import "./internal/type-extensions.js";
+
+import { PLUGIN_ID } from "./internal/constants.js";
```
kanej, line 8: What should the pattern be here for PLUGIN_ID? It is used across the plugin - are we happy with this pattern of `constants.js` and importing it here?

v-next/hardhat-keystore/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,43 @@
...
+        userConfig,
+        resolveConfigurationVariable,
+      );
+
+      const defaultKeystoreFilePath = await getKeystoreFilePath();
+
+      return {
+        ...resolvedConfig,
+        keystore: {
+          filePath: defaultKeystoreFilePath,
```
kanej, line 36: This addition of a file path to config makes out testing far easier (we can set the keystore file).

v-next/hardhat-keystore/src/internal/loaders/file-manager.ts
```diff
@@ -0,0 +1,26 @@
+import type { FileManager, UnencryptedKeystoreFile } from "../types.js";
```
ChristopherDedominici, line 1: Can this be removed? It looks like a wrapper around a wrapper, maybe the whole logic can go inside the KeystoreFileLoader?

v-next/hardhat-keystore/src/internal/tasks/delete.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  await remove({ key }, keystoreLoader, interruptions);
+};
+
+export const remove = async (
+  { key }: TaskDeleteArguments,
+  keystoreLoader: KeystoreLoader,
+  interruptions: UserInteractions,
+): Promise<void> => {
+  checkMissingKeyTaskArgument(key, "keystore delete");
```
ChristopherDedominici, line 28: I think that this checks (also for the other tasks) is not needed. There is a default check to ensure that the key is passed

v-next/hardhat-keystore/src/internal/loaders/file-manager.ts
kanej: This is here entirely to enable tests. It narrows the API for mocking to the narrowest point we can get away with. My intent is that by having just this one point of file mocking we can remove a fair chunk of the helpers for higher level mocking in the tests.

v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (keystoreLoader === undefined) {
+        keystoreLoader =
+          await _setupLoaderWithContextBasedUserInterruptions(context);
+      }
+
+      if (!(await keystoreLoader.isKeystoreUninitialized())) {
+        return next(context, variable);
+      }
+
+      const keystore = await keystoreLoader.loadKeystore();
```
kanej, line 30: Do we want to guard against file loading errors (i.e. permissions problems) here and return the next value, or do we expose those exceptions immediately?

Currently we fail early.

v-next/hardhat-keystore/package.json
```diff
@@ -0,0 +1,75 @@
...
+    "hardhat"
+  ],
+  "scripts": {
+    "lint": "pnpm prettier --check && pnpm eslint",
+    "lint:fix": "pnpm prettier --write && pnpm eslint --fix",
+    "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
+    "prettier": "prettier \"**/*.{ts,js,md,json}\"",
+    "test": "node --import tsx/esm --test --test-reporter=@ignored/hardhat-vnext-node-test-reporter \"test/*.ts\" \"test/!(fixture-projects|helpers)/**/*.ts\"",
+    "test:only": "node --import tsx/esm --test --test-only --test-reporter=@ignored/hardhat-vnext-node-test-reporter \"test/*.ts\" \"test/!(fixture-projects|helpers)/**/*.ts\"",
+    "test:coverage": "c8 --reporter html --reporter text --all --exclude test --exclude src/internal/types.ts --exclude src/internal/ui/direct-user-interruption-manager.ts --src src node --import tsx/esm --test --test-reporter=@ignored/hardhat-vnext-node-test-reporter \"test/!(fixture-projects|helpers)/**/*.ts\"",
```
kanej, line 29: We added additional exclusions from c8 coverage here rather than putting them in the files.

The odd one is direct-user-interruption-manager which is a copy of the version under utils.

v-next/hardhat-keystore/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+// The keystore plugin should be disabled in CI environments.
+// We implement this by returning a Keystore plugin
+// with its functionality turned off (i.e. no hook handlers or tasks)
+const chosenPlugin: HardhatPlugin = isCi()
+  ? continuousIntegrationKeystorePlugin
+  : hardhatKeystorePlugin;
+
+export default chosenPlugin;
```
kanej, line 82: Note this is how CI disabling is implemented for the keystore plugin. We subtly replace the plugin returned on an `isCi()` detection.

The tests are careful to import not the `default` export but the full plugin directly.

v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
```diff
@@ -0,0 +1,50 @@
...
+  ConfigurationVariableHooks,
+  HookContext,
+} from "@ignored/hardhat-vnext/types/hooks";
+
+import { FileManagerImpl } from "../loaders/file-manager.js";
+import { KeystoreFileLoader } from "../loaders/keystore-file-loader.js";
+
+export default async (): Promise<Partial<ConfigurationVariableHooks>> => {
+  // Use a cache with hooks since they may be called multiple times consecutively.
+  let keystoreLoader: KeystoreLoader | undefined;
```
kanej, line 13: We can the loader here as shared state between runs of the handler.

It can't be initialized here as we need access to the hook context (for the user interruptions).

v-next/hardhat-keystore/src/internal/keystores/unencrypted-keystore.ts
```diff
@@ -0,0 +1,50 @@
...
+import { assertHardhatInvariant } from "@ignored/hardhat-vnext-errors";
+
+export class UnencryptedKeystore implements Keystore {
+  readonly #keystoreData: UnencryptedKeystoreFile;
+
+  constructor(keystoreData: UnencryptedKeystoreFile) {
+    this.#keystoreData = keystoreData;
+  }
+
+  public static createEmptyUnencryptedKeystoreFile(): UnencryptedKeystoreFile {
```
kanej, line 12: This is the factory for keystore files, it is used across the tests as well as in the production code.

v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
kanej: Note there is no caching in the tasks. I assume we don't expect running of the tasks from scripts?

v-next/hardhat-keystore/src/internal/loaders/keystore-file-loader.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import {
+  assertHardhatInvariant,
+  HardhatError,
+} from "@ignored/hardhat-vnext-errors";
+
+import { UnencryptedKeystore } from "../keystores/unencrypted-keystore.js";
+import { unencryptedKeystoreFileSchema } from "../types-validation.js";
+
+export class KeystoreFileLoader implements KeystoreLoader {
```
kanej, line 16: The loader takes responsibility for:

- reading from the file - including validating the file
- writing back an updated keystore
- caching the keystore

v-next/hardhat-keystore/src/internal/tasks/delete.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const remove = async (
+  { key, force }: TaskDeleteArguments,
+  keystoreLoader: KeystoreLoader,
+  interruptions: UserInteractions,
+): Promise<void> => {
+  if (!(await keystoreLoader.isKeystoreUninitialized())) {
+    process.exitCode = 1;
+    return interruptions.displayNoKeystoreSetErrorMessage();
```
kanej, line 30: I am not a fan of this pattern.

We have to add opt outs in the tests to deal with `process.exit`. I think we need framework level support in tasks for dealing with failed tasks.

v-next/hardhat-keystore/src/internal/types-validation.ts
```diff
@@ -0,0 +1,11 @@
+import type { UnencryptedKeystoreFile } from "./types.js";
+import type { ZodSchema } from "zod";
+
+import { z } from "zod";
+
+export const unencryptedKeystoreFileSchema: ZodSchema<UnencryptedKeystoreFile> =
+  z.object({
+    _format: z.literal("hh-unencrypted-keystore"),
+    version: z.number(),
```
kanej, line 9: Should this be:

```
_format: string,
_version: number
```

Or combine into `_format`?

kanej: It can be deferred for the full encrypted version.

v-next/hardhat-keystore/src/internal/type-extensions.ts
```diff
@@ -0,0 +1,9 @@
+import "@ignored/hardhat-vnext/types/config";
+
+declare module "@ignored/hardhat-vnext/types/config" {
+  export interface HardhatConfig {
```
kanej, line 4: The intent here is to expand the Hardhat config but not allow users to specify a file path to a keystore.

v-next/hardhat-keystore/src/internal/types.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+export interface KeystoreLoader {
+  isKeystoreUninitialized: () => Promise<boolean>;
+  createUnsavedKeystore: () => Promise<Keystore>;
+  loadKeystore: () => Promise<Keystore>;
+  saveToKeystoreFile: () => Promise<void>;
+}
+
+export interface FileManager {
```
kanej, line 25: This exists only for testing, creating a mocking boundary.

v-next/hardhat-keystore/src/internal/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  _format: Format;
+  version: number;
+  keys: {
+    [key: string]: string;
+  };
+}
+
+export interface Keystore {
+  listKeys(): Promise<string[]>;
+  hasKey(key: string): Promise<boolean>;
```
kanej, line 11: I added `hasKey` to separate out the checking for a key from the reading, as checking does not imply decrypting.

v-next/hardhat-keystore/src/internal/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  addNewValue(key: string, value: string): Promise<void>;
+  removeKey(key: string): Promise<void>;
+  readValue(key: string): Promise<string>;
+  toJSON(): UnencryptedKeystoreFile;
```
kanej, line 15: The keystore itself takes responibility for converting into a KeystoreFile object where that object is serializable by `JSON.stringify`.

v-next/hardhat-keystore/test/helpers/mock-file-manager.ts
```diff
@@ -0,0 +1,75 @@
...
+  public setKeystoreFile(keystoreFile: UnencryptedKeystoreFile): void {
+    this.#keystoreFile = keystoreFile;
+  }
+
+  public fileExists: Mock<(_absolutePath: string) => Promise<boolean>> =
+    mock.fn(async (_absolutePath: string): Promise<boolean> => {
+      return this.#keystoreFile !== null;
+    });
+
+  public writeJsonFile: Mock<
```
kanej, line 51: Some of the tests use `mock` from `node:test` to assert that a call wasn't made (i.e. write to file didn't happen).
Do we prefer an explicit custom `wasWriteJsonCalled` boolean to be added and updated, or are we happy with this `mock.fn` pattern?

v-next/hardhat-keystore/test/helpers/setup-keystore-file-location-override-at.ts
```diff
@@ -0,0 +1,34 @@
+import type { HardhatPlugin } from "@ignored/hardhat-vnext/types/plugins";
+
+export const setupKeystoreFileLocationOverrideAt = (
+  keystoreFilePath: string,
+): HardhatPlugin => {
+  const hardhatKeystoreFileLocationOverridePlugin: HardhatPlugin = {
+    id: "hardhat-keystore-file-location-override",
+    hookHandlers: {
```
kanej, line 8: This inline hookhandler triggers a warning about not having inline in production.

When we shifted it to dynamic imports we ran into the problem of passing in state (i.e. `keystoreFilePath`) to the hook handler.

v-next/hardhat-keystore/.eslintrc.cjs
```diff
@@ -0,0 +1,3 @@
+const { createConfig } = require("../../config-v-next/eslint.cjs");
+
+module.exports = createConfig(__filename, { onlyHardhatError: false });
```
alcuadrado, line 3: Why `onlyHardhatError: false`? I think it should be true.

v-next/hardhat-keystore/package.json
alcuadrado: I think we can move that functionality from `hardhat` to `hardhat-utils` instead.

v-next/hardhat-keystore/src/index.ts
alcuadrado: I think a better alternative is to just have

```
if (isCi()) {
  return next(...);
}
```

in the `fetchValue` hook handler.

Why? Because the current approach changes the CI setup significantly (e.g. some tasks disappear) and it's harder to communicate and understand.

Instead, if we just don't fetch the values, that'd be conceptually simpler.

v-next/hardhat-keystore/src/internal/hook-handlers/config.ts
alcuadrado: No, in general, we shouldn't change the product just for testing.

In this case in particular, this can lead to people using this path to point the keystone to a file within their repository, which is something we want to avoid, at least for the first version.

v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
alcuadrado: This works as expected. The reason we export factories, and not the hook handlers directly, is precisely to be able to do this.

v-next/hardhat-keystore/src/internal/types-validation.ts
alcuadrado: +1 on defering

v-next/hardhat-keystore/src/internal/loaders/keystore-file-loader.ts
alcuadrado: Maybe add this as a doc comment in its interface?

v-next/hardhat-keystore/src/internal/loaders/keystore-file-loader.ts
```diff
@@ -0,0 +1,88 @@
...
+    await this.#fileManager.writeJsonFile(this.#keystoreFilePath, keystoreFile);
+  }
+
+  #throwIfInvalidKeystoreFormat(keystore: UnencryptedKeystoreFile): void {
+    try {
+      unencryptedKeystoreFileSchema.parse(keystore);
+    } catch (error) {
+      throw new HardhatError(
+        HardhatError.ERRORS.KEYSTORE.INVALID_KEYSTORE_FILE_FORMAT,
+        error instanceof Error ? error : undefined,
```
alcuadrado, line 84: Please use `ensureError` instead of this

v-next/hardhat-keystore/src/internal/utils/get-keystore-file-path.ts
alcuadrado: Already got to this in a comment above. I think we should go with 3.

v-next/hardhat-keystore/src/internal/ui/direct-user-interruption-manager.ts
```diff
@@ -0,0 +1,87 @@
+import type { UserInterruptionManager } from "@ignored/hardhat-vnext/types/user-interruptions";
+
+import { HardhatError } from "@ignored/hardhat-vnext-errors";
+
+/**
+ * A class that handles user interruptions directly on the console
+ * and explicitly does NOT go through the user interruptions
+ * system of the Hook system.
+ */
+export class DirectUserInterruptionManager implements UserInterruptionManager {
```
alcuadrado, line 10: I don't get this. I think it shouldn't exist, it conflicts with the user interruptions system. It's meant to be used "as is" by a plugin like this, so that others can override its behavior.

v-next/hardhat-keystore/src/internal/tasks/get.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const get = async (
+  { key }: TaskGetArguments,
+  keystoreLoader: KeystoreLoader,
+  interruptions: UserInteractions,
+): Promise<void> => {
+  if (!(await keystoreLoader.isKeystoreInitialized())) {
+    process.exitCode = 1;
+    return interruptions.displayNoKeystoreSetErrorMessage();
```
alcuadrado, line 29: This is not an interruption, it's normal output coming from the main and only task being executed at this point in time.

Same for the rest.

The only real interruption from this plugin should be "hey, i need the password to the keystore", as that doesn't come from a task, but from a hook.

v-next/hardhat-keystore/src/internal/tasks/set.ts
```diff
@@ -0,0 +1,61 @@
...
+  interruptions: UserInteractions,
+): Promise<void> => {
+  if (!(await validateKey(key))) {
+    process.exitCode = 1;
+    return interruptions.displayInvalidKeyErrorMessage(key);
+  }
+
+  // TODO: Bring in the setup password flow when implementing
+  // the encrypted version.
+  // await interruptions.setUpPassword();
```
alcuadrado, line 36: Note that this is also not an interruption, despite it being requesting the password. It still comes from the main task.

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
```diff
@@ -0,0 +1,120 @@
...
+  public async displayNoKeysInfoMessage(): Promise<void> {
+    await this.#displayMessage("The keystore does not contain any keys.");
+  }
+
+  public async displayValueInfoMessage(value: string): Promise<void> {
+    await this.#displayMessage(value);
+  }
+
+  async #displayMessage(message: string): Promise<void> {
+    await this.#userInterruptions.displayMessage(PLUGIN_ID, message);
```
alcuadrado, line 111: I think this is a mistake; this is abusing the user interruption system. This should be just `console.log`.

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  async #requestSecretInput(inputDescription: string): Promise<string> {
+    return this.#userInterruptions.requestSecretInput(
```
alcuadrado, line 115: Instead of holding a reference to the user interruption manager I think this could just return the messages, so that it's approximately equally testable, and you don't have to couple this.

Note that if you do want to test that the interruption gets that message, you can set a dynamic hook handler from the test, assert the message, and then remove it.

v-next/hardhat-keystore/src/internal/tasks/delete.ts
alcuadrado: Can you expand on this?

v-next/hardhat-keystore/src/internal/utils/validate-key.ts
```diff
@@ -0,0 +1,9 @@
+export async function validateKey(key: string): Promise<boolean> {
+  const KEY_REGEX = /^[a-zA-Z_]+[a-zA-Z0-9_]*$/;
```
alcuadrado, line 2:
```suggestion
  TODO: Revist this once encryption is implemented.
  const KEY_REGEX = /^[a-zA-Z_]+[a-zA-Z0-9_]*$/;
```

alcuadrado: I'm not even sure if we should validate that. Special chars are good for passwords.

v-next/hardhat-keystore/test/helpers/mock-file-manager.ts
alcuadrado: I think this is fine

v-next/hardhat-keystore/test/helpers/setup-keystore-file-location-override-at.ts
alcuadrado: Let's ignore those warnings for now

v-next/hardhat-keystore/src/internal/utils/validate-key.ts
ChristopherDedominici: This regex is for the KEY validation (same naming convention as ENV variables), not the password. The password regex is [this one](https://github.com/NomicFoundation/hardhat/pull/5654/files#diff-5d0d1b07b3cc14c33477602b9a9261a21abf6ee2ef0a2774c98a64b5fb7f87be)

v-next/hardhat-keystore/src/internal/types-validation.ts
ChristopherDedominici: _format and _version removed [here](https://github.com/NomicFoundation/hardhat/pull/5654/commits/96c7c1caf690bc711f70bd9f55f1b991249ddab2)

v-next/hardhat-keystore/src/internal/utils/validate-key.ts
alcuadrado: ohhh, got it!

v-next/hardhat-keystore/src/internal/types-validation.ts
alcuadrado: Wait, why? I think they made sense The deferred decision was whether they should be one or two fields.

ChristopherDedominici: oh I though we were planning to introduce them later, I'll revert the commit

ChristopherDedominici: Btw, should we use `_version` instead of `version`?

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
```diff
@@ -0,0 +1,120 @@
...
+    await this.#displayMessage(setupMsg);
+    await this.#displayMessage(passwordRulesMsg);
+    await this.#displayMessage("");
+
+    let password: string | undefined;
+
+    while (password === undefined) {
+      password = await this.#requestSecretInput(`Enter your password: `);
+
+      if (!PASSWORD_REGEX.test(password)) {
```
alcuadrado, line 32: I think this is application logic, and shouldn't be hidden here in this UI-level module. I completely missed it until @ChristopherDedominici corrected me that the other regex validates the key, not the password.

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
```diff
@@ -0,0 +1,120 @@
...
+      "The password must have at least 8 characters, one uppercase letter, one lowercase letter, and one special character.";
+
+    await this.#displayMessage("\n👷🔐 Hardhat-Keystore 🔐👷\n");
+    await this.#displayMessage(setupMsg);
+    await this.#displayMessage(passwordRulesMsg);
+    await this.#displayMessage("");
+
+    let password: string | undefined;
+
+    while (password === undefined) {
```
alcuadrado, line 29: What's this while here? What does `undefined` mean?

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
```diff
@@ -0,0 +1,120 @@
...
+      password = await this.#requestSecretInput(`Enter your password: `);
+
+      if (!PASSWORD_REGEX.test(password)) {
+        password = undefined;
+        await this.#displayMessage(chalk.red("Invalid password!"));
+      }
+    }
+
+    let confirmPassword: string | undefined;
+    while (confirmPassword === undefined) {
```
alcuadrado, line 39: Same

v-next/hardhat-keystore/test/integration/tasks.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    keystoreFile.keys.myKey1 = "myValue1";
+    keystoreFile.keys.myKey2 = "myValue2";
+
+    await _overwriteKeystoreFileWith(keystoreFilePath, keystoreFile);
+
+    hre = await createHardhatRuntimeEnvironment({
+      plugins: [
+        hardhatKeystorePlugin,
+        setupKeystoreFileLocationOverrideAt(keystoreFilePath),
```
alcuadrado, line 46: This is pretty clever.

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  public async setUpPassword(): Promise<string> {
+    const PASSWORD_REGEX = /^(?=.*[A-Z])(?=.*[a-z])(?=.*[\W_]).{8,}$/;
+
+    const setupMsg =
+      "This is the first time you are using the keystore, please set a password.";
+    const passwordRulesMsg =
+      "The password must have at least 8 characters, one uppercase letter, one lowercase letter, and one special character.";
+
+    await this.#displayMessage("\n👷🔐 Hardhat-Keystore 🔐👷\n");
```
alcuadrado, line 22: Please only add this kind of formatting after discussing it. We want to have a cohesive messaging style through Hardhat, which requires discussing it and designing it.

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
ChristopherDedominici: Small note: this code is currently not used, it will be used when the encryption is enabled.

To answer to your question:

There are two steps when setting up the password for the keystore:
1. the user must enter a password
2. the user must confirm the password

in both scenarios there is a while lookup, why?
- step 1: loop until the password is valid -it must match the regex (the process can be interrupted by the user)
- step 2: loop until the passwords do NOT match (the process can be interrupted by the user)

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
ChristopherDedominici: I created [an issue](https://github.com/NomicFoundation/hardhat/issues/5729) to track this

v-next/hardhat-keystore/.eslintrc.cjs
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/5654/commits/232307dd3a2d64e8a943f810a13c981e5ac4ee3f)

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
kanej: We have talked it over. This code is not run in this version, but we didn't want to lose it. For now it is used only from a test. So we could delete it and bring it back later.

However, we are suggesting the next steps of:

1. Create a util helper function `isValidPassword` that (for the moment) returns a boolean
2. Inject in the validation function as a parameter to `setupPassword`

v-next/hardhat-keystore/src/index.ts
kanej: Done with this commit: [`e7b5ff9` (#5654)](https://github.com/NomicFoundation/hardhat/pull/5654/commits/e7b5ff98e49a0234cb4a7a084d55e2df10b6914d)

kanej: Though I ended up having to put in some `process.env.CI` toggling magic in the tests as the Hook fetch value integration tests run in our CI.

v-next/hardhat-keystore/src/internal/tasks/delete.ts
kanej: It forces the plugin author to put reversing `process.exitCode = undefined` into tests where the task finishes in a state where task failure should be reported.

I can imagine three approachs:
* live with plugin authors having do `process.exitCode` twidling in their tests - though I think we should think through how this composes for `./scripts`.
* enhance the return results of task to allow passing along exitCode information, the test can then check for this while the Hardhat cli sets the `process.exitCode` - this complicates the return result for tasks called from scripts
* enhance the context that the tasks run in to provide a side channel for this information, so

```ts
context.setAsFailedTask()
return myTaskResult;
```

I don't think we have to solve this here - but I think writing a task properly involves specifying which return results should be interpreted as a fail for the cli task.

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
kanej: Based on the UserInteractions and interruptions comments I made the following moves:

* Changed user interactions into `UserDisplayMessages` where each static method just returns the string of the method
* In `get`, `list`, `delete`, `set` use poor man's dependency injection to inject a `consoleLog` defaulting to `console.log` allowing the tests to be ported
* Remove the interruptions class that is no longer needed.


[`f0efbb0` (#5654)](https://github.com/NomicFoundation/hardhat/pull/5654/commits/f0efbb0df60a0317da4f5bf4abf88682c6474ece)

v-next/hardhat-keystore/src/internal/ui/request-secret-input.ts
```diff
@@ -0,0 +1,53 @@
+import { HardhatError } from "@ignored/hardhat-vnext-errors";
+
+export async function requestSecretInput(
```
alcuadrado, line 3: For a follow-up PR: I think we should move this to utils, we have similar things in a few places, including one implementation that uses enquirer.

v-next/hardhat-keystore/src/internal/ui/user-interactions.ts
alcuadrado: To be revisited when we implement encryption

alcuadrado (approved):

PR #5658, opened 2024-08-20
refactor!: Remove the experimental MessageTrace hook

Xanewok:
This removes the `MessageTrace` from the (experimental) public API and only hides it behind our internal API, where it is used for stack trace logic.

We want to move this type to Rust to decrease the overhead of the value passing across the FFI and it would allow us to remove/not export this type entirely from Rust when porting the stack trace logic.

@fvictorio you mentioned that we broke this since the EDR-powered release and we have not heard back from anyone that this is broken, correct?

cc @fvictorio @alcuadrado since we talked about this recently


Xanewok:
Let me know what kind of breakage should we associate this with in our changesets. We should probably mention this but I don't think anything above patch makes sense as this was very explicitly experimental and the note stating it's about to be removed has been there for few years already, it seems.

fvictorio:
> Let me know what kind of breakage should we associate this with in our changesets. We should probably mention this but I don't think anything above patch makes sense as this was very explicitly experimental and the note stating it's about to be removed has been there for few years already, it seems.

Patch is fine.

> is this expected?

That seems like a false negative.

fvictorio (approved):

PR #5661, opened 2024-08-21
ci: test:integration if the task is available

galargh:

---

This PR enables running `test:integration` task in CI (if the task is present for the package). This currently affects only the test reporter. It was discovered as part of #5437

This PR will be ready for review once the CI is 🟢


galargh:
I've only just realised we're already doing this! Closing now.

PR #5664, opened 2024-08-21
Add await in provider creation

fvictorio:
In https://github.com/NomicFoundation/hardhat/pull/5478 I added some debug logs to investigate a false negative we have now and then on EDR's CI. But I forgot to await the provider creation, so these logs are pretty useless as they are.

This PR just adds the proper await.


schaable (approved):
LGTM

PR #5670, opened 2024-08-24
docs: add pnpm to document

iosh:

---

- Add `pnpm` to document
- Change 'frameborder' to 'frameBorder' to resolve terminal warning.

~~By default, pnpm does not automatically install peer dependencies. To enable this behavior, users must set auto-install-peers=true in the .npmrc file. So I need to list all packages as I would when using yarn.~~

Update: pnpm is automatically install peer dependencies, if user want use the package in code, user  need to add it as a dependency

### DIFF
--- a/docs/src/content/hardhat-runner/docs/advanced/create-task.md
+++ b/docs/src/content/hardhat-runner/docs/advanced/create-task.md
@@ -79,7 +79,7 @@ yarn add --dev @nomicfoundation/hardhat-toolbox @nomicfoundation/hardhat-network
 :::tab{value="pnpm"}
 
 ```
-pnpm add -D @nomicfoundation/hardhat-toolbox @nomicfoundation/hardhat-network-helpers @nomicfoundation/hardhat-chai-matchers @nomicfoundation/hardhat-ethers @nomicfoundation/hardhat-verify chai@4 ethers hardhat-gas-reporter solidity-coverage @typechain/hardhat typechain @typechain/ethers-v6
+pnpm add -D @nomicfoundation/hardhat-toolbox @nomicfoundation/hardhat-network-helpers chai@4 ethers
 ```
 
 :::
--- a/docs/src/content/hardhat-runner/docs/advanced/create-task.md
+++ b/docs/src/content/hardhat-runner/docs/advanced/create-task.md
@@ -50,7 +50,7 @@ Tasks in Hardhat are asynchronous JavaScript functions that get access to the [
 
 For our example, we will use the [`@nomicfoundation/hardhat-toolbox`](/hardhat-runner/plugins/nomicfoundation-hardhat-toolbox), which includes the [ethers.js](https://docs.ethers.org/v6/) library to interact with our contracts.
 
-::::tabsgroup{options="npm 7+,npm 6,yarn"}
+::::tabsgroup{options="npm 7+,npm 6,yarn,pnpm"}
 
 :::tab{value="npm 7+"}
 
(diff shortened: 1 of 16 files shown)


alcuadrado:
> By default, pnpm does not automatically install peer dependencies.

According to the docs it does install peer dependencies: https://pnpm.io/npmrc#auto-install-peers

iosh:
> > By default, pnpm does not automatically install peer dependencies.
> 
> According to the docs it does install peer dependencies: https://pnpm.io/npmrc#auto-install-peers

User need to add auto-install-peers=true to the `.npmrc` file .

alcuadrado:
I think you may be correct. I created this issue in the pnpm repo to better understand the situation: https://github.com/pnpm/pnpm/issues/8463

alcuadrado:
I believe you are right, @iosh, and we need to add pnpm to the doc and manually list most peer dependencies, at least the ones that the user would require/import.

iosh:
> I believe you are right, @iosh, and we need to add pnpm to the doc and manually list most peer dependencies, at least the ones that the user would require/import.

I tried npm, which allows users to import/require packages not added to package.json. However, pnpm doesn't allow this, so I think this PR is  okay.

docs/src/content/hardhat-runner/docs/advanced/create-task.md
```diff
@@ -76,6 +76,14 @@ yarn add --dev @nomicfoundation/hardhat-toolbox @nomicfoundation/hardhat-network
 
 :::
 
+:::tab{value="pnpm"}
+
+```
+pnpm add -D @nomicfoundation/hardhat-toolbox @nomicfoundation/hardhat-network-helpers @nomicfoundation/hardhat-chai-matchers @nomicfoundation/hardhat-ethers @nomicfoundation/hardhat-verify chai@4 ethers hardhat-gas-reporter solidity-coverage @typechain/hardhat typechain @typechain/ethers-v6
```
alcuadrado, line 82: My understanding is that with pnpm you don't need to install all the peer dependencies, but only the ones that you will be importing directly. I believe that in this case it'd be something like 

```
pnpm add -D @nomicfoundation/hardhat-toolbox @nomicfoundation/hardhat-network-helpers chai@4 ethers
```

alcuadrado (approved):
Thanks again for this PR, and for your patience @iosh. I was confused about the topic. 

I believe that we can be a bit less verbose (see my comment), but we can play safe and do the full list for now. WDYT?

iosh:
@alcuadrado  Thank you for your guidance, I agree with your suggestions. I have another question - I want to contribute to Hardhat, but I am unsure where to start. Do you have any suggestions for me? I want to spend some time to Hardhat every week.

alcuadrado:
> at least the ones that the user would require/import.

We've got a confirmation from a maintainer of pnpm https://github.com/pnpm/pnpm/issues/8463#issuecomment-2325409745

> I want to contribute to Hardhat, but I am unsure where to start. Do you have any suggestions for me? I want to spend some time to Hardhat every week.

Maybe look for simple bugs in the issue and tag any of us to learn more about it. Or hit me up on telegram.

iosh:
I have update the pnpm installation command to only list the packages that the user needs to require/import.

> Maybe look for simple bugs in the issue and tag any of us to learn more about it. Or hit me up on telegram.

Hardhat has many issues, I don't know which one i can do.  maybe  add like `good first issue` or `help wanted` label to issue. or just let me know this issues i can do....

alcuadrado:
This is a good one: https://github.com/NomicFoundation/hardhat/issues/5696

PR #5675, opened 2024-08-26
ci: trigger v-next ci on changes to scripts

galargh:

---

This PR ensures we run CI on PRs to v-next when we modify `scripts`. This was discovered in https://github.com/NomicFoundation/hardhat/pull/5671

There is a more general question whether we want to run the v-next CI on any PR to the v-next branch since we require `ci` and `lint` checks in merge queue. This PR doesn't address this fully.

### DIFF
--- a/.github/workflows/v-next-ci.yml
+++ b/.github/workflows/v-next-ci.yml
@@ -10,12 +10,14 @@ on:
       - "v-next/**"
       - "config-v-next/**"
       - "pnpm-lock.yaml"
+      - "scripts/**"
   pull_request:
     paths:
       - ".github/workflows/v-next-ci.yml"
       - "v-next/**"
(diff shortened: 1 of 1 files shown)


alcuadrado:
Sorry, I got confused and merged the other alternative, @galargh. 

I believe that running on every PR given that we filter out most jobs, shouldn't be too bad, but I'd go with your recommendation.

galargh:
> Sorry, I got confused and merged the other alternative, @galargh.
> 
> I believe that running on every PR given that we filter out most jobs, shouldn't be too bad, but I'd go with your recommendation.

That was exactly my thinking, too! But I wanted to present both options for this particular case. If it turns out it's too much, we can revisit this again and try to impose filters again. I'll close this now.

PR #5685, opened 2024-08-27
Render the node-test-reporter integration test results as SVG

alcuadrado:
@sz-piotr I went ahead and created a script to regenerate the fixtures and render them as SVG at the same time, so that they are easier to review on Github


galargh (approved):
This is lovely, thanks 🙇 I'll add the skip for node_internals in the original PR.

PR #5692, opened 2024-08-29
chore: implement additional test cases for the node test reporter (part 2)

galargh:

---

This is a follow-up to https://github.com/NomicFoundation/hardhat/pull/5669 and it is related to https://github.com/NomicFoundation/hardhat/issues/5437

This PR covers adding more test cases that we chatted about. I'll be proposing changes to the reporter's output in a separate PR(s).

Newly added test cases:
- an error thrown at the top level of a test file
- a `describe.only` and `it.only` executed without the `--test-only` flag
- a `test.only`, `describe.only` and `it.only` executed with the `--test-only` flag
- nested `it`s
- an error with nested causes
- a `test.todo` without a callback

The IT frameworks improvements that support the newly added test cases:
-  support for custom `RunOptions` per fixture test - This was needed to be able to test `test.only` with the `--test-only` flag enabled. If one puts a `options.json` file inside the fixture test, it is going to be parsed by both the IT test runner, as well as our script that regenerates the fixtures.
- saving the actual output produced by the test when ran by the IT test runner - I found it useful to have the output saved to a file on a disk. I save it as `result.actual.txt` inside the fixture test and I gitignore these files because they do not need to be checked in.
- accept a fixture name as an input to regenerate fixtures script - I found it helpful to be able to regenerate fixtures selectively.
- normalize lines like `at async node:...` - We missed that flavour of node internal traces before.
- remove all the lines from the output until we encounter one starting with `Node.js` - This, currently, only affects the test case where we throw an error at the top level of the test file. In such a case, the output first lists the error details, then prints the node version and then continues to print the output from the reporter. I decided to propose removing that output from the checked-in results instead of just skipping it during normalization because we do not have control over that part in the reporter.
    - before:
        ```
        /Users/galargh/GitHub/NomicFoundation/hardhat/v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/execution-failure-test/test.ts:1
        throw new Error("error");
              ^


        Error: error
            at <anonymous> (/Users/galargh/GitHub/NomicFoundation/hardhat/v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/execution-failure-test/test.ts:1:7)
            at ModuleJob.run (node:internal/modules/esm/module_job:262:25)
            at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:485:26)
            at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:109:5)

        Node.js v22.4.1

          [31m1) integration-tests/fixture-tests/execution-failure-test/test.ts[39m [31m[3m(98ms)[23m[39m


        [32m0 passing[39m[90m (130ms)[39m[31m[39m
        [31m1 failing[39m

          1) integration-tests/fixture-tests/execution-failure-test/test.ts:

          [31mTest file execution failed (exit code 1).[39m
          [90m    Did you forget to await a promise?[39m

        ```
    - after:
        ```

          [31m1) integration-tests/fixture-tests/execution-failure-test/test.ts[39m [31m[3m(98ms)[23m[39m


        [32m0 passing[39m[90m (130ms)[39m[31m[39m
        [31m1 failing[39m

          1) integration-tests/fixture-tests/execution-failure-test/test.ts:

          [31mTest file execution failed (exit code 1).[39m
          [90m    Did you forget to await a promise?[39m

        ```


v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/execution-failure-test/test.ts
```diff
@@ -0,0 +1 @@
+throw new Error("error");
```
alcuadrado, line 1: Note for a future pr: we need to change the output generated in this case, as it has nothing to do with promises.

v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/only-test-with-the-flag/test.ts
```diff
@@ -0,0 +1,21 @@
+import assert from "node:assert/strict";
```
alcuadrado, line 1: I think adding another file to this test case would be useful. Something like:

`ignored-tests.ts`:
```ts
import { describe, it, test } from "node:test";

describe("test excluded because they don't have a .only", () => {
   it("ignored", () => {});
});
```

The reason I think it's worthwhile is I believe that right now we'd display it as a passed test, which we should avoid in a future pr.

alcuadrado (commented):
Looks good! I like that options idea.

I just left a comment about one test.

v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/only-test-with-the-flag/test.ts
galargh: That's a great point! I added test cases that should be ignored when run with the flag to the fixture that is generated without the flag, but forgot to add it here 🤦 I added it, now, in https://github.com/NomicFoundation/hardhat/pull/5692/commits/fb775b7988cc083212bd55aa38dff95f85f08e28

galargh:
I also added a test case for an unawaited `it` inside an `it` - https://github.com/NomicFoundation/hardhat/pull/5692/commits/473cb03214209ce325ad5d977e71e5e7714faae2

It does seem to work just fine, though.

alcuadrado:
> It does seem to work just fine, though.

I think we need to force it to go through the event loop. I somehow was getting this error, that's why I added an error message related to promises when you throw from the top-level 😅

I think we can merge this and tackle that on a follow up pr when fixing that error message.

alcuadrado (approved):

PR #5693, opened 2024-08-29
Zod utils improvements

alcuadrado:
This PR introduces a few improvements to the zod utils package and how we use it:

- It disables `isolatedModules` in that package as it's super inconvenient and just an internal package.
- It introduces a `unionType` helper, equivalent to `z.union` but with a custom error message.
- It introduces a `sensitiveUrlType,` which is a variant of `sensitiveStringType.`
- It adds linter rules to prevent us from using `z.union` in the future.

While building this, I discovered a misconfiguration in `eslint`, so I fixed the config and every new linter error I found after that.

### DIFF
--- a/config-v-next/eslint.cjs
+++ b/config-v-next/eslint.cjs
@@ -257,6 +257,12 @@ function createConfig(
           message:
             "Don't use assert.doesNotReject. Just await the async-function-call/promise directly, letting the error bubble up if rejected",
         },
+        {
+          selector:
+            "CallExpression[callee.object.name='z'][callee.property.name=union]",
+          message:
+            "Use the unionType helper from the zod utils package instead, as it provides better error messages.",
+        },
       ],
       "@typescript-eslint/restrict-plus-operands": "error",
       "@typescript-eslint/restrict-template-expressions": [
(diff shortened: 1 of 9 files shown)


config-v-next/eslint.cjs
```diff
@@ -426,7 +426,7 @@ function createConfig(
       files: ["src/**/*.ts"],
       rules: {
         "no-restricted-syntax": [
-          "error",
+          ...config.rules["no-restricted-syntax"],
```
schaable, line 429: 🤦‍♂️ good catch!

v-next/hardhat-zod-utils/src/index.ts
```diff
@@ -1,56 +1,47 @@
...
 
+/**
+ * A Zod untagged union type that returns a custom error message if the value
+ * is missing or invalid.
+ */
+export const unionType = (
+  types: Parameters<typeof z.union>[0],
+  errorMessage: string,
+) =>
+  // eslint-disable-next-line no-restricted-syntax -- This is the only place we allos z.union
```
schaable, line 15:
```suggestion
  // eslint-disable-next-line no-restricted-syntax -- This is the only place we allow z.union
```

schaable (approved):
Looks good! I committed a typo fix.

alcuadrado:
Thanks for fixing that typo and reviewing the PR!

config-v-next/eslint.cjs
alcuadrado: All our `no-restricted-syntax` magic was being skipped! 😅

PR #5695, opened 2024-08-30
Move `core` into `hardhat`

alcuadrado:


alcuadrado:
Comment about zod utils tests: it could be `unknown` instead of `any`. We should document why the utils expects a config (i.e. the path starts from the root if you do that)

alcuadrado:
This was reviewed with @schaable on a call.

PR #5698, opened 2024-08-31
`core` in `hardhat` continuation

alcuadrado:
This PR is the continuation of the refactor to include `core` as part of `hardhat`.

It cleans up a few different things:

- `re-exports` clean up: There were multiple re-exports of types and values, as they were previously exported from `core`. Now, only the exported modules (i.e. those in `src/`) re-export things.
- Moved all the logic in the exported modules into `internal/`.
- I cleaned up the configuration loading and HRE initialization logic.
- I moved some unrelated things out of the `core/` directory.
- Implemented a change to make sure that the built-in plugins' type-extensions are loading in every module that `hardhat` exports.
- Removed an outdated linter rule.
- Fixed all TODOs I had left in the tests, primarily by using the HRE instead of trying to mock it.


v-next/hardhat/test/internal/core/hook-manager.ts
```diff
@@ -36,11 +35,13 @@ describe("HookManager", () => {
 
   describe("plugin hooks", () => {
     describe("running", () => {
-      let hookManager: HookManager;
+      let hre: HardhatRuntimeEnvironment;
+      let foceConfigValidationErrorFromPlugin: boolean;
```
alcuadrado, line 39: Typo here

alcuadrado:
I reviewed this with @kanej on a call

PR #5700, opened 2024-09-02
feat: replace util.inspect with a custom stack parser

galargh:

---

This PR updates how we format errors. It replaces usage of `util.inspect` with a custom error (stack trace) parser.

The implemented solution formats error chains recursively (see https://github.com/NomicFoundation/hardhat/blob/b94c936cb862a6dca841d7b9b00fd496a601e52b/v-next/hardhat-node-test-reporter/src/error-formatting.ts#L53 for a more detailed description). It is inspired by `util.inspect` and https://www.npmjs.com/package/@voxpelli/node-test-pretty-reporter (just as the initial implementation of the reporter).

Custom stack parser will allow us to have greater control over what we display to the users. Some examples:
- custom location formatting implemented in this PR (file URL handing, path relativisation, etc.; see https://github.com/NomicFoundation/hardhat/blob/b94c936cb862a6dca841d7b9b00fd496a601e52b/v-next/hardhat-node-test-reporter/src/error-formatting.ts#L228)
- hiding node/test reporter internals implemented in a follow-up PR (see https://github.com/NomicFoundation/hardhat/pull/5714)

This PR also adds 2 new integration test cases, relevant to stack trace parsing:
- circular references in the error chain
- aggregate errors

Finally, it adds some unit tests around stack parsing and location formatting.


v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -12,7 +11,17 @@ import {
   isTestFileExecutionFailureError,
 } from "./node-test-error-utils.js";
 
-// TODO: Clean up the node internal fames from the stack trace
+/**
+ * This interface represents the result of parsing a single stack trace line,
+ * e.g.: at Object.<anonymous> (file:///Users/user/project/test.js:1:1)
+ */
+interface StackReference {
+  context: string | undefined;
```
alcuadrado, line 19: Can we add docs to `context` and `location`? As they can feel a bit ambiguous.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -37,44 +46,99 @@ export function formatError(error: Error): string {
...
-      .substring(indexOfMessage + error.message.length)
-      .replace(/^(\r?\n)*/, "");
-  } else {
-    title = error.message;
-    stack = error.stack ?? "";
+/**
+ * This function takes an error and formats it into a human-readable string.
+ *
+ * The error is formatted as follows:
+ * - The error message is the first line of the error stack if the stack
```
alcuadrado, line 56: I'm not sure i understand the first line

v-next/hardhat-node-test-reporter/src/error-formatting.ts
alcuadrado: The doc in `parseStackLine` helped a ton to understand this

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -85,31 +149,96 @@ function isDiffableError(
...
+ * - at <context> (<location>)
+ * - at <location>:<lineNumber>:<columnNumber>
+ * - at <location>:<lineNumber>
+ * - at <location>
+ *
+ * @param line
+ * @returns
+ */
+export function parseStackLine(line: string): StackReference | undefined {
+  const regex = /^at (?:(.+?) \()?(?:([^\(].+?)(?::(\d+))?(?::(\d+))?)\)?$/;
```
alcuadrado, line 180: What do you think about doing something like this?

```ts
const regex = new RegExp(
  '^at ' +                         // Start with "at " 
  '(?:' +                           // Non-capturing group for optional context:
  '(.+?) \\(' +                     //   - Captures the context (lazy match) followed by " (" 
  ')?' +                            // End optional context group
  '(?:' +                           // Non-capturing group for location, line number, and column:
  '([^\\(].+?)' +                   //   - Captures the location (anything except a "(" at the start, lazy match)
  '(?::(\\d+))?' +                  //   - Optional: captures the line number prefixed by ":"
  '(?::(\\d+))?' +                  //   - Optional: captures the column number prefixed by ":"
  ')?' +                            // End non-capturing group for location and numbers
  '\\)?$',                          // Optional closing ")" at the end and end of the string
  ''
);
```

Note that it was generated by chatgpt, so it may be incorrect 😅

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -85,31 +149,96 @@ function isDiffableError(
...
+  if (location.startsWith("node:")) {
+    return location;
+  }
+  if (location === "<anonymous>") {
+    return location;
+  }
+  const locationPath = location.startsWith("file://")
+    ? fileURLToPath(location, { windows })
+    : location;
+  return locationPath.replace(`${cwd}${sep}`, "");
```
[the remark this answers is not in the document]
alcuadrado: It may even be worth a test case. Something like `/not/inside/<cwd>`. 

The current implementation would break that path.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -37,44 +46,99 @@ export function formatError(error: Error): string {
...
+ * @returns The formatted error
+ */
+function formatSingleError(
+  error: Error,
+  prefix: string = "",
+  depth: number = 0,
+): string {
+  const stackLines = (error.stack ?? "").split("\n");
+
+  let message = error.message.split("\n")[0];
```
alcuadrado, line 87: Why only the first line? We use a ton of multi-line error messages.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -37,44 +46,99 @@ export function formatError(error: Error): string {
...
+  if (stackLines.length > 0 && stackLines[0].includes(message)) {
+    message = stackLines[0];
   }
+  message = message.replace(" [ERR_ASSERTION]", "").replace(/:$/, "");
```
alcuadrado, line 91: I had already forgotten about this 😅 Maybe we should add this behavior to the doc block above.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -37,44 +46,99 @@ export function formatError(error: Error): string {
...
+  if (prefix !== "") {
+    message = `[${prefix}]: ${message}`;
+  }
 
-  const diffResult = getErrorDiff(error);
+  const diff = getErrorDiff(error);
+
+  const stackReferences: StackReference[] = stackLines
+    .map(parseStackLine)
+    .filter((reference) => reference !== undefined);
```
alcuadrado, line 101: Should we really filter these out, print them "as is", or in a slightly different format?

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -37,44 +46,99 @@ export function formatError(error: Error): string {
...
+
+  const stack = stackReferences.map(formatStackReference).join("\n");
+
+  let formattedError = depth === 0 ? chalk.red(message) : chalk.grey(message);
+  if (diff !== undefined) {
+    formattedError += `\n${diff}\n`;
+  }
+  if (stack !== "") {
+    formattedError += `\n${chalk.gray(indent(stack, 4))}`;
```
alcuadrado, line 110: Can we make the different numbers here constants? the different indentations, the max depth before truncating, etc.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -37,44 +46,99 @@ export function formatError(error: Error): string {
...
+  }
 
-  if (diffResult === undefined) {
-    return `${title}
-${stack}`;
+  if (isAggregateError(error)) {
+    // Only the first aggregate error in a chain survives serialization
```
alcuadrado, line 114: I feel like this is something important, but I can't understand it. Can you expand on it? In the comment I mean

v-next/hardhat-node-test-reporter/test/error-formatting.ts
```diff
@@ -0,0 +1,380 @@
+import assert from "node:assert/strict";
```
alcuadrado, line 1: There's now a `nop.ts` file in this same folder that can now be deleted.

v-next/hardhat-node-test-reporter/test/error-formatting.ts
```diff
@@ -0,0 +1,380 @@
...
+  {
+    line: "AssertionError [ERR_ASSERTION]: Multiplication result is not correct",
+    reference: undefined,
+  },
+  {
+    line: "",
+    reference: undefined,
+  },
+  {
+    line: "at (/home/user/project/index.js:10:7)",
```
alcuadrado, line 16:
```suggestion
    // This is an invalid case, as it doesn't have a context, yet the location is between parens
    line: "at (/home/user/project/index.js:10:7)",
```

alcuadrado (commented):
This looks good! I just left a few comments, mostly about docs. But LGTM

alcuadrado (approved):
Actually, if you want, merge it after addressing the comments.

There's this one comment, which we can discuss on slack https://github.com/NomicFoundation/hardhat/pull/5700/files#r1749235686

v-next/hardhat-node-test-reporter/src/error-formatting.ts
galargh: Sure, TDLR is that if we do something like this in a test:
```
throw new AggregateError([new AggregateError([new Error("A"), new Error("B")])])
```

Then, the object that we get in the reporter is something like this (notice missing `errors` in the nested `AggregateError`):
```
{
  name: "AggregateError"
  errors: [
    {
      name: "AggregateError"
    }
  ]
}
```

I'll reword the comment for clarity.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
galargh: That's a great idea, I'll add it - I'll double check if LLM was correct here ;)

v-next/hardhat-node-test-reporter/src/error-formatting.ts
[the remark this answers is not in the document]
galargh: Great catch! You're totally right about that.

galargh:
I implemented all the suggestions from the review, including passing on the unmatched stack lines as-is. Thank you so much for the review 🙇 It was really helpful :D

v-next/hardhat-node-test-reporter/src/error-formatting.ts
alcuadrado: oh, that's unfortunate. Hopefully they'll fix it.

PR #5708, opened 2024-09-03
fix(hardhat-viem): type consistency + retryCount 0 on devnet

TateB:

added some small type changes to allow for more extensibility, as well ensuring there are 0 retries when on developer networks to avoid wasting time when testing for errors

### Type changes

- added initialisation of `ContractTypesMap` as empty so it can be referenced without an initial contract compiled (e.g. for an external lib)
- moved the types under `HardhatRuntimeEnvironment.viem` to their own interface `HardhatViemHelpers` which allows the `viem` object to be extended with other types

### Retry count change

the default `retryCount` on the custom transport is 3, meaning in most cases RPC requests are retried 3 times.
when writing tests with expected reverts from write calls this is ok, since the `sendTransaction` function specifies a retry count of 0. however, when testing read calls with expected reverts this isn't the case, and there is time wasted in retrying calls that will always revert. 

to fix read call tests (and anything else i'm not aware of), i changed the retry count to be 0 when on dev nets. this should be fine since:
- dev nets are local and produce consistent results
- ethers doesn't have retries by default anyway

### DIFF
--- a/.changeset/kind-radios-begin.md
+++ b/.changeset/kind-radios-begin.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-viem": patch
+---
+
+Moved types to `HardhatViemHelpers` and initialized `ContractTypesMap` as empty for better extensibility. Improved performance by disabling retries in dev nets (thanks @TateB!)
--- a/packages/hardhat-viem/src/internal/clients.ts
+++ b/packages/hardhat-viem/src/internal/clients.ts
@@ -13,6 +13,26 @@ import type {
   WalletClient,
 } from "../types";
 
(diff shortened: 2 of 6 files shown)


schaable (approved):
Looks good! I’ve tested the following:

1) `ContractTypesMap` is accessible without compiling.
2) The `hre.viem` types are working as expected.
3) The viem sample project tests still pass.

schaable:
Released in [@nomicfoundation/hardhat-viem@2.0.5](https://github.com/NomicFoundation/hardhat/releases/tag/%40nomicfoundation%2Fhardhat-viem%402.0.5).

Issue #3272, opened 2022-10-13
RPC response crash for debug_traceTransaction due to stringify


Archethect:
Context:

I am setting up a local hardhat node in combination with blockscout to do some local tests.
When blockscout starts fetching data through the RPC endpoint, the hardhat node fails with the following error:

```
/usr/src/app/node_modules/hardhat/src/internal/hardhat-network/jsonrpc/handler.ts:155
    res.end(JSON.stringify(rpcResp));
                 ^
RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at JsonRpcHandler._sendResponse (/usr/src/app/node_modules/hardhat/src/internal/hardhat-network/jsonrpc/handler.ts:155:18)
    at Server.JsonRpcHandler.handleHttp (/usr/src/app/node_modules/hardhat/src/internal/hardhat-network/jsonrpc/handler.ts:48:12)
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
```

In my case the calls that brake it are:
```
{"jsonrpc":"2.0", "method":"debug_traceTransaction", "params":["0xe751022dd3074ae061939331ceca6b05e2f679c9f92aff6a9e9ae9a033b6fc84", {"tracer": "callTracer"}], "id":1}
```

Hardhat version: v2.10.1
Node version: v14

I suppose the issue is related to https://github.com/NomicFoundation/hardhat/pull/2512 and https://github.com/NomicFoundation/edr/issues/392

hadfieldn:
Running into this issue as well.

fvictorio:
Thanks for reporting this @Archethect, I can confirm this is a bug.

srinjoychakravarty:
Running into this exact same issue as of yesterday:

<img width="964" alt="Screenshot 2022-11-17 at 5 09 36 AM" src="https://user-images.githubusercontent.com/1481037/202420121-646952c1-b594-4b28-a64c-e37384171f53.png">


>     "hardhat": "^2.12.2",
package.json

SuperHero123123:
> Thanks for reporting this @Archethect, I can confirm this is a bug.

meet the same bug today, is there anyway to resolve this?

unparalleled-js:
I noticed you (or blockscout) is using `"tracer": "callTracer"`, but when looking at the code and the docs, I don't see that as an option anywhere. I would like to use the call tracer as well but can't seem to get it to work.

[...]

PR #5710, opened 2024-09-03
Add streaming response

Ivan-Feofanov:
Big JSON response fix for issue #3272

### DIFF
--- a/.changeset/warm-beans-tie.md
+++ b/.changeset/warm-beans-tie.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Use streaming response to avoide failure on big response data.
--- a/.changeset/warm-beans-tie.md
+++ b/.changeset/warm-beans-tie.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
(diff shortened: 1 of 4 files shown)


alcuadrado:
Great catch!

.changeset/warm-beans-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Use streaming response to avoide failure on big response data.
```
alcuadrado, line 5:
```suggestion
Hardhat node can now handle large response objects by streaming them.
```

fvictorio (approved):
Thanks @Ivan-Feofanov! I can confirm this works.

Something to keep in mind is that this still doesn't work if you connect to the Hardhat node using `hh --network localhost` and then send a `debug_traceTransaction` request, because at some point the response is converted into a string, and we ran into the same limitation on the other side. To handle that properly, we would need a streaming json _parser_, not just a stringifier.

But I think we can ignore that use case for now, since doing the same thing using the in-process network does work.

I'll wait for the CI to pass and then merge this.

PR #5713, opened 2024-09-03
`ethereum-cryptography` cleanup

alcuadrado:
This PR introduces two changes:

- It unpins the version of `ethereum-cryptography` in `hardhat`, which had no reason to be pinned.
- It uses `ethereum-cryptography` in `hardhat-utils` to be consistent with the dependencies that we use, hence decreasing the total number of dependencies.


alcuadrado:
Dependencies went from 46 to 38 :) /cc @sz-piotr

galargh (approved):
That's awesome 🥇

PR #5714, opened 2024-09-04
feat: hide node and test runner internals from the stack

galargh:

---

This PR builds up on custom stack trace parsing introduced in https://github.com/NomicFoundation/hardhat/pull/5700 and implements stack trace filtering. It adds a removal of the node stack references.

Only the node stack references grouped at the end of the stack are removed. They are not removed at all if there are no other stack lines to display to the user.

The stack traces should be much cleaner and easier to comprehend after this change.


v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -96,10 +97,27 @@ function formatSingleError(
...
+  let stackReferences: StackReference[] = stackLines
     .map(parseStackLine)
     .filter((reference) => reference !== undefined);
 
+  // Remove all the stack references exposing test runner internals
+  // User code starts after: (Suite|Test|TestHook|...).runInAsyncScope
+  const runInAsyncScopeIndex = stackReferences.findIndex((reference) => {
+    return (
+      reference.context !== undefined &&
+      reference.context.endsWith(".runInAsyncScope")
```
alcuadrado, line 109: Can we make this check a bit stricter?  Maybe also check that the location is `node:...`?

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -96,10 +97,27 @@ function formatSingleError(
...
+    );
+  });
+  if (runInAsyncScopeIndex !== -1) {
+    stackReferences = stackReferences.slice(0, runInAsyncScopeIndex);
+  }
+
+  // Remove all the stack references originating from node
```
alcuadrado, line 116: I think we should probably remove all the bottom ones, instead of all of them.

Imagine something like this:

```
- node:http/something/something.js:123
- myFileThatDoesAnHttpRequest.js:345
- node:something/else.js:78
```

Id't like to see 

```
- node:http/something/something.js:123
- myFileThatDoesAnHttpRequest.js:345
```

and not just

```
- myFileThatDoesAnHttpRequest.js:345
```

as the exception isn't coming from my code.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -96,10 +97,27 @@ function formatSingleError(
...
 
+  // Remove all the stack references exposing test runner internals
+  // User code starts after: (Suite|Test|TestHook|...).runInAsyncScope
+  const runInAsyncScopeIndex = stackReferences.findIndex((reference) => {
+    return (
+      reference.context !== undefined &&
+      reference.context.endsWith(".runInAsyncScope")
+    );
+  });
+  if (runInAsyncScopeIndex !== -1) {
```
alcuadrado, line 112: I think we should also check if it's the last entry, and not remove it in that case. Not sure if that would ever happen, but if it does, it would be confusing, as we would show no stack trace.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
alcuadrado: What if all the frames come from node? That'd be some sort of internal error, and in that case we probably don't want to remove anything. Maybe not even the test runner ones.

alcuadrado (commented):
This is starting to look so cool :) I just left a few comments about a few edge cases that I think we should treat differently, and, if possible, have tests for them.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
galargh: Actually, this is just a special case of a `node:...` reference now that I think about it. I think we can do without it altogether.

I'm going to remove it.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
galargh: I'll remove that check altogether. But I'll apply the sentiment to how we handle the node references i.e. if references from node is all we have, we should display them to the user.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
galargh: That's a great point. I'm going to update the proposed implementation to remove only continuously grouped node references, and not to remove them at all if there are only node references in the stack.

galargh (commented):
I really like the idea of removing only the node references grouped at the end of the stack, and only if the rest of the stack is not empty. I'm going to update the PR with that change shortly.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
alcuadrado: Oh, that's true! I hadn't thought about it

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -108,17 +109,43 @@ function formatSingleError(
...
       }
     } else {
-      stackLines.push(formatStackReference(reference));
+      const formattedStackReference = formatStackReference(reference);
+      // Put all the node references on a separate stack. If we don't encounter
+      // any non-node references after that, we will discard them
+      if (isNodeLocation(reference.location)) {
+        // Check if the location is strictly a node location if the node stack
+        // is empty. Otherwise, it is OK to accept <anonymous> and index locations
+        if (nodeLines.length > 0 || isNodeLocation(reference.location, true)) {
```
alcuadrado, line 129: Why is there an `||` here? I don't understand the first part.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -312,3 +336,17 @@ export function formatLocation(
     return locationPath;
   }
 }
+
+function isNodeLocation(location: string, strict: boolean = false): boolean {
```
alcuadrado, line 340: We should probably documment what strict means, or we'll forget it.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -108,17 +109,43 @@ function formatSingleError(
 ): string {
   const messageLines = [];
   const stackLines = [];
+  const nodeLines = [];
 
   for (const line of error.stack?.split("\n") ?? []) {
```
alcuadrado, line 114: It took me a while to understand why i found this harder to follow than what I was expecting, and I think I just got it.

If you were iterrating in reverse order, the algorithm is much simpler/closer to how we dicussed:

```ts
let foundNodeFrame = false;
let stackLines = [];

const lines = error.stack?.split("\n") ?? [];
for (let i = lines.length - 1; i >= 0; i--) {
  const line = lines[i];
  //...
  if (!isNodeLocation(reference.location) && !foundNodeFrame) {
    // We remove all the frames below the first non-node ones
    stackLines = [];
    foundNodeFrame = true; 
  }
   
  stackLines.push(formattedLine);
}

stackLines.reverse();
```

Note that this is pseudocode though, i haven't run it

v-next/hardhat-node-test-reporter/src/error-formatting.ts
galargh: I removed the notion of a strict check in the newest iteration. Instead, I added a new function [hasNodeContext](https://github.com/NomicFoundation/hardhat/pull/5714/commits/bfe40154dd252547b04e67bd3b47f47d30ecadf8#diff-561a47ff066e39f4e07ec5fdc8265f35643511c1776154bd916966b8f2add17aR365) which checks whether the context of a reference is for one of the node types. Currently, for the purpose of the function, `Array`, `Promise` and `SafePromise` are considered node types. We can explicitly expand this definition as we find new cases we want to support. I think this approach is safer and easier to understand.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
galargh: That's a fair point. It got a tad bit complicated with trying to extract lines from which we build up a message and discarding node lines, but not all of them, ...

This got me thinking and I put together another approach. I'm curious to see what you think about it.

First, I added a new interface [StackLine](https://github.com/NomicFoundation/hardhat/pull/5714/commits/bfe40154dd252547b04e67bd3b47f47d30ecadf8#diff-561a47ff066e39f4e07ec5fdc8265f35643511c1776154bd916966b8f2add17aR46) which can hold both the original line that we attempt to parse and a `StackReference` if we succeed to parse the line into it. With this we can [run all the lines through parseStackLine](https://github.com/NomicFoundation/hardhat/pull/5714/commits/bfe40154dd252547b04e67bd3b47f47d30ecadf8#diff-561a47ff066e39f4e07ec5fdc8265f35643511c1776154bd916966b8f2add17aR119) straight away and not worry wether we got undefined back.

Now, among all the parsed lines, we can find [the first one which has a defined stack reference](https://github.com/NomicFoundation/hardhat/pull/5714/commits/bfe40154dd252547b04e67bd3b47f47d30ecadf8#diff-561a47ff066e39f4e07ec5fdc8265f35643511c1776154bd916966b8f2add17aR122). With that in hand, we can [build a message](https://github.com/NomicFoundation/hardhat/pull/5714/commits/bfe40154dd252547b04e67bd3b47f47d30ecadf8#diff-561a47ff066e39f4e07ec5fdc8265f35643511c1776154bd916966b8f2add17aR132) from all the lines from the beginning up to that line.

We can also find [the last line which does **NOT** have a reference to node internals](https://github.com/NomicFoundation/hardhat/pull/5714/commits/bfe40154dd252547b04e67bd3b47f47d30ecadf8#diff-561a47ff066e39f4e07ec5fdc8265f35643511c1776154bd916966b8f2add17aR127). With that, we can easily [build the stack](https://github.com/NomicFoundation/hardhat/pull/5714/commits/bfe40154dd252547b04e67bd3b47f47d30ecadf8#diff-561a47ff066e39f4e07ec5fdc8265f35643511c1776154bd916966b8f2add17aR156) from all the lines between the first line with any stack reference and that line. With a special case being that the entire stack is node internals only.

Let me know what you think :)

alcuadrado (approved):
Nice change!

PR #5715, opened 2024-09-04
Remove hardhat-change-network plugin

0xNeshi:
The plugin is unmaintained, [asked if it should be removed](https://github.com/NomicFoundation/hardhat/pull/5703#issuecomment-2325571462) and got an [affirmative response](https://github.com/NomicFoundation/hardhat/pull/5703#issuecomment-2327205196).


alcuadrado (approved):
Thanks

PR #5716, opened 2024-09-04
Docs update config to allow access to rpc from external ip on specific port using hostname flag

AwaisBinKaleem:
- [x] added hostname flag in docs that was not available in docs.

---


alcuadrado:
Thanks! I just updated this a bit to make it more correct with how it works.

alcuadrado (approved):

PR #5724, opened 2024-09-05
chore: improve the test reporter's messaging

galargh:

---

Please note that this PR is currently based on https://github.com/NomicFoundation/hardhat/pull/5714

This PR implements improvements to our test reporter that were discussed in and around https://github.com/NomicFoundation/hardhat/pull/5669 and https://github.com/NomicFoundation/hardhat/pull/5692

Here's a breakdown of the proposed changes together with links to their UI diffs (if you prefer me to break the PR down into separate ones, please let me know):
1. If there is an error in a suite, display it in the tree view. In particular, this is relevant to when a suite fails in before or describe. Previously, an issue like that was only displayed in the summary at the end. After the change, there will also be an indication of a failure in the tree view. This is useful because, in such cases, the failure in the suite tends to be the more interesting one. It is introduced in https://github.com/NomicFoundation/hardhat/pull/5724/commits/d9bca644ff946e5cefe52cc23f358f971c4e68b1
2. Do not display top-level file passes. This tends to be an indication that we successfully imported a file and there are no tests defined in it. It seems to be a waste of space to display it. It is introduced in https://github.com/NomicFoundation/hardhat/pull/5724/commits/3957b42bf20a5422e598e933fb1b4524db6cd84f
3. Do not suggest that a user forgot to await a promise when file execution fails. Usually, the issue is different, e.g. the file throws, so the suggestion can be misleading. It is introduced in https://github.com/NomicFoundation/hardhat/pull/5724/commits/15a5eab55b5b22af7a2f0ca55ff966d1b03513a6
4. When displaying unused diagnostic messages, make sure they are unique. Also, make them hardhat specific. In particular, make sure the suggestion about rerunning only tests, uses the hardhat `--only` flag. It is introduced in https://github.com/NomicFoundation/hardhat/pull/5724/commits/270f889aebf57c8c7e2693941c452493ab2f2a47
5. Remove `ERR_TEST_FAILURE` from the formatted error details. Similarly to how we improve the assertion error message. It is introduced in https://github.com/NomicFoundation/hardhat/pull/5724/commits/bfb1415d52407d16655847ed69a39b95c34e7810
6. Do not show cancelled by parent errors in the summary. These can be resolved by fixing the issue in the parent. This is why I suggest to link to the details of the parent error from the test error instead. It is introduced in https://github.com/NomicFoundation/hardhat/pull/5724/commits/a698e685fc7c3f803ef8b8816009b0749c04aa9c
7. Show tests failing with cancelled by parent errors in grey. After introduction of 5., these tests duplicate failure index of parent errors. I think we should guide the user towards the parent error more by making the line of the test itself grey. It is introduced in https://github.com/NomicFoundation/hardhat/pull/5724/commits/bc0a738fa548cf15392df9d823f2fd36ac5c44f9

Additionally, I removed the usage of async describe block from hardhat tests since it doesn't work well with node's test reporter even though it is supported (https://github.com/NomicFoundation/hardhat/pull/5724/commits/fee6fb50e5909b8f263fa210331e29fbfebda5f5). We could think about making it into a linting rule, maybe?


alcuadrado:
> 1. If there is an error in a suite, display it in the tree view. In particular, this is relevant to when a suite fails in before or describe. Previously, an issue like that was only displayed in the summary at the end. After the change, there will also be an indication of a failure in the tree view. This is useful because, in such cases, the failure in the suite tends to be the more interesting one. It is introduced in [d9bca64](https://github.com/NomicFoundation/hardhat/commit/d9bca644ff946e5cefe52cc23f358f971c4e68b1)

This is a much better experience.

> Do not display top-level file passes. This tends to be an indication that we successfully imported a file and there are no tests defined in it. It seems to be a waste of space to display it. It is introduced in https://github.com/NomicFoundation/hardhat/commit/3957b42bf20a5422e598e933fb1b4524db6cd84f

yay!

> Do not suggest that a user forgot to await a promise when file execution fails. Usually, the issue is different, e.g. the file throws, so the suggestion can be misleading. It is introduced in https://github.com/NomicFoundation/hardhat/commit/15a5eab55b5b22af7a2f0ca55ff966d1b03513a6

Makes sense. See below my other comment about this.

> When displaying unused diagnostic messages, make sure they are unique. Also, make them hardhat specific. In particular, make sure the suggestion about rerunning only tests, uses the hardhat --only flag. It is introduced in https://github.com/NomicFoundation/hardhat/commit/270f889aebf57c8c7e2693941c452493ab2f2a47

I think this change is needed, but this implementation breaks the usage of this reporter outside of hardhat (e.g. in our own tests).

I think we can have two exports in the package:
* One is the one that we already have, and can be used by non-hardhat tests.
* The other one would be a function that constructs a reporter, and can receive params, like how the only flag is called.

This will be helpful in a comment below.

> Remove ERR_TEST_FAILURE from the formatted error details. Similarly to how we improve the assertion error message. It is introduced in https://github.com/NomicFoundation/hardhat/commit/bfb1415d52407d16655847ed69a39b95c34e7810

Makes sense

> Do not show cancelled by parent errors in the summary. These can be resolved by fixing the issue in the parent. This is why I suggest to link to the details of the parent error from the test error instead. It is introduced in https://github.com/NomicFoundation/hardhat/commit/a698e685fc7c3f803ef8b8816009b0749c04aa9c

This is great. Much cleaner.

> Show tests failing with cancelled by parent errors in grey. After introduction of 5., these tests duplicate failure index of parent errors. I think we should guide the user towards the parent error more by making the line of the test itself grey. It is introduced in https://github.com/NomicFoundation/hardhat/commit/bc0a738fa548cf15392df9d823f2fd36ac5c44f9

Great idea!

> I removed the usage of async describe block from hardhat tests since it doesn't work well with node's test reporter even though it is supported (https://github.com/NomicFoundation/hardhat/commit/fee6fb50e5909b8f263fa210331e29fbfebda5f5). We could think about making it into a linting rule, maybe?

Unfortunately, this doesn't work, and we need to support that case. The reason is that our users will definitely do that. For example:

```
describe("Testing contract Foo", async () => {
    const Foo = await getFooFactory();
    
    it(...);
})
```

This already happens in practice, I'm trying to get Mocha to show an error in those situations, as they don't support async describes, but node does.

I went deeper into the reporter's code, and I understand when this happens and how to detect it:

```ts
/**
 * Returns true the event represents a failure due to trying to create a subtest
 * when its parent has already finished.
 *
 * This normally happens if you are running in `only` and the parent is async,
 * as the parent gets cancelled but continues running an async operation, and
 * try to create a subtests after it.
 */
export function isSubtestCreationAbortedError(event: TestEvent): boolean {
  if (event.type !== "test:fail") {
    return false;
  }

  const error = event.data.details.error;

  if (!("code" in error) || !("failureType" in error)) {
    return false;
  }

  return (
    error.code === "ERR_TEST_FAILURE" &&
    error.failureType === "parentAlreadyFinished"
  );
}
```

It may also be generated when the users forget an `await` and/or call to `it/test` after some weird async behavior, so I don't think it's entirely safe to silence it.

What if we do this?
- Expose a separate entry point with options for hardhat, as mentioned above.
- Pass it a flag when running in `only` mode.
- Ignore these events in `only` mode.
- Print them otherwise.

(we would need to add tests for this)

alcuadrado:
Also, let's revert the change to our test that does that. If we are going to support it, it's better to dogfood it.

alcuadrado (commented):
I just left a few comments

v-next/hardhat-node-test-reporter/src/reporter.ts
```diff
@@ -138,6 +157,22 @@ export function hardhatTestReporter(
             break;
           }
 
+          // We do not want to display failures caused by the parent already
+          // having finished when running with the --test-only flag because
+          // they are false negatives.
+          // We still display the failed suites, because failures for the nested
+          // tests (cancelled by parent errors) were already displayed.
```
alcuadrado, line 164: This doesn't work, it looks like an error. I think a good trade-off would be to detect that situation and maybe display those canceled tests/suites as a particular case in the summary at the end of the test run if needed.

This is a complex case, so we can continue addressing it in a follow up PR.

galargh:
I'll look closer at it tomorrow and prepare the PR for a re-review properly, but I reverted any custom handling for the async describe case and the diff from node v22.8.0 looks quite promising - https://github.com/NomicFoundation/hardhat/pull/5724/commits/5599b925907e7efe20db99fdff268790d28b4533

galargh:
> I think this change is needed, but this implementation breaks the usage of this reporter outside of hardhat (e.g. in our own tests).
> 
> I think we can have two exports in the package:
> 
> One is the one that we already have, and can be used by non-hardhat tests.
> The other one would be a function that constructs a reporter, and can receive params, like how the only flag is called.

I think this was a great suggestion and I implemented it straight away. Now we have the following 2 exports:
- [hardhatTestReporter](https://github.com/NomicFoundation/hardhat/pull/5724/files#diff-65ba3306b59a9065745c45f2156502dff63cd8738ebafc4ad1dde91e3a821579R59) which accepts [RunOptions](https://github.com/NomicFoundation/hardhat/pull/5724/files#diff-a827f46c1d67d8833d6960464bbdae96b4318e26f425e1f18a286737d5b87981R4) and [HardhatTestReporterConfig](https://github.com/NomicFoundation/hardhat/pull/5724/files#diff-65ba3306b59a9065745c45f2156502dff63cd8738ebafc4ad1dde91e3a821579R31); the former will give us the ability to detect whether we're running with an only flag or not, for example; while the latter we can use to customise how the reporter works - to begin with, we can get a custom message to display when the runner encounters some .only tests
- [customReporter](https://github.com/NomicFoundation/hardhat/pull/5724/files#diff-65ba3306b59a9065745c45f2156502dff63cd8738ebafc4ad1dde91e3a821579R51) which is the default export and uses the defaults that should work for any node project

> It may also be generated when the users forget an await and/or call to it/test after some weird async behavior, so I don't think it's entirely safe to silence it.

After the experimentation and discussions we had on Slack, I do agree that silencing might not be the way forward in this case. Actually, I don't think we should do that even when running in the only mode. I now reverted the changes related to that silencing "parent already finished" errors.

> This doesn't work well, as it looks like an error. I think a good trade-off would be to detect that situation and maybe display those canceled tests/suites as a particular case in the summary at the end of the test run if needed.
> 
> This is a complex case, so we can continue addressing it in a follow up PR.

As agreed, this PR will not introduce any special behaviour for "parent already finished" errors as it's a complex issue without a clearly right way forward. We'll deal with that in a separate PR if needed.

One thing worth noting is that some of the issues we were dealing with were addressed in `v22.8.0` release of Node 🥳 We can see that clearly in this diff - https://github.com/NomicFoundation/hardhat/commit/5599b925907e7efe20db99fdff268790d28b4533

This should be the result of https://github.com/nodejs/node/pull/54423 as far as I can tell. With that PR in place, we can only see the "parent already finished" errors in nested async describes. Something like:

```js
describe("a suite", () => {
  // If we remove this, no tests will be executed and no failures will be reported.
  it.only("an only test", async () => {
  	assert.equal(1, 1);
  });

  describe("a nested suite with an async callback", async () => {
    const one = await new Promise((resolve) => setTimeout(() => resolve(1), 0));

    it("a nested test", async () => {
      assert.equal(one, 1);
    });

    it.only("a nested only test", async () => {
      assert.equal(one, 1);
    });
  });
});
```

The issue is also only visible (reported by the test runner) if there is some block marked as only at the same level as the nested async describe.

All in all, triggering the issue requires a muuuch more complex setup now. We could consider whether trying to handle that edge case is worth pursuing.

v-next/hardhat-node-test-reporter/src/reporter.ts
```diff
@@ -102,20 +109,32 @@ export function hardhatTestReporter(
...
             const parts = event.data.message.split(" ");
             const count = parseInt(parts[1], 10) - topLevelFilePassCount;
             event.data.message = `${parts[0]} ${count}`;
           }
+          if (event.data.message.startsWith("fail")) {
+            const parts = event.data.message.split(" ");
+            const count =
+              parseInt(parts[1], 10) - parentAlreadyFinishedFailureCount;
+            event.data.message = `${parts[0]} ${count}`;
+          }
```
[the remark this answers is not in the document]
galargh: That's a good catch. I now moved it to after `processGlobalDiagnostics` which I think is much cleaner, see [`8332321` (#5724)](https://github.com/NomicFoundation/hardhat/pull/5724/commits/83323218854d422c2dfd1021976ac26b69a60e9b)

v-next/hardhat-node-test-reporter/src/reporter.ts
galargh: I replied in a longer form in https://github.com/NomicFoundation/hardhat/pull/5724#issuecomment-2343028639 TLDR I removed any attempt to handle that case from this PR + the newest node makes the output for these cases a lot better (e.g. in `hardhat` package, `test:only` works as expected now).

alcuadrado (approved):
Great news about 22.8.0. Let's merge this now, and we'll see how this progresses.

Issue #4725, opened 2024-01-03
[hardhat-chai-matchers] Edge case with `not.to.be.reverted` and view functions


jubeira:
### Version of Hardhat

hardhat-chai-matchers 2.0.3 / hardhat-ethers 3.0.5

### What happened?

The (not) `reverted` matcher throws an error when calling a `view` function returning ``bytes32` that does not revert.

This is because the code inside `expect` actually returns a string, and `reverted.ts` interprets that return value as a hash. Then, when it tries to get the transaction receipt it fails at [this line](https://github.com/NomicFoundation/hardhat/blob/%40nomicfoundation/hardhat-chai-matchers%402.0.3/packages/hardhat-chai-matchers/src/internal/reverted/reverted.ts#L37).

Tested with latest tag `2.0.3`. Probably also happens with other types that can be interpreted as a hash?

### Minimal reproduction steps

Suppose I have library with a view function that doesn't revert, but returns `bytes32`:

Solidity:
```
contract Test {
    bytes32 private _value;

    function setValue(bytes32 value) external {
        _value = value;
    }

    function getValueBytes32() external view returns (bytes32) {
        return _value;
    }

    function getValueUint() external view returns (uint256) {
        return uint256(_value);
    }
}
```

Test:

```
it('is ok', async () => {
    await expect(test.getValueUint()).not.to.be.reverted;
});

it('throws', async () => {
    await expect(test.getValueBytes32()).not.to.be.reverted;
});
```

The error in the second scenario is:
```
HardhatChaiMatchersAssertionError: Assertion error: receipt should not be null
```

A simple workaround is to just not use `expect` in these cases:

```
it('is ok as a workaround', async () => {
    test.getValueBytes32(); // does not revert
});
```

But ideally `not.reverted` should work as with non-view functions and the other return values.

### Search terms

chai-matchers, not.to.be.reverted

fvictorio:
Makes sense, thanks for opening this!

RenanSouza2:
Hey @jubeira, 

First a disclaimer, I am not from nomic foundation so I am not speaking for them.

From what I looked into the code it would require a major refactor to encompass this case,

My suggestion is to simply not use an `expect` in this case as `expect(...).to.not.revert` is implicit as any error would stop the test

fvictorio:
> My suggestion is to simply not use an expect in this case as expect(...).to.not.revert is implicit as any error would stop the test

Yes, that's a very good point!

[...]

PR #5727, opened 2024-09-07
fix: hardhat-chai-matchers error message when event is overloaded

### LINKED ISSUE #5696: Support overload
### Describe the feature

See original issue [here](https://github.com/NomicFoundation/hardhat/issues/5690)

When there is an event with the same name in the parent contract, but the parameters are different, `to.emit (contractAddress, EventName)` will result in an error where the event cannot be found.

Repro example:

.sol code
```
contract Bar {
  event MyEvent(string);
}

contract Foo is Bar {
  event MyEvent(uint);

  function f() public {
    emit MyEvent(3);
  }
}
```

.js code - how to trigger the error
```
  const foo = await hre.ethers.deployContract("Foo");

  await expect(foo.f()).to.emit(foo, "MyEvent")
```

.js code - workaround to make the test pass 
```
it("test", async function () {
  const foo = await hre.ethers.deployContract("Foo");

  await expect(foo.f()).to.emit(foo, "MyEvent(uint)")
})
```

### Search terms

_No response_

iosh:

---

Throw an error if the provided event name is ambiguous and request a clear event name from the user.

This will be resolve #5696


alcuadrado:
Hey! This looks really good! Thanks a ton!

I'll let @schaable review it, as he's more familiar with this part of the codebase.

Maybe this issue will also be of your interest: https://github.com/NomicFoundation/hardhat/issues/4725

schaable (approved):
Looks great! I’ve tested the following scenarios:

1) A contract that inherits from another contract and adds an overloaded event to the one defined in the parent contract.
2) A contract that defines multiple overloads of the same event.
3) Calling `emit` with an unknown event.

Screenshots:
Before:
previous
1) `ambiguous event description` is shown:
after
2) `ambiguous event description` is shown:
multiple
3) `event doesn't exist` is shown:
event2

PR #5731, opened 2024-09-09
Fix provider events when init is called

fvictorio:
Provider events are not working when `provider.init` is called explicitly, instead of the provider being implicitly initialized with the first request. The problem was that the events were only forwarded in the `_getOrInitProvider` function instead of that being done in `init`. This PR fixes this and adds a test for that scenario.

### DIFF
--- a/packages/hardhat-core/src/internal/core/providers/lazy-initialization.ts
+++ b/packages/hardhat-core/src/internal/core/providers/lazy-initialization.ts
@@ -40,6 +40,19 @@ export class LazyInitializationProviderAdapter implements EthereumProvider {
         this._initializingPromise = this._providerFactory();
       }
       this.provider = await this._initializingPromise;
+
+      // Copy any event emitter events before initialization over to the provider
+      const recordedEvents = this._emitter.eventNames();
+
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/src/internal/core/providers/lazy-initialization.ts
```diff
@@ -40,6 +40,19 @@ export class LazyInitializationProviderAdapter implements EthereumProvider {
...
+
+      for (const event of recordedEvents) {
+        const listeners = this._emitter.rawListeners(event) as Listener[];
+        for (const listener of listeners) {
+          this.provider.on(event, listener);
+          this._emitter.removeListener(event, listener);
+        }
+      }
+
+      this.provider.setMaxListeners(this._emitter.getMaxListeners());
```
schaable, line 55: I know this is how it's done in the original code, but shouldn't we do it before the loop?

alcuadrado:
Great catch!

schaable:
Released in [Hardhat v2.22.11](https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.22.11).

PR #5734, opened 2024-09-09
Move the global-dir functions from hardhat to harhdat-utils

ChristopherDedominici:
PROBLEM
The `global-dir` functions could also be used inside plugins or other packages other than hardhat. 
So they should not be in an internal file isolated in hardhat.

SOLUTION
Move the functions from the internal folder from  the hardhat package to the hardhat-utils package.

The available directories paths are:
- config
- cache 
- telemetry

The default package name use to generate the path is `hardhat` but for all of them it can be specified a different package name (e.g.: if a plugin does not want to store info in the same hardhat path).


v-next/hardhat-utils/src/global-dir.ts
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Returns the configuration directory path for a given package (defaults to "hardhat").
+ * Ensures that the directory exists before returning the path.
+ *
+ * @param packageName The name of the package for which to generate paths. Defaults to "hardhat" if no package name is provided.
+ * @returns The path to the hardhat configuration directory.
+ * @throws FileSystemAccessError for any error.
+ */
+export async function getConfigDir(
+  packageName: string = HARDHAT_PACKAGE_NAME,
```
ChristopherDedominici, line 14: I added the possibility to pass a `package name` in every exported function so that user can override the default behavior for all of them

alcuadrado (approved):

PR #5738, opened 2024-09-11
fix: hardhat-chai-matchers - add check byte32 string to reverted matcher

### LINKED ISSUE #4725: [hardhat-chai-matchers] Edge case with `not.to.be.reverted` and view functions

iosh:

---

 Add when receipt is null, check if value is a byte32 string. minimal changes were in PRto resolve #4725 

I see the following comment in the code

```js

        // If the subject of the assertion is not connected to a transaction
        // (hash, receipt, etc.), then the assertion fails.
        // Since we use `false` here, this means that `.not.to.be.reverted`
        // assertions will pass instead of always throwing a validation error.
        // This allows users to do things like:
        //   `expect(c.callStatic.f()).to.not.be.reverted`
        assert(false, "Expected transaction to be reverted");
```

So i think maybe we can change the check to :

```js
      if (
        isTransactionResponse(value) ||
        (typeof value === "string" && isValidTransactionHash(value))
      ) {
        const hash = typeof value === "string" ? value : value.hash;

        // if (!isValidTransactionHash(hash)) {
        //   throw new TypeError(
        //     `Expected a valid transaction hash, but got '${hash}'`
        //   );
        // }

```
This way, we can enable other data types to pass the test, but they will be broken change.

### DIFF
--- a/packages/hardhat-chai-matchers/src/internal/reverted/reverted.ts
+++ b/packages/hardhat-chai-matchers/src/internal/reverted/reverted.ts
@@ -3,7 +3,11 @@ import type EthersT from "ethers";
 import { buildAssert } from "../../utils";
 import { REVERTED_MATCHER } from "../constants";
 import { assertIsNotNull, preventAsyncMatcherChaining } from "../utils";
-import { decodeReturnData, getReturnDataFromError } from "./utils";
+import {
+  decodeReturnData,
+  getReturnDataFromError,
+  parseBytes32String,
+} from "./utils";
 
 export function supportReverted(
   Assertion: Chai.AssertionStatic,
@@ -35,6 +39,14 @@ export function supportReverted(
         }
 
         const receipt = await getTransactionReceipt(hash);
+
+        if (receipt === null) {
+          // If the receipt is null, maybe the string is a bytes32 string
+          if (isByte32String(hash)) {
+            assert(false, "Expected transaction to be reverted");
+            return;
+          }
+        }
 
         assertIsNotNull(receipt, "receipt");
         assert(
@@ -133,4 +145,13 @@ function isTransactionReceipt(x: unknown): x is { status: number } {
 
 function isValidTransactionHash(x: string): boolean {
   return /0x[0-9a-fA-F]{64}/.test(x);
+}
+
+function isByte32String(v: string): boolean {
+  try {
+    parseBytes32String(v);
+    return true;
+  } catch {
+    return false;
+  }
 }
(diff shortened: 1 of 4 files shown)


packages/hardhat-chai-matchers/src/internal/reverted/reverted.ts
```diff
@@ -36,6 +40,14 @@ export function supportReverted(
 
         const receipt = await getTransactionReceipt(hash);
 
+        if (receipt === null) {
+          // If the receipt is null, maybe the string is a bytes32 string
+          if (isByte32String(hash)) {
+            assert(false, "Expected transaction to be reverted");
```
iosh, line 46: I think it also can throw an error:
```js
 throw new TypeError(
            `Expected a valid transaction hash, but got bytes32  '${hash}'`
       )
```

schaable (approved):
Looks good. Tested with the code example provided in the issue and it works. Thanks @iosh!

Issue #5723, opened 2024-09-05
Error: Unsupported language version '0.8.27'


majinbot:
### Describe the feature

Hello!

can you release an update to support https://github.com/ethereum/solidity/releases/tag/v0.8.27

Confucian-e:
+1

schaable:
Released in [Hardhat v2.22.11](https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.22.11).

schaable:
@majinbot that's interesting. I just compiled the sample project using the config and dependencies you mentioned without any errors. Could you share a link to your repo so I can take a closer look?

alcuadrado:
I think the error is in "@openzeppelin/contracts-upgradeable"

Juglipaff:
For me, this is mitigated by manualy providing the latest `@nomicfoundation/slang` commit in `package.json`, deleting `node_modules` with `package-lock.json`, and doing `npm i`.

```json
"overrides": {
    "@openzeppelin/upgrades-core": {
      "@nomicfoundation/slang": "git://github.com/NomicFoundation/slang.git#15c437c6c4902cd43e0027f750ba59e8f22f47f9"
    }
 }
 ```

Note that this is a temporary solution before the new `@nomicfoundation/slang` version release.

majinbot:
> For me, this is mitigated by manually providing the latest `@nomicfoundation/slang` commit in `package.json`, deleting `node_modules` with `package-lock.json`, and doing `npm i`.
> 
> ```json
> "overrides": {
>     "@openzeppelin/upgrades-core": {
>       "@nomicfoundation/slang": "git://github.com/NomicFoundation/slang.git#15c437c6c4902cd43e0027f750ba59e8f22f47f9"
>     }
>  }
> ```
> 
> Note that this is a temporary solution before the new `@nomicfoundation/slang` version release.

Mmmh, doesnt work for me with Bun, it should support npm overrides just fine 🤷

PR #5740, opened 2024-09-11
feat: add support for solc versions 0.8.25, 0.8.26, and 0.8.27

schaable:
This PR adds support for Solidity versions 0.8.25, 0.8.26, and 0.8.27, and updates the default/latest Solidity version to 0.8.27.

Closes https://github.com/NomicFoundation/hardhat/issues/5723


fvictorio (approved):

galargh:
@schaable I restored the option to run tests for all solc versions in CI (in https://github.com/NomicFoundation/hardhat/pull/5745) and kicked off a run that has the changes introduced in this PR - https://github.com/NomicFoundation/hardhat/actions/runs/10846884277. As you can see it fails with OOMs, do you think that's something we should look into more?

schaable:
> @schaable I restored the option to run tests for all solc versions in CI (in #5745) and kicked off a run that has the changes introduced in this PR - [NomicFoundation/hardhat/actions/runs/10846884277](https://github.com/NomicFoundation/hardhat/actions/runs/10846884277). As you can see it fails with OOMs, do you think that's something we should look into more? Other than that, it looks good to me, but wanted to check with you first.

To be honest, I’m not sure. It doesn’t seem directly related to adding the `solc` versions, so I’m inclined to think it's not an issue caused by that, but I could be wrong. @alcuadrado @fvictorio, wdyt?

alcuadrado:
@fvictorio, what's the status of these tests now that more things are ported to Rust? Should we still run them here?

fvictorio:
@alcuadrado yes. We are working right now to complete the migration. After that (and after helping adapts solc's CI to run on EDR) we'll be able to remove the stack traces tests from the Hardhat repo.

alcuadrado (approved):
Appproving despite the OOM that @galargh discovered, as that's a CI workflow that we don't usually run, and will be moved to EDR soon.

PR #5741, opened 2024-09-12
add mock artifacts plugins

kanej:
While the build system is under development, this PR adds a temporary artifacts plugin that adds a fake artifacts manager to the `hre` under `hre.artifacts`.

This `hre.artifacts` will be replaced, but for the moment you can code against it by leveraging `createMockHardhatRuntimeEnvironment` in your tests. This will apply a mock artifacts plugin that you can use in your tests like so:

```ts
describe("createMockHardhatRuntimeEnvironment", () => {
  it("should allow plugins that leverage the artifact hre object", async () => {
    // arrange
    const exampleArtifact: Artifact = {
      _format: "hh-sol-artifact-1",
      contractName: "MyContract",
      sourceName: "source.sol",
      abi: [],
      bytecode: "0x",
      linkReferences: {},
      deployedBytecode: "0x",
      deployedLinkReferences: {},
    };

    const myPlugin: HardhatPlugin = {
      id: "my-plugin",
      tasks: [
        task("hello-artifact-using-world", "Tests artifact loading")
          .setAction(async ({}, hre: HardhatRuntimeEnvironment) => {
            return hre.artifacts.readArtifact("MyContract");
          })
          .build(),
      ],
    };

    const mockHre = await createMockHardhatRuntimeEnvironment({
      plugins: [myPlugin],
    });

    await mockHre.artifacts.saveArtifact(exampleArtifact);

    // act
    const helloArtifactUsingWorld = mockHre.tasks.getTask(
      "hello-artifact-using-world",
    );

    const result = await helloArtifactUsingWorld.run({});

    // Assert
    assert.equal(result, exampleArtifact);
  });
});
```

### DIFF
--- a/v-next/hardhat-errors/src/descriptors.ts
+++ b/v-next/hardhat-errors/src/descriptors.ts
@@ -209,6 +209,15 @@ Please add the property "type" with the value "module" in your package.json to e
 Please [report it](https://github.com/nomiclabs/hardhat/issues/new) to help us improve Hardhat.`,
       shouldBeReported: true,
     },
+    NOT_IMPLEMENTED_ERROR: {
+      number: 101,
+      messageTemplate: "Not implemented: {message}",
+      websiteTitle: "Not implemented",
+      websiteDescription: `A code path that has not been implemented was unexpectedly triggered.
+
+Please [report it](https://github.com/nomiclabs/hardhat/issues/new) to help us improve Hardhat.`,
+      shouldBeReported: true,
+    },
   },
   PLUGINS: {
     PLUGIN_NOT_INSTALLED: {
--- a/v-next/hardhat/src/internal/builtin-plugins/artifacts/hookHandlers/hre.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/artifacts/hookHandlers/hre.ts
@@ -1,0 +1,13 @@
+import type { HardhatRuntimeEnvironmentHooks } from "../../../../types/hooks.js";
+
+import { ArtifactsManagerImplementation } from "../../../artifacts/artifacts-manager.js";
+
+export default async (): Promise<Partial<HardhatRuntimeEnvironmentHooks>> => {
+  const handlers: Partial<HardhatRuntimeEnvironmentHooks> = {
+    created: async (_context, hre): Promise<void> => {
+      hre.artifacts = new ArtifactsManagerImplementation();
+    },
+  };
+
+  return handlers;
+};
--- a/v-next/hardhat/src/internal/core/hre.ts
+++ b/v-next/hardhat/src/internal/core/hre.ts
@@ -1,4 +1,5 @@
 import type { UnsafeHardhatRuntimeEnvironmentOptions } from "./types.js";
+import type { ArtifactsManager } from "../../types/artifacts.js";
 import type {
   HardhatUserConfig,
   HardhatConfig,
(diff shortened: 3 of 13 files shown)


v-next/hardhat/src/internal/core/hre.ts
```diff
@@ -134,7 +134,14 @@ export class HardhatRuntimeEnvironmentImplementation
       globalOptionDefinitions,
     );
 
-    await hooks.runSequentialHandlers("hre", "created", [hre]);
```
kanej, line 137: I switched this as the mock artifact plugin (which depends on the artifact plugin) was being run before the artifacts plugin. This is because `runSequentialHandlers` runs in reverse priority order.

I am unsure if this is the right move. But plugin authors who enhance the `hre` are likely to want to execute after other `hre` enhancements they are depending on.

kanej: This change has been dropped. We fixed the underlying bug with the `runSequentialHandlers` order.

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -206,6 +206,15 @@ Please add the property "type" with the value "module" in your package.json to e
       websiteTitle: "Invariant violation",
       websiteDescription: `An internal invariant was violated. This is probably caused by a programming error in Hardhat or in one of the used plugins.
 
+Please [report it](https://github.com/nomiclabs/hardhat/issues/new) to help us improve Hardhat.`,
+      shouldBeReported: true,
+    },
+    NOT_IMPLEMENTED_ERROR: {
```
kanej, line 212: I initially piggy backed on the assertion error. But I suspect it is better for us to be explicit so we can track unimplemented sections of the code base.

v-next/hardhat/src/types/artifacts.ts
```diff
@@ -0,0 +1,237 @@
...
+ *
+ * This file has just enough information to deploy the contract and interact
+ * with an already deployed instance of it.
+ *
+ * For debugging information and other extra information, you should look for
+ * its companion DebugFile, which should be stored right next to it.
+ *
+ * Note that DebugFiles are only generated for Solidity contracts.
+ */
+export interface Artifact<AbiT extends Abi = Abi> {
```
kanej, line 112: This is the form taken over from Ignition.

v-next/hardhat/src/internal/builtin-plugins/artifacts/hookHandlers/hre.ts
```diff
@@ -0,0 +1,13 @@
+import type { HardhatRuntimeEnvironmentHooks } from "../../../../types/hooks.js";
+
+import { ArtifactsManagerImplementation } from "../../../artifacts/artifacts-manager.js";
```
alcuadrado, line 3: I'll add this to the guidelines tomorrow, but this is the only  hook handler category where we want to load things lazily, as it's always executed.

alcuadrado (approved):
I left a minor comment, but looks good to me

v-next/hardhat/src/internal/builtin-plugins/artifacts/hookHandlers/hre.ts
kanej: I switched the `ArtifactsManagerImplementation` to be lazily based on the pattern in the network builtin plugin.

PR #5743, opened 2024-09-12
Introduce the Zod `conditionalUnionType` helper

alcuadrado:
This PR adds a new helper to our zod utils.

I originally planned to remove the old `unionType`, but it's still valid.

I also introduced another helper that's meant to be used with `conditionalUnionType`: `incompatibleFieldType`. See its docs.

### DIFF
--- a/v-next/hardhat-mocha-test-runner/src/hookHandlers/config.ts
+++ b/v-next/hardhat-mocha-test-runner/src/hookHandlers/config.ts
@@ -1,7 +1,9 @@
 import type { ConfigHooks } from "@ignored/hardhat-vnext/types/hooks";
 
+import { isObject } from "@ignored/hardhat-vnext-utils/lang";
 import { resolveFromRoot } from "@ignored/hardhat-vnext-utils/path";
 import {
+  conditionalUnionType,
   unionType,
   validateUserConfigZodType,
 } from "@ignored/hardhat-vnext-zod-utils";
@@ -72,10 +74,17 @@ const mochaConfigType = z.object({
 
 const userConfigType = z.object({
   mocha: z.optional(mochaConfigType),
-  test: unionType(
-    [z.object({ mocha: z.string().optional() }), z.string()],
-    "Expected a string or an object with an optional 'mocha' property",
-  ).optional(),
+  paths: z
+    .object({
+      test: conditionalUnionType(
+        [
+          [(data) => typeof data === "string", z.string()],
+          [isObject, z.object({ mocha: z.string().optional() })],
+        ],
+        "Expected a string or an object with an optional 'mocha' property",
+      ).optional(),
+    })
+    .optional(),
 });
 
 export default async (): Promise<Partial<ConfigHooks>> => {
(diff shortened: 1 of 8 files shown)


v-next/hardhat-mocha-test-runner/src/hookHandlers/config.ts
```diff
@@ -72,10 +74,17 @@ const mochaConfigType = z.object({
 
 const userConfigType = z.object({
   mocha: z.optional(mochaConfigType),
-  test: unionType(
```
schaable, line 75: was the original config path a bug? I noticed you changed it from `config.test` to `config.paths.test`, which makes sense, but I’m wondering why it was like that initially. Did we miss updating it in a previous PR?

v-next/hardhat-zod-utils/src/index.ts
```diff
@@ -42,16 +145,22 @@ export const configurationVariableType = z.object({
 /**
  * A Zod type to validate Hardhat's SensitiveString values.
  */
-export const sensitiveStringType = unionType(
-  [z.string(), configurationVariableType],
+export const sensitiveStringType = conditionalUnionType(
+  [
+    [(data) => typeof data === "string", z.string()],
+    [(data) => typeof data === "object", configurationVariableType],
```
schaable, line 151: we can use `isObject` here

v-next/hardhat-zod-utils/test/index.ts
```diff
@@ -32,3 +41,146 @@ describe("unionType", () => {
...
+    it("should return the provided error message if no condition matches", async () => {
+      const shouldUseString = conditionalUnionType(
+        [
+          [(data) => typeof data === "string", z.string()],
+          [(data) => typeof data === "string", z.string().url()],
+        ],
+        "No match",
+      );
+
+      // Both conditions match, but we only use the first one
```
schaable, line 79: this comment doesn't apply

v-next/hardhat-zod-utils/test/index.ts
```diff
@@ -32,3 +41,146 @@ describe("unionType", () => {
...
+        "No match",
+      );
+
+      assertParseResult(shouldUseString.safeParse(123), "No match", []);
+    });
+
+    it("Should have the path to the nested error if a condition matches", () => {
+      const shouldUseString = conditionalUnionType(
+        [
+          [(data) => typeof data === "object", z.object({ foo: z.string() })],
```
schaable, line 100: not that it matters too much here, but I'd still use isObject in tests, as it could be viewed or copied by another dev later on.

schaable (approved):
Great additions 💯! I’ve left a few minor comments.

v-next/hardhat-mocha-test-runner/src/hookHandlers/config.ts
alcuadrado: Oh, it was a bug, and the fix went unnoticed in this PR. I was meaning to send a separate one.

v-next/hardhat-zod-utils/src/index.ts
alcuadrado: good point

v-next/hardhat-zod-utils/test/index.ts
alcuadrado: Good catch

v-next/hardhat-zod-utils/test/index.ts
alcuadrado: yep

PR #5745, opened 2024-09-13
ci: add test-all-solc-versions input to the core ci

galargh:

---

This PR restores the ability to run the tests for all solc versions in CI on demand. It doesn't change the default of running the tests only for the compiler versions marked as latest. 

Here's an example run - https://github.com/NomicFoundation/hardhat/actions/runs/10846884277 

Note that it currently fails with OOM errors :( These are the standard machines sizes - https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories Do you think we should act on it?

### DIFF
--- a/.github/workflows/hardhat-core-ci.yml
+++ b/.github/workflows/hardhat-core-ci.yml
@@ -22,7 +22,7 @@ on:
       test-all-solc-versions:
         description: Test all solc versions
         required: false
-        default: 'false'
+        default: "false"
 
 defaults:
   run:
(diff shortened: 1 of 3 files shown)


schaable (approved):
I'm going to give this an approval because it looks good to my eyes, but you probably know a lot more about workflows than me, so I'd ask for another reviewer if you're not totally sure about the changes 😄

PR #5751, opened 2024-09-16
fix: rework `runSequentialHandlers` plugin ordering

kanej:
The plugin reverse ordering of handlers makes sense for the chained
running of hooks but is wrong for the sequential running, where
dependent plugins should run before the dependee.

We make several moves:

- remove the `getHandlers` method from the API as it is only used
  internally to the hook manager
- create equivalents to `getHandlers` with explicit logic for reversing
  the plugin order for chained running, and keeping the dependency order
  for sequential.

## Reviewers Suggestion

This PR should be reviewed a commit at a time to avoid the jitter from the test restructuring.

### DIFF
--- a/v-next/hardhat/src/internal/core/hook-manager.ts
+++ b/v-next/hardhat/src/internal/core/hook-manager.ts
@@ -57,25 +57,6 @@ export class HookManagerImplementation implements HookManager {
     this.#context = context;
   }
 
-  public async getHandlers<
-    HookCategoryNameT extends keyof HardhatHooks,
-    HookNameT extends keyof HardhatHooks[HookCategoryNameT],
-  >(
-    hookCategoryName: HookCategoryNameT,
-    hookName: HookNameT,
-  ): Promise<Array<HardhatHooks[HookCategoryNameT][HookNameT]>> {
-    const pluginHooks = await this.#getPluginHooks(hookCategoryName, hookName);
-
-    const dynamicHooks = await this.#getDynamicHooks(
-      hookCategoryName,
-      hookName,
-    );
-
-    const r = [...dynamicHooks, ...pluginHooks];
-
-    return r;
-  }
-
   public registerHandlers<HookCategoryNameT extends keyof HardhatHooks>(
     hookCategoryName: HookCategoryNameT,
     hookHandlerCategory: Partial<HardhatHooks[HookCategoryNameT]>,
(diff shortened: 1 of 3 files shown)


v-next/hardhat/src/types/hooks.ts
```diff
@@ -321,8 +308,11 @@ export interface HookManager {
   ): Promise<Awaited<Return<HookT>>>;
 
   /**
-   * Runs all the handlers for a hook in the same order that `getHandlers`
-   * returns them.
+   * Runs all the handlers for a hook in the following priority order:
+   *  - Dynamically registered handlers come first, in the reverse order they
+   *   were registered.
```
kanej, line 313: There is an argument that the registered handlers should also not be reversed. But that was a step further than was needed to fix the underlying bug.

Are we happy that for sequential runs the ordering of registered handlers is done in the same was as for the handler chain?

v-next/hardhat/src/internal/core/hook-manager.ts
```diff
@@ -205,6 +196,42 @@ export class HookManagerImplementation implements HookManager {
...
+    const reversedPluginHooks = (
+      await this.#getPluginHooks(hookCategoryName, hookName)
+    ).reverse();
+
+    const dynamicHooks = await this.#getDynamicHooks(
+      hookCategoryName,
+      hookName,
+    );
+
+    return [...dynamicHooks, ...reversedPluginHooks];
```
alcuadrado, line 232:
```suggestion
    const reverseDynamicHooks = await this.#getDynamicHooks(
      hookCategoryName,
      hookName,
    ).reversed();

    return [...reversedPluginHooks, ...reversedDynamicHooks];
```

Shouldn't this be like this?

v-next/hardhat/src/types/hooks.ts
```diff
@@ -252,23 +252,6 @@ export interface HardhatRuntimeEnvironmentHooks {
...
-  /**
-   * Returns an array with the existing handlers for a hook, in priority order.
-   *
-   * The priority is defined like this:
-   *  - Dynamically registered handlers come first, in the reverse order they
-   *   were registered.
-   *  - Plugin handlers come last, in the same reverse of the resolved plugins
-   *  list, as seen in `HardhatConfig#plugins`.
-   */
-  getHandlers<
```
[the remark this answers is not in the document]
alcuadrado: makes sense, it's confusing otherwise.

v-next/hardhat/src/types/hooks.ts
alcuadrado: As susggested in another comment, I think this should be the reverse of the other one.

v-next/hardhat/src/internal/core/hook-manager.ts
kanej: I have applied the complete reverse approach:

```ts
return this.#getHandlersInChainedRunningOrder(hookCategoryName, hookName).reverse();
```

The tests and doc comments have been updated to reflect the complete reordering.

See [`3dd37fe` (#5751)](https://github.com/NomicFoundation/hardhat/pull/5751/commits/3dd37fe886712111f7ba183f94e6c9d97e73ccbd)

alcuadrado (approved):

PR #5753, opened 2024-09-16
V3 hardhat-network-helpers

ChristopherDedominici:
## Issue to track TODOs as soon as V3 node is ready
[Link to issue](https://github.com/NomicFoundation/hardhat/issues/5757)

## Short overview on how to use the package

```javascript
const { networkHelpers } = await hre.network.connect();

// Network helpers methods exposed via `networkHelpers`
await networkHelpers.mine();

// Time methods exposed via `time`
await networkHelpers.time.increase(1);

// Duration methods exposed via `duration`
networkHelpers.time.duration.days(1);
```

## Questions: input needed

- the old `network-helpers` package has a type called `NumberLike` - [see it here](https://github.com/NomicFoundation/hardhat/blob/d725c09783da3750beb5d80c81774490a1ea58bf/packages/hardhat-network-helpers/src/types.ts#L10C1-L10C25) - that defines the types `EthersBigNumberLike` and `BNLike` that are used to support the types exported by the packages `ethers` and `ethereumjs-util`. Do we still want to support those types or just support `number, bigint, string` (as we do in `hh-utils` in the method `toBigInt`)? `NumberLike` is used in the code in situations like this:

```javascript
  public async mineUpTo(blockNumber: NumberLike): Promise<void> {
    ...code...
  }
```

I think we could remove `NumberLike` and just have a few tests that ensure that the code is able to manage the types from `EthersBigNumberLike` and `BNLike`. In doing so, we will have the dependencies to these external packages only as devDependencies. Note: probably a manual casting will be needed.

This decision will also impact [this conversion file](https://github.com/NomicFoundation/hardhat/pull/5753/files#diff-30112b7b1e81968714c60ad4b852c69e62e8060d66db41af0d8ed5d80d8a9d94) because then the methods exposed by `hh-utils` could be used (e.g.: `toBigInt`). Currently it's not possible to use it because there is a mismatch between types, `toBigInt` does not support `EthersBigNumberLike` and `BNLike`.

- In the check for valid addresses the old code uses `isValidChecksumAddress` from the `ethereumjs-util` package. Should we create our own function to check the checksum so this dependency can be removed? Note that the `ethereumjs-util` package is also the one mentioned above.

- In old tests a few network errors were tested even though they were coming from the provider (which we should assume as a black box). I kept them as precaution but do we want to remove them? Also because they do not throw HardhatErrors, so `assert.rejects` must be used which is against the code style guidelines. [See example here](<https://github.com/NomicFoundation/hardhat/pull/5753/files#diff-12f04d034c08be7676e1bd5107d0f790722645bca4f407818833c99c4e7db917:~:text=await%20assert.rejects(networkHelpers.dropTransaction(txHash))%3B>)

- the methods [advanceBlock](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-network-helpers/src/helpers/time/advanceBlock.ts) and [advanceBlockTo](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-network-helpers/src/helpers/time/advanceBlockTo.ts) were marked as deprecated in the old code, so I did not ported them.

- the old code was also exposing the function `loadFixture`. I did not included it in the `NetworkHelpers` class and I kept it separated because it should only be used in tests. Do we want to keep it separated or do we want to include it in the `NetworkHelpers` class?

- a few tests [like here at bottom of the file](https://github.com/NomicFoundation/hardhat/pull/5753/files#diff-a61172afdf0b13979daec282b02a9689319771816c345649b8f311eed6356b19) cannot currently be executed because in V3 there is not yet the possibility to modify a running node. These tests are using the property `allow-blocks-same-timestamp`. Will be this property added in V3? Or these tests can be removed? 

- as described in the design doc, all the methods are declared inside the network helpers class. The file is quite long but also easy to navigate. So I think that if we want we can split the methods in different files and then wrap them in the NetworkHelper class, but I think that at this point it's more a choice of style preference than actual functionalities.

- The way to use the network-helpers is changed compared to V2, should we update the docs now or later?

- in the design dock is mentioned to set all the tests to `skip`, but I think that for simplicity of use we should use the approach mentioned in the `tests` section (see next).

- I added the plugin in the example project but currently there is no `test` command configured so it cannot be used. 

## Tests

The tests cannot be run on the CI because they need a running Hardhat node and currently V2 does not have one. But they can re run manually on the local machine. To do so, you need to:

1. Start a node in Hardhat V2: `npx hardhat node`
2. Run the tests: `pnpm test:tmp`

The command `pnpm test:tmp` has been added as temporary solution because the tests MUST run in a synchronous way since they all share the same Hardhat V2 node. So if they were executed in parallel (this is the default nodeJS test runner behavior), they would interfere with each other.

A TMP file called `run-tests.ts` has been added to the root of the project to run the tests in a synch way, for the reason just mentioned.

The tests are currently placed in the `test-tmp` folder and not the `test` folder. This is needed to keep the script `test` in the `package.json` file unchanged (otherwise it will fail in the CI with `Check that the npm scripts are consistent`). This also allows to not have to add `skips` in all the tests. So the tests can just be locally executed with the 2 steps mentioned above without touching any code.

The command `pnpm test:tmp` and the script `run-tests.ts` will be removed as soon as the V3 node is ready. Also, the way on how the network is initialized will be changed with the V3 node implementation [see here](https://github.com/NomicFoundation/hardhat/pull/5753/files#diff-e73f1c51852cd28dc50d02b9bd31318253d0a9c987850f98155fb6f774c5b193:~:text=export%20async%20function%20initializeNetwork()%3A%20Promise%3C%7B)

If you want to run the tests using an INFURA_URL, you can set the env variable `INFURA_URL` to the desired URL. If the env variable is not set, the test will be skipped.

Currently the `pnpm test` is not running any tests, it is an empty file for the reasons mentioned above.

## TODOs once V3 node is ready
- delete the `run-tests.ts` file
- delete the `pnpm test:tmp` command
- update the `initializeNetwork` helper function at the path `./test/helpers/helpers.ts`
- move the tests from the `test-tmp` folder to the `test` fodler and delete the `test-tmp` folder
- update the tests that are using the `allow-blocks-same-timestamp` property (if we still support the property, see previous section about the questions)


v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
```diff
@@ -0,0 +1,452 @@
...
+
+export class NetworkHelpers {
+  readonly #provider: EthereumProvider;
+
+  public time: Time;
+
+  constructor(provider: EthereumProvider) {
+    this.#provider = provider;
+
+    this.time = new Time(this, provider);
```
ChristopherDedominici, line 29: Needed because the `time` class uses methods from the network helpers class. It also goes in the opposite direction: the network helpers class uses methods from the time class.

v-next/hardhat-network-helpers/test/index.ts
```diff
@@ -0,0 +1 @@
+// TODO: revisit as soon as the V3 node is ready: move tests from TMP folder to this folder
```
ChristopherDedominici, line 1: See the PR description for understanding why this is needed

alcuadrado:
I'm starting to review this PR, but I'll answer here first

> ## Questions: input needed
> * the old `network-helpers` package has a type called `NumberLike` - [see it here](https://github.com/NomicFoundation/hardhat/blob/d725c09783da3750beb5d80c81774490a1ea58bf/packages/hardhat-network-helpers/src/types.ts#L10C1-L10C25) - that defines the types `EthersBigNumberLike` and `BNLike` that are used to support the types exported by the packages `ethers` and `ethereumjs-util`. Do we still want to support those types or just support `number, bigint, string` (as we do in `hh-utils` in the method `toBigInt`)? `NumberLike` is used in the code in situations like this:
> 
> ```js
>   public async mineUpTo(blockNumber: NumberLike): Promise<void> {
>     ...code...
>   }
> ```
> 
> I think we could remove `NumberLike` and just have a few tests that ensure that the code is able to manage the types from `EthersBigNumberLike` and `BNLike`. In doing so, we will have the dependencies to these external packages only as devDependencies. 
I think we can remove it altogether, and drop support for `BNLike` entirely. This will put this package in alignment to hh-utils.

> * In the check for valid addresses the old code uses `isValidChecksumAddress` from the `ethereumjs-util` package. Should we create our own function to check the checksum so this dependency can be removed? Note that the `ethereumjs-util` package is also the one mentioned above.

Let's create a `hh-utils/address` and copy paste that `isValidChecksumAddress` with a comment saying that it comes from ethereumjs-util. Don't forget to also copy its tests. This part of the codebase mostly: https://github.com/ethereumjs/ethereumjs-monorepo/blob/47f388bfeec553519d11259fee7e7161a77b29b2/packages/util/src/account.ts#L440-L478

> * In old tests a few network errors were tested even though they were coming from the provider (which we should assume as a black box). I kept them as precaution but do we want to remove them? Also because they do not throw HardhatErrors, so `assert.rejects` must be used which is against the code style guidelines. [See example here](https://github.com/NomicFoundation/hardhat/pull/5753/files#diff-12f04d034c08be7676e1bd5107d0f790722645bca4f407818833c99c4e7db917:~:text=await%20assert.rejects(networkHelpers.dropTransaction(txHash))%3B)

using `asser.rejects` for this kind of error is ok. We should probably add helpers for them later.

> * the methods [advanceBlock](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-network-helpers/src/helpers/time/advanceBlock.ts) and [advanceBlockTo](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-network-helpers/src/helpers/time/advanceBlockTo.ts) were marked as deprecated in the old code, so I did not ported them.

Makes sense.

> * the old code was also exposing the function `loadFixture`. I did not included it in the `NetworkHelpers` class and I kept it separated because it should only be used in tests. Do we want to keep it separated or do we want to include it in the `NetworkHelpers` class?

I'm not sure if I understand your point here. `loadFixture`, like the rest of the helpers, is used in our users' own tests.

> * a few tests [like here at bottom of the file](https://github.com/NomicFoundation/hardhat/pull/5753/files#diff-a61172afdf0b13979daec282b02a9689319771816c345649b8f311eed6356b19) cannot currently be executed because in V3 there is not yet the possibility to modify a running node. These tests are using the property `allow-blocks-same-timestamp`. Will be this property added in V3? Or these tests can be removed?

It will.

> * as described in the design doc, all the methods are declared inside the network helpers class. The file is quite long but also easy to navigate. So I think that if we want we can split the methods in different files and then wrap them in the NetworkHelper class, but I think that at this point it's more a choice of style preference than actual functionalities.

If it doesn't change the API that we expose to the users, and can be done quickly, it could be good.

> * The way to use the network-helpers is changed compared to V2, should we update the docs now or later?

We'll get to the docs later.

> * I added the plugin in the example project but currently there is no `test` command configured so it cannot be used.

It should still be usable from a script, right?

> ## Tests
> The tests cannot be run on the CI because they need a running Hardhat node and currently V2 does not have one. But they can re run manually on the local machine. To do so, you need to:
> 
> 1. Start a node in Hardhat V2: `npx hardhat node`
> 2. Run the tests: `pnpm test:tmp`
> 
> The command `pnpm test:tmp` has been added as temporary solution because the tests MUST run in a synchronous way since they all share the same Hardhat V2 node. So if they were executed in parallel (this is the default nodeJS test runner behavior), they would interfere with each other.
> 
> A TMP file called `run-tests.ts` has been added to the root of the project to run the tests in a synch way, for the reason just mentioned.
> 
> The tests are currently placed in the `test-tmp` folder and not the `test` folder. This is needed to keep the script `test` in the `package.json` file unchanged (otherwise it will fail in the CI with `Check that the npm scripts are consistent`). This also allows to not have to add `skips` in all the tests. So the tests can just be locally executed with the 2 steps mentioned above without touching any code.
> 
> The command `pnpm test:tmp` and the script `run-tests.ts` will be removed as soon as the V3 node is ready. Also, the way on how the network is initialized will be changed with the V3 node implementation [see here](https://github.com/NomicFoundation/hardhat/pull/5753/files#diff-e73f1c51852cd28dc50d02b9bd31318253d0a9c987850f98155fb6f774c5b193:~:text=export%20async%20function%20initializeNetwork()%3A%20Promise%3C%7B)
> 
> If you want to run the tests using an INFURA_URL, you can set the env variable `INFURA_URL` to the desired URL. If the env variable is not set, the test will be skipped.
> 
> Currently the `pnpm test` is not running any tests, it is an empty file for the reasons mentioned above.
> 
> ## TODOs once V3 node is ready
> * delete the `run-tests.ts` file
> * delete the `pnpm test:tmp` command
> * update the `initializeNetwork` helper function at the path `./test/helpers/helpers.ts`
> * move the tests from the `test-tmp` folder to the `test` fodler and delete the `test-tmp` folder
> * update the tests that are using the `allow-blocks-same-timestamp` property (if we still support the property, see previous section about the questions)

I think this approach, as a temporarily way to unblock yourself, makes sense. The sync execution got me confused at first, but it's a good quick way to achieve that. Another one would be to initialize the network with random ports, so they can also run in parallel. Just letting you know, i don't think this needs to be changed.

alcuadrado:
> Let's create a `hh-utils/address`

Self-correction: `hh-utils/eth` exists, so let's add it there.

v-next/hardhat-network-helpers/src/types.ts
```diff
@@ -0,0 +1,34 @@
...
+  toNumber(): number;
+  toString(base?: number): string;
+}
+
+export type NumberLike =
+  | number
+  | bigint
+  | string
+  | EthersBigNumberLike
+  | BNLike;
```
alcuadrado, line 15: These are the ones that I think we can remove.

alcuadrado (commented):
I think this looks good. Let's resolve the answer/commetns in the main discussion and then get it merged

ChristopherDedominici:
The method to check an address checksum has been added [in this commit](https://github.com/NomicFoundation/hardhat/pull/5753/commits/d34d20e4c464386009c5f35c0bd1d0ec8e8605b3). Main points:
- I added an helper method in the `bytes.ts` called `utf8StringToBytes`
- I ported the `ethereumjs` function + tests. I adopted the function to use the `hh-utils` methods and I removed the `eip1191ChainId` parameter because currently we do not use it. If needed, it can be added later

v-next/hardhat-network-helpers/src/types.ts
ChristopherDedominici: The 'ethers' and 'ethereumjs-util' dependencies have been removed in [this commit](https://github.com/NomicFoundation/hardhat/pull/5753/commits/5375f74d50984adfd393b04779f1a9550e6348e1).

Now NumberLike only supports number | bigint | string. Tests have been modified as consequence

ChristopherDedominici:
Regarding `loadFixture`: what I meant is that `loadFixture` was in a separate folder in the old code. Should we keep this separation and not include `loadFixture` in the NetworkHelper class but allow it to be imported like this:

```js
import { loadFixture } from "@ignored/hardhat-vnext-network-helpers/load-fixture";
```

ChristopherDedominici:
The logic for the classes `NetworkHelpers` and `Time` have been moved into separate files in [this commit](https://github.com/NomicFoundation/hardhat/pull/5753/commits/4bdd7562b8d811a75428d4442410ebd940ae1f41)

ChristopherDedominici:
The logic for the classes `NetworkHelpers` and `Time` have been moved into separate files in [this commit](https://github.com/NomicFoundation/hardhat/pull/5753/commits/43f1e434ed0098eebe72aefcdcb888003d2ddb76)

ChristopherDedominici:
I added an issue with the TODO list for the tasks that must be done as soon as the V3 node is ready

v-next/hardhat-network-helpers/src/types.ts
alcuadrado: great

v-next/hardhat-utils/src/eth.ts
```diff
@@ -13,6 +17,18 @@ export function isAddress(value: unknown): boolean {
   return typeof value === "string" && /^0x[0-9a-f]{40}$/i.test(value);
 }
 
+/**
+ * Checks if a value is an Ethereum address and if the checksum is valid.
+ *
+ * @param value The value to check.
+ * @returns True if the value is an Ethereum address with a valid checksum, false otherwise.
+ */
+export async function isValidChecksumAddress(value: unknown): Promise<boolean> {
```
alcuadrado, line 26: This is nitpicky, but this could be sync. I'd also be ok with this remaining async.

v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
```diff
@@ -0,0 +1,281 @@
...
+   *
+   * @example
+   * const { networkHelpers } = await hre.network.connect();
+   * const snapshot = await networkHelpers.takeSnapshot();
+   * await snapshot.restore(); // Restores the blockchain state
+   */
+  public takeSnapshot(): Promise<SnapshotRestorer> {
+    return takeSnapshot(this.#provider);
+  }
+}
```
alcuadrado, line 281: Let's add `loadFixture` like this here too, instead of exporting it on a separate file. The reason being that now we'll have multiple providers.

alcuadrado (approved):
I left a minor comment, but this looks good. Feel free to merge it after addressing that :)

PR #5759, opened 2024-09-18
refactor: Streamline `console.sol` generation

Xanewok:

---

I'm working on porting console.log from Hardhat (https://github.com/NomicFoundation/edr/issues/641). To better understand which exactly `console.log` functions we support and why, I refactored the build script for clarity in the process.

I separated helper logic (e.g. permutation generator) and upfront declare which functions we support:
- single parameter `logUint, ..., logBytes32` and
- permutations of up to 4 of (`uint256`, `string`, `bool`, `address`), including empty `log()`

Since the calculation logic was mixed with rendering (and had to take into account duplicate signatures, for example), I disentangled it and streamlined the (data) flow. Hopefully this will be useful to anyone trying to understand this or modify in the future.

While making the changes, I also changed the function selectors in the logger.ts file to be a hexadecimal integer rather than a decimal one. This is the format everyone expects when working with it so there's no use in using decimals if a hexadecimal integer is natively supported as an object key.

In the future, we can remove the faulty `(u)int` ABI encoding; I tried to highlight and and document the backwards-compatibility here.


packages/hardhat-core/scripts/console-library-generator.ts
```diff
@@ -1,44 +1,131 @@
...
+          params: [param],
+        };
+      })
+    )
+    // Also generate the function definitions for `log` for permutations of
+    // up to 4 parameters using the `types` (uint256, string, bool, address).
+    .concat(
+      [...Array(4)].flatMap((_, paramCount) => {
+        return Array.from(
+          genPermutations(TYPES.length, paramCount + 1),
```
Xanewok, line 91:
```suggestion
      [...Array(5)].flatMap((_, paramCount) => {
        return Array.from(
          genPermutations(TYPES.length, paramCount),
```

As a side-note, this base case can handle generating the empty `log()` just fine but shifted its order in the console.sol file. This is probably meaningless but I wanted to avoid making any changes; let me know if you'd like me to not explicitly handle `log()` and handle this with this code, instead.

fvictorio: That's a good point, but yes, let's leave it as is.

fvictorio (approved):

PR #5762, opened 2024-09-18
Network Manager M1 remaining tasks

schaable:
This PR addresses most of the tasks outlined in [this document](https://www.notion.so/nomicfoundation/Network-Manager-M1-missing-work-2d207a7d96944978a9cc13c4839a6a0a?pvs=4), except for the "Types" section, which will be addressed in a separate PR. I suggest reviewing one commit at a time for an easier review process.

Summary:
- Added tests for:
  - All config hooks (`extendUserConfig`, `validateUserConfig`, `resolveUserConfig`).
  - `HTTPProvider`: `jsonRpcWrapper` and `close` functions.
  - `NetworkManager`: `connect` function and all network hooks (`newConnection`, `closeConnection`, `onRequest`).
  - `NetworkConnection`: class initialization and `close` function (through `NetworkManager`)
- Fixed bugs found during testing.
- Refactored code for better readability and to facilitate testing.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
@@ -5,6 +5,7 @@ import type {
 import type { HardhatUserConfigValidationError } from "../../../types/hooks.js";
 
 import {
+  conditionalUnionType,
   sensitiveUrlType,
   unionType,
   validateUserConfigZodType,
@@ -16,24 +17,23 @@ const chainTypeSchema = unionType(
   "Expected 'l1', 'optimism', or 'unknown'",
 );
 
+const userGasSchema = conditionalUnionType(
+  [
+    [(data) => typeof data === "string", z.literal("auto")],
+    [(data) => typeof data === "number", z.number().int().positive().safe()],
+    [(data) => typeof data === "bigint", z.bigint().positive()],
+  ],
+  "Expected 'auto', a safe int, or bigint",
+);
+
 const httpNetworkUserConfigSchema = z.object({
   type: z.literal("http"),
   chainId: z.optional(z.number().int()),
   chainType: z.optional(chainTypeSchema),
   from: z.optional(z.string()),
-  gas: z.optional(
-    unionType(
-      [z.literal("auto"), z.number().int().safe(), z.bigint()],
-      "Expected 'auto', a safe int, or bigint",
-    ),
-  ),
+  gas: z.optional(userGasSchema),
   gasMultiplier: z.optional(z.number()),
-  gasPrice: z.optional(
-    unionType(
-      [z.literal("auto"), z.number().int().safe(), z.bigint()],
-      "Expected 'auto', a safe int, or bigint",
-    ),
-  ),
+  gasPrice: z.optional(userGasSchema),
 
   // HTTP network specific
   url: sensitiveUrlType,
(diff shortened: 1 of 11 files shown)


v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -0,0 +1,669 @@
...
+
+        /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+        -- testing invalid network type for js users */
+        const validationErrors = await validateUserConfig(config as any);
+
+        assert.ok(
+          validationErrors.length > 0,
+          "validation errors should be present",
+        );
+        // TODO: the error message should be "Expected a URL or a Configuration Variable"
```
schaable, line 471: The interesting thing is that this validation uses `conditionalUnionType` behind the scenes, so it might be bugged as well.

alcuadrado: I think this is working as intended concerning the `conditionalUnionType`. Its main idea is to return errors that are as specific as possible.

`unionType` does the contrary. It always returns a less specific error message.

Maybe for this particular case we should use `unionType`.

v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
```diff
@@ -23,15 +23,15 @@ const httpNetworkUserConfigSchema = z.object({
   from: z.optional(z.string()),
   gas: z.optional(
     unionType(
-      [z.literal("auto"), z.number().int(), z.bigint()],
-      "Expected 'auto', number, or bigint",
+      [z.literal("auto"), z.number().int().safe(), z.bigint()],
```
alcuadrado, line 26:
```suggestion
      [z.literal("auto"), z.number().int().positive().safe(), z.bigint()],
```

I think we are missing the `.positive()`.

We can also create a `gasSchema` to avoid duplicating it.

Finally, I think this would be better with a `conditionalUnionType`, as very few people understand what a safe number is. A `conditionalUnionType` will give more specific errors:

```
> z.number().int().positive().safe().safeParse(-1).error.errors[0].message
'Number must be greater than 0'
> z.number().int().positive().safe().safeParse(1.2).error.errors[0].message
'Expected integer, received float'
> z.number().int().positive().safe().safeParse(1e18).error.errors[0].message
'Number must be less than or equal to 9007199254740991'
```

alcuadrado (approved):
I left a few comments, but LGTM

v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -0,0 +1,669 @@
...
+        /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+        -- testing invalid network type for js users */
+        configWithNonSafeIntGas as any,
+      );
+
+      assert.ok(
+        validationErrors.length > 0,
+        "validation errors should be present",
+      );
+      // TODO: the error message should be "Expected 'auto', a safe int, or bigint"
```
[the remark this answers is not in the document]
alcuadrado: Same as https://github.com/NomicFoundation/hardhat/pull/5762#discussion_r1775731095

v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
schaable: the problem is that `unionType` expects the first argument to be an array of zod types, and we're using `isObject` in `sensitiveUrlType`. Maybe we could replace it with `z.object()` or `z.record()`, but we'd need to make it generic.

v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
schaable: I've updated the code to use a `gasSchema`, and replaced the `unionType` with a `conditionalUnionType`. I've also added the `.positive` check for both `int` and `bigint`. See [`87f5940` (#5762)](https://github.com/NomicFoundation/hardhat/pull/5762/commits/87f5940a467955eab4663db12716030c95c895fe)

v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
schaable: I've updated `sensitiveUrlType` and `sensitiveStringType` to use `unionType` (and renamed them to schemas). However, there's a bug in Zod that prevents the union from using the provided `errorMap` when the string schema is `z.string().url()` (and likely other validations). See: https://github.com/NomicFoundation/hardhat/blob/ae009033ab5cb130042b33707f17a1ca12631f06/v-next/hardhat-zod-utils/src/index.ts#L156-L160

PR #5764, opened 2024-09-19
feat!: Stop supporting console.log selectors with "(u)int" aliases

Xanewok:
- [X] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team (@alcuadrado, @fvictorio).

---

As discussed in Slack, this removes the support for selectors that use the faulty ABI encoding. To reiterate, I searched these selectors in GitHub and I found either someone's Hardhat forks or tools that do not seem to have much usage.

We started shipping correct functions in console.log since https://github.com/NomicFoundation/hardhat/commit/bb17b503b76b4c7a29b6263d81022b93459994fb (Jul 2022, Hardhat 2.10.2) and it looks like Foundry introduced a patch that normalized our faulty selectors at least since https://github.com/foundry-rs/foundry/pull/6394/files#diff-4726f7deca3fcac31136a3902fd173c49a78f3c3b6322b9bf98c7226b0ebc10a (Nov 2023). I also bumped the changeset to a minor version, as this has a non-zero potential of breaking things.

Because the chances of someone deploying a contract that contains a console.log are slim and even then the value seems to be low, this should be a relatively safe change to make; we can bring it back or re-introduce a grace period then.

@DaniPopes since you worked on this from the Foundry's side of things, do you think there's something else that we'd need to be aware of? Would you still support these fixed selectors in Foundry?


fvictorio (approved):

DaniPopes:
We support both kinds of `console`, and treat them exactly the same since https://github.com/foundry-rs/foundry/pull/6504. Before this, people were using `console2` from forge-std because the `console` calls would not show up in traces.

More recently, we made `console2` have the same implementation as `console` in https://github.com/foundry-rs/forge-std/pull/573; this is currently still using the aliases for backwards compatibility, but if you're going to stop supporting it we can also follow.

I'm not aware of any breakage other than very old Hardhat/Foundry versions, so we wouldn't mind dropping support for it entirely since it would simplify code and reduce binary size.

Xanewok:
Sounds good, thanks for the context!

We're waiting for @alcuadrado sign off here; we can then drop the support for the alias signatures moving forward.

alcuadrado:
Looks good! Thanks, @DaniPopes, for chiming in.

Xanewok:
This is blocked until we release a Hardhat version that uses the EDR-powered stack tracing.

PR #5769, opened 2024-09-23
Run solidity tracing logic in EDR

fvictorio:


packages/hardhat-core/test/internal/hardhat-network/stack-traces/test.ts
```diff
@@ -335,22 +340,15 @@ function compareStackTraces(
...
-      decodedMessage = actualMessage.decodeError();
-    }
+    // actual.message is a ReturnData in revert errors but in custom errors
+    // we need to decode it
+    const decodedMessage =
+      "message" in actual
+        ? actual.message
+        : "returnData" in actual &&
+          new ReturnData(actual.returnData).isErrorReturnData()
+        ? new ReturnData(actual.returnData).decodeError()
```
kanej, line 350: This is a quibble, and not one I would block on, but I would be tempted to make `isErrorReturnData` a static, and pass `actual.returnData` in as an argument to that static method.

fvictorio: I'm not sure if we can have static methods using napi-rs though. If we can, yes, I agree both with the idea and with doing it in a separate PR.

Btw, these tests will be removed in a follow-up PR, so even if we do it it will only show up in the EDR repo.

kanej: In which case, this is fine.

kanej (approved):

PR #5774, opened 2024-09-24
feat: implement alpha of hardhat --init for v3

galargh:

---

This PR implements the alpha version of `hardhat --init` for Hardhat v3. This is a work-in-progress as the [design document](https://www.notion.so/nomicfoundation/Forking-re-visited-3cbcd925f12e45f79e8387ddf481945d?d=1a78dccad23e42598b91ee151c251d9b#9eb61126d63a435c8328f78ca8fbfd5d) (pending approval).

###### Templates

This PR adds a new top-level directory to the `hardhat` package - [templates](https://github.com/NomicFoundation/hardhat/tree/4c2238ee42546a8ee9f00a7dcf53505f1ffdf0c6/v-next/hardhat/templates). In that directory, you'll find templates from which `hardhat` projects can be initiated.

Each template consists of a `package.json` and other template files. As the templates are part of the workspace, they can reference workspace packages - `hardhat` itself, in particular. During project initialization, the workspace references are replaced with the currently running version of `hardhat`.

###### "Command"

[initHardhat](https://github.com/NomicFoundation/hardhat/blob/4c2238ee42546a8ee9f00a7dcf53505f1ffdf0c6/v-next/hardhat/src/internal/cli/init/init.ts#L74) serves as the entry point to the project initialization flow. Its' structure evolved from what I encountered here to begin with and it should make sense to follow that evolution while reviewing the PR commit by commit. However, with the amount of changes introduced to the flow, it might be easier to treat it as an entirely new form and review altogether as-is.

Reading `initHardhat`'s implementation should give the reader an idea of what the initialization flow looks like as it only calls other, helper (hopefully, well-named) functions. These other functions contain more and more specific implementation logic.

###### Tests

I refactored and introduced new project initialization tests in the [init directory](https://github.com/NomicFoundation/hardhat/tree/4c2238ee42546a8ee9f00a7dcf53505f1ffdf0c6/v-next/hardhat/test/internal/cli/init).

Where it made sense, I tried to test the initialization logic in isolation. However, I also preserved the tests that run the init command as a whole. One important thing not covered by the tests is the process of installing project dependencies - I felt like this would be too time-consuming.

###### Initializing an empty project

tmp2x3s5ysc-ascii

###### Trying to re-initialize a project

tmp764c5ygp-ascii

###### Initializing a partially initialized project

tmpagc8byz2-ascii


v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -145,6 +145,8 @@ Please double check whether you have multiple versions of the same plugin instal
       websiteTitle: "Hardhat project already created",
       websiteDescription: `Cannot create a new Hardhat project, the current folder is already associated with a project.`,
     },
+    // TODO: This error is no longer relevant, is it OK to remove it?
+    // Should we update the following error numbers afterwards?
     NOT_INSIDE_PROJECT_ON_WINDOWS: {
```
galargh, line 150: Should/could we remove this error now?

v-next/hardhat/src/internal/cli/init/package-manager.ts
```diff
@@ -0,0 +1,153 @@
...
+import { exists } from "@ignored/hardhat-vnext-utils/fs";
+import semver from "semver";
+
+type PackageManager = "npm" | "yarn" | "pnpm";
+
+/**
+ * getPackageManager returns the name of the package manager used in the workspace.
+ * It determines this by checking the presence of package manager specific lock files.
+ *
+ * @param workspace The path to the workspace to initialize the project in.
```
galargh, line 13:
```suggestion
 * @param workspace The path to the workspace where the package manager should be checked.
```

v-next/hardhat/src/internal/cli/init/package-manager.ts
```diff
@@ -0,0 +1,153 @@
...
+  const command = packageManagerToCommand[packageManager];
+  command.push(...dependencies);
+  return command;
+}
+
+/**
+ * installsPeerDependenciesByDefault returns true if the package manager
+ * installs peer dependencies by default.
+ *
+ * @param workspace The path to the workspace to initialize the project in.
```
galargh, line 57:
```suggestion
 * @param workspace The path to the workspace where the package manager will operate.
```

v-next/hardhat/src/internal/utils/package.ts
```diff
@@ -17,3 +18,27 @@ export async function getHardhatVersion(): Promise<string> {
 
   return packageJson.version;
 }
+
+export async function getLatestHardhatVersion(): Promise<string> {
```
galargh, line 22: I think we should remove it after all and use the current hardhat version instead. By doing that we can add proper testing to the project templates (as they are coupled with the initialization process).

v-next/hardhat/templates/javascript-with-examples/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "javascript-with-examples",
+  "description": "A JavaScript Hardhat project with examples",
+  "type": "module",
+  "devDependencies": {
+    "@ignored/hardhat-vnext": "workspace:^3.0.0-next.3",
+    "@ignored/hardhat-vnext-node-test-runner": "workspace:^3.0.0-next.2"
```
galargh, line 7: A project won't be able to initialize from this template right now because we haven't published `hardhat-vnext-node-test-runner` yet.

v-next/hardhat/templates/typescript-with-examples/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "typescript-with-examples",
+  "description": "A TypeScrpit Hardhat project with examples",
+  "type": "module",
+  "devDependencies": {
+    "@ignored/hardhat-vnext": "workspace:^3.0.0-next.3",
+    "@ignored/hardhat-vnext-node-test-runner": "workspace:^3.0.0-next.2",
```
galargh, line 7: A project won't be able to initialize from this template right now because we haven't published `hardhat-vnext-node-test-runner` yet.

v-next/hardhat-errors/src/descriptors.ts
alcuadrado: yes, this can be removed now

alcuadrado: When you remove them, the tests will force you to update the error numbers

v-next/hardhat/templates/empty-javascript/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "empty-javascript",
+  "description": "An empty JavaScript Hardhat project",
+  "type": "module",
+  "devDependencies": {
+    "@ignored/hardhat-vnext": "2.0.0"
```
alcuadrado, line 6: Why do some of these tempaltes have `workspace:` dependencies and others don't?

galargh: I wanted to test the path with `hardhat --init` actually installing the dependencies from `npm`. 

The way I propose handling `workspace:` deps from the templates is to replace them with whatever `hardhat` version we're running the `--init` from. That way, we won't have to update the hardhat version in template package.json(s) all the time. At the same time, we'll be able to test the templates themselves as they are and where they are with a workspace version of `hardhat`.

The one downside is that when we're running `--init` from an unreleased version of `hardhat`, we cannot just install the project deps from `npm` because they are not there yet. If we wanted to handle that better, maybe we could introduce some test flag/env var for that?

galargh: Having said that, I think we should reference the workspace packages in the template `package.json`s.

```suggestion
    "@ignored/hardhat-vnext": "workspace:^3.0.0-next.3"
```

v-next/hardhat/templates/empty-typescript/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "empty-typescript",
+  "description": "An empty TypeScrpit Hardhat project",
+  "type": "module",
+  "devDependencies": {
+    "@ignored/hardhat-vnext": "2.0.0",
```
galargh, line 6:
```suggestion
    "@ignored/hardhat-vnext": "workspace:^3.0.0-next.3",
```

v-next/hardhat/src/internal/utils/package.ts
```diff
@@ -17,3 +18,27 @@ export async function getHardhatVersion(): Promise<string> {
...
+    return cachedLatestHardhatVersion;
+  }
+
+  const packageJson: PackageJson = await readClosestPackageJson(
+    import.meta.url,
+  );
+
+  const packageName = packageJson.name;
+
+  const latestHardhat = await fetch(
```
alcuadrado, line 33: I think I misunderstood this other PR: https://github.com/NomicFoundation/hardhat/pull/5789

We just want to avoid using `fetch` everywhere, not during tests. Instead, we should be using `hardhat-utils/request`.

That's why I thought that a linter rule would be better.

v-next/hardhat/src/internal/config-loading.ts
```diff
@@ -45,14 +45,14 @@ export async function resolveHardhatConfigPath(
  * @returns The absolute path to the closest Hardhat config file.
  * @throw HardhatError if no Hardhat config file can be found.
  */
-export async function findClosestHardhatConfig(): Promise<string> {
-  let hardhatConfigPath = await findUp("hardhat.config.ts");
+export async function findClosestHardhatConfig(from?: string): Promise<string> {
```
alcuadrado, line 48: This function is public, so we shouldn't change it's signature for testing, or change the export from `src/hre.ts` to be an alias that always passes `undefined` here.

If we do the alias, we should also have a docblock there.

v-next/hardhat/src/internal/cli/init/subprocess.ts
```diff
@@ -0,0 +1,24 @@
...
+  console.log(command, args.join(" "));
+  const child = nodeSpawn(command, args, options);
+  await new Promise<void>((resolve, reject) => {
+    child.on("close", (code) => {
+      if (code !== 0) {
+        reject(
+          new Error(
+            `Command "${command} ${args.join(" ")}" exited with code ${code}`,
+          ),
+        );
```
alcuadrado, line 19:
```suggestion
        );
        return;
```

if `code !== 0`, this promise will be both rejected and resolved otherwise.

v-next/hardhat/src/internal/cli/init/subprocess.ts
```diff
@@ -0,0 +1,24 @@
+import type { LastParameter } from "../../../types/utils.js";
+
+import { spawn as nodeSpawn } from "node:child_process";
+
+export async function spawn(
+  command: string,
+  args: string[],
+  options: LastParameter<typeof nodeSpawn>,
+): Promise<void> {
+  console.log(command, args.join(" "));
```
alcuadrado, line 10: You forgot this debug log

v-next/hardhat/src/internal/cli/init/template.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      if (!(await isDirectory(pathToTemplate))) {
+        return;
+      }
+
+      // Validate that the the template has a package.json file
+      if (!(await exists(pathToPackageJson))) {
+        // eslint-disable-next-line no-restricted-syntax -- this is a valid usecase for the PackageJsonNotFoundError
+        throw new PackageJsonNotFoundError(pathToPackageJson);
+      }
```
alcuadrado, line 52: I don't think the eslint error should be ignored. In Hardhat we intentionally only throw `HardhatError`s, which are more helpful for a user.

But I'd say that more than a validation, this is checking an invariant that should always hold.

I would replace this with something like

```ts
assertHardhatInvarian(await exists(pathToPackageJson), `package.json for template ${name} is missing`);
```

The reason is that if that is `false`, it's an important bug that we need to fix.

v-next/hardhat/src/internal/cli/init/template.ts
```diff
@@ -0,0 +1,83 @@
...
+  findClosestPackageRoot,
+  PackageJsonNotFoundError,
+  type PackageJson,
+} from "@ignored/hardhat-vnext-utils/package";
+
+export interface Template {
+  name: string;
+  packageJson: PackageJson;
+  path: string;
+  files: string[];
```
alcuadrado, line 20: Maybe adding a comment or making it explicit in the name that these paths are relative and the above one is absolute?

v-next/hardhat/src/internal/cli/init/init.ts
```diff
@@ -1,36 +1,184 @@
...
+export async function getWorkspace(workspace?: string): Promise<string> {
+  // Ask the user for the workspace to initialize the project in if it was not provided
+  if (workspace === undefined) {
+    workspace = await promptForWorkspace();
+  }
+
+  workspace = path.resolve(workspace);
+
+  if (!(await exists(workspace))) {
+    throw new HardhatError(HardhatError.ERRORS.GENERAL.UNSUPPORTED_OPERATION, {
```
alcuadrado, line 174: This should have its own error descriptor

v-next/hardhat/src/internal/cli/init/init.ts
```diff
@@ -45,9 +193,195 @@ async function throwIfCwdAlreadyInsideProject() {
...
+  const workspaceDependencies = workspacePkg.devDependencies ?? {};
+  const dependenciesToInstall = Object.entries(templateDependencies)
+    .filter(([name]) => workspaceDependencies[name] === undefined)
+    .map(([name, version]) => {
+      // If the version is workspace:, replace it with the current version of Hardhat
+      if (version.startsWith("workspace:")) {
+        return `"${name}@${hardhatVersion}"`;
+      }
+      return `"${name}@${version}"`;
+    });
```
alcuadrado, line 353: This part won't work, as the versions of every package in the mono repo are independent. I think it's better not to use `workspace:` in the templates and handle those versions manually (i.e. bump them when appropriate).

alcuadrado: Thanks for documenting the behavior in the `README.md`, btw.

That would be useful in the future.

v-next/hardhat/src/internal/cli/init/package-manager.ts
```diff
@@ -0,0 +1,153 @@
...
+ * @returns True if the package manager installs peer dependencies by default, false otherwise.
+ */
+export async function installsPeerDependenciesByDefault(
+  workspace: string,
+  packageManager: PackageManager,
+  version?: string,
+  config?: Record<string, string>,
+): Promise<boolean> {
+  switch (packageManager) {
+    case "npm":
```
alcuadrado, line 70: Oh, nice, I thought this function was going to be `packageManager !== 'yarn'`. But this is more robust.

v-next/hardhat/src/internal/cli/init/package-manager.ts
```diff
@@ -0,0 +1,153 @@
...
+  const pathToYarnLock = path.join(workspace, "yarn.lock");
+  const pathToPnpmLock = path.join(workspace, "pnpm-lock.yaml");
+
+  if (await exists(pathToYarnLock)) {
+    return "yarn";
+  }
+  if (await exists(pathToPnpmLock)) {
+    return "pnpm";
+  }
+  return "npm";
```
alcuadrado, line 28: What if instead of always defaulting to `npm` here we add a check based on `import.meta.filename`? That way we can know what was used to install Hardhat itself.

Something like:

```ts
if (import.meta.filename.includes("yarn")) {
  return "yarn";
}

if (import.meta.filename.includes("pnpm")) {
  return "pnpm";
}
```

I think we should prioritize the lockfiles, but this can help a bit.

v-next/hardhat/src/internal/cli/init/prompt.ts
```diff
@@ -0,0 +1,90 @@
+import type { Template } from "./template.js";
+
+import { HardhatError } from "@ignored/hardhat-vnext-errors";
+
+export async function promptForWorkspace(): Promise<string> {
```
alcuadrado, line 5: I think this function would be better if it is aware of CWD and handles that instead, leading to a nicer user experience.

I guess this would mean two things:

1) We use `"."` as `initial`. ie. the relative path to CWD from CWD.
2) When we get a response, we do `resolveFromRoot(worspaceResponse, process.cwd())`. — This function is in hh-utils.

alcuadrado (approved):
I left a few comments, but this looks good. Feel free to address them and merge if the changes are not large.

Another comment that can't be left here, but I think it's necessary:

* Now we import a few dependencies in init, so I think we should import it dynamically in `main.ts`: only when it's needed.

v-next/hardhat/src/internal/utils/package.ts
galargh: Yes! That makes so much more sense now. I replaced `fetch` with `getRequest` now and reverted the `doNotUseFetch` PR in https://github.com/NomicFoundation/hardhat/pull/5801

I also restored using network (i.e. accessing npm registry) during tests for now - I think we can explore it as a generic follow-up item.

v-next/hardhat/src/internal/config-loading.ts
galargh: `findClosestHardhatConfig` is not exported from `src/hre.ts` and, other than `initHardhat` now, it was only ever used within the boundries of `config-loading.ts` - in `resolveHardhatConfigPath` to be exact. `resolveHardhatConfigPath` is also the function that's exported from `src/hre.ts`, btw. 

Do you think it would be OK to modify it then or should we leave it as is and maybe introduce a second function (something like `findClosestHardhatConfigFrom(from: string)`)?

v-next/hardhat/src/internal/cli/init/subprocess.ts
galargh: Great catch! Thank you 🙇

v-next/hardhat/src/internal/cli/init/subprocess.ts
galargh: Logging the command that is to be executed was intentional because I felt it communicated better to the user what's happening, but I do 100% agree that it looks odd here. I moved the `console.log` to the caller now where we have all the other UI handling. It definitely feels much more in-place there.

v-next/hardhat/src/internal/cli/init/template.ts
galargh: That's an excellent suggestion. We should definitely catch this before the code reaches our users. I really like the invariant solution - I added it now.

v-next/hardhat/src/internal/cli/init/init.ts
galargh: Of course! It also helped me find another place where I used `UNSUPPORTED_OPERATION` where in fact we already had a more specific error available. Fixed both now.

v-next/hardhat/src/internal/cli/init/init.ts
galargh: Ay, of course, I forgot that we don't do mass-releases of all the packages. I removed replacing the `workspace:` prefixed versions with current hardhat version. Instead, I added stripping `workspace:` prefixes. That way we should still be able to quickly test the templates with the workspace packages while delivering concrete versions (that we have full control over) to our users. In practice, during project initialization:

```json
  "devDependencies": {
    "typescript": "~5.5.0"
  }
```

would become

```json
  "devDependencies": {
    "@ignored/hardhat-vnext": "^3.0.0-next.3",
    "typescript": "~5.5.0"
  }
```

Let's iterate on it in a follow-up if needed.

v-next/hardhat/src/internal/cli/init/package-manager.ts
galargh: Won't the `import.meta.filename` just contain an absolute path to `package-manager.js`?

v-next/hardhat/src/internal/cli/init/prompt.ts
galargh: Good suggestion, I updated the prompt to use paths relative to current working directory. Thanks :)

v-next/hardhat/test/internal/cli/init/init.ts
```diff
@@ -1,150 +1,192 @@
...
-        process.env.HARDHAT_CREATE_EMPTY_TYPESCRIPT_HARDHAT_CONFIG = "true";
-
-        await assertRejectsWithHardhatError(
-          async () => initHardhat(),
-          HardhatError.ERRORS.GENERAL.ONLY_ESM_SUPPORTED,
-          {},
-        );
-      });
+  describe("when force is false", () => {
+    it("should copy the template files to the workspace while skipping existing files", async () => {
```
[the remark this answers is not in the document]
galargh: The previous one tests copying files with `force=true`. Then, the expectation is for any clashing template files that already exist in the workspace to be overwritten. When `force=false`, we want to skip clashing files instead. I added some inline comments to make it a little clearer.

v-next/hardhat/src/internal/config-loading.ts
kanej: Agreed I think `resolveHardhatConfigPath` is in the public API, not `findClosestHardhatConfig`.

I think we leave the `from` addition rather than introducing a second function.

alcuadrado: oops, sorry, i misread the reexport 😅 @galargh

PR #5776, opened 2024-09-25
Define the `HookContext` as a partial view of the `HRE`

alcuadrado:
This PR addresses a limitation in Hardhat's core architecture discovered by @ChristopherDedominici.

## Explanation of the problem

Chris is working on the first few hook handlers that use the builtin plugins, and in particular, the extensions to the `HRE` that they make.

This was problematic, as `HookHandler`s don't receive the `HRE`, but a `HookContext` instead. The problem was that any extension to the `HRE` wasn't reflected in the `HookContext`.

The' HookContext' intends to prevent `HookHandler`'s from accessing the `TaskManager` and running tasks, as these are separate concerns in v3, and not to prevent them from accessing the builtin plugins.

## Proposed solution

This solution has a type-level and runtime-level component.

### Types

Instead of defining the `HookContext` as a separate interface, define it as a derived type of the `HRE`:

```ts
export type HookContext = Omit<HardhatRuntimeEnvironment, "tasks">;
```

This way, every extension to the `HardhatRuntimeEnvironment` is also available in the `HookContext`.

### Runtime

To make the values (as in runtime values, not types) of the extensions available through the `HookContext`, without creating a new hook nor restructuring the system, I used `Object.create` to create the `HookContext`, and overwrote `tasks` with `undefined`.

Contrary to making a shallow copy at initialization, defining the HRE as its prototype means that any future modification to the HRE will also be visible.

### DIFF
--- a/v-next/hardhat/src/internal/core/hre.ts
+++ b/v-next/hardhat/src/internal/core/hre.ts
@@ -118,15 +118,6 @@ export class HardhatRuntimeEnvironmentImplementation
 
     const interruptions = new UserInterruptionManagerImplementation(hooks);
 
-    const hookContext: HookContext = {
-      hooks,
-      config,
-      globalOptions,
-      interruptions,
-    };
-
-    hooks.setContext(hookContext);
-
     const hre = new HardhatRuntimeEnvironmentImplementation(
       extendedUserConfig,
       config,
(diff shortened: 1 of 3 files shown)


schaable (approved):
Looks good! Using Object.create was really smart 💯

PR #5787, opened 2024-09-27
Port 'hardhat-ethers' plugin from V2 to V3

ChristopherDedominici:
# Short overview on how to use the package

```ts
const { ethers } = await hre.network.connect();

// ethers functionalities
ethers.isAddress("0x1234567890123456789012345678901234567890");

// ethers.Provider functionalities
await ethers.provider.getBlockNumber();

// Hardhat helper methods
await ethers.getSigners();
```

# Issue to track TODOs as soon as V3 node is ready

[Link to issue](https://github.com/NomicFoundation/hardhat/issues/5788)

# Questions: input needed

- In the [old v2 code](https://github.com/NomicFoundation/hardhat/blob/198612d378f1e7c09fbf44b6e7622c3d36618829/packages/hardhat-ethers/src/signers.ts#L59) there is a property called `blockGasLimit` that is assigned to `gasLimit` under specific conditions. But currently the new `NetworkConfig` does not have this property. Is it gonna be supported in the future? Or this logic can be removed?

- in the code I kept the utils functions from `ethers` (e.g.: `isHexString`). The reason behind this choice is to have the code logic as close as possible to the original `ethers` library. Do we want to keep these ethers utils'functions or do we want to use the hardhat ones from `hardhat-utils`?

- the old code was using`sinon`, specifically in [this file](https://github.com/NomicFoundation/hardhat/blob/198612d378f1e7c09fbf44b6e7622c3d36618829/packages/hardhat-ethers/test/provider-events.ts), to ensure that the listener functions were called correctly. I replaced sinon with the builtin nodeJS test runner mocks. Is this ok?

- the old V2 code had a few TODOs. I kept them, since this PR is only about porting the code from V2 to V3. Is this ok?

- As mentioned in the [design doc](https://www.notion.so/nomicfoundation/hardhat-ethers-104578cdeaf580fa9ab0e9172952e33a), the current implementation is using the `mocked ArtifactsManager`. For the tests, I copied the code of the mocked artifact manager from [this test example](https://github.com/NomicFoundation/hardhat/blob/198612d378f1e7c09fbf44b6e7622c3d36618829/v-next/hardhat/test/internal/example-mock-artifacts-plugin-using-test.ts) and I modified it a bit, [see here](https://github.com/NomicFoundation/hardhat/pull/5787/files#r1778893511). Is this ok?
Also, as mentioned in one of the team syncing calls, the artifacts used in the tests are currently stored in files but they just include the artifact properties and not the code that originated them. See the [artifacts folder here](https://github.com/NomicFoundation/hardhat/pull/5787/files#r1778891066). This is because we currently do not know if we're going to use this approach (in this case the artifacts should have additional information like the source code, etc.) or if we're going to use the old approach of compiling the code during the tests. Is it ok to keep the artifact in this format right now?

- in the old V2 code the hardhat helpers were separated in functions, [see here](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-ethers/src/internal/helpers.ts). In V3 I grouped them into a class ([see here](https://github.com/NomicFoundation/hardhat/pull/5787/files#r1780713401)), to simplify the usage of the dependencies that before were imported from the HardhatRuntimeEnvironment. For example, before the functionalities available in the `hre` (like hre.artifacts ) had to be passed several times across functions, now they can be directly accessed via `this`. This approach does not affect how the methods are exposed to the end user. Is this ok?

# Notes

- As mentioned in the [design doc](https://www.notion.so/nomicfoundation/hardhat-ethers-104578cdeaf580fa9ab0e9172952e33a), there are a few parts of the code that can only be implemented when V3 has all the features that V2 has. This is the list of those parts:
    - tests
    - property `blockGasLimit` (if it's still gonna be supported - see the question asked in the previous section)

To find these portions of code that require the missing V3 features you can search in the code for `TODO: enable when V3 is ready`.

- In the old V2 code there were a lot of disabled eslint rules. I removed most of them, but a few are required. In the code, they can be found by searching for `eslint-disable-next-line`. In the `ethers-utils.ts` file all of them have been kept, because almost all the methods in the file are a verbatim copy of the `ethers` methods. There are also a few disabled eslint rules in the tests, because there are tests that check errors that are not thrown by Hardhat, so there is a warning when using the `assert.rejects` method provided by the nodeJS test runner.

- The old V2 files where quite long, but as agreed in one of the team syncing calls, in V3 we will keep the same file structure as in V2.

# Tests

For an explanation on why we currently need a workaround to run the tests, please read the [design doc here](https://www.notion.so/nomicfoundation/hardhat-ethers-104578cdeaf580fa9ab0e9172952e33a).

How to run the tests:

- start a V2 hardhat node with `npx hardhat node`
- in the `hardhat-ethers` package run the command `pnpm test:tmp`

If you want to run the tests using an INFURA_URL, you can set the env variable INFURA_URL to the desired URL. If the env variable is not set, the test will be skipped.


v-next/hardhat-ethers/test-tmp/helpers/artifacts/ambiguous-library.ts
```diff
@@ -0,0 +1,14 @@
+import type { Artifact } from "@ignored/hardhat-vnext/types/artifacts";
```
ChristopherDedominici, line 1: These are the mocked artifacts mentioned in the PR description that currently do not have the source code that generates them, as discussed during the team meeting sync

v-next/hardhat-ethers/test-tmp/helpers/artifact-manager-mock.ts
```diff
@@ -0,0 +1,122 @@
...
+  readonly #artifacts: Map<string, Artifact>;
+  readonly #artifactsPaths: Map<string, string>;
+
+  constructor(
+    mockedArtifacts?: Array<{ artifactName: string; fileName: string }>,
+  ) {
+    this.#artifacts = new Map();
+    this.#artifactsPaths = new Map();
+
+    // An array of elements, where every element has an artifact name and artifact file name, is passed as argument during initialization and stored in the map.
```
ChristopherDedominici, line 21: This is the code I copied from the mocked ArtifactsManager PR. As mentioned in the PR description, I modified the logic slightly to suit my specific test scenarios

v-next/hardhat-ethers/src/types.ts
```diff
@@ -0,0 +1,73 @@
+import type { HardhatEthersProvider } from "./internal/hardhat-ethers-provider/hardhat-ethers-provider.js";
```
ChristopherDedominici, line 1: I need a more careful review of the types here, to be sure that they properly align with the V3 artifact implementation

v-next/hardhat-ethers/src/internal/initialization.ts
```diff
@@ -0,0 +1,47 @@
...
+  return {
+    ...ethers,
+
+    provider,
+
+    getSigner: (address: string) => hardhatHelpers.getSigner(address),
+    getSigners: () => hardhatHelpers.getSigners(),
+    getImpersonatedSigner: (address: string) =>
+      hardhatHelpers.getImpersonatedSigner(address),
+    getContractFactory: hardhatHelpers.getContractFactory.bind(hardhatHelpers),
```
ChristopherDedominici, line 39: The bind is necessary because otherwise in the function the "this" that refers to the HardhatHelpers class will be overwritten

v-next/hardhat-ethers/test-tmp/plugin-functionalities.ts
```diff
@@ -0,0 +1,1261 @@
...
+  initializeTestEthers,
+} from "./helpers/helpers.js";
+
+describe("Ethers plugin", () => {
+  let ethers: HardhatEthers;
+  let artifactsManager: ArtifactsManager;
+
+  beforeEach(async () => {
+    // Declare all the artifacts that we need during the test
+    ({ ethers, artifactsManager } = await initializeTestEthers([
```
ChristopherDedominici, line 31: these are the tests with a heavy usage of artifacts

v-next/hardhat-ethers/src/internal/hardhat-helpers/hardhat-helpers.ts
```diff
@@ -0,0 +1,473 @@
+import type {
```
ChristopherDedominici, line 1: As mentioned in the PR description:

in the old V2 code the hardhat helpers were separated in functions, [see here](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-ethers/src/internal/helpers.ts). In V3 I grouped them into a class, to simplify the usage of the dependencies that before were imported from the HardhatRuntimeEnvironment. For example, before the functionalities available in the `hre` (like hre.artifacts ) had to be passed several times across functions, now they can be directly accessed via `this`

v-next/hardhat-ethers/src/internal/signers/signers.ts
```diff
@@ -0,0 +1,280 @@
...
+          try {
+            await provider.send("hardhat_metadata");
+            isHardhatNetwork = true;
+          } catch {}
+
+          if (isHardhatNetwork) {
+            // WARNING: this assumes that the hardhat node is being run in the
+            // same project which might be wrong
+            // TODO: enable when V3 is ready: blockGasLimit currently missing in networkConfig
+            // gasLimit = networkConfig.blockGasLimit;
```
ChristopherDedominici, line 70: As mentioned in the PR description, this is the property (blockGasLimit) that is currently not supported in `networkConfig`

v-next/hardhat-ethers/src/internal/initialization.ts
alcuadrado: i understand the reason behind this, but not sure why it's different in the next line. I think everything could use the same syntax

v-next/hardhat-ethers/src/internal/signers/deep-copy.ts
```diff
@@ -0,0 +1,54 @@
...
+  "boolean",
+  "function",
+  "number",
+  "string",
+  "symbol",
+];
+
+export async function deepCopy<T = any>(value: T): Promise<T> {
+  // The function 'deepClone' from 'hardhat-utils' cannot be used to replace this function, it won't properly clone
+  // the value.
```
alcuadrado, line 15: I'd add an explanation about the "why" to this comment

alcuadrado: I saw that dis code comes from v2, so keeping it makes sense

v-next/hardhat-ethers/src/internal/signers/signers.ts
```diff
@@ -0,0 +1,280 @@
...
+    networkConfig: NetworkConfig,
+    address: string,
+  ): Promise<HardhatEthersSigner> {
+    this.networkName = networkName;
+    this.networkConfig = networkConfig;
+
+    // depending on the config, we set a fixed gas limit for all transactions
+    let gasLimit: bigint | undefined;
+
+    if (networkName === "hardhat") {
```
alcuadrado, line 40: This changed. Now this check would be `netorkConfig.type !== "http"`.

I think we should create an issue to address this entire section in a follow up pr.

v-next/hardhat-ethers/src/internal/signers/deep-copy.ts
ChristopherDedominici: I added a comment to better explain why `deepClone` cannot be used to replace `deepCopy`: https://github.com/NomicFoundation/hardhat/pull/5787/commits/8348448bf64bebb099fb158c6e9bb268211167b7

v-next/hardhat-ethers/src/internal/signers/signers.ts
ChristopherDedominici: I created an issue to track it and I added it in the PR description too.

https://github.com/NomicFoundation/hardhat/issues/5796

v-next/hardhat-ethers/package.json
```diff
@@ -0,0 +1,75 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/NomicFoundation/hardhat",
+    "directory": "v-next/ethers"
+  },
+  "author": "Nomic Foundation",
+  "license": "MIT",
+  "type": "module",
+  "exports": {
+    ".": "./dist/src/index.js"
```
alcuadrado, line 15: I think we should also have an export for the types

ChristopherDedominici: Yes, good catch. [Done here](https://github.com/NomicFoundation/hardhat/pull/5787/commits/5a170acac329dc58922f1a11a37fa813cc874dcb)

v-next/hardhat-ethers/src/types.ts
alcuadrado: I double checked this, I think it looks good.

alcuadrado (approved):

PR #5789, opened 2024-09-27
feat: add doNotUseFetch helper to the test utils

galargh:

---

I found it helpful during the draft implementation of https://github.com/NomicFoundation/hardhat/pull/5774. 

The idea is to override the global reference to fetch with a function that always throws an error. This is a very crude approach to simulating network connectivity issues when using fetch. We can explore this area further in the future if needed. However, at this time, this met my needs.

### DIFF
--- a/v-next/hardhat-test-utils/src/global.ts
+++ b/v-next/hardhat-test-utils/src/global.ts
@@ -1,0 +1,16 @@
+import { after, before } from "node:test";
+
+export function doNotUseFetch(): void {
+  let originalFetch: typeof global.fetch;
+
+  before(async () => {
+    originalFetch = global.fetch;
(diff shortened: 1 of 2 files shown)


alcuadrado (commented):
Interesting PR, I think I'd just rename the path to `/http` instead of `/global`, as it'd be easier to discover in the future.

alcuadrado (approved):

alcuadrado:
Now that I think of, maybe it'd be better to use a linter rule

galargh:
> Now that I think of, maybe it'd be better to use a linter rule that directs you to the utils package

Do you mean one that does that when someone tries to overwrite `global.fetch` directly?

PR #5798, opened 2024-10-02
fix: bump find-up dependency

kiriyaga-txfusion:

---

Since your plugin uses older versions of `find-up` and because, we had to extend your register and script-runner. However, we couldn't resolve these dependencies because they rely on internal/errors, which also uses outdated versions. None of the usual solutions, like ignoring or module declarations, worked. Many other dependencies, such as mocha, c8, and eslint, have updated to version 5.0.0. It would be helpful if you could also update your versions

### DIFF
--- a/packages/hardhat-core/package.json
+++ b/packages/hardhat-core/package.json
@@ -122,7 +122,7 @@
     "env-paths": "^2.2.0",
     "ethereum-cryptography": "^1.0.3",
     "ethereumjs-abi": "^0.6.8",
-    "find-up": "^2.1.0",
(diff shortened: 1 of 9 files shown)


kiriyaga-txfusion:
Hello @kanej,
Any news about this PR?

alcuadrado:
Thanks for sending this PR. Unfortunately, if conflicts with https://github.com/NomicFoundation/hardhat/pull/5767

alcuadrado:
Actually, this help debug the other pr. Thanks you.

`find-up` 7.0.0 somehow breaks mocha!

PR #5801, opened 2024-10-03
Revert "feat: add doNotUseFetch helper to the test utils"

galargh:
Reverts NomicFoundation/hardhat#5789

This can be reverted because we should not use `fetch` directly anyway. We should use `request` from `utils` instead which uses `undici` internally.

We can explore the ability to disable net connect for `undici` globally in a follow-up PR. The rough idea that we could explore would be to use `setGlobalDispatcher`/`getGlobalDispatcher` to control the global dispatcher/agent set for tests.

See https://github.com/NomicFoundation/hardhat/pull/5774#discussion_r1785096505 for details.


schaable:
Hey @galargh, I have some experience on mocking requests with `undici`, let me know if you need a hand. You can also check out `v-next/hardhat/test/internal/builtin-plugins/network-manager/http-provider.ts` for an example.

schaable (approved):

galargh:
> Hey @galargh, I have some experience on mocking requests with `undici`, let me know if you need a hand. You can also check out `v-next/hardhat/test/internal/builtin-plugins/network-manager/http-provider.ts` for an example.

Thanks! Temporarily, I decided just not to worry about it. I looked at `initializeTestDispatcher` for potential inspiration for the future. We'll ping you when/if we have to revisit this.

PR #5806, opened 2024-10-04
feat: create a mocha/ethers template project

kanej:
As part of the Alpha prep this adds a mocha/ethers template project (used during `hardhat --init`), that tries to recreate the functionality (we are currently able to support`f of the v2 Ethers typescript project in v3.

It pulls in the `v-next` versions of the mocha test runner, the test helpers and `hardhat-ethers`.

It adds a the `Lock.sol` and `Lock.ts` test file. The test file has been modified to allow for running the tests while we still have rough edges.

Specifically the rough edges are:
- we need to import plugin types to get the full connection type
- we need new patterns for sharing `connection` among tests (I added a `beforeEach`)
- we don't have chai matchers so I added `chai-as-promised` and adapted the tests, sometime crudely.

### DIFF
--- a/v-next/hardhat/templates/mocha-ethers/test/Lock.ts
+++ b/v-next/hardhat/templates/mocha-ethers/test/Lock.ts
@@ -1,0 +1,166 @@
+import hre from "@ignored/hardhat-vnext";
+import { NetworkConnection } from "@ignored/hardhat-vnext/types/network";
+import { describe, it, before } from "mocha";
+
+// TODO: Chai as promised support needs to be added either directly
+// or through the updated `chai-matchers` package.
+import * as chai from "chai";
+import chaiAsPromised from "chai-as-promised";
+chai.use(chaiAsPromised);
+const { expect } = chai;
+
+describe("Lock", function () {
+  let networkConnection: NetworkConnection<"l1">;
+
+  // We define a fixture to reuse the same setup in every test.
+  // We use loadFixture to run this setup once, snapshot that state,
+  // and reset Hardhat Network to that snapshot in every test.
+  async function deployOneYearLockFixture() {
+    const ONE_YEAR_IN_SECS = 365 * 24 * 60 * 60;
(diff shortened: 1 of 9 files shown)


v-next/hardhat/templates/mocha-ethers/test/Lock.ts
```diff
@@ -0,0 +1,156 @@
...
+
+import "@ignored/hardhat-vnext-ethers";
+import "@ignored/hardhat-vnext-network-helpers";
+
+// TODO: Chai as promised support needs to be added either directly
+// or through the updated `chai-matchers` package.
+import * as chai from "chai";
+import chaiAsPromised from "chai-as-promised";
+chai.use(chaiAsPromised);
+const { expect } = chai;
```
kanej, line 12: I applied the `chaiAsPromised` directly here.
I don't think we currently have support `.mochars.json` files to allow us to setup up supporting setup files.

v-next/hardhat/templates/mocha-ethers/test/Lock.ts
```diff
@@ -0,0 +1,153 @@
...
+
+describe("Lock", function () {
+  let connection: NetworkConnection<"l1">;
```
kanej, line 12: Now that the connection is not ambient to the global context we need a new pattern for tests.
I have no strong feelings here so opted for a shared `connection` that is explicitly typed for the expected network connection and that is setup for each test.

I suspect we may want to go with something other than `connection` as the variable name. Maybe `networkConnection` is less ambiguous?

schaable: +1 to `networkConnection`. That's the name I used in the `network-manager` tests.

schaable: I'm not sure we need to reset it between tests, though.

schaable (approved):
Looks good, just a minor comment regarding the `networkConnection` reset.

PR #5809, opened 2024-10-08
feat: replace spec with a custom hardhat solidity test reporter

galargh:

---

This PR introduces a custom hardhat solidity test reporter in place of the spec reporter that we used to use.

It is an implementation of the design outline in https://www.notion.so/nomicfoundation/Runner-Reporter-Design-11f578cdeaf580b8b01ddc43ae589f05

This PR doesn't change the output formatting.

###### Testing

The runner/reporter setup have been tested manually. It would be great to introduce a more formalised E2E testing for the solidity testing task, but it is left as a follow-up for now.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/types.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/types.ts
@@ -1,0 +1,17 @@
+import type { SuiteResult } from "@ignored/edr";
+import type { Readable } from "node:stream";
+
+export type TestStatus = "Success" | "Failure" | "Skipped";
+
+export type TestsStream = Readable;
+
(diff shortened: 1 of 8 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity-test/types.ts
```diff
@@ -0,0 +1,17 @@
+import type { SuiteResult } from "@ignored/edr";
+import type { Readable } from "node:stream";
+
+export type TestStatus = "Success" | "Failure" | "Skipped";
```
alcuadrado, line 4: I think EDR changed how they export enums now, so this can probably be improved in a later PR, so that we avoid the duplication.

alcuadrado (approved):

PR #5819, opened 2024-10-11
feat: add bytecode module with resolveLinkedBytecode and linkBytecode…

schaable:
Added a `bytecode` module which exports two functions:
 - `resolveLinkedBytecode`: which takes a contract artifact and a map of library names to their addresses, performs validations, and returns the linked bytecode.
 - `linkBytecode`: which takes a contract artifact and the resolved library links, and returns the linked bytecode without performing any validations.

Links:
[Design document](https://www.notion.so/nomicfoundation/Bytecode-library-linking-11b578cdeaf580faa434ca6951860b99?pvs=4)

### DIFF
--- a/v-next/hardhat-utils/src/bytecode.ts
+++ b/v-next/hardhat-utils/src/bytecode.ts
@@ -1,34 +1,17 @@
 import type { PrefixedHexString } from "./hex.js";
+import type {
+  Artifact,
+  LibraryAddresses,
+  LibraryLink,
+} from "./internal/bytecode.js";
 
-import {
-  AmbiguousLibraryNameError,
-  InvalidLibraryAddressError,
-  MissingLibrariesError,
-  OverlappingLibrariesError,
-  UnnecessaryLibraryError,
-} from "./errors/bytecode.js";
-import { isAddress } from "./eth.js";
 import { getPrefixedHexString, getUnprefixedHexString } from "./hex.js";
-
-interface Artifact {
-  bytecode: string;
-  linkReferences: {
-    [sourceName: string]: {
-      [libraryName: string]: Array<{ start: number; length: number }>;
-    };
-  };
-}
-
-interface LibraryLink {
-  sourceName: string;
-  libraryName: string;
-  libraryFqn: string;
-  address: string;
-}
-
-interface LibraryAddresses {
-  [contractName: string]: PrefixedHexString;
-}
+import {
+  checkAmbiguousOrUnnecessaryLinks,
+  checkMissingLibraryAddresses,
+  checkOverlappingLibraryNames,
+  checkProvidedLibraryAddresses,
+} from "./internal/bytecode.js";
 
 /**
  * Resolves the linked bytecode for a given contract artifact by ensuring all
@@ -90,7 +73,8 @@ export function linkBytecode(
   let linkedBytecode = bytecode;
 
   for (const { sourceName, libraryName, address } of libraries) {
-    const contractLinkReferences = linkReferences[sourceName]?.[libraryName];
+    const contractLinkReferences =
+      linkReferences[sourceName]?.[libraryName] ?? [];
     const unprefixedAddress = getUnprefixedHexString(address);
 
     for (const { start, length } of contractLinkReferences) {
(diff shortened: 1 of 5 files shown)


v-next/hardhat-utils/src/bytecode.ts
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ * Links the bytecode of a contract artifact with the provided library addresses.
+ * This function does not perform any validation on the provided libraries.
+ *
+ * @param artifact The contract artifact containing the bytecode and link references.
+ * @param libraries An array of LibraryLink objects representing the libraries to be linked.
+ * @returns The linked bytecode with all provided libraries correctly linked.
+ */
+export function linkBytecode(
```
kanej, line 68: Are we expecting this to be used outside of `resolveLinkedBytecode`? Or is this exported for for testing?

schaable: I don’t expect to see a lot of usage of `linkBytecode`, but the idea was to export it for cases where validation and error handling need to be handled externally.

v-next/hardhat-utils/src/internal/bytecode.ts
```diff
@@ -0,0 +1,155 @@
...
+import {
+  AmbiguousLibraryNameError,
+  InvalidLibraryAddressError,
+  MissingLibrariesError,
+  OverlappingLibrariesError,
+  UnnecessaryLibraryError,
+} from "../errors/bytecode.js";
+import { isAddress } from "../eth.js";
+
+export interface Artifact {
```
kanej, line 12: How do we think about types within the hh-utils?

I was tempted to suggest that we should be leveraging a top level `Artifact` type, but that would imply a new package that the utils could depend on that is shared with hardhat.

So would a fair description of our rule be, that these are the minimal types that describe the util functions signature and we will take the pain of keeping them in sync with any core types (e.g. core `Artifact`)?

v-next/hardhat-utils/src/bytecode.ts
```diff
@@ -0,0 +1,89 @@
...
+import {
+  checkAmbiguousOrUnnecessaryLinks,
+  checkMissingLibraryAddresses,
+  checkOverlappingLibraryNames,
+  checkProvidedLibraryAddresses,
+} from "./internal/bytecode.js";
+
+/**
+ * Resolves the linked bytecode for a given contract artifact by ensuring all
+ * required libraries are correctly linked.
```
kanej, line 18:
```suggestion
 * Resolves the linked bytecode for a given contract artifact by substituting
 * the required library placeholders within the bytecode with the provided 
 * library addresses.
```

v-next/hardhat-utils/src/errors/bytecode.ts
```diff
@@ -0,0 +1,48 @@
...
+Please provide the fully qualified name for these libraries.`);
+  }
+}
+
+export class UnnecessaryLibraryError extends CustomError {
+  constructor(formattedLibraries: string) {
+    super(`The following libraries are not referenced by the contract:
+${formattedLibraries}
+
+Please provide only the libraries that are actually needed.`);
```
kanej, line 26:
```suggestion
Please provide only the libraries that are needed.`);
```
Stop being mean to our poor users.

v-next/hardhat-utils/src/errors/bytecode.ts
```diff
@@ -0,0 +1,48 @@
+import { CustomError } from "../error.js";
+
+export class InvalidLibraryAddressError extends CustomError {
+  constructor(formattedLibraries: string) {
```
kanej, line 4: Is there an argument here for moving the formatting code that precedes the throwing of these errors into the custom errors themselves?

So passing a structured type in the constructor parameter and converting it to a string here for the message rather than putting the formatting code at the call point?

v-next/hardhat-utils/test/bytecode.ts
```diff
@@ -0,0 +1,327 @@
...
+          name: "InvalidLibraryAddressError",
+          message: `The following libraries have invalid addresses:
+\t* "Library": "0xnotanaddress"
+
+Please provide valid Ethereum addresses for these libraries.`,
+        });
+      });
+
+      it("should throw AmbiguousLibraryNameError if a library name matches multiple needed libraries", () => {
+        const artifact = {
```
kanej, line 41: Should we add the `Artifact` type explicitly here, to be consistent with `libraries`?

```suggestion
        const artifact: Artifact = {
```

kanej (changes requested):
This is looking good.

I left a few comments and minor suggestions - but my only real question is double checking on types within the utils package.

v-next/hardhat-utils/src/internal/bytecode.ts
schaable: That’s correct. Types in hh-utils should remain minimal, used internally, and not exposed in the public API. Unfortunately, this means we’ll need to manually keep them in sync with the types in Hardhat or plugins.

v-next/hardhat-utils/src/errors/bytecode.ts
schaable: The idea behind this approach was to keep the error classes simple. However, your point about shifting the responsibility of formatting to the error classes is valid. I’ll update the code accordingly.

v-next/hardhat-utils/test/bytecode.ts
kanej: Sorry, I should have said that this was a nitpick and that I was happy either way.

kanej (approved):
Looks good, let get it merged.

PR #5825, opened 2024-10-15
Port hardhat-viem to V3

schaable:
This PR depends on [this PR](https://github.com/NomicFoundation/hardhat/pull/5819)

Design doc [here](https://www.notion.so/nomicfoundation/hardhat-viem-port-to-v3-a5f66176a18148479c4c655ec0f903b9)

# What this PR addresses:
- port the hardhat-viem plugin to v3:
  - the plugin is initialized when a new network connection is made.
  - the client functions have been ported.
    - public and wallet clients are extended with L2 actions when the chain type is optimism.
  - the contract functions have been ported.
    - the types have been updated to work dynamically, avoiding type generation.
  - general cleanup:
    - removed a lot of internal functions that were originally made to simplify testing in v2
    - removed lodash.memoize and implement cache for the functions that used it
  - ported and added tests
    - the e2e test were tested locally by running a hardhat v2 node in the background. There is an issue where the test suite takes a lot of time to complete when running these tests, so I skipped them for now. I suspect it may have to do with the v2 node.
  - updated the example-project

# Notes on the implementation 
- There are some TODOs in the tests. I'll revisit this once we have the EDR provider.


v-next/hardhat-viem/.eslintignore
```diff
@@ -0,0 +1 @@
+/test/fixture-projects
```
schaable, line 1: I'm not sure if this is the best way to ignore the `fixture-projects`. Let me know if you have a better approach.

v-next/hardhat-viem/src/chains.ts
```diff
@@ -0,0 +1,139 @@
...
+import type * as viemT from "viem";
+
+import { HardhatError } from "@ignored/hardhat-vnext-errors";
+import { extractChain } from "viem";
+import * as chainsModule from "viem/chains";
+import { hardhat, anvil } from "viem/chains";
+
+/* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+-- TODO: this assertion should not be necessary */
+const chains = Object.values(chainsModule) as viemT.Chain[];
```
schaable, line 12: I had to cast this, but the [example](https://viem.sh/docs/utilities/extractChain.html#usage) in the Viem docs doesn’t require a cast. Not entirely sure why, though.

v-next/hardhat-viem/src/chains.ts
```diff
@@ -0,0 +1,139 @@
...
+    NetworkMethod.HARDHAT_METADATA,
+  );
+  isHardhatNetworkCache.set(provider, isHardhat);
+
+  return isHardhat;
+}
+
+const isAnvilNetworkCache = new Map<EthereumProvider, boolean>();
+
+export async function isAnvilNetwork(
```
schaable, line 101: This was previously called `isFoundryNetwork`, but since the `foundry` chain has been marked as deprecated in the `viem/chains` module (and `anvil` is now recommended), I renamed it accordingly.

v-next/hardhat-viem/src/clients.ts
```diff
@@ -0,0 +1,157 @@
...
+  if (chainType === "optimism") {
+    walletClient = walletClient.extend(walletActionsL2());
+  }
+
+  /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions --
+  We need to assert the type because TS gets confused with the conditional type */
+  return walletClient as GetWalletClientReturnType<ChainTypeT>;
+}
+
+export async function getDefaultWalletClient<
```
schaable, line 108: We might want to export this in the future, but for now, it's kept internal in case we decide to take a different approach for retrieving a default account.

v-next/hardhat-viem/src/contracts.ts
```diff
@@ -0,0 +1,233 @@
...
+  deployContractConfig: DeployContractConfig = {},
+): Promise<ContractReturnType<ContractName>> {
+  const {
+    client,
+    confirmations = 1,
+    libraries = {},
+    ...deployContractParameters
+  } = deployContractConfig;
+
+  if (confirmations < 0) {
```
schaable, line 38: This used to be placed lower in the code, but since it’s just a parameter check, it's better to perform the validation earlier. This way, we can fail faster in case of an error.

v-next/hardhat-viem/src/contracts.ts
```diff
@@ -0,0 +1,233 @@
...
+  const artifact = await artifacts.readArtifact(contractName);
+  let bytecode;
+  try {
+    bytecode = resolveLinkedBytecode(artifact, libraries);
+  } catch (error) {
+    ensureError(error);
+
+    throw new HardhatError(HardhatError.ERRORS.VIEM.LINKING_CONTRACT_ERROR, {
+      contractName,
+      error: error.message,
```
schaable, line 225: The actual error, along with the list of offending libraries, comes from `resolveLinkedBytecode`. I think centralizing the error format is a good approach, but another option could be returning an object containing the validation errors and formatting them within the consumer plugin.

v-next/hardhat-viem/test/utils.ts
```diff
@@ -0,0 +1,36 @@
+import type {
+  EthereumProvider,
+  RequestArguments,
+} from "@ignored/hardhat-vnext/types/providers";
+
+import EventEmitter from "node:events";
+
+export class MockEthereumProvider
```
schaable, line 8: We should consider moving this to `hardhat-test-utils`.

v-next/hardhat-viem/src/chains.ts
iosh: This ensures a clear type for chainId to meet the requirements of the extractChain function.
```ts
export function checkChainId(chainId: number): (typeof chains)[number]["id"] {
	const findId = chains.find((chain) => chain.id === chainId);
	if (findId === undefined) {
		throw new HardhatError(HardhatError.ERRORS.VIEM.NETWORK_NOT_FOUND, {
			chainId,
		});
	}
	return findId.id;
}

export async function getChainId(
	provider: EthereumProvider,
): Promise<ReturnType<typeof checkChainId>> {
	const cachedChainId = chainIdCache.get(provider);
	if (cachedChainId !== undefined) {
		return checkChainId(cachedChainId);
	}

	const chainId = Number(await provider.request({ method: "eth_chainId" }));
	chainIdCache.set(provider, chainId);

	return checkChainId(chainId);
}

```

v-next/hardhat-viem/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "tsx": "^4.11.0",
+    "typescript": "~5.5.0",
+    "typescript-eslint": "7.7.1",
+    "viem": "^2.21.17"
+  },
+  "dependencies": {
+    "@ignored/hardhat-vnext-errors": "workspace:^3.0.0-next.3",
+    "@ignored/hardhat-vnext-utils": "workspace:^3.0.0-next.3"
+  },
+  "peerDependencies": {
```
kanej, line 68: Comparing against v2, it had a dependency on `typescript`. Has the generation of types in the compilation pipeline removed that need?

v-next/hardhat-viem/src/accounts.ts
```diff
@@ -0,0 +1,22 @@
+import type { EthereumProvider } from "@ignored/hardhat-vnext/types/providers";
```
kanej, line 1: This file and the others, excluding `index.ts` and the type files should move to an `./internal` folder.

v-next/hardhat-viem/src/accounts.ts
```diff
@@ -0,0 +1,22 @@
...
+import type * as viemT from "viem";
```
kanej, line 2: Could we import `Address` directly here?

v-next/hardhat-viem/src/accounts.ts
```diff
@@ -0,0 +1,22 @@
...
+
+const accountsCache = new Map<EthereumProvider, viemT.Address[]>();
```
kanej, line 4: I am tempted here to say that we should associate the accounts cache as a private property of the `connection.viem` object, and pass it in as a parameter here - to avoid the global.

Or as an alternative use [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap), so that providers+cached accounts get garbage collected when the connection is no longer used.

v-next/hardhat-viem/src/chains.ts
schaable: Hi @iosh, thanks for your feedback! However, this PR is intended for internal review by the team, as there is specific context around v3 that external contributors might not have. For example, we need the `module` option to be `"node16"` to support ESM, so changing it to `"CommonJS"` wouldn't work. 

We’ll start accepting feedback once we have a more stable version of v3. In the meantime, feel free to contribute to or provide feedback on v2 issues. Look for issues not labeled as `v-next`.

v-next/hardhat-viem/package.json
schaable: To be honest, I don't remember why we added it as a peer dependency. It might have something to do with type generation. The [commit](https://github.com/NomicFoundation/hardhat/commit/dee451bcbc2c4f0ca3966e32a33f645a5052e0c1) doesn't provide much context either. I suggest we leave it as-is for now and revisit if we run into issues during testing next week.

v-next/hardhat-viem/src/accounts.ts
schaable: I’ve done it this way to make it easier to differentiate when a type comes from `viem` versus `hardhat-viem`, since all viem types follow the format `viemT.Type`. I don’t believe this impacts performance, as it’s just type annotations, but I think this is the only plugin that uses this approach. I don’t have a strong opinion on keeping it, so I’ll leave the decision to you. If we decide to change it, I’d suggest being consistent and applying it across other files. wdyt?

v-next/hardhat-viem/src/accounts.ts
schaable: I don’t think we can pass it if it’s `private`, right? And I’m not a fan of exposing it as a `public` property. I like the idea of using a `WeakMap` though, since the concern is valid. We should apply it here and in the `chains.ts` file.

kanej: I was thinking something along the lines of:

```ts
const accountsCache = ...

connection.viem = {
  _accountsCache: accountsCache
  getAccounts: () => {
    return getAccounts(provider, accountsCache)
  }
}
```

But I am happy with the WeakMap approach. Lets do that.

v-next/hardhat-viem/src/accounts.ts
kanej: I think this is a nitpick and finely balanced.

My instinct here is to be brutally verbose with something like:

```ts
import type { Address as ViemAddress } from "viem";
```

And I come down on this side, and only just, because it gives us a general rule so that you can treat `import type * as xxx` as suspicious, rather than almost always suspicious but not in these case X,Y,Viem.

I am happy either way on this one - make a call and then feel free to resolve this conversation.

v-next/hardhat-viem/src/chains.ts
```diff
@@ -0,0 +1,139 @@
...
+  chainIdCache.set(provider, chainId);
+
+  return chainId;
+}
+
+export function isDevelopmentNetwork(chainId: number): boolean {
+  return chainId === 31337;
+}
+
+enum NetworkMethod {
```
kanej, line 75: I think we should just use constants here, as part of our war on enums.

v-next/hardhat-viem/src/chains.ts
```diff
@@ -0,0 +1,139 @@
...
+    // If the chain couldn't be found and we can't detect the development
+    // network we throw an error
+    if (chain === undefined) {
+      throw new HardhatError(HardhatError.ERRORS.VIEM.NETWORK_NOT_FOUND, {
+        chainId,
+      });
+    }
+
+    // If the chain is a development network but we couldn't detect which one
+    // we throw an error
```
kanej, line 48:
```suggestion
    // If the chain is a development network but not one of our supported
    // development networks (e.g. Hardhat, Anvil) then throw
```

v-next/hardhat-viem/src/chains.ts
```diff
@@ -0,0 +1,139 @@
...
+    throw new HardhatError(
+      HardhatError.ERRORS.VIEM.UNSUPPORTED_DEVELOPMENT_NETWORK,
+    );
+  }
+
+  return chain;
+}
+
+const chainIdCache = new Map<EthereumProvider, number>();
```
kanej, line 57: All of these caches should become WeakMaps. But what do you think of moving them all to the top of the file, to give a hint to the reader to be on guard?

v-next/hardhat-viem/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+No plugins dependencies.
+
+## Tasks
+
+This plugin creates no additional tasks.
+
+## Environment extensions
+
+This plugins adds a `viem` object to the Hardhat Runtime Environment which provides a minimal set of capabilities for interacting with the blockchain.
```
kanej, line 45: This section needs to be reworked.

I would do it minimally - we are going to review all these readmes. But we shouldn't be factually misleading.

v-next/hardhat-viem/src/accounts.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5825/commits/04b4261526f9da3ba62cd946f6c063bbffbf0681). We should revisit this after the alpha, as the structure is not consistent across plugins.

v-next/hardhat-viem/src/accounts.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5825/commits/45cddb988679b153bc4edadd4af083d85e46a151). We should apply the same changes to other plugins, like `hardhat-ethers`.

v-next/hardhat-viem/src/accounts.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5825/commits/dde02e9c4d406f5413dcd39f7f9d393bd5307a8a).

v-next/hardhat-viem/src/chains.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5825/commits/2abbebd0ab4dfb3026e15631ef412270223dc423).

v-next/hardhat-viem/README.md
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5825/commits/1c81b6cbd79a33d3bde26da3554911710bad5884).

v-next/hardhat-viem/src/internal/clients.ts
```diff
@@ -0,0 +1,162 @@
...
+    ...parameters,
+  });
+
+  if (chainType === "optimism") {
+    publicClient = publicClient.extend(publicActionsL2());
+  }
+
+  /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions --
+  We need to assert the type because TS gets confused with the conditional type */
+  return publicClient as GetPublicClientReturnType<ChainTypeT>;
```
kanej, line 50: I took a brief look here, but have no suggestions.
Do we consider this technical debt or just a limitation in TS? @alcuadrado

v-next/hardhat-viem/src/internal/clients.ts
```diff
@@ -0,0 +1,162 @@
...
+  if (chainType === "optimism") {
+    walletClient = walletClient.extend(walletActionsL2());
+  }
+
+  /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions --
+  We need to assert the type because TS gets confused with the conditional type */
+  return walletClient as GetWalletClientReturnType<ChainTypeT>;
+}
+
+export async function getDefaultWalletClient<
```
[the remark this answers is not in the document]
schaable: Yes, I left a comment about it [here](https://github.com/NomicFoundation/hardhat/pull/5825#discussion_r1806901923), but it was collapsed.

v-next/hardhat-viem/src/internal/clients.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  return getWalletClient(
+    provider,
+    chainType,
+    defaultAccount,
+    walletClientConfig,
+  );
+}
+
+export async function getTestClient(
```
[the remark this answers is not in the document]
schaable: Yes, `connection.viem.getTestClient(...)` returns a test client, so you can extend it as mentioned in the docs. Keep in mind that this has nothing to do with the OP Stack, it's just extending the client with the base public and wallet actions.

v-next/hardhat-viem/src/internal/clients.ts
kanej: It turns our there are limitations in typescript.

kanej (approved):
This looks good.

I left some questions for my own understanding, and some comments:

* passing the parent error
* replacing the README

Take a look at those comments, but I am approving the PR.

v-next/hardhat-viem/src/types.ts
```diff
@@ -0,0 +1,264 @@
...
+   */
+  getContractAt: <ContractName extends string>(
+    contractName: ContractName,
+    address: ViemAddress,
+    getContractAtConfig?: GetContractAtConfig,
+  ) => Promise<ContractReturnType<ContractName>>;
+}
+
+export type GetPublicClientReturnType<ChainTypeT extends ChainType | string> =
+  ChainTypeT extends "optimism" ? OpPublicClient : PublicClient;
```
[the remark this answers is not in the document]
schaable: Correct. However, this would also depend on `viem` adding more chainType-specific actions.

v-next/hardhat-viem/test/contracts.ts
```diff
@@ -0,0 +1,619 @@
...
+      // timeout, the test could hang indefinitely if something goes wrong, such
+      // as blocks not being mined or the contract not being deployed correctly.
+      // This specific timeout helps avoid hitting the much higher global timeout
+      // for tests.
+      // TODO: This test is not working, but I'm unable to debug it because the test
+      // runner hangs after the test is done and does not show the diff. The hanging
+      // issue should be resolved once we're using the hardhat in-memory network.
+      // I suspect the issue with the test itself has to do with the test client
+      // not being able to control the network properly.
+      it.skip("should wait for confirmations", { timeout: 500 }, async () => {
```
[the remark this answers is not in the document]
kanej: Or put another way, I think we tackle this with the test updates for the EDR Provider

v-next/hardhat-viem/src/types.ts
```diff
@@ -0,0 +1,264 @@
...
+   *
+   * @param contractName The name of the contract to deploy. This is required
+   * to return the correct contract type.
+   * @param constructorArgs The arguments to pass to the contract's constructor.
+   * @param deployContractConfig A configuration object. See
+   * {@link DeployContractConfig} for more details.
+   * @returns The deployed contract instance.
+   */
+  deployContract: <ContractName extends string>(
+    contractName: ContractName,
```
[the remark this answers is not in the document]
schaable: Agreed, but I'd call it `contractNameOrFqn` to keep it shorter. I'll add it to the list of small tweaks to do after the alpha, if that's ok.

v-next/hardhat-viem/test/hook-handlers/network.ts
```diff
@@ -0,0 +1,55 @@
...
+
+describe("hook-handlers/network", () => {
+  describe("newConnection", () => {
+    let connection: NetworkConnection<"unknown">;
+
+    before(async () => {
+      const hre = await createHardhatRuntimeEnvironment({
+        plugins: [HardhatViem],
+      });
+      connection = await hre.network.connect();
```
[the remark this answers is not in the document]
schaable: I don't think so, but I'll be even safer once we have the EDR provider, as hardhat is the default network.

v-next/hardhat-viem/test/fixture-projects/default-ts-project/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "default-ts-project",
```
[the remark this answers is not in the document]
schaable: Technically, viem should work with js as well, but I doubt anyone actually uses it that way 😄

v-next/hardhat-viem/src/internal/contracts.ts
```diff
@@ -0,0 +1,233 @@
...
+  const artifact = await artifacts.readArtifact(contractName);
+  let bytecode;
+  try {
+    bytecode = resolveLinkedBytecode(artifact, libraries);
+  } catch (error) {
+    ensureError(error);
+
+    throw new HardhatError(HardhatError.ERRORS.VIEM.LINKING_CONTRACT_ERROR, {
+      contractName,
+      error: error.message,
```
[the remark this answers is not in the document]
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5825/commits/5574552324334597458e117457eb88ebb3bb3cea).

v-next/hardhat-viem/README.md
```diff
@@ -0,0 +1,315 @@
+[![npm](https://img.shields.io/npm/v/@nomicfoundation/hardhat-viem.svg)](https://www.npmjs.com/package/@nomicfoundation/hardhat-viem) [![hardhat](https://hardhat.org/buidler-plugin-badge.svg?1)](https://hardhat.org)
+
+# hardhat-viem
```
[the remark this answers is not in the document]
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/5825/commits/656b37579e27b1b5257e4b3bee612df165b00151).

v-next/hardhat-viem/package.json
alcuadrado: typescript as a peer dependency was an error.

v-next/hardhat-viem/src/internal/chains.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export async function getChain(provider: EthereumProvider): Promise<ViemChain> {
+  const chainId = await getChainId(provider);
+
+  const chain = extractChain({
+    chains,
+    id: chainId,
+  });
+
+  if (isDevelopmentNetwork(chainId) || chain === undefined) {
```
alcuadrado, line 29: I think this is probably too restrictive. I think it's better to do a check like:

```
if (hardhat network) {


}

if (anvil) {

}

// maybe here check if chain id is 1337 or 31337

```

The reason is that when forking people tend to change the chainid

v-next/hardhat-viem/src/internal/contracts.ts
```diff
@@ -0,0 +1,233 @@
...
+  }
+  if (confirmations === 0) {
+    throw new HardhatError(HardhatError.ERRORS.VIEM.INVALID_CONFIRMATIONS, {
+      error:
+        "deployContract does not support 0 confirmations. Use sendDeploymentTransaction if you want to handle the deployment transaction yourself.",
+    });
+  }
+
+  const [publicClient, walletClient, { abi, bytecode }] = await Promise.all([
+    client?.public ?? getPublicClient(provider, "l1"),
```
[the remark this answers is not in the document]
alcuadrado: Maybe it's worth adding a comment, as it's a bit surprising.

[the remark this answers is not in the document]
alcuadrado: Maybe `"unknown"` would be less surprising.

PR #5828, opened 2024-10-16
feat: format the solidity test output better

galargh:

---

This PR builds up on https://github.com/NomicFoundation/hardhat/pull/5809 and introduces custom output formatting described in https://www.notion.so/nomicfoundation/Reporter-Output-Formatting-11f578cdeaf580408d12fb03e7799e81.

It has been tested manually and visually verified.

The the reporter produces for the example project after the changes:
Screenshot 2024-10-16 at 12 11 03

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
@@ -3,8 +3,13 @@ import type {
   TestReporterResult,
   TestStatus,
 } from "./types.js";
+import type { TestResult } from "@ignored/edr";
+
+import { inspect } from "node:util";
 
 import chalk from "chalk";
+
+import { formatArtifactId } from "./formatters.js";
 
 /**
  * This is a solidity test reporter. It is intended to be composed with the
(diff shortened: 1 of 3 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -14,47 +19,119 @@ import chalk from "chalk";
...
       case "suite:result": {
         const { data: suiteResult } = event;
-        for (const testResult of suiteResult.testResults) {
-          testResultCount++;
-          let name = suiteResult.id.name + " | " + testResult.name;
-          if ("runs" in testResult?.kind) {
-            name += ` (${testResult.kind.runs} runs)`;
+        const suiteTestCount = suiteResult.testResults.length;
+
+        if (suiteTestCount === 0) {
```
alcuadrado, line 36: When can this happen? An empty test contract, for example?

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -14,47 +19,119 @@ import chalk from "chalk";
...
+        // NOTE: The test results are in reverse run order, so we reverse them
+        // again to display them in the correct order.
+        for (const testResult of suiteResult.testResults.reverse()) {
+          const name = testResult.name;
           const status: TestStatus = testResult.status;
+          const details = [
+            ["duration", `${testResult.durationMs} ms`],
+            ...Object.entries(testResult.kind),
+          ]
+            .map(([key, value]) => `${key}: ${value}`)
```
alcuadrado, line 62: When `value` is a `Buffer`, we should format it as a 0x-prefixed hex string.

alcuadrado: I think this can also happen later in this code.

alcuadrado (approved):
Left a minor comment, but LGTM

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
galargh: Yes, exactly or an abstract one only.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
galargh: Got it! I'll introduce a utility `bufferToHex` copied from v2 and use it here in a follow-up PR.

galargh: Here is the follow-up PR - https://github.com/NomicFoundation/hardhat/pull/5847

PR #5829, opened 2024-10-16
feat: enhance the solidity test artifacts discovery

galargh:

---

<details><summary>Outdated</summary>
This PR makes the check whether an artifact is a test artifact more accurate. It implements the design described in https://www.notion.so/nomicfoundation/Test-Artifacts-Discovery-11f578cdeaf580bdb7e5c4268c5f248d

The base cases of the newly added `isTestArtifact` method have been tested using unit tests and manually.

It introduces 2 new dependencies:
- `@nomicfoundation/slang` - 2.6M - direct 
- `@bytecodealliance/preview2-shim` - 524K	- transitive
</details>

This PR splits the solidity test runner input preparation into 2 phases: getting the artifacts, and filtering which artifacts are the test artifacts.

It also establishes the test artifact filter as:
1. The source of the artifact has the `.t.sol` extension.
2. The source of the artifact exists within the root of the project.


galargh:
With https://github.com/NomicFoundation/edr/pull/705 in place, we could get away with being less strict when it comes to test contracts discovery. It still could be useful if, let's say, we wanted to ensure we don't spend time compiling files that have `.t.sol` extension but don't define any tests, but I don't think it's that important.

If the reviewers agree, I'd propose simplifying the `isTestArtifact` function by removing the source content checks, which in turn will allow us to remove the dependency on slang.

alcuadrado (approved):
Let's merge this, as it's an improvement, but in a follow up PR we should switch this to be based on files, and not artifacts. This will allow users to run (and compile) a single file.

galargh:
Agreed, that's exactly what I discovered when playing around with compilation for the test task. I'm going to share my thoughts on that later today in a doc.

PR #5835, opened 2024-10-18
Docs: Add JavaScript require statements for upgradeable smart contract via proxy in ignition docs

wesleysmyth:
This includes JS require statements that were missing from the upgradeable smart contract ignition docs that are required for the code to execute properly.

### DIFF
--- a/docs/src/content/ignition/docs/guides/upgradeable-proxies.md
+++ b/docs/src/content/ignition/docs/guides/upgradeable-proxies.md
@@ -227,6 +227,9 @@ module.exports = demoModule;
 Next it's time to upgrade our proxy to a new version. To do this, we'll create a new file within our `ignition/modules` directory called `UpgradeModule.js` (or `UpgradeModule.ts` if you're using TypeScript). Inside this file, we'll again break up our module into two parts. To start, we'll write our `UpgradeModule`:
 
 ```js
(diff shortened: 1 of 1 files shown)


docs/src/content/ignition/docs/guides/upgradeable-proxies.md
```diff
@@ -227,6 +227,9 @@ module.exports = demoModule;
 Next it's time to upgrade our proxy to a new version. To do this, we'll create a new file within our `ignition/modules` directory called `UpgradeModule.js` (or `UpgradeModule.ts` if you're using TypeScript). Inside this file, we'll again break up our module into two parts. To start, we'll write our `UpgradeModule`:
 
 ```js
+const { buildModule } = require("@nomicfoundation/hardhat-ignition/modules");
+const proxyModule = require("./ProxyModule");
```
zoeyTM, line 231: @wesleysmyth it was assumed that a user would be able to infer this much on their own, however being explicit is not necessarily a bad thing. If we add these lines, then we also need a typescript version that uses import instead. You can look elsewhere in this file for a reference on what the formatting should look like.

zoeyTM (changes requested):

zoeyTM:
Closing this as its been inactive for a bit and isn't strictly necessary to be merged. If the comment is addressed later we can reopen and merge.

PR #5811, opened 2024-10-09
feat: add an EDR Provider to the network manager

kanej:
This is a draft PR to allow me to work through the changes required in supporting the EDR Provider and to document them for the design process.

It adds an EDR Provider based on the Hardhat v2 code. The only testing that has been done is manual, based on this test added to the example project:

```ts
describe("Example EDR based test", () => {
  it("should work get the block number from the EDR Network", async () => {
    const connection = await hre.network.connect();

    const blockNumberAtStart = await connection.provider.request({
      method: "eth_blockNumber",
      params: [],
    });

    assert.equal(blockNumberAtStart, `0x0`);

    await connection.networkHelpers.mine(5);

    const blockNumberAfter = await connection.provider.request({
      method: "eth_blockNumber",
      params: [],
    });

    assert.equal(blockNumberAfter, `0x5`);
  });
});
```

To make the above test work, a `hardhat` network was added to the default networks available and the name of the default network was changed from `localhost` to `hardhat`.

## TODO

- [x] Turn isolated modules back on and pass in literals to EDR
- [x] Bring in the new HH-utils version once it is merged: https://github.com/NomicFoundation/hardhat/pull/5811/files#r1797222462
- [ ] bring back Sentry reporting: https://github.com/NomicFoundation/hardhat/pull/5811/files#r1797220231
- [ ] Wire through the tracing config/logging config: https://github.com/NomicFoundation/hardhat/pull/5811/files#r1797262169
- [ ] Define a default `from` value in the EDR Network config, assumedly coming from accounts: https://github.com/NomicFoundation/hardhat/pull/5811/files#r1797263157
- [ ] Actually set proper defaults for the other EDR Network configs, rather than letting co-pilot suggest them: https://github.com/NomicFoundation/hardhat/pull/5811/files#r1797265405
- [ ] Determine if we can remove the property `#vmTracer`: https://github.com/NomicFoundation/hardhat/pull/5811/files#r1797274889
- [ ] Look at some over zealous `any` usage to get out of a type jam: https://github.com/NomicFoundation/hardhat/pull/5811/files#r1797280135
- [ ] Share up logic with the HttpProvider where it makes sense: https://github.com/NomicFoundation/hardhat/pull/5811/files#r1797283389
- [ ] Move compilation result parameter checking to EDR and deal with the exception handling: https://github.com/NomicFoundation/hardhat/pull/5811/files#r1797284569
- [ ] Cleanup, rationalize and move type files for EDR
- [ ] Tests, actual tests

PR #5840, opened 2024-10-22
feat: add Optimism provider

Wodann:
A draft PR for adding Optimism support in EDR using the `@ignored/edr-optimism` package, intended for the alpha.

### DIFF
--- a/packages/hardhat-core/package.json
+++ b/packages/hardhat-core/package.json
@@ -102,7 +102,7 @@
   "dependencies": {
     "@ethersproject/abi": "^5.1.2",
     "@metamask/eth-sig-util": "^4.0.0",
-    "@nomicfoundation/edr": "^0.6.4",
+    "@ignored/edr-optimism": "^0.6.4",
     "@nomicfoundation/ethereumjs-common": "4.0.4",
     "@nomicfoundation/ethereumjs-tx": "5.0.4",
     "@nomicfoundation/ethereumjs-util": "9.0.4",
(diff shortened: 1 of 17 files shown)


schaable:
@Wodann I've [merged](https://github.com/NomicFoundation/hardhat/pull/5811/commits/79de6904dbafb8202111acdf3b81bddadf66d0ef) this PR into the `feat/edr-provider` branch, so we can close this. There are a couple of enums, like `StackTraceEntryType` and `MineOrdering`, that still need to be converted to constants. I've casted the values for now to get things working, but we should revisit this in the future to clean it up.

Wodann:
> @Wodann I've [merged](https://github.com/NomicFoundation/hardhat/pull/5811/commits/79de6904dbafb8202111acdf3b81bddadf66d0ef) this PR into the `feat/edr-provider` branch, so we can close this. There are a couple of enums, like `StackTraceEntryType` and `MineOrdering`, that still need to be converted to constants. I've casted the values for now to get things working, but we should revisit this in the future to clean it up.

Are enums prohibited?

Could you provide an example of what you'd like to be generated instead of an enum?

Wodann:
> The issue is that we have [`isolatedModules`](https://www.typescriptlang.org/tsconfig/#isolatedModules) set to true to ensure compatibility with single-file transpilation tools like `tsx`. With `isolatedModules`, `const enum`s can’t be inlined, since this setting assumes each file compiles independently rather than together.
> 
> Instead of using enums, it would be better to generate constants, similar to what was done for `SpecId`.

Thanks for explaining! I'll add a task to do that and add you as a reviewer.

Wodann:
[Issue added](https://github.com/NomicFoundation/edr/issues/721)

PR #5841, opened 2024-10-22
chore: explicitly opt-in to node test events

galargh:

---

This is a follow-up to the discussion that emerged under https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1729529455497859

### DIFF
--- a/v-next/hardhat-node-test-reporter/src/reporter.ts
+++ b/v-next/hardhat-node-test-reporter/src/reporter.ts
@@ -244,45 +244,15 @@ export function hardhatTestReporter(
           yield event.data.message;
           break;
         }
-        case "test:plan": {
-          // Do nothing
-          break;
-        }
-        case "test:enqueue": {
-          // Do nothing
-          break;
-        }
-        case "test:dequeue": {
-          // Do nothing
-          break;
-        }
-        case "test:watch:drained": {
-          // Do nothing
-          break;
-        }
-        case "test:complete": {
-          // Do nothing
-          break;
-        }
         case "test:coverage": {
           yield chalk.red("\nTest coverage not supported by this reporter\n");
           break;
         }
-        // @ts-expect-error -- A new event type was added in https://github.com/nodejs/node/pull/54851
-        // but @types/node doesn't know about it yet. Once it does, the dependency should be updated
-        // and this comment removed.
-        case "test:summary": {
-          // Do nothing
-          break;
-        }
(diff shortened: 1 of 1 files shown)


v-next/hardhat-node-test-reporter/src/reporter.ts
```diff
@@ -244,45 +244,15 @@ export function hardhatTestReporter(
...
-        // but @types/node doesn't know about it yet. Once it does, the dependency should be updated
-        // and this comment removed.
-        case "test:summary": {
-          // Do nothing
-          break;
-        }
-        /* eslint-disable-next-line @typescript-eslint/switch-exhaustiveness-check --
-        We have this extra check here becase we know the @types/node type is
-        unreliable */
         default: {
```
kanej, line 251: This looks good.
I was tempted to say we should keep the `do nothing` tests so we can explicitly see what events we are skipping (and consider whether we should add the actions for them).
But I think the complexity cost of keeping `@ts-expect-error` annoys me more than a theoretical future readability payoff.
What do you think?

galargh: That was my thinking exactly. Also, when we want to reconsider what events we use, we should look at node directly anyway since @types/node, sadly, might be lagging behind. Considering all that, a more concise switch statement might be a better choice here.

kanej: Agreed.

kanej (approved):

v-next/hardhat-node-test-reporter/src/reporter.ts
```diff
@@ -244,45 +244,15 @@ export function hardhatTestReporter(
...
-          const _isNever: never = event;
-          void _isNever;
-
-          yield chalk.red(`Unsuported node:test event:`, event);
+          // NOTE: Do nothing; we explicitly opt-in to the events we want to
```
alcuadrado, line 252: Maybe we can log it with `debug`

galargh: I think if we do that we might run into some ordering edge cases as the reporter doesn't really log anything at the moment; it's just a string generator that is later piped through to the stdout.

I do agree with the sentiment though, it could be nice to be able to enable extra logging of unknown events.

alcuadrado: Good point. Let's play safe.

PR #5846, opened 2024-10-24
fix: pass HTTP to EDR

### LINKED ISSUE #5845: Setting custom httpHeaders does not work after EDR introduction
### Version of Hardhat

2.22.13

### What happened?

Custom authorisation headers no longer work when forking node.

Latest working version is 2.20.1 so I suppose EDR is the one to blame.

Follow the guide about how to provide custom HTTP headers: https://github.com/NomicFoundation/hardhat/blob/2f5b7a8be71f2183c2992383bcc582bc4c02626b/docs/src/content/hardhat-network/docs/guides/forking-other-networks.md?plain=1#L139

### Search terms

httpHeaders EDR

Wodann:
Fixes #5845 

A test validating the fix was added in https://github.com/NomicFoundation/edr/pull/710

### DIFF
--- a/.changeset/neat-moose-compare.md
+++ b/.changeset/neat-moose-compare.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed broken HTTP headers for forked configurations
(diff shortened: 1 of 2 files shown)


kanej (approved):
LGTM.

kanej:
@schaable we should take this addition across in the EDR Provider.

PR #5847, opened 2024-10-24
chore: format buffers in solidity test output as hex strings

galargh:

---

This is a follow-up to https://github.com/NomicFoundation/hardhat/pull/5828#discussion_r1813556874

### DIFF
--- a/v-next/hardhat-utils/src/hex.ts
+++ b/v-next/hardhat-utils/src/hex.ts
@@ -95,17 +95,7 @@ export function hexStringToNumber(hexString: string): number {
  * @returns PrefixedHexString The hexadecimal representation of the bytes.
  */
 export function bytesToHexString(bytes: Uint8Array): PrefixedHexString {
-  return bufferToHexString(Buffer.from(bytes));
-}
-
-/**
- * Converts a Buffer to a hexadecimal string.
- *
- * @param buffer The buffer to convert.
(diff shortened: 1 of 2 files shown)


v-next/hardhat-utils/src/hex.ts
```diff
@@ -95,7 +95,17 @@ export function hexStringToNumber(hexString: string): number {
...
+  return bufferToHexString(Buffer.from(bytes));
+}
+
+/**
+ * Converts a Buffer to a hexadecimal string.
+ *
+ * @param buffer The buffer to convert.
+ * @returns PrefixedHexString The hexadecimal representation of the buffer.
+ */
+export function bufferToHexString(buffer: Buffer): PrefixedHexString {
```
alcuadrado, line 107: I think we don't need this. It may not be entirely obvious, but `Buffer` extends `Uint8Array`, so you can just use `bytesToHexString`.

`hardhat-utils` intentionally tries not to have Node types in its interface.

alcuadrado (approved):
Apart from the comment I left, this LGTM

v-next/hardhat-utils/src/hex.ts
galargh: Ah, OK, I understand now. Thanks, I'll revert that part of the PR and use bytesToHexString instead.

PR #5859, opened 2024-10-26
Rename the `"unknown"` `ChainType` to `"generic"` to be aligned with EDR

alcuadrado:


schaable (approved):
Looks good. Another small improvement we could do here is to make the `NetworkConnection` type parameter default to `GenericChainType` when none is passed, so we can do `let connection: NetworkConnection;` instead of `let connection: NetworkConnection<GenericChainType>;`

PR #5862, opened 2024-10-27
Unified test workflow and small improvements

alcuadrado:
This is a PR with some small improvements to have a first iteration of a unified test workflow for the alpha. This will be improved later.

It does a few things:

1. Set proper names to the `node:test` and `mocha` plugins.
2. Introduces a `builtin:test` plugin, which defines a `test` task, which runs all of its subtasks.
3. Updates the `mocha`, and `node:test` plugins so that they define subtasks — This allows them to coexist 🥳 /cc @kanej 
4. Sets the solidity test runner as a subtask.
5. Fix a bug in the solidity test runner that made the task return before the output was printed.
6. Sets all the required `dependencies:` in the builtin plugins.

Unrelated improvements:

1. Fix `pnpm run clean` in the example project.
2. Fix a viem example in the example project.
3. Implement `noCompile` everywhere it was needed (mostly replacing TODOs)
4. Implemented a TODO in the `console` task. 
5. Make the build-system print warnings and errors in quiet mode.

I also added a test in the example project which uses `forge-std`.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
@@ -8,7 +8,7 @@ interface TestActionArguments {
   noCompile: boolean;
 }
 
-const runScriptWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
+const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
   { noCompile },
   hre,
 ) => {
@@ -31,4 +31,4 @@ const runScriptWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
   }
 };
 
-export default runScriptWithHardhat;
+export default runAllTests;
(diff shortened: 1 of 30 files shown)


v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import chalk from "chalk";
+
+import { formatTaskId } from "../../core/tasks/utils.js";
+
+interface TestActionArguments {
+  noCompile: boolean;
+}
+
+const runScriptWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
```
galargh, line 11:
```suggestion
const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
```

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -0,0 +1,34 @@
...
+    await subtask.run({ noCompile: true });
+    console.log();
+  }
+
+  if (process.exitCode !== 0) {
+    console.error("Test run failed");
+  }
+};
+
+export default runScriptWithHardhat;
```
galargh, line 34:
```suggestion
export default runAllTests;
```

galargh (approved):

PR #5864, opened 2024-10-27
Fix automatic sender hook handler

alcuadrado:
This PR fixes the automatic sender hook handler, as it didn't support the case where no account was present.

@ChristopherDedominici can you take a look?

Also, some hook handlers throw errors where they should return JSON-RPC response with `InternalError`s.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/automatic-sender-provider.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/automatic-sender-provider.ts
@@ -8,20 +8,23 @@ import { Sender } from "./sender.js";
  * ensuring dynamic selection of the sender for all JSON-RPC requests without requiring manual input.
  */
 export class AutomaticSender extends Sender {
+  #alreadyFetchedAccounts = false;
   #firstAccount: string | undefined;
 
-  protected async getSender(): Promise<string> {
-    if (this.#firstAccount === undefined) {
+  protected async getSender(): Promise<string | undefined> {
+    if (this.#alreadyFetchedAccounts === false) {
       const accounts = await this.provider.request({
         method: "eth_accounts",
       });
 
+      // TODO: This shouldn't be an exception but a failed JSON response!
       assertHardhatInvariant(
-        Array.isArray(accounts) && typeof accounts[0] === "string",
-        "accounts should be an array and accounts[0] should be a string",
+        Array.isArray(accounts),
+        "eth_accounts response should be an array",
       );
 
       this.#firstAccount = accounts[0];
+      this.#alreadyFetchedAccounts = true;
     }
 
     return this.#firstAccount;
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/fixed-sender-provider.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/fixed-sender-provider.ts
@@ -15,7 +15,7 @@ export class FixedSender extends Sender {
     this.#sender = sender;
   }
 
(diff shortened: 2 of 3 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/automatic-sender-provider.ts
```diff
@@ -8,20 +8,23 @@ import { Sender } from "./sender.js";
  * ensuring dynamic selection of the sender for all JSON-RPC requests without requiring manual input.
  */
 export class AutomaticSender extends Sender {
+  #alreadyFetchedAccounts = false;
```
alcuadrado, line 11: Note that as `#firstAccount` can be undefined, I added a second flag to avoid re-requesting.

v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/automatic-sender-provider.ts
```diff
@@ -8,20 +8,23 @@ import { Sender } from "./sender.js";
...
 
-  protected async getSender(): Promise<string> {
-    if (this.#firstAccount === undefined) {
+  protected async getSender(): Promise<string | undefined> {
+    if (this.#alreadyFetchedAccounts === false) {
       const accounts = await this.provider.request({
         method: "eth_accounts",
       });
 
+      // TODO: This shouldn't be an exception but a failed JSON response!
```
alcuadrado, line 20: This drove me crazy tbh. Throwing here makes things much more complex, as viem gets confused.

jxom: Is this an issue on Viem's side?

v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/automatic-sender-provider.ts
ChristopherDedominici: We could add a test here, I can do it if you want because I exactly know where to add it. It will only take me 5 minutes

v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/automatic-sender-provider.ts
ChristopherDedominici: I'm confused here: is this TODO for later or it should be implemented in this PR? If it's for later, is there an issue tracking it?

alcuadrado: @ChristopherDedominici: later

@jxom, I don't think this is an issue. It's just that we used to have a bug where this code entered an infinite recursion, and viem was doing many retries, so I spent a long time thinking that the other bug appeared again 😅

Is there a way to throw from a provider so that viem just bubbles/ignores the error?

v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/automatic-sender-provider.ts
alcuadrado: Sure, but in a follow up PR

v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/automatic-sender-provider.ts
jxom: You can pass `retryCount: 0` as second param

PR #5865, opened 2024-10-27
Small improvements to the config resolution of EDR networks

alcuadrado:
This is a small PR to that I put in place to be able to enable forking in EDR networks. It is not complete, though.

It does just a few things:

1. Changes the EDR network config resolution so that `chainType` can be left `undefined`, as that's important to be able to change them during connection.
2. Resolved the `chainType` during connection.
3. Added the `forkConfig` to the resolved network config, without any actual resolution.
4. Sparkled some TODOs.

The reason this PR is important is that optimism needs to be tested in forked mode.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
@@ -143,8 +143,26 @@ export class NetworkManagerImplementation {
       );
 
       if (resolvedNetworkConfig.type === "edr") {
+        if (
+          resolvedChainType !== "optimism" &&
+          resolvedChainType !== "generic" &&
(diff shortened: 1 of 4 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -143,8 +143,26 @@ export class NetworkManagerImplementation {
       );
 
       if (resolvedNetworkConfig.type === "edr") {
+        if (
+          resolvedChainType !== "optimism" &&
```
schaable, line 147: Not to be resolved in this PR, but I'm starting to think that all these keywords need to be constants.

schaable (approved):

PR #5866, opened 2024-10-27
Run `v-next`'s CI on chained PRs

alcuadrado:
As we move towards polishing the next version, we are probably going to be creating a ton of chained/dependent PRs.

Currently, `v-next-ci` doesn't run for those, leading to undetected errors until you merge some PRs.

In this PR I changed how we filter which PRs to run this workflow on. The workflow always runs, but I first check if the `./v-next` folder exists. If it doesn't exist, I skip the main jobs, leading to all the jobs being skipped.

### DIFF
--- a/.github/workflows/v-next-ci.yml
+++ b/.github/workflows/v-next-ci.yml
@@ -6,9 +6,6 @@ on:
     branches:
       - v-next
   pull_request:
-    # TODO: Remove the filter when v-next is merged into main
-    branches:
-      - v-next
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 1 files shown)


alcuadrado:
@galargh I'm merging this to unblock all my other PRs, but I'd appreciate any comments/ideas.

.github/workflows/v-next-ci.yml
```diff
@@ -6,18 +6,35 @@ on:
     branches:
       - v-next
   pull_request:
-    # TODO: Remove the filter when v-next is merged into main
-    branches:
-      - v-next
```
galargh, line 11: We could add a `paths` filter here. It wouldn't be perfect but running CI whenever `v-next` folder is modified should catch the vast majority of the changes.

```
    paths:
      - .github/workflows/v-next-ci.yml
      - v-next/**
```

galargh (commented):
Post-merge LGTM 👍 One suggestion if we wanted to get rid of the extra job and the if statements.

PR #5868, opened 2024-10-28
Move remaining v2 providers (LocalAccounts and HDWallet)

ChristopherDedominici:
Link to [design doc](https://www.notion.so/nomicfoundation/Port-missing-V2-providers-121578cdeaf58043802dd5b52233bb91)

Changes in this PR:
- add the missing accounts provider from V2:
  - LocalAccounts 
  - HDWallet
- use zod for RPC type validation
- replace V2 npm packges (e.g.: `@nomicfoundation/ethereumjs-tx`) with `micro-eth-signer`

NOTE:
HdWallet extends LocalAccounts. 
LocalAccounts is a bit different compared to the other V2 providers. The other V2 providers only modify the RPC requests, but under certain specific circumstances, this one directly returns a response to the request.
To handle this scenario LocalAccounts exposes 2 methods:
- [resolveRequest](https://github.com/NomicFoundation/hardhat/pull/5868/files#r1822995090): resolve request if need, returns null otherwise
- [modifyRequest](https://github.com/NomicFoundation/hardhat/pull/5868/files#r1822996112): modify the request as the other V2 providers
This also impact the hook handler: now if the request is directly resolved by LocalAccounts or HDWallet, the `next` function is not called, but the response is returned directly. See [here](https://github.com/NomicFoundation/hardhat/pull/5868/files#diff-135b8c817cb995a1e4b06fd45da884a3d7730ed0f86ae2d740712f70031dfece)

TODOs after alpha:
- double check the new zod typs


v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/local-accounts.ts
```diff
@@ -0,0 +1,328 @@
...
+  constructor(
+    provider: EthereumProvider,
+    localAccountsHexPrivateKeys: string[],
+  ) {
+    super(provider);
+
+    this.#initializePrivateKeys(localAccountsHexPrivateKeys);
+  }
+
+  public async resolveRequest(
```
ChristopherDedominici, line 42: Directly resolve the request

v-next/hardhat/src/internal/builtin-plugins/network-manager/json-rpc-request-modifiers/accounts/local-accounts.ts
```diff
@@ -0,0 +1,328 @@
...
+          jsonRpcRequest.id,
+          signTyped(typedMessage, privateKey),
+        );
+      }
+    }
+
+    return null;
+  }
+
+  public async modifyRequest(jsonRpcRequest: JsonRpcRequest): Promise<void> {
```
ChristopherDedominici, line 129: modify request like all the other V2 providers

PR #5869, opened 2024-10-28
update the v-next packates in @ignore

kanej:
This is a release of the latest `v-next` packages to our `@ignored` namespace to support testing.


.changeset/curly-wombats-wink.md
```diff
@@ -2,7 +2,6 @@
 "@ignored/hardhat-vnext-errors": patch
 "@ignored/hardhat-vnext-utils": patch
 "@ignored/hardhat-vnext": patch
-"@ignored/hardhat-vnext-build-system": major
```
kanej, line 5: We no longer have this package, so I am removing it from changeset config and history.

.changeset/mean-chefs-learn.md
```diff
@@ -0,0 +1,15 @@
+---
+"@ignored/hardhat-vnext-viem": major
```
kanej, line 2: `hardhat-viem` is the only new package, everything else has been released before.

PR #5870, opened 2024-10-28
feat: define solidity test sources in config

galargh:

---

This PR:
- adds a `solidity` key to the `paths.tests` user config (defaults to `paths.sources.solidity` value)
- removes the dependency on the `compile` task from the `test:solidity` task
- adds compilation of `paths.tests.solidity` to the `test:solidity` task

To compile a subset of sources for solidity tests, I use the build function from the build system programmatically (and prepare required inputs in the solidity test task action).

To make the artifact discovery work properly, I had to add `contractArtifactsGenerated` property to the `CacheHitFileBuildResult`. This is because given a compilation job, we need to be able to tell what artifacts were generated, even if that job was cached.

One alternative/extension considered was to define the solidity test paths per build profile. This would have allowed greater level of configuration. We decided not to proceed with that, at least for now.


v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -19,25 +31,46 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
...
-  const testSuiteIds = (
+  const rootFilePaths = (
     await Promise.all(
-      artifacts.map(async (artifact) => {
-        if (await isTestArtifact(hre.config.paths.root, artifact)) {
-          return artifact.id;
-        }
-      }),
+      hre.config.paths.tests.solidity
+        .map((p) => resolveFromRoot(hre.config.paths.root, p))
```
alcuadrado, line 37: This would change based on the other comments

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -19,25 +31,46 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
...
+        }),
     )
-  ).filter((artifact) => artifact !== undefined);
+  ).flat(1);
+
+  const buildOptions: BuildOptions = {
+    // NOTE: The uncached sources will still be compiled event if `noCompile`
+    // is true. We could consider adding a `cacheOnly` option to support true
+    // `noCompile` behavior.
+    force: !noCompile,
```
alcuadrado, line 56: If I run `npx hardhat test solidity` it will force and recompile everything

alcuadrado: Maybe add an issue and address it after we have the compilation cache in place, and when we have the split between production and test compilation caches or while implementing that.

alcuadrado: We should also consider removing this option

v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
```diff
@@ -1,66 +1,119 @@
...
-      name: hardhatArtifact.contractName,
-      solcVersion: buildInfo.solcVersion,
-      source: hardhatArtifact.sourceName,
-    };
+/**
+ * This function asserts that the given Solidity build results are successful.
+ * It throws a HardhatError if the build results indicate that the compilation
+ * job failed.
+ */
+export function throwIfSolidityBuildFailed(
```
alcuadrado, line 35: For a future PR: we should use the same implementation here and during `compile`.

alcuadrado: mostly to have a consistent ui

v-next/hardhat/test/internal/core/hook-manager.ts
```diff
@@ -257,8 +257,9 @@ describe("HookManager", () => {
                       _context: HookContext,
                       givenHre: HardhatRuntimeEnvironment,
                     ): Promise<void> => {
-                      givenHre.config.paths.tests =
-                        "./test-folder-from-plugin1";
+                      givenHre.config.paths.tests = {
```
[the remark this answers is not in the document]
galargh: I remember now! `givenHre.config.paths` is of type `TestPathsConfig`, which is an empty interface that is extended by the `solidity-test` plugin with `solidity` key. We shouldn't have been assigning a string to it, but my guess is it worked because the interface wasn't being extended in this package before.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -78,7 +78,10 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     const localFilesToCompile = (
       await Promise.all(
         this.#options.soliditySourcesPaths.map((dir) =>
-          getAllFilesMatching(dir, (f) => f.endsWith(".sol")),
+          getAllFilesMatching(
+            dir,
+            (f) => f.endsWith(".sol") && !f.endsWith(".t.sol"),
+          ),
```
galargh, line 84: I'm saying here that we should **NOT** compile `.t.sol` files present in solidity sources in the default compile task.

v-next/hardhat/test/internal/core/hook-manager.ts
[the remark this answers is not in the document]
galargh: `givenHre.config.paths.tests` is of type `TestPathsConfig` which is an empty interface to begin with. In this PR, it gets extended by a built-in `solidity-test` plugin with a `solidity` key. After this, ts starts flagging this assignment as an issue so we can either use something it knows about here or ignore the error.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -1,43 +1,68 @@
...
 interface TestActionArguments {
   timeout: number;
-  noCompile: boolean;
+  force: boolean;
+  quiet: boolean;
 }
 
 const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
-  { timeout, noCompile },
+  { timeout, force, quiet },
```
galargh, line 28: I decided to remove the `noCompile` flag from here until we figure out the right semantics for it. 

I also added `force` and `quiet` flags as they are present in the compile task and make sense here as well.

galargh (commented):
As discussed in a call and on Slack, I made the semantics of `paths.tests.solidity` equal to that of `mocha` and `nodeTest`.

I also removed the `noCompile` flag from the solidity test task, moved the common helpers from the test plugin to solidity one, and, most importantly, changed the way in which the tests are discovered (`.t.sol` from solidity sources and all `.sol` files from solidity test sources).

v-next/hardhat/src/internal/builtin-plugins/solidity-test/index.ts
```diff
@@ -16,8 +21,12 @@ const hardhatPlugin: HardhatPlugin = {
...
       .addFlag({
-        name: "noCompile",
-        description: "Don't compile the project before running the tests",
+        name: "force",
+        description: "Force compilation even if no files have changed",
+      })
+      .addFlag({
+        name: "quiet",
+        description: "Makes the compilation process less verbose",
       })
```
alcuadrado, line 30: Agreed with removing `noCompile`.

Now, I'm not sure if these new flags are valuable tbh. And with this names they are definitely confusing, as they are flags to the compilation process, but belong to a test task. I think we should do the same and remove them, at least for now.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -1,43 +1,68 @@
...
+    ])
+  ).flat(1);
+
+  const buildOptions: BuildOptions = {
+    force,
+    buildProfile: hre.globalOptions.buildProfile,
+    mergeCompilationJobs: shouldMergeCompilationJobs(
+      hre.globalOptions.buildProfile,
+    ),
+    quiet,
```
alcuadrado, line 51: quiet should always be true

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -1,43 +1,68 @@
...
+      ...hre.config.paths.sources.solidity.map(async (dir) => {
+        return getAllFilesMatching(dir, (f) => f.endsWith(".t.sol"));
       }),
-    )
-  ).filter((artifact) => artifact !== undefined);
+    ])
+  ).flat(1);
+
+  const buildOptions: BuildOptions = {
+    force,
```
alcuadrado, line 46: force should always be force

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -1,43 +1,68 @@
...
+      hre.globalOptions.buildProfile,
+    ),
+    quiet,
+  };
 
-  if (testSuiteIds.length === 0) {
-    return;
-  }
+  // NOTE: We compile all the sources together with the tests
+  const rootFilePaths = [...rootSourceFilePaths, ...rootTestFilePaths];
```
alcuadrado, line 55: No need to compile non-test files, the test files include any necessary file as a dependency.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: looks good

v-next/hardhat/test/internal/core/hook-manager.ts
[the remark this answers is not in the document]
alcuadrado: Oh, I see! the only reason this issue was hit before is because the other plugins that extend `TestPathsConfig` are non-builtin.

alcuadrado (approved):
Looking good. I left two small comments. Feel about the root files and the task's flags. Feel free to merge after approaching them.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/index.ts
galargh: OK, no worries. I'll remove them for now.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
galargh: I would have sworn that didn't work for me before, but I cannot replicate what I vaguely remember now so I'll remove the rootSourceFilePaths compilation.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/index.ts
fvictorio: I think removing `noCompile` broke the main `test` task, which always sets that param to true for all test subtasks.

galargh: Ay! You're right! Sorry about that. I'm on it!

PR #5877, opened 2024-10-28
Change how we print the titles of the test tasks

alcuadrado:


kanej (approved):
Looks good.
We should revisit some of the spacing in the output, but that is for later.

PR #5880, opened 2024-10-29
update templates to include basic solidity tests and add a viem template

kanej:
Update the templates for the Hardhat v3 Alpha:

1. Enhances the Mocha/Ethers template by adding:

* simple Solidity test contracts from the example project
* an example Solidity test using `forge-std`
* a script deploying a simple contract to a local optimism Hardhat
  network

2. Adds a Node Test Runner/Viem Template with:

* brings across the lock contract taken from v2
* the viem tests for lock taken from v2 and adapted
* simple Solidity test contracts from the example project
* an example Solidity test using `forge-std`
* a script deploying a simple contract to a local optimism Hardhat
  network

3. Removes all other templates (they will only distract in the Alpha)


v-next/hardhat/templates/mocha-ethers/hardhat.config.ts
```diff
@@ -7,6 +7,17 @@ import HardhatNetworkHelpers from "@ignored/hardhat-vnext-network-helpers";
 const config: HardhatUserConfig = {
   plugins: [HardhatMochaTestRunner, HardhatEthers, HardhatNetworkHelpers],
   solidity: "0.8.24",
+  defaultNetwork: "hardhat",
+  networks: {
+    "local-base": {
+      chainId: 8453,
+      type: "edr",
+      chainType: "optimism",
+      gas: "auto",
```
kanej, line 16: This seems required for the moment, which is unexpected. We should have defaults here.

alcuadrado (approved):

PR #5887, opened 2024-10-29
Prevent a shell injection in `--init`

alcuadrado:


galargh (approved):
Good call, especially if at some point we want to add support for remotely hosted templates.

PR #5888, opened 2024-10-29
Fix the EDR Network user config validation

alcuadrado:
Just a small PR fixing a zod schema that was too restrictive.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
@@ -193,12 +193,12 @@ const edrNetworkUserConfigAccountsSchema = conditionalUnionType(
 
 const edrNetworkUserConfigSchema = z.object({
   type: z.literal("edr"),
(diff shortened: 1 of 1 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
```diff
@@ -193,12 +193,12 @@ const edrNetworkUserConfigAccountsSchema = conditionalUnionType(
...
 const edrNetworkUserConfigSchema = z.object({
   type: z.literal("edr"),
-  chainId: z.number().int(),
+  chainId: z.number().int().optional(),
   chainType: z.optional(chainTypeSchema),
-  from: z.optional(z.string()),
-  gas: userGasSchema,
-  gasMultiplier: z.number(),
-  gasPrice: userGasSchema,
+  from: z.optional(z.string()).optional(),
```
schaable, line 198: this should be redundant

schaable (approved):
Just a small comment, but lgtm!

PR #5892, opened 2024-10-29
Template improvements

alcuadrado:
This PR includes a first round of improvements to the template projects, it's still WIP, but can be merged as is.


v-next/hardhat/templates/node-test-runner-viem/test/Lock.ts
```diff
@@ -1,11 +1,11 @@
...
 import assert from "node:assert/strict";
-import hre from "@ignored/hardhat-vnext";
+import { network } from "@ignored/hardhat-vnext";
 import { getAddress, parseGwei } from "viem";
-import { NetworkConnection } from "@ignored/hardhat-vnext/types/network";
 
-describe("Lock", function () {
-  let networkConnection: NetworkConnection<"l1">;
+describe("Lock", async function () {
+  const { viem, networkHelpers } = await network.connect();
```
alcuadrado, line 7: What do you think about this async operation in the describe? @kanej @galargh 

It makes the tests way simpler.

v-next/hardhat/templates/mocha-ethers/test/Lock.ts
```diff
@@ -1,16 +1,19 @@
...
-import chaiAsPromised from "chai-as-promised";
-chai.use(chaiAsPromised);
-const { expect } = chai;
+import "./setup.js";
+
+import { HardhatEthers } from "@ignored/hardhat-vnext-ethers/types";
 
 describe("Lock", function () {
-  let networkConnection: NetworkConnection<"l1">;
+  let networkHelpers: any; // TODO: We need to export this type in @ignored/hardhat-vnext-network-helpers
```
alcuadrado, line 9: We need to export the `NetworkHelpers` as an interface in the plugin. It's now a class, but it should be a quick conversation into an interface. Slightly more involved than just copy and replace, because it has some nested objects.

v-next/hardhat/templates/node-test-runner-viem/test/Lock.ts
kanej: I think it cleans up the code. But I do find it surprising - I expect resource initialization code to be in the before/beforeEach.

I was initially concerned that this code might get run even if the describe is skipped, but I can confirm that isn't the case.

Would we use a similar pattern in the `mocha` template?

kanej (approved):
I have left a comment on async usage to avoid an explicit `NetworkConnection`, but however that gets resolved the other changes make sense.

v-next/hardhat/templates/node-test-runner-viem/test/Lock.ts
alcuadrado: `mocha` intentionally doesn't support that. So we can't do it there.

Maybe we can initialize the network connection on the first provider call? I think the code style improvement is massive.

alcuadrado: In reality, given that this is ESM, these calls can be in the top-level of the file.

PR #5893, opened 2024-10-29
Use the `DefaultChainType` as default type parameter

alcuadrado:
WDYT, @schaable ?


schaable (approved):
LGTM! Next step should be to update the code to remove the type parameter when using `NetworkConnection`. But I don't think it's a priority.

PR #5899, opened 2024-10-30
Fix some race conditions when lazy loading things in the HRE hook handlers

alcuadrado:


kanej (approved):
I have approved but added a follow on task to add in an integration test: https://github.com/NomicFoundation/hardhat/issues/5902

PR #5916, opened 2024-10-31
Fix hardhat-viem so that it works with EDR simulating Op

alcuadrado:


alcuadrado:
Reviewed it with @schaable on a call

PR #5922, opened 2024-11-01
chore: add optimism example script to viem template

kanej:
Pull across the example script showing sending a tx against Optimism into the viem template.

### DIFF
--- a/v-next/hardhat/templates/node-test-runner-viem/README.md
+++ b/v-next/hardhat/templates/node-test-runner-viem/README.md
@@ -27,8 +27,37 @@ npx hardhat3 test
 
 ### Multi-chain support
 
-To deploy a contract to an in-memory Hardhat node simulating Base (an Optimism l2 chain), run:
+Hardhat network now supports simulating Optimism L2 chains. You can see an example of a local Hardhat network configured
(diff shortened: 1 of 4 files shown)


v-next/hardhat/templates/node-test-runner-viem/README.md
```diff
@@ -27,8 +27,37 @@ npx hardhat3 test
...
+
+A script demonstrating the sending of an L2 transaction is included as `./scripts/send-op-tx.ts`, to run it against the local `edrOp` Optimism network, run:
+
+```shell
+npx hardhat3 run scripts/send-op-tx.ts --network edrOp
+```
+
+To run the same script against the Optimism mainnet, first set the `OPTIMISM_PRIVATE_KEY` config variable, with the private key of the account you want to use to send the transaction:
+
+> WARNING: the private key is stored unencrypted to file. Full encryption will be included in a future release.
```
kanej, line 52: Would it be better to rely on `.env` variables here, so that at least we are being explicit?

alcuadrado: We discussed this, and added a big warning in `keystore set`

Issue #3430, opened 2022-12-15
Add `--output` parameter to `flatten` task


fvictorio:
You can use `>` to create a file with the output of the `flatten` task, but this is not discoverable for people that are not very familiar with the shell. Plus, AFAIK that doesn't work on Windows's cmd.

We should add a `--output` parameter with a path to the file that will be created.

```
hh flatten --output Flattened.sol
```

If the file exists, it should be overriden.

kmjones1979:
Any update on this feature request? I stumbled across this problem today when using yarn and trying to use hardhat flatten. I think this would be a really useful feature!

s-di-cola:
@fvictorio I have created a PR for this. This is my first PR for this project, therefore any improvement is greatly appreciated :)

PR #5947, opened 2024-11-15
feat: adds --output to flatten

s-di-cola:
---
Implements --output parameter for flatten task as requested in #3430. This allows users to specify an output file path for the flattened contract, improving usability especially on Windows systems where shell redirection may not work consistently.


packages/hardhat-core/test/builtin-tasks/flatten.ts
```diff
@@ -412,6 +415,18 @@ describe("Flatten task", () => {
...
+    it("should write to an output file when the parameter output is specified", async function () {
+      const outputFile = "flatten.sol";
+      await this.env.run(TASK_FLATTEN, {
+        files: ["contracts/A.sol", "contracts/D.sol"],
+        output: outputFile,
+      });
+      const expected = await getExpectedSol();
+      const actual = readFileSync(outputFile, "utf8");
+      assert.equal(actual, expected);
+      removeSync(outputFile);
```
fvictorio, line 427: Instead of creating the file in the cwd and then removing it, I think it's slightly better to create it in a temporary dir. You can do it like this:

```typescript
const outputFile = path.resolve(os.tmpdir(), "flattened.sol");
```

(Give or take; I haven't tried it)

fvictorio:
This looks good, but I'll let someone from the Hardhat team actually review it in depth and approve it. This also needs a changeset, of course.

packages/hardhat-core/test/builtin-tasks/flatten.ts
s-di-cola: Good point :) 

Addressed with `const outputFile = `${tmpdir()}/flatten.sol`;`

packages/hardhat-core/test/builtin-tasks/flatten.ts
```diff
@@ -412,6 +414,21 @@ describe("Flatten task", () => {
...
+      try {
+        await this.env.run(TASK_FLATTEN, {
+          files: ["contracts/A.sol", "contracts/D.sol"],
+          output: outputFile,
+        });
+        const expected = await getExpectedSol();
+        const actual = readFileSync(outputFile, "utf8");
+        assert.equal(actual, expected);
+      } finally {
+        removeSync(outputFile);
```
galargh, line 428: I put this in a try/finally block to ensure we clean up the temp file on failed runs, too.

galargh (approved):
Great PR, thanks for the contribution! I'll wait for the CI to pass and merge 🥇

PR #5953, opened 2024-11-18
fix: copy .gitignore files during project initialization

galargh:

---

This PR introduces the agreed upon workaround for `npm`s inability to publish `.gitignore` files (see https://github.com/npm/npm/issues/3763). 

Going through the comments on the linked issue, it seems that the most common convention for working around this issue is exactly what we came up with, too - which is to copy `gitignore` template files as `.gitignore` workspace files.

I left `.gitignore` files in the templates because it is useful to have, for example, `node_modules` gitignored during development.

### DIFF
--- a/v-next/hardhat/src/internal/cli/init/init.ts
+++ b/v-next/hardhat/src/internal/cli/init/init.ts
@@ -263,6 +263,32 @@ export async function ensureProjectPackageJson(
 }
 
 /**
+ * The following two functions are used to convert between relative workspace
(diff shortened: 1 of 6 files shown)


galargh:
Interestingly enough `pnpm pack` does respect explicit `templates/**/.gitignore` in `files` of `package.json` but I'm not sure if we can/should rely on it (in particular, I don't think `pnpm changeset publish` would respect it but I haven't tried).

schaable (approved):
Makes sense and it's well documented. LGTM!

PR #5959, opened 2024-11-20
address the follow-up tasks related to --init

### LINKED ISSUE #5954: hardhat --init cleanup
The Hardhat v3 alpha includes `--init` support. Testing for alpha has shown we need some follow up tasks.

## TODO

- [ ] Tests for installation of `--init`, this doesn't need to be e2e but it should test at a minimum it should prove that spawn works on different platforms.
- [ ] `--init` doesn’t show “help us improve…”
- [ ] `--init` doesn’t check for existing dependencies.
    - e.g. if you have `viem` already installed in the package, it just installs it again, potentially a different version
    - Maybe we should check this before even copying any file and fail
        - except for a hardhat clash, which will be common

galargh:

---

Resolves #5954

- [x] Tests for installation of `--init`, this doesn't need to be e2e but it should test at a minimum it should prove that spawn works on different platforms.
- [x] `--init` doesn’t show “help us improve…”
- [x] `--init` doesn’t check for existing dependencies.
    - e.g. if you have `viem` already installed in the package, it just installs it again, potentially a different version
    - Maybe we should check this before even copying any file and fail
        - except for a hardhat clash, which will be common
        
#### Tests for installation

We had the test for installation committed, but marked as skipped. I now enabled it, but I added a conditional skip statement to that test.

Now, if a flag `HARDHAT_DISABLE_SLOW_TESTS=true` is present, the long running tests will be skipped.

The test I reenabled tests installing dependencies for all the templates using `npm` (`pnpm` and `yarn` remain untested for now).

#### Help us improve...

I added the telemetry consent question as the first one in the init flow (immediately after the welcome message). My thinking is that it will allow us to collect data about aborted init flows.

Please let me know if you prefer me to move the question to the end of the flow.

#### Existing dependencies handling

Checking if a dependency already exists before proceeding with installation was already put in place. The behaviour that I opted for in the original version was to skip installation of the dependencies that are already installed. There are some adjustments that I introduced in this PR.

Firstly, we used to only browse the `devDependencies` when checking if a dependency is installed or not. Now, we'll also check `dependencies` just in case.

Secondly, I added version checks. Now, if a dependency is already installed, we'll check if the installed version (or a version range) satisfies the version (or a version range) specified in the template. If it doesn't, we'll suggest upgrading/downgrading the installed version to the version from the template. I intentionally kept the updates separate from the installs of completely new dependencies because I think it is quite likely a user might want to proceed with one while skipping the other. Also, I suspect the messaging around both will evolve in slightly different directions.


alcuadrado:
Thanks, I'll review this one to make sure i understand the installation process

dav3yblaz3 (commented):
lets get this going for testing team

dav3yblaz3 (commented):
let us give this a try team thank you

alcuadrado (approved):

alcuadrado:
Let's move the telemetry consent prompt to be the last question before installation of dependencies

PR #5983, opened 2024-11-26
fix(hardhat-foundry): config clashes condition

wiasliaw:

---

https://github.com/NomicFoundation/hardhat/blob/ad3df5b78691202d399cbfc5a726de16f83dc6e7/packages/hardhat-foundry/src/index.ts#L69-L71

The code above is to solve config collision between foundry and hardhat. However, the condition will cause a weird behavior when foundry config is the same as hardhat config.

## i.e.

foundry config:
```toml
[profile.default]
cache_path = "sample_cache"
```

hardhat config:
```js
/** @type import('hardhat/config').HardhatUserConfig */
module.exports = {
  paths: {
    cache: path.resolve(__dirname, "sample_cache"),
  }
}
```

Run `hardhat compile` and the cache folder will be generated in path `{project_root}/cache_hardhat`.

### DIFF
--- a/packages/hardhat-foundry/src/index.ts
+++ b/packages/hardhat-foundry/src/index.ts
@@ -66,7 +66,7 @@ extendConfig((config, userConfig) => {
     config.paths.root,
     foundryConfig.cache_path
   );
-  if (config.paths.cache === foundryCachePath) {
+  if (config.paths.cache !== foundryCachePath) {
     config.paths.cache = "cache_hardhat";
   }
 


antico5:
This is the intended behavior, we want to avoid foundry and hardhat using the same directory for cache, so when they are set to be the same, we change hardhat's cache directory to `cache_hardhat`. Is there a reason why you would want both caches to be on the same directory?

wiasliaw:
@antico5 sorry for replying late.

afaik, both hardhat and foundry use sloc for compilation. cache files should not be different with the same version of the solc.

antico5:
There might be cases where both could use the same cache directory without conflict, but it's wise to keep them separated to prevent potential clashes. E.g. you can specify different solc versions on the different frameworks. Also each framework may use the cache directory for other purposes besides just the solc cache and from that perspective it's convenient that they are on separate directories.

Issue #5827, opened 2024-10-16
Add resolution for the 'accounts' property in edr configurations


ChristopherDedominici:
Related to [this issue](https://github.com/NomicFoundation/hardhat/pull/5817)

We need to port the `account` property from V2 to V3 for all the configurations. 
Currently in V3 we only support the resolution of the `http` configuration and not the `edr` one.
The resolution for the `http accounts` has already been implemented [here](https://github.com/NomicFoundation/hardhat/pull/5817).

As soon as we support the `edr` configuration, we need to add the resolution for the `accounts` property. The validation for edr is already implemented in [this PR](https://github.com/NomicFoundation/hardhat/pull/5817)

PR #5994, opened 2024-11-27
Rework providers

ChristopherDedominici:
DESCRIPTION
This PR reworks the logic for the providers.

Main Points:
- the providers are now stored in an array, making it easier to track which provider is being executed by stepping through a simple for loop. This replaces the previous solution, which relied on recursive behavior and was harder to follow.
- e2e tests have been simplified. The old tests have been replaced with a single e2e file, which is easier to maintain.
- minor renaming adjustments have been made for improved clarity.

QUESTIONS
- the v2 code was using[ this function](https://github.com/NomicFoundation/hardhat/blob/9ad82f53684abd6ce7e89a297233d08a8228298f/packages/hardhat-core/src/internal/core/jsonrpc/types/base-types.ts#L150) to convert from number to rpc quantity. The v3 code is following [this logic](https://github.com/NomicFoundation/hardhat/blob/fd3b0cc9f7fcb09462174c96b4ee0a9a1ee7cdbd/v-next/hardhat-utils/src/hex.ts#L13). I think the logic is the same, but I'd like to have a double check on this.


v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
```diff
@@ -39,7 +42,20 @@ export class LocalAccounts extends ChainId {
     this.#initializePrivateKeys(localAccountsHexPrivateKeys);
   }
 
-  public async resolveRequest(
+  public async handle(
+    jsonRpcRequest: JsonRpcRequest,
+  ): Promise<JsonRpcRequest | JsonRpcResponse> {
+    const response = await this.#resolveRequest(jsonRpcRequest);
```
ChristopherDedominici, line 48: This handler can modify a JSON-RPC request and return a JSON-RPC response. In the v2 implementation, the execution order first handled the response logic and then processed the request modification. I have retained the same approach here. [See the v2 code here](https://github.com/NomicFoundation/hardhat/blob/9ad82f53684abd6ce7e89a297233d08a8228298f/packages/hardhat-core/src/internal/core/providers/accounts.ts#L45)

v-next/hardhat/test/internal/builtin-plugins/network-manager/request-handlers/e2e.ts
```diff
@@ -0,0 +1,271 @@
+import type { HttpNetworkHDAccountsConfig } from "../../../../../src/types/config.js";
```
ChristopherDedominici, line 1: Are these e2e tests comprehensive enough, or is there an additional scenario you believe would be worth including?

v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/hanlders-array.ts
```diff
@@ -0,0 +1,124 @@
...
+    // have all the transaction fields available, including the
+    // gasPrice / maxFeePerGas & maxPriorityFeePerGas.
+    //
+    // We never use those when using EDR Network, as we sign within the
+    // EDR Network itself. This means that we don't need to provide all the
+    // fields, as the missing ones will be resolved there.
+    //
+    // EDR Network handles this in a more performant way, so we don't use
+    // the AutomaticGasPrice for it.
+    if (networkConfig.type === "http") {
```
ChristopherDedominici, line 58: Is this the if condition that can be removed because also supported in EDR networks?
See related issue here: https://github.com/NomicFoundation/hardhat/issues/5827

alcuadrado: Let's remove this one, but not the one in line 92. Let's let EDR handle the signing of transactions from its own accounts.

v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/network.ts
```diff
@@ -11,14 +12,18 @@ import type {
...
-  // to apply the appropriate modifiers to each request.
-  // When a connection is closed, it is removed from the map. Refer to "closeConnection" at the end of the file.
-  const jsonRpcRequestModifiers: Map<number, JsonRpcRequestModifier> =
-    new Map();
+  // This map is essential for managing multiple network connections in Hardhat V3.
+  // Since Hardhat V3 supports multiple connections, we use this map to track each one
+  // and associate it with the corresponding handlers array.
+  // When a connection is closed, its associated handlers array is removed from the map.
+  // See the "closeConnection" function at the end of the file for more details.
+  const requestHandlersPerConnection: Map<number, RequestHandler[]> = new Map();
```
alcuadrado, line 26: I think we should change it to this, so that the array can still be GC'd if the user doesn't call `close()`.

```suggestion
  const requestHandlersPerConnection: WeakMap<NetworkConnection, RequestHandler[]> = new Map();
```

alcuadrado: To be clear, we should still keep close, as it's still valuable to offer control to the users, but if the GC decides to collect the network connection, it should also collect its `RequestHandler[]`.

v-next/hardhat/test/internal/builtin-plugins/network-manager/request-handlers/e2e.ts
```diff
@@ -0,0 +1,271 @@
...
+    // Use the localhost network for these tests because the modifier is only
+    // applicable to HTTP networks. EDR networks do not require this modifier.
+    const connection = await hre.network.connect("localhost");
+
+    const res = await connection.provider.request({
+      method: "eth_sendTransaction",
+      params: [
+        {
+          to: "0x0000000000000000000000000000000000000012",
+          maxFeePerGas: "0x99",
```
alcuadrado, line 116: this value here confuses me a bit. Why is it necessary?

v-next/hardhat/test/internal/builtin-plugins/network-manager/request-handlers/e2e.ts
```diff
@@ -0,0 +1,271 @@
...
+              Math.floor((LATEST_BASE_FEE_IN_MOCKED_PROVIDER * 9) / 8),
+            ),
+          ],
+          reward: [["0x4"]],
+        },
+        eth_accounts: ["0x123006d4548a3ac17d72b372ae1e416bf65b8eaf"],
+      },
+    );
+
+    // Use the localhost network for these tests because the modifier is only
```
alcuadrado, line 107: This comment will be outdated ones we remove the if mentioned in another comment.

alcuadrado: Same for the most of the other tests

alcuadrado (commented):
I left a few comments, but this looks good. It's now definitely simpler to reason about it.

v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/hanlders-array.ts
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/5994/commits/2fa55f6cc21d79ec78f11dca321ff12057613355)

v-next/hardhat/test/internal/builtin-plugins/network-manager/request-handlers/e2e.ts
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/5994/commits/2fa55f6cc21d79ec78f11dca321ff12057613355)

v-next/hardhat/test/internal/builtin-plugins/network-manager/request-handlers/e2e.ts
ChristopherDedominici: It's not strictly necessary. It's more of an additional check to be sure that there is no conflict with the logic if `maxFeePergas` is present. But it can be removed, I did it [here](https://github.com/NomicFoundation/hardhat/pull/5994/commits/71d8935de87bebcd7b06c7d6ec70a25b517dc415)

alcuadrado (approved):

PR #6007, opened 2024-11-29
Make totalDifficulty field optional, as spec has changed.

nebasuke:
Make totalDifficulty field optional, as spec has changed. See https://github.com/ethereum/go-ethereum/issues/25505


schaable (approved):
LGTM!

PR #6008, opened 2024-12-01
Update test-contracts.md

ebulku:
Every step of this guide is running smoothly in Windows, except this command that is setting also the environment variable.


docs/src/content/hardhat-runner/docs/guides/test-contracts.md
```diff
@@ -378,6 +378,12 @@ The Hardhat Toolbox also includes the [`hardhat-gas-reporter`](https://github.co
 REPORT_GAS=true npx hardhat test
 ```
 
+For Windows users, set the environment variable using PowerShell with `$env:REPORT_GAS="true"`
```
kanej, line 381:
```suggestion
For Windows users, set the environment variable for the PowerShell session with `$env:REPORT_GAS="true"`:
```

kanej (approved):
I have tested this on Windows. It triggers the gas reporting.

PR #6010, opened 2024-12-02
Add check for test networks in hardhat-network-helpers

ChristopherDedominici:
This PR ensures that every method performing an RPC request in the `hardhat-network-helpers` plugin verifies whether the network used for the request is a `test` network. If the network is not a `test` network, an error is thrown. This is because `hardhat-network-helpers` is designed to work exclusively with test networks.

Supported test networks:
- hardhat
- zksync
- anvil


v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
```diff
@@ -29,93 +33,162 @@ import { Time } from "./time/time.js";
...
     return stopImpersonatingAccount(this.#provider, address);
   }
 
-  public takeSnapshot(): Promise<SnapshotRestorer> {
+  public async takeSnapshot(): Promise<SnapshotRestorer> {
+    await this.throwIfNotDevelopmentNetwork();
     return takeSnapshot(this.#provider);
   }
+
+  public async throwIfNotDevelopmentNetwork(): Promise<void> {
```
ChristopherDedominici, line 152: The method is marked as public because it also needs to be used in the `Time` class. However, it will not appear when a user lists available methods during coding, as it is not included in the `NetworkHelpersI` interface.

Other solutions considered:
- Marking the method as private and passing it through the constructor of the `Time` class.
- Applying a decorator to the class that would automatically enforce the check logic on every method.

alcuadrado:
@sz-piotr I think you can review this one to get more familiar with this plugin, as an example of a plugin that extends the network.

v-next/hardhat-network-helpers/src/internal/network-helpers/time/time.ts
```diff
@@ -27,10 +24,12 @@ export class Time implements TimeI {
   }
 
   public async increase(amountInSeconds: NumberLike): Promise<number> {
+    await this.#networkHelpers.throwIfNotDevelopmentNetwork();
```
galargh, line 27: Is the reason why we have to call it explicitly because interaction with the `NetworkHelpers` inside the `increase` implementation happen too late and we want to get the error as early as possible?

v-next/hardhat-network-helpers/test/helpers/mocked-rpc-method.ts
galargh, line 1: Could we give it a more descriptive name maybe? One that suggests that the method being mocked is only `web3_clientVersion`.

v-next/hardhat-network-helpers/test/helpers/mocked-rpc-method.ts
```diff
@@ -0,0 +1,41 @@
...
+        nextContext: HookContext,
+        nextNetworkConnection: NetworkConnection<ChainTypeT>,
+        nextJsonRpcRequest: JsonRpcRequest,
+      ) => Promise<JsonRpcResponse>,
+    ) => {
+      if (jsonRpcRequest.method === "web3_clientVersion") {
+        return {
+          jsonrpc: "2.0",
+          id: jsonRpcRequest.id,
+          result: "non-test-network",
```
galargh, line 32: Do you think it's worth testing the scenarios where we use other development network names that we support? If so, we could make this value configurable here and reuse that mock for the happy cases as well.

galargh (commented):
Great work 🥇 The implementation looks solid! I left a couple of comments to better understand the alternatives.

v-next/hardhat-network-helpers/test/helpers/mocked-rpc-method.ts
ChristopherDedominici: It would complicate the testing a bit, because I will have to use a different logic to pass the mocked value. I think this is a good compromise between testability and coverage

v-next/hardhat-network-helpers/test/helpers/mocked-rpc-method.ts
ChristopherDedominici: done [here](https://github.com/NomicFoundation/hardhat/pull/6010/commits/6df65f59cae5412a832fee7f783a6f98f47807a4)

v-next/hardhat-network-helpers/src/internal/network-helpers/time/time.ts
ChristopherDedominici: Not sure I fully understand the question, but I believe that the answer I gave in a previous comment also works here. The method `throwIfNotDevelopmentNetwork` must be called when a method is used, and not during the class initialization

v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
galargh: Got it! Makes sense, thanks for giving me the extra context. In that case, I think the current implementation totally works.

v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
```diff
@@ -29,93 +33,162 @@ import { Time } from "./time/time.js";
...
+        );
+
+        this.#version = version;
+
+        this.#isDevelopmentNetwork =
+          this.#version.toLowerCase().startsWith("hardhat") ||
+          this.#version.toLowerCase().startsWith("zksync") ||
+          this.#version.toLowerCase().startsWith("anvil");
+      } catch (e) {
+        this.#isDevelopmentNetwork = false;
```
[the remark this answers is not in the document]
ChristopherDedominici: Good question! I think it should be ok. @alcuadrado do you see any reason why the error should not be re-thrown?

v-next/hardhat-network-helpers/src/internal/network-helpers/time/time.ts
galargh: Here, I was thinking what if we didn't `throwIfNotDevelopmentNetwork` at all and just waited for it to fail when we call something on the `networkHelpers`. But I'm assuming we want to prevent this call - https://github.com/NomicFoundation/hardhat/blob/48d64ebed36dac283593e18e75db3348705cc630/v-next/hardhat-network-helpers/src/internal/network-helpers/time/helpers/increase.ts#L15 - from happening in the first place. Which answers my question.

ChristopherDedominici: Correct, we want to prevent that call from happening

v-next/hardhat-network-helpers/test/helpers/mocked-rpc-method.ts
galargh: Ah, now I see how we pass this mock through the config. Alright, that makes sense.

v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
```diff
@@ -29,93 +33,162 @@ import { Time } from "./time/time.js";
...
+          typeof version === "string",
+          `"version" should be a string`,
+        );
+
+        this.#version = version;
+
+        this.#isDevelopmentNetwork =
+          this.#version.toLowerCase().startsWith("hardhat") ||
+          this.#version.toLowerCase().startsWith("zksync") ||
+          this.#version.toLowerCase().startsWith("anvil");
```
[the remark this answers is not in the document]
galargh: Cool! That works for me. Thanks for sharing the version string.

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -798,6 +798,22 @@ This might be caused by using hardhat_reset and loadFixture calls in a testcase.
...
+      websiteTitle:
+        "Hardhat network helpers can only be used with the Hardhat Network",
+      websiteDescription:
+        "Hardhat network helpers can only be used with the Hardhat Network",
+    },
+    CAN_ONLY_BE_USED_WITH_HARDHAT_NETWORK_VERSIONED: {
+      number: 915,
+      messageTemplate: `This helper can only be used with the Hardhat Network. You are connected to "{networkName}", whose identifier is "{version}".`,
+      websiteTitle:
+        "Hardhat network helpers can only be used with the Hardhat Network",
```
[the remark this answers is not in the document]
ChristopherDedominici: Good question! I thought the same but my assumptions was that as long as the keys are not the same (which is impossible because TS would complain) we should be safe.

But to be even safer I changed the message [here](https://github.com/NomicFoundation/hardhat/pull/6010/commits/1dde1f173b64ec11e45edc490d7d7b0ff4d3a8c7)

v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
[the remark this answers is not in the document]
alcuadrado: Agreed. I think it'd be better to just let the error bubble up. We'd retest in the next call, if any.

v-next/hardhat-errors/src/descriptors.ts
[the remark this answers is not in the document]
alcuadrado: I think your take is right, but it wouldn't hurt to use a slightly different title.

v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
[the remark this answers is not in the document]
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/6010/commits/124fba1481075aaf021faadd89148dce6511b24f)

galargh (approved):
Great! Thank you for your help understanding the changeset 🙇

PR #6017, opened 2024-12-05
micro-eth-signer issue when deploying local accounts

ChristopherDedominici:
The branch for this PR was checked out from [this other PR](https://github.com/NomicFoundation/hardhat/pull/5994).

This PR adds logic to handle transactions sent to the null address:

- Updates the version of `micro-eth-signer` to handle transactions directed to the null address. [See release notes](https://github.com/paulmillr/micro-eth-signer/releases/tag/0.12.2).
- If both the "to" field and the "data" field are undefined, an error is thrown.
- If the "to" field is undefined but the "data" field is not, the "to" address is set to "0x", which is the expected address in `micro-eth-signer` for processing transactions to the null address.

NOTES:

If the "to" field and the "data" field are explicitly set to null, a `zod` validation error will be thrown.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
@@ -276,15 +276,22 @@ export class LocalAccountsHandler extends ChainId implements RequestHandler {
       };
     });
 
-    // TODO: Fix after the alpha release
-    assertHardhatInvariant(
-      txData.to !== undefined,
-      "The alpha version doesn't support deploying contracts with local accounts yet",
-    );
+    if (txData.to === undefined && txData.data === undefined) {
+      throw new HardhatError(
+        HardhatError.ERRORS.NETWORK.DATA_FIELD_CANNOT_BE_NULL_WITH_NULL_ADDRESS,
+      );
+    }
 
-    const checksummedAddress = addr.addChecksum(
-      bytesToHexString(txData.to).toLowerCase(),
-    );
+    let checksummedAddress;
+    if (txData.to === undefined) {
+      // This scenario arises during contract deployment. The npm package "micro-eth-signer" does not support
+      // null or undefined addresses. Therefore, these values must be converted to "0x", the expected format.
+      checksummedAddress = "0x";
+    } else {
+      checksummedAddress = addr.addChecksum(
+        bytesToHexString(txData.to).toLowerCase(),
+      );
+    }
 
     assertHardhatInvariant(
       txData.nonce !== undefined,
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
@@ -285,7 +285,7 @@ export class LocalAccountsHandler extends ChainId implements RequestHandler {
     let checksummedAddress;
     if (txData.to === undefined) {
       // This scenario arises during contract deployment. The npm package "micro-eth-signer" does not support
-      // null or undefined addresses. Therefore, these values must be converted to "0x", the expected format.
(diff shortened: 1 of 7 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
```diff
@@ -276,15 +276,22 @@ export class LocalAccountsHandler extends ChainId implements RequestHandler {
       };
     });
 
-    // TODO: Fix after the alpha release
```
ChristopherDedominici, line 279: Can this code be already deleted? Tests are passing so I assume so

alcuadrado: I'm not sure i understood this comment. Haven't you replaced this code in this same pr?

v-next/hardhat/test/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
```diff
@@ -451,6 +451,62 @@ describe("LocalAccountsHandler", () => {
   });
 
   describe("modifyRequest", () => {
+    describe("sending transactions to the null address", () => {
+      it("should succeed when the data field is not empty", async () => {
+        const tx = {
+          // null address: the "to" field is not required
```
alcuadrado, line 457: Given that this situation is kind of tricky, i think we should also add a test with `to: null`. That should behave exactly like this one (i.e. the zod validation probably needs to be updated).

alcuadrado (commented):
I left two small comments, but it looks good.

v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
ChristopherDedominici: Yes, let me rephrase: I wanna double check that is ok to already remove this code. Or if it should stay for now. In that case I would have to skip the tests for the moment

v-next/hardhat/test/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
ChristopherDedominici: So for the `to` property the `null` value should be allowed and it should be handled as we do for `undefined`?

ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/6017/commits/d08c6e08e410bcbec71e5c4cd7a3bce8869d3142)

v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
alcuadrado: Oh, I see! No, the new version without this assert is correct. I just added it as a quick fix during the retreat.

v-next/hardhat/src/internal/builtin-plugins/network-manager/rpc/types/address.ts
```diff
@@ -14,6 +14,7 @@ export const rpcAddress: ZodType<Uint8Array> = conditionalUnionType(
       z.instanceof(Uint8Array),
     ],
     [isAddress, z.string()],
+    [(data) => data === null, z.null()],
```
alcuadrado, line 17: wouldn't this also affect `from`?

ChristopherDedominici: Yes, I allowed the null value only for the "to" field [here](https://github.com/NomicFoundation/hardhat/pull/6017/commits/756248a4fd0b0e3fdeb3e290e161f0697d71bb19)

alcuadrado (approved):

PR #6025, opened 2024-12-06
feat: port the node task to hardhat v3

galargh:

---

In this PR, I port the node task from v2 to v3. I describe my design choices in https://www.notion.so/nomicfoundation/Node-Task-154578cdeaf5802c870ee4c7ea7a05ba (this doc should be helpful in guiding the review)

## Follow-ups

- https://github.com/NomicFoundation/hardhat/issues/6040
- https://github.com/NomicFoundation/hardhat/issues/6042

## Testing

1. I successfully started a node using `hardhat3 node` and deployed a contract from the v2 sample project to it using ignition.
2. I added a test case in which I start a JSON RPC server from an EDR provider, create an additional HTTP provider pointing at the server, and compare responses to an `eth_chainId` request. The test suite could definitely be expanded.


v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
galargh, line 1: This is a direct port of https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/hardhat-network/jsonrpc/handler.ts

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/server.ts
galargh, line 1: This is a direct port of https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-core/src/internal/hardhat-network/jsonrpc/server.ts

galargh:
@schaable Tagging you for a review, too, as per our initiative to spread the process. https://www.notion.so/nomicfoundation/Node-Task-154578cdeaf5802c870ee4c7ea7a05ba should be helpful for a little more context.

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
```diff
@@ -0,0 +1,294 @@
...
+} from "../../network-manager/edr/errors.js";
+import {
+  isJsonRpcRequest,
+  isJsonRpcResponse,
+  isSuccessfulJsonRpcResponse,
+} from "../../network-manager/json-rpc.js";
+import { ProviderError } from "../../network-manager/provider-errors.js";
+
+export class JsonRpcHandler {
+  readonly #provider: EIP1193Provider;
```
schaable, line 25: I believe this should be an `EthereumProvider`

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
```diff
@@ -0,0 +1,294 @@
...
+
+const _handleError = (error: any): JsonRpcResponse => {
+  // extract the relevant fields from the error before wrapping it
+  let txHash: string | undefined;
+  let returnData: string | undefined;
+
+  if (error.transactionHash !== undefined) {
+    txHash = error.transactionHash;
+  }
+  if (error.data !== undefined) {
```
schaable, line 254: minor: the optional chaining operator shouldn't be necessary within this code block.

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
```diff
@@ -0,0 +1,294 @@
...
+      code: error.code,
+      message: error.message,
+    },
+  };
+
+  const data: any = {
+    message: error.message,
+  };
+
+  if (txHash !== undefined) {
```
schaable, line 284: why is this block empty?

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
```diff
@@ -0,0 +1,294 @@
...
+  #sendResponse(
+    res: ServerResponse,
+    rpcResp: JsonRpcResponse | JsonRpcResponse[],
+  ) {
+    res.statusCode = 200;
+    res.setHeader("Content-Type", "application/json");
+    res.end(JSON.stringify(rpcResp));
+  }
+
+  async #handleSingleRequest(req: JsonRpcRequest): Promise<JsonRpcResponse> {
```
schaable, line 149: minor: I’d suggest moving the `handleRequest` function code into this function and renaming this one to `handleRequest`. Having both can be a bit confusing.

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
```diff
@@ -0,0 +1,294 @@
...
+      rpcResp = _handleError(error);
+    }
+
+    // Validate the RPC response.
+    if (!isJsonRpcResponse(rpcResp)) {
+      // Malformed response coming from the provider, report to user as an internal error.
+      rpcResp = _handleError(new InternalError("Internal error"));
+    }
+
+    if (rpcReq !== undefined) {
```
schaable, line 169: Isn’t this guaranteed to be non-undefined due to the earlier `isJsonRpcRequest` check?

schaable (commented):
Hey Piotr! Overall, this looks good, but I’ve left a couple of comments. Feel free to make a decision on the ones marked as minor and consider them resolved.

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
```diff
@@ -0,0 +1,294 @@
...
+      result,
+    };
+  };
+}
+
+const _readJsonHttpRequest = async (req: IncomingMessage): Promise<any> => {
+  let json;
+
+  try {
+    const buf = await getRawBody(req);
```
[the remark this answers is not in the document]
galargh: Sounds great, I'll make the change and add something like this to the utils!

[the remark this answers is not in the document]
galargh: It seems we can do it even simpler:
```
    const bytes: number[] = [];
    for await (const chunk of req) {
        bytes.push(...chunk);
    }
    const text = new TextDecoder("utf-8").decode(new Uint8Array(bytes));
```

I was thinking of putting this inline for now and abstracting it later if/when we have more usecases for this. Do you think that's a good idea?

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/server.ts
```diff
@@ -0,0 +1,101 @@
...
+}
+
+export interface JsonRpcServerConfig {
+  hostname: string;
+  port: number;
+
+  provider: EIP1193Provider;
+}
+
+export class JsonRpcServer implements IJsonRpcServer {
```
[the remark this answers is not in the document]
galargh: Good catch! I copied it from v2, but it totally makes sense. I'll add it to our eng guidelines, too.

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
galargh: A great catch, I have no idea how that happened.

galargh:
Thanks @schaable - that was really helpful 🙇

v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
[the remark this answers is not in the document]
schaable: I think it's a good idea. Go for it!

schaable (approved):
LGTM!

PR #6030, opened 2024-12-09
WIP - Node test runner e2e test

### LINKED ISSUE #5964: Add e2e test for node test runner
We want to avoid the issue of Node test runner not working on windows due to paths and tsx usage. An end to end test that runs the Node test runner (and the Mocha test runner) would have shown this issue in our CI.

Perhaps the simplest approach is to run the example project npm scripts for tests (i.e. `npx hardhat3 test`).

antico5:
This PR adds a simple test script to example-project, which runs node and mocha tests. It also removes it from the CI filter, so it's not excluded from the CI run.

Closes #5964


v-next/example-project/package.json
```diff
@@ -17,7 +17,8 @@
     "lint:fix": "pnpm prettier --write",
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
     "build": "tsc --build .",
-    "clean": "rimraf dist"
+    "clean": "rimraf dist",
+    "test": "hardhat3 test node && hardhat3 test mocha"
```
galargh, line 21: Like suggested in the meeting, we can take inspiration from https://github.com/NomicFoundation/hardhat/blob/v-next/v-next/hardhat/package.json#L49

So something along the lines of the following should work:

```suggestion
    "test": "hardhat3 test node && hardhat3 test mocha",
    "pretest": "pnpm build && pnpm install"
```

antico5: It worked! should be good to go now

galargh (approved):
That's great, thanks 🏅

galargh:
We're failing on Windows, that's interesting! I can help you debug next week if you need a hand.

galargh:
> > We're failing on Windows, that's interesting! I can help you debug next week if you need a hand.
> 
> I created #6051 for this. Is it ok that we handle that separate issue, and when it's fixed, rebase and merge this pr?

Sorry, I didn't see this question before. In general, I don't think we should merge PRs that would create an expectation of failing CI. If we were to proceed with the change without the fix, I would have suggested to disable the e2e on Windows until we have a fix in place. 

Luckily, we do have the fix now so the PR should be ready to go once the CI passes 😄

antico5:
> > > We're failing on Windows, that's interesting! I can help you debug next week if you need a hand.
> > 
> > 
> > I created #6051 for this. Is it ok that we handle that separate issue, and when it's fixed, rebase and merge this pr?
> 
> Sorry, I didn't see this question before. In general, I don't think we should merge PRs that would create an expectation of failing CI. If we were to proceed with the change without the fix, I would have suggested to disable the e2e on Windows until we have a fix in place.
> 
> Luckily, we do have the fix now so the PR should be ready to go once the CI passes 😄

That makes sense! thank you :) I'll go ahead and merge

PR #6043, opened 2024-12-10
Add support tests for project root resolution

ChristopherDedominici:
[This PR](https://github.com/NomicFoundation/hardhat/pull/5517) modified the project root resolution logic but did not include tests. 
This PR addresses that by adding the missing tests.


v-next/hardhat/test/internal/hre-intialization.ts
```diff
@@ -26,6 +27,52 @@ describe("HRE intialization", () => {
   });
 
   describe("createHardhatRuntimeEnvironment", () => {
+    describe("project root", () => {
+      describe("no path param is passed", () => {
+        it("should use the CWD as the default path", async () => {
```
alcuadrado, line 32: This isn't the expected behavior. What happens is that it looks for the closest package.json starting from the CWD, and uses its directory.

ChristopherDedominici: The tests have been modified

alcuadrado (approved):

PR #6046, opened 2024-12-11
Fix plugin dependency checks and set npmPackage values on plugins

### LINKED ISSUE #6035: Include `npmPackage` in all our plugins
Installing a viem project, deleting viem, and trying to connect to a network doesn’t properly detect the installation issue — it fails with a more generic “module not found” error.

We are missing `npmPackage` from all of our plugins’ `index.ts` files. This should allow for a more specific error based on expected dependencies not being installed.

Feedback from Pato on the Hardhat Alpha.

antico5:
This PR addresses #6035 

The original scope of the issue was to set the `npmPackage` property of our hardhat3 plugins, so when an import fails we can give detailed information if a given plugin package is missing, or some of its peer dependencies.

When running a test scenario, specifically using `hardhat-viem` and not having the `viem` peer dep installed, I noticed that even after setting the `npmPackage` property on our plugin, the error message wasn't helping. This lead me to debug the core `plugins/detect-plugin-npm-dependency-problems.ts` module. 

I noticed that the `require.resolve` function was not working, and after thorough debugging I found out that the `createRequire` function doesn't work properly when passing an absolute path to a directory without a trailing forward slash. Adding the trailing forward slash to the directory path would make the `require` instance work.

After that, I started to get the error that `package.json` was not being exported by our modules, and we need to `require()` those to find the peer dependencies of packages. Then I added an export of `package.json` to all of our modules.

Closes #6035


alcuadrado (approved):
Approved as a temporary fix. We should make this work without exporting `package.json`. Luckily #6125 already does it for the resolver.

PR #6047, opened 2024-12-11
Add validations for HardhatUserConfig.paths

ChristopherDedominici:
This PR introduces validation for the `paths` property in the `HardhatUserConfig` object.

To verify the validity of all the paths, we ensure they are string values. Path resolution (including the check for the existence of a path) is handled later in the code when the path is resolved.


v-next/hardhat/src/internal/core/config-validation.ts
```diff
@@ -110,6 +124,63 @@ export function collectValidationErrorsForUserConfig(
...
+    }
+  }
+
+  return validationErrors;
+}
+
+function validatePath(
+  filePath: unknown,
+  pathName: "cache" | "artifacts" | "tests" | "sources",
+  index?: number, // If the filePath is part of an array, pass its index to make the error message more specific
```
kanej, line 164:
```suggestion
  index?: number,
```
I think we can leave the comment out, the usage is pretty clear; and the function is not exported so the invocation in practice is in the function above.

I think you either go with a full JSDOC comment on the `validatePath` function to include the param, or no comment and rely on context. Both work.

kanej (approved):
I left two suggestions, but this looks good.

zoeyTM (approved):
agree with the changes John suggested, but LGTM

Issue #2888, opened 2022-07-03
Feature: Deno support


sambacha:
I am guessing v3 of Hardhat is being planned. With that said It may be benifical to give Deno a serious consideration. 

###  Native web worker support 
[see this issue about NodeJs having web worker support](https://github.com/nodejs/node/issues/43583)

### sane FFI [see https://deno.land/manual/runtime/ffi_api](https://deno.land/manual/runtime/ffi_api)
- If Hardhat is going to adopt a Rust based EVM solution this would be easier I would imagine than working with Nodejs as is

### X-Typescript-Types
HRE could adopt providing responses  with a custom X-TypeScript-Types HTTP header when the types (.d.ts) is defined. This is useful for deno type checks ([link](https://deno.land/manual/typescript/types#using-x-typescript-types-header)).

![](https://esm.sh/embed/assets/sceenshot-deno-types.png)

>**Note**
> The idea being that as `solc` supports imports via URL, there may be use cases in having type checking work like that (am unsure of this just an idea). This would be for browsers 

> Example (Browser): pass the ?no-check query to disable the X-TypeScript-Types header if some types are incorrect:

```typescript
import SafeTransferLib from "https://abi.storage/rari-capital/solmate@0.7.0/~/SafeTransferLib?no-check"
```
> [see esm.sh/](https://esm.sh/)

### Possibility for Interoperable packages 

All required remote dependencies are referenced in a file called `deps.ts` and the required methods and classes are re-exported. The dependent local modules then reference the deps.ts rather than the remote dependencies. If now for example one remote dependency is used in several files, upgrading to a new version of this remote dependency is much simpler as this can be done just within deps.ts.

```typescript
/**
 * deps.ts
 *
 * This module re-exports the required methods from the dependant remote module.
 */
 export { SafeTransferLib } from "https://x.nest.land/ramda@0.7.0/source/index.js?pin=0x0000DeployedAddress
```

[see https://deno.land/manual/examples/manage_dependencies](https://deno.land/manual/examples/manage_dependencies)

### Maintain NodeJS Compatibility easily with `dnt` 

Deno to NodeJs [`dnt`](https://deno.land/manual/node/dnt) allows you to develop your Deno module mostly as-is and use a single Deno script to build, type check, and test an npm package in an output directory. Once built, you only need to npm publish the output directory to distribute it to Node.js users.

### WebAssembly and Rust 

[...]

PR #6053, opened 2024-12-14
Fix tsx.register() crash on Deno

SebastienGllmt:
This fixes the issue mentioned [here](https://github.com/NomicFoundation/hardhat/issues/2888#issuecomment-2524080632)

This PR definitely doesn't make Deno easy to use because of the `forge` issue mentioned [here](https://github.com/NomicFoundation/hardhat/issues/2888#issuecomment-2496323103), but it at least makes it usable

### DIFF
--- a/v-next/hardhat/src/internal/cli/main.ts
+++ b/v-next/hardhat/src/internal/cli/main.ts
@@ -41,12 +41,8 @@ import { printVersionMessage } from "./version.js";
 
 export interface MainOptions {
   print?: (message: string) => void;
-  registerTsx?: true;
+  registerTsx?: boolean;
   rethrowErrors?: true;
-}
-
-declare global {
-  const Deno: any;
 }
 
 export async function main(
@@ -102,10 +98,9 @@ export async function main(
       return;
     }
 
-    if (typeof Deno === "undefined") {
-      if (options.registerTsx) {
-        register();
-      }
+    
+    if (options.registerTsx) {
+      register();
     }
 
     const userConfig = await importUserConfig(configPath);
--- a/v-next/hardhat/src/internal/cli/main.ts
+++ b/v-next/hardhat/src/internal/cli/main.ts
@@ -98,8 +98,7 @@ export async function main(
       return;
     }
 
-    
-    if (options.registerTsx) {
+    if (options.registerTsx === true) {
       register();
     }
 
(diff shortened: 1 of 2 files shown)


v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -45,6 +45,10 @@ export interface MainOptions {
   rethrowErrors?: true;
 }
 
+declare global {
```
alcuadrado, line 48: What if instead we modify the `MainOptions` a bit, so that it only defaults `registerTsx` to true if it's running on Node.js? That way we don't need any `declare` nor dependency.

I believe we can detect node like this:

```ts
const isNode = typeof process === 'object' &&
               process?.versions?.node !== undefined;
```

alcuadrado: We could also default it to false, and do the `isNode` evaluation in `src/index.ts`.

alcuadrado:
Hey! Thanks a ton for this PR! I just left a comment with an idea, but in general I'm in favor of this change.

v-next/hardhat/src/internal/cli/main.ts
SebastienGllmt: Checking if something "isNode" is unfortunately hard

The reason is that a lot of existing code has some equivalent of "isNode" to differentiate backend vs frontend. To make sure those projects still work in different runtimes like Deno and Bun, both of the return `true` for this kind of isNode process despite the fact they aren't actually nodejs

SebastienGllmt: I made the type `Deno: any` specifically to avoid having to bring in any dependency so that it's just type stuff that disappears when you actually compile

alcuadrado: > To make sure those projects still work in different runtimes like Deno and Bun, both of the return true for this kind of isNode process despite the fact they aren't actually nodejs

Oh, I was scared that would be the situation :( but is understandable.

I guess we should go for a blocklist approach, and not register it under runtimes we know it's not needed.

Then maybe we can detect Deno without any type declaration with `"Deno" in globalThis`.

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -98,7 +98,7 @@ export async function main(
       return;
     }
 
-    if (options.registerTsx) {
+    if (options.registerTsx === true) {
```
SebastienGllmt, line 101: we need `=== true` to satisfy the linter (because it complains `if (options.registerTsx)` is not clear if this is meant as a non-null check or as a check that the variable is true)

alcuadrado (approved):

PR #6068, opened 2024-12-19
Port flatten task to v3

### LINKED ISSUE #5647: Implement `npx hardhat flatten` task in v3
Implement the `npx hardhat flatten` task in the v3 command line:

```shell
Hardhat version 2.22.8

Usage: hardhat [GLOBAL OPTIONS] flatten [...files]

POSITIONAL ARGUMENTS:

  files The files to flatten 

flatten: Flattens and prints contracts and their dependencies. If no file is passed, all the contracts in the project will be flattened.

```

This task will take advantage of the new build system.

antico5:
This PR introduces an implementation for the `flatten` task. Both the implementation and the tests were ported and adapted from v2. 

Details to point out:
- The usage of the `tsort` package was replaced by `toposort`, which has 30x weekly download count and specially, it's typed.
- On v2, there were multiple subtasks associated with flatten. The most important being (shortened): `FLATTEN`, `GET_FLATTENED_SOURCE`, `GET_FLATTENED_SOURCE_AND_METADATA`. I wasn't sure there was a need to define more than one task, so I went with this approach. Maybe I could move the printing (contract and warnings)  to a separate one, in case just the flattened file and metadata need to be consumed by an external plugin.
- For testing I used several fixture projects, which was the approach taken in v2. I noticed the engineering guidelines suggest to avoid this unless necessary, but I'd find it dificult to test this without using them. If we wanted that, we'd have to think about a mock filesystem/resolver.
- For asserting stdio I hooked the console.log and console.warn implementations on the test cases. I tried to find an injection approach but it wasn't clear to me how, since tasks don't allow function arguments.
- On v2 tests, flattened contracts were compiled. I didn't include that part on v3 tests, since I'm not sure it's in scope. As I understand it, the flatten task doesn't guarantee the the output is valid, compilable solidity code.

Closes #5647


v-next/hardhat/package.json
```diff
@@ -68,6 +68,7 @@
     "@types/debug": "^4.1.4",
     "@types/node": "^20.14.9",
     "@types/semver": "^7.5.8",
+    "@types/toposort": "^2.0.7",
```
alcuadrado, line 71: We have an implementation of topological sort, which we should reuse

v-next/hardhat/test/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -0,0 +1,446 @@
...
+    };
+    console.warn = (...args: unknown[]) => {
+      consoleWarnBuffer.push(args.join(" "));
+    };
+  });
+
+  afterEach(() => {
+    console.log = originalConsoleLog;
+    console.warn = originalConsoleWarn;
+  });
```
alcuadrado, line 62: Instead of doing this, the task action can accept these params

```
print = console.log,
warn = console.warn
```

which are not exposed to the CLI, nor hooked into it in any way.

Then, to test it, we just need to call the task action directly.

alcuadrado:
> * The usage of the `tsort` package was replaced by `toposort`, which has 30x weekly download count and specially, it's typed.

Also added a comment about this, but we already have an implementation of top sort.

> * On v2, there were multiple subtasks associated with flatten. The most important being (shortened): `FLATTEN`, `GET_FLATTENED_SOURCE`, `GET_FLATTENED_SOURCE_AND_METADATA`. I wasn't sure there was a need to define more than one task, so I went with this approach. Maybe I could move the printing (contract and warnings)  to a separate one, in case just the flattened file and metadata need to be consumed by an external plugin.

Single task is the way to go. If we need to make something extensible, we can add hooks.

> * For testing I used several fixture projects, which was the approach taken in v2. I noticed the engineering guidelines suggest to avoid this unless necessary, but I'd find it dificult to test this without using them. If we wanted that, we'd have to think about a mock filesystem/resolver.

I guess it's fine for this case.

> * For asserting stdio I hooked the console.log and console.warn implementations on the test cases. I tried to find an injection approach but it wasn't clear to me how, since tasks don't allow function arguments.

See the comments.

> * On v2 tests, flattened contracts were compiled. I didn't include that part on v3 tests, since I'm not sure it's in scope. As I understand it, the flatten task doesn't guarantee the the output is valid, compilable solidity code.

I think this is ok.

v-next/hardhat/test/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -0,0 +1,446 @@
...
+
+async function assertFlattenedFilesResult(flattened: string) {
+  const expected = await getExpectedSol();
+
+  assert.equal(flattened, expected);
+}
+
+async function createHRE() {
+  return createHardhatRuntimeEnvironment({}, {}, process.cwd());
+}
```
galargh, line 43: Can we just do `createHardhatRuntimeEnvironment({})` inline?

v-next/hardhat/test/internal/builtin-plugins/flatten/task-action.ts
galargh: If we wanted to do that, here's an example - https://github.com/NomicFoundation/hardhat/blob/00119c33b4636f4580be9df87c0e4f5b654b405d/v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts#L16

We do, however, assert on the returned values already so it might not be as necessary to inspect the entire stdio. It might still be useful to suppress it though.

v-next/hardhat/test/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -0,0 +1,446 @@
...
+
+        for (let i = 0; i + 1 < runs; i++) {
+          assert.equal(flattenedFiles[i], flattenedFiles[i + 1]);
+        }
+      });
+    });
+  });
+
+  describe("SPDX licenses and pragma abicoder directives", () => {
+    describe("Flatten files that dont contain SPDX licenses or pragma directives", () => {
```
galargh, line 163:
```suggestion
    describe("Flatten files that don't contain SPDX licenses or pragma directives", () => {
```

v-next/hardhat/test/fixture-projects/flatten-task/consistent-build-info-names/.gitignore
galargh, line 1: We don't really need this file in this fixture project I think.

v-next/hardhat/test/fixture-projects/flatten-task/consistent-build-info-names/hardhat.config.js
```diff
@@ -0,0 +1,7 @@
+// This project is compiled from scratch multiple times in the same test, which
+// produces a lot of logs. We override this task to omit those logs.
```
galargh, line 2: In the v3 implementation of the task, we don't do the compilation so I don't think we need this either. In fact, I think we should be able to remove all the `hardhat.config.js`/`hardhat.config.ts` from these projects and still get the tests to work as expected. To be verified though.

galargh (commented):
I think it's a good idea to use an internal version of topological sort implementation if we have most of it already in place.

Other than that most of my comments are minor and, as I suspect, mostly directed at the v2 implementation of the task. Feel free to take into account only where it makes sense to you.

As for the testing side of this, I think there are some good candidates for lower level testing - like helper functions, regexes, etc. We don't necesarilly have to add this range of tests in this iteration though. The higher level tests, the E2E ones do seem reasonable to me and do cover the core of the task so I'd be happy to proceed with them.

v-next/hardhat/package.json
antico5: I removed the dependency and just used the dfs approach for walking and sorting the files. The order is deterministic but it's different than in v2. As discussed in the call, we are ok with that change.

v-next/hardhat/test/internal/builtin-plugins/flatten/task-action.ts
antico5: I added two extra params for the log and warn functions. Using ReplOptions seemed a bit overkill for these because it's just output (perfect fit for the console action althought).

v-next/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -0,0 +1,300 @@
...
+    rootPaths.toSorted(), // We sort them to have a deterministic order
+    config.paths.root,
+    config.solidity.remappings,
+  );
+
+  let flattened = "";
+
+  // Return empty string when no files are resolved
+  if (Array.from(dependencyGraph.getAllFiles()).length === 0) {
+    return { flattened, metadata: null };
```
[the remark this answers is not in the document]
antico5: Since this is a port, I'd try to minimize the changes to the external interface

v-next/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -0,0 +1,300 @@
...
+    normalizedText = commentLicenses(normalizedText);
+    normalizedText = commentPragmaAbicoderDirectives(normalizedText);
+
+    // Write files without imports, with commented licenses and pragma abicoder directives
+    flattened += `\n\n// File ${file.sourceName}\n`;
+    flattened += `\n${normalizedText}\n`;
+  }
+
+  // Print the flattened file
+  console.log(flattened);
```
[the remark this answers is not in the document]
antico5: For now we just match v2 behavior. It's printed to stdout.

v-next/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -0,0 +1,300 @@
...
+    console.warn(
+      chalk.yellow(
+        `\nThe following file(s) do NOT specify SPDX licenses: ${filesWithoutLicenses.join(
+          ", ",
+        )}`,
+      ),
+    );
+  }
+
+  if (pragmaDirective !== "" && filesWithoutPragmaDirectives.length > 0) {
```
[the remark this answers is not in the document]
antico5: Yes, it's acceptable that no files have a pragma abicoder directive. In that case, we don't want to show any warning.

v-next/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -0,0 +1,300 @@
...
+        directivesByImportance.indexOf(directive)
+      ) {
+        directive = normalizedPragma;
+      }
+
+      // Update the most important pragma directive for the current file
+      if (
+        directivesByImportance.indexOf(normalizedPragma) >
+        directivesByImportance.indexOf(fileMostImportantDirective)
+      ) {
```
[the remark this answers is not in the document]
antico5: With the current implementation, non supported directives are just ignored

v-next/hardhat/test/fixture-projects/flatten-task/consistent-build-info-names/.gitignore
antico5: It's necessary for the node_modules folder to be version controlled.

v-next/hardhat/test/fixture-projects/flatten-task/consistent-build-info-names/hardhat.config.js
antico5: That's correct. I just removed all of the config files

v-next/hardhat/test/internal/builtin-plugins/flatten/task-action.ts
galargh: Oh no, I wasn't suggesting repl options. That's task specific. I just wanted to show you how we set up args that are only exposed for testing.

antico5:
@galargh I just pushed a commit with all the changes, and rebased on top of v-next. Should be good to review

v-next/hardhat/test/internal/builtin-plugins/flatten/task-action.ts
antico5: Ahh I see. Thanks !

galargh:
> @galargh I just pushed a commit with all the changes, and rebased on top of v-next. Should be good to review

Thanks! I'll re-review today 💪

v-next/hardhat/package.json
galargh: @antico5 Let's remove the `@types/toposort` dependency, too.

galargh (approved):
Great! Let's just remove the `@types/toposort` dependency and we should be good to go 🥇

PR #6073, opened 2024-12-23
Replace solc dependency with a thin solidity_compile wrapper

### LINKED ISSUE #6071: trim down solcjs and keep only the wrapper
Extracted out of https://github.com/NomicFoundation/hardhat/issues/5990 as it will be easier to review as a separate contribution.

galargh:

---

Resolves #6071

This is a follow-up to https://github.com/NomicFoundation/hardhat/pull/6056

In this PR, I replace the solc-js dependency with a thin wrapper over the `solidity_compile` method. 

It was created by extracting the parts of the solc-js package (https://github.com/ethereum/solc-js) that we were actually using.


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/solcjs-wrapper.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export interface SolcWrapper {
+  compile: CompileWrapper;
+}
+
+export type CompileWrapper = (input: string) => string;
+
+export default function wrapper(solc: Solc): SolcWrapper {
+  const version = bindVersion(solc);
+  const isVersion6OrNewer = semver.gte(versionToSemver(version()), "0.6.0");
```
galargh, line 37:
```suggestion
  const semverVersion = versionToSemver(version());
  const isVersion6OrNewer = semver.gte(semverVersion, "0.6.0");
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/solcjs-wrapper.ts
```diff
@@ -0,0 +1,138 @@
...
+  const reset = bindReset(solc);
+  const compile = bindCompile(solc, isVersion6OrNewer);
+
+  if (compile === undefined) {
+    // eslint-disable-next-line no-restricted-syntax -- should we use a HardhatError here?
+    throw new Error(
+      'Could not find the "compile" function in the solc library',
+    );
```
galargh, line 45: I'm wondering if we should throw a `HardhatError` here?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/solcjs-wrapper.ts
```diff
@@ -0,0 +1,138 @@
...
+  // eslint-disable-next-line @typescript-eslint/naming-convention -- this is a C function
+  _solidity_compile?: Compile | null;
+}
+
+type Reset = () => string;
+type Version = () => string;
+type Compile = (
+  input: string,
+  callbackPtr: number | null,
+  callbackContextPtr?: null,
```
galargh, line 26: This is an odd type, but we either want to skip this argument or provide an explicit null. That depends on the version of solc that we're going to be using.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/solcjs-wrapper.ts
```diff
@@ -0,0 +1,138 @@
...
+
+function compileWrapper(
+  isVersion6OrNewer: boolean,
+  compile: Compile,
+  reset?: Reset,
+): CompileWrapper {
+  return (input: string): string => {
+    const output = isVersion6OrNewer
+      ? compile(input, null, null)
+      : compile(input, null);
```
galargh, line 61: The underlying compile functions expect explicit values as far as I understand. Also, their arguments differ depending on the version being used.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/solcjs-wrapper.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    if (reset !== undefined) {
+      // Explicitly free memory.
+      //
+      // NOTE: cwrap() of "compile" will copy the returned pointer into a
+      //       Javascript string and it is not possible to call free() on it.
+      //       reset() however will clear up all allocations.
+      reset();
+    }
```
galargh, line 70: If I'm not mistaken, we should be doing that before compilation.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/solcjs-runner.ts
```diff
@@ -28,7 +26,8 @@ async function main() {
   const solcjsPath = process.argv[2];
   const solc = await getSolcJs(solcjsPath);
 
-  const output = solc.compile(input);
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- the input read from the stdin should be a string
+  const output = solc.compile(input as string);
```
schaable, line 30: Looking at the `readStream` implementation, it seems we are already returning a `string`. Could we type the return statement to avoid casting it here?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/solcjs-wrapper.ts
schaable: I'd say yes, unless we expect this to never happen. In that case, I'd use an invariant error.

schaable (approved):
LGTM, but I'd wait for @alcuadrado's review on this one since he's more familiar with solc

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/solcjs-runner.ts
galargh: Of course, good catch!

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/solcjs-wrapper.ts
galargh: Done!

galargh:
> LGTM, but I'd wait for @alcuadrado's review on this one since he's more familiar with solc

I've only just realised I merged the solc replacement as part of https://github.com/NomicFoundation/hardhat/pull/6056. Now, I also merged this one because after merging the other one, this one ended up consisting only of the changes requested in the review.

@alcuadrado I put together a branch on which I revert the solc wrapper changes altogether in case we want to roll them back for now: https://github.com/NomicFoundation/hardhat/compare/v-next...revert-solcjs-compile-wrapper

PR #6081, opened 2024-12-26
Update npm package link for hardhat-gasless-deployer plugin

emretepedev:
While browsing new hardhat plugins, I noticed that the npm URL for the **hardhat-gasless-deployer** plugin was returning a 404 error. The issue was that the direct link was provided instead of the package name. I fixed the URL to make it easier for others to find.

### DIFF
--- a/docs/src/content/hardhat-runner/plugins/plugins.ts
+++ b/docs/src/content/hardhat-runner/plugins/plugins.ts
@@ -819,7 +819,7 @@ const communityPlugins: IPlugin[] = [
     name: "hardhat-gasless-deployer",
     author: "Ahmed Ali",
     authorUrl: "https://twitter.com/0xAhmedAli",
-    npmPackage: "https://www.npmjs.com/package/hardhat-gasless-deployer",
+    npmPackage: "hardhat-gasless-deployer",
     description: "Deploy contracts with Hardhat using Gas Station Network",
     tags: ["GSN", "Gasless", "Deployment"],
   },


docs/src/content/hardhat-runner/plugins/plugins.ts
```diff
@@ -819,7 +819,7 @@ const communityPlugins: IPlugin[] = [
     name: "hardhat-gasless-deployer",
     author: "Ahmed Ali",
     authorUrl: "https://twitter.com/0xAhmedAli",
-    npmPackage: "https://www.npmjs.com/package/hardhat-gasless-deployer",
+    npmPackage: "hardhat-gasless-deployer",
```
schaable, line 822: This is interesting because some plugins use a full URL while others use the name. Since the name is unique, I suggest we standardize by updating all plugins in the list to use only the name and remove the URL. @alcuadrado, what do you think?

emretepedev: Hi @schaable! As far as I can see, no other plugins in the list use a full URL for the `npmPackage` field.


The field specifically responsible for redirecting to npm is `npmPackage`, whereas others like authorUrl are used for different types of links.

schaable: You're correct. I did a quick search, and the field with the full URL is `authorUrl`, which looks fine.

schaable (approved):

PR #6086, opened 2024-12-27
Translate path separators when parsing npm direct imports

### LINKED ISSUE #6051: Compilation is failing on windows
Running compile on example-project on windows fails with:

```
Error HHE1223: You are tying to resolve the npm file "@openzeppelin/contracts/token/ERC20/ERC20.sol", but it doesn't exist within its package.
```

Found this after adding example-project to the CI as part of the e2e tests

Example failed run: https://github.com/NomicFoundation/hardhat/actions/runs/12314374318/job/34374354685?pr=6030

galargh:

---

Resolves #6051

The issue was that we used POSIX path separators in the `dependenciesToCompile`, but then we weren't translating them to system-aware paths anywhere along the way. Eventually, it made it all the way through to https://github.com/NomicFoundation/hardhat/blob/v-next/v-next/hardhat-utils/src/fs.ts#L120, which expects system-specific path separators, and that's where we errored out.

The resolver for npm paths is https://github.com/NomicFoundation/hardhat/blob/v-next/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts#L250. I added the path separator translator to https://github.com/NomicFoundation/hardhat/blob/v-next/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts#L254 because it seemed like the most natural option.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts
@@ -1318,6 +1318,7 @@ export class ResolverImplementation implements Resolver {
         path: string;
       }
     | undefined {
+    // NOTE: We assume usage of path.posix.sep in the direct import
     const directImportPattern =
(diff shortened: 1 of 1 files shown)


antico5 (approved):
Great find!

alcuadrado (approved):
great catch!

PR #6088, opened 2024-12-27
Implement SolidityCallSite.toString()

### LINKED ISSUE #6067: Replace `source-map-support` functionality
In Hardhat 2 we used a library `source-map-support`. We need to replace that functionality.

## Notes

- Source-map-support lib converted class into string
- we don’t use `source-map-support` so we need a custom inspect function that replicate
- Add directly to SolidityCallSite
- The conversion can be ad-hoc for the key errors, it does not need to be generic

antico5:
Solidity stack traces should be stringified properly now.

Extracted the logic from source-map-support package.

Closes #6067

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
@@ -309,7 +309,7 @@ export class SolidityError extends Error {
   }
 }
 
-class SolidityCallSite implements NodeJS.CallSite {
+export class SolidityCallSite implements NodeJS.CallSite {
   readonly #sourceName: string | undefined;
   readonly #contract: string | undefined;
   readonly #functionName: string | undefined;
@@ -335,7 +335,7 @@ class SolidityCallSite implements NodeJS.CallSite {
     return undefined;
   }
 
-  public getFileName() {
+  public getFileName(): string {
     return this.#sourceName ?? "unknown";
   }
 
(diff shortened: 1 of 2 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
```diff
@@ -420,7 +420,41 @@ class SolidityCallSite implements NodeJS.CallSite {
...
+
+    if (this.getLineNumber() !== null) {
+      fileLocation += `:${this.getLineNumber()}`;
+    }
+
+    let line = "";
+    const functionName = this.getFunctionName();
+    let addSuffix = true;
+    const isConstructor = this.isConstructor();
+    const isMethodCall = !(this.isToplevel() || isConstructor);
```
schaable, line 435: I've checked the definition of `isToplevel()`, and it always seems to return `false`. Am I missing something?

schaable (approved):
Looks good. Just left one question.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
antico5: Some of these CallSite interface functions are returning fixed values, but probably that's temporary. Since I'm not really sure, I didn't exclude the logic from the toString() function, in case in the future those functions may be of use

alcuadrado: agreed with keeping the v8 logic just in case

PR #6120, opened 2025-01-07
feat: adopt EDR multichain

Wodann:
This PR tracks changes needed to backport EDR's multi-chain architecture to Hardhat 2.

After merging this PR, Hardhat will no longer support `smock`, as per the maintainer's own archival of the repo.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
@@ -230,6 +230,7 @@ export class EdrProviderWrapper
       genesisState.push({
         address: privateToAddress(privateKey),
         balance: BigInt(account.balance),
+        code: new Uint8Array(), // Empty account code, removing potential delegation code when forking
       });
 
       return account.privateKey;


packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToEdr.ts
```diff
@@ -195,7 +212,7 @@ export function edrRpcDebugTraceToHardhat(
     structLogs.shift();
   }
 
-  let returnValue = rpcDebugTrace.output?.toString("hex") ?? "";
+  let returnValue = rpcDebugTrace.output?.toString() ?? "";
```
fvictorio, line 215: This chnge surprised me because it looks wrong but it actually works fine.

Why it looks wrong: the type of `rpcDebugTrace.output` is allegedly `Uint8Array` (see [here](https://github.com/NomicFoundation/edr/blob/0047135e068f06ca314ced43636ad4d9e110ed9a/crates/edr_napi/index.d.ts#L361)). But the `.toString` method of `Uint8Array` does _not_ format the result in hexadecimal.

Why it works: the actual value received here is a string. Calling `.toString` on a string just returns the same value. That string is already a hex representation of the output.

This is not a blocker, but seems worth investigating. Why is EDR's `index.d.ts` wrong?

PR #6123, opened 2025-01-08
Remove shell lang from some md snippets

fvictorio:
There are many markdown snippets that have `sh` as the info string but whose content is not shell code. This makes them be incorrectly highlighted. But the main issue is that some angle brackets are escaped:

image

To be honest, the fact that removing the info string fixes this doesn't make a lot of sense to me. I don't know why they are escaped only when a language is specified. I looked down the remark/rehype rabbit hole, but it was too deep so I went with the fix that works but that I don't understand.


kanej:
Closing as clearly generated by a low value LLM.

kanej (approved):
The generated preview does not have the escaping issue.

PR #6125, opened 2025-01-08
Support package exports

### LINKED ISSUE #5992: npm package export support within Solidity import resolution
Hardhat 3's Solidity import resolution works well with npm packages installed locally (i.e. under `./node_modules`). It does not however work with npm packages which use package.json exports.
To provide first class support for npm as a package manager for Solidity, we should support these exports as part of Solidity import resolution.

antico5:
This PR adds support for package.json#exports for packages that distribute solidity files.

Implementation details:
- Source names for npm modules use the subpath without resolving, e.g. `npm/mypkg@1.2.3/Foo.sol` instead of `npm/mypkg@1.2.3/src/Foo.sol`, for an import of the kind `import 'mypkg/Foo.sol'`. This is arbitrary
- Module resolving was changed from using `require.resolve` to manual lookup using the help of `require.resolve.paths`, to support packages that don't export their package.json
- Exports resolving make use of the `resolve.exports` package. In this case we avoid using `require.resolve` because it may resolve to different packages in the filesystem for a given package name and version.

Known issues:
- When trying to import a module with wrong casing, in a module that uses exports, I'm not sure how we can suggest the correct casing. E.g. `mypkg/TheContract.sol` resolves to `mypkg/src/TheContract.sol`. If user does `import 'mypkg/thecontract.sol'`, I know that the correct resolved path is `src/TheContract.sol`, but I'm not sure how to <unresolve> it back to `TheContract.sol`. Asking for suggestions on how to approach this

Closes #5992


v-next/hardhat-utils/src/package.ts
```diff
@@ -102,6 +105,34 @@ export async function findClosestPackageRoot(
...
+  const lookupPaths = require.resolve.paths(packageName) ?? [];
+
+  const pathToTest = [...packageName.split("/"), "package.json"];
+
+  for (const lookupPath of lookupPaths) {
+    const packageJsonPath = path.join(lookupPath, ...pathToTest);
+
+    try {
+      await fs.promises.access(packageJsonPath, fs.constants.R_OK);
+      return await fs.promises.realpath(packageJsonPath);
```
alcuadrado, line 129: This should use the utils' `fs` module

v-next/hardhat-utils/src/package.ts
```diff
@@ -102,6 +105,34 @@ export async function findClosestPackageRoot(
   return path.dirname(packageJsonPath);
 }
 
+/**
+ * Finds the package json for a given package
+ * @param from the absolute path from where to start the search
+ * @param packageName the name of the package to find
+ * @returns the absolute real path (resolved symlinks) of the package.json
+ */
+export async function findPackageJson(
```
alcuadrado, line 114: We should at at least one or two tests of this

v-next/hardhat/src/types/solidity/resolved-file.ts
```diff
@@ -1,3 +1,5 @@
+import type { Exports } from "resolve.exports";
```
alcuadrado, line 1: We should duplicate this type to avoid re-exporting a dependency type

alcuadrado:
Note to self:

Double check that this situation:

```
import "foo/a.sol";
```

where `foo` has
```
exports: {
  "./*.sol": "./src/*.sol"
}
```

and `./src/a.sol` doesn't exist.

throws that "foo/a.sol" doesn't exist, and not "foo/src/a.sol".

alcuadrado:
@antico5: my changes are ready to review

alcuadrado:
sure, @antico5

alcuadrado (approved):

PR #6128, opened 2025-01-09
chore: check if wasm compiler is already downloaded

galargh:

---

When updating how solidity compilers are downloaded, I forgot to add a check if a compiler is already downloaded before the call to download the wasm compiler. This PR fixes that.


schaable (approved):
Looks good!

PR #6129, opened 2025-01-09
feat: implement the solidity compilation cache

galargh:

---

In this PR, I implement the solidity compilation cache as outlined in the [associated design doc](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d).

The implementation diverges from the design doc in 1 areas:
- [when the compiler output cache is cleaned up](https://github.com/NomicFoundation/hardhat/pull/6129#discussion_r1935539979)


v-next/hardhat-utils/src/crypto.ts
galargh, line 1: #### Context

The changes to this file are related to the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

After the changes, `createNonCryptographicHashId` is an asynchronous function. I also changed the hashing algorithm from MD5 to SHA-1. This is OK because the only place where we currently use this function is the `getBuildId` method of the `CompilationJobImplementation`.

v-next/hardhat-utils/src/fs.ts
galargh, line 1: #### Context

Here, I add two new helper functions, `getAccessTime` and `getSize` which return last access time and size of a file respectively.

They are needed to implement the `clean` part of the [caching component](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf5804da858d661034584ec) which removes cache entries based on file size and "age".

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/artifacts.ts
galargh, line 1: #### Context

It is necessary to change the signatures of `getBuildInfo` and `getBuildInfoOutput` because they both use `getBuildId` which turned asynchronous as part of the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
galargh, line 1: #### Context

The [build system changes](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf5800aa96afaf3846ed2bd) are the core part of introducing solidity compilation caching.

The best part of the changes is concentrated in the `build` method of the `SolidityBuildSystemImplementation` where we now interact with the cache.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -118,11 +131,44 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
 
     const compilationJobs = [...new Set(compilationJobsPerFile.values())];
 
-    // TODO: Filter the compilation jobs based on the cache
+    await Promise.all(
+      compilationJobs.map(async (compilationJob) =>
+        compilationJob.getBuildId(),
+      ),
+    );
```
galargh, line 138: Computing build IDs is time consuming as it involves hashing all the compiler input sources. That's why we do it asynchronously. Once a compilation job build ID is computed, it will be returned straight away on the subsequent requests.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -131,10 +177,27 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
+    void Promise.all(
+      uncachedSuccessfulResults.map(async (result) => {
+        return this.#compilerOutputCache.set(
+          await result.compilationJob.getBuildId(),
+          result.compilerOutput,
+        );
+      }),
+    ).then(() => {
+      return this.#compilerOutputCache.clean();
+    });
```
galargh, line 196: [When should we clean up the cache](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf5807099dae5e86efdc624) was one of the open questions stated in the design doc.

To do so from the `cleanArtifacts` method was suggested. I do like this suggestion as then all the "clean" operations would be grouped together. However, when working on the final implementation, I realised it might be more beneficial to call the cache clean from build after all.

As you can see in the snippet, this allows us to ensure that clean is only ever called after all the writes have finished. That way we can avoid (or at least postpone) implementing synchronisation mechanisms between writes and cleans in the cache component.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/cache.ts
galargh, line 1: #### Context

This is the implementation of the [caching component](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf5804da858d661034584ec).

It is expected to support caching compiler outputs to begin with. We can extend/change it as needed in the future.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+  readonly #path: string;
+  readonly #defaultMaxAgeMs: number;
+  readonly #defaultMaxSize: number;
+
+  constructor(
+    basePath: string,
+    namespace: string,
+    version: string,
+    defaultMaxAgeMs: number = 7 * 24 * 60 * 60 * 1000, // 1 week
+    defaultMaxSize: number = 2 * 1024 * 1024 * 1024, // 2 GB
```
galargh, line 23: These are the default values we agreed upon [here](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf580f3aaa6d946581e7ac7). They are not exposed to the user at the moment.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+    const filePath = path.join(this.#path, key);
+    await writeJsonFile(filePath, value);
+  }
+
+  public async get(key: string): Promise<T | undefined> {
+    const filePath = path.join(this.#path, key);
+    return (await exists(filePath)) ? readJsonFile<T>(filePath) : undefined;
+  }
+
+  public async clean(maxAgeMs?: number, maxSize?: number): Promise<void> {
```
galargh, line 40: Being able to override the default max age and size is really useful for testing purposes.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+    const files = await getAllFilesMatching(this.#path);
+    const fileInfos = await Promise.all(
+      files.map(async (file) => ({
+        file,
+        atimeMs: (await getAccessTime(file)).getTime(),
+        size: await getSize(file),
+      })),
+    );
+
+    const sortedFileInfos = fileInfos.sort((a, b) => a.atimeMs - b.atimeMs);
```
galargh, line 53: We sort the files in the ascending order of their last access time here.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+    const filesToRemove: string[] = [];
+
+    for (const fileInfo of sortedFileInfos) {
+      if (fileInfo.atimeMs < minAtimeMs || size > maxSize) {
+        filesToRemove.push(fileInfo.file);
+        size -= fileInfo.size;
+      } else {
+        break;
+      }
+    }
```
galargh, line 70: We remove the files starting from the front of the list (the "oldest" entries) until the total size of the cache is smaller than the maximum cache size **and** the oldest entry is newer than the maximum age of a cache entry.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
galargh, line 1: #### Context

The changes in this file are concerned with the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a). They are related to the effort to make the `getBuildId` asynchronous and to cache resolved file content hashes.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -20,6 +21,10 @@ export class CompilationJobImplementation implements CompilationJob {
 
   #buildId: string | undefined;
   #solcInput: CompilerInput | undefined;
+  #solcInputWithoutSources: Omit<CompilerInput, "sources"> | undefined;
+  #resolvedFiles: ResolvedFile[] | undefined;
```
galargh, line 25: By adding these two new properties on the compilation job, we ensure they are only ever computed once per compilation job.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -20,6 +21,10 @@ export class CompilationJobImplementation implements CompilationJob {
...
+
+  static readonly #sourceContentHashCache = new Map<string, string>();
```
galargh, line 27: In [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a) section of the design doc, I proposed caching resolved file hashes on the instances of `ProjectResolvedFile` and `NpmResolvedFile` respectively. 

When I started working on the final implementation, I realised that `ProjectResolvedFile` and `NpmResolvedFile` are currently interfaces rather than classes. So, in order to add a `getContentHash` method on them, we would have to first turn them into ones.

Before doing that, I thought I'd propose starting with a private static resolved file content hash cache on the `CompilationJobImplementation` instead. At least until `CompilationJobImplementation` is the only consumer of said cache.

I'm not attached to this idea though, so if you have better suggestions or if you think we should implement the original proposal, please let me know.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -108,20 +137,58 @@ export class CompilationJobImplementation implements CompilationJob {
...
+      resolvedFiles.map(async (file) => {
+        sources[file.sourceName] = {
+          hash: await this.#getSourceContentHash(file),
+        };
+      }),
+    );
+
+    const sortedSources = Object.fromEntries(
+      Object.entries(sources).sort((a, b) => a[0].localeCompare(b[0])),
+    );
```
galargh, line 175: We need to sort the sources because the sources map might be populated out of order which does affect serialisation.

v-next/hardhat/src/types/solidity/build-system.ts
```diff
@@ -147,13 +147,9 @@ export type FileBuildResult =
...
 export interface CacheHitFileBuildResult {
   type: FileBuildResultType.CACHE_HIT;
-  // TODO: Should we remove this? It is a buildId of an already existing build
-  // info.
-  // NOTE: The buildId and contractArtifactsGenerated are useful when one uses
-  // the build system programatically and wants to find out what artifacts
-  // were generated for a given file, and with what configuration.
   buildId: string;
   contractArtifactsGenerated: string[];
+  warnings: CompilerOutputError[];
```
galargh, line 152: Since we're storing the entire compiler outputs in the cache, we're able to extra warnings out of them even when they're retrieved from cache.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -55,43 +54,76 @@ describe(
...
+    };
+
+    before(async () => {
+      expectedArtifactsPath = path.join(process.cwd(), "artifacts");
+      expectedCachePath = path.join(process.cwd(), "cache");
+      await remove(expectedArtifactsPath);
+      await remove(expectedCachePath);
+      const hooks = new HookManagerImplementation(process.cwd(), []);
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- We don't care about hooks in this context
+      hooks.setContext({} as HookContext);
```
galargh, line 92: When we call `build()` it is asserted that the hook context is not `undefined`. We don't actually use hooks in this tests.

A proper `HookContext` contains `hre`. I didn't want to create a full `hre` for this test. Hence, the type cast.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
galargh: One alternative that I can think of is as follows. What if we add an optional `content.hash` property on both `ProjectResolvedFile` and `NpmPackageResolvedFile`, and have a method `getResolvedFileContentHash(resolvedFile: ResolvedFile)` that returns that property if it is set and otherwise sets it on the interface and then returns it. The downside to this is that after such change, `ResolvedFile`s would no longer be immutable.

v-next/hardhat-utils/src/crypto.ts
galargh, line 1: #### Context

The changes to this file are related to the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

After the changes, `createNonCryptographicHashId` is an asynchronous function. I also changed the hashing algorithm from MD5 to SHA-1. This is OK because the only place where we currently use this function is the `getBuildId` method of the `CompilationJobImplementation`.

v-next/hardhat-utils/src/fs.ts
galargh, line 1: #### Context

Here, I add two new helper functions, `getAccessTime` and `getSize` which return last access time and size of a file respectively.

They are needed to implement the `clean` part of the [caching component](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf5804da858d661034584ec) which removes cache entries based on file size and "age".

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/artifacts.ts
galargh, line 1: #### Context

It is necessary to change the signatures of `getBuildInfo` and `getBuildInfoOutput` because they both use `getBuildId` which turned asynchronous as part of the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

v-next/hardhat-utils/src/crypto.ts
```diff
@@ -17,17 +15,22 @@ export async function keccak256(bytes: Uint8Array): Promise<Uint8Array> {
...
-export function createNonCryptographicHashId(data: string): string {
-  return createHash("md5").update(data).digest("hex");
+export async function createNonCryptographicHashId(
+  data: string,
+): Promise<string> {
+  const message = new TextEncoder().encode(data);
+  const buffer = await crypto.subtle.digest("SHA-1", message);
+  const array = Array.from(new Uint8Array(buffer));
+  return array.map((b) => b.toString(16).padStart(2, "0")).join("");
 }
```
galargh, line 36: #### Context

The changes to this file are related to the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

After the changes, `createNonCryptographicHashId` is an asynchronous function. I also changed the hashing algorithm from MD5 to SHA-1. This is OK because the only place where we currently use this function is the `getBuildId` method of the `CompilationJobImplementation`.

v-next/hardhat-utils/src/fs.ts
```diff
@@ -407,6 +407,50 @@ export async function getChangeTime(absolutePath: string): Promise<Date> {
...
+    return stats.atime;
+  } catch (e) {
+    ensureError<NodeJS.ErrnoException>(e);
+    if (e.code === "ENOENT") {
+      throw new FileNotFoundError(absolutePath, e);
+    }
+
+    throw new FileSystemAccessError(e.message, e);
+  }
+}
```
galargh, line 430: #### Context

Here, I add two new helper functions, `getAccessTime` and `getSize` which return last access time and size of a file respectively.

They are needed to implement the `clean` part of the [caching component](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf5804da858d661034584ec) which removes cache entries based on file size and "age".

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/artifacts.ts
```diff
@@ -104,9 +104,9 @@ declare module "@ignored/hardhat-vnext/types/artifacts" {
 }`;
 }
 
-export function getBuildInfo(
+export async function getBuildInfo(
```
galargh, line 107: #### Context

It is necessary to change the signatures of `getBuildInfo` and `getBuildInfoOutput` because they both use `getBuildId` which turned asynchronous as part of the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolved-file.ts
galargh, line 1: #### Context

I add implementations of `ProjectResolvedFile` and `NpmPackageResolvedFile` interfaces here. These are necessary in order to lazily store content hashes on them, as proposed as part of the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

v-next/hardhat/src/types/solidity/resolved-file.ts
galargh, line 1: #### Context

I add `getContentHash()` methods on the `ProjectResolvedFile` and `NpmPackageResolvedFile` interfaces here. These functions are needed to implement the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolved-file.ts
```diff
@@ -0,0 +1,66 @@
...
+  NpmPackageResolvedFile,
+  ProjectResolvedFile,
+  ResolvedNpmPackage,
+} from "../../../../types/solidity.js";
+
+import { createNonCryptographicHashId } from "@ignored/hardhat-vnext-utils/crypto";
+
+import { ResolvedFileType } from "../../../../types/solidity.js";
+
+export class ProjectResolvedFileImplementation implements ProjectResolvedFile {
```
galargh, line 12: #### Context

I add implementations of `ProjectResolvedFile` and `NpmPackageResolvedFile` interfaces here. These are necessary in order to lazily store content hashes on them, as proposed as part of the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

v-next/hardhat/src/types/solidity/resolved-file.ts
```diff
@@ -35,14 +35,15 @@ export interface ResolvedNpmPackage {
...
 export enum ResolvedFileType {
   PROJECT_FILE = "PROJECT_FILE",
-  NPM_PACKGE_FILE = "NPM_PACKAGE_FILE",
+  NPM_PACKAGE_FILE = "NPM_PACKAGE_FILE",
 }
 
 /**
  * A file that's part of the Hardhat project (i.e. not installed through npm).
  */
 export interface ProjectResolvedFile {
```
galargh, line 44: #### Context

I add `getContentHash()` methods on the `ProjectResolvedFile` and `NpmPackageResolvedFile` interfaces here. These functions are needed to implement the [`getBuildId` optimisations](https://www.notion.so/nomicfoundation/solidity-compilation-cache-17b578cdeaf580e3ba10c0d907fd6a4d?pvs=4#182578cdeaf58080aa28e353759ff37a).

v-next/hardhat-utils/src/crypto.ts
```diff
@@ -17,17 +15,22 @@ export async function keccak256(bytes: Uint8Array): Promise<Uint8Array> {
  *
  * This function is primarily intended for generating unique identifiers from
  * a given input string.
- * It uses the MD5 hash algorithm, which is not cryptographically secure, but
+ * It uses the SHA-1 hash algorithm, which is not cryptographically secure, but
```
alcuadrado, line 18: this LGTM

v-next/hardhat-utils/src/crypto.ts
```diff
@@ -17,17 +15,22 @@ export async function keccak256(bytes: Uint8Array): Promise<Uint8Array> {
...
  */
-export function createNonCryptographicHashId(data: string): string {
-  return createHash("md5").update(data).digest("hex");
+export async function createNonCryptographicHashId(
+  data: string,
+): Promise<string> {
+  const message = new TextEncoder().encode(data);
+  const buffer = await crypto.subtle.digest("SHA-1", message);
+  const array = Array.from(new Uint8Array(buffer));
+  return array.map((b) => b.toString(16).padStart(2, "0")).join("");
```
alcuadrado, line 35:
```suggestion
  const buffer = await crypto.subtle.digest("SHA-1", message);
  return Buffer.from(buffer).toString("hex");
```

This is probably much faster, as it doesn't iterate in js.

v-next/hardhat-utils/src/fs.ts
```diff
@@ -407,6 +407,50 @@ export async function getChangeTime(absolutePath: string): Promise<Date> {
...
+
+/**
+ * Retrieves the size of a file.
+ *
+ * @param absolutePath The absolute path to the file.
+ * @returns The size of the file in bytes.
+ * @throws FileNotFoundError if the path does not exist.
+ * @throws FileSystemAccessError for any other error.
+ */
+export async function getSize(absolutePath: string): Promise<number> {
```
alcuadrado, line 440: What about this? So that it's clear that it's for file, and not directories?

```suggestion
export async function getFileSize(absolutePath: string): Promise<number> {
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/artifacts.ts
alcuadrado: (these comments are really helpful)

v-next/hardhat/src/types/solidity/build-system.ts
alcuadrado: Nice! 

Note to self: are we remapping these?

v-next/hardhat-utils/src/crypto.ts
galargh: Oh nice! It totally makes sense. I haven't thought of that initially.

v-next/hardhat-utils/src/fs.ts
galargh: Yes! I like it!

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+    defaultMaxAgeMs: number = 7 * 24 * 60 * 60 * 1000, // 1 week
+    defaultMaxSize: number = 2 * 1024 * 1024 * 1024, // 2 GB
+  ) {
+    this.#path = path.join(basePath, namespace, version);
+    this.#defaultMaxAgeMs = defaultMaxAgeMs;
+    this.#defaultMaxSize = defaultMaxSize;
+  }
+
+  public async set(key: string, value: T): Promise<void> {
+    const filePath = path.join(this.#path, key);
```
alcuadrado, line 31: It's unclear to me if these files have any extension. I think I would add `.json` to play safe.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+    await writeJsonFile(filePath, value);
```
alcuadrado, line 32: Given that these can be pretty large files, I think we should write to a temporary name first, and then rename it to the intended name. That way, if hardhat dies while writing it, we wouldn't corrupt the file.

```
    const filePath = path.join(this.#path, key);
    const tmpFilePath = filePath + ".tmp";
    
    await writeJsonFile(tmpFilePath, value);
    await move(tmpFilePath, filePath)
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -0,0 +1,74 @@
...
+    let size = sortedFileInfos.reduce(
+      (acc, fileInfo) => acc + fileInfo.size,
+      0,
+    );
+    const minAtimeMs = new Date().getTime() - maxAgeMs;
+
+    const filesToRemove: string[] = [];
+
+    for (const fileInfo of sortedFileInfos) {
+      if (fileInfo.atimeMs < minAtimeMs || size > maxSize) {
```
alcuadrado, line 64: This will always remove files that are too large.

I think the `maxSize` logic should only be applied if the total size of the files exceeds `maxSize`, and in that case you delete the LRU ones until the total size is below that threshold again.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado, line 138:
```suggestion
    // We precompute the build ids here, which are cached internally in each compilation job
    await Promise.all(
      compilationJobs.map(async (compilationJob) =>
        compilationJob.getBuildId(),
      ),
    );
```

What about adding a comment like this?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -131,10 +177,27 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
-    const isSuccessfulBuild = results.every(
-      (result) => !this.#hasCompilationErrors(result),
+    const uncachedResults = results.filter((result) => !result.cached);
+    const uncachedSuccessfulResults = uncachedResults.filter(
+      (result) => !this.#hasCompilationErrors(result.compilerOutput),
     );
 
+    // NOTE: We're not waiting for the writes and clean to finish because we
+    // will only care about the result of these operations in subsequent runs
+    void Promise.all(
```
alcuadrado, line 187: I think this is unsafe, which we would be ok though if we implement this change: https://github.com/NomicFoundation/hardhat/pull/6129/files#r1944738230

I think I'd move it to its own private method (a sync `#updateCache`?), and better explain why it's safe.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: I think once we update the cache to include partial recompilation we may want to revisit this, but it works at the moment.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -204,6 +267,18 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
           continue;
         }
 
+        if (result.cached) {
+          resultsMap.set(formatRootPath(publicSourceName, root), {
+            type: FileBuildResultType.CACHE_HIT,
+            buildId,
+            contractArtifactsGenerated:
+              contractArtifactsGenerated.get(publicSourceName) ?? [],
+            warnings: errors,
```
alcuadrado, line 276: self-answer: yes, these warnings have been remapped

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
alcuadrado: great point. What about adding a comment?

galargh: Done :D

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
galargh: Of course, done 📖

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
galargh: It was! You're right! I did move it to a separate method and added a longer docstring with a callout that it might have to be reconsidered in the future. Great catch, thank you 🙇

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: ok for now, but this will break soon, when we add hooks to the build system.

alcuadrado (approved):

PR #6143, opened 2025-01-14
chore: check test task options before passing on noCompile flag

galargh:

---

As @fvictorio rightly pointed out in https://github.com/NomicFoundation/hardhat/pull/5870#discussion_r1914587375, we shouldn't have removed the `noCompile` flag from the `test solidity` task as all the "test" tasks are expected to have it implemented.

In this PR, I remove that expectation by adding a `options.has("noCompile")` check to the parent `test` task.


alcuadrado:
I think we should probably explore fixing the issue that franco raised instead.

alcuadrado (approved):

PR #6158, opened 2025-01-19
Fix ESM_PROJECT_WITHOUT_CJS_CONFIG heuristic

fvictorio:
This fixes the "latest dependency versions" CI job. I don't know why this started happening now, but I confirmed manually that these extra checks are needed.

### DIFF
--- a/packages/hardhat-core/src/internal/core/config/config-loading.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-loading.ts
@@ -29,11 +29,14 @@ export function importCsjOrEsModule(filePath: string): any {
     const imported = require(filePath);
     return imported.default !== undefined ? imported.default : imported;
   } catch (e: any) {
-    if (
-      e.code === "ERR_REQUIRE_ESM" ||
+    // An ESM project that has a Hardhat config with a .js extension will fail to be loaded,
+    // because Hardhat configs can only be CJS but a .js extension will be interpreted as ESM.
+    // The kind of error we get in these cases depends on the Node.js version.
+    const node20Heuristic = e.code === "ERR_REQUIRE_ESM";
+    const node22Heuristic =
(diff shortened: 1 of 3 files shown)


packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -29,7 +29,11 @@ export function importCsjOrEsModule(filePath: string): any {
     const imported = require(filePath);
     return imported.default !== undefined ? imported.default : imported;
   } catch (e: any) {
-    if (e.code === "ERR_REQUIRE_ESM") {
+    if (
+      e.code === "ERR_REQUIRE_ESM" ||
+      e.message === "module is not defined" ||
```
alcuadrado, line 34: Do you mind adding a comment about when these cases happen? @fvictorio

alcuadrado: Other than that, LGTM

fvictorio: I realized that this fix, while useful, has nothing to do with the latest dependencies CI job. The problem is that I cloned a fresh Hardhat repo, deleted the lockfile, and ran the tests, and got a failure for the test that checks this. But this failed because I was using node v22; the error in the CI was different :man_facepalming:

I still think having this is useful, and double-checked manually that it's necessary, although I don't understand why this doesn't fail when the CI is run with v22.

Anyway, the latest dependencies job was failing because solhint published a new version that makes our "no-errors" fixture project have an error. I fixed that in this PR too.

alcuadrado (approved):

packages/hardhat-core/src/internal/core/config/config-loading.ts
alcuadrado: ouch

PR #6164, opened 2025-01-20
feat: cleanup provider and stack traces logic

schaable:
This is the final cleanup PR for the network stack in preparation for the alpha (and ideally the beta). It includes:

 - Normalized the EDR and HTTP providers to make them as similar as possible.
 - Removed the emission of the `hardhatNetworkReset` event in the `request` method, as it no longer makes sense in v3.
 - Free resources for garbage collection on `close` (unset the `provider` for the EDR provider and the `dispatcher` for the HTTP provider) and throw an error if the `request` method is called after closing.
 - EDR Provider – `request` method:
   - Removed `InvalidInputError` logic, as it is already handled by `getJsonRpcRequest`.
   - Replaced assertions on `jsonRpcResponse.result` with `assertHardhatInvariant`.
   - Cleaned up the `#handleEdrResponse` function.
 - Improved functions in `stack-trace-solidity-errors.ts` by removing TODOs and enhancing overall code flow.
 - Made other minor adjustments to improve code clarity and flow.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
@@ -22,7 +22,10 @@ import {
   l1GenesisState,
   l1HardforkFromString,
 } from "@ignored/edr-optimism";
-import { assertHardhatInvariant } from "@ignored/hardhat-vnext-errors";
+import {
+  assertHardhatInvariant,
+  HardhatError,
+} from "@ignored/hardhat-vnext-errors";
 import { toSeconds } from "@ignored/hardhat-vnext-utils/date";
 import { numberToHexString } from "@ignored/hardhat-vnext-utils/hex";
 import { deepEqual } from "@ignored/hardhat-vnext-utils/lang";
@@ -109,9 +112,9 @@ interface EdrProviderConfig {
 }
 
 export class EdrProvider extends BaseProvider {
-  readonly #provider: Readonly<Provider>;
   readonly #jsonRpcRequestWrapper?: JsonRpcRequestWrapperFunction;
 
+  #provider: Provider | undefined;
   #nextRequestId = 1;
 
   /**
@@ -176,6 +179,10 @@ export class EdrProvider extends BaseProvider {
   public async request(
     requestArguments: RequestArguments,
   ): Promise<SuccessfulJsonRpcResponse["result"]> {
+    if (this.#provider === undefined) {
+      throw new HardhatError(HardhatError.ERRORS.NETWORK.PROVIDER_CLOSED);
+    }
+
     const { method, params } = requestArguments;
 
     const jsonRpcRequest = getJsonRpcRequest(
@@ -190,6 +197,11 @@ export class EdrProvider extends BaseProvider {
       jsonRpcResponse = await this.#jsonRpcRequestWrapper(
         jsonRpcRequest,
         async (request) => {
+          assertHardhatInvariant(
+            this.#provider !== undefined,
+            "The provider is not defined",
+          );
+
           const stringifiedArgs = JSON.stringify(request);
           const edrResponse =
             await this.#provider.handleRequest(stringifiedArgs);
@@ -244,7 +256,8 @@ export class EdrProvider extends BaseProvider {
   }
 
   public async close(): Promise<void> {
-    // TODO: what needs cleaned up?
+    // Clear the provider reference to help with garbage collection
+    this.#provider = undefined;
   }
 
   async #handleEdrResponse(
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
@@ -39,6 +39,10 @@ import { getJsonRpcRequest, isFailedJsonRpcResponse } from "../json-rpc.js";
 import { getGlobalEdrContext } from "./edr-context.js";
 import { InvalidArgumentsError, ProviderError } from "./errors.js";
 import { createSolidityErrorWithStackTrace } from "./stack-traces/stack-trace-solidity-errors.js";
+import {
+  isDebugTraceResult,
+  isEdrProviderErrorData,
+} from "./type-validation.js";
 import { clientVersion } from "./utils/client-version.js";
 import { ConsoleLogger } from "./utils/console-logger.js";
 import {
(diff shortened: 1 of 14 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
```diff
@@ -212,23 +215,35 @@ export function edrRpcDebugTraceToHardhat(
   });
 
   // REVM trace adds initial STOP that Hardhat doesn't expect
+  // ASK: is this still valid?
```
alcuadrado, line 218: This isn't resolved

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
```diff
@@ -212,23 +215,35 @@ export function edrRpcDebugTraceToHardhat(
...
     returnValue = "";
   }
 
   return {
-    failed: !rpcDebugTrace.pass,
-    gas: Number(rpcDebugTrace.gasUsed),
-    returnValue,
+    failed: !debugTraceResult.pass,
+    gas: Number(debugTraceResult.gasUsed),
+    returnValue, // ASK: do we expect an unprefixed hex string here?
```
alcuadrado, line 231: This isn't resolved

v-next/hardhat/src/types/providers.ts
```diff
@@ -88,17 +88,6 @@ export interface EthereumProvider extends EIP1193Provider {
   ): void;
 }
 
-/**
- * A message emitted by the provider as a result of an eth_subscription.
- */
-export interface EthSubscription {
```
alcuadrado, line 94: This shouldn't be deleted. It's our implementation of a standard.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
```diff
@@ -280,27 +278,38 @@ function getMessageFromLastStackTraceEntry(
...
+  ) {
     super(message);
-    this.stackTrace = stackTrace;
-  }
 
-  // TODO: Can we bring this back with isolated declarations?
-  // public [inspect](): string {
-  //   return this.inspect();
-  // }
+    Object.defineProperty(this, Symbol.for("nodejs.util.inspect.custom"), {
```
alcuadrado, line 306: Nice trick!

No need to define "inspect" in 314. We can use `value: () => { ... }`

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
```diff
@@ -20,48 +17,52 @@ import {
...
 ): SolidityError {
-  const previousPrepareStackTrace = Error.prepareStackTrace;
-  Error.prepareStackTrace = (error, stack) => {
-    if (previousStack !== undefined) {
-      stack = previousStack;
-    } else {
-      // We remove error management related stack traces
-      stack.splice(0, 1);
-    }
+  const originalPrepareStackTrace = Error.prepareStackTrace;
```
alcuadrado, line 26: we should assert that this isn't undefined

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
```diff
@@ -20,48 +17,52 @@ import {
...
+
+    const solidityError = new SolidityError(
+      message,
+      stackTrace,
+      data,
+      transactionHash,
+    );
+
+    /* eslint-disable-next-line @typescript-eslint/no-unused-expressions
+    -- As the stack property is lazy-loaded in JavaScript, we need to access it
```
alcuadrado, line 58:
```suggestion
    -- As the stack property is lazy-loaded in v8, we need to access it
```

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -283,66 +283,61 @@ export class EdrProvider extends BaseProvider {
       }
 
       if (stackTrace !== null) {
-        error = encodeSolidityStackTrace(
-          jsonRpcResponse.error.message,
+        // Pass data and transaction hash from the original error
+        /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+        -- If we have a stack trace, we know that the json rpc response data
+        is an object with the data and transactionHash fields */
+        const responseErrorData = responseError.data as {
```
alcuadrado, line 290: Let's assert this, as this field has a lot of historic nuance, and if an assumption is incorrect, we should detect it and fix it asap.

alcuadrado: Also, note for @fvictorio @nebasuke: while this error data is now returned as `JsonRpcReponse.data`, it could as well be part of our own metadata object, which would allow us to have stronger type guarantees.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -283,66 +283,61 @@ export class EdrProvider extends BaseProvider {
...
-  #emitLegacySubscriptionEvent(subscription: string, result: any) {
+  #emitLegacySubscriptionEvent(subscription: string, result: unknown) {
     this.emit("notification", {
       subscription,
       result,
     });
   }
 
   #emitEip1193SubscriptionEvent(subscription: string, result: unknown) {
-    const message: EthSubscription = {
```
alcuadrado, line 337: There's a comment about this below, but let's keep the type here.

alcuadrado (commented):
I left a few comments.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
schaable: @Wodann is going to double-check this, but for now it will remain as it is. I'll change the ASK for a TODO for easier tracking.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
schaable: Can you answer this one? Depending on the response, we may be able to simplify the logic.

PR #6186, opened 2025-01-23
feat: add testFiles option to test solidity

### LINKED ISSUE #6176: Support running specific tests in Solidity Test
The Private Alpha of Solidity test supports only running the full test suite.
We should provide command line options for running only a specific test. We should explore what other options for test filtering should be made available.

galargh:
Resolved #6176

---

This introduces an option to run only a subset of solidity tests. The CLI interface I implemented is exactly the same as with the `test node` task.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/index.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/index.ts
@@ -11,6 +11,11 @@ const hardhatPlugin: HardhatPlugin = {
   },
   tasks: [
     task(["test", "solidity"], "Run the Solidity tests")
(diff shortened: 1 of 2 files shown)


ChristopherDedominici (approved):
LGTM. Just a question: is it worth to add a simple test just to cover this new addition?

alcuadrado (approved):

galargh:
> LGTM. Just a question: is it worth to add a simple test just to cover this new addition?

We're yet to set up solidity test plugin tests. For now, I created the issue for that - https://github.com/NomicFoundation/hardhat/issues/6192 - and I assigned it to myself. I'll jump right on it after we clear the tasks we planned for this sprint.

PR #6199, opened 2025-01-29
chore: remove duplicated information from assertion error details

galargh:

---

I noticed that the information about the diff in the assertion error details is duplicated. We used to deduplicate it before the changes I made to the general structure of the reporter. In this PR, I restore the diff deduplication for assertion errors in our custom reporter.

Before:
```
...
1 failing

  1) crypto
       createNonCryptographicHashId
         Should create a non-cryptographic hash-based identifier:

   AssertionError: Expected values to be strictly equal:
   + actual - expected
   
   + 'aaf4c61ddcc5e8a2dabede0f3b482cd9aea9434d'
   - '5d41402abc4b2a76b9719d911017c592'
   - Expected
   + Received
   
   - 5d41402abc4b2a76b9719d911017c592
   + aaf4c61ddcc5e8a2dabede0f3b482cd9aea9434d
   
       at TestContext.<anonymous> (test/crypto.ts:20:14)
```

After:
```
...
1 failing

  1) crypto
       createNonCryptographicHashId
         Should create a non-cryptographic hash-based identifier:

   AssertionError: Expected values to be strictly equal
   - Expected
   + Received
   
   - 5d41402abc4b2a76b9719d911017c592
   + aaf4c61ddcc5e8a2dabede0f3b482cd9aea9434d
   
       at TestContext.<anonymous> (test/crypto.ts:20:14)
```


alcuadrado:
I found an issue when reviewing this, and I went ahead and fixed it and added test for this functionality.

Can you review my changes, @galargh ? Thanks

alcuadrado (commented):
(see the other comment)

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -123,14 +123,38 @@ function formatSingleError(
...
-    message = message.match(/^(AssertionError:\s.*)\:/)?.[1] ?? message;
+    // If we got a diff, we try to remove any diff from the message, which can
+    // have different formats.
+
+    // Format 1: A line starting with "+ actual - expected".
+    let match = message.match(/^(.*?)\n\s*?\+\sactual\s-\sexpected/s);
+
+    // Format 2: A line starting with "{error.actual} !==".
+    if (match === null && "actual" in error) {
+      match = message.match(/^(.*?)\n\s*?(?:.*?) !==/s);
```
galargh: In https://github.com/NomicFoundation/hardhat/pull/6199/commits/9b0d5b19d37346739c84398b1de76d8b24c4a9db, I updated the first matching group to be greedy. That way we match everything up to the line with `!==`. I also updated the tests to catch if that was not the case.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -123,14 +123,38 @@ function formatSingleError(
...
+    }
+
+    // Format 3: AssertionError: {message}:
+    match = match ?? message.match(/^(AssertionError:\s.*)\:/);
```
galargh, line 147: I'm thinking that maybe we could remove this altogether because we're not really sure what we're removing here, right? In 1 and 2, we're removing known diff formats, which should cover most of the cases. And if we new diff formats are added, we can always add them.

galargh (commented):
Great catch with the custom messages @alcuadrado 🥇 

I fixed one more issue with `!==` diffs, and posted one comment on the 3rd diff removal clause. Let me know what you think :D

v-next/hardhat-node-test-reporter/src/error-formatting.ts
alcuadrado: That's a good point!

v-next/hardhat-node-test-reporter/src/error-formatting.ts
alcuadrado: Looks good.

v-next/hardhat-node-test-reporter/src/error-formatting.ts
galargh: OK, I'm going to remove it then 💪

alcuadrado (approved):

PR #6200, opened 2025-01-29
chore: pass build info as buffer

agostbiro:


agostbiro:
Incorporated in https://github.com/NomicFoundation/hardhat/pull/6203

Issue #6198, opened 2025-01-28
Improve tests for validation errors in the network manager


schaable:
Rework the tests to use `assertValidationErrors` and add any missing test cases.

PR #6201, opened 2025-01-30
Network manager tests

schaable:
The PR adds tests to the network manager, specifically for the config resolution phase and providers. It also fixes a bug in `edrRpcDebugTraceToHardhat`, where the `output` was mistakenly treated as a `Buffer` instead of a `string` (due to an incorrect type from EDR). Finally, it groups the `ProviderError`s into a single file and normalizes their constructors.

Follow-up issues:
 - https://github.com/NomicFoundation/edr/issues/783
 - https://github.com/NomicFoundation/hardhat/issues/6198


v-next/hardhat/src/internal/builtin-plugins/network-manager/provider-errors.ts
```diff
@@ -41,7 +55,75 @@ export class ProviderError extends CustomError implements ProviderRpcError {
...
+
+export class InternalError extends ProviderError {
+  public static readonly CODE = -32603;
+
+  constructor(message: string = "Internal error", parent?: Error) {
+    super(message, InternalError.CODE, parent);
+  }
+}
+
+// TODO: not used, should we remove it?
```
ChristopherDedominici, line 104: There are several TODOs in this PR, maybe add them in the PR description with a link to a GitHub issue so we can track them

v-next/hardhat/test/internal/builtin-plugins/network-manager/http-provider.ts
```diff
@@ -91,6 +92,48 @@ describe("http-provider", () => {
...
+      provider.on(EDR_NETWORK_REVERT_SNAPSHOT_EVENT, () => {
+        eventEmitted = true;
+      });
+
+      await provider.request({
+        method: "evm_revert",
+        params: ["0x1"],
+      });
+
+      assert.ok(eventEmitted, "The evm_revert event should be emitted");
```
ChristopherDedominici, line 134: This assertion might have concurrency issues. It could happen that the event is emitted but the variable `eventEmitted` is set to `true` after the `assert`. Maybe you can wrap the code into a promise like:

```
await new Promise<void>((resolve, reject) => {
  provider.on(EDR_NETWORK_REVERT_SNAPSHOT_EVENT, () => {
    eventEmitted = true;
    resolve();
  });

  await provider.request({
    method: "evm_revert",
    params: ["0x1"],
  });
});

assert.ok(eventEmitted, "The evm_revert event should be emitted");
```

v-next/hardhat/src/internal/builtin-plugins/network-manager/provider-errors.ts
schaable: Done!

v-next/hardhat/test/internal/builtin-plugins/network-manager/config-resolution.ts
```diff
@@ -0,0 +1,726 @@
...
+        userConfig.allowBlocksWithSameTimestamp,
+      );
+      assert.equal(
+        edrNetworkConfig.allowUnlimitedContractSize,
+        userConfig.allowUnlimitedContractSize,
+      );
+      assert.equal(
+        edrNetworkConfig.blockGasLimit,
+        // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- cast for testing
+        BigInt(userConfig.blockGasLimit as number),
```
ChristopherDedominici, line 153: Why not using the method `toBigInt` and avoid the casting?

v-next/hardhat/test/internal/builtin-plugins/network-manager/config-resolution.ts
```diff
@@ -0,0 +1,726 @@
...
+        "initialDate is not being set to the current time by default",
+      );
+      assert.equal(edrNetworkConfig.loggingEnabled, false);
+      assert.equal(edrNetworkConfig.minGasPrice, 0n);
+      assert.equal(edrNetworkConfig.networkId, 31337);
+      assert.equal(edrNetworkConfig.throwOnCallFailures, true);
+      assert.equal(edrNetworkConfig.throwOnTransactionFailures, true);
+    });
+
+    it("should use the provided chainId for the networkId if it is not provided", () => {
```
ChristopherDedominici, line 213: maybe "default" is better instead of the first "provdided"?

ChristopherDedominici: Or maybe I'm misunderstanding the message

schaable: yeah, the logic is basically: use the provided `networkId`, if it's not present, use the provided `chainId`, if it's not present, use the default value. Here we're testing that second part.

v-next/hardhat/test/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -0,0 +1,244 @@
...
+  describe("EdrProvider#request", () => {
+    it("should emit an event when the method is evm_revert", async () => {
+      let eventEmitted = false;
+      const { provider } = await hre.network.connect();
+
+      provider.on(EDR_NETWORK_REVERT_SNAPSHOT_EVENT, () => {
+        eventEmitted = true;
+      });
+
+      await provider.request({
```
ChristopherDedominici, line 34: Same comment about concurrency issue as here: https://github.com/NomicFoundation/hardhat/pull/6201/files#r1935901466

ChristopherDedominici (approved):
I left a few minor comments but a part from that LGTM 😎

v-next/hardhat/test/internal/builtin-plugins/network-manager/http-provider.ts
schaable: Done!

PR #6202, opened 2025-01-30
ci: make list-packages outputs smaller

galargh:

---

In this PR, I address the issue described here - https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1738270381236179

`list-packages` started producing too big outputs.

To address this, I:
- added the `base` parameter to the call to `dorny/paths-filter` (this way, this workflow will be detecting changed files in relation to `v-next` rather than `main; we should have had it set in the first place)
- moved the check whether `bundle` should be run or not to  inside the `list-packages` job (this way, we don't have to put all the changed files in the `list-packages` job outputs)


alcuadrado (approved):
Good catch!

PR #6203, opened 2025-01-30
chore: bump EDR to 0.8.0

nebasuke:
Bump EDR to [0.8.0](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.8.0).

This fixes the following issues:

### Minor Changes
* https://github.com/NomicFoundation/edr/commit/af26624ae0b705d70bee15d4f4b76e017f5b4dfd: Improved provider initialization performance by passing build info as buffer which avoids FFI copy overhead

### Patch Changes
* https://github.com/NomicFoundation/edr/commit/4ffb4f60ea1c49977c685dddb702445579d1f41c: fix: ignore unknown opcodes in source maps
* https://github.com/NomicFoundation/edr/commit/5c8c3ddcbd932d47d3fc8705f3e8943a45a4dfdb: Fixed crash when loading EDR on Windows without a C Runtime library installed
* https://github.com/NomicFoundation/edr/commit/2b9b805a48cd340c05ac44af189ecda752270807: Improved stack trace generation performance by eliminating one-way branching in the bytecode trie

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
@@ -12,13 +12,13 @@ import type {
   Response,
   SubscriptionEvent,
   HttpHeader,
+  TracingConfigWithBuffers,
 } from "@nomicfoundation/edr";
 import { Common } from "@nomicfoundation/ethereumjs-common";
 import picocolors from "picocolors";
 import debug from "debug";
 import { EventEmitter } from "events";
 import fsExtra from "fs-extra";
-import semver from "semver";
 
 import { requireNapiRsModule } from "../../../common/napi-rs";
 import {
@@ -33,7 +33,6 @@ import {
 import { isErrorResponse } from "../../core/providers/http";
 import { getHardforkName } from "../../util/hardforks";
 import { ConsoleLogger } from "../stack-traces/consoleLogger";
-import { FIRST_SOLC_VERSION_SUPPORTED } from "../stack-traces/constants";
 import { encodeSolidityStackTrace } from "../stack-traces/solidity-errors";
 import { SolidityStackTrace } from "../stack-traces/solidity-stack-trace";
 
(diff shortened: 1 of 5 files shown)


agostbiro (commented):
Thanks, LGTM, but I'd defer to @kanej to approve

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -506,19 +505,14 @@ export async function createHardhatNetworkProvider(
...
+): Promise<TracingConfigWithBuffers | undefined> {
   if (artifacts !== undefined) {
-    const buildInfos = [];
-
     const buildInfoFiles = await artifacts.getBuildInfoPaths();
 
     try {
-      for (const buildInfoFile of buildInfoFiles) {
-        const buildInfo = await fsExtra.readJson(buildInfoFile);
-        if (semver.gte(buildInfo.solcVersion, FIRST_SOLC_VERSION_SUPPORTED)) {
```
agostbiro, line 518: This check is now done in EDR: https://github.com/NomicFoundation/edr/blob/403bd351ded6a75adddb3879003245e4559e5354/crates/edr_solidity/src/artifacts.rs#L83

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -165,7 +164,7 @@ export class EdrProviderWrapper
   public static async create(
     config: HardhatNetworkProviderConfig,
     loggerConfig: LoggerConfig,
-    tracingConfig?: TracingConfig
+    tracingConfig?: TracingConfigWithBuffers
```
kanej, line 167: I was initially worried about this move as it is changing the API, but EdrProviderWrapper is in internal and is not exposed to the user. The only place create is invoked, uses `makeTracingConfig` which has had its return type altered as well.

Put another way this signature change looks safe to me.

kanej (approved):

PR #6205, opened 2025-01-31
feat: add accounts info printing to the hardhat node task

### LINKED ISSUE #6042: Add accounts info printing to the hardhat node task
This is a follow-up task to https://github.com/NomicFoundation/hardhat/issues/5615

It is about printing account information from the network config once it starts being resolved.

Below, you'll find a draft of the implementation.

Implementation:
```
import type {
  EdrNetworkAccountConfig,
  EdrNetworkAccountsConfig,
  GenesisAccount,
} from "../../../types/config.js";

import {
  bytesToHexString,
  hexStringToBytes,
} from "@ignored/hardhat-vnext-utils/hex";
import chalk from "chalk";
import { addr } from "micro-eth-signer";

import { derivePrivateKeys } from "../network-manager/request-handlers/handlers/accounts/derive-private-keys.js";

export function normalizeEdrNetworkConfigAccounts(
  accounts: EdrNetworkAccountsConfig,
): EdrNetworkAccountConfig[] {
  const normalizedAccounts: EdrNetworkAccountConfig[] = [];

  if (accounts !== undefined) {
    if (Array.isArray(accounts)) {
      normalizedAccounts.push(...accounts);
    } else if (accounts !== undefined) {
      const privateKeys = derivePrivateKeys(
        accounts.mnemonic,
        accounts.path,
        accounts.initialIndex,
        accounts.count,
        accounts.passphrase,
      );
      for (const privateKey of privateKeys) {
        normalizedAccounts.push({
          privateKey: bytesToHexString(privateKey),
          balance: accounts.accountsBalance,
        });
      }
    }
  }

  return normalizedAccounts;
}

export function printEdrNetworkConfigAccounts(
  accounts: Array<EdrNetworkAccountConfig | GenesisAccount>,
): void {
  if (accounts.length === 0) {
    return;
  }

  console.log("Accounts");
  console.log("========");

  // NOTE: In v2, we were printing the warning only if the default config was used.
  console.log();
  printPublicPrivateKeysWarning();
  console.log();

  for (const [index, account] of accounts.entries()) {
    const address = addr
      .fromPrivateKey(hexStringToBytes(account.privateKey))
      .toLowerCase();
    const balance = (BigInt(account.balance) / 10n ** 18n).toString(10);

    console.log(`Account #${index}: ${address} (${balance} ETH)`);
    // TODO: Should we print the private key as well?
    // console.log(`Private Key: ${account.privateKey}`);

    console.log();
  }

  // NOTE: In v2, we were printing the warning only if the default config was used.
  printPublicPrivateKeysWarning();
  console.log();
}

// NOTE: In v2, we were printing the warning only if the default config was used.
// Because of that we were certain that the printed accounts were publicly known.
function printPublicPrivateKeysWarning(): void {
  console.log(
    chalk.bold(
      "WARNING: Funds sent on live network to accounts with publicly known private keys WILL BE LOST.",
    ),
  );
}
```

Usage:
```
import { normalizeEdrNetworkConfigAccounts, printEdrNetworkConfigAccounts } from "./helpers.js";

...

  printEdrNetworkConfigAccounts(normalizeEdrNetworkConfigAccounts(networkConfig.accounts))

```

galargh:

---

Closes #6042

In this PR, I add accounts' info printing to the node task. This is behaviour equivalent to this seen in Hardhat v2.


v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -80,9 +83,16 @@ export const DEFAULT_EDR_NETWORK_HD_ACCOUNTS_CONFIG_PARAMS: EdrNetworkDefaultHDA
...
-  return deepEqual(account, DEFAULT_EDR_NETWORK_HD_ACCOUNTS_CONFIG_PARAMS);
+  return deepEqual(
+    {
+      ...accounts,
+      mnemonic: await accounts.mnemonic.get(),
+      passphrase: await accounts.passphrase.get(),
+    },
+    DEFAULT_EDR_NETWORK_HD_ACCOUNTS_CONFIG_PARAMS,
+  );
 }
```
galargh, line 96: I had to update this function as before, it would always return false since the types of `menominic` and `passphrase` were different in `EdrNetworkHDAccountsConfig` and `DEFAULT_EDR_NETWORK_HD_ACCOUNTS_CONFIG_PARAMS`.

schaable: Uh, I've addressed this differently in a PR that was merged today. Instead of delegating the resolution of the config variables to the compare function, I used the getters before calling the function. You can check it out here: [link](https://github.com/NomicFoundation/hardhat/blob/a3e1f123be84a06e0406631a7a36c6c05909323b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts#L258). This is temporary though, as EDR network accounts may not use config variables in the future. See https://github.com/NomicFoundation/hardhat/issues/6210

v-next/hardhat/src/internal/builtin-plugins/node/helpers.ts
```diff
@@ -0,0 +1,58 @@
...
+import chalk from "chalk";
+import { addr } from "micro-eth-signer";
+
+import { isDefaultEdrNetworkHDAccountsConfig } from "../network-manager/edr/edr-provider.js";
+import { normalizeEdrNetworkAccountsConfig } from "../network-manager/edr/utils/convert-to-edr.js";
+
+export async function printEdrNetworkConfigAccounts(
+  config: EdrNetworkAccountsConfig,
+): Promise<void> {
+  const accounts = await normalizeEdrNetworkAccountsConfig(config);
```
schaable, line 13: @alcuadrado this will need to be resolved differently if we move the derivation process to EDR

v-next/hardhat/src/internal/builtin-plugins/node/task-action.ts
```diff
@@ -126,7 +130,14 @@ const nodeAction: NewTaskActionFunction<NodeActionArguments> = async (
 
   // TODO(https://github.com/NomicFoundation/hardhat/issues/6040): Add build info watcher here
 
-  // TODO(https://github.com/NomicFoundation/hardhat/issues/6042): Add accounts info printing here
+  // NOTE: Before creating the node, we check if the input network config is of type edr.
+  // We only proceed if it is. Hence, we can assume that the outputnetwork config is of type edr as well.
```
schaable, line 134:
```suggestion
  // We only proceed if it is. Hence, we can assume that the output network config is of type edr as well.
```

schaable (commented):
In general looks good. I left some comments. It would be good to have some basic tests as well.

v-next/hardhat/src/internal/builtin-plugins/node/helpers.ts
alcuadrado: Can you add it to your issue?

alcuadrado: good point btw, I hadn't noticed that

alcuadrado: Actually, this is an interactive task that prints to the console (printing is slow), where 200-300ms may not matter.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
galargh: I see! Yeah, that works too. Would you prefer for the resolution to stay outside `isDefaultEdrNetworkHDAccountsConfig`? Personally, I like the fact that by moving it to the function, we can type the input argument.

schaable: Either way it's fine. If you're going to keep this implementation, could you rebase and update the other call?

v-next/hardhat/src/internal/builtin-plugins/node/helpers.ts
schaable: Agreed, but seems weird to keep the code in the frontend just for this logic. Btw the issue to move the derivation to EDR was going to be created by @nebasuke

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
galargh: Of course! And thank you for the review 🙇 I'll add a test for this as well.

alcuadrado:
There are a few stylistic changes missing:

- An empty line between each account
- An empty line between the warning and the accounts
- The address and private key should be aligned. That is: the `"Account #${n}:"` and `"Private Key:"` messages require right padding

alcuadrado (commented):
Looking good except for some changes that are needed to make the output readable.

galargh:
> There are a few stylistic changes missing:
> 
> * An empty line between each account
> * An empty line between the warning and the accounts
> * The address and private key should be aligned. That is: the `"Account #${n}:"` and `"Private Key:"` messages require right padding

I lost the newlines when I made the helper function return string. I'm on it!

As for the alignment, do you mean something like this?
```
Account 1:   <address>
Private Key: <key>

...

Account 10: <address>
Private Key: <key>
```

PR #6208, opened 2025-01-31
Exploration task: identify where to place the new hooks for the build system

ChristopherDedominici:
This PR is currently on hold. We need to sync up to decide which hooks to keep and where they should go


kanej:
I am going to close this PR as the build system has gone through significant changes since it was put together - we will add hooks on a case by case basis.

PR #6212, opened 2025-01-31
fix: error when re-using the solc settings object

### LINKED ISSUE #6204: Error when re-using the solc settings object
Re-using the solc settings object for versions that default to a different EVM version results in a very obscure error:

```
Invalid EVM version requested.
```

**Reproduction steps**

Create a Hardhat config with this content:

```js
const compilerSettings = {}

module.exports = {
  solidity: {
    compilers: [
      { settings: compilerSettings, version: '0.5.17' },
      { settings: compilerSettings, version: '0.8.28' },
    ]
  }
};
```

And add two Solidity files:

```solidity
// contracts/Foo.sol
// SPDX-License-Identifier: Unlicense
pragma solidity ^0.5.0;

contract Foo {}
```
```solidity
// contracts/Bar.sol
// SPDX-License-Identifier: Unlicense
pragma solidity ^0.8.0;

contract Bar {}
```

Then run `hh compile`.

Notice that replacing the config with this makes the issue go away:

```js
module.exports = {
  solidity: {
    compilers: [
      { settings: {}, version: '0.5.17' },
      { settings: {}, version: '0.8.28' },
    ]
  }
};
```

So it's very likely we are mutating something we shouldn't.

Thanks @shark0der for reporting this!

galargh:

---

Closes #6204

This fix ensures that we don't mutate the `comiler.settings` when resolving the compiler.

### DIFF
--- a/.changeset/witty-clouds-compete.md
+++ b/.changeset/witty-clouds-compete.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix error when re-using the solc settings object
--- a/packages/hardhat-core/src/internal/core/config/config-resolution.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-resolution.ts
@@ -401,7 +401,7 @@ function normalizeSolidityConfig(
 function resolveCompiler(compiler: SolcUserConfig): SolcConfig {
   const resolved: SolcConfig = {
     version: compiler.version,
-    settings: compiler.settings ?? {},
+    settings: { ...(compiler.settings ?? {}) },
   };
 
   if (semver.gte(resolved.version, "0.8.20")) {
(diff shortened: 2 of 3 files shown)


galargh:
I'm putting it back in draft because I don't fully understand what's going on.

We do a deep clone here - https://github.com/NomicFoundation/hardhat/pull/6212/commits/6a341bd254c8436cb4a93681f3c52b0015bc65b2

That's way before solidity config resolution.

But then, if we have multiple compilers configured and they reuse the same settings object, then there is still only one settings object being mutated by each call to resolveConfig, even after the deep clone. It seems like the deep clone doesn't actually clone the settings separately. So if they were pointing at the same memore block before, they will still do so after the deep clone.

I added a test case which can help showcase that - https://github.com/NomicFoundation/hardhat/pull/6212/commits/6a341bd254c8436cb4a93681f3c52b0015bc65b2

This still solves the issue at hand, so we could merge it if we want the exact described problem fixed, but I think we might have to do more changes around how we perform user config deep clone in general.

PR #6213, opened 2025-01-31
Clean up and implement the ArtifactManager interface

alcuadrado:
While looking at a part of the codebase I noticed that we never addressed tons of TODOs in the `ArtifactManager` interface.

This PR does that, cleaning up the interface, and updating the different implementations.

### DIFF
--- a/v-next/hardhat-ethers/test/helpers/artifact-manager-mock.ts
+++ b/v-next/hardhat-ethers/test/helpers/artifact-manager-mock.ts
@@ -67,7 +67,7 @@ export class MockArtifactsManager implements ArtifactsManager {
     });
   }
 
-  public async getAllFullyQualifiedNames(): Promise<string[]> {
+  public async getAllFullyQualifiedNames(): Promise<ReadonlySet<string>> {
     throw new HardhatError(HardhatError.ERRORS.INTERNAL.NOT_IMPLEMENTED_ERROR, {
       message: "Not implemented in MockArtifactsManager",
     });
@@ -81,7 +81,7 @@ export class MockArtifactsManager implements ArtifactsManager {
     });
   }
 
-  public async getAllBuildInfoIds(): Promise<string[]> {
+  public async getAllBuildInfoIds(): Promise<ReadonlySet<string>> {
     throw new HardhatError(HardhatError.ERRORS.INTERNAL.NOT_IMPLEMENTED_ERROR, {
       message: "Not implemented in MockArtifactsManager",
     });
@@ -98,6 +98,12 @@ export class MockArtifactsManager implements ArtifactsManager {
   public async getBuildInfoOutputPath(
     _buildInfoId: string,
   ): Promise<string | undefined> {
+    throw new HardhatError(HardhatError.ERRORS.INTERNAL.NOT_IMPLEMENTED_ERROR, {
+      message: "Not implemented in MockArtifactsManager",
+    });
+  }
+
+  public async clearCache(): Promise<void> {
     throw new HardhatError(HardhatError.ERRORS.INTERNAL.NOT_IMPLEMENTED_ERROR, {
       message: "Not implemented in MockArtifactsManager",
     });
--- a/v-next/hardhat-ethers/test/helpers/artifact-manager-mock.ts
+++ b/v-next/hardhat-ethers/test/helpers/artifact-manager-mock.ts
@@ -81,7 +81,7 @@ export class MockArtifactsManager implements ArtifactsManager {
     });
   }
 
-  public async getBuildInfoIds(): Promise<string[]> {
+  public async getAllBuildInfoIds(): Promise<string[]> {
     throw new HardhatError(HardhatError.ERRORS.INTERNAL.NOT_IMPLEMENTED_ERROR, {
       message: "Not implemented in MockArtifactsManager",
     });
--- a/v-next/hardhat-ethers/test/helpers/artifact-manager-mock.ts
+++ b/v-next/hardhat-ethers/test/helpers/artifact-manager-mock.ts
@@ -1,13 +1,8 @@
 import type {
   Artifact,
   ArtifactsManager,
-  BuildInfo,
   GetAtifactByName,
 } from "@ignored/hardhat-vnext/types/artifacts";
-import type {
-  CompilerInput,
-  CompilerOutput,
-} from "@ignored/hardhat-vnext/types/solidity";
 
 import { HardhatError } from "@ignored/hardhat-vnext-errors";
 
@@ -28,6 +23,10 @@ export class MockArtifactsManager implements ArtifactsManager {
         this.#artifactsPaths.set(artifactName, fileName);
       }
     }
+  }
+
+  public async saveArtifact(artifact: Artifact): Promise<void> {
+    this.#artifacts.set(artifact.contractName, artifact);
   }
 
   public async readArtifact<ContractNameT extends string>(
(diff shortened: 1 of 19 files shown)


alcuadrado:
I think we should add way more test to the `ArtifactsManager` though. While it's simple, it has some path manipulation, which is easy to get wrong in windows

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifacts-manager.ts
```diff
@@ -1,318 +1,405 @@
 import type {
   ArtifactsManager,
   GetAtifactByName,
-  BuildInfo,
 } from "../../../types/artifacts.js";
 
+import { EOL } from "node:os";
```
schaable, line 6: Didn't know about this. We should probably start using it everywhere.

v-next/hardhat/test/test-helpers/mock-artifacts-manager.ts
```diff
@@ -17,6 +16,10 @@ export class MockArtifactsManager implements ArtifactsManager {
     this.#artifacts = new Map();
   }
 
+  public async saveArtifact(artifact: Artifact): Promise<void> {
```
schaable, line 19: same as the other file

schaable (commented):
Left a few comments

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifacts-manager.ts
alcuadrado: TBH I only used it because it was there. I think node translates `\n` to EOL everywhere.

v-next/hardhat-ethers/test/helpers/artifact-manager-mock.ts
```diff
@@ -30,6 +25,10 @@ export class MockArtifactsManager implements ArtifactsManager {
     }
   }
 
+  public async saveArtifact(artifact: Artifact): Promise<void> {
```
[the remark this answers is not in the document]
alcuadrado: This is specific to the mock, it's not part of the artifact manager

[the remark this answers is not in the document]
alcuadrado: `async` or not doesn't matter in this case tbh. I just left it as it was when I found it.

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifacts-manager.ts
```diff
@@ -1,318 +1,405 @@
...
+ * same time, using this interface to organize the data.
  */
-export class ArtifactsManagerImplementation implements ArtifactsManager {
+interface FsData {
+  bareNameToFullyQualifiedNameMap: Map<string, ReadonlySet<string>>;
+  allArtifactPaths: ReadonlySet<string>;
+  allFullyQualifiedNames: ReadonlySet<string>;
+}
+
+export class ArticlesManagerImplementation implements ArtifactsManager {
```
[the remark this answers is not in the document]
alcuadrado: 😳

schaable (approved):
Pre-approving, the only required change would be the typo 😄

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifacts-manager.ts
```diff
@@ -1,318 +1,405 @@
...
-   * FQN `contracts/Foo.sol:Bar`
+   * @param fullyQualifiedName The fully qualified name of the contract whose
+   * artifact is being requested.
+   * @returns The path to the artifact, which may or may not exist.
    */
-  #getFullyQualifiedNameFromPath(absolutePath: string): string {
-    const sourceName = this.#replaceBackslashes(
-      path.relative(this.#artifactsPath, path.dirname(absolutePath)),
+  #getArtifactPathFromFullyQualifiedName(fullyQualifiedName: string): string {
+    // TODO: Cache this?
```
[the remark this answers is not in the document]
alcuadrado: Actually, caching them simplifies the code. I just did it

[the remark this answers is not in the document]
alcuadrado: See dad80ea908fe0012437a8f268c3c9ad7673772da

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifacts-manager.ts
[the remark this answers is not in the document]
alcuadrado: See 90bea42e8717ab80a00c57e77b67f99675a876b2

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
```diff
@@ -0,0 +1,405 @@
...
+    const relativePath = path.relative(this.#artifactsPath, artifactPath);
+    const sourceName = path.dirname(relativePath).split(path.sep).join("/");
+    const contractName = path.basename(relativePath, ".json");
+    return `${sourceName}:${contractName}`;
+  }
+
+  #getSimilarContractNames(
+    contractNameOrFullyQualifiedName: string,
+    bareNameToFullyQualifiedNameMap: Map<string, ReadonlySet<string>>,
+  ): string[] {
```
galargh, line 214: I found it a little confusing that this function can also return similar fully qualified names. Would it make sense to split it in two? `getSimilarContractNames` and `getSimilarFullyQualifiedNames`?

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
```diff
@@ -0,0 +1,405 @@
...
+      case 0:
+        return "";
+      case 1:
+        return `Did you mean "${similarNames[0]}"?`;
+      default:
+        return `We found some that were similar:
+
+${similarNames.map((n) => `  * ${n}`).join(EOL)}
+
+Please replace "${contractNameOrFullyQualifiedName}" for the correct contract name wherever you are trying to read its artifact.
```
galargh, line 245: Could we check if we got contract name or fully qualified name here and update the message accordingly? I think it's going to be easier to communicate with users that way.

Also, I don't think this error is exclusively related to reading artifacts so maybe it would be a good idea to make the messaging here more general.

galargh (approved):
I left a few comments here and there. Nothing blocking, though. Great work 🏆 

Since GitHub decided not to display the diff for `artifact-manager.ts` based on its' predecessor `artifacts-manager.ts`, I reviewed the former from scratch and ignored the latter.

I also only just skimmed through the levenshtein distance calculation as I'm assuming the source is trustworthy.

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
```diff
@@ -0,0 +1,405 @@
+import type {
+  ArtifactManager,
+  GetAtifactByName,
+} from "../../../types/artifacts.js";
+
+import { EOL } from "node:os";
```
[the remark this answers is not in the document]
galargh: Duplicate - https://github.com/NomicFoundation/hardhat/pull/6213#discussion_r1939431497

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
alcuadrado: Great point! I changed into a slightly different approach in 4c8e874afaedd9e593050eb8678f9891a8c922c8, which I believe is simpler.

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
alcuadrado: I changed it to reflect the contract name type.

I'd be ok with leaving "reading" as the action here. But open to better alternatives.

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
galargh: Lovely! Yes, that's great!

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
galargh: > I changed it to reflect the contract name type.

Amazing! Thanks!

> I'd be ok with leaving "reading" as the action here. But open to better alternatives.

To be fair, now, reading it again, I think "read" makes sense here after all. Sorry for the noise.

galargh (approved):

PR #6214, opened 2025-01-31
Enable stack traces in the EDR provider

### LINKED ISSUE #5981: EDR not getting compilation results in v3
We need to provide compilation results to EDR to allow it to provide enhanced tracing information as Hardhat 2 does.

There are multiple elements to consider:

* recompilations leading to new results
* multiple connections each needed updated
* resource allocation/sharing

alcuadrado:
While cleaning up the `ArtifactsManager` interface in #6213 I noticed that we may have a misalignment between the build info formats that EDR was expecting and the ones we generated, so I took a stab at this.

This PR those a few things:

* Sets the correct types in our own EDR `node-types`, as for some reason EDR is using some `any`s.
* It loads and converts all the build infos into the expected format.
* Adds the artifacts plugin as a dependency of the network manager plugin, so that it can load the build infos
* Fixes an issue related to how we were creating the solc inputs.

While this closes #5981, there's going to be a bit of follow up work.

1. Once Agost releases a new version of `@ignored/edr`, we should do something similar for the solidity tests. Except without transforming the build infos into HH2 build infos.
2. Once Remco adapts the interface of `@ignored/edr-optimism`, we should stop parsing and transforming the build infos in the `NetworkManager`.
3. We should test this integration, as right now its only "tested" in the example project.


alcuadrado:
@sz-piotr as I changed how the solc inputs are generated, some build ids changed, and that broke a test. Can you check if my fix is correct?

v-next/example-project/test/node/example-test.ts
```diff
@@ -22,4 +22,31 @@ describe("Example EDR based test", () => {
...
+
+    const Revert = await connection.ethers.getContractFactory("Revert");
+    const revert = await Revert.deploy([]);
+
+    try {
+      await revert.boom();
+    } catch (error) {
+      assert.ok(error instanceof Error, "Expected an error to be thrown");
+
+      assert.ok(
```
schaable, line 37: no need to change it, but there's an [assert.match](https://nodejs.org/api/assert.html#assertmatchstring-regexp-message) that better captures this type of assertion.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -204,6 +214,10 @@ export class NetworkManagerImplementation implements NetworkManager {
             chainType: resolvedChainType as ChainType,
           },
           jsonRpcRequestWrapper,
+          tracingConfig: {
```
schaable, line 217: have you seen any performance hit after adding this? I thought it would be disabled by default and enabled through config.

schaable (approved):
Looks good, but left a question regarding performance.

v-next/example-project/test/node/example-test.ts
alcuadrado: Thanks! I looked for something with "includes" semantics and couldn't find any 😅

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
alcuadrado: It's meant to be enabled by default, so if there's any performance issue we should address it.

Right now it will have a small performance problem, which is being address in EDR. Note that we are parsing and transforming the build infos, just to pass it to EDR.

In the near future we'll just pass the `Uint8Array` read from the file, directly.

galargh (approved):
Looks good! In https://github.com/NomicFoundation/hardhat/commit/7ebad19444c213687c98581593cc22528e797f22 (which is part of https://github.com/NomicFoundation/hardhat/pull/6129), I refactored the `build-system.ts` as I found it very annoying up to update.

Having said that, we can go ahead with these changes first, and I'll deal with resolving the conflicts.

PR #6216, opened 2025-02-02
Fix regex pattern for bigint parameter validation

PixelPil0t1:
In file `docs/src/content/ignition/docs/guides/deploy.md`:

Changed:
- `/d+n/` → `/\d+n/`

Reason for change:
The original regex pattern `/d+n/` would incorrectly match one or more literal 'd' characters followed by 'n'. The corrected pattern `/\d+n/` properly matches one or more digits followed by 'n', which is the correct format for bigint literals in JavaScript (e.g. "1000n").

This fix ensures the documentation accurately describes how to specify bigint values in parameter files.


alcuadrado (approved):
Thanks!

PR #6223, opened 2025-02-04
Prague support

schaable:
TODO:
 - [x] Bump EDR.
 - [x] Fix TODOs in `packages/hardhat-core/src/internal/hardhat-network/provider/utils/convertToEdr.ts`.
 - [ ] Bump Ethers once https://github.com/ethers-io/ethers.js/issues/4916 is released and update `formatTransactionResponse` in `packages/hardhat-ethers/src/internal/ethers-utils.ts` to include the `authorizationList`. [Not-a-blocker]
 - [ ] Bump all `@ledgerhq` dependencies in `hardhat-ledger` once signing type 4 transactions is supported by ledger. [Not-a-blocker]
 
 Follow-up PR for setting prague as the default hardfork (to be merged once prague is activated in mainnet): https://github.com/NomicFoundation/hardhat/pull/6237


packages/hardhat-core/src/internal/core/providers/accounts.ts
```diff
@@ -249,54 +246,118 @@ export class LocalAccountsProvider extends ProviderWrapperWithChainId {
...
+        "maxFeePerGas should be defined"
       );
+
+      transaction = Transaction.prepare({
+        type: "eip7702",
+        to: checksummedAddress,
+        nonce: txData.nonce,
+        chainId: txData.chainId ?? normalizeToBigInt(chainId),
+        value: txData.value !== undefined ? txData.value : 0n,
+        data: txData.data !== undefined ? bytesToHex(txData.data) : "",
```
fvictorio, line 309: Same for some fields below.

```suggestion
        value: txData.value ?? 0n,
        data: txData.data ?? "",
```

fvictorio (approved):
I think `x ?? y` is always more readable than `x !== undefined ? x : y`, although I know that the semantics are different. But I'm pretty sure in the cases where the latter pattern is used in this PR, turning it into "undefined or null" (the `??` semantics) would be fine.

packages/hardhat-core/src/internal/core/providers/accounts.ts
```diff
@@ -249,54 +246,114 @@ export class LocalAccountsProvider extends ProviderWrapperWithChainId {
...
+          nonce,
+          yParity: bytesToInt(yParity),
+          r: bytesToBigInt(r),
+          s: bytesToBigInt(s),
+        };
+      }
+    );
+
+    let checksummedAddress;
+    if (txData.to === undefined || txData.to === null) {
```
alcuadrado, line 281: Note that, as in v3, we should fail if there's no address and there's no data either, because that's not supported by `micro-eth-signer`.

In theory it is supported by ethereum, but it's almost certainly a user error.

packages/hardhat-core/src/internal/core/providers/accounts.ts
```diff
@@ -117,11 +117,7 @@ export class LocalAccountsProvider extends ProviderWrapperWithChainId {
       // if we don't manage the address, the method is forwarded
       const privateKey = this._getPrivateKeyForAddressOrNull(address);
       if (privateKey !== null) {
-        return signTypedData({
-          privateKey,
-          version: SignTypedDataVersion.V4,
-          data: typedMessage,
-        });
+        return signTyped(typedMessage, privateKey, false);
```
alcuadrado, line 120: Are we sure we want a false here?

packages/hardhat-core/src/internal/core/providers/accounts.ts
```diff
@@ -249,54 +246,114 @@ export class LocalAccountsProvider extends ProviderWrapperWithChainId {
...
+        chainId: txData.chainId ?? normalizeToBigInt(chainId),
+        value: txData.value !== undefined ? txData.value : 0n,
+        data: txData.data !== undefined ? bytesToHex(txData.data) : "",
+        gasPrice: txData.gasPrice ?? 0n,
+        gasLimit: txData.gasLimit,
+      });
     }
 
-    const signedTransaction = transaction.sign(privateKey);
+    const signedTransaction = transaction.signBy(privateKey, false);
```
alcuadrado, line 354: Same with the false here

alcuadrado:
I think we can merge this, and then we revisit the extra entropy thing.

packages/hardhat-core/src/internal/core/providers/accounts.ts
alcuadrado: I also find it better, unless `null` has some special meaning here.

packages/hardhat-core/src/internal/core/providers/accounts.ts
alcuadrado: This can be addressed later though

packages/hardhat-core/src/internal/core/providers/accounts.ts
schaable: I believe we do. Using `extraEntropy` will make the signature different every time, which not only messes up our tests but could also affect user tests and end up being a breaking change.

alcuadrado (approved):

packages/hardhat-core/src/internal/core/providers/accounts.ts
alcuadrado: Let's merge with `false`. I created a follow up.

PR #6225, opened 2025-02-04
feat: hook up global exception handler in Hardhat 3

### LINKED ISSUE #6222: Hook up global exception handler in Hardhat 3
Sentry services have been added in Hardhat 3. However we have not hooked up the node global exception handler for the command line so that exceptions are reported to Sentry.
We want to take across the same capability as exists in Hardhat 2.

## TODO

- [x] Review the approach to global exception handling in HH2
- [x] Port the approach to HH3
- [x] Check with the test Sentry environment that exceptions are coming through

galargh:

---

Closes #6222

TODOs:
- [x] I hooked up the Sentry reporter in the catch block inside the CLI 
- [x] I verified the messages can be sent to Sentry using the setup we have in the `v-next` branch - https://nomic-labs.sentry.io/issues/?project=4507685793103872&query=is%3Aresolved&referrer=issue-list&statsPeriod=1h&utc=true
- [x] I checked that EDR panics, as of now, are **NOT** send to Sentry because they are not raised by Hardhat.
    - See https://github.com/NomicFoundation/hardhat/pull/6225#discussion_r1941965616 for more details.
- [x] I uncommented the codepath that ensures the Sentry reporter is flushed before we continue.

Follow-up work items:
- [ ] Set up alerting to Slack
    - Only Sentry team admins/moderators can set up alerts and I am not one. 
    - The alerting rule should be added here: https://nomic-labs.sentry.io/alerts/rules/?project=4507685793103872&statsPeriod=1h&utc=true
    - It should be similar to the one we have configured for Hardhat v2: https://nomic-labs.sentry.io/alerts/rules/buidler/1079337/details/

### DIFF
--- a/v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
+++ b/v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
@@ -24,7 +24,7 @@ export async function sendErrorTelemetry(
   configPath: string = "",
 ): Promise<boolean> {
   const instance = await Reporter.getInstance();
-  return instance.reportError(error, configPath);
+  return instance.reportErrorViaSubprocess(error, configPath);
 }
 
 // ATTENTION: this function is exported for testing, do not directly use it in production
@@ -86,7 +86,7 @@ class Reporter {
     this.#instance = undefined;
   }
 
-  public async reportError(
+  public async reportErrorViaSubprocess(
     error: Error,
     configPath: string = "",
   ): Promise<boolean> {
--- a/v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
+++ b/v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
@@ -2,6 +2,7 @@ import {
   HardhatError,
   HardhatPluginError,
 } from "@ignored/hardhat-vnext-errors";
+import { flush } from "@sentry/node";
 import debug from "debug";
 
 import { ProviderError } from "../../../builtin-plugins/network-manager/provider-errors.js";
@@ -9,7 +10,6 @@ import { getHardhatVersion } from "../../../utils/package.js";
 import { isTelemetryAllowed } from "../telemetry-permissions.js";
 
 import { getSubprocessTransport } from "./transport.js";
-import { flush } from "@sentry/node";
 
 const log = debug("hardhat:cli:telemetry:sentry:reporter");
 
@@ -104,7 +104,9 @@ class Reporter {
     captureException(error);
 
     // NOTE: Alternatively, we could close the reporter when we exit the process.
-    await flush(50);
+    if (!(await flush(50))) {
+      log("Failed to flush events");
+    }
 
     return true;
   }
--- a/v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
+++ b/v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
@@ -9,6 +9,7 @@ import { getHardhatVersion } from "../../../utils/package.js";
 import { isTelemetryAllowed } from "../telemetry-permissions.js";
 
 import { getSubprocessTransport } from "./transport.js";
+import { flush } from "@sentry/node";
 
 const log = debug("hardhat:cli:telemetry:sentry:reporter");
 
@@ -102,8 +103,8 @@ class Reporter {
 
     captureException(error);
 
-    // TODO: should we run it? Hook at the end of hardhat execution?
-    // await flush(50);
+    // NOTE: Alternatively, we could close the reporter when we exit the process.
+    await flush(50);
 
     return true;
   }
--- a/v-next/hardhat/src/internal/cli/telemetry/sentry/subprocess.ts
+++ b/v-next/hardhat/src/internal/cli/telemetry/sentry/subprocess.ts
@@ -43,8 +43,6 @@ try {
   const anonymizedEvent = await anonymizer.anonymize(sentryEvent);
 
   if (anonymizedEvent.success) {
-    // NOTE: It is quite surprising to find this check here.
-    // Should we move it to the reporter's #shouldBeReported method?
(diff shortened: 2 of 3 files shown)


v-next/hardhat/src/internal/cli/telemetry/sentry/subprocess.ts
```diff
@@ -43,6 +43,8 @@ try {
   const anonymizedEvent = await anonymizer.anonymize(sentryEvent);
 
   if (anonymizedEvent.success) {
+    // NOTE: It is quite surprising to find this check here.
+    // Should we move it to the reporter's #shouldBeReported method?
     if (anonymizer.raisedByHardhat(anonymizedEvent.event)) {
       await sendEventToSentry(anonymizedEvent.event);
     }
```
galargh, line 50: Because of this check, panics originating in EDR won't be send to Sentry. An example of `anonymizedEvent.event` that represents an EDR panic that fails the `anonymizer.raisedByHardhat` check:

```json
{
  "event_id": *,
  "platform": "node",
  "timestamp": *,
  "extra": {
    "nodeVersion": "v22.10.0",
    "hardhatVersion": "3.0.0-next.16"
  },
  "exception": {
    "values": [
      {
        "type": "Error",
        "value": "task 20 panicked",
        "stacktrace": {
          "frames": []
        }
      }
    ]
  }
}
```

The panic was created by changing the log handler in the EDR provider to:

```rust
use crate::{data::ProviderData, spec::ProviderSpec, time::TimeSinceEpoch, ProviderError};

pub fn handle_set_logging_enabled_request<
    ChainSpecT: ProviderSpec<TimerT>,
    TimerT: Clone + TimeSinceEpoch,
>(
    data: &mut ProviderData<ChainSpecT, TimerT>,
    is_enabled: bool,
) -> Result<bool, ProviderError<ChainSpecT>> {
    panic!("Forcing a panic for demonstration purposes");

    data.logger_mut().set_is_enabled(is_enabled);
    Ok(true)
}
```

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -102,8 +103,10 @@ class Reporter {
 
     captureException(error);
 
-    // TODO: should we run it? Hook at the end of hardhat execution?
-    // await flush(50);
+    // NOTE: Alternatively, we could close the reporter when we exit the process.
+    if (!(await flush(50))) {
```
alcuadrado, line 107: This function is actually reporting the error via a subprocess. In the subprocess we do flush, so I believe it's unnecessary to flush here, and it may slow things down (i.e. HH feeling hung after printing an error).

@schaable @fvictorio can you confirm?

alcuadrado: I guess the only thing that may make flushing needed is if `captureException` enqueues the subprocess creation.

If that's the case, let's flush for now, but maybe also change our strategy soon?

alcuadrado: Also, what about renaming to `reportErrorViaSubprocess`

galargh: > I guess the only thing that may make flushing needed is if captureException enqueues the subprocess creation.

Yes, that's what I was thinking. 

> Also, what about renaming to reportErrorViaSubprocess

Sounds good, I'll rename it straight away.

v-next/hardhat/src/internal/cli/telemetry/sentry/subprocess.ts
alcuadrado: I guess we can start by sending everything that matches `/panicked$/`, and we refine it further if it gets too spammy (but i think we'd be fine).

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
alcuadrado: > This function is actually reporting the error via a subprocess. In the subprocess we do flush, so I believe it's unnecessary to flush here, and it may slow things down (i.e. HH feeling hung after printing an error).

Let's leave it for now, but I think ti would be great to consider taking more control from Sentry so that the workflow is clear.

galargh: Sounds good, in case we want to revisit it, I wrote down what happens on the sentry side of things when we flush the client.

This is the top-level flush implementation: https://github.com/getsentry/sentry-javascript/blob/aa8704a59639d1226ff9a5e5fa97f3f00bc83a9f/packages/core/src/exports.ts#L210

This is the client's flush implementation: https://github.com/getsentry/sentry-javascript/blob/aa8704a59639d1226ff9a5e5fa97f3f00bc83a9f/packages/node/src/sdk/client.ts#L61

We don't initialize tracing so we'd only end up calling the flush from super there which is: https://github.com/getsentry/sentry-javascript/blob/develop/packages/core/src/client.ts#L295

Here, we wait for the client to be done processing messages (they might be being handled by one of the integrations for examle): https://github.com/getsentry/sentry-javascript/blob/develop/packages/core/src/client.ts#L802

And once that's done, we flush the transport layer i.e. wait for all the messages to leave it's buffer i.e. get passed down to the subprocess: https://github.com/getsentry/sentry-javascript/blob/aa8704a59639d1226ff9a5e5fa97f3f00bc83a9f/packages/core/src/transports/base.ts#L39

v-next/hardhat/src/internal/cli/telemetry/sentry/subprocess.ts
galargh: @nebasuke is currently looking into it more comprehensively as per https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1738753918142849?thread_ts=1738747172.375189&cid=C03P6B72ZHU

So I'm going to leave it as-is for this PR, and we'll handle that in the follow-up.

PR #6227, opened 2025-02-05
ignition: port ignition plugin

kanej:
This migrates `hardhat-ignition` the Ignition Hardhat plugin to Hardhat 3.

image

The following tasks have been ported:

- [x] deploy
- [x] status
- [x] deployments
- [x] transactions
- [x] wipe
- [x] visualize
- [ ] ~verify~

The verify task shows a message that it has not been implemented:

image

### Running Manually

The Ignition plugin has been added to the example project, along with an Ignition module. You can run a deploy with:

```shell
pnpm hardhat3 ignition deploy ./ignition/modules/Apollo.ts
```

### Tests

The tests are all passing. Three are skipped:

* `should deploy user interfered transaction on second run` - which is skipped in Hardhat 2 Ignition
* two verify tests

## Known bugs

There is a known bug in the visualize report. This will be addressed in a follow up PR.

- [ ] Text is not appearing on visualize mermaid diagram

## Review points

My apologies to the reviewer. The commits are targeted. Moving through them one at a time should be possible.

The commits split down into:

1. Code formatting and linting updates to the latest eslint/typescript/prettier
2. Many individual commits turn on tests one test a time
3. A small number of updates to the logic of the plugin, these have been individually commented upon - see my comments for the details.

## Follow-up tasks

- [ ] Rework Ignition UI to be based off of the HH3 hook api
- [ ] Rewire the ledger checks to work with user interruptions
- [ ] Review config setup
- [ ] Revisit test fixture setup - can we improve on performance with precompiled artifacts
- [ ] Deal with task default values like deploymentId where currently we are using `""` to represent no option given by user
- [ ] Bring back `verify` task
- [ ] Convert tests to Node Test Runner
- [ ] Get rid of `hardhat_setLedgerOutputEnabled - Method not supported` message on the node
- [ ] Re-enable all eslint rules
- [ ] Add an integration test to ensure that if the user sets the ignition path, that is coming through in the config
- [ ] Revisit the visualization test - it requires a vite build of the ui which significantly slows down the dev flow of testing


v-next/hardhat/src/internal/core/hre.ts
```diff
@@ -205,6 +205,7 @@ async function resolveUserConfig(
     plugins: sortedPlugins,
     tasks: config.tasks ?? [],
     paths: resolvePaths(projectRoot, configPath, config.paths),
+    defaultNetwork: config.defaultNetwork ?? "hardhat",
```
kanej, line 208: This needs confirmed. I suspect this is a fix for a bug whereby the `defaultNetwork` from config was being ignored and always set to `hardhat`.

v-next/hardhat/src/internal/core/plugins/detect-plugin-npm-dependency-problems.ts
```diff
@@ -78,7 +78,7 @@ export async function detectPluginNpmDependencyProblems(
 
     const { satisfies } = await import("semver");
 
-    if (!satisfies(installedVersion, versionSpec)) {
+    if (!satisfies(installedVersion, versionSpec.replace("workspace:", ""))) {
```
kanej, line 81: Satisfies failed on the version spec having the network prefix.

v-next/hardhat-ignition/src/internal/tasks/transactions.ts
```diff
@@ -0,0 +1,58 @@
...
+    );
+  } catch (e) {
+    if (e instanceof IgnitionError && shouldBeHardhatPluginError(e)) {
+      throw new HardhatError(HardhatError.ERRORS.IGNITION.INTERNAL_ERROR, e);
+    }
+    throw e;
+  }
+
+  // TODO: HH3 revisit looking up the network name for display
+  const networkName = (await hre.network.connect()).networkName;
```
kanej, line 47: We use the network name for forming the etherscan links. Is there a better way of getting the network name at this point?

v-next/hardhat-ignition/src/helpers/hardhat-artifact-resolver.ts
```diff
@@ -0,0 +1,39 @@
...
+import { readJsonFile } from "@ignored/hardhat-vnext-utils/fs";
+
+export class HardhatArtifactResolver implements ArtifactResolver {
+  readonly #hre: HardhatRuntimeEnvironment;
+
+  constructor(_hre: HardhatRuntimeEnvironment) {
+    this.#hre = _hre;
+  }
+
+  public async getBuildInfo(
```
kanej, line 17: This has been rework to use the new build info methods.
Previously FQNs and contract names were treated separately.

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
```diff
@@ -0,0 +1,337 @@
...
+    // }
+
+    if (result.type !== "SUCCESSFUL_DEPLOYMENT") {
+      process.exitCode = 1;
+    }
+
+    return result;
+  } catch (e) {
+    if (e instanceof IgnitionError && shouldBeHardhatPluginError(e)) {
+      throw new HardhatError(HardhatError.ERRORS.IGNITION.INTERNAL_ERROR, e);
```
kanej, line 285: For the moment we are choosing to wrap `IgnitionError`'s with an internal `HardhatError`.
We intend to move IgnitionErrors into HardhatError proper during the Alpha.

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
```diff
@@ -0,0 +1,337 @@
...
+    // if (result.type === "SUCCESSFUL_DEPLOYMENT" && verify) {
+    //   console.log("");
+    //   console.log(chalk.bold("Verifying deployed contracts"));
+    //   console.log("");
+
+    //   await hre.run({ scope: "ignition", task: "verify" }, { deploymentId });
+    // }
+
+    if (result.type !== "SUCCESSFUL_DEPLOYMENT") {
+      process.exitCode = 1;
```
kanej, line 279: Are we happy with this behaviour. We set exitCode as 1 but return from the task.
Is this the right pattern in HH3? Should we be throwing?

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
```diff
@@ -0,0 +1,337 @@
...
+    parameters: parametersInput,
+    deploymentId: givenDeploymentId,
+    defaultSender,
+    reset,
+    verify,
+    strategy: strategyName,
+    writeLocalhostDeployment,
+  },
+  hre: HardhatRuntimeEnvironment,
+): Promise<DeploymentResult | null> => {
```
kanej, line 56: This was void. I modified it to allow tests to determine if a deployment was a success or not.
This is however, exporting a core type.
One option would be to return a `hardhat-plugin` specific type, or something simpler like boolean.
Another option is to return void and alter the test.

v-next/hardhat-ignition/src/helpers.ts
```diff
@@ -0,0 +1,5 @@
+export { HardhatArtifactResolver } from "./helpers/hardhat-artifact-resolver.js";
+export { errorDeploymentResultToExceptionMessage } from "./helpers/error-deployment-result-to-exception-message.js";
```
kanej, line 2: What types should we export to match this?
Do we re-export core types that are in these helper function's signatures?

v-next/hardhat/src/internal/core/hre.ts
alcuadrado: correct

v-next/hardhat/src/internal/core/plugins/detect-plugin-npm-dependency-problems.ts
alcuadrado: good catch

v-next/hardhat-ignition/src/internal/tasks/transactions.ts
alcuadrado: Not right now, but we can revisit it later.

v-next/hardhat-ignition/src/index.ts
```diff
@@ -754,3 +754,66 @@
...
+import { emptyTask, task } from "@ignored/hardhat-vnext/config";
+import { ArgumentType } from "@ignored/hardhat-vnext/types/arguments";
+
+import { PLUGIN_ID } from "./internal/constants.js";
+
+const hardhatIgnitionPlugin: HardhatPlugin = {
+  id: PLUGIN_ID,
+  npmPackage: "@ignored/hardhat-vnext-ignition",
+  tasks: [
+    emptyTask("ignition", "Store your keys in a secure way").build(),
```
alcuadrado, line 17: Incorrect message

v-next/hardhat-ignition/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,70 @@
...
+  const ignition: IgnitionUserConfig = userConfig.ignition ?? {};
+
+  return {
+    ...resolvedConfig,
+    networks: updatedNetworks,
+    paths: {
+      ...paths,
+      ignition: resolveFromRoot(
+        resolvedConfig.paths.root,
+        paths.ignition ?? "ignition",
```
alcuadrado, line 63: ignition will never be here, as it wasn't resolved by core. It should be fetched from the user config.

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
```diff
@@ -14,16 +35,289 @@ interface TaskDeployArguments {
...
+      method: "eth_chainId",
+    }),
+  );
+
+  const deploymentId = resolveDeploymentId(givenDeploymentId, chainId);
+
+  const deploymentDir =
+    connection.networkName === "hardhat" && !writeLocalhostDeployment
+      ? undefined
+      : path.join(hre.config.paths.ignition, "deployments", deploymentId);
```
alcuadrado, line 91: To be revisited after the alpha

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
```diff
@@ -14,16 +35,289 @@ interface TaskDeployArguments {
...
+    method: "eth_accounts",
+  })) as string[];
+
+  const artifactResolver = new HardhatArtifactResolver(hre);
+
+  const executionEventListener = new PrettyEventHandler();
+
+  const strategyConfig = hre.config.ignition.strategyConfig?.[strategyName];
+
+  try {
```
alcuadrado, line 200: all this ledger related things can be removed

alcuadrado: Now they'll be better served by the user interruptions system

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
alcuadrado: i think that's fine

v-next/hardhat-ignition/test/deploy/nonce-checks/error-on-pending-user-transactions.ts
```diff
@@ -71,7 +70,7 @@ describe.skip("execution - error on pending user transactions", () => {
     const outsideFoo = await deployPromise;
     assert.equal(
       outsideFoo,
-      "0x3d0ac80b8daf180b4d03e0ff107caa7089b5494cdbd81ba9d7619cc4d710caae",
+      "0x94581d6d5b6c86965a35291786d92fc802940c55113878103d26f8131744e96d",
```
alcuadrado, line 73: are we using the same addresses?

v-next/hardhat-ignition/src/helpers.ts
alcuadrado: not sure i understand this

v-next/hardhat-ignition/test/deploy/nonce-checks/error-on-pending-user-transactions.ts
kanej: I think when I added extra contracts to the minimal project, it changed the compilation job - and so the bytecode. Or is that just a post-hoc rationalization?

alcuadrado: That would only change the address if we were using `create2`. With normal deployments addresses are computed exclusively based on the sender and nonce

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
kanej: Do you mean how we setup the deployment directory, or the place we are calculating and setting the deployment dir - as in, this should be done during config resolution in a hook?

v-next/hardhat-ignition/src/helpers.ts
kanej: We have a packages export for `helpers` (they are used from the companion plugins).
In `hardhat-ignition-core` we use the API extractor to ensure that all exported types are available.
Should we be doing the same for the helpers here?

v-next/hardhat-ignition/src/index.ts
kanej: I have updated the message to `"Deploy your smart contracts using Hardhat Ignition"`, taken directly from the Ignition scope entry in HH2.

v-next/hardhat-ignition/src/internal/hook-handlers/config.ts
kanej: My mental model of this change:
1. `resolveUserConfig` is a hook that injects into the chain that resolves user config into Hardhat config
2. the `resolvedConfig` here is coming from early links in the chain and is hence the Hardhat config in the process of being built up - not a direct copy of the user config.
3. The intent here is a HardhatUser config with a populated `paths.ignition` entry. To do that I should use the passed in `userConfig.paths.ignition` and fallback to the `"ignition"` default otherwise.

kanej: I have updated the line at 62 to:
 
```ts
userConfig.paths?.ignition ?? "ignition",
```

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
kanej: To confirm, we should remove these ledger related code blocks now, and add a task for after `hardhat-ledger` becoming available to re-enable Ignition support?

v-next/hardhat-ignition/test/deploy/nonce-checks/error-on-pending-user-transactions.ts
kanej: I will dig in and get a better answer on why the address changed.

kanej: I compared the artifact json files for `foo` between HH3 and the version in the Ignition repo. The bytecodes match.
I am going to look at whether the sender address is going through in `hardhat-viem` next.

kanej: We have discussed this out of band. This looks like the correct change.
The string changed here is not the address of the contract but is instead the transaction hash coming from viem's `deployContract`.
The `0x3d0ac80b8daf180b4d03e0ff107caa7089b5494cdbd81ba9d7619cc4d710caae` hash is what HH2's in-memory Hardhat network will give. HH2's http network (aka `npx hardhat node`) gives the hash `0x94581d6d5b6c86965a35291786d92fc802940c55113878103d26f8131744e96d`; the same hash as HH3's in-memory Hardhat network gives for the transaction.
The theory is that the gas limit parameter is different in HH2 between in-memory/http, but is the same in HH3 between in-memory/http but different from Hh2 in-memory.
This is expected as we have intentionally brought in-memory inline with http in HH3.
We judge this assertion change to be safe.

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
kanej: I have removed the ledger related code in this commit: https://github.com/NomicFoundation/hardhat/pull/6227/commits/e5dd73f45d5154bd5db8ea7982b7755581ccc29b

This included removing some tests. We will revisit ledger+Ignition integration during the port of `hardhat-ignition`. The task has been added here: https://github.com/NomicFoundation/hardhat/issues/5646

v-next/hardhat-ignition/src/helpers.ts
kanej: We will revisit the type exports after the initial public alpha

v-next/hardhat-ignition/src/internal/tasks/deploy.ts
kanej: We should change the network name check, to a check on network type - EDR networks are ephemeral.

alcuadrado (approved):

PR #6228, opened 2025-02-05
fix: use unexistent exportation from upgrade module

EnThib:

---

It's not much but it can cause confusion in the documentation. Above the documentation makes us export the "DemoModule" and here we use "useModule" with the proxy module which is not supposed to be exported, I therefore deduce that we must use the import which was introduced above.

Thank you for taking the time to read me and have a good day.


alcuadrado:
I think this change makes sense. Can you double check it, @zoeyTM ?

Also, I think there's another mention of `proxyModule` a bit below.

zoeyTM:
Yep, this looks right, good catch! There are a couple more references to this mistake that should be fixed on lines 315, 323, 356, 364, and 394. If you would correct those as well then we can get this merged @EnThib

zoeyTM (changes requested):
see my comment above ^ this looks good though!

EnThib:
@zoeyTM It's done :)

zoeyTM (approved):
LGTM!

PR #6229, opened 2025-02-06
Ignition: fix for bug when we fail to save transaction hash

zoeyTM:
I don't think we actually have an issue to track this now that I think about it. 

In this PR

- added new journal message `TRANSACTION_PREPARE_SEND` that is written to journal immediately before sending a transaction to the network
- added new logic to execution engine that checks the deployment state for any execution state that has a nonce but no transactions and throws if it finds one, alerting the user and asking them to find the appropriate transaction on a block explorer and use the new `track-tx` command to add it to the deployment
- added a new `track-tx` command that allows users to add a missing transaction back to their deployment. accounts for several edge cases 
- tests for all of the above


packages/hardhat-ignition-core/src/track-transaction.ts
```diff
@@ -0,0 +1,251 @@
...
+              deploymentState,
+              deploymentLoader
+            );
+
+            return;
+          }
+          // case 2: the user sent a different transaction that replaced ours
+          // so we check their transaction for the required number of confirmations
+          else {
+            return checkConfirmations(
```
alcuadrado, line 154: I'm not sure if I understand this part, let's look at it on a call. Other than that LGTM.

alcuadrado (approved):
Re-reviewed this, and now I get it. I think the else in the newline was confusing me.

packages/hardhat-ignition-core/src/track-transaction.ts
alcuadrado: self-answer: This works because it's the else to the `if` in line 104. Hence, the transaction matches the sender and nonce, but not data and value. i.e. a replacement.

Issue #5722, opened 2024-09-05
encrypted keystore


kanej:
The v3 `hardhat-keystore` should encrypt the secrets on disk within the keystore file.

See the design doc for details: https://www.notion.so/nomicfoundation/hardhat-keystore-ec80978a6de54098b60350a436b3e666

## Follow up tasks:
- [ ] currently, tasks accept a `key` as a parameter to identify a specific secret. However, this may cause confusion, as it could be mistaken for the key used in the encryption logic. We should consider a clearer name, such as `id`
- [ ] currently, if the password is incorrect, the keystore decryption fails and throws an error. Should we handle this failure more gracefully?
![Image](https://github.com/user-attachments/assets/085858e5-46d0-4460-a0c5-6bf5d88d75fc)

Tmthang1601:
Hi @kanej, I am unable to access the design document linked in the task description for the v3 hardhat-keystore encryption details: https://www.notion.so/nomicfoundation/hardhat-keystore-ec80978a6de54098b60350a436b3e666. It seems that access is restricted, and I receive a message saying, 'Page non accessible. You will be able to access this page once your application is approved.'

Could you please grant me the necessary permissions to view this document, or alternatively, provide the relevant details or a different link where I can access the information needed for this task?

![image](https://github.com/user-attachments/assets/17a40625-cb22-46a2-9984-356fb0fb729d)

Tmthang1601:
Thank you @kanej for the clarification. I appreciate your response and understand that some design documents are kept internal. As someone interested in contributing to Hardhat, I'm curious if there are any specific areas or tasks where community involvement is currently welcome? I'd be excited to learn more about how I can support the project in a meaningful way, whether it's through smaller tasks or areas that are open for community contributions.
> Hey @Tmthang1601 thanks for the interest. Issues marked `v-next` are intended for the Hardhat team. The design doc is internal. Are you looking for good tasks to contribute on?

PR #6232, opened 2025-02-06
Encrypted keystore

ChristopherDedominici:
For a list of follow-up tasks related to user experience improvements (not security improvements), see the [parent issue here](https://github.com/NomicFoundation/hardhat/issues/5722)

## QUESTIONS:
- user inputs are handled via `requestSecretInput` from `hre.interruptions`, but outputs are not currently managed by `hre.interruptions`. I believe they should be. Should we address this in this PR, or should I add it to the follow-up tasks?


v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -706,25 +706,12 @@ Try using another mnemonic or deriving less keys.`,
     },
   },
   KEYSTORE: {
-    INVALID_KEYSTORE_FILE_FORMAT: {
```
ChristopherDedominici, line 709: Removed because it is no longer needed, as user input is now handled via the `requestSecretInput` hook.

v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
```diff
@@ -39,7 +41,16 @@ export default async (): Promise<Partial<ConfigurationVariableHooks>> => {
         return next(context, variable);
       }
 
-      return keystore.readValue(variable.name);
+      const password = await askPassword(
+        context.interruptions.requestSecretInput.bind(context.interruptions),
```
ChristopherDedominici, line 45: `bind` is needed because `requestSecretInput` is a method of `hre.interruptions`, and when passed as an argument, it loses its original this context. Since method calls rely on the object reference for this, passing it as a standalone function removes that association. To ensure it retains the correct context, we need to explicitly bind this using `.bind(hre.interruptions)`.

This approach will be used also later in other files

v-next/hardhat-keystore/src/internal/keystores/encryption.ts
```diff
@@ -0,0 +1,630 @@
...
+ * @returns The derived master key.
+ */
+function deriveMasterKey({
+  password,
+  salt,
+}: {
+  password: string;
+  salt: Uint8Array;
+}): Uint8Array {
+  password = password.normalize(PASSWORD_NORMALIZATION_FORM);
```
ChristopherDedominici, line 553: I need a double check that the `normalization` is only required here

v-next/hardhat-keystore/src/internal/keystores/password.ts
```diff
@@ -0,0 +1,58 @@
...
+import { UserDisplayMessages } from "../ui/user-display-messages.js";
+
+export async function setUpPassword(
+  requestSecretInput: (
+    interruptor: string,
+    inputDescription: string,
+  ) => Promise<string>,
+  consoleLog: (text: string) => void = console.log,
+): Promise<string> {
+  const PASSWORD_REGEX = /^.{8,}$/;
```
ChristopherDedominici, line 13: Currently, we only require a minimum of 8 characters. Do we want to introduce additional rules?

v-next/hardhat-keystore/src/internal/keystores/password.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  consoleLog(UserDisplayMessages.keystoreBannerMessage());
```
ChristopherDedominici, line 15: How it currently looks like:
image

v-next/hardhat-keystore/src/internal/keystores/password.ts
```diff
@@ -0,0 +1,58 @@
...
+    );
+
+    if (!PASSWORD_REGEX.test(password)) {
+      password = undefined;
+      consoleLog(chalk.red(UserDisplayMessages.passwordRequirementsError()));
+    }
+  }
+
+  let confirmPassword: string | undefined;
+  while (confirmPassword === undefined) {
```
ChristopherDedominici, line 36: This loop will continue indefinitely until the confirmation password matches the original password.
This check ensures that the user saves the correct password, preventing spelling mistakes.

v-next/hardhat-keystore/test/keystores/encryption.ts
```diff
@@ -0,0 +1,992 @@
...
+            masterKey,
+            encryptedKeystore: corruptedKeystore,
+          }),
+        (e) => e instanceof InvalidHmacError,
+      );
+    });
+  });
+});
+
+describe("password normalization", () => {
```
ChristopherDedominici, line 962: I added this additional test to validate that the normalization is working

v-next/hardhat-keystore/src/internal/keystores/keystore.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  constructor(keystoreData: EncryptedKeystore) {
+    this.#keystoreData = keystoreData;
+  }
+
+  public toJSON(): EncryptedKeystore {
+    return this.#keystoreData;
+  }
+
+  public async listKeys(): Promise<string[]> {
```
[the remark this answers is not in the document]
alcuadrado: This is expected

v-next/hardhat-keystore/src/internal/keystores/encryption.ts
```diff
@@ -0,0 +1,630 @@
+import { siv } from "@noble/ciphers/aes";
```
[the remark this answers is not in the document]
alcuadrado: I'll get to that tomorrow

v-next/hardhat-errors/src/descriptors.ts
alcuadrado: I don't get this comment. Maybe it's in the wrong line?

v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
alcuadrado: Won't this ask for the password every time?

I think we want to cache the masterkey in line 17

v-next/hardhat-keystore/src/internal/keystores/encryption.ts
alcuadrado: Yes, I think this is correct.

alcuadrado:
> user inputs are handled via requestSecretInput from hre.interruptions, but outputs are not currently managed by hre.interruptions. I believe they should be. Should we address this in this PR, or should I add it to the follow-up tasks?

Actually, I think they are fine now. As the output is printed from tasks.

alcuadrado:
Still reviewing this, but looks good.

I think something that we should do is rename `key` to `id`.

ChristopherDedominici:
> Still reviewing this, but looks good.
> 
> I think something that we should do is rename the secret's `key` to `id`.

@alcuadrado yes, I added this as a follow up task with another one too, see here: https://github.com/NomicFoundation/hardhat/issues/5722

I added them as follow-up tasks in case we want to close this PR ASAP. However, if there's time, I'll address them right away

v-next/hardhat-errors/src/descriptors.ts
ChristopherDedominici: Yes, This comment is for the following lines, the ones where I removed `USERINTERRUPTION_NOT_IMPLEMENTED`.

`INVALID_KEYSTORE_FILE_FORMAT` has been removed because the validation is now performed via the HMAC and not via zod, so an encryption error will be thrown

v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
ChristopherDedominici: Yes, good catch. I'll address it and add a test to be sure the password is only asked once

ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/6232/commits/dbaae55973f019c0520f48e72c9fa72b8901cd7a)

agostbiro (commented):
Hey, great job on this PR! 

I have one concern regarding HMAC verification of the file on disk. 

I noticed in a [comment](https://github.com/NomicFoundation/hardhat/pull/6232/files#r1950952694) that we don't want to verify the HMAC when listing keys to avoid prompting the user for a password. I think this is ok if the sole purpose of listing the keys is to display them to the user and it comes with a warning that the authenticity of the file is not verified.

But if there is further logic depending on the presence of a key and if an attacker can control whether a key is present, that is dangerous. For example, if a fallback mechanism is introduced in the future, the attacker can trigger the fallback mechanism by removing a key from the file on disk.

So I'd suggest creating two incompatible key store types to be used in different tasks: a read only `UnverifiedKeyStore` and read/write `VerifiedKeyStore` where read operations are verified too. Then it's easy to catch in code review whether the specific task needs verification and if the proper warnings are present.

v-next/hardhat-keystore/src/internal/keystores/keystore.ts
```diff
@@ -0,0 +1,69 @@
...
+    return Object.keys(this.#keystoreData.secrets);
```
agostbiro, line 24: Linking review comment: https://github.com/NomicFoundation/hardhat/pull/6232#pullrequestreview-2611368134

v-next/hardhat-keystore/src/internal/keystores/keystore.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  public async hasKey(key: string): Promise<boolean> {
+    return Object.keys(this.#keystoreData.secrets).includes(key);
```
agostbiro, line 28: Linking review comment: https://github.com/NomicFoundation/hardhat/pull/6232#pullrequestreview-2611368134

v-next/hardhat-keystore/src/internal/keystores/password.ts
agostbiro: This is in line with [OWASP recommendation.](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#implement-proper-password-strength-controls) If we want to improve on this, we could provide password strength feedback to the user or [check for compromised password](https://haveibeenpwned.com/API/v3#PwnedPasswords), but I think these are nice to haves only.

v-next/hardhat-keystore/src/internal/keystores/encryption.ts
```diff
@@ -0,0 +1,630 @@
...
+      "Decryption failed: make sure you are using the right password/key and that your encrypted data isn't corrupted",
+      cause,
+    );
+  }
+}
+
+export class SecretNotFoundError extends CustomError {
+  public readonly key: string;
+
+  constructor(key: string) {
```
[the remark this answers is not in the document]
ChristopherDedominici: I agree, it is confusing. There is a work item to track it, see here: https://github.com/NomicFoundation/hardhat/pull/6232#issuecomment-2653360769

v-next/hardhat-keystore/src/internal/keystores/encryption.ts
```diff
@@ -0,0 +1,630 @@
...
+    super(`Invalid hmac in keystore`);
+  }
+}
+
+/// ////////////////////////////////////////////////////////////////////////////
+// Generic crypto utils
+/// ////////////////////////////////////////////////////////////////////////////
+
+/**
+ * Encrypts the utf-8 encoded value using the master key, and a new random iv.
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6232/commits/09ff1ab810c291c6559a0579f2f75cad77422b79

v-next/hardhat-keystore/src/internal/keystores/encryption.ts
```diff
@@ -0,0 +1,630 @@
...
+  masterKey: Uint8Array;
+  encryptedKeystore: EncryptedKeystore;
+  keyToRemove: string;
+}): EncryptedKeystore {
+  if (!(keyToRemove in encryptedKeystore.secrets)) {
+    // eslint-disable-next-line no-restricted-syntax -- We don't throw HardhatErrors here
+    throw new SecretNotFoundError(keyToRemove);
+  }
+
+  validateHmac({ masterKey, encryptedKeystore });
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6232/commits/9244d97d86e2fe6848b6794a53d3f560d8f27f00

v-next/hardhat-keystore/src/internal/tasks/delete.ts
```diff
@@ -32,7 +42,7 @@ export const remove = async (
 
   const keystore = await keystoreLoader.loadKeystore();
 
-  const keys = await keystore.listKeys();
+  const keys = await keystore.listUnverifiedKeys();
 
   if (!keys.includes(key)) {
```
ChristopherDedominici, line 47: Here we check if the key we want to delete exists using the unverified method `listUnverifiedKeys`. Related to this discussion:

ChristopherDedominici:
> `UnverifiedKeyStore`

This has been addressed here: https://github.com/NomicFoundation/hardhat/pull/6232/commits/4cb0d2528e3e101d34055ecef5f068984860cb4a

The function `listKeys` has been renamed to `listUnverifiedKeys` to clearly indicate that no integrity check is performed.
The only spot where there could be an issue is here: https://github.com/NomicFoundation/hardhat/pull/6232/files#r1952835150

v-next/hardhat-keystore/src/internal/keystores/keystore.ts
ChristopherDedominici: See here: https://github.com/NomicFoundation/hardhat/pull/6232#issuecomment-2653983599

v-next/hardhat-errors/src/descriptors.ts
alcuadrado: Oh, I see. I would expect the `INVALID_READLINE_OUTPUT` to also be removed

alcuadrado: as in: this plugin shouldn't need readline

v-next/hardhat-keystore/src/internal/keystores/password.ts
alcuadrado: agreed to start with this

v-next/hardhat-keystore/src/internal/loaders/file-manager.ts
```diff
@@ -13,14 +14,12 @@ export class FileManagerImpl implements FileManager {
 
   public writeJsonFile(
     absolutePathToFile: string,
-    keystoreFile: UnencryptedKeystoreFile,
+    keystoreFile: EncryptedKeystore,
   ): Promise<void> {
     return writeJsonFile(absolutePathToFile, keystoreFile);
```
alcuadrado, line 19: Can we write to `"${absolutePathToFile}.tmp"`, and then move that to `absolutePathToFile` once it's ready? That would minimize the chances of getting a corrupted file.

v-next/hardhat-keystore/src/internal/tasks/delete.ts
alcuadrado: I think I'd go for asking for the password first. Failing on the side of security.

ChristopherDedominici:
@alcuadrado @agostbiro based on @agostbiro feedback I modified the code in [this commit ](https://github.com/NomicFoundation/hardhat/pull/6232/commits/67af98902b2313603571dafc5a221a73b4f6f1c2) in the following way:

- I added a function called `doesKeyExist` (with corresponding tests) that first checks the `HMAC integrity` and then verifies if the key exists.
- In the hook used to fetch secrets, the key existence check is now performed using the `doesKeyExist` function, ensuring the file is validated before any action is taken.
- In the `delete`, `get`, and `set` tasks, the key existence check is also done using the `doesKeyExist` function, ensuring the file is validated before performing any actions.

Please double-check that everything looks good to you.

v-next/hardhat-keystore/src/internal/tasks/delete.ts
ChristopherDedominici: See here: https://github.com/NomicFoundation/hardhat/pull/6232#issuecomment-2656758272

agostbiro:
Thanks @ChristopherDedominici ! Keystore internals LGTM.

v-next/hardhat-keystore/src/internal/keystores/encryption.ts
```diff
@@ -0,0 +1,659 @@
...
+ * @param masterKey The master key to use.
+ * @param encryptedKeystore - The encrypted keystore object containing the secrets.
+ * @param key - The name of the secret to check for existence.
+ * @returns True if the key is present in the keystore, otherwise false.
+ *
+ * @remarks
+ * This function first calls `validateHmac` to verify the cryptographic integrity of
+ * the keystore before checking for the existence of the specified key.
+ */
+export function doesKeyExist({
```
alcuadrado, line 245: Tiny comment: can we move this to the `Keystore primitives` section of this file? e.g. before `addSecretToKeystore`.

v-next/hardhat-keystore/src/internal/loaders/file-manager.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6232/commits/88b9b5c12a5b73bd2f5d71f4a109096f505eaa5d

v-next/hardhat-keystore/src/internal/keystores/encryption.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6232/commits/c2ed9d04b3f919350945539553cbaade2fb115fd

alcuadrado (approved):
🥳

PR #6235, opened 2025-02-07
feat: integrate stack traces into solidity test

### LINKED ISSUE #6183: Integrate Stack traces into Solidity Test

galargh:

---

Closes #6183

In this PR, I integrate stack traces into solidity tests (as per https://nomicfoundation.slack.com/archives/C03UN893ZT4/p1738858331089219).

This is an example of the test solidity UI implemented in this PR. It is subject to change based on user feedback.

Screenshot 2025-02-07 at 09 37 45


v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
```diff
@@ -68,7 +68,7 @@ export function createSolidityErrorWithStackTrace(
   }
 }
 
-function encodeStackTraceEntry(
+export function encodeStackTraceEntry(
```
galargh, line 71: I export this and `getMessageFromLastStackTraceEntry` for use in `solidity-test` built-in plugin so that stack trace entry parsing doesn't have to be duplicated.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -20,7 +20,6 @@ const solidityTestUserConfigType = z.object({
       write: z.array(z.string()).optional(),
     })
     .optional(),
-  trace: z.boolean().optional(),
```
galargh, line 23: As per the release notes, the `trace` option is no longer supported.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -117,27 +122,60 @@ export async function* testReporter(
...
+        for (const entry of stackTrace.reverse()) {
+          const callsite = encodeStackTraceEntry(entry);
+          if (callsite !== undefined) {
+            stackTraceStack.push(`  at ${callsite.toString()}`);
+          }
+        }
+
+        if (stackTraceMessage !== undefined || stackTraceStack.length > 0) {
+          yield `${stackTraceMessage ?? "Stack trace:"}\n${chalk.grey(stackTraceStack.join("\n"))}\n`;
+        }
```
galargh, line 145: This formats to:
```
Either the message retrieved from the stack trace or the literal string "Stack trace:"
  at callsite 1
  at callsite 2
```

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -117,27 +122,60 @@ export async function* testReporter(
...
+      }
+
+      if (
+        failure.decodedLogs !== undefined &&
+        failure.decodedLogs !== null &&
+        failure.decodedLogs.length > 0
+      ) {
```
galargh, line 152: I added extra `!== null` checks to all the properties of `TestResult` we inspect because I noticed they sometimes reach us as `null`s even though that's not reflected in their types (I noticed that for `failure.reason` but decided to apply the checks to everything just in case).

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -117,27 +122,60 @@ export async function* testReporter(
...
+        yield `Decoded Logs:\n${chalk.grey(failure.decodedLogs.map((log) => `  ${log}`).join("\n"))}\n`;
```
galargh, line 153: I added an extra newline before printing out the logs here to make it consistent with how stack traces are displayed. It should make the logs more readable. See the example output in the description.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -117,27 +122,60 @@ export async function* testReporter(
...
+      if (
+        failure.counterexample !== undefined &&
+        failure.counterexample !== null
+      ) {
         const counterexamples = Array.isArray(failure.counterexample)
           ? failure.counterexample
           : [failure.counterexample];
 
         for (const counterexample of counterexamples) {
-          const details = `{\n${Object.entries(counterexample)
```
galargh, line 134: I removed the `{}` enclosing from the counterexample as it didn't provide any additional info but took up valuable space. Not having it also make it more consistent with the rest. See the example output in the description.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
```diff
@@ -54,27 +56,95 @@ export function throwIfSolidityBuildFailed(
...
+        await artifactManager.getBuildInfoOutputPath(buildId);
+
+      assertHardhatInvariant(
+        buildInfoPath !== undefined,
+        "buildInfoPath should not be undefined",
+      );
+      assertHardhatInvariant(
+        buildInfoOutputPath !== undefined,
+        "buildInfoOutputPath should not be undefined",
+      );
```
galargh, line 88: I decided to assert these as Hardhat invariants because of the way we currently call this function i.e. immediately after a build which should have created these build info files. If the usage pattern changes, these might not hold as invariants anymore and should be transformed into other errors instead.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
```diff
@@ -54,27 +56,95 @@ export function throwIfSolidityBuildFailed(
...
-  const artifacts: EdrArtifact[] = [];
+  const solcVersions = Object.fromEntries(
+    buildInfos.map(({ buildInfo }) => {
+      const solidityBuildInfo: SolidityBuildInfo = JSON.parse(
+        new TextDecoder("utf-8").decode(buildInfo),
+      );
+
+      return [solidityBuildInfo.id, solidityBuildInfo.solcVersion];
+    }),
+  );
```
galargh, line 122: We need the build ID to solc version mapping in order to create artifacts in the format EDR expects.

I decided to retrieve this mapping from the build infos, which we have to create and pass over to EDR anyway. By doing so, I was able to avoid reading the build info file from disk for the second time.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
```diff
@@ -54,27 +56,95 @@ export function throwIfSolidityBuildFailed(
...
+      return result.contractArtifactsGenerated.map((artifactPath) => ({
+        source,
+        buildId: result.buildId,
+        artifactPath,
+      }));
+    })
+    .flat();
+
+  return Promise.all(
+    contractArtifacts.map(async ({ source, buildId, artifactPath }) => {
```
galargh, line 135: I parallelised reading artifacts from disk here in hope that it will speed up the preparation for running solidity tests.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
fvictorio: cc @agostbiro

agostbiro: I can repro it in EDR integration tests, but not in a standalone NAPI-rs project. I created an [issue](https://github.com/NomicFoundation/edr/issues/796) and set it aside for now until after the alpha release.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
alcuadrado: Can you explain this in the comments?

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
```diff
@@ -181,7 +181,7 @@ function sourceReferenceToSolidityCallsite(
   );
 }
 
-function getMessageFromLastStackTraceEntry(
+export function getMessageFromLastStackTraceEntry(
```
alcuadrado, line 184: Actually, reusing this adds a ton of network-simulation-level concepts to the error messages. For example, `"Transaction reverted: library was called directly"`, while there's no concept of transactions here. It also adds a ton of noise that's only present for backwards compatibility with existing libraries, like `"VM Exception while processing transaction:"`.

Ideally we'd have a cleaner version of the messages in the solidity tests reporter. 

I guess the simpler change, at least for now, would be duplicating this function in this same file (with a different name) and polishing the error messages a bit.

`encodeStackTraceEntry` doesn't seem to have this problem.

alcuadrado (commented):
I left a comment regarding error messages, apart from that LGTM

v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
```diff
@@ -54,27 +56,95 @@ export function throwIfSolidityBuildFailed(
...
  */
 export async function getArtifacts(
   results: SuccessfulSolidityBuildResults,
-  artifactsRootPath: string,
+  buildInfos: BuildInfoAndOutput[],
 ): Promise<EdrArtifact[]> {
-  const artifacts: EdrArtifact[] = [];
+  const solcVersions = Object.fromEntries(
+    buildInfos.map(({ buildInfo }) => {
+      const solidityBuildInfo: SolidityBuildInfo = JSON.parse(
```
alcuadrado, line 116: There's a problem here. We shouldn't be needing to parse the build infos in hardhat.

I believe the problem comes from EDR needing an

```ts
export interface Artifact {
  /** The identifier of the artifact. */
  id: ArtifactId
  /** The test contract. */
  contract: ContractData
}

/** The identifier of a Solidity contract. */
export interface ArtifactId {
  /** The name of the contract. */
  name: string
  /** Original source file path. */
  source: string
  /** The solc semver string. */
  solcVersion: string
}
```

which includes the `solcVersion`, which is already provided in the build infos that we pass it.

I guess ideally we just pass the build infos, build info outputs, and an array of something like this:

```
export interface ContractIdentifier {
  buildId: string;
  sourceName: string;
  contractName: string;
}
```

That removes the need to deal with artifacts here, where it's not really needed.

agostbiro: Thanks Pato, responding on Slack.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
galargh: That's a great point! Actually, revisiting this piece of code, I started thinking if we even have to try and retrieve the message from the traces at all. My reasoning is that we might be getting the same information via `TestResult.reason` already anyway. 

If my thinking is correct (which I'm yet to double-check), then we could do [`5a25255` (#6235)](https://github.com/NomicFoundation/hardhat/pull/6235/commits/5a25255e26d98c950d519238919f8d0a7a6d4567) which would result in this example output:

Screenshot 2025-02-07 at 16 54 56

alcuadrado: I think that looks good

alcuadrado (approved):
If the new change works with other failure reasons (i.e. supports the same cases than the previous implementation), this should be ready

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
galargh: I'm checking in with Agost about this to make sure. I'll report our findings back here.

galargh: Agost is the best! So it seems `forge` uses `TestResult.reason` as their main message for the test outputs, which tells me it should contain all the crucial information we need.

I'd say, let's go with this, and if we find ourselves lacking detail at some point, we can look again if we can find it somewhere among trace props.

PR #6238, opened 2025-02-07
Don't parse build infos

alcuadrado:
A small PR updating the build system so that we don't need to parse the build infos when running solidity tests.


alcuadrado:
I also double-checked that we aren't parsing them anywhere, except for the network manager (until we get the new API to pass build infos in the provider)

galargh (approved):
🚀

PR #6250, opened 2025-02-10
feat: send EDR panics to sentry

### LINKED ISSUE #6218: Ensure if EDR crashes we get an exception
Currently if EDR crashes we get a panic - we should try and catch and display a proper error message and send it to sentry.

galargh:

---

Closes #6218

In this PR, I propose to:
- enclose each interactions with the [internal EDR provider](https://github.com/NomicFoundation/hardhat/blob/72b6e43b0bfbbf4021de56a33b2f989824817464/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts#L137) in a try-catch block
- wrap the unknown errors caught during such interactions in a newly introduced `UnknownError` which extends `ProviderError`
- wrap the unknown errors caught during fetch requests to the [HTTP provider](https://github.com/NomicFoundation/hardhat/blob/72b6e43b0bfbbf4021de56a33b2f989824817464/v-next/hardhat/src/internal/builtin-plugins/network-manager/http-provider.ts#L182) in the same `UnknownError`
- allow for the `UnknownError`s to be reported to sentry

Example report: https://nomic-labs.sentry.io/issues/6286399601/?project=4507685793103872&referrer=project-issue-stream

The alternative we discussed earlier is to search for `panic` string within the error message and allow for such errors to be reported.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/http-provider.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/http-provider.ts
@@ -11,6 +11,7 @@ import type {
 } from "@ignored/hardhat-vnext-utils/request";
 
 import { HardhatError } from "@ignored/hardhat-vnext-errors";
+import { ensureError } from "@ignored/hardhat-vnext-utils/error";
 import { sleep, isObject } from "@ignored/hardhat-vnext-utils/lang";
 import {
   getDispatcher,
@@ -31,7 +32,11 @@ import {
   isFailedJsonRpcResponse,
   parseJsonRpcResponse,
 } from "./json-rpc.js";
-import { ProviderError, LimitExceededError } from "./provider-errors.js";
+import {
+  ProviderError,
+  LimitExceededError,
+  UnknownError,
+} from "./provider-errors.js";
 
 const TOO_MANY_REQUEST_STATUS = 429;
 const MAX_RETRIES = 6;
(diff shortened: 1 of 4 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/http-provider.ts
```diff
@@ -218,7 +225,8 @@ export class HttpProvider extends BaseProvider {
         throw new LimitExceededError(undefined, e);
       }
 
-      throw e;
+      // eslint-disable-next-line no-restricted-syntax -- allow throwing ProviderError
+      throw new UnknownError(e.message, e);
```
alcuadrado, line 229: This may report every HTTP-related error. We can start with reporting them, but we may need to filter most out later. LGTM though

galargh: My thinking was to err on the side of reporting too much rather than too little.

alcuadrado (approved):
LGTM, but can you rename the PR before merging? I think it got an incorrect name by mistake

v-next/hardhat/src/internal/builtin-plugins/network-manager/http-provider.ts
alcuadrado: agreed

PR #6256, opened 2025-02-11
fix: wait for the caching to finish before returning from build

galargh:

---

This PR fixes an error in the `SolidityBuildSystemImplementation.build` method we observed when running tests in the Windows environment.

```
# Error: Test "should recompile the project when given the same input as in the previous call but the force flag is set" at ..\\..\\..\\internal\\builtin-plugins\\solidity\\build-system\\build-system.ts:1:6740 generated asynchronous activity after the test ended. This activity created the error "FileNotFoundError: File C:\\Users\\runneradmin\\AppData\\Local\\Temp\\hardhat-tests-solidity-build-system-implementation\\cache\\compiler-output\\v1\\8c3d2c1b5b88fb28c81dca6411cca903fbc8047d.json.tmp not found" and would have caused the test to fail, but instead triggered an unhandledRejection event.
```

The issue is that we were not awaiting writes to cache to finish before returning from the build method. In this PR, I propose to add an await for all the interactions with cache to finish just before the return statement.

##### Before

14/100 jobs running hardhat tests on Windows failed due to an async not being resolved before a test finished.

https://github.com/NomicFoundation/hardhat/actions/runs/13269013804

###### After

All 100 jobs running hardhat on Windows suceeded.

https://github.com/NomicFoundation/hardhat/actions/runs/13269290566

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -184,7 +184,9 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       (result) => !this.#hasCompilationErrors(result.compilerOutput),
     );
 
-    this.#cacheCompilationResults(uncachedSuccessfulResults);
+    const cachingCompilationResults = this.#cacheCompilationResults(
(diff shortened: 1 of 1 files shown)


alcuadrado (approved):
Good catch! 👏

PR #6258, opened 2025-02-11
fix: fix asynchronous issue with resolveAddress in populate function

0xlora:

---

In the `populate` function, the `resolveAddress` call was not awaited, causing the returned `Promise<string>` to be assigned directly to `pop.to`, which expects a `string`. I've added `await` to correctly resolve the address before assignment. This ensures the function works as expected, resolving the promise to the actual address string.


ChristopherDedominici:
Hi @0xlora, thanks for opening the PR.

I took a look, and the code is working as expected, it's not supposed to wait for the promise to resolve at that point. This is because all the `pop` properties are pending promises that get resolved within the `resolveProperties` function ([see here](https://github.com/NomicFoundation/hardhat/blob/2924ab3846c4a942e5c2278b08246c23dbb1c611/packages/hardhat-ethers/src/signers.ts#L296)).

Inside the `resolveProperties` function, there's this snippet:
```
  const results = await Promise.all(
    keys.map((k) => Promise.resolve(value[k as keyof T]))
  );
```

This loops through all the properties and resolves them ([link to code](https://github.com/NomicFoundation/hardhat/blob/2924ab3846c4a942e5c2278b08246c23dbb1c611/packages/hardhat-ethers/src/internal/ethers-utils.ts#L98)).

PR #6259, opened 2025-02-11
fix: display all unused diagnostics at the end of the test report

galargh:

---

This is a quick fix for an issue where we wouldn't display error details that were passed to us via `test:diagnostics` events. This was due to the fact that we accidentally swallowed such messages when formatting the unused diagnostics events.

As a follow-up to this, when the diagnostic message has a file property and that test file has failed to execute, we should display the message next to the file failure instead.

### DIFF
--- a/v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/asynchronous-activity-test/result.txt
+++ b/v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/asynchronous-activity-test/result.txt
@@ -1,0 +1,14 @@
+
+  [90m[32m✔[39m[90m foo[39m
+
+  [31m1) integration-tests/fixture-tests/asynchronous-activity-test/test.ts[39m [31m[3m(203ms)[23m[39m
+
+
+[32m1 passing[39m[90m (234ms)[39m[31m[39m
+[31m1 failing[39m
+[34mℹ[39m Error: Test "foo" at integration-tests/fixture-tests/asynchronous-activity-test/test.ts:1:29 generated asynchronous activity after the test ended. This activity created the error "Error" and would have caused the test to fail, but instead triggered an uncaughtException event.
+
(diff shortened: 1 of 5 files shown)


v-next/hardhat-node-test-reporter/src/formatting.ts
```diff
@@ -130,6 +130,7 @@ export function formatUnusedDiagnostics(
       if (isTestOnlyMessage(message)) {
         return testOnlyMessage ?? message;
       }
+      return message;
```
galargh, line 133: This is why we weren't displaying the error details at all before.

v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/asynchronous-activity-test/result.txt
```diff
@@ -0,0 +1,14 @@
+
+  [90m[32m✔[39m[90m foo[39m
+
+  [31m1) integration-tests/fixture-tests/asynchronous-activity-test/test.ts[39m [31m[3m(203ms)[23m[39m
+
+
+[32m1 passing[39m[90m (234ms)[39m[31m[39m
+[31m1 failing[39m
+[34mℹ[39m Error: Test "foo" at integration-tests/fixture-tests/asynchronous-activity-test/test.ts:1:29 generated asynchronous activity after the test ended. This activity created the error "Error" and would have caused the test to fail, but instead triggered an uncaughtException event.
```
galargh, line 9: TODO(@galargh): We need to handle the path separators here correctly so that the integration tests pass on Windows.

v-next/hardhat-node-test-reporter/src/formatting.ts
alcuadrado: Ouch 😓

Great catch

alcuadrado (approved):
Approving despite a failing test. I guess we should fix that and merge this.

PR #6263, opened 2025-02-12
feat: turn package into esm on init

galargh:

---

### DIFF
--- a/v-next/hardhat/src/internal/cli/init/init.ts
+++ b/v-next/hardhat/src/internal/cli/init/init.ts
@@ -28,7 +28,7 @@ import {
   installsPeerDependenciesByDefault,
 } from "./package-manager.js";
 import {
-  promptForEsm,
+  promptForMigrateToEsm,
   promptForForce,
   promptForInstall,
   promptForTemplate,
@@ -40,7 +40,7 @@ import { getTemplates } from "./template.js";
 
 export interface InitHardhatOptions {
   workspace?: string;
-  esm?: boolean;
+  migrateToEsm?: boolean;
   template?: string;
   force?: boolean;
   install?: boolean;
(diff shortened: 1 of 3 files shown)


v-next/hardhat/src/internal/cli/init/prompt.ts
```diff
@@ -20,6 +20,25 @@ export async function promptForWorkspace(): Promise<string> {
...
+): Promise<boolean> {
+  ensureTTY();
+
+  const { default: enquirer } = await import("enquirer");
+
+  const esmResponse = await enquirer.prompt<{ esm: boolean }>([
+    {
+      name: "esm",
+      type: "confirm",
+      message: `Hardhat only supports ESM projects. Would you like to set the type for ${pathToPackageJson} to "module" now?`,
```
alcuadrado, line 34: is this a relative path from the CWD?

alcuadrado: We have `shortenPath` in `-utils/path`

v-next/hardhat/src/internal/cli/init/init.ts
```diff
@@ -39,6 +40,7 @@ import { getTemplates } from "./template.js";
 
 export interface InitHardhatOptions {
   workspace?: string;
+  esm?: boolean;
```
alcuadrado, line 43: Can we add a name that better describes what this does?

alcuadrado (approved):
I left two comments, but pre-approving

v-next/hardhat/src/internal/cli/init/prompt.ts
galargh: Good call! It's an absolute path. I changed the name to reflect that and called `shortenPath` on it.

v-next/hardhat/src/internal/cli/init/init.ts
galargh: I renamed this to `migrateToEsm`.

PR #6264, opened 2025-02-12
Stub coverage task and chainType argument on test task

antico5:
- Added a stub `coverage` task
- Added a stub `chainType` option to the solidity-tests task


kanej (changes requested):
Agreed. Lets leave gas reporting. The user's can't include it as a plugin.
Can I scope creep and ask you to add another stubbed out task, `verify`:

```shell
Hardhat version 2.22.5

Usage: hardhat [GLOBAL OPTIONS] verify [--constructor-args <INPUTFILE>] [--contract <STRING>] [--force] [--libraries <INPUTFILE>] [--list-networks] [address] [...constructorArgsParams]

OPTIONS:

  --constructor-args    Path to a Javascript module that exports the constructor arguments 
  --contract            Fully qualified name of the contract to verify. Skips automatic detection of the contract. Use if the deployed bytecode matches more than one contract in your project 
  --force               Enforce contract verification even if the contract is already verified. Use to re-verify partially verified contracts on Blockscout 
  --libraries           Path to a Javascript module that exports a dictionary of library addresses. Use if there are undetectable library addresses in your contract. Library addresses are undetectable if they are only used in the contract constructor 
  --list-networks       Print the list of supported networks 

POSITIONAL ARGUMENTS:

  address               Address of the contract to verify 
  constructorArgsParams Contract constructor arguments. Cannot be used if the --constructor-args option is provided (default: [])

verify: Verifies a contract on Etherscan or Sourcify

```
The options and details aren't needed. I think only the positionals make sense in this context.

kanej (changes requested):
These both work:
image

Can we make both yellow, and eliminate the newline before the chain type warning.

kanej (approved):
LGTM!

I did a manual run through:
image

image

PR #6268, opened 2025-02-12
Expose default publicClient in hardhat-viem

schaable:
This PR should not be merged yet as there is a concurrency issue when deploying multiple contracts using the same public client. The issue occurs randomly and causes a timeout.

An example can be found in `v-next/hardhat-viem/test/contracts.ts` in the test case: _'should be able to deploy a contract with both normal and constructor libraries linked'_.


schaable:
I can confirm the hanging test is still happening. I believe it’s a concurrency issue, since it seems to go away if the libraries are deployed one after the other. That said, I wasn’t able to identify the root cause. The test hangs but eventually passes, so we don’t get any actual error. I added some logging, and everything looks normal compared to runs with different public clients.
I’m going to close this for now and we can revisit it in the future if needed.

PR #6303, opened 2025-02-17
Use proper quotes in error messages

fvictorio:


alcuadrado (approved):

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -1544,7 +1538,7 @@ Please use the fully qualified name of the contract to disambiguate it.`,
     },
     UNKNOWN_TRANSACTION_TYPE: {
       number: 1708,
-      messageTemplate: `Unknown transaction type: {type}`,
+      messageTemplate: `Unknown transaction type: "{type}"`,
       websiteTitle:
         "Hardhat Ignition was unable to display an unknown transaction type",
       websiteDescription:
```
ChristopherDedominici, line 1544: I can't add a comment on the correct line but probably you missed line 1549:

```
... using bigint notation: \`{value}n\``,
```

should be
```
using bigint notation: "{value}n"`,
```

ChristopherDedominici (approved):
Just a small comment. Other than that, LGTM

v-next/hardhat-errors/src/descriptors.ts
alcuadrado: I fixed it

PR #6305, opened 2025-02-17
fix: make the root paths in solidity test unique

galargh:

---

I discovered that when `paths.sources` and `paths.tests.solidity` share common directories, we might end up with duplicate root paths. This PR ensures they are unique. Also, it adds the source file path information to the hardhat invariant violation error message which was helpful in debugging.

Context: https://nomicfoundation.slack.com/archives/C069TLP61BQ/p1739656187561829

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
@@ -66,6 +66,9 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
         }),
       ])
     ).flat(1);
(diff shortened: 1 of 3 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -66,6 +66,9 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
         }),
       ])
     ).flat(1);
+    // NOTE: We remove duplicates in case there is an intersection between
```
alcuadrado, line 69: I think this should go outside of the if

alcuadrado (changes requested):
Great catch! Only a small comment

Issue #6310, opened 2025-02-17
[Typechain] `getContractAt` override is not working


ChristopherDedominici:
### Version of Hardhat

3.0.0-next.18

### What happened?

The `getContractAt` override is not functioning correctly, resulting in improper type generation. The issue stems from an incorrect type definition in the `hardhat-ethers` plugin.

### Minimal reproduction steps

Generate the TypeChain types and use the `getContractAt` method. The generated types will not be applied correctly

PR #6312, opened 2025-02-17
[Typechain] getContractAt override is not working 

ChristopherDedominici:
See issue description [here](https://github.com/NomicFoundation/hardhat/issues/6310)


v-next/example-project/test/node/example-test.ts
```diff
@@ -27,7 +27,7 @@ describe("Example EDR based test", () => {
     const connection = await hre.network.connect();
 
     const Revert = await connection.ethers.getContractFactory("Revert");
-    const revert = await Revert.deploy([]);
+    const revert = await Revert.deploy();
```
ChristopherDedominici, line 30: I also applied this fix in the `example-project` because passing an array to this function is incorrect.

schaable (approved):

PR #6318, opened 2025-02-18
Sync templates with tutorial

fvictorio:


kanej (approved):
I won't speak to the content.
I did a manual run through, and able to run the relevant commands and scripts.
The `README.md` should be updated in `Sending a Transaction to Optimism Sepolia` as you the send script now names the network so `--network` will be ignored.

PR #6334, opened 2025-02-18
chore: deduplicate build ids before getting build infos

galargh:

---

This change greatly improves solidity testing performance. Before, we would read the same build infos over and over again (for every compilation resul - test root path).

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
@@ -67,11 +67,12 @@ export async function getBuildInfos(
   results: SuccessfulSolidityBuildResults,
   artifactManager: ArtifactManager,
 ): Promise<BuildInfoAndOutput[]> {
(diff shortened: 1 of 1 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
```diff
@@ -67,11 +67,12 @@ export async function getBuildInfos(
   results: SuccessfulSolidityBuildResults,
   artifactManager: ArtifactManager,
 ): Promise<BuildInfoAndOutput[]> {
-  const buildIds = await Promise.all(
+  let buildIds = await Promise.all(
     Array.from(new Set(results.values())).map(async ({ compilationJob }) =>
       compilationJob.getBuildId(),
     ),
   );
+  buildIds = Array.from(new Set(buildIds));
```
kanej, line 75: Nit: I would have two different variables.

kanej (approved):
Awesome. Lets get this merged.

PR #6335, opened 2025-02-18
Init projects with pnpm when invoking hardhat from pnpx

### LINKED ISSUE #6321: init with pnpx installs with wrong package manager
In an empty directory, initializing a project with `pnpx` installs dependencies with `npm` and creates a `package-lock.json`. Instead, `pnpm` should be used and a `pnpm-lock.yaml` file should be created.

In effect we are detecting the wrong package manager.

Reported by @fvictorio

antico5:
Uses the `npm_config_user_agent` env variable to tell if it's been invoked by `pnpm dlx` or `pnpx`

I also tried other 2 approaches but discarded them:
- Using `process.argv`: the binary is set as "node" even though initially invoked by pnpx
- Using `process._`: this doesnt work on windows cmd/powershell

From what I've researched and manually tested, `npm_config_user_agent` is set on both type of operating systems

Closes #6321

### DIFF
--- a/v-next/hardhat/src/internal/cli/init/package-manager.ts
+++ b/v-next/hardhat/src/internal/cli/init/package-manager.ts
@@ -19,10 +19,14 @@ export async function getPackageManager(
   const pathToYarnLock = path.join(workspace, "yarn.lock");
   const pathToPnpmLock = path.join(workspace, "pnpm-lock.yaml");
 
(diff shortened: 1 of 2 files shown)


galargh (approved):
I think this should be OK because even if `process.env.npm_config_user_agent` is not available, we'll just fall back to the original behaviour.

PR #6361, opened 2025-02-19
Fix ignition -viem and -ethers type-extensions

alcuadrado:


alcuadrado:
This can be validated in the example project

zoeyTM (approved):

Issue #6358, opened 2025-02-19
Ignition in hardhat config not coming through


kanej:
The ignition field is not being added to the HardhatUserConfig type in either ethers or viem hardhat projects.
The issue is that, while the ethers and viem plugins do add the base ignition plugin itself to the project, they do not extend/include the base plugins type extensions. For a user, this can be fixed by [adding the base plugin to the project directly](https://github.com/NomicFoundation/hardhat-v3-alpha-testing/blob/main/ignition-deploying-via-create2/hardhat.config.ts#L12), but we should fix this internally as well.

PR #6362, opened 2025-02-19
Fix missing ignition field in HardhatUserConfig type

zoeyTM:
fixes https://github.com/NomicFoundation/hardhat/issues/6358


alcuadrado:
Oops, we had a race condition. I sent https://github.com/NomicFoundation/hardhat/pull/6361

The difference is that this version actually loads the plugin, and if there's an installation issue (e.g. a missing peer dependency), hardhat would break with a Node.js error. 

If we do `export type * from "..."` instead, we can print our own error instead.

PR #6394, opened 2025-02-21
Fixed Broken Discord Link in Final Thoughts Tutorial

gap-editor:
**In the file `docs/src/content/tutorial/final-thoughts.md`, I updated the Discord link because the previous one was leading to a GitHub 404 error page."**

This keeps everything clear and professional. Let me know if you need any adjustments! 🚀


kanej:
This is an intentional trade-off we have made. Thanks for the suggestion!

Issue #6331, opened 2025-02-18
Ugly error on bad password in keystore


kanej:
currently, if the password is incorrect, the keystore decryption fails and throws an error. Should we handle this failure more gracefully?

![Image](https://github.com/user-attachments/assets/e3c7059b-2253-4bda-8cbe-798220813781)

We should catch and throw a Hardhat specific error.

PR #6396, opened 2025-02-21
Show hardhat error on bad password

schaable:
Fixes https://github.com/NomicFoundation/hardhat/issues/6331


ChristopherDedominici (approved):

.changeset/short-zoos-arrive.md
```diff
@@ -0,0 +1,6 @@
+---
```
ChristopherDedominici, line 1: I need a double check on this changeset, is this format acceptable for the `v-next`?

kanej: Do you mean have more than one entry for the packages? Yes. The versions of each will be bumped and the changelog will get the text bellow as an entry.

kanej (approved):
The updated error message on a bad password:

image

Issue #5785, opened 2024-09-27
Hardhat-ledger plugin error using rpc nodes that don't support eth_accounts


daveroga:
### Version of Hardhat

2.22.10

### What happened?

Trying to deploy smart contracts through hardhat and `hardhat-ledger` plugin (1.0.3) to Polygon zkEVM networks (Cardona or Mainnet) is not possible because it throws an error when trying to get accounts with `eth_accounts`.

In hardhat-ledger provider code I see that is failing in this code when trying to get accounts through `eth_accounts` when the rpc node is not supporting this in the API:
```
  if (args.method === "eth_accounts") {
      const accounts = (await this._wrappedProvider.request(args));
      return [...accounts, ...this.options.accounts];
  }
```

### Minimal reproduction steps

1. Install `hardhat-ledger` plugin following the instructions here https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-ledger
2. Try to deploy a smart contract to zkEVM Cardona with some faucet through `ledgerAccounts` address with any Ledger device.

### Search terms

hardhat-ledger zkevm getaccounts eth_accounts

kanej:
Thanks for the reported error.
I suspect we want there is to guard against `eth_accounts` throwing and return `[]` as if there were no unlocked accounts.

farhad-zada:
Is this problem solved? i am still getting this

daveroga:
> Is this problem solved? i am still getting this

Not solved yet. We still have a patch for this in our project https://github.com/iden3/contracts/blob/master/patches/%40nomicfoundation%2Bhardhat-ledger%2B1.0.3.patch

farhad-zada:
> > Is this problem solved? i am still getting this
> 
> Not solved yet. We still have a patch for this in our project https://github.com/iden3/contracts/blob/master/patches/%40nomicfoundation%2Bhardhat-ledger%2B1.0.3.patch

I don't encounter it. i just changed the RPC to public one

kanej:
Hey we released a general fix for deprecated `eth_accounts` in `v2.24.2`: https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.24.2

daveroga:
> Hey we released a general fix for deprecated `eth_accounts` in `v2.24.2`: https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%402.24.2

[...]

Issue #5786, opened 2024-09-27
Hardhat-ledger plugin error using blockchains that don't support EIP-1559 format for transactions


daveroga:
### Version of Hardhat

2.22.10

### What happened?

Trying to deploy smart contracts through hardhat and `hardhat-ledger` plugin (1.0.3) to Polygon zkEVM networks (Cardona or Mainnet) is not possible because it throws an error when trying to send signed transaction with `eth_sendRawTransaction` and EIP-1559 format.
```
{"jsonrpc":"2.0","id":17,"error":{"code":-32000,"message":"RPC error response: error getting from address"}}
```
Debugging it I see that is sending the transaction with type EIP-1559 that is not supported by zkEVM now (https://support.polygon.technology/support/solutions/articles/82000897869-does-polygon-zkevm-support-the-eip-1559-format-for-transactions-).
```
{
  "chainId": "2442",
  "type": "EIP-1559",
  ...
}
```
So at this moment to support deploying to Polygon zkEVM networks this ledger-plugin should send legacy transaction (type=0) instead calcultaing then `gasPrice` too.
This will produce legacy transaction format:
```
{
  "chainId": "2442",
  "type": "LegacyTransaction",
  ...
}
```

### Minimal reproduction steps

1. Install `hardhat-ledger` plugin following the instructions here https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-ledger
2. Patch first getting accounts in hardhat-ledger `provider` for zkEVM networks with empty array [] because `eth_accounts` is not supported in zkEVM node now (https://github.com/NomicFoundation/hardhat/issues/5785).
3. Try to deploy a smart contract to zkEVM Cardona with some faucet through `ledgerAccounts` address with any Ledger device.

kanej:
Thanks for the reproduction steps we will take a look.

michalbrabec:
Estimate the effort for V3 and prepare a design doc as appropriate

ChristopherDedominici:
Fixed in ledger for hardhat v3: https://github.com/NomicFoundation/hardhat/pull/6983

PR #6398, opened 2025-02-24
Fix hardhat-ledger in some chains

daveroga:

---

Fixes issues: #5785 , #5786

After fixing these issues with this code, I'm able to deploy with Ledger to networks that weren't supported before like Polygon zkEVM.

### DIFF
--- a/packages/hardhat-ledger/src/provider.ts
+++ b/packages/hardhat-ledger/src/provider.ts
@@ -120,7 +120,20 @@ export class LedgerProvider extends ProviderWrapperWithChainId {
     }
 
     if (args.method === "eth_accounts") {
-      const accounts = (await this._wrappedProvider.request(args)) as string[];
+      // some rpcs return "the method has been deprecated: eth_accounts" error
+      let accounts: string[];
+      try {
+        accounts = (await this._wrappedProvider.request(args)) as string[];
+      } catch (error) {
+        if (
+          error instanceof Error &&
+          error.message.includes("deprecated: eth_accounts")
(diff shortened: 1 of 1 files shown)


antico5:
@daveroga We just released a new version of hardhat (2.24.2) which wraps calls to eth_accounts, defaulting to empty array. Would it be possible for you to test with this new version and see if you are still having issues deploying to polygon or zkEVM ?

daveroga:
> @daveroga We just released a new version of hardhat (2.24.2) which wraps calls to eth_accounts, defaulting to empty array. Would it be possible for you to test with this new version and see if you are still having issues deploying to polygon or zkEVM ?

@antico5 I tested and seems that issue for `eth_accounts` is solved but I still have error for only supporting legacy transactions on zkEVM:
```
ProviderError: RPC error response: only legacy transactions are supported
```

I solved this issue in my patch by forcing legacy transaction before signing tx in `_ethSendTransaction` if EIP1559 fields were not present.

```
   ...

   if (!hasEip1559Fields) {
      baseTx.type = 0; // Legacy transaction type
    }

    const txToSign =
      ethers.Transaction.from(baseTx).unsignedSerialized.substring(2);

  ...
```

kanej:
Thanks @daveroga, you are right #5786 is still an issue.
We are looking now at porting the `hardhat-ledger` plugin to Hardhat 3. That should give us an opportunity to test chains that require legacy transactions and validate your approach.

michalbrabec:
Resolved by #6983

Issue #6392, opened 2025-02-21
[Typechain] generate types only for contracts and not test files


ChristopherDedominici:
### Version of Hardhat

3.0.0-next.23

### What happened?

`types` should be generated only for contracts and not for the solidity test files

### Minimal reproduction steps

Run `npx hardhat test`. The `types` will be generated for the Solidity test file, overriding the previously generated contract types.

PR #6400, opened 2025-02-24
[Typechain] generate types only for contracts and not test files

ChristopherDedominici:
[Parent issue link](https://github.com/NomicFoundation/hardhat/issues/6392)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -645,6 +638,13 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       duplicatedContractNamesDeclarationFilePath,
       getDuplicatedContractNamesDeclarationFile(duplicatedNames),
     );
+
+    await this.#hooks.runHandlerChain(
```
ChristopherDedominici, line 642: Pato mentioned that probably this is the best place to add a hook that can emit the artifacts paths.
His message:
`I think that we should change the moment when we generate the typechain types. Its now on build, but i believe it should be in cleanArtifacts. That already generates some typescript types for the artifacts. Especially when there are clashes between names`

.changeset/calm-clouds-work.md
```diff
@@ -0,0 +1,6 @@
+---
```
ChristopherDedominici, line 1: I need a double check on this changeset, is this format acceptable for the `v-next`?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
galargh: This would mean we generate types only when the entire project is compiled - see https://github.com/NomicFoundation/hardhat/blob/4554b50d1889f4efb165d63ba997bb1debe4057f/v-next/hardhat/src/internal/builtin-plugins/solidity/tasks/compile.ts#L37 - which, as far as I understand, is not necessarily what we want.

To be fair, I also find hooking types generation to the cleanup step a little confusing. It is not obvious to know what the type generation has to do with cleanup without deeply understanding the internals. I think there might be some potential for the refactor in the cleanup function, but that's definitely out of the scope of this PR.

However, as far as when the type generation should get triggered, maybe it should be when the artifacts are emitted after all. But maybe we could add some option to the `build` and consequently `emitArtifacts` to control whether artifact declaration files - https://github.com/NomicFoundation/hardhat/blob/4554b50d1889f4efb165d63ba997bb1debe4057f/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts#L527 - are emitted. Then, we could either propagate that option via the hook system or check whether the declaration file exists for a given artifact on the typechain side. There definitely might be better options, just thinking out loud.

alcuadrado: I think your points are valid, but would require further exploration, @galargh . I'd say we merge this PR and revaluate this soon.

One reason that can make the generation require clenup to be run is that you can have name clashes.

alcuadrado (approved):

Issue #6409, opened 2025-02-26
Ignition mermaid diagram centering is broken for small deployments


zoeyTM:
If the module is small and simple enough, the visualization report is off center:

![Image](https://github.com/user-attachments/assets/7e4e4799-a923-4eb6-8585-194922762e52)

We should center the mermaid diagram even for small modules, but ensure larger examples continue to work.

PR #6410, opened 2025-02-26
Override max-width to fix ignition graph centering

zoeyTM:
fixes https://github.com/NomicFoundation/hardhat/issues/6409


kanej (approved):
I have tested this manually with Chrome and Firefox.

Issue #6336, opened 2025-02-19
stdout maxBuffer length exceeded


kanej:
Reported during the Solidity Test testing for the Initial Public Alpha.
This is the output from running hardhat test solidity in seaport on MacOS ARM machine:

```sh
An unexpected error occurred:

RangeError [ERR_CHILD_PROCESS_STDIO_MAXBUFFER]: stdout maxBuffer length exceeded
    at Socket.onChildStdout (node:child_process:487:14)
    at Socket.emit (node:events:518:28)
    at Socket.emit (node:domain:489:12)
    at addChunk (node:internal/streams/readable:561:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)
    at Readable.push (node:internal/streams/readable:392:5)
    at Pipe.onStreamRead (node:internal/stream_base_commons:189:23)

If you think this is a bug in Hardhat, please report it here: https://hardhat.org/report-bug
```

Reproduction script: https://github.com/galargh/solidity-testing-testing/blob/main/issues/stdio-maxbuffer.sh

Hardhat config used: https://github.com/galargh/solidity-testing-testing/blob/main/repositories.json#L398

The initial investigation:

* Case 1 (WASM solc): solc fails because it runs out of memory. known issue in solc — https://github.com/ethereum/solidity/issues/13966
* Case 2 (Native solc — Merging compilation jobs): We compile a HUGE compilation job, so the buffer to store the output runs out.
* Case 3 (Native solc — Not merging compilation jobs, by modifying the code a big): Hardhat runs out of memory because the new build system keeps all the solc outputs in memory, which is normally prohibitively expensive.

How should we deal with the cases?

* Case 1) not much to do on our end.
* Case 2) Maybe we can merge less aggressively.
* Case 3) This should be reworked. It's important.

## TODO

- [ ] Design task on limiting the out of memory issue in the new build system

Issue #6328, opened 2025-02-18
Investigate error code 143 in constrained environments


kanej:
During the Solidity Test benchmarking several repos failed to run tests with a error code 143 in Github Actions. The tests could be run locally.

The repos included: 
* Uniswap/UniswapX
* Vectorized/solady
* pancakeswap/pancake-v4-core

Disabling optimizers cleared the issue for `Vectorized/solady`.

*Link to an example:* https://github.com/galargh/solidity-testing-testing/actions/runs/13348852314/job/37282728543#step:13:690

PR #6411, opened 2025-02-26
fix: compiler outputs might exceed the max buffer size

galargh:

---

Resolves: #6336
Resolves: #6328

In this PR, I start using streams when handling the solc compiler outputs to support the compilation of very large codebases where the compilation outputs might exceed the maximum buffer size/string length.

#### Changes to the build system and cache

I added two new file system access functions:
- `readJsonFileAsStream`: opens the file as a stream, passes it to the streaming JSON parser, picks the full parsed object from the JSON values stream
- `writeJsonFileAsStream`: parses the JSON object as a stream of tokens, writes it to the file

These functions allow us to avoid storing the full compiler output as a string, which can exceed the maximum string length at times.

To implement these functions, I added two new dependencies:
- [@streamparser/json-node](https://github.com/juanjoDiaz/streamparser-json): this is a completely new dependency
- [json-stream-stringify](https://github.com/Faleij/json-stream-stringify): this dependency was present in Hardhat 2

These functions are used when interacting with compiler output objects, in the cache and when emitting the build info output.

#### Changes to the compiler

Previously, the compiler would execute the underlying solc compiler using the `execFile` function. With that function, we would end up waiting for the entire output to be returned to us as a string. Unfortunately, for very large compiler outputs, the resulting output would exceed the max buffer size and if we increased the buffer size further, we would eventually hit the maximum string length limit.

I removed the usage of `execFile` from the compiler and replaced it with `spawn`, where we pipe the output to a temporary file which we later read using the newly added `readJsonFileAsStream` function. This allows us to avoid creating a string of the compiler output.

###### Deduplicate compiler output file creation

At the moment, we create a file with compiler output three times, in the Compiler, when saving the output to the cache, and when emitting the build info output. We should consider creating it only once and then copying/moving it when needed. This will require changing the build info output format which currently has some extra information apart from the compiler output.

#### Testing

I rerun the solidity testing testing suite using this branch which cleared all the 143s and max buffer size exceeded errors. My hypothesis on why the 143 were cleared as well is that in these cases, we weren't exceeding the max buffer size, but we were running very near it, which caused the memory pressure.

https://github.com/galargh/solidity-testing-testing/actions/runs/13542801435


alcuadrado (approved):
Looks good as a first move, but I'll think about how to avoid reading/writing so much data.

Issue #6377, opened 2025-02-20
Invariant error when running chai matchers and load fixture


kanej:
In a `mocha/ethers` Hardhat 3 project update the test to:

```
// tests/Counter.ts

import { network } from "@ignored/hardhat-vnext";
import { HardhatEthers } from "@ignored/hardhat-vnext-ethers/types";
import { NetworkHelpers } from "@ignored/hardhat-vnext-network-helpers/types";
import { NetworkConnection } from "@ignored/hardhat-vnext/types/network";
import { EthereumProvider } from "@ignored/hardhat-vnext/types/providers";

import { expect } from "chai";

let connection: NetworkConnection;
let ethers: HardhatEthers;
let helpers: NetworkHelpers;
let provider: EthereumProvider;

describe("Counter", function () {
  beforeEach(async () => {
    connection = await network.connect();
    ethers = connection.ethers;
    helpers = connection.networkHelpers;
    provider = connection.provider;
  });

  async function deployTokenFixture() {
    const counter = await ethers.deployContract("Counter");
    return { counter };
  }

  it("load fixture", async function () {
    await helpers.loadFixture(deployTokenFixture);
  });

  it("load fixture and assert changeEtherBalance", async function () {
    const { counter } = await helpers.loadFixture(deployTokenFixture);

    await expect(counter.inc()).to.changeEtherBalance(provider, counter, 0);
  });
});
```

You will get one failing test:
```
  1 failing

  1) Counter
       load fixture and assert changeEtherBalance:
     HardhatError: HHE100: An internal invariant was violated: There should be only 1 transaction in the block
      at assertHardhatInvariant (node_modules/@nomicfoundation/hardhat-errors/src/errors.ts:227:11)
      at getBalanceChange (node_modules/@nomicfoundation/hardhat-ethers-chai-matchers/src/internal/matchers/changeEtherBalance.ts:104:3)
      at async Promise.all (index 0)
      at async Context.<anonymous> (test/Counter.ts:36:5)
```

See this thread for more context: https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1740053709804449

PR #6412, opened 2025-02-26
Invariant error when running chai matchers and load fixture

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/6377

Fix explanation: Hardhat now supports multiple connections. The `snapshots` array used in the `loadFixture` function was declared globally, which meant it was shared across multiple connections, leading to the failure described in the issue.
To fix this problem, I moved the `snapshots` array to the `network-helpers` class declaration. Since a new class is created for every connection, the `snapshots` array will now be unique for each connection.

### DIFF
--- a/v-next/hardhat-network-helpers/src/internal/network-helpers/helpers/load-fixture.ts
+++ b/v-next/hardhat-network-helpers/src/internal/network-helpers/helpers/load-fixture.ts
@@ -2,11 +2,10 @@ import type { Fixture, NetworkHelpers, Snapshot } from "../../../types.js";
 
 import { HardhatError } from "@nomicfoundation/hardhat-errors";
 
-let snapshots: Array<Snapshot<any>> = [];
-
 export async function loadFixture<T>(
   networkHelpers: NetworkHelpers,
   fixture: Fixture<T>,
+  snapshots: Array<Snapshot<any>>,
 ): Promise<T> {
   if (fixture.name === "") {
     throw new HardhatError(
@@ -20,10 +19,14 @@ export async function loadFixture<T>(
     try {
       await snapshot.restorer.restore();
 
-      snapshots = snapshots.filter(
+      const filteredSnapshots = snapshots.filter(
         (s) =>
           Number(s.restorer.snapshotId) <= Number(snapshot.restorer.snapshotId),
       );
+
+      // Modify the array by reference
+      snapshots.length = 0;
+      snapshots.push(...filteredSnapshots);
     } catch (e) {
       if (
         HardhatError.isHardhatError(e) &&
@@ -52,6 +55,7 @@ export async function loadFixture<T>(
   }
 }
 
-export function clearSnapshots(): void {
-  snapshots = [];
+export function clearSnapshots(snapshots: Array<Snapshot<any>>): void {
+  // Modify the array by reference
+  snapshots.length = 0;
 }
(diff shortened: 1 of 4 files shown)


v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
```diff
@@ -50,7 +52,7 @@ export class NetworkHelpers implements NetworkHelpersI {
   }
 
   public clearSnapshots(): void {
-    clearSnapshots();
+    clearSnapshots(this.#snapshots);
```
fvictorio, line 55: This is a bit weird, I'd just do `this.#snapshots = []` here and remove the `clearSnapshots` function in `load-fixture.ts`.

The `clearSnapshots` function only existed because there was some module-level state in that file, but with that gone it doesn't make sense anymore.

v-next/hardhat-network-helpers/src/internal/network-helpers/helpers/load-fixture.ts
```diff
@@ -20,10 +19,14 @@ export async function loadFixture<T>(
...
 
-      snapshots = snapshots.filter(
+      const filteredSnapshots = snapshots.filter(
         (s) =>
           Number(s.restorer.snapshotId) <= Number(snapshot.restorer.snapshotId),
       );
+
+      // Modify the array by reference
+      snapshots.length = 0;
+      snapshots.push(...filteredSnapshots);
```
fvictorio, line 29: Ugh, this is painful. Maybe consider the following approach:

- Make this function return both the result (the `T` value) and the new snapshots
- In `NetworkHelpers.loadFicture`, receive both the result and the new snapshots, update the instance's `snapshots`, and return the result

There might be a more elegant approach, but at least this avoids mutating the argument.

galargh: I agree with this. We don't even have to return two things from here. It'd be enough to return `Snapshot<T>` I think.

Then, we could move the filtering to the caller. Something along the lines of:
```
const snapshot = loadFixture(this, fixture, this.#snapshots);
this.#snapshots = this.#snapshots.filter((s) => {
  // NOTE: Using lower than to ensure the snapshot we got back is not in the list
  return Number(s.restorer.snapshotId) < Number(snapshot.restorer.snapshotId),
});
this.#snapshots.push(snapshot);
return snapshot.data;
```

v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/6412/commits/b4bb6992fe5adbf58b9df5221f3df7522d91dccd)

v-next/hardhat-network-helpers/src/internal/network-helpers/helpers/load-fixture.ts
ChristopherDedominici: Done [here](https://github.com/NomicFoundation/hardhat/pull/6412/commits/026506766278a800819be14b0e21df25033f3e15).

I prefer to keep all the logic within the `loadFixture` function, returning both `snapshots` and `snapshotData`. I want the `NetworkHelpers` class to remain as agnostic as possible regarding the `loadFixture` logic.

galargh: With `snapshots` being owned by the `NetworkHelpers`, I think that's a shared responsibility already, but that's just an implementation details and a matter of opinion. I think it's good to go now. Thank you for making the change 🙇

galargh (approved):

fvictorio (approved):

PR #6420, opened 2025-02-26
build: upgrade to EDR v0.9.0

Wodann:
Upgrades EDR to [v0.9.0](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.9.0), introducing support for the Prague hardfork.

This contains no changes in the code's API. However, it does contain breaking changes in the JSON-RPC. Namely:
- all call request & transaction request types now contain optional [EIP-7702-specifc fields](https://github.com/NomicFoundation/edr/pull/814)
- all transaction getters can now return (optional) [EIP-7702-specific fields](https://github.com/NomicFoundation/edr/pull/814)
- all block getters can now return (optional) [EIP-7685-specific fields](https://github.com/NomicFoundation/edr/pull/814)

All Prague EIPs are extensively tested inside EDR. However, it might be desirable to add some smoke tests to validate end-to-end functionality. Additionally, the Hardhat team might need to do some work to expose configuration of the Prague hardfork to the end user.

### DIFF
--- a/.changeset/ninety-beans-warn.md
+++ b/.changeset/ninety-beans-warn.md
@@ -1,0 +1,6 @@
+---
+"hardhat": patch
+---
+
+Upgraded EDR to [v0.9.0](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.9.0):
+- Added support for the Prague hardfork
--- a/packages/hardhat-core/package.json
+++ b/packages/hardhat-core/package.json
@@ -100,7 +100,7 @@
   "dependencies": {
     "@ethersproject/abi": "^5.1.2",
     "@metamask/eth-sig-util": "^4.0.0",
-    "@nomicfoundation/edr": "^0.8.0",
+    "@nomicfoundation/edr": "^0.9.0",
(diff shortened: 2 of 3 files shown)


ChristopherDedominici (approved):

Wodann:
After some debugging and searches, I figured out that https://github.com/prebuild/prebuild-install/issues/203 was the problem. I looked up how to force an override of a transitive dependency and I came up with https://github.com/NomicFoundation/hardhat/pull/6420/commits/5272eb1dad2cafccf386da2d09e7f82e5d161223 as a fix.

Just running `pnpm i` didn't override the transitive dependency, so I had to manually remove some entries from the PNPM lockfile, before re-running `pnpm i`. That fixed the issue, but I should be clear in stating that this was the result of experimentation rather than know-how. As such, please review the change carefully to make sure that I didn't inadvertently break something.

cc: @ChristopherDedominici, @alcuadrado

ChristopherDedominici (approved):
If this workaround unblocks us, I think it's fine. 
Just a follow-up question: should we create an issue to track this, so we can revert the change later once the issue with `prebuild` is resolved?

PR #6421, opened 2025-02-27
Fix for mermaid centering on v3

zoeyTM:
fixes https://github.com/NomicFoundation/hardhat/issues/6409 for v3


kanej (approved):
I was able to test this locally on Chrome/Firefox for smaller and larger modules.

PR #6428, opened 2025-02-28
chore: disable the v-next changeset release

kanej:
This job will not work with the constraints of a build queue (i.e. you cannot commit and push within the job).

This build is effectively disabled - to avoid any danger (and a failing build) lets completely comment out the job.

We will look to revive an improved version of changeset releases in the coming weeks.

### DIFF
--- a/.github/workflows/v-next-changesets-release.yml
+++ b/.github/workflows/v-next-changesets-release.yml
@@ -1,94 +1,94 @@
-name: Changesets
+# name: Changesets
 
-on:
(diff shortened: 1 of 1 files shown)


ChristopherDedominici (approved):

.github/workflows/v-next-changesets-release.yml
galargh, line 1: Why don't we remove this? We can restore it from git history when needed.

PR #6434, opened 2025-03-03
ci: run the changeset check in the merge queue

galargh:

---

This should unblock the merge queue.

### DIFF
--- a/.github/workflows/check-changeset-added.yml
+++ b/.github/workflows/check-changeset-added.yml
@@ -4,6 +4,7 @@
 name: Check that the PR has a changeset
 
 on:
+  merge_group:
   pull_request:
     branches:
       - main
--- a/.github/workflows/check-changeset-added.yml
+++ b/.github/workflows/check-changeset-added.yml
@@ -26,6 +26,15 @@ jobs:
       - uses: actions/github-script@v7
         with:
           script: |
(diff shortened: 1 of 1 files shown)


kanej (approved):
The docs on the `merge_queue` event are here: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#merge_group

kanej:
The changeset check when run on the `merge_queue` event fails. We try and read the PR, as the `merge_queue` is made up of multiple PRs this fails with a url does not exist error.

So options:
1. Update the job to confirm that every PR in the merge queue has a changeset
2. On the merge_queue event short-circuit and return with no further check

I am going to try putting together 1.

galargh:
> The changeset check when run on the `merge_queue` event fails. We try and read the PR, as the `merge_queue` is made up of multiple PRs this fails with a url does not exist error.
> 
> So options:
> 
> 1. Update the job to confirm that every PR in the merge queue has a changeset
> 2. On the merge_queue event short-circuit and return with no further check
> 
> I am going to try putting together 1.

Ah, right! Sorry, I didn't analyze it to be fair. I can take care of it tomorrow if you get stuck.

kanej:
I reverted to short-circuiting in the merge_queue context.

Issue #6315, opened 2025-02-17
[Typechain] Investigate how to support the `attach` method


ChristopherDedominici:
### Describe the feature

In TypeChain for ethers-v6, the `attach` method is not supported. Should we explore a way to support it in Hardhat?

PR #6442, opened 2025-03-04
[Typechain] Investigate how to support the attach method

ChristopherDedominici:
Fixes [this](https://github.com/NomicFoundation/hardhat/issues/6315)

Depends on: [NomicFoundation/hardhat#6400](https://github.com/NomicFoundation/hardhat/pull/6400).
This dependency exists because test files are still being compiled and do not extend ethers. However, since test files should not be compiled, this is not an issue.


v-next/hardhat-typechain/src/internal/generate-types.ts
```diff
@@ -140,3 +142,37 @@ export function addJsExtensionsIfNeeded(content: string): string {
       `import ${imports} from ${quote}${path}/index.js${quote};`,
   );
 }
+
+function addSupportForAttachMethod(modifiedContent: string): string {
```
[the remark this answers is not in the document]
galargh: Do we expect similar needs to modify `Typechain` to be a common occurrence in the future? If so, have we considered forking it and becoming maintainers ourselves, at least for our own needs.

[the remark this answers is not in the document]
ChristopherDedominici: We discussed this possibility in the design doc but discarded it due to several maintenance complications

v-next/hardhat-typechain/src/internal/generate-types.ts
```diff
@@ -140,3 +142,37 @@ export function addJsExtensionsIfNeeded(content: string): string {
...
+  const pattern = /class\s+(\w+)__factory/; // Pattern to find the contract name in factory files
```
galargh, line 147: Just a note for other reviewers because I checked it myself, `\w+` is equivalent to ` [a-zA-Z0-9_]+`. So it should cover all the characters allowed in contract names 👍

v-next/hardhat-typechain/src/internal/generate-types.ts
```diff
@@ -140,3 +142,37 @@ export function addJsExtensionsIfNeeded(content: string): string {
...
+  const match = modifiedContent.match(pattern);
```
galargh, line 148: Can there only be one such match in a file?

v-next/hardhat-typechain/src/internal/generate-types.ts
```diff
@@ -140,3 +142,37 @@ export function addJsExtensionsIfNeeded(content: string): string {
...
+
+  if (match === null) {
+    // File is not a factory file, so there is no need to modify it
+    return modifiedContent;
+  }
+
+  const contractName = match[1];
+
+  // Insert the "attach" snippet right before the "connect" method
+  const insertPoint = modifiedContent.lastIndexOf("static connect(");
```
galargh, line 158: Can there be other classes (in particular, other factory classes) that also define the static connect method in the same file?

If so, then maybe we could search between the index of the contract name match and the first match of `^}` after that? If not, scratch that ;)

v-next/hardhat-typechain/src/internal/generate-types.ts
```diff
@@ -140,3 +142,37 @@ export function addJsExtensionsIfNeeded(content: string): string {
...
+  `;
+
+  modifiedContent =
+    modifiedContent.slice(0, insertPoint) +
+    attachMethod +
+    modifiedContent.slice(insertPoint);
+
+  // Import the "Addressable" type as it is required by the "attach" method
+  modifiedContent = modifiedContent.replace(
+    "/* eslint-disable */",
```
galargh, line 173: Is it always guaranteed for this comment to be at the top where all the imports are?

v-next/hardhat-typechain/src/internal/generate-types.ts
galargh, line 146: Do you think we could add a rough outline of the expected structure of the file being modified in a comment? I think it would make it easier to reason about the replacements.

I'm thinking something along the lines of, very simple:
```
/** We expect the structure of the modifiedContent to be:
 * /* eslint disable */
 * import
 *
 * ...
 * 
 * class contract_name__factory {
 *   ...
 *   static connect(abcdef) {}
 * }
 */
```

v-next/hardhat-typechain/src/internal/generate-types.ts
ChristopherDedominici: Good question! 
Yes, there is only one match. Only one class per file that creates the contract class. For example:
```
export class Revert__factory extends ContractFactory {
```

Even in this scenario, where two contracts are defined in the same `.sol` file, two separate files will be created.

```
contract Revert {
  function boom() public pure {
    foo();
  }

  function foo() internal pure {
    revert("Boom");
  }
}

contract Revert2 {
  function boom() public pure {
    foo();
  }

  function foo() internal pure {
    revert("Boom");
  }
}
```

v-next/hardhat-typechain/src/internal/generate-types.ts
ChristopherDedominici: No only one method `connect` is defined. The file that I am modifying is coming from a hardcoded template from `Typechain`, [see here:](https://github.com/dethcrypto/TypeChain/blob/4c9773403a2aa66bf6f5fa066e8ef1b00769c71b/packages/target-ethers-v6/src/codegen/index.ts#L255)

v-next/hardhat-typechain/src/internal/generate-types.ts
ChristopherDedominici: Yes, since the files are generated from a hardcoded template in the `typechain` package, the same applies to the above comment about `connect`.

v-next/hardhat-typechain/src/internal/generate-types.ts
ChristopherDedominici: [Done here](https://github.com/NomicFoundation/hardhat/pull/6442/commits/287102a4ea8f3579edcb89e88f48aecb8a52d621)

Issue #6111, opened 2025-01-06
Support `pnpm dlx hardhat --init`


kanej:

aychtang:
@kanej To confirm Franco's assessment the `isYarnProject` and `isPnpmProject` functions in `project-creation.ts` are checking for the existence of the package manager lock files. I can help to test adding the check against `process.env.npm_config_user_agent` as a first validation step.

kanej:
> [@kanej](https://github.com/kanej) To confirm Franco's assessment the `isYarnProject` and `isPnpmProject` functions in `project-creation.ts` are checking for the existence of the package manager lock files. I can help to test adding the check against `process.env.npm_config_user_agent` as a first validation step.

Yes. The checks should be enhanced to first make their determinations off of `process.env.npm_config_user_agent` and fallback to the current approach otherwise.

Let me know if you have any other questions!

antico5:
For hardhat v3 (original issue description), this has been fixed 3 weeks ago: https://github.com/NomicFoundation/hardhat/pull/6335
Aychtang submitted a PR for this issue on v2

aychtang:
Ah, my mistake. I'm new to the codebase and just realised about the v3 branch on `v-next` which is related to the label.

I suppose this fix would also be relevant in v2 for now, but let me know if you would like me to close the PR instead if you're focused on v3.

aychtang:
This should be closed by #6335 and the PR for the fix into v2 has also been merged now at #6445.

PR #6445, opened 2025-03-04
fix: correct package manager selection in CLI install

aychtang:
(It seems there are no existing tests for the CLI installation yet, so I have kept the same level of testing for now and added my manual test below).
- [x] The change was previously discussed on an Issue or with someone from the team.

**Description:**
* (issue #6111) Franco noticed that the install script defaults to using `npm` in some cases, noting that its probably to do with checking the directories package manager related files.
* Confirmed that the checks in `project-creation.ts` are checking for related .lock files, and therefore if the project isn't set up correctly then the CLI could use the wrong package manager.
* Added a new function `getProjectTypeFromUserAgent` which extracts the first part of `process.env.npm_config_user_agent`, this is then added as the first check in the `isYarnProject` and `isPnpmProject` functions. 
* **Note:** This also removes the need to make a file system check in most cases.
* **Note:** `process.env.npm_config_user_agent` might not be available on all OS, and we still default back to initial behaviour for now.
* Rebuilt, packaged, and tested the change by printing the results of `getProjectTypeFromUserAgent` at start of CLI code.

Manual CLI test output:
```
howardtang hardhat-test % pnpm dlx $(pwd)/hardhat-2.22.19.tgz init
 WARN  3 deprecated subdependencies found: ethereumjs-abi@0.6.8, glob@8.1.0, inflight@1.0.6
Packages: +243
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 243, reused 242, downloaded 1, added 243, done
aychtang/hardhat
userAgent pnpm/8.8.0 npm/? node/v20.8.0 darwin arm64
projectType pnpm
```

### DIFF
--- a/.changeset/hot-birds-cover.md
+++ b/.changeset/hot-birds-cover.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Use `npm_config_user_agent` to determine what package manager to use for project creation
--- a/packages/hardhat-core/src/internal/cli/project-creation.ts
+++ b/packages/hardhat-core/src/internal/cli/project-creation.ts
@@ -554,12 +554,24 @@ function isInstalled(dep: string) {
   return dep in allDependencies;
 }
 
+function getProjectTypeFromUserAgent() {
+  const userAgent = process.env.npm_config_user_agent;
+  // Get first part of user agent string
+  const [projectType] = userAgent?.split("/") ?? [];
+  return projectType;
+}
+
 async function isYarnProject() {
-  return fsExtra.pathExists("yarn.lock");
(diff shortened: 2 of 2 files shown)


galargh (approved):
Great job! Thank you for your contribution 🙇 I allowed myself to add a changeset to your PR. Once the CI is through, I'm going to merge this 🚀

PR #6446, opened 2025-03-04
fix: test the shouldUpdateDependency helper thoroughly

galargh:

---

This is a follow-up based on the comment here - https://github.com/NomicFoundation/hardhat/pull/6426#discussion_r1979582780

### DIFF
--- a/v-next/hardhat/src/internal/cli/init/init.ts
+++ b/v-next/hardhat/src/internal/cli/init/init.ts
@@ -3,7 +3,10 @@ import type { PackageJson } from "@nomicfoundation/hardhat-utils/package";
 
 import path from "node:path";
 
-import { HardhatError } from "@nomicfoundation/hardhat-errors";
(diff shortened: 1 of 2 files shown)


v-next/hardhat/src/internal/cli/init/init.ts
```diff
@@ -529,3 +528,34 @@ function showStarOnGitHubMessage() {
   console.log();
   console.log(chalk.cyan("     https://github.com/NomicFoundation/hardhat"));
 }
+
+// NOTE: This function is exported for testing purposes only.
```
kanej, line 532: Nit: I don't think you need the comment. I think our internal/external folder structure covers us.
We could look at adding better eslint rules for unused exports, but I think that is a general thing.

kanej (approved):
Looks good. Burn the woods down, I like it.

PR #6458, opened 2025-03-10
Prevent losing typechain types when compiling a subset of contracts

### LINKED ISSUE #6311: [Typechain] modify the plugin to generate types for a subset of contracts
### Describe the feature

### Search terms

_No response_

antico5:
When compiling a subset of contracts, only the generated artifacts were being passed to typechain, losing the types of any previously generated artifacts.

This PR changes that, so besides passing the current generated artifacts, it passes all available artifacts in the project.

I had to adapt the test cases a bit since apparently on windows the generated files have a different directory structure than in linux/mac when having multiple contracts.

Closes #6311

### DIFF
--- a/.changeset/sour-mayflies-speak.md
+++ b/.changeset/sour-mayflies-speak.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-typechain": patch
+---
+
+Fix typechain overwrite when compiling a subset of contracts
--- a/v-next/hardhat-typechain/src/internal/hook-handlers/solidity.ts
+++ b/v-next/hardhat-typechain/src/internal/hook-handlers/solidity.ts
@@ -13,8 +13,18 @@ export default async (): Promise<Partial<SolidityHooks>> => {
         artifacts: Map<CompilationJob, ReadonlyMap<string, string[]>>,
       ) => Promise<void>,
     ) {
-      const artifactsPaths = Array.from(artifacts.values()).flatMap(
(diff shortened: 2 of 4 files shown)


ChristopherDedominici:
Cool solution, I like it! 
The only downside is that we do have to generate types for all the contracts, even though only a portion is compiled, but I don’t see any other solution

ChristopherDedominici (approved):

antico5:
> Cool solution, I like it! The only downside is that we do have to generate types for all the contracts, even though only a portion is compiled, but I don’t see any other solution.
> 
> @alcuadrado wdyt?

I've checked typechain-ethers and there's no option to "merge" newly added artifacts into an existing type file (hardhat.d.ts). It just uses whatever is passed on allArtifacts

ChristopherDedominici:
> I think this PR was replaced by the (just merged) #6400

@alcuadrado @antico5 
No, [the PR](https://github.com/NomicFoundation/hardhat/pull/6400) that Pato mentioned is generating types only for the contracts and not for the test contracts. This one, however, is introducing the possibility to generate types only for a subset of the contracts.

In the current version of `hh-typechain`, types can only be generated for all contracts at once. However, this PR allows users to generate types for a specific portion of the contracts without losing the previously generated types for contracts that are not being compiled.

So these 2 PRs are addressing different issues.

@antico5 You do have merge conflicts, though, because the hook used has changed

alcuadrado (approved):

kanej:
I am closing this as we need to rethink our approach after the changes to partial recompilation.

PR #6464, opened 2025-03-11
Use default value for rpcCachePath when running solidity tests

### LINKED ISSUE #6459: Caching when forking in Solidity Test runs
Update Solidity test to pass in a caching location for fork caching:

https://github.com/NomicFoundation/edr/blob/a8b55ff5eb29e98b2285aa8abfa5935f250ccc40/crates/edr_napi/index.d.ts#L514

This can likely be done manually right now, but it should be passed in by default.

antico5:
`config.solidityTest.rpcCachePath` defaults to `edr-cache`, which is the directory that edr uses when running a node in fork mode.

Closes #6459


v-next/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -121,6 +121,11 @@ export async function resolveSolidityTestUserConfig(
   testsPath = typeof testsPath === "object" ? testsPath.solidity : testsPath;
   testsPath ??= "test";
 
+  const solidityTest = {
+    rpcCachePath: "edr-cache",
```
galargh, line 125: Shouldn't we be using a subdirectory under our [global hardhat cache directory](https://github.com/NomicFoundation/hardhat/blob/0ea038914eeec40eb6a7db532bed94a276ddda28/v-next/hardhat-utils/src/global-dir.ts#L28)?

Do you also happen to know why ["there are no cache cleanup implications."](https://github.com/NomicFoundation/hardhat/issues/6459#issuecomment-2715036312)? 

I'm a little confused as to who owns/is responsible for this cache dir. Is it hardhat or is it edr?

alcuadrado: cleaning is done by the user with `hardhat clean`. That cache is never automatically cleaned.

alcuadrado, line 125: This should be inside `config.paths.cache`. A subfolder I mean.

There should be a single cache directory in the root of the project.

alcuadrado (changes requested):
Left some comments re the location of the cache

v-next/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
galargh: Ah, sorry, I meant - shouldn't we use `hre.config.paths.cache` instead of the `edr-cache`?

v-next/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -121,6 +123,16 @@ export async function resolveSolidityTestUserConfig(
   testsPath = typeof testsPath === "object" ? testsPath.solidity : testsPath;
   testsPath ??= "test";
 
+  const defaultRpcCachePath = path.join(
+    resolvedConfig.paths.cache,
+    "edr-cache",
```
galargh, line 128: Should we just name it `edr` since it's going to be in a directory named `cache`? This is a non-blocking suggestion, feel free to take it or leave it.

```suggestion
    "edr",
```

galargh (approved):
Thank you 🙇

alcuadrado (approved):

PR #6466, opened 2025-03-12
chore: bump `@ignored/edr` to `0.10.0-alpha.1`

agostbiro:
Bump `@ignored/edr` to `0.10.0-alpha.1`. Only adapts the code to preserve the functionality after the interface change to `TestResult.stackTrace`, doesn't handle the new warnings.


galargh (approved):
Thank you! I'll take care of the handling of new warnings in a new PR.

PR #6467, opened 2025-03-12
Increase default http request timeout from 30s to 300s

### LINKED ISSUE #6439: Increase HTTP timeouts
The default timeout values in the HTTP Hardhat networks, and wherever we have HTTP/TCP timeouts should be higher.

We should consider aligning with unidici, (maybe something like 200s). We are currently closer to 20s.

This will require manually testing to confirm we have not introduced issues.

antico5:
Closes #6439 

Scanned the v3 codebase for all potential outgoing http requests. Since they all go through utils/request, it all cleanly resolves to a single default value for timeouts (which we use for connection, headers and body timeout). Increased it to match unidici's defaults.

### DIFF
--- a/.changeset/eleven-bees-serve.md
+++ b/.changeset/eleven-bees-serve.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-utils": patch
+---
+
+Increase http timeout
--- a/v-next/hardhat-utils/src/request.ts
+++ b/v-next/hardhat-utils/src/request.ts
@@ -23,7 +23,7 @@ import {
   handleError,
 } from "./internal/request.js";
 
-export const DEFAULT_TIMEOUT_IN_MILLISECONDS = 30_000;
+export const DEFAULT_TIMEOUT_IN_MILLISECONDS = 300_000; // Aligned with unidici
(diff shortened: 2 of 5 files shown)


schaable:
Isn't the original issue referring to the default HTTP network timeout, defined in `resolveHttpNetwork` within `v-next/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts`?

schaable (changes requested):
> Isn't the original issue referring to the default HTTP network timeout, defined in `resolveHttpNetwork` within `v-next/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts`?

I'll assume we're referring to the timeout defined in the config resolution, not the default timeout set in `hardhat-utils`, since the issue mentions a timeout of about 20s, which matches the value defined in config resolution.

antico5:
> Isn't the original issue referring to the default HTTP network timeout, defined in `resolveHttpNetwork` within `v-next/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts`?

Makes sense, I completely missed the default value there. Thanks for pointing it out

schaable (approved):
LGTM!

PR #6470, opened 2025-03-13
Error when using a low interval mining value and mismatching timestamp

### LINKED ISSUE #6012: Hardhat 3 timestamps are broken if `intervalMining` is set to less than 1000
This was reported by @SebastienGllmt in Telegram.

> timestamps are broken if intervalMining is set to less than 1000 on 3.0.0-next.16
> the block timestamp advances faster than real clock time

antico5:
Closes #6012 

I put this logic inside the `resolveUserConfig` hook since by that point it was already zod-checked. I also considered the `validateUserConfig` hook but if the config had wrong type, it would show both the warning and the zod error (also it could attempt to compare two non-number values).

Let me know if there's a better place to move the warning logic.

### DIFF
--- a/.changeset/tasty-deers-doubt.md
+++ b/.changeset/tasty-deers-doubt.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Warn when using a low interval mining value and mismatching timestamp
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
@@ -75,7 +75,6 @@ export async function resolveUserConfig(
 
   const resolvedNetworks: Record<string, NetworkConfig> = {};
 
-  // resolve edr and http networks
   for (const [networkName, networkConfig] of Object.entries(networks)) {
     if (networkConfig.type !== "http" && networkConfig.type !== "edr") {
       throw new HardhatError(HardhatError.ERRORS.NETWORK.INVALID_NETWORK_TYPE, {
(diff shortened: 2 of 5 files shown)


schaable:
> Closes #6012
> 
> I put this logic inside the `resolveUserConfig` hook since by that point it was already zod-checked. I also considered the `validateUserConfig` hook but if the config had wrong type, it would show both the warning and the zod error (also it could attempt to compare two non-number values).
> 
> Let me know if there's a better place to move the warning logic.

@antico5 I don't think this scenario is different from other config validation errors, so I'd avoid placing the logic outside the `validateUserConfig` hook. In v2, validation and resolution were spread across multiple places, making the code harder to maintain. We should aim for better organization in v3.

Since this validation depends on two different fields in the config, a good place for it would be `refineEdrNetworkUserConfig`, which runs after individual network config fields are validated.

Regarding the warning, I'd suggest using a zod error instead, as the configuration isn't valid and will cause issues. @kanej wdyt?

antico5:
> > Closes #6012
> > I put this logic inside the `resolveUserConfig` hook since by that point it was already zod-checked. I also considered the `validateUserConfig` hook but if the config had wrong type, it would show both the warning and the zod error (also it could attempt to compare two non-number values).
> > Let me know if there's a better place to move the warning logic.
> 
> @antico5 I don't think this scenario is different from other config validation errors, so I'd avoid placing the logic outside the `validateUserConfig` hook. In v2, validation and resolution were spread across multiple places, making the code harder to maintain. We should aim for better organization in v3.
> 
> Since this validation depends on two different fields in the config, a good place for it would be `refineEdrNetworkUserConfig`, which runs after individual network config fields are validated.
> 
> Regarding the warning, I'd suggest using a zod error instead, as the configuration isn't valid and will cause issues. @kanej wdyt?

I'd agree using the `validate` step if this was meant to throw an error, but on the issue discussion they concluded to use just a warning. 

I can still move the logic to the validate step and just duplicate the type checking for the interval value. Will push a commit changing that and the test cases

fvictorio:
> I'd agree using the `validate` step if this was meant to throw an error, but on the issue discussion they concluded to use just a warning.

A warning for HH2, mainly to avoid a (theoretical) breaking change. But I think in HH3 this should be an error.

alcuadrado:
Actually, @schaable convinced me that we should change this to an error

alcuadrado (changes requested):
Let's make it an error

antico5:
@alcuadrado ready for re-review

schaable (approved):

PR #6480, opened 2025-03-17
Restructure validation tests and add tests for missing fields

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat/issues/6198

### DIFF
--- a/v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
@@ -1,6 +1,9 @@
 import type {
   EdrNetworkConfigOverride,
+  EdrNetworkUserConfig,
+  HardhatUserConfig,
   HttpNetworkConfigOverride,
+  HttpNetworkUserConfig,
   NetworkConfig,
 } from "../../../../src/types/config.js";
 import type { NetworkHooks } from "../../../../src/types/hooks.js";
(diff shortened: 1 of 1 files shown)


v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -355,431 +363,1943 @@ describe("NetworkManagerImplementation", () => {
...
+            },
+          ]);
         });
 
-        it("should allow valid private keys with missing hex prefix", async () => {
-          networkConfig.accounts = [
-            "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+        it("should fail with invalid HttpNetworkHDAccountsConfig", async () => {
+          const accountsValuesToTest: Array<
+            [Partial<any>, ExpectedValidationError[]]
```
schaable, line 542: isn't `Partial<any>` pretty much the same as `any`?

v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -355,431 +363,1943 @@ describe("NetworkManagerImplementation", () => {
...
+                { mnemonic: "valid", initialIndex: 1, count: 1, path: 123 },
+                [
+                  {
+                    path: ["networks", "hardhat", "accounts", "path"],
+                    message: "Expected string, received number",
+                  },
+                ],
+              ],
+              // todo: is this expected? the zod type has every field optional, so passing an object with an unrelated field doesn't fail
+              [
```
schaable, line 896: I believe this is Zod’s default behavior for unrecognized keys in an object:

> By default, Zod object schemas strip out unrecognized keys during parsing. You can disallow unknown keys with .strict().

I don’t see this as an issue because extra keys won’t be used by hardhat and might even be needed by plugins in the future. Enabling `.strict()` would prevent plugins from extending the configuration.

schaable (approved):
Looks great! Left 2 small comments, but consider it approved.

v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
zoeyTM: Yes, this appears to be a mistake. I'll push an update!

PR #6490, opened 2025-03-20
[Investigation] Config overrides should be based on merging the user config, not the resolved

ChristopherDedominici:


v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
```diff
@@ -5,6 +5,10 @@ export default async (): Promise<Partial<HardhatRuntimeEnvironmentHooks>> => ({
   created: async (context, hre) => {
     let networkManager: NetworkManager | undefined;
 
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- TMP
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- TMP
+    const userConfigNetworks = (hre as any).userConfig.networks;
```
ChristopherDedominici, line 10: `hre` already have `userConfig.networks` but it does not currenlty exist in the type definition (I wonder why?)

v-next/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts
```diff
@@ -149,6 +149,28 @@ export function resolveNetworkConfigOverride(
   );
 }
 
+export function resolveNetworkConfig(
```
ChristopherDedominici, line 152: an helper method to resolve the configuration based on the `type` property. I didn't fin an already existing one

v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
ChristopherDedominici: this configuration is NOT resolved yet

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -111,17 +111,15 @@ export class NetworkManagerImplementation implements NetworkManager {
         );
       }
 
-      const normalizedNetworkConfigOverride =
-        await normalizeNetworkConfigOverride(
-          networkConfigOverride,
-          this.#networkConfigs[resolvedNetworkName],
-        );
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/consistent-type-assertions -- tmp
+      const newConfig: NetworkUserConfig = {
```
ChristopherDedominici, line 115: since the configurations are NOT resolved, a merge between object using `...` is possible

v-next/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts
schaable: this is fine, but you could improve it further by also using it in the hook handler

v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
schaable: wait, you're saying that we're setting `hre.userConfig` somewhere in the code?

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
schaable: the problem is that spread only does a shallow merge:

```
a = { b: { c: 1 } }
a1 = { b: { d: 1 }, e: 3 }
console.log({ ...a, ...a1 }) // { b: { d: 1 }, e: 3 }
```

If you want to override just mining.interval, you’d ideally only pass { mining: { interval: 123 } } as an override to the connect call. Using a spread forces you to pass all the mining properties, even if you only want to change one.

v-next/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts
```diff
@@ -109,82 +106,6 @@ export function resolveEdrNetwork(
...
- * This function processes the network configuration override using the appropriate
- * resolver (either HTTP or EDR) and ensures only the values explicitly specified
- * in the override are included in the final result, preventing defaults from
- * overwriting the user's configuration.
- *
- * @param networkUserConfigOverride The user's network configuration override.
- * @param resolveConfigurationVariable A function to resolve configuration variables.
- * @returns A Partial<NetworkConfig> containing the resolved values from the override.
- */
-export function resolveNetworkConfigOverride(
```
ChristopherDedominici, line 123: This method was removed as the validation logic has been simplified. 
The user config is now merged with the Hardhat config and fully re-validated (so `pickResolvedFromOverrides` can be deleted too). Instead of selectively resolving overrides, it's sufficient to use `resolveHttpNetwork` or `resolveEdrNetwork`.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -111,17 +112,14 @@ export class NetworkManagerImplementation implements NetworkManager {
         );
       }
 
-      const normalizedNetworkConfigOverride =
-        await normalizeNetworkConfigOverride(
-          networkConfigOverride,
-          this.#networkConfigs[resolvedNetworkName],
-        );
+      const newConfig = mergeConfigOverride(
```
alcuadrado, line 115: I think this logic, `mergeConfigOverride` is probably now outdated. What do you think, @schaable ?

alcuadrado: I think how we could do with a deep merge, without anything fancy.

ChristopherDedominici: It doesn't seem like we have a deep merge in `hh-utils`. Should I add one?

schaable: I would say go for it, but keep it simple. You can probably reuse the `mergeConfigOverride` code. I'd leave functions, symbols, and classes out, since we probably don't want to merge those.

v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
```diff
@@ -1,10 +1,15 @@
 import type { HardhatRuntimeEnvironmentHooks } from "../../../../types/hooks.js";
 import type { NetworkManager } from "../../../../types/network.js";
+import type { HardhatRuntimeEnvironmentImplementation } from "../../../core/hre.js";
 
 export default async (): Promise<Partial<HardhatRuntimeEnvironmentHooks>> => ({
   created: async (context, hre) => {
     let networkManager: NetworkManager | undefined;
 
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- although `userConfig` is present on the hre, it isn’t typed in the official definition
+    const userConfigNetworks = (hre as HardhatRuntimeEnvironmentImplementation)
```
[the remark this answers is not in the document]
schaable: @alcuadrado can you clarify this?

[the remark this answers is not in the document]
alcuadrado: oh, this was clarified on slack. This should be properly declared in the HRE type.

schaable (approved):

PR #6492, opened 2025-03-23
Replace the usual banner with an `AlphaBanner` component

alcuadrado:


alcuadrado:
@fzeoli @kanej @fvictorio I think this works correctly, but can you double check (manually / visually) before we merge it?

@fzeoli, the design didn't specify at which width I should jump between the two layouts. I opted for `smd` , which is `592px`, using the smaller banner, and everything above that the full banner. We can also pick `md` (`1000px`) for that cut.

docs/src/components/ui/AlphaBanner.tsx
```diff
@@ -0,0 +1,103 @@
...
+// depending on dark/light mode.
+const alphaBannerBackground = "#4F00A3" as const;
+const alphaBannerText = "#F2F2F2" as const;
+const alphaBannerCTA = "#C4C4C4" as const;
+const alphaBannerBraket = "#EDCF00" as const;
+
+const AlphaBannerContainer = styled.section`
+  font-family: ChivoRegular, sans-serif;
+  user-select: none;
+  z-index: 100;
```
kanej, line 16: This z-index interferes with the popup from `TOOLS` which shows on hover over the tools button:

image


```suggestion
  z-index: 5;
```

kanej (changes requested):
I did a manual test using chrome set to laptop/ipad/mobile to confirm that the banner works across our pages.
It is present on every page including the home page.
I found one issue which was that the tools menu item has a popup menu that is obscured by the banner. I have proposed a change reducing its z-index. A quick look at the codebase suggests a z-index of 5 will make it fall within our expected use case (animated things should appear above it).

fvictorio (approved):

kanej (approved):

PR #6496, opened 2025-03-24
Prepare for next Node LTS version - branch: v-next

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/6441

Intermediate step: support node v23 while waiting for v24

PR #6497, opened 2025-03-24
Prepare for next Node LTS version - branch: main

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/pull/6496

Intermediate step: support node v23 while waiting for v24.

Note:
I had to reorganize a few files in `ignition` because class syntax like `private` has been replaced with the `#` prefix.
Example:
```
private _blockNumber = 10;

becomes:

#blockNumber = 10;
```


packages/hardhat-solpp/test/tests.ts
```diff
@@ -102,22 +102,23 @@ describe("Solpp plugin", function () {
...
-
-        await expectErrorAsync(async () =>
-          this.env.run("hardhat-solpp:run-solpp", {
-            files,
-            opts,
-          })
-        );
-      });
-    });
+    // The test is commented out because otherwise it will fail the test suite in WIndows after the update to node 24
```
ChristopherDedominici, line 105: This test was already skipped, but after the update to Node 24, it somehow caused the entire test suite to fail on Windows. See here: https://github.com/NomicFoundation/hardhat/actions/runs/15000928741/job/42147457507

packages/hardhat-chai-matchers/test/contracts.ts
```diff
@@ -1,9 +1,12 @@
 import {
   BaseContract,
-  BaseContractMethod,
+  // eslint-disable-next-line prettier/prettier
```
ChristopherDedominici, line 3: Needed because otherwise the lint task will fail

packages/hardhat-ethers/test/helpers.ts
```diff
@@ -1,5 +1,6 @@
 import { assert } from "chai";
-import { ContractRunner, Signer } from "ethers";
+// eslint-disable-next-line prettier/prettier
```
ChristopherDedominici, line 2: As above: needed because otherwise the lint task will fail

packages/hardhat-core/src/internal/cli/constants.ts
```diff
@@ -1 +1 @@
-export const SUPPORTED_NODE_VERSIONS = ["^18.0.0", "^20.0.0", "^22.0.0"];
+export const SUPPORTED_NODE_VERSIONS = ["^20.0.0", "^22.0.0", "^24.0.0"];
```
alcuadrado, line 1:
```suggestion
// TODO: Remove 18 on June 30, 2025
export const SUPPORTED_NODE_VERSIONS = ["^18.0.0", "^20.0.0", "^22.0.0", "^24.0.0"];
```

According to [this](https://hardhat.org/hardhat-runner/docs/reference/stability-guarantees) we still need to support it. Other than that, this looks good.

alcuadrado: I guess there are some tests below that would need to be updated

packages/hardhat-solpp/test/tests.ts
```diff
@@ -102,22 +102,23 @@ describe("Solpp plugin", function () {
...
-
-        await expectErrorAsync(async () =>
-          this.env.run("hardhat-solpp:run-solpp", {
-            files,
-            opts,
-          })
-        );
-      });
-    });
+    // The test is commented out because it causes the test suite to fail on Windows after the update to Node 24
```
alcuadrado, line 105: Ok with this, this plugin is defacto deprecated

packages/hardhat-core/src/internal/cli/constants.ts
alcuadrado: No need to read 18 to the ci though

alcuadrado (approved):
The `#private` thing is unexpected, but ok.

I left a small comment. Apart from that, it looks ready.

PR #6499, opened 2025-03-24
Fix infinite loop in `SolcConfigSelector`

alcuadrado:
😳


galargh (approved):
Nice one! Great find 🥇

PR #6504, opened 2025-03-25
Resolve all dependencies when using submodules in `after`

zoeyTM:
This issue has two commits:

1. The first commit resolves the original issue (https://github.com/NomicFoundation/hardhat-ignition/issues/857) by recursively collecting all futures from all submodules of a module given as a dependency in an `after` option of a future.
2. The second commit extends the above logic to apply to futures passed as dependencies in an `after` option as well. The reason for this change is to cover what I'm calling a combined "malaga-index" case:

```typescript
const IndexMod = buildModule("IndexMod", (m) => {
  const { F1 } = m.useModule(Mod1);
  const { F2 } = m.useModule(Mod2);
  const { F3 } = m.useModule(Mod3);

  const lastCall = m.call(F3, "mustBeCalledByTwoSeparateContracts");
  const NewF1 = m.contractAt("F1", F1, { after: [lastCall], id: "NewF1" });

  return { NewF1, F2, F3 };
});

const moduleDefinition = buildModule("DeployModule", (m) => {
  const { NewF1 } = m.useModule(IndexMod);

  const failingCall = m.call(NewF1, "throwsIfNotCalledTwice");

  return {};
});
```

In the above example, before the fix in the second commit, the final future calling the function `throwsIfNotCalledTwice` would fail because the second call to `mustBeCalledByTwoSeparateContracts` inside `Mod2` had not completed yet as there was no direct dependency link between that call and `failingCall`. It's a similar problem to the original issue, however this one would still occur with the original fix if the failing call was depending on a future inside the index module instead of the index module itself (the malaga rule). 

This fix essentially applies the malaga rule logic to the new index module assumption that any submodule included in a module that is used as a dependency should have all of its futures and submodules futures and so on added to the original futures dependencies.

### DIFF
--- a/packages/hardhat-ignition-core/src/internal/utils/adjacency-list.ts
+++ b/packages/hardhat-ignition-core/src/internal/utils/adjacency-list.ts
@@ -32,6 +32,16 @@ export class AdjacencyList {
     this._list.set(from, toSet);
   }
 
+  public hasDependency({ from, to }: { from: string; to: string }): boolean {
+    const toSet = this._list.get(from);
+
+    if (toSet === undefined) {
+      return false;
+    }
+
+    return toSet.has(to);
+  }
+
   public deleteDependency({ from, to }: { from: string; to: string }) {
     const toSet = this._list.get(from) ?? new Set<string>();
 
--- a/packages/hardhat-ignition-core/src/internal/utils/adjacency-list.ts
+++ b/packages/hardhat-ignition-core/src/internal/utils/adjacency-list.ts
@@ -32,16 +32,6 @@ export class AdjacencyList {
     this._list.set(from, toSet);
   }
 
-  public hasDependency({ from, to }: { from: string; to: string }): boolean {
-    const toSet = this._list.get(from);
-
-    if (toSet === undefined) {
-      return false;
-    }
-
-    return toSet.has(to);
-  }
-
   public deleteDependency({ from, to }: { from: string; to: string }) {
     const toSet = this._list.get(from) ?? new Set<string>();
 
--- a/.changeset/many-falcons-relate.md
+++ b/.changeset/many-falcons-relate.md
@@ -1,0 +1,6 @@
+---
+"@nomicfoundation/ignition-core": patch
+"@nomicfoundation/hardhat-ignition": patch
+---
+
+Resolve all dependencies when using submodules in `after`
--- a/packages/hardhat-ignition-core/src/internal/batcher.ts
+++ b/packages/hardhat-ignition-core/src/internal/batcher.ts
@@ -130,29 +130,7 @@ export class Batcher {
   ): boolean {
     const dependencies = batchState.adjacencyList.getDependenciesFor(futureId);
 
-    return [...dependencies].every((depId) => {
-      // We distinguish between module and future ids here, as the future's always have `#` and the modules don't.
-      if (/#/.test(depId)) {
-        return batchState.visitState[depId] === VisitStatus.VISITED;
-      }
-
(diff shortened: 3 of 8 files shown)


packages/hardhat-ignition-core/src/internal/utils/adjacency-list.ts
```diff
@@ -32,6 +32,16 @@ export class AdjacencyList {
     this._list.set(from, toSet);
   }
 
+  public hasDependency({ from, to }: { from: string; to: string }): boolean {
```
alcuadrado, line 35: I think this isn't used, right?

alcuadrado:
Great work, @zoeyTM!

What do you think about this diff that would simplify the batcher? 

Basically, it only adds futures to the adjacency list graph, so part of the batcher can be deleted. 

I run the tests and they pass.

```diff
diff --git a/packages/hardhat-ignition-core/src/internal/batcher.ts b/packages/hardhat-ignition-core/src/internal/batcher.ts
index 26bb2348d..3c4ef0b90 100644
--- a/packages/hardhat-ignition-core/src/internal/batcher.ts
+++ b/packages/hardhat-ignition-core/src/internal/batcher.ts
@@ -130,29 +130,7 @@ export class Batcher {
   ): boolean {
     const dependencies = batchState.adjacencyList.getDependenciesFor(futureId);
 
-    return [...dependencies].every((depId) => {
-      // We distinguish between module and future ids here, as the future's always have `#` and the modules don't.
-      if (/#/.test(depId)) {
-        return batchState.visitState[depId] === VisitStatus.VISITED;
-      }
-
-      return this._checkModuleDependencyIsComplete(depId, batchState);
-    });
-  }
-
-  /**
-   * This is needed because moduleIds are not present in the visit state
-   * causing an infinite loop when checking whether a depenedency is visited if that dependency is a module.
-   */
-  private static _checkModuleDependencyIsComplete(
-    moduleId: string,
-    batchState: BatchState
-  ) {
-    const dependencies = Object.keys(batchState.visitState).filter((futureId) =>
-      futureId.startsWith(moduleId)
-    );
-
-    return dependencies.every(
+    return [...dependencies].every(
       (depId) => batchState.visitState[depId] === VisitStatus.VISITED
     );
   }
diff --git a/packages/hardhat-ignition-core/src/internal/utils/adjacency-list-converter.ts b/packages/hardhat-ignition-core/src/internal/utils/adjacency-list-converter.ts
index 6ec84ac2a..ef3a3dad0 100644
--- a/packages/hardhat-ignition-core/src/internal/utils/adjacency-list-converter.ts
+++ b/packages/hardhat-ignition-core/src/internal/utils/adjacency-list-converter.ts
@@ -16,7 +16,11 @@ export class AdjacencyListConverter {
 
     for (const future of futures) {
       for (const dependency of future.dependencies) {
-        dependencyGraph.addDependency({ from: future.id, to: dependency.id });
+        // We only add Futures to the dependencry graph, modules are handled
+        // in method call below, by adding their futures to the graph.
+        if (isFuture(dependency)) {
+          dependencyGraph.addDependency({ from: future.id, to: dependency.id });
+        }
 
         this._optionallyAddDependenciesFromSubmodules(
           dependencyGraph,
```

alcuadrado:
PS: I can also push a commit with that diff if you want. I have it locally.

packages/hardhat-ignition-core/src/internal/utils/adjacency-list.ts
zoeyTM: Ah yep, good catch!

Issue #6505, opened 2025-03-26
Show inferred errors in Solidity stack traces


fvictorio:
Inferred errors are not shown in Solidity stack traces. EDR should be returning them, so I think the problem is that Hardhat is not using them.

Here's a minimal example. Take this contract:

```solidity
interface Bar {
  function g() external;
}

contract Foo {
  function f() public {
    Bar(0x1111111111111111111111111111111111111111).g();
  }
}
```

If you write a Solidity test for it:

```solidity
import { Foo } from "./Foo.sol";

contract FooTest {
    function test() public {
        Foo foo = new Foo();
        foo.f();
    }
}
```

Then this test will fail with the following error:

```
Reason: EvmError: Revert
  at Foo.f (contracts/Foo.sol:10)
  at FooTest.test (contracts/Foo.t.sol:9)
```

But if you test it with typescript:

```ts
import { it } from "node:test";
import { network } from "hardhat";

it("Foo", async function () {
  const { viem } = await network.connect()

  const foo = await viem.deployContract("Foo")

  await foo.write.f();
})
```

You'll get the inferred error:

```
       Details: Transaction reverted: function call to a non-contract account
```

galargh:
Resolved in https://github.com/NomicFoundation/hardhat/pull/6506

Issue #6475, opened 2025-03-17
Handle new warnings produced by EDR stack trace


galargh:
Follow-up to https://github.com/NomicFoundation/hardhat/pull/6466

galargh:
Resolved in https://github.com/NomicFoundation/hardhat/pull/6506

PR #6506, opened 2025-03-26
feat: support stack trace warnings and infer error messages from stack trace

galargh:

---

This is a follow-up to https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1742980461593939

It resolved https://github.com/NomicFoundation/hardhat/issues/6505 and https://github.com/NomicFoundation/hardhat/issues/6475

Here's an example run which showcases both error messages inferred from stack trace and new stack trace warnings - https://github.com/galargh/hardhat/actions/runs/14082082256/job/39437613164

Please note that the notes are not part of the actual output.

```
...
✘ Run Failed: 443 tests, 440 passed, 3 failed, 0 skipped (duration: 12062 ms)

Failure (1): setUp()
# 💁 NOTE: This message would have been previously EvmError: Revert
Reason: Transaction reverted without a reason string
  at TokenLocker.constructor (test/vault/VaultReentrancy.t.sol:22)
  at TokenLocker.<unknown> (test/vault/VaultReentrancy.t.sol:18)

Failure (2): test_getTickAtSqrtRatio_matchesJavascriptImplWithin1()
Reason: EvmError: Revert
# 💁 NOTE: This is a new entry in the test failure details
Stack Trace Warning: The test is not safe to replay because it uses impure cheatcodes: function ffi(string[] calldata commandInput) external returns (bytes memory result);

Failure (3): test_getSqrtRatioAtTick_matchesJavaScriptImplByOneHundrethOfABip()
Reason: EvmError: Revert
Stack Trace Warning: The test is not safe to replay because it uses impure cheatcodes: function ffi(string[] calldata commandInput) external returns (bytes memory result);
```

### DIFF
--- a/.changeset/pretty-lizards-roll.md
+++ b/.changeset/pretty-lizards-roll.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Display solidity stack trace generation warnings to the users and try to infer better error messages from said stack traces
--- a/.changeset/pretty-lizards-roll.md
+++ b/.changeset/pretty-lizards-roll.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Display solidity stack trace generation warnings to the users and try to infer better error messages from said stack traces
+Infer and display better error messages from stack traces ([#6505](https://github.com/NomicFoundation/hardhat/issues/6505))
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
@@ -126,7 +126,7 @@ export async function* testReporter(
 
       let reason: string | undefined;
       if (stackTrace?.kind === "StackTrace") {
-        reason = getMessageFromLastStackTraceEntry(stackTrace.entries[0]);
+        reason = getMessageFromLastStackTraceEntry(stackTrace.entries[stackTrace.entries.length - 1]);
       }
       if (reason === undefined || reason === "") {
         reason = failure.reason ?? "Unknown error";
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
@@ -126,7 +126,9 @@ export async function* testReporter(
 
       let reason: string | undefined;
       if (stackTrace?.kind === "StackTrace") {
-        reason = getMessageFromLastStackTraceEntry(stackTrace.entries[stackTrace.entries.length - 1]);
+        reason = getMessageFromLastStackTraceEntry(
+          stackTrace.entries[stackTrace.entries.length - 1],
+        );
       }
       if (reason === undefined || reason === "") {
         reason = failure.reason ?? "Unknown error";
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
@@ -8,12 +8,10 @@ import type { TestResult } from "@ignored/edr";
 import { bytesToHexString } from "@nomicfoundation/hardhat-utils/hex";
 import chalk from "chalk";
 
-import {
-  encodeStackTraceEntry,
-  getMessageFromLastStackTraceEntry,
-} from "../network-manager/edr/stack-traces/stack-trace-solidity-errors.js";
+import { encodeStackTraceEntry } from "../network-manager/edr/stack-traces/stack-trace-solidity-errors.js";
 
 import { formatArtifactId } from "./formatters.js";
+import { getMessageFromLastStackTraceEntry } from "./stack-trace-solidity-errors.js";
 
 /**
  * This is a solidity test reporter. It is intended to be composed with the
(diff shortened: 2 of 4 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -119,33 +122,46 @@ export async function* testReporter(
...
 
-        if (stackTraceStack.length > 0) {
-          yield `${chalk.grey(stackTraceStack.join("\n"))}\n`;
-        }
+          if (stackTraceStack.length > 0) {
+            yield `${chalk.grey(stackTraceStack.join("\n"))}\n`;
+          }
+          break;
+        case "UnexpectedError":
+          yield `Stack Trace Warning: ${chalk.grey(stackTrace.errorMessage)}\n`;
```
agostbiro, line 154: This used to throw before. I think it's ok to turn it into a warning, but wanted to make sure we're all on the same page about this.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -119,33 +122,46 @@ export async function* testReporter(
...
-        failure.reason !== undefined &&
-        failure.reason !== null &&
-        failure.reason !== ""
-      ) {
-        yield `Reason: ${chalk.grey(failure.reason)}\n`;
+      const stackTrace = failure.stackTrace();
+
+      let reason: string | undefined;
+      if (stackTrace?.kind === "StackTrace") {
+        reason = getMessageFromLastStackTraceEntry(stackTrace.entries[stackTrace.entries.length - 1]);
```
agostbiro: The messages from `getMessageFromLastStackTraceEntry` are written with a Hardhat Network use-case in mind and they can be a bit weird in a Solidity-testing context. There is more discussion of this here: https://github.com/NomicFoundation/hardhat/pull/6235#discussion_r1946455490

For me it's ok to merge like this, but wanted to make sure we discuss this (cc @fvictorio )

agostbiro (approved):
Thanks, LGTM, but I made two comments where other reviewers might have different conclusions.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
galargh: I quickly created a duplicate of the original message inference function in solidity-test plugin and run a first pass at the error message simplification. I haven't yet fully analysed which even make sense in the context of solidiy-tests and which don't but I wanted to put it out there as quickly as possible because I think there might be value into opening up these deliberations to a wider audience. Here's the commit in question - https://github.com/NomicFoundation/hardhat/pull/6506/commits/543ea3c6862dfb1ad934325e2fb172d216e57809

.changeset/pretty-lizards-roll.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Display solidity stack trace generation warnings to the users and try to infer better error messages from said stack traces
```
kanej, line 5:
```suggestion
Infer and display better error messages from stack traces ([#6505](https://github.com/NomicFoundation/hardhat/issues/6505))
```

fvictorio:
I only skimmed the code, but I tested this with the same example I used in #6505 and now it works nicely!

```
Failure (1): test()
Reason: Function call to a non-contract account
  at Foo.f (contracts/Foo.sol:10)
  at FooTest.test (contracts/Foo.t.sol:9)
```

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
alcuadrado: Is this an unexpected error in the stack traces generation? Or when? Maybe the name can be more explicit.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -119,33 +120,48 @@ export async function* testReporter(
...
+          break;
+        case "UnsafeToReplay":
+          if (stackTrace.globalForkLatest === true) {
+            yield `Stack Trace Warning: ${chalk.grey("The test is not safe to replay because a fork url without a fork block number was provided.")}\n`;
+          }
+          if (stackTrace.impureCheatcodes.length > 0) {
+            yield `Stack Trace Warning: ${chalk.grey(`The test is not safe to replay because it uses impure cheatcodes: ${stackTrace.impureCheatcodes.join(", ")}`)}\n`;
+          }
+          break;
+        case "HeuristicFailed":
```
alcuadrado, line 162: What does this mean, @agostbiro? Does it make sense to just ignore it? How is it different than `UnexpectedError`?

alcuadrado (approved):
I left a few questions, but in general this looks good to me

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
galargh: We get `HeuristicFailed` when the process of generating stack trace goes through, but the stack trace is empty i.e. we couldn't find any useful information. We don't get any additional information on why that might have been though. We could inform the user about it but I'm not sure if there's anything actionable on their part we could display.

`UnexpectedError`, on the other hand, happens when the stack trace generation throws.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
galargh: Yes, that's during stack trace generation.

PR #6510, opened 2025-03-27
[Hardhat-utils] Throw an exception if the mutex times out

ChristopherDedominici:
Proposal:
In this PR, I added logic to throw an error if the mutex times out. This can help identify issues related to function timeouts.

Wdyt? @alcuadrado @galargh

### DIFF
--- a/v-next/hardhat-utils/src/synchronization.ts
+++ b/v-next/hardhat-utils/src/synchronization.ts
@@ -15,6 +15,7 @@ import debug from "debug";
 
 import { ensureError } from "./error.js";
 import { FileSystemAccessError } from "./errors/fs.js";
+import { MutexTimeoutError } from "./errors/synchronization.js";
 import { sleep } from "./lang.js";
 
 const log = debug("hardhat:util:multi-process-mutex");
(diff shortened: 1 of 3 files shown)


v-next/hardhat-utils/src/synchronization.ts
```diff
@@ -76,16 +77,38 @@ export class MultiProcessMutex {
   async #executeFunctionAndReleaseMutex<T>(f: () => Promise<T>): Promise<T> {
     log(`Mutex acquired at path '${this.#mutexFilePath}'`);
 
+    const controller = new AbortController();
+
     try {
-      return await f();
+      // `Promise.race` returns as soon as the first promise settles (resolves or rejects).
+      // This means `f()` will continue running in the background even if `#throwOnMutexTimeout`
```
galargh, line 84: To confirm, this is the behaviour we want to keep, right?

galargh (commented):
I like this direction 👍 One thing that I think would be worth changing is getting rid of `#isMutexFileTooOld()` since `#executeFunctionAndReleaseMutex` should always release the mutex within the set timeout now.

alcuadrado:
I don't understand this change. It isn't a timeout on the mutex itself, but rather on the execution of the function protected by the mutex.

Shouldn't throwing on timeout of the mutex be done in here

```ts
        // If the mutex file is too old, it likely indicates a stale lock, so the file should be removed
        log(
          `Current mutex file is too old, removing it at path '${this.#mutexFilePath}'`,
        );
        this.#deleteMutexFile();
```

I guess I'm not understanding the motivation, which are the errors we are trying to improve/debug/understand?

PR #6511, opened 2025-03-27
Custom error without `Interface` parameter

### LINKED ISSUE #6484: Support custom error selectors in `revertedWithCustomError` Chai matcher
### Describe the feature

What is the reasoning behind requiring an interface to be passed to `revertedWithCustomError`?  The error name string itself should be enough to run the assertion.

In my case, the contract I'm testing takes a custom error selector as a parameter and reverts with that error using assembly.  The errors do not exist in the contract ABI, so I have to pass in a fake interface to the matcher.

```typescript
const customError = 'TestError';

const placeholder = {
  interface: new ethers.Interface([`error ${ customError }()`]),
};

const selector = placeholder.interface.getError(customError)?.selector!;

await expect(
  contract.revertWithCustomErrorSelector(selector),
).to.be.revertedWithCustomError(placeholder, customError);
```

If this process could be encapsulated within the matcher code, the matcher could be used like this:

```typescript
await expect(...).to.be.revertedWithCustomError(customError);
// or
await expect(...).to.be.revertedWithCustomError(selector);
```

Maybe the `withArgs` extension needs the full interface to process its arguments.  In that case, maybe require the full error signature?

```typescript
await expect(...).to.be.revertedWithCustomError('TestError(uint256)`);
```

### Search terms

_No response_

ItsNickBarry:
partially resolves #6484

Draft because I haven't been able to test - after running `pnpm build`, I tried `yarn link` and installing the tarball generated by `pnpm pack`, but the package doesn't work at all (even before the changes in this PR). If there's a build step I'm missing, please let me know.  Or just close this if you all are sure you don't want to improve the custom error matchers.

### DIFF
--- a/packages/hardhat-chai-matchers/src/internal/constants.ts
+++ b/packages/hardhat-chai-matchers/src/internal/constants.ts
@@ -9,5 +9,7 @@ export const EMIT_MATCHER = "emit";
 export const REVERTED_MATCHER = "reverted";
 export const REVERTED_WITH_MATCHER = "revertedWith";
 export const REVERTED_WITH_CUSTOM_ERROR_MATCHER = "revertedWithCustomError";
+export const REVERTED_WITH_CUSTOM_ERROR_SIG_MATCHER =
+  "revertedWithCustomErrorSig";
 export const REVERTED_WITH_PANIC_MATCHER = "revertedWithPanic";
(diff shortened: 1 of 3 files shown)


kanej:
Hey @ItsNickBarry, I am going to close this PR. If you are wanting to push it forward let us know.

Issue #6503, opened 2025-03-25
Flaky test in CI


ChristopherDedominici:
This is the 2nd time I get this random error in the CI, link to failed job: https://github.com/NomicFoundation/hardhat/actions/runs/14064321889/job/39382851902?pr=6490.

And also here: https://github.com/NomicFoundation/hardhat/actions/runs/14070087338/job/39402074162?pr=6491

Error:

![Image](https://github.com/user-attachments/assets/396c18e2-d4ef-44ae-81fb-6e8fa7cbdff8)

galargh:
Resolved in https://github.com/NomicFoundation/hardhat/pull/6512

PR #6512, opened 2025-03-27
test: remove timeouts from the object cache test

galargh:

---

This resolves https://github.com/NomicFoundation/hardhat/issues/6503

I ran the test after changes a 100 times on each runner:
- https://github.com/NomicFoundation/hardhat/actions/runs/14118656535
- https://github.com/NomicFoundation/hardhat/actions/runs/14118548054

### DIFF
--- a/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/cache.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/cache.ts
@@ -5,6 +5,7 @@ import { beforeEach, describe, it } from "node:test";
 
 import { useTmpDir } from "@nomicfoundation/hardhat-test-utils";
 import {
+  getAccessTime,
   getAllFilesMatching,
   getFileSize,
 } from "@nomicfoundation/hardhat-utils/fs";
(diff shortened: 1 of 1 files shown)


v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -69,42 +70,44 @@ describe("ObjectCache", () => {
...
       const filesAfter = await getAllFilesMatching(cachePath);
       assert.deepEqual(filesAfter, []);
     });
 
-    it("should remove everything with the max size set to 0", async () => {
+    it("should remove everything with the max size set to -1", async () => {
       const filesBefore = await getAllFilesMatching(cachePath);
       assert.notDeepEqual(filesBefore, []);
-      await cache.clean(undefined, 0);
+      await cache.clean(undefined, -1);
```
galargh, line 86: This is unrelated to flakiness. I just thought it would be a more accurate test as technically we could have an 0 byte entry in the cache so setting max size to 0 wouldn't remove it.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -69,42 +70,44 @@ describe("ObjectCache", () => {
...
       const filesAfter = await getAllFilesMatching(cachePath);
-      assert.notDeepEqual(filesAfter, []);
+      // NOTE: It would be nice to check whether we removed some but not all files,
+      // but that check has proven to be flaky, especially on Intel macOS, due to
+      // its heavy reliance on timing of the operations.
       assert.notDeepEqual(filesAfter, filesBefore);
     });
 
     it("should remove something with the max size set to some value", async () => {
+      await cache.set(randomUUID(), testValue);
```
galargh, line 108: This wasn't flaky before either. I just noticed the logic in this test can be simplified without compromising what we actually want to test here.

ChristopherDedominici (approved):

PR #6524, opened 2025-03-31
feat: add watcher to hardhat node

galargh:

---

This PR adds build info watcher to the node task. It uses chokidar instead of built-in fs.watch because https://github.com/paulmillr/chokidar?tab=readme-ov-file#why

The way we use watcher here is similar to how we used it in HH2. See https://github.com/NomicFoundation/hardhat/blob/b914c6c00f1ef8e47a895bb75a38b766882d792c/packages/hardhat-core/src/builtin-tasks/utils/watch.ts

However, there is one major difference. In HH3, build infos are split between the build info file and the build info output file. We can only start reading them in the watcher once writing to both has finished. To do that, I took inspiration from chokidar's [awaitWriteFinished](https://github.com/paulmillr/chokidar/blob/main/src/index.ts#L751) and implemented `waitUntilBuildFinished` function, which waits for writes to both the build info file and the build info output file to finish.

The entire flow could be depicted as follows:
1. A build info file (or build info output file) is added
2. We receive a notification about it from chokidar
3. We extract the build ID from the the file path we received
4. We check if both the build info file and the build info output file already exist; if not, we abort
5. We check if we haven't started waiting for writes to the build info file and the build info output file to finish before; if we have, we abort
6. We start waiting for writes to the build info file and the build info output file to finish, and mark it so that we don't wait for the writes related to the same build id more than once
7. We read the build info file and the build info output file
8. We update the node with new compilation results
9. We clear the mark acquired in step 6.


v-next/hardhat/src/internal/builtin-plugins/node/task-action.ts
```diff
@@ -129,7 +138,58 @@ const nodeAction: NewTaskActionFunction<NodeActionArguments> = async (
...
-  // TODO(https://github.com/NomicFoundation/hardhat/issues/6040): Add build info watcher here
+  const buildInfoDirPath = path.join(
+    hre.config.paths.artifacts,
+    BUILD_INFO_DIR_NAME,
+  );
+  const addCompilationResult = async (modifiedPath: string) => {
+    log(`Detected change in ${modifiedPath}`);
+
+    // NOTE: We're ignoring the output file here, because the build info files
+    // are modified after the output files
```
galargh, line 149: Actually, we write to them concurrently. So we need to somehow detect when both build info file and its output have been written to.

galargh:
This is replaced by https://github.com/NomicFoundation/hardhat/pull/6928

PR #6529, opened 2025-04-01
fix: pass all artifacts to edr when executing solidity tests

### LINKED ISSUE #6522: Solidity tests task is not passing non-test artifacts to EDR
### Feedback

The `getCode/getDeployedCode` cheatcodes require all the project's artifacts to be available. But Hardhat is only passing test artifacts to EDR, meaning that the code of non-test contracts (which are the ones you are most likely to want) cannot be accessed.

Reproduction:

```solidity
// contracts/Foo.sol
pragma solidity ^0.8.28;

contract Foo {}
```

```solidity
// contracts/Foo.t.sol
pragma solidity ^0.8.28;

import { Test } from "forge-std/Test.sol";
import { Foo } from "./Foo.sol";

contract FooTest is Test {
    function test() public {
        bytes memory barBytecode = vm.getCode("contracts/Foo.sol:Foo");
    }
}
```

This test will fail with "No matching artifact found". If you change the argument to `"contracts/Foo.t.sol:FooTest"` it does work.

galargh:

---

Resolves #6522

In this PR, I ensure the `compile` task is executed before the test sources compilation in the `test solidity` task. I also ensure that all the artifacts present in the artifacts folder after these two compilations are passed over to EDR for the solidity tests execution.

Context: https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1743408250467489

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
@@ -31,10 +31,11 @@ import { run } from "./runner.js";
 interface TestActionArguments {
   testFiles: string[];
   chainType: string;
+  noCompile: boolean;
 }
 
 const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
-  { testFiles, chainType },
+  { testFiles, chainType, noCompile },
   hre,
 ) => {
   let rootFilePaths: string[];
@@ -52,7 +53,9 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
   // NOTE: We run the compile task first to ensure all the artifacts for them are generated
   // Then, we compile just the test sources. We don't do it in one go because the user
   // is likely to use different compilation options for the tests and the sources.
-  await hre.tasks.getTask("compile").run();
+  if (noCompile === false) {
+    await hre.tasks.getTask("compile").run();
+  }
 
   if (testFiles.length > 0) {
     rootFilePaths = testFiles.map((f) =>
@@ -76,15 +79,17 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
   // the tests.solidity paths and the sources paths
   rootFilePaths = Array.from(new Set(rootFilePaths));
 
-  const buildOptions: BuildOptions = {
-    force: false,
-    buildProfile: hre.globalOptions.buildProfile,
-    quiet: true,
-  };
+  if (noCompile === false) {
+    const buildOptions: BuildOptions = {
+      force: false,
+      buildProfile: hre.globalOptions.buildProfile,
+      quiet: true,
+    };
 
-  const results = await hre.solidity.build(rootFilePaths, buildOptions);
+    const results = await hre.solidity.build(rootFilePaths, buildOptions);
 
-  throwIfSolidityBuildFailed(results);
+    throwIfSolidityBuildFailed(results);
+  }
 
   const buildInfos = await getBuildInfos(hre.artifacts);
   const artifacts = await getArtifacts(hre.artifacts);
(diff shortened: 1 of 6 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
```diff
@@ -109,46 +98,75 @@ export async function getBuildInfos(
...
+
+    assertHardhatInvariant(
+      solcVersion !== undefined,
+      "solcVersion should not be undefined",
+    );
+
+    const id = {
+      name: artifact.contractName,
+      solcVersion,
+      source: artifact.sourceName,
```
galargh, line 158: I wonder what value should we use for npm artifacts here?

agostbiro (commented):
Hey thanks for this PR!

I might be wrong, but I think we should change `task-action.ts:L67` to include all `*.sol` files not just `*.t.sol`, so that all the contracts in Solidity sources get built even if they aren't part of the dependency graph of a test file. The reason for this is that users could try to load them with the `getCode/getDeployedCode` cheatcodes.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -49,6 +49,8 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
     return;
   }
 
+  await hre.tasks.getTask("compile").run();
```
galargh, line 52: > I might be wrong, but I think we should change task-action.ts:L67 to include all *.sol files not just *.t.sol, so that all the contracts in Solidity sources get built even if they aren't part of the dependency graph of a test file. The reason for this is that users could try to load them with the getCode/getDeployedCode cheatcodes.

@agostbiro I'll reply here because this line is why the one you mentioned is correct. So what will be happening in the task after the changes introduced in this PR is:
1. We compile all the sources, excluding the test sources. We do that by calling the compile task. We don't compile the sources together with the test sources because we want the user to be able to use different build options for this step.
2. We compile all the test sources using the test build options.
3. We collect everything from artifacts folder to pass it to EDR. These include compilation artifacts from 1. and 2.
4. We find the test suite IDs that we want to pass to EDR. We do that by looking at all the artifacts and checking if:
  a. the source of the artifact matches one of the requested test sources
  b. the artifact contains any tests.
5. We pass 3. and 4. to EDR for the test execution.

agostbiro: Ah thanks for the explanation, sounds good! Could you add the first two points as comments?

agostbiro (approved):

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
galargh: Of course, done :)

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -49,6 +49,11 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
     return;
   }
 
+  // NOTE: We run the compile task first to ensure all the artifacts for them are generated
+  // Then, we compile just the test sources. We don't do it in one go because the user
+  // is likely to use different compilation options for the tests and the sources.
+  await hre.tasks.getTask("compile").run();
```
alcuadrado, line 55: I think we need to reintroduce the `--no-compile` here, otherwise running `hardhat test` will compile twice.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-results.ts
alcuadrado: It should be the `artifact.inputSourceName`, otherwise EDR won't find it.

We should polish the `sourceName` vs `inputSourceName` vs `publicSourceName` situation.

alcuadrado: but for now, something like `artifact.inputSourceName ?? artifact.sourceName`.

alcuadrado (changes requested):
Left two comments, otherwise it looks good to me.

alcuadrado (approved):
🚀

PR #6533, opened 2025-04-03
chore: fix circular dependency handling in internal resolution

0xsenty:

---

fixed an issue where circular dependencies could cause an infinite loop.
now, already processed dependencies are skipped to prevent revisiting them.


alcuadrado:
Can you take a look, @galargh ? If you don't think it's useful feel free to just close it, as this may be a bot.

alcuadrado:
Actually, let's be more strict with our policy here.

alcuadrado:
Hi, 

PR #6535, opened 2025-04-03
Fix Solidity compilation in subprocesses

alcuadrado:
This PR modifies the two solidity `ICompiler` implementations, making them more robust.


alcuadrado:
Inspired (AKA stolen from) https://github.com/NomicFoundation/hardhat-vscode/pull/642

packages/hardhat-core/src/internal/solidity/compiler/index.ts
```diff
@@ -17,32 +18,21 @@ export class Compiler implements ICompiler {
...
+        [scriptPath, this._pathToSolcJs],
+        JSON.stringify(input),
+        {
+          maxBuffer: 1024 * 1024 * 500,
+        }
+      );
+
+      output = stdout;
+    } catch (e: any) {
+      throw new HardhatError(ERRORS.SOLC.CANT_RUN_NATIVE_COMPILER, {}, e);
```
antico5, line 34: Should this be `ERRORS.SOLC.SOLCJS_ERROR`?

alcuadrado: Good catch!

alcuadrado:
> Do we have integration tests for this, at least the happy path ? both for running solcjs and native compiler

Yes, actually, I think we use the compilers too much in integration tests.

packages/hardhat-core/src/internal/solidity/compiler/index.ts
antico5: @alcuadrado I think the error types are swapped between socljs and native implementations :sweat_smile:

antico5 (approved):

PR #6536, opened 2025-04-04
Fix broken link

Hopium21:
Replaced a broken ENS docs link with a working one that points to the ENS smart contracts page:

 Old: https://docs.ens.domains/deploying-ens-on-a-private-chain#migration-file-example

 New: https://docs.ens.domains/contracts

The previous link returned a 404. The new one still gives useful context for the ENS contracts used in this Hardhat example.

I recommend this replacement, but if there's a more specific or updated link, feel free to suggest it — happy to update the PR!


alcuadrado (approved):

alcuadrado:
Not exactly the same, but close enough.

PR #6539, opened 2025-04-04
HH3 plugin migration guide

ItsNickBarry:
This is a draft.  I will edit as I better understand HH3.

Some particular issues:
* It's not absolutely clear how to choose between tasks and hooks.  Most people migrating plugins will probably not bother with hooks if they're not convinced, because tasks still work.
* I don't fully understand the boundaries and interactions between ESM and Typescript.  Is `tsconfig.json` outside of the scope of this guide?  Are there any other required configuration options that I happened to already have set?

### DIFF
--- a/docs/src/content/hardhat3-alpha/learn-more/plugin-migration.md
+++ b/docs/src/content/hardhat3-alpha/learn-more/plugin-migration.md
@@ -159,6 +159,24 @@ export default async (): Promise<Partial<SolidityHooks>> => ({
 });
 ```
 
+Like tasks, hooks must be registeredon your `HardhatPlugin` object:
+
+```typescript
+// index.ts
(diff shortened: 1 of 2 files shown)


docs/src/content/hardhat3-alpha/learn-more/plugin-migration.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Global Options
+
+```typescript
+// index.ts
+import { globalOption } from 'hardhat/config';
+import { ArgumentType } from 'hardhat/types/arguments';
+import type { HardhatPlugin } from 'hardhat/types/plugins';
+
+declare module 'hardhat/types/global-options' {
```
ItsNickBarry, line 217: Is it important to import the module first?  The examples I've seen have done that, but skipping doesn't cause any problems in my plugins.

PR #6541, opened 2025-04-07
Fix compilation error with bignumber.js latest types

fvictorio:
The latest version of bignumber.js has breaking changes in its types, even if it's a minor bump. This change works both with the currently installed version and with the latest. This is just a dev dependency anyway, so it's not a big deal.


kanej (approved):
LGTM.

PR #6544, opened 2025-04-07
Show Hardhat 3 banner based on json config

schaable:
Display a banner when creating a project (only js/ts projects) or running a task. The banner configuration is fetched from a JSON file in the Hardhat repo via an HTTP request.

Example config:

```json
{
  "enabled": false,
  "formattedMessages": [],
  "minSecondsBetweenDisplays": 0,
  "minSecondsBetweenRequests": 0
}
```

The banner supports customizable messages with color and multiline formatting. Multiple messages can be defined, and one is randomly selected for display each time.

### Screenshots
In the above examples, the banner configuration is read from `https://raw.githubusercontent.com/schaable/hardhat/refs/heads/main/banner-config.json` and `console.log`s have been added for testing purposes:

Project creation - TS project
project-creation-ts

Project creation - JS project
project-creation-js

Project creation - Empty project (no banner)
project-creation-empty

Running a task
task-run

Running a task --- No cache file
task-run--no-cache

Running a task --- Banner not shown, time between commands is less than `minSecondsBetweenDisplays`
task-run--no-banner

Running a task --- Request failed, use cache
task-run--banner-request-failed

Running a task --- Banner not shown, request failed and there is no cache
task-run---no-request-no-cache

Running a task --- Request cache expired and a new config is fetched
task-run--full-flow-1
task-run--full-flow-2

Running a task --- timers
task-run---times

### DIFF
--- a/packages/hardhat-core/src/internal/cli/banner-manager.ts
+++ b/packages/hardhat-core/src/internal/cli/banner-manager.ts
@@ -144,7 +144,7 @@ interface BannerCache {
 
 async function readCache(): Promise<BannerCache> {
   const cacheDir = await getCacheDir();
-  const versionNotifierCachePath = path.join(cacheDir, BANNER_CACHE_FILE_NAME);
+  const bannerCacheFilePath = path.join(cacheDir, BANNER_CACHE_FILE_NAME);
 
   let cache: BannerCache = {
     bannerConfig: undefined,
@@ -152,7 +152,7 @@ async function readCache(): Promise<BannerCache> {
     lastRequestTime: 0,
   };
   try {
-    const fileContents = await fs.readFile(versionNotifierCachePath, "utf-8");
+    const fileContents = await fs.readFile(bannerCacheFilePath, "utf-8");
     cache = JSON.parse(fileContents);
   } catch (error) {
     log(
@@ -167,14 +167,11 @@ async function readCache(): Promise<BannerCache> {
 
 async function writeCache(cache: BannerCache) {
   const cacheDir = await getCacheDir();
-  const versionNotifierCachePath = path.join(cacheDir, BANNER_CACHE_FILE_NAME);
+  const bannerCacheFilePath = path.join(cacheDir, BANNER_CACHE_FILE_NAME);
 
   try {
     await fs.mkdir(cacheDir, { recursive: true });
-    await fs.writeFile(
-      versionNotifierCachePath,
-      JSON.stringify(cache, null, 2)
-    );
+    await fs.writeFile(bannerCacheFilePath, JSON.stringify(cache, null, 2));
   } catch (error) {
     log(
       `Error writing cache file:  ${


packages/hardhat-core/src/internal/cli/beta-notifier.ts
```diff
@@ -0,0 +1,146 @@
...
+      if (
+        timeSinceLastRequest <
+        this._bannerConfig.minSecondsBetweenRequests * 1000
+      ) {
+        return [() => {}, Promise.resolve()];
+      }
+    }
+
+    const bannerConfigRequest = request(
+      "https://raw.githubusercontent.com/NomicFoundation/hardhat/refs/heads/main/beta-banner.json",
```
alcuadrado, line 51: nit: I'd make this a contstant

packages/hardhat-core/src/internal/cli/beta-notifier.ts
```diff
@@ -0,0 +1,146 @@
...
+        bannerConfig,
+        lastDisplayTime,
+        lastRequestTime
+      );
+    }
+
+    return this._instance;
+  }
+
+  public async sendBannerConfigRequest(): Promise<[() => void, Promise<void>]> {
```
alcuadrado, line 35: I think we could hide the controller abort inside this function. We may even be able to make this private.

packages/hardhat-core/src/internal/cli/beta-notifier.ts
```diff
@@ -0,0 +1,146 @@
...
+import { getCacheDir } from "../util/global-dir";
+
+interface Hardhat3Banner {
+  enabled: boolean;
+  formattedMessages: string[];
+  minSecondsBetweenDisplays: number;
+  minSecondsBetweenRequests: number;
+}
+
+export class Hardhat3BannerManager {
```
alcuadrado, line 12: I'd add `debug` logs, otherwise this will be impossible to debug.

packages/hardhat-core/src/internal/cli/beta-notifier.ts
alcuadrado: Another nit: I'd remove the word `beta`. We may keep using this after the beta.

packages/hardhat-core/src/internal/cli/beta-notifier.ts
alcuadrado: Like, we could implement a function like this, which showBanner would use

```js
import { request } from "undici";

async function fetchData(url, timeout = undefined) {
  // This should be a `debug` log
  console.log("Trying to fetch data, with timeout", timeout);

  const controller = new AbortController();

  const req = request(url, {
    method: "GET",
    signal: controller.signal,
  });

  const requestAborted = new Error("Request aborted: timeout reached");

  if (timeout !== undefined) {
    setTimeout(() => {
      controller.abort(requestAborted);
    }, timeout);
  }

  try {
    const res = await req;

    const text = await res.body.text();

    return text;
  } catch (err) {
    if (err === requestAborted) {
      // This should be a `debug` log
      console.log("Request aborted", err);
    } else {
      // This should be a `debug` log
      console.error("Unexpected error", err);
    }
  }

  return undefined;
}

console.log(await fetchData("http://example.com", 10));

console.log("\n\n\n\n\n\n");

console.log(await fetchData("http://example.com", 500));

console.log("\n\n\n\n\n\n");

console.log(await fetchData("http://example.com"));
```

alcuadrado (approved):

PR #6546, opened 2025-04-07
build: upgrade to @ignored/edr-optimism@0.10.0-alpha.1

Wodann:
Updates to @ignored/edr-optimism@0.10.0-alpha.1

### DIFF
--- a/v-next/hardhat-zod-utils/src/index.ts
+++ b/v-next/hardhat-zod-utils/src/index.ts
@@ -28,6 +28,7 @@ export interface HardhatUserConfigValidationError {
  * If you have a type that's complex, like an object or array, you must use
  * {@link conditionalUnionType}.
  */
+// TODO: improve the return type of this function to be more specific
 export const unionType = (
   types: Parameters<typeof z.union>[0],
   errorMessage: string,
@@ -79,6 +80,7 @@ export const unionType = (
  * returns true.
  * @returns The conditional union ZodType.
  */
+// TODO: improve the return type of this function to be more specific
 export const conditionalUnionType = (
   cases: Array<[predicate: (data: unknown) => boolean, zodType: ZodType<any>]>,
   noMatchMessage: string,
@@ -164,7 +166,9 @@ export const resolvedConfigurationVariableSchema = z.object({
 /**
  * A Zod type to validate Hardhat's SensitiveString values.
  */
-export const sensitiveStringSchema = unionType(
+export const sensitiveStringSchema: z.ZodType<
+  string | z.infer<typeof configurationVariableSchema>
+> = unionType(
   [z.string(), configurationVariableSchema],
   "Expected a string or a Configuration Variable",
 );
@@ -178,7 +182,9 @@ export const sensitiveStringSchema = unionType(
  * As a workaround, we provide the error message directly in the url() call.
  * We should remove this when the issue is fixed.
  */
-export const sensitiveUrlSchema = unionType(
+export const sensitiveUrlSchema: z.ZodType<
+  string | z.infer<typeof configurationVariableSchema>
+> = unionType(
   [
     z.string().url("Expected a URL or a Configuration Variable"),
     configurationVariableSchema,
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
@@ -22,8 +22,8 @@ import type {
 } from "@ignored/edr-optimism";
 
 import {
-  optimismGenesisState,
-  optimismHardforkFromString,
+  opGenesisState,
+  opHardforkFromString,
   l1GenesisState,
   l1HardforkFromString,
 } from "@ignored/edr-optimism";
@@ -388,9 +388,9 @@ async function getProviderConfig(
     networkConfig.forking !== undefined
       ? [] // TODO: Add support for overriding remote fork state when the local fork is different
       : networkConfig.chainType === "optimism"
-        ? optimismGenesisState(
-            optimismHardforkFromString(
-              // TODO: Optimism conversion is not implemented yet
+        ? opGenesisState(
+            opHardforkFromString(
+              // TODO: OP conversion is not implemented yet
               hardhatHardforkToEdrSpecId(networkConfig.hardfork),
             ),
           )
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
@@ -11,6 +11,7 @@ import type {
   RequestArguments,
   SuccessfulJsonRpcResponse,
 } from "../../../../types/providers.js";
+import type { RequireField } from "../../../../types/utils.js";
 import type { DefaultHDAccountsConfigParams } from "../accounts/constants.js";
 import type { JsonRpcRequestWrapperFunction } from "../network-manager.js";
 import type {
@@ -125,7 +126,7 @@ export const EDR_NETWORK_DEFAULT_PRIVATE_KEYS: string[] = [
 ];
 
 interface EdrProviderConfig {
-  networkConfig: EdrNetworkConfig;
+  networkConfig: RequireField<EdrNetworkConfig, "chainType">;
   loggerConfig?: LoggerConfig;
   tracingConfig?: TracingConfigWithBuffers;
   jsonRpcRequestWrapper?: JsonRpcRequestWrapperFunction;
@@ -152,12 +153,6 @@ export class EdrProvider extends BaseProvider {
     const providerConfig = await getProviderConfig(networkConfig);
 
     let edrProvider: EdrProvider;
-
-    // The chainType can't be undefined, as it's always set on connect()
-    assertHardhatInvariant(
-      networkConfig.chainType !== undefined,
-      "The chain type is undefined",
-    );
 
     // We need to catch errors here, as the provider creation can panic unexpectedly,
     // and we want to make sure such a crash is propagated as a ProviderError.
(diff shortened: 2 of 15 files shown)


v-next/hardhat-zod-utils/src/index.ts
```diff
@@ -164,7 +166,9 @@ export const resolvedConfigurationVariableSchema = z.object({
 /**
  * A Zod type to validate Hardhat's SensitiveString values.
  */
-export const sensitiveStringSchema = unionType(
+export const sensitiveStringSchema: z.ZodType<
+  string | z.infer<typeof configurationVariableSchema>
+> = unionType(
```
schaable, line 171: This might seem unrelated, but it actually helped me improve the typings in `refineEdrNetworkUserConfig`, as some fields were being inferred as `any`.

A similar improvement can be made for `unionType` and `conditionalUnionType`, but those cases are a bit trickier, so I added TODOs to handle them in a separate PR.

Do you think we should create issues for those as well?

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
```diff
@@ -102,49 +152,6 @@ export function hardhatHardforkToEdrSpecId(hardfork: string): string {
   }
 }
 
-export function edrSpecIdToHardhatHardfork(specId: string): HardforkName {
```
schaable, line 105: We don’t seem to use this anywhere. Maybe it was left behind during a refactor that removed the calls?

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
```diff
@@ -274,28 +281,32 @@ export async function normalizeEdrNetworkAccountsConfig(
...
-    });
-  }
-
-  return edrChains;
+  return (
+    Array.from(chains)
+      // Skip chains that don't match the expected chain type
+      .filter(([_, config]) => {
+        if (chainType === GENERIC_CHAIN_TYPE) {
+          // When "generic" is requested, include both "generic" and "l1" chains
```
schaable, line 291: Can you confirm this? If the passed `chainType` (i.e., the "resolved" chain type) is `"generic"`, we want to return both `"generic"` and `"l1"` hardforks from the `chains` map.

v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
```diff
@@ -221,54 +220,102 @@ const baseNetworkUserConfigSchema = z.discriminatedUnion("type", [
...
       enableTransientStorage,
     } = networkConfig;
 
-    if (hardforkGte(hardfork, HardforkName.LONDON)) {
-      if (minGasPrice !== undefined) {
+    if (hardfork !== undefined && !isValidHardforkName(hardfork, chainType)) {
+      ctx.addIssue({
+        code: z.ZodIssueCode.custom,
+        path: ["hardfork"],
+        message: `Invalid hardfork name ${hardfork} for chainType ${chainType}. Expected ${getHardforks(
```
schaable, line 240: I think we should decide on a consistent way to wrap values (single quotes, double quotes, etc.) and normalize all validation and error messages. I'm pretty sure we already have this tracked somewhere.

v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
```diff
@@ -221,54 +220,102 @@ const baseNetworkUserConfigSchema = z.discriminatedUnion("type", [
...
+            });
+          }
+          hardforkHistoryIdx++;
+        }
+        chainIdx++;
+      }
+    }
+
+    const resolvedHardfork = hardfork ?? getLatestHardfork(chainType);
+    if (chainType === L1_CHAIN_TYPE || chainType === GENERIC_CHAIN_TYPE) {
```
schaable, line 280: I’ve added an extra validation here, since this should now only be validated for `"l1"` or `"generic"` chain types.

v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
```diff
@@ -202,7 +201,7 @@ const edrNetworkUserConfigSchema = z.object({
   enableRip7212: z.optional(z.boolean()),
   enableTransientStorage: z.optional(z.boolean()),
   forking: z.optional(edrNetworkForkingUserConfigSchema),
-  hardfork: z.optional(hardforkNameSchema),
+  hardfork: z.optional(z.string()),
```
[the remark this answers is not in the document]
schaable: Not correct, as `hardfork` also depend on the `chainType`. This is validated in `refineEdrNetworkUserConfig`

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/types/hardfork.ts
```diff
@@ -18,21 +22,84 @@ export enum HardforkName {
...
+const L1_LATEST_HARDFORK: L1HardforkName =
+  L1_HARDFORK_ORDER[L1_HARDFORK_ORDER.length - 1];
+const OP_LATEST_HARDFORK: OpHardforkName =
+  OP_HARDFORK_ORDER[OP_HARDFORK_ORDER.length - 1];
+
+export function getLatestHardfork(chainType: ChainType): string {
+  return chainType === OPTIMISM_CHAIN_TYPE
+    ? OP_LATEST_HARDFORK
+    : L1_LATEST_HARDFORK;
+}
```
Wodann, line 53: Important to note for EDR:

What EDR refers to as `opLatestHardfork()` or `l1LatestHardfork` refers to the latest activated hardfork. It could however be that internally there are supported pre-release hardfork. E.g. when clients are starting to develop Osaka, EDR might already have that hardfork as an option with a preliminary implementation.

This would adversely affect this logic if Hardhat adds Osaka  to the `HardforkName` enum.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
Wodann: Seems safe to remove

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
Wodann: Generic is a more permissive version of L1. Thus, all L1 one chains can be treated as generic chains.

Generic chains cannot be treated as L1 chains.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -153,6 +153,12 @@ export class EdrProvider extends BaseProvider {
 
     let edrProvider: EdrProvider;
 
+    // The chainType can't be undefined, as it's always set on connect()
+    assertHardhatInvariant(
```
Wodann, line 157: Could this not be handled at the type level by "resolving" a `networkConfig` with `chainType?: string` to a `ResolvedNetworkConfig` type with `chainType: string`?

v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
Wodann: Looks correct to me

v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
```diff
@@ -221,54 +220,102 @@ const baseNetworkUserConfigSchema = z.discriminatedUnion("type", [
...
+
+      if (
+        !hardforkGte(resolvedHardfork, L1HardforkName.CANCUN, chainType) &&
+        enableTransientStorage === true
+      ) {
         ctx.addIssue({
           code: z.ZodIssueCode.custom,
-          message:
-            "minGasPrice is not valid for networks with EIP-1559. Try an older hardfork or remove it.",
+          message: `'enableTransientStorage' is not supported for hardforks before 'cancun'. Please use a hardfork from 'cancun' onwards to enable this feature.`,
```
Wodann, line 305: Is the `enableTransientStorage` config option not obsolete? I thought that was merely added for people to test transient storage before support was added in Cancun

schaable: Yes, but removing it should not be part of this PR.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
schaable: The `EdrNetworkConfig` is the resolved version of the config, and `chainType` is optional there. However, the `EdrProvider` is only used within the network manager, and if you look at where `.create` is called, we always pass a `chainType`.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/types/hardfork.ts
schaable: I agree this needs another iteration. Maybe we could assign a "state" to each hardfork and pick the latest one that isn’t a pre-release. That said, this might be good enough for the alpha.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
schaable: So it should be the other way around? If "l1" is passed, we include both "generic" and "l1"? That would mean "generic" chains won’t pass any chains to EDR,  is that correct?

Wodann: No, I think it's correct as is.

E.g. A user can define a generic chain type for Arbitrum. In that case it should not be sent to an L1 chain.

alcuadrado (approved):

PR #6547, opened 2025-04-08
Fix compilation error with bignumber.js latest types

fvictorio:
Same as https://github.com/NomicFoundation/hardhat/pull/6541 but after fixing that I only re-compiled in `hardhat-core` and didn't realize this was also happening in another package.


alcuadrado (approved):
Confirming that this does not need a changeset.

PR #6552, opened 2025-04-09
Refactor error categories to support subcategories

zoeyTM:
Closes #6368 
Closes #6314

PR #6562, opened 2025-04-10
Move pluginId to the error category from the error descriptors

alcuadrado:
This is a small improvement on top of #6552.

I moved the `pluginId` definition from each error descriptor (where it's easy to forget it), to the error categories object.

I didn't add any changeset because it can be seen as covered by #6552's changeset.


alcuadrado:
This leads to the errors being displayed like this:

image

zoeyTM (approved):
LGTM

PR #6566, opened 2025-04-11
chore: fix up changesets for next release

kanej:
This is a small tweak to our changesets and changeset config in prep for the next release.

The changes:

- update the text of some of the changesets to work better in the release notes
- add some missing links to issues for more context
- fix some changeset patches to allow for a clean `pnpm changeset version`
- update the version of `@nomicfoundation/hardhat-toolbox-viem` to `4.0.0-next.3`, as there is already a `v3.0.0` for the toolbox in HH2. This matches the pattern from `hardhat-ethers`


.changeset/pre.json
```diff
@@ -20,7 +20,7 @@
     "@nomicfoundation/hardhat-toolbox-viem": "3.0.0",
     "@nomicfoundation/hardhat-typechain": "2.0.0",
     "@nomicfoundation/hardhat-utils": "2.0.0",
-    "@nomicfoundation/hardhat-toolbox-mocha-ethers": "3.0.0",
+    "@nomicfoundation/hardhat-toolbox-mocha-ethers": "2.0.0",
```
ChristopherDedominici, line 23: You changed the `mocha toolbox` from `3.0.0` to `2.0.0`. What about `viem toolbox` at line 20, is it correct for it to be at `3.0.0`?

.changeset/polite-clocks-obey.md
```diff
@@ -1,7 +1,6 @@
 ---
```
ChristopherDedominici, line 1: I don't see a changeset for the `viem` toolbox, is it expected because the current changeset does not need any changes?

kanej: Yes. The current changeset for viem looks good.

.changeset/pre.json
kanej: The versions in `pre` are used on the initial setup of pre-release config. We are already beyond that, so whichever version we put in here does not have a bearing.
I have set `hardhat-toolbox-mocha-ethers` to 2.0.0 inline with the other new packages here.
I have set `hardhat-toolbox-viem` to 3.0.0 which is the current latest version on npm.
When running `pnpm changeset version` both will be updated:
* `hardhat-toolbox-mocha-ethers@3.0.0-next.4`
* `hardhat-toolbox-viem@4.0.0-next.4`

alcuadrado (approved):

Issue #6555, opened 2025-04-10
Update opt-in flow based on new privacy policy


kanej:
The Hardhat 3 Alpha should default to opt-in when asking for permission to send anonymous analytics to GA/Sentry.
We should double check the updated flow.
We should also check that we are anonymizing the Sentry Server name correctly

PR #6573, opened 2025-04-14
feat: enable telemetry by default

galargh:

---

Resolves https://github.com/NomicFoundation/hardhat/issues/6555

In this PR I:
- remove the telemetry prompt from the init flow and main CLI startup
- enable telemetry by default
- add a `telemetry` task through which one can enable or disable telemetry explicitly
- set the server name to a generic `<user-server>` string on the Sentry client to anonymize server name tags

The merge of this PR should be coordinated with Manu.


.changeset/thin-peas-end.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Removed timeout from the telemetry consent prompt and anonymized server name tag on events sent to Sentry
```
galargh, line 5:
```suggestion
Removed telemetry consent prompt, made telemetry enabled by default, added a task through which telemetry can be explicitly enabled/disabled, and anonymized server name tag on events sent to Sentry
```

v-next/hardhat/src/internal/cli/telemetry/analytics/analytics.ts
```diff
@@ -24,25 +24,25 @@ const SESSION_ID = Math.random().toString();
 const ENGAGEMENT_TIME_MSEC = "10000";
 
 // Return a boolean for testing purposes to verify that analytics are not sent in CI environments
-export async function sendTelemetryConsentAnalytics(
-  consent: boolean,
+export async function sendTelemetryConfigAnalytics(
+  enabled: boolean,
 ): Promise<boolean> {
```
galargh, line 29: FYI, I also renamed helper functions that dealt with telemetry permissions, like this one for example, so that they refer to whether telemetry is enabled or disabled rather than whether telemetry consent is granted or revoked. I thought this was more in line with the new task you proposed and its flags.

If you prefer to stay with the previous nomenclature outside the task, let me know. I can restore it in no time.

alcuadrado: makes a ton of sense

v-next/hardhat/src/internal/cli/telemetry/telemetry-permissions.ts
```diff
@@ -95,58 +60,44 @@ export function isTelemetryAllowedInEnvironment(): boolean {
...
 
-  return undefined;
+  return true;
 }
 
-async function getTelemetryConsentFilePath() {
+async function getTelemetryConfigFilePath() {
   const configDir = await getTelemetryDir();
-  return path.join(configDir, "telemetry-consent.json");
+  return path.join(configDir, "telemetry-config.json");
```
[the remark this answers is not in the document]
alcuadrado: +1

This also makes sense, because it invalidates all the previous ones.

alcuadrado (approved):

alcuadrado:
Approved, but don't merge yet, as that means that this may get released accidentally.

Issue #6559, opened 2025-04-10
Minimal port of assertions to viem


kanej:
Taking inspiration from the `hardhat-chai-matchers` plugin create an assertion Hardhat plugin for use in the `viem` stack.

Design doc: https://www.notion.so/nomicfoundation/Hardhat-Viem-Matchers-1ce578cdeaf5805c8f58d2944a51591a#1ce578cdeaf5806dbe21e3fbd2896b92

PR #6574, opened 2025-04-14
Viem matchers first iteration

ChristopherDedominici:
[Parent issue](https://github.com/NomicFoundation/hardhat/issues/6559)


v-next/hardhat-viem-matchers/test/fixture-projects/hardhat-project/contracts/Events.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: MIT
```
ChristopherDedominici, line 1: These methods were copied from `chai-matchers`. Most are currently unused but will be utilized once more extensive testing is added.

v-next/hardhat-viem-matchers/test/internal/matchers/balances-have-changed.ts
```diff
@@ -0,0 +1,91 @@
...
+        to: aliceWalletClient.account.address,
+        value: 1000000000000000000000n,
+      });
+
+      const publicClient = await viem.getPublicClient();
+      await publicClient.waitForTransactionReceipt({ hash });
+    }, [
+      {
+        address: bobWalletClient.account.address,
+        amount: -1000000023255859375000n, // TODO: gas fees that change the value?
```
ChristopherDedominici, line 58: I noticed that the balance variation in the sender is bigger than the amount sent. I assume it's because of gas fee?

v-next/hardhat-viem-matchers/src/internal/matchers/revert/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+      hasReverted = true;
+
+      assertHardhatInvariant(
+        typeof error.details === "string",
+        "The error details should be a string",
+      );
+
+      reason = error.details
+        .split("reverted with reason string ")[1]
+        .replaceAll("'", "");
```
fvictorio, line 27: I don't think this is very future-proof. Ideally we want an API-stable way of getting the return data, and then we decode it ourselves. I think this will be tricky, for fun JSON-RPC reasons, but it's important.

v-next/hardhat-viem-matchers/src/internal/matchers/revert/core.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    const { args } = decodeErrorResult({ data });
+
+    return args.map(String).join(", ");
+  }
+
+  assert.fail("The function was expected to revert, but it did not.");
+}
+
+function extractRevertData(
```
ChristopherDedominici, line 26: // Function to extract the revert reason from a Viem error

v-next/hardhat-viem-matchers/src/internal/matchers/emit/emit-with-args.ts
```diff
@@ -0,0 +1,84 @@
...
+  assert.notEqual(
+    abiEvents.length,
+    0,
+    `Event "${eventName}" not found in the contract ABI`,
+  );
+
+  assert.equal(
+    abiEvents.length,
+    1,
+    `There should be only one event named "${eventName}" in the contract ABI`,
```
ChristopherDedominici, line 51: Is it ok to have more than one? If yes, should I take the last one?

v-next/hardhat-viem-matchers/src/internal/hook-handlers/network.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export default async (): Promise<Partial<NetworkHooks>> => {
+  const handlers: Partial<NetworkHooks> = {
+    async newConnection<ChainTypeT extends ChainType | string>(
+      context: HookContext,
+      next: (
+        nextContext: HookContext,
+      ) => Promise<NetworkConnection<ChainTypeT>>,
+    ) {
+      const connection: NetworkConnection<ChainTypeT> = await next(context);
```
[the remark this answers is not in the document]
kanej: Not a new convention. I pushed a cleanup commit removing the explicit type.

v-next/hardhat-viem-matchers/src/internal/matchers/emit/emit-with-args.ts
kanej: You can have event overloading like method overloading (so same name but different parameters).
How is this handled in chai matchers? Does it require you to pass an unambigous name in the case of overloaded events e.g. "MyEvent(address,address,bytes32)"

fvictorio:
I haven't reviewed the code in depth, but the API and set of assertions LGTM for a first version.

fvictorio (approved):

v-next/hardhat-viem-matchers/src/types.ts
```diff
@@ -0,0 +1,78 @@
...
+import type {
+  Address,
+  ContractEventName,
+  Hash,
+  ReadContractReturnType,
+  WriteContractReturnType,
+} from "viem";
+
+export interface HardhatViemMatchers {
+  utils: HardhatViemMatchersUtils;
```
kanej, line 14: Thinking it over, I think I agree @ChristopherDedominici.
Lets remove the `utils` object (and the associated code and tests). They feel like useful utils but not really `viem` related.

ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6574/commits/9f77b09ecc103f68ca645023a2055cb5dabd67b2

v-next/hardhat-viem-matchers/src/internal/matchers/balances-have-changed.ts
```diff
@@ -0,0 +1,72 @@
+import type { HardhatViemHelpers } from "@nomicfoundation/hardhat-viem/types";
+import type { ChainType } from "hardhat/types/network";
+import type { Address, Hash } from "viem";
+
+import assert from "node:assert/strict";
+
+import { isHash } from "@nomicfoundation/hardhat-utils/eth";
+
+export async function balancesHaveChanged<
```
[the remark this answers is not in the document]
ChristopherDedominici: I do not have a strong opinion, I omitted `ether` because I thought it was implied but I can re-added if we think it's better

v-next/hardhat-viem-matchers/src/internal/matchers/revert/core-revert.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import assert from "node:assert/strict";
+
+import { ensureError } from "@nomicfoundation/hardhat-utils/error";
+import { decodeErrorResult } from "viem";
+
+import { extractRevertData } from "./extract-revert-data.js";
+import { isDefaultRevert } from "./is-default-revert.js";
+
+export async function handleRevert(
```
kanej, line 11: Can we rename the file to `handle-revert.ts`, as there is only one function in here.

v-next/hardhat-viem-matchers/src/internal/matchers/revert/core-revert-with-custom-error.ts
```diff
@@ -0,0 +1,82 @@
...
+import { ensureError } from "@nomicfoundation/hardhat-utils/error";
+import { decodeErrorResult } from "viem";
+
+import { extractRevertData } from "./extract-revert-data.js";
+import {
+  DEFAULT_REVERT_REASON_SELECTOR,
+  isDefaultRevert,
+} from "./is-default-revert.js";
+
+export async function handleRevertWithCustomError<
```
kanej, line 19: Can we rename the file to `handle-revert-with-custom-error.ts`, as there is only one function in here.

v-next/hardhat-viem-matchers/src/internal/matchers/revert/core-revert-with-custom-error.ts
```diff
@@ -0,0 +1,82 @@
...
+        `Expected error, but the type is "${abiItem.type}".`,
+      );
+
+      assert.equal(
+        abiItem.name,
+        customErrorName,
+        `Expected error name: "${customErrorName}", but found "${abiItem.name}".`,
+      );
+
+      return Array.isArray(args) ? [...args] : [];
```
kanej, line 65: If `args` is not array does that imply there was no args?

v-next/hardhat-viem-matchers/test/internal/matchers/emit/emit-with-args.ts
```diff
@@ -0,0 +1,171 @@
...
+            error,
+            `Event "SameEventDifferentArgs" with argument count 4 not found in the contract ABI`,
+            false,
+            true,
+          ),
+      );
+    });
+  });
+
+  it("should not throw if no args are passed and nothing is emitted", async () => {
```
kanej, line 121: Should this not throw a validation error internally, as in calling `emitWithArgs` but expecting no args is unexpected?

ChristopherDedominici: It depends on how we want to handle it. If we choose to be more restrictive, we should require the array to contain at least one argument with the expected parameters.
This scenario was discussed with Franco, and we agreed to take a more relaxed approach and allow it

v-next/hardhat-viem-matchers/src/internal/matchers/revert/core-revert.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6574/commits/caced8309cdc5b39462f7669f334c6c4c3cedca0

v-next/hardhat-viem-matchers/src/internal/matchers/revert/core-revert-with-custom-error.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6574/commits/caced8309cdc5b39462f7669f334c6c4c3cedca0

v-next/hardhat-viem-matchers/src/internal/matchers/emit/emit-with-args.ts
```diff
@@ -0,0 +1,84 @@
...
+    assert.ok(
+      parsedLog !== undefined,
+      `No arguments in the event logs, are you sure you are targeting an event with arguments?`,
+    );
+
+    const abiEvent = abiEvents[0];
+
+    let parsedLogCount = 0;
+    for (const [index, param] of abiEvent.inputs.entries()) {
+      assertHardhatInvariant(
```
kanej, line 62: Can you talk me through this? Are we saying you can't assert unnamed args for an event?

ChristopherDedominici: Good comment! It made me to revisit the logic and spot a bug, which I've addressed here: https://github.com/NomicFoundation/hardhat/pull/6574/commits/733ee7230c996bd772565bd3cfa15c4b6c89778f

There is also a comment in the code explaining the logic

kanej (approved):
Great work. Lets get this merged!

PR #6576, opened 2025-04-15
add ensureNodeErrnoExceptionError function to assert the error

### LINKED ISSUE #5605: Remove unsafe casting with ensureError<T>
Using `ensureError` with a type parameter casts the error but doesn't perform any runtime validations. For example:

```typescript
try {
  throw new Error();
} catch (e) {
  ensureError<NodeJS.ErrnoException>(e);
  if (e.code === "ENOENT") {
    // ...
  }
}
```

In this case, we can't guarantee that `e.code` exists at runtime. In fact, it won't exist because we're throwing a regular `Error`. To address this, we should remove the type casting and instead use a check like `"code" in e` to ensure the property exists on the error object. We could wrap this logic within a `ensureNodeJSError` helper function to avoid repetition.

iosh:

---

resolve #5605 

Add ensureNodeJSError function to replace unsafe  ensureError<T>

### DIFF
--- a/v-next/hardhat-utils/src/error.ts
+++ b/v-next/hardhat-utils/src/error.ts
@@ -79,6 +79,31 @@ export function ensureError<ErrorT extends Error>(
 }
 
 /**
+ * Ensures that the provided value is a NodeJS.ErrnoException with a string 'code'.
+ * @param thrown The value to check.
+ * @throws The value if its not an error or if it doesn't have a code property.
+ * @example
+ * ```ts
+ * try {
+ *   await fs.promises.readFile("non-existing-file.txt");
+ * } catch (error) {
+ *   ensureNodeErrnoExceptionError(error);
+ *   console.error(error.code);
+ * }
+ * ```
+ */
+
+export function ensureNodeErrnoExceptionError(
+  thrown: unknown,
+): asserts thrown is NodeJS.ErrnoException & Error & { code: string } {
+  ensureError(thrown);
+
+  if (!("code" in thrown) || typeof thrown.code !== "string") {
+    throw thrown;
+  }
+}
+
+/**
  * Throws an error for an unreachable code path. This function is typically
  * used in a default case of a switch statement where all possible values of
  * the switched variable should be handled in other cases. If the default case
--- a/v-next/hardhat-utils/src/error.ts
+++ b/v-next/hardhat-utils/src/error.ts
@@ -92,7 +92,6 @@ export function ensureError<ErrorT extends Error>(
  * }
  * ```
  */
-
 export function ensureNodeErrnoExceptionError(
   thrown: unknown,
 ): asserts thrown is NodeJS.ErrnoException & Error & { code: string } {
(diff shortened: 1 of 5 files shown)


v-next/hardhat-utils/src/error.ts
```diff
@@ -78,6 +78,31 @@ export function ensureError<ErrorT extends Error>(
...
+ * ```ts
+ * try {
+ *   await fs.promises.readFile("non-existing-file.txt");
+ * } catch (error) {
+ *   ensureNodeErrnoExceptionError(error);
+ *   console.error(error.code);
+ * }
+ * ```
+ */
+
```
alcuadrado, line 95:
```suggestion

```

alcuadrado (approved):

v-next/hardhat-utils/src/error.ts
```diff
@@ -78,6 +78,30 @@ export function ensureError<ErrorT extends Error>(
...
+ * ```
+ */
+export function ensureNodeErrnoExceptionError(
+  thrown: unknown,
+): asserts thrown is NodeJS.ErrnoException & Error & { code: string } {
+  ensureError(thrown);
+
+  if (!("code" in thrown) || typeof thrown.code !== "string") {
+    throw thrown;
+  }
```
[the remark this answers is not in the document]
alcuadrado: Ignore this

v-next/hardhat-utils/src/fs.ts
```diff
@@ -241,19 +241,22 @@ export async function readJsonFileAsStream<T>(
 
     return result;
   } catch (e) {
-    ensureError<NodeJS.ErrnoException>(e);
+    ensureError(e);
```
schaable, line 244: Can't you use `ensureNodeErrnoExceptionError` to avoid the `if ("code" in e)` check?

v-next/hardhat-utils/src/fs.ts
```diff
@@ -318,7 +321,7 @@ export async function writeJsonFileAsStream<T>(
 
     await pipeline(jsonStream, fileWriteStream);
   } catch (e) {
-    ensureError<NodeJS.ErrnoException>(e);
+    ensureError(e);
```
schaable, line 324: same here

schaable (commented):
Left a small comment but looks great!

v-next/hardhat-utils/src/fs.ts
iosh: The `e` here could be either thrown by `new Error` or by `fs` , so I don't think we can use `ensureNodeErrnoExceptionError` there instead.

v-next/hardhat-utils/src/fs.ts
iosh: This is also a mix of two types of errors.

v-next/hardhat-utils/src/fs.ts
schaable: And you’re absolutely right :) I didn’t notice the `new Error` in the function code.

schaable (approved):

Issue #6580, opened 2025-04-16
Instrument the code on compilation job creation when coverage is enabled


galargh:
... and save metadata to `artifacts/build-info/${buildId}.coverage.json`

See https://www.notion.so/nomicfoundation/Code-Coverage-in-Hardhat-3-1cf578cdeaf580a4b640f394cb70da40?pvs=4#1d6578cdeaf5800ca46ac9bf244ff369

PR #6589, opened 2025-04-16
feat: instrument code before compilation and save metadata next to build info

galargh:

---

Resolves https://github.com/NomicFoundation/hardhat/issues/6580

This PR adds instrumentation to the compilation job when coverage is enabled and saving metadata produced by the instrumentation in the build info folder.

Please note that the actual call to EDR to instrument the code and get the metadata will be added in a subsequent PR once that feature is available. This PR only sets up the skeleton for the process.


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -71,23 +88,41 @@ export class CompilationJobImplementation implements CompilationJob {
...
+    let coverageMetadata: CoverageMetadata = {};
 
     const resolvedFiles = this.#getResolvedFiles();
 
     for (const file of resolvedFiles) {
-      sources[file.sourceName] = {
-        content: file.content.text,
-      };
+      if (this.#coverage && file.type === ResolvedFileType.PROJECT_FILE) {
+        // TODO: Call EDR to instrument the content of the file and obtain the metadata
```
galargh, line 101: Please note that this TODO will be filled in once the instrumentation feature is available in EDR. This will be added in a future PR.

v-next/hardhat/src/types/solidity/solidity-artifacts.ts
```diff
@@ -60,3 +61,18 @@ export interface SolidityBuildInfoOutput {
    */
   readonly output: CompilerOutput;
 }
+
```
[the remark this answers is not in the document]
alcuadrado: Actually, maybe the right file is `v-next/hardhat/src/types/coverage.ts`.

v-next/hardhat/src/types/solidity/build-system.ts
```diff
@@ -45,6 +45,11 @@ export interface BuildOptions {
    * If `true`, the build process doesn't print any output.
    */
   quiet?: boolean;
+
+  /**
+   * If `true`, the root files will be instrumented before compilation.
```
alcuadrado, line 50: I think this comment is out of sync with respect to the implementation. In particular, which files get instrumented.

v-next/hardhat/src/internal/builtin-plugins/solidity/index.ts
```diff
@@ -13,6 +13,13 @@ const hardhatPlugin: HardhatPlugin = {
       );
       return artifactsPlugin;
     },
+    async () => {
+      // NOTE: builtin:solidity depends on buitin:coverage because it needs to
```
alcuadrado, line 17: This may lead to a circular dependency? I'd say we proceed with this as is, and then we'll see.

alcuadrado (approved):
I think we should make the coverage metadata internal, both the types and the actual files. We are sure that these will be pretty unstable, so we can't expose them as part of the public API.

Apart from that, looks good.

v-next/hardhat/src/types/solidity/build-system.ts
galargh:
```suggestion
   * If `true`, the project files will be instrumented before compilation.
```

v-next/hardhat/src/types/solidity/solidity-artifacts.ts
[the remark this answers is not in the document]
galargh: I think we'll still need to advertise some types publicly for external plugins, like node/mocha test runners, to consume. They're going to need, at least, information about the structure of coverage reports.

I do agree, they don't have to know about metadata. So I think it would make sense to keep those under the coverage plugin.

galargh:
I'm closing this PR for now. I will reuse its' content to implement https://www.notion.so/nomicfoundation/Instrumentation-metadata-and-node-test-coverage-data-collection-1df578cdeaf580eb8847dd730a496df5 once it's finalised.

PR #6595, opened 2025-04-17
feat: collect coverage data from EDR providers created inside node:test runners

galargh:

---

Resolves https://github.com/NomicFoundation/hardhat/issues/6581

This PR adds coverage data collection to the node:test runner.

The flow of data collection from node:test workers is as follows:
1. Coverage plugin adds coverage manager to the hre context on creation. It also registers network hooks (to be described later in step 4).
2. Node test runner clears the state of the coverage manager before any tests are run. This ensures that coverage hits collected during the run are only for that particular run.
3. Node test runner starts node test workers with a tiny helper imported. It adds a piece of code to be executed after the tests inside the worker are complete (to be described later in step 5).
4. Node test workers execute the tests. When a new EDR connection is created, the underlying provider is saved in the coverage manager (this is handled via a `newConnection` hook registered in step 1). When a connection is closed, the coverage hits are collected from the underlying EDR provider and stored in the coverage manager (this is handled via a `closeConnection` hook registered in step 1).
5. Once all the tests are complete, the coverage manager is instructed to dump all the collected coverage hits to disk to the `./cache/coverage/hits/${RANDOM_UUID}.json` file (this is handled by the helper added in step 3). In that steps, any unclosed connections are also handled properly.
6. Node test runner loads all the hits dumped to disk since step 2 and processes them further. This is not implemented yet.


v-next/hardhat-node-test-coverage/package.json
```diff
@@ -0,0 +1,67 @@
...
+    "ethereum",
+    "smart-contracts",
+    "hardhat"
+  ],
+  "scripts": {
+    "lint": "pnpm prettier --check && pnpm eslint",
+    "lint:fix": "pnpm prettier --write && pnpm eslint --fix",
+    "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
+    "prettier": "prettier \"**/*.{ts,js,md,json}\"",
+    "test": "node --import tsx/esm --test \"test/**/*.ts\"",
```
galargh, line 27: We do not use our test reporter here because it would create a dependency cycle.

v-next/hardhat-node-test-coverage/src/collector.ts
```diff
@@ -0,0 +1,52 @@
...
+    const currentReport: CoverageReport = await readJsonFile(
+      process.env.HARDHAT_NODE_TEST_COVERAGE_PATH,
+    );
+    const updatedReport: CoverageReport = {
+      markerIds: [...currentReport.markerIds, ...report.markerIds],
+    };
+    await writeJsonFile(
+      process.env.HARDHAT_NODE_TEST_COVERAGE_PATH,
+      updatedReport,
+    );
```
galargh, line 50: We use a file to communicate coverage report back to the test node task because `TestContext.diagnostic()` method is not available in the context of the `after()` block (which is of type `TestContext | SuiteContext`).

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -103,6 +131,17 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
...
 
+    if (hre.globalOptions.coverage === true) {
+      assertHardhatInvariant(
+        process.env.HARDHAT_NODE_TEST_COVERAGE_PATH !== undefined,
+        "HARDHAT_NODE_TEST_COVERAGE_PATH should be defined",
+      );
+      const report: CoverageReport = await readJsonFile(
+        process.env.HARDHAT_NODE_TEST_COVERAGE_PATH,
+      );
+      log("Coverage report", report);
```
galargh, line 142: Please note the display of the report as well as bubbling of the data up the stack will be handled separately.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -71,12 +71,17 @@ export class NetworkManagerImplementation implements NetworkManager {
...
       "network",
       "newConnection",
-      [],
-      async (_context) =>
+      [networkName, chainType, networkConfigOverride],
+      async (
+        _nextContext,
+        nextNetworkName,
+        nextChainType,
+        nextNetworkConfigOverride,
```
galargh, line 79: I modify the `newConnection` hook arguments because we need to be able to modify the network config override in a hook before the network is created.

I also expose network name and chain type in the same manner, even though I don't need it for coverage support specifically, but I think it could be useful for some to know these values ahead of network initialization, too.

.changeset/tiny-walls-tan.md
```diff
@@ -0,0 +1,8 @@
+---
+"@nomicfoundation/hardhat-ethers-chai-matchers": patch
```
alcuadrado, line 2: I think this is missing a few packages that got their network hook handlers updated.

v-next/hardhat-node-test-coverage/src/collector.ts
```diff
@@ -0,0 +1,44 @@
...
+before(() => {
+  const networkHooks: Partial<NetworkHooks> = {
+    newConnection: async (
+      context,
+      networkName,
+      chainType,
+      networkConfigOverride,
+      next,
+    ) => {
+      // TODO: Define networkConfigOverride.observability.codeCoverage.onCollectedCoverageCallback
```
alcuadrado, line 25: Note that these shouldn't be exposed as part of the config. The Hardhat config of EDR networks shouldn't match the EDR internal config format. They will evolve independently.

v-next/hardhat-node-test-coverage/src/collector.ts
```diff
@@ -0,0 +1,44 @@
...
+      // if this is an EDR network config
+      return next(context, networkName, chainType, networkConfigOverride);
+    },
+  };
+
+  hre.hooks.registerHandlers("network", networkHooks);
```
alcuadrado, line 31: I thought this was going to be done in the `coverage` builtin plugin, with a non-dynamic handler, that would be a `nop` if `globalOptions.coverage` is false.

v-next/hardhat-node-test-coverage/src/collector.ts
alcuadrado: Having written that, I think this approach won't work. Because this handler won't be able to access that internal config field.

v-next/hardhat-node-test-coverage/src/collector.ts
```diff
@@ -0,0 +1,44 @@
+import type { CoverageReport } from "hardhat/types/coverage";
```
alcuadrado, line 1: I was expecting this file to live in the coverage plugin, so that we don't create a new package/dependency.

alcuadrado: Actually, we probably want it in the `node-test-runner` plugin instead.

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -67,7 +79,21 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
...
-  process.env.NODE_OPTIONS = `--import "${tsx.href}"`;
+  const imports = [tsx.href];
+
+  if (hre.globalOptions.coverage === true) {
+    const coverage = new URL(
+      import.meta.resolve("@nomicfoundation/hardhat-node-test-coverage"),
+    );
+    imports.push(coverage.href);
+    process.env.HARDHAT_NODE_TEST_COVERAGE_DIR_PATH = path.join(
+      os.tmpdir(),
```
alcuadrado, line 90: I think we have a helper for this in the `utils/fs`, if not, we should.

alcuadrado:
I think high level this looks good. We need to sync a bit on:

- How to split the behavior across packages and plugins
- How the data is collected from EDR (probably need to chat about this with Remco)
- How this feature is configured

alcuadrado:
Note that the modification to the `newConnection` hook looks good, but I think it wouldn't work, as we don't want to expose edr's config types "as is", but rather to HH to have its own config types, and expose that to the user. 

This would probably mean a change on EDR's API or making the network manager somewhat aware of the coverage functionality, or exposing some API for it.

v-next/hardhat-node-test-coverage/src/collector.ts
galargh: I see your point! In that case, I think modifying the EDR API so that we pull the coverage rather than register a callback might in fact be a better option.

That way, we could revert the modification of new connection hook signature, add a small indicator in network manager that controls whether EDR networks should be created with coverage enabled or not, and add a get call for the coverage data here once the tests are complete.

v-next/hardhat-node-test-coverage/src/collector.ts
galargh: The issue is that in the context of node:test worker, the `globalOptions.coverage` is currently `false`. We could enable it by setting `HARDHAT_COVERAGE=true` env var for the worker which would ensure the global option is set correctly.

v-next/hardhat-node-test-runner/src/task-action.ts
galargh: We don't yet, but it definitely makes sense to have one.

v-next/hardhat-node-test-coverage/src/collector.ts
galargh: Alternatively, as you're saying, we could make network manager aware of the coverage global option and give it a responsibility to act as an intermediary in handling the callbacks.

v-next/hardhat-node-test-coverage/src/collector.ts
galargh: Got you! My thinking was that it would be good to keep it separate but seeing how tiny it's going to be in the end, I don't think there's much point to it.

v-next/hardhat-node-test-runner/src/task-action.ts
galargh: I exposed the new utility here: https://github.com/NomicFoundation/hardhat/pull/6595/commits/f08b2fa9b32d58ecc28c785a7828e2b1225e5f61

v-next/hardhat-node-test-coverage/src/collector.ts
galargh: I deleted the `hardhat-node-test-coverage` package here: https://github.com/NomicFoundation/hardhat/pull/6595/commits/5f11babac5f45b0eeaa381420d513fb1e4f33b76

And replaced it with a file within `node-test-runner` here: https://github.com/NomicFoundation/hardhat/pull/6595/commits/6214a39955f6ff99c1c996c015b096759a011370

v-next/hardhat-node-test-coverage/src/collector.ts
galargh: I addressed the global options in a subprocess issue here: https://github.com/NomicFoundation/hardhat/blob/6214a39955f6ff99c1c996c015b096759a011370/v-next/hardhat-node-test-runner/src/task-action.ts#L82

As for moving handlers registration to the coverage plugin, I think we should wait with that until the final decision on the EDR API. 

I also made some observations on the condition under which we could move the entire collection logic to the coverage plugin. See https://github.com/NomicFoundation/hardhat/pull/6595/commits/47ab90c8be85071c1a84752e5c100ab7fcfd8f0d

.changeset/tiny-walls-tan.md
galargh: I updated the changesets here: https://github.com/NomicFoundation/hardhat/pull/6595/commits/74748831e478922ad8136f498e7589ec104b5365

v-next/hardhat-node-test-coverage/src/collector.ts
galargh: I did move the logic to the coverage plugin in https://github.com/NomicFoundation/hardhat/pull/6595/commits/a37b7dd07e58ea19294e564058f45916b8cafe47 and I like how tiny the collector import on the node test worker side has become.

galargh:
I'm going to put this PR back into a draft mode for now. Then, I think we should reach an agreement with EDR on the collection APIs for provider. Finally, I'm gonig to write a small design doc for this addition specifically as there are a couple of non obvious trade offs that we have to weigh in.

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/network.ts
```diff
@@ -0,0 +1,54 @@
...
+        nextContext: HookContext,
+      ) => Promise<NetworkConnection<ChainTypeT>>,
+    ) {
+      const connection = await next(context);
+
+      if (context.globalOptions.coverage === true) {
+        if (connection.provider instanceof EdrProvider) {
+          const { getOrCreateCoverageManager } = await import(
+            "../coverage-manager.js"
+          );
```
galargh, line 20: We access it via the getter so that we operate on an instance of the internal coverage manager type rather than the public one. I think technically we could force hast `context.coverage` to an internal type but it doesn't seem right either.

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -67,7 +67,19 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
   }
 
   const tsx = new URL(import.meta.resolve("tsx/esm"));
-  process.env.NODE_OPTIONS = `--import "${tsx.href}"`;
+  const imports = [tsx.href];
+
+  if (hre.globalOptions.coverage === true) {
+    // NOTE: Saving the environment variable to ensure node:test workers operate
+    // with the information that the global --coverage option is enabled
+    process.env.HARDHAT_COVERAGE = "true";
```
galargh, line 75: This feels quite hacky to be fair. Maybe we should come up with a more general way of passing down the process flags to subprocesses.

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -0,0 +1,106 @@
...
+  #coveragePath?: string;
+  #hitsPath?: string;
+
+  #getCoveragePath(): string {
+    if (this.#coveragePath === undefined) {
+      const coveragePath = process.env.HARDHAT_COVERAGE_PATH;
+      assertHardhatInvariant(
+        coveragePath !== undefined,
+        "HARDHAT_COVERAGE_PATH was not set",
+      );
```
galargh, line 38: This is to be set in the hre hook handler. One alternative would be to operate inside a well-known directory, like root level `/coverage` directory for example, and just make sure that it's cleared every time we start a new coverage collection.

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/hre.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export default async (): Promise<Partial<HardhatRuntimeEnvironmentHooks>> => {
+  const handlers: Partial<HardhatRuntimeEnvironmentHooks> = {
+    created: async (_context, hre): Promise<void> => {
+      if (process.env.HARDHAT_COVERAGE_PATH === undefined) {
+        // NOTE: Saving the environment variable so that any subprocesses that
+        // inherit the env will operate within the same coverage path
+        process.env.HARDHAT_COVERAGE_PATH =
+          await getEmptyTmpDir("hardhat-coverage");
+      }
```
galargh, line 38: This feels really hacky, too. As mentioned in https://github.com/NomicFoundation/hardhat/pull/6595/files#r2054463033, if this was a well-known directory, then we could maybe have an extra `clear` method on the coverage manager instead. Then the expected flow in the test tasks would be:
1. `clear` before any tests are run
2. `save` inside the test workers once the tests are complete
3. `load` after all the tests are done

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/network.ts
galargh: This is the alternative I had in mind:

https://github.com/NomicFoundation/hardhat/compare/coverage-6581...coverage-6581-alt#diff-257ddac31ce1a21f944e89cd8a9575932b1d0a099a5f12a73aaac13f7b1589adR19-R21 

Actually, maybe it wouldn't be that bad. Especially, in comparison to this singleton.

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
galargh: I did explore this idea a bit further and I think it is clearer than using the temporary folder. See https://github.com/NomicFoundation/hardhat/compare/coverage-6581...coverage-6581-alt

The only question is, what directory should we use. In that exploration, I added a new config path `hre.config.paths.coverage` which points to `./coverage` by default.

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/hre.ts
galargh: I explored this proposed API with additional `clear` method in https://github.com/NomicFoundation/hardhat/compare/coverage-6581...coverage-6581-alt and, in my opinion, clearing the coverage hits dir explicitly from the test task seems quite alright. Let me know what you think.

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -67,7 +67,23 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
...
```
galargh, line 75: This feels a little hacky. Maybe we should come up a more general way for passing flags to subprocesses.

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/network.ts
```diff
@@ -0,0 +1,53 @@
...
+        nextContext: HookContext,
+      ) => Promise<NetworkConnection<ChainTypeT>>,
+    ) {
+      const connection = await next(context);
+
+      if (context.globalOptions.coverage === true) {
+        if (connection.provider instanceof EdrProvider) {
+          const coverageManager: CoverageManager =
+            // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- We want to access internal coverage manager methods
+            context.coverage as unknown as CoverageManager;
```
galargh, line 21: This does feel a little hacky, too, but the alternative I explored, with making the coverage manager implementation a singleton that can be accessed outside the hre context doesn't feel right either.

galargh:
I'm closing this PR for now. I will reuse its' content to implement https://www.notion.so/nomicfoundation/Instrumentation-metadata-and-node-test-coverage-data-collection-1df578cdeaf580eb8847dd730a496df5 once it's finalised.

PR #6605, opened 2025-04-20
Merge solc output selection config

### LINKED ISSUE #6551: Cannot add `storageLayout` to compiler output selection
### Feedback

Using version Hardhat v3 Alpha version `3.0.0-next.3`, I cannot change the Solidity compiler settings to include `storageLayout` in the output.

---

Steps to reproduce:
1. Create a sample Hardhat v3 Alpha project according to https://hardhat.org/hardhat3-alpha, by using `npx hardhat@next --init`
2. Under the `default` profile, add an `outputSelection` section with `storageLayout`, for example:
```
  plugins: [
    HardhatNodeTestRunner,
    HardhatViem,
    HardhatNetworkHelpers,
    HardhatKeystore,
    HardhatIgnitionViem,
  ],
  solidity: {
    profiles: {
      default: {
        version: "0.8.28",
        settings: {
          outputSelection: {
            '*': {
              '*': ['storageLayout'],
            },
          },
        }
      },
...
```
3. Run `npx hardhat clean && npx hardhat compile`

Notice `storageLayout` is not included in the Solidity input's `outputSelection` in the build info dir, which only has:
```
  "input": {
    "language": "Solidity",
    "settings": {
      "outputSelection": {
        "*": {
          "*": [
            "abi",
            "evm.bytecode",
            "evm.deployedBytecode",
            "evm.methodIdentifiers",
            "metadata"
          ],
          "": [
            "ast"
          ]
        }
      },
...
```
And the Solidity output json does not contain storage layout.

---

Other steps I've tried:

Even if not using build profiles (without using `profiles` in hardhat.config.ts), the same issue occurs.
For example:
```
  plugins: [
    HardhatNodeTestRunner,
    HardhatViem,
    HardhatNetworkHelpers,
    HardhatKeystore,
    HardhatIgnitionViem,
  ],
  solidity: {
    version: "0.8.28",
    settings: {
      optimizer: {
        enabled: true,
        runs: 100,
      },
      outputSelection: {
        '*': {
          '*': ['storageLayout'],
        },
      },
    },
...
```
The other settings (e.g. `optimizer`) are included in the resulting Solidity input json, but `storageLayout` is not.

ItsNickBarry:
closes #6551

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
@@ -99,54 +99,56 @@ export class CompilationJobImplementation implements CompilationJob {
     // from other files (e.g. new Foo()), and it won't output its bytecode if
     // it's not asked for. This would prevent EDR from doing any runtime
     // analysis.
-    const defaultOutputSelection: CompilerInput["settings"]["outputSelection"] =
-      {
-        "*": {
-          "": ["ast"],
-          "*": [
-            "abi",
-            "evm.bytecode",
-            "evm.deployedBytecode",
-            "evm.methodIdentifiers",
-            "metadata",
-          ],
-        },
-      };
+    type OutputSelection = CompilerInput["settings"]["outputSelection"];
+    const defaultOutputSelection: OutputSelection = {
+      "*": {
+        "": ["ast"],
+        "*": [
+          "abi",
+          "evm.bytecode",
+          "evm.deployedBytecode",
+          "evm.methodIdentifiers",
+          "metadata",
+        ],
+      },
+    };
 
-    const configOutputSelection = settings.outputSelection ?? {};
-    const outputSelection: typeof defaultOutputSelection = {};
+    // TODO: export this function somewhere as helper
+    const mergeOutputSelection = (
+      selectionA: OutputSelection,
+      selectionB: OutputSelection,
+    ): OutputSelection => {
+      const result: OutputSelection = {};
 
-    const fileKeys = Array.from(
-      new Set([
-        ...Object.keys(defaultOutputSelection),
-        ...Object.keys(configOutputSelection),
-      ]),
-    ).sort();
-
-    for (const fileKey of fileKeys) {
-      const contractKeys = Array.from(
-        new Set([
-          ...Object.keys(defaultOutputSelection[fileKey] ?? {}),
-          ...Object.keys(configOutputSelection[fileKey] ?? {}),
-        ]),
+      const fileKeys = Array.from(
+        new Set([...Object.keys(selectionA), ...Object.keys(selectionB)]),
       ).sort();
 
-      for (const contractKey of contractKeys) {
-        const values = Array.from(
+      for (const fileKey of fileKeys) {
+        const contractKeys = Array.from(
           new Set([
-            ...Object.keys(
-              defaultOutputSelection[fileKey]?.[contractKey] ?? [],
-            ),
-            ...Object.keys(configOutputSelection[fileKey]?.[contractKey] ?? []),
+            ...Object.keys(selectionA[fileKey] ?? {}),
+            ...Object.keys(selectionB[fileKey] ?? {}),
           ]),
         ).sort();
 
-        if (values.length > 0) {
-          outputSelection[fileKey] = {};
-          outputSelection[fileKey][contractKey] = values;
+        for (const contractKey of contractKeys) {
+          const values = Array.from(
+            new Set([
+              ...Object.keys(selectionA[fileKey]?.[contractKey] ?? []),
+              ...Object.keys(selectionB[fileKey]?.[contractKey] ?? []),
+            ]),
+          ).sort();
+
+          if (values.length > 0) {
+            result[fileKey] = {};
+            result[fileKey][contractKey] = values;
+          }
         }
       }
-    }
+
+      return result;
+    };
 
     return {
       language: "Solidity",
@@ -155,7 +157,10 @@ export class CompilationJobImplementation implements CompilationJob {
         evmVersion:
           settings.evmVersion ??
           getEvmVersionFromSolcVersion(this.solcConfig.version),
-        outputSelection,
+        outputSelection: mergeOutputSelection(
+          defaultOutputSelection,
+          settings.outputSelection,
+        ),
         remappings: this.#remappings.map(formatRemapping),
       },
     };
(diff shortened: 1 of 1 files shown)


ItsNickBarry:
I suppose the object keys also need to be sorted, but the default is not sorted.  `""` should come before `"*"`.

ItsNickBarry:
This would be useful as an exported helper function so that plugins can make additions to the output selection without overwriting.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -113,9 +113,46 @@ export class CompilationJobImplementation implements CompilationJob {
         },
       };
 
-    // TODO: Deep merge the user output selection with the default one
     const outputSelection = defaultOutputSelection;
 
+    const configOutputSelection = settings.outputSelection ?? {};
+
+    for (const fileKey in configOutputSelection) {
```
zoeyTM, line 120: Would prefer this pattern over the `hasOwnProperty` checks

```suggestion
    for (const fileKey of Object.keys(configOutputSelection)) {
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -113,9 +113,46 @@ export class CompilationJobImplementation implements CompilationJob {
...
+      if (
+        !Object.prototype.hasOwnProperty.call(configOutputSelection, fileKey)
+      ) {
+        continue;
+      }
```
zoeyTM, line 125:
```suggestion
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -113,9 +113,46 @@ export class CompilationJobImplementation implements CompilationJob {
...
+
+      const configFileOutputSelection = configOutputSelection[fileKey] ?? {};
+      outputSelection[fileKey] ??= {};
+
+      for (const contractKey in configFileOutputSelection) {
```
zoeyTM, line 130: Same as above 

```suggestion
      for (const contractKey of Object.keys(configFileOutputSelection)) {
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -113,9 +113,46 @@ export class CompilationJobImplementation implements CompilationJob {
...
+        if (
+          !Object.prototype.hasOwnProperty.call(
+            configFileOutputSelection,
+            contractKey,
+          )
+        ) {
+          continue;
+        }
```
zoeyTM, line 138:
```suggestion
```

zoeyTM (changes requested):
Will approve after a couple of minor changes, but this looks great overall. Thanks!

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -102,20 +102,41 @@ export class CompilationJobImplementation implements CompilationJob {
...
-    // TODO: Deep merge the user output selection with the default one
     const outputSelection = defaultOutputSelection;
 
+    const configOutputSelection = settings.outputSelection ?? {};
+
+    for (const fileKey of Object.keys(configOutputSelection).sort()) {
+      const configFileOutputSelection = configOutputSelection[fileKey] ?? {};
+      outputSelection[fileKey] ??= {};
+
+      for (const contractKey of Object.keys(configFileOutputSelection).sort()) {
```
ItsNickBarry, line 124: I don't think sorting the keys is strictly necessary because they order of keys isn't expected to change, but it's better to keep the object in a normalized state.

ItsNickBarry: Maybe not, actually.  I can't quite contrive a situation in which a plugin would reasonably change key order without adding or removing keys.

ItsNickBarry:
@zoeyTM What do you think of this?  Everything is sorted and empty entries are ignored.

```typescript
const configOutputSelection = settings.outputSelection ?? {};
const outputSelection: typeof defaultOutputSelection = {};

const fileKeys = Array.from(
  new Set([
    ...Object.keys(defaultOutputSelection),
    ...Object.keys(configOutputSelection),
  ]),
).sort();

for (const fileKey of fileKeys) {
  const contractKeys = Array.from(
    new Set([
      ...Object.keys(defaultOutputSelection[fileKey] ?? {}),
      ...Object.keys(configOutputSelection[fileKey] ?? {}),
    ]),
  ).sort();

  for (const contractKey of contractKeys) {
    const values = Array.from(
      new Set([
        ...Object.keys(
          defaultOutputSelection[fileKey]?.[contractKey] ?? [],
        ),
        ...Object.keys(configOutputSelection[fileKey]?.[contractKey] ?? []),
      ]),
    ).sort();

    if (values.length > 0) {
      outputSelection[fileKey] = {};
      outputSelection[fileKey][contractKey] = values;
    }
  }
}
```

ItsNickBarry:
Shouldn't all of this go in the config hook?

ItsNickBarry:
I was thinking that both users and plugin developers would expect to be able to pass in an output selection object with the same standard structure.  The object isn't even documented (https://github.com/ethereum/solidity/issues/15744), so it's uncomfortable to manipulate directly.

The HH solidity module is also a plugin, and doesn't need to behave any differently.  Isn't this a config resolution problem that should be done in the resolveConfig hook?

This is what the code looks like without a helper:

```typescript
const settings = compiler.settings;
settings.outputSelection ??= {};
settings.outputSelection['*'] ??= {};
settings.outputSelection['*']['*'] ??= [];

if (!settings.outputSelection['*']['*'].includes('storageLayout')) {
  settings.outputSelection['*']['*'].push('storageLayout');
}
```

It's not that bad, but needs to be duplicated in at least 4 places:
* `builtin:solidity`
* `hardhat-docgen`
* `hardhat-storage-layout-diff` (in two places)

I'd rather do something like this:

```typescript
const STORAGE_LAYOUT_OUTPUT_SELECTION = { '*': { '*': ['storageLayout'] } };

settings.outputSelection = mergeOutputSelection(settings.outputSelection, STORAGE_LAYOUT_OUTPUT_SELECTION);
```

Then, nobody has to think about any of these issues:
* overwriting other plugins' entries
* nonexistent entries (especially if Hardhat waits until just before compilation to add the "*" defaults)
* duplicate entries
* changes in sort order leading to unnecessary recompilation

kanej:
I am closing this PR, thanks @ItsNickBarry, the approach was encoded in these two PRs:

* https://github.com/NomicFoundation/hardhat/pull/6668
* https://github.com/NomicFoundation/hardhat/pull/6678

The discussion around exposing the helper function to plugin authors I am going to capture in a separate issue: https://github.com/NomicFoundation/hardhat/issues/6704.

Issue #6603, opened 2025-04-19
A way to disable chainId checking if chainId is greater than 2^32 - 1


sugardave:
### Describe the feature

I am using Hardhat to write smart contracts to deploy to a new EVM-compatible chain's testnet where the chain ID is `200820172034`.  Unfortunately, deploying with ignition or my own deploy script fails with an error from the `micro-eth-signer` package's `validateFields` method that the value
```
'must be >= 1 and <= 2**32-1, not 200820172034'
```
I initially opened an issue at [https://github.com/paulmillr/micro-eth-signer/issues/34](https://github.com/paulmillr/micro-eth-signer/issues/34) to ask for a way to override the check, but the response was to change the way Hardhat prepares a `Transaction` to set `strict = false`.  However, I haven't really been able to figure out how to do that simply.

My current workaround uses `patch-package` to increase the range of acceptable values for `maxChainId` in `micro-eth-signer` so I can keep my Hardhat installation and configuration as close to "regular" as I can.

It would be great if there were a way to specify that either the `chainId` verification be ignored, some way to tell Hardhat you want it to prepare all/some transactions with `strict = false`, or to pass the `type = legacy` prop to the `chainId` method in [micro-eth-signer/src/tx.js](https://github.com/paulmillr/micro-eth-signer/blob/baa4b8c922c3253b125e3f46b1fce6dee7c33853/src/tx.ts#L496-L500)

alcuadrado:
I think it would be better to do pass `strict = false` only if the `chainId` is greater than that number.

We should implement this in v2 and v3.

alcuadrado:
This is the file in v2: https://github.com/NomicFoundation/hardhat/blob/52f091ee1df22833c59e44dcf6191538213bffc3/packages/hardhat-core/src/internal/core/providers/accounts.ts

This is the file in v3: https://github.com/NomicFoundation/hardhat/blob/v-next/v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts

Both look similar.

I think we should add a good comment explaining that it's a workaround to this issue.

kanej:
Can I confirm, with @iosh merged PR this is now resolved for v2 but needs ported to v3?

schaable:
> Can I confirm, with [@iosh](https://github.com/iosh) merged PR this is now resolved for v2 but needs ported to v3?

[...]

PR #6607, opened 2025-04-20
feat: support chainId values above 2^32 - 1 for local account transac…

iosh:
…tions

---

#6603 

This PR add chainId check. When chainId larger than 2^32 - 1， set `strict=false` and pass this to Transaction.prepare function

### DIFF
--- a/.changeset/weak-frogs-care.md
+++ b/.changeset/weak-frogs-care.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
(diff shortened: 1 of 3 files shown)


schaable (approved):
Looks great! Thanks for fixing this ❤️

schaable:
@iosh I expanded the comment a bit and simplified the logic using a ternary — hope you don’t mind. 🙂

Issue #6596, opened 2025-04-17
Some `GlobalOptions` and `TaskArguments` should be able to have `undefined` as default value


alcuadrado:
Right now we have many task arguments and global options that use `""` as default, when in reality they just don't want any default, and compute the right value at runtime. For example the default build profile depends on the task you are running.

We should make a few changes to allow this:

* [ ] `ArgumentType` should have an extra element `STRING_WITHOUT_DEFAULT`, which maps to `string | undefined`
* [ ] When parsing and resolving the arguments we should set `undefined` as default.
* [ ] We should update the help messages to handle that type.
* [ ] We should update all the places were we set `""` as default and use `STRING_WITHOUT_DEFAULT` instead.

PR #6615, opened 2025-04-23
Allow undefined for default values in global options and other argument types

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat/issues/6596


v-next/hardhat-ignition/src/index.ts
```diff
@@ -29,19 +29,19 @@ const hardhatIgnitionPlugin: HardhatPlugin = {
         type: ArgumentType.FILE,
         description:
           "A relative path to a JSON file to use for the module parameters",
-        defaultValue: "", // TODO: HH3 check this comes through correctly
+        defaultValue: "", // TODO: should this also be STRING_WITHOUT_DEFAULT?
```
kanej, line 32: Lets resolve whether this arg should be a File.

kanej: Currently it appears that the File and String types are treated the same. So some questions:

- Should we remove the File type
- Should this parameter just become a string
- Should there be a FILE_WITHOUT_DEFAULT

v-next/hardhat/src/internal/core/config-validation.ts
```diff
@@ -457,6 +472,7 @@ export function validatePositionalArguments(
 
     if (arg.defaultValue !== undefined) {
       switch (arg.type) {
+        case ArgumentType.STRING_WITHOUT_DEFAULT:
         case ArgumentType.STRING:
         case ArgumentType.FILE: {
           if (
```
kanej, line 478: Why do we need array checks here?
When we look at usages, arrays do not appear expressible in the type system.
The message suggests you can set to an array of strings but the type system blocks it.

v-next/hardhat/src/internal/core/global-options.ts
```diff
@@ -80,7 +80,7 @@ export function buildGlobalOptionDefinition<T extends ArgumentType>({
   type?: T;
   defaultValue: ArgumentTypeToValueType<T>;
 }): OptionDefinition {
-  const argumentType = type ?? ArgumentType.STRING;
+  const argumentType = type ?? ArgumentType.STRING_WITHOUT_DEFAULT;
```
kanej, line 83: This would go away if we remove the optionality on type for global options.
If we don't think it should be optional should we keep it as `ArgumentType.string`?

v-next/hardhat-ignition/src/index.ts
kanej: We decided to add `FILE_WITHOUT_DEFAULT`

v-next/hardhat/src/internal/core/global-options.ts
kanej: We are keeping type optionality and setting the type of `T` to `STRING` explicitly, so this should be reverted so that string is the runtime default as well

kanej (approved):
LGTM!

PR #6628, opened 2025-04-24
Disable hardhat-toolbox-mocha-ethers flaky test until issue is fixed

ChristopherDedominici:


v-next/hardhat-toolbox-mocha-ethers/test/index.ts
```diff
@@ -11,7 +11,7 @@ describe("hardhat-toolbox-mocha-ethers", function () {
   describe("all the expected plugins are available", function () {
     useFixtureProject("toolbox");
 
-    it("should not throw because all the plugins should exist", async function () {
+    it.skip("should not throw because all the plugins should exist", async function () {
```
kanej, line 14:
```suggestion
    // TODO: This test should be brought back once an intermittent bug in EDR is resolved
    it.skip("should not throw because all the plugins should exist", async function () {
```

kanej (approved):
Pre-approving, but I suggested an additional comment.

PR #6630, opened 2025-04-24
hardhat-vscode integration support

antico5:
This PR is meant to introduce all changes necessary for supporting hardhat3 projects on our vscode extension.

At the time, I aimed to introduce the bare minimal changes that would allow this integration:
- Export the DependencyResolver class. The LSP can build its own dependency graph using this
- Introduce a hook for reading solidity source files. This is necessary so we can read unsaved files from the editor
- Inject a "readSourceFile" function down from the build system to the dependency resolver. As a first approach I was passing the HookManager down, but was suggested to instead use a function to avoid coupling. Then I had to define a helper factory and export it since I also need to use it from the LSP.

### DIFF
--- a/v-next/hardhat/package.json
+++ b/v-next/hardhat/package.json
@@ -20,6 +20,7 @@
     "./config": "./dist/src/config.js",
     "./hre": "./dist/src/hre.js",
     "./plugins": "./dist/src/plugins.js",
+    "./dependency-resolver": "./dist/src/dependency-resolver.js",
     "./types/arguments": "./dist/src/types/arguments.js",
     "./types/artifacts": "./dist/src/types/artifacts.js",
     "./types/config": "./dist/src/types/config.js",
--- a/v-next/hardhat/package.json
+++ b/v-next/hardhat/package.json
@@ -20,7 +20,7 @@
     "./config": "./dist/src/config.js",
     "./hre": "./dist/src/hre.js",
     "./plugins": "./dist/src/plugins.js",
-    "./dependency-resolver": "./dist/src/dependency-resolver.js",
+    "./lsp-helpers": "./dist/src/lsp-helpers.js",
     "./types/arguments": "./dist/src/types/arguments.js",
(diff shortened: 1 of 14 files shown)


v-next/hardhat/package.json
```diff
@@ -20,6 +20,7 @@
     "./config": "./dist/src/config.js",
     "./hre": "./dist/src/hre.js",
     "./plugins": "./dist/src/plugins.js",
+    "./lsp-helpers": "./dist/src/lsp-helpers.js",
```
antico5, line 23: prepend /internal

v-next/hardhat/src/lsp-helpers.ts
```diff
@@ -0,0 +1,3 @@
+export { importUserConfig } from "./internal/config-loading.js";
```
antico5, line 1: move to hre exports

v-next/hardhat/src/internal/builtin-plugins/solidity/type-extensions.ts
```diff
@@ -114,5 +114,11 @@ declare module "../../../types/hooks.js" {
         artifactPaths: string[],
       ) => Promise<void>,
     ) => Promise<void>;
+
+    readSourceFile: (
+      context: HookContext,
+      absPath: string,
```
antico5, line 120: change to absolutePath, nextAbsolutePath

v-next/hardhat/src/types/solidity/build-system.ts
```diff
@@ -2,6 +2,8 @@ import type { CompilationJob } from "./compilation-job.js";
 import type { CompilerOutput, CompilerOutputError } from "./compiler-io.js";
 import type { SolidityBuildInfo } from "./solidity-artifacts.js";
 
+export type { Resolver } from "../../internal/builtin-plugins/solidity/build-system/resolver/types.js";
```
antico5, line 5: move to lsp helpers

alcuadrado (approved):

PR #6631, opened 2025-04-24
feat: bump @ignored/edr to 0.10.0-alpha.3

agostbiro:
Bumps `@ignored/edr` to `0.10.0-alpha.3` which brings the following changes:

- Back-ports fuzz and invariant testing improvements from Foundry 1.0 (commit: a2ecefce6), notably:
  - Significant performance improvements
  - Support for `afterInvariant` function
  - Support for linked artifacts in `get*Code` cheatcodes
  - Various bug fixes
- fix: support different solc versions for libs and contracts in linker

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -96,8 +96,8 @@ importers:
   v-next/hardhat:
     dependencies:
       '@ignored/edr':
-        specifier: 0.10.0-alpha.2
-        version: 0.10.0-alpha.2
+        specifier: 0.10.0-alpha.3
+        version: 0.10.0-alpha.3
       '@ignored/edr-optimism':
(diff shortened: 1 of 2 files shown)


kanej (approved):
Looks good. Thanks for the detailed changeset!

kanej:
The linting error does not appear to be in `v-next`. The pnpm lock file doesn't appear to have changed anything other than EDR.

kanej (changes requested):
It looks like there is a subtle change to `TestResult` that trips our eslint rule.

counterexample's array option changed from an array to an object:

Previous Type
image

New types
image

image

I think we need to update the testReporter code to deal with the new structure.

galargh (approved):

kanej (approved):

Issue #6612, opened 2025-04-22
Move `chains` to network config top level and enhance network definitions


schaable:
To do:
 - Move chains to a top-level field in the network config
 - Add the necessary information for verification tools (Etherscan, Sourcify, Blockscout)
 - Include any required information to support Viem
 - Add any missing relevant networks, such as Arbitrum, Avalanche, and Polygon POS
 - Remove hardforks in the `hardforkHistory` map as EDR no longer requires them, unless the user wants to intentionally overwrite them.

PR #6635, opened 2025-04-25
Chains config refactor

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/6612

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
@@ -103,7 +103,9 @@ export async function resolveUserConfig(
 
   return {
     ...resolvedConfig,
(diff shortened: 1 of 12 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -104,7 +104,7 @@ export async function resolveUserConfig(
   return {
     ...resolvedConfig,
     chainDescriptors: await resolveChainDescriptors(
-      resolvedConfig.chainDescriptors,
+      userConfig.chainDescriptors,
```
schaable, line 107: I’m not sure this is correct, but we’re doing the same thing with the networks. Shouldn’t we always use the resolved version to account for other hook handler changes? Doing so causes type issues, though, since I’d be resolving a field that’s already resolved.

alcuadrado (approved):

PR #6645, opened 2025-04-28
build: upgrade to @ignored/edr-optimism v0.13.0-alpha.1

Wodann:
This PR upgrades `@ignored/edr-optimism` to v0.13.0-alpha.1, including the following changes:
- Added instrumenting of source code for statement code coverage measurement
- Added code coverage to the provider. It can be configured through the ProviderConfig


galargh:
@Wodann Is it OK if we merge this upgrade as-is and later add actual coverage handling on the hardhat side?

galargh (approved):

Wodann:
> @Wodann Is it OK if we merge this upgrade as-is and later add actual coverage handling on the hardhat side?

Yeah, if you'd like, it shouldn't break anything.

Would you like me to remove the `TODO` comment in that case?

Issue #6647, opened 2025-04-29
Contract reverted without a reason string in Solidity test


agostbiro:
### Version of Hardhat

3.0.0-next.5

### What happened?

Hardhat doesn't report revert reason in a Solidity test.

### Minimal reproduction steps

**Steps to reproduce**

1. Check out https://github.com/agostbiro/hh3-repro
2. `pnpm i`
3. `pnpm hardhat test solidity`

**Expected result**

The `test_IncByZero` test should fail with the reason "incBy: increment should be positive".

**Actual result**

The `test_IncByZero` test fails with the reason: "Contract reverted without a reason string"

**Notes**

I verified that EDR returns the revert reason "incBy: increment should be positive" as part of the `TestResult`.

### Search terms

solidity test revert reason

galargh:
This has been resolved in https://github.com/NomicFoundation/hardhat/pull/6648. A more general follow-up issue has been created - https://github.com/NomicFoundation/hardhat/issues/6651

PR #6648, opened 2025-04-29
fix: remove contract reverted without a reason string message from solidity tests

galargh:

---

This fixes the issue reported in #6647.

As a follow-up, we should review the logic of extracting a more detailed error message from stack trace (https://github.com/NomicFoundation/hardhat/blob/999a8153bdedde5435432a28cd7cafd30fac8a90/v-next/hardhat/src/internal/builtin-plugins/solidity-test/stack-trace-solidity-errors.ts#L6) than the failure reason we get here (https://github.com/NomicFoundation/hardhat/blob/999a8153bdedde5435432a28cd7cafd30fac8a90/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts#L132). cc @kanej @agostbiro 

#### Testing

I tested this on the reproduction repo provided here - https://github.com/agostbiro/hh3-repro

Before:
```
Failure (1): test_IncByZero()
Reason: Contract reverted without a reason string
  at Counter.incBy (contracts/Counter.sol:15)
  at CounterTest.test_IncByZero (contracts/Counter.t.sol:18)
```

After:
```
Failure (1): test_IncByZero()
Reason: revert: incBy: increment should be positive
  at Counter.incBy (contracts/Counter.sol:15)
  at CounterTest.test_IncByZero (contracts/Counter.t.sol:18)
```

### DIFF
--- a/.changeset/chatty-rocks-reply.md
+++ b/.changeset/chatty-rocks-reply.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Replace "Contract reverted without a reason string" message with a more detailed failure reason in solidity tests
--- a/.changeset/chatty-rocks-reply.md
+++ b/.changeset/chatty-rocks-reply.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Replace "Contract reverted without a reason string" message with a more detailed failure reason in solidity tests
+Replace "Contract reverted without a reason string" message with a more detailed failure reason in solidity tests ([#6647](https://github.com/NomicFoundation/hardhat/issues/6647))
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/stack-trace-solidity-errors.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/stack-trace-solidity-errors.ts
@@ -46,13 +46,15 @@ export function getMessageFromLastStackTraceEntry(
     case StackTraceEntryType.DIRECT_LIBRARY_CALL_ERROR:
       return `Library was called directly`;
 
+    /* These types are not associate with a more detailed error message */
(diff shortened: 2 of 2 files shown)


agostbiro (approved):
Thanks, LGTM

.changeset/chatty-rocks-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Replace "Contract reverted without a reason string" message with a more detailed failure reason in solidity tests
```
kanej, line 5:
```suggestion
Replace "Contract reverted without a reason string" message with a more detailed failure reason in solidity tests ([#6647](https://github.com/NomicFoundation/hardhat/issues/6647))
```

v-next/hardhat/src/internal/builtin-plugins/solidity-test/stack-trace-solidity-errors.ts
```diff
@@ -46,13 +46,15 @@ export function getMessageFromLastStackTraceEntry(
     case StackTraceEntryType.DIRECT_LIBRARY_CALL_ERROR:
       return `Library was called directly`;
 
+    /* These types are not associate with a more detailed error message */
     case StackTraceEntryType.UNRECOGNIZED_CREATE_ERROR:
     case StackTraceEntryType.UNRECOGNIZED_CONTRACT_ERROR: {
-      return "Contract reverted without a reason string";
+      return undefined;
```
kanej, line 52: To check I have understood. For these error types, we now fallback to the reason that is passed from EDR (rather than a hard coded string on the Hardhat side). That fallback is managed by passing undefined which indicates we should read the error message from the EDR side?

galargh: Yes, that's correct!

PR #6652, opened 2025-04-30
[WIP - not ready for review] Collect instrumentation metadata and node:test coverage data

galargh:

---

This PR implements https://www.notion.so/nomicfoundation/Instrumentation-metadata-and-node-test-coverage-data-collection-1df578cdeaf580eb8847dd730a496df5

Please note the design document has not yet been approved and is subject to change. Hence, this PR is not ready for review yet.

The other blocker for this PR is the fact that the instrumentation process adds `hardhat/coverage.sol` imports to the files. These imports are currently unresolvable because `coverage.sol` is not available yet. This should be addressed before marking this PR as ready for review. In particular, this might mean that we have to move the instrumentation earlier in the process i.e. before dependency graph construction.

### DIFF
--- a/v-next/hardhat/package.json
+++ b/v-next/hardhat/package.json
@@ -33,7 +33,8 @@
     "./types/user-interruptions": "./dist/src/types/user-interruptions.js",
     "./types/utils": "./dist/src/types/utils.js",
     "./types/solidity": "./dist/src/types/solidity.js",
-    "./console.sol": "./console.sol"
+    "./console.sol": "./console.sol",
+    "./internal/coverage": "./dist/src/internal/builtin-plugins/coverage/exports.js"
   },
   "keywords": [
     "ethereum",
--- a/v-next/hardhat/package.json
+++ b/v-next/hardhat/package.json
@@ -89,7 +89,7 @@
   },
   "dependencies": {
     "@ignored/edr": "0.10.0-alpha.3",
-    "@ignored/edr-optimism": "0.13.0-alpha.0",
+    "@ignored/edr-optimism": "0.13.0-alpha.1",
     "@nomicfoundation/hardhat-errors": "workspace:^3.0.0-next.5",
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.0-next.5",
     "@nomicfoundation/hardhat-zod-utils": "workspace:^3.0.0-next.5",
--- a/v-next/hardhat/src/internal/builtin-plugins/coverage/helpers.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/coverage/helpers.ts
@@ -1,0 +1,36 @@
+import type { HardhatRuntimeEnvironment } from "../../../types/hre.js";
+import type { HardhatRuntimeEnvironmentImplementation } from "../../core/hre.js";
+
+import { getOrCreateGlobalHardhatRuntimeEnvironment } from "../../hre-intialization.js";
+
+export function unsafelyCastHardhatRuntimeEnvironmentImplementation(hre: HardhatRuntimeEnvironment): HardhatRuntimeEnvironmentImplementation {
+  const hreImplementation =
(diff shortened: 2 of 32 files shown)


v-next/hardhat/src/internal/builtin-plugins/coverage/helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * NOTE: The following helpers interact with the global HRE instance only;
+ * This is OK because:
+ * - They are intended for the internal use only. They are exposed via the
+ *   internal public API only.
+ * - We know the HRE has been initialized by the time they are used.
+ */
+
+export async function clearCoverageData(): Promise<void> {
```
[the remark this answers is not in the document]
galargh: These functions could also take in a string ID as an argument. Then each test task would operate under its' own namespace under coverage manager. This would mean fewer chances for any potential races and this could actually solve how the data generated for the sub-test-tasks could be aggregated in the parent test task. I'm going to write a short doc on this.

v-next/hardhat/package.json
```diff
@@ -60,6 +61,7 @@
     "src/",
     "templates/",
     "console.sol",
+    "coverage.sol",
```
alcuadrado, line 64: I don't think this is needed.

v-next/hardhat/src/internal/builtin-plugins/coverage/helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+  const hre = await getOrCreateGlobalHardhatRuntimeEnvironment();
```
alcuadrado, line 36: I think doing `await import("../../.../index.js")` here would be better, as it ensure that the HRE has the same initialization as the one the user would expect. i.e. importing the same file that the user would import as `import hre from "hardhat"`

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/solidity.ts
```diff
@@ -0,0 +1,93 @@
...
+        const { source, metadata } = addStatementCoverageInstrumentation(
+          fileContent,
+          sourceName,
+          solcVersion,
+          COVERAGE_LIBRARY_PATH,
+        );
+        // NOTE: We need to cast the hre to the internal HardhatRuntimeEnvironmentImplementation
+        // because the coverage manager (hre._coverage) is not exposed via the public interface
+        const hreImplementation =
+          unsafelyCastAsHardhatRuntimeEnvironmentImplementation(context);
```
alcuadrado, line 36: This part is slightly more confused than needed, as the context is not an HRE.

```
assertHardhatInvariant("_coverage" in context && context._coverage instanceof CoverageManager, "Expected _coverage to be defined in the HookContext, as it's should be defined in the HRE");
```

would be simpler to understand.

alcuadrado (approved):
I left some comments, but it looks good in general

v-next/hardhat/src/internal/builtin-plugins/coverage/helpers.ts
alcuadrado: Having reviewed the other PR now, maybe this is better done on a separate PR, to avoid the need for rebasing the other one.

v-next/hardhat/package.json
galargh: You're right! I added it here before doing the trick with modifying solc inputs.

v-next/hardhat/src/internal/builtin-plugins/coverage/helpers.ts
galargh: I merged the other PR into this one and then addressed this. I added the import and, instead of the cast, I did the assert of `HardhatRuntimeEnvironmentImplementation` instead. Similarly, to what you suggested with getting coverage manager from hook context.

```
const { default: hre } = await import("../../../index.js");
assertHardhatInvariant(hre instanceof HardhatRuntimeEnvironmentImplementation, "...");
```

Issue #6545, opened 2025-04-07
creation inside of workspace causes errors


3commascapital:
### Version of Hardhat

3.0.0-next.3

### What happened?

errors while creating a sample project

### Minimal reproduction steps

create a npm workspace, with defined workspaces inside of a `packages` folder. cd into your `packages` folder, try to run init command `npx hardhat@next --init` find that it requires the new folder to be created before init begins: 
```sh
Error HHE17: Workspace "xxx/packages/contracts-hh" not found
```

schaable:
Hi @3commascapital, thanks for reporting this! This is expected behavior, but I agree that Hardhat should create the folder for you, like most project initialization tools do.

ChristopherDedominici:
Fixed with https://github.com/NomicFoundation/hardhat/pull/6653

PR #6653, opened 2025-04-30
Create folder if not exists during `--init` 

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/6545

### DIFF
--- a/v-next/hardhat-errors/src/descriptors.ts
+++ b/v-next/hardhat-errors/src/descriptors.ts
@@ -397,11 +397,11 @@ Please add the property "type" with the value "module" in your package.json to e
         websiteTitle: "Template not found",
         websiteDescription: `The template you provided is not found. Please check the documentation to learn which templates are available.`,
       },
-      WORKSPACE_NOT_FOUND: {
+      WORKSPACE_MUST_BE_A_DIRECTORY: {
         number: 17,
-        messageTemplate: `Workspace "{workspace}" not found`,
-        websiteTitle: "Workspace not found",
-        websiteDescription: `The workspace you provided does not exist. Please ensure that the workspace exists and try again.`,
+        messageTemplate: `Workspace "{workspace}" must be a directory`,
+        websiteTitle: "Workspace must be a directory",
+        websiteDescription: `The workspace you provided is not a directory. Please ensure that the workspace is a directory and try again.`,
       },
       INVALID_HEX_STRING: {
         number: 18,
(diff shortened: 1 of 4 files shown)


v-next/hardhat/test/internal/cli/init/init.ts
```diff
@@ -389,30 +397,38 @@ describe("initHardhat", async () => {
...
-        const workspaceFiles = template.files.map(
-          relativeTemplateToWorkspacePath,
-        );
-        for (const file of workspaceFiles) {
-          const pathToFile = path.join(process.cwd(), file);
-          assert.ok(await exists(pathToFile), `File ${file} should exist`);
-        }
-      },
-    );
+    // Verifies that non-existent folders are created during initialization instead of throwing an error
```
galargh, line 400: Can we make it a separate test? I don't think there's a need to check it for every template.

v-next/hardhat/test/internal/cli/init/init.ts
```diff
@@ -382,38 +390,84 @@ describe("installProjectDependencies", async () => {
...
-        for (const file of workspaceFiles) {
-          const pathToFile = path.join(process.cwd(), file);
-          assert.ok(await exists(pathToFile), `File ${file} should exist`);
-        }
-      },
-    );
-  }
+    for (const template of templates) {
+      // NOTE: This test uses network to access the npm registry
+      it(
```
ChristopherDedominici, line 402: This test is the same, only the formatting has changed because it’s now wrapped in an additional `describe`

v-next/hardhat/test/internal/cli/init/init.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6653/commits/7cb1dee8c15de06fd31dbb980f1bcbdac025fba4

v-next/hardhat/test/internal/cli/init/init.ts
galargh: FYI, when reviewing you can hide whitespace changes - https://github.com/NomicFoundation/hardhat/pull/6653/files?diff=split&w=1

It's ideal for cases like this one

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -397,11 +397,11 @@ Please add the property "type" with the value "module" in your package.json to e
         websiteTitle: "Template not found",
         websiteDescription: `The template you provided is not found. Please check the documentation to learn which templates are available.`,
       },
-      WORKSPACE_NOT_FOUND: {
```
galargh, line 400: @ChristopherDedominici Just one more q, it's OK to change this because we're still in alpha, right?

PR #6662, opened 2025-05-05
Fix instance of dead code

ItsNickBarry:
Without this change, the `if` block on line 41 (now line 39) does not run.


schaable (approved):
I don't believe this is technically dead code, but in v2 it was only used by `hardhat-vyper`. We can remove it and add it back if we find out we need it in the future.

schaable:
@ItsNickBarry there's a related test, can you remove it as well?

ItsNickBarry:
Which test?  The only change is that the `log` is now able to run.  I searched the logged string but didn't find anywhere that it's tested.

alcuadrado (approved):

alcuadrado:
We just reviewed this. You were right! Thanks, @ItsNickBarry!

Issue #6672, opened 2025-05-06
Improve the API of network.connect by supporting param objects


kanej:
Update the signature of the `network.connect` method to take a params object: {network: "foo", chainType: "bar"}

In effect we move from:

```ts
export interface NetworkManager {
  connect<ChainTypeT extends ChainType = DefaultChainType>(
    networkName?: string,
    chainType?: ChainTypeT,
    networkConfigOverride?: NetworkConfigOverride,
  ): Promise<NetworkConnection<ChainTypeT>>;
}
```

to

```
export interface NetworkManager {
  connect<ChainTypeT extends ChainType = DefaultChainType>(
   networkConnectionParams?: NetworkConnectionParams
  ): Promise<NetworkConnection<ChainTypeT>>;
}
```

Where the `networkConnectionParams` extends `NetworkConfigOverride` to add `networkName` and `chainType`.

`networkConnectionParams` should be optional so that `network.connect()` remains a valid invocation.

Our goal is to let connections to be completely optional in `networkName` and `chainType` while still providing `networkConfigOverrider` without having to weirdly pass undefined.

This will be a breaking change to the API which we should flag in the changeset.

PR #6674, opened 2025-05-06
Update `network.connect` to take a params object

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat/issues/6672


v-next/hardhat/src/types/network.ts
```diff
@@ -40,11 +40,16 @@ export type DefaultChainType = ChainTypeConfig extends {
 interface to allow the user to change the default chain type. */
 export interface ChainTypeConfig {}
 
+export type NetworkConnectionParams<
+  ChainTypeT extends ChainType | string = DefaultChainType,
+> = NetworkConfigOverride & {
+  networkName?: string;
```
alcuadrado, line 46: I think we should have two overloads of connect.
`connect(network?: string)`

and this one with object. 

A few comments about the object:
- Can we do `network` instead of network name?
- move NetworkConfigOverride to an `override` field

alcuadrado (approved):
This looks good to me. I have question though. 

Should the field be `override` or `overrides`? I'm not a native speaker, so I'd defer it to you. I think it as "configuration overrides", but it may be wrong.

PS: I just noticed that I wrote `override` before. it may have been a typo. But anyways, let's do whatever looks more natural.

zoeyTM:
@alcuadrado since the type itself was already named `NetworkConfigOverride` I think its fine to leave as-is. To completely answer your question, I think either spelling would work.

PR #6668, opened 2025-05-06
Merge user's output selection with our defaults

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat/issues/6551

PR #6678, opened 2025-05-07
Dedupe and sort merged solc outputSelection

zoeyTM:
A follow up PR to resolve [this conversation](https://github.com/NomicFoundation/hardhat/pull/6668/files/a6c7e85b770720a0f5d2ec417730e954693375a7#r2075254322)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -124,12 +124,31 @@ export class CompilationJobImplementation implements CompilationJob {
...
     };
   }
 
+  #dedupeAndSortOutputSelection(
+    outputSelection: CompilerInput["settings"]["outputSelection"],
+  ): CompilerInput["settings"]["outputSelection"] {
+    const dedupedOutputSelection: CompilerInput["settings"]["outputSelection"] =
+      {};
+
+    for (const [sourceName, contracts] of Object.entries(outputSelection)) {
```
alcuadrado, line 139: I think doing `for (const sourceName of Object.keys(..).sort())` would be safer.

Same with the other for.

alcuadrado (approved):

PR #6681, opened 2025-05-07
Enable Node 24 in v-next's CI

alcuadrado:


alcuadrado:
The `node-test-reporter` integration tests can't run with multiple versions of Node, as they depend on the behavior of it. The reason is that they are snapshot tests, and they will fail as soon as there's a version change. We should either make them support multiple snapshots, or run them only on a single version.

On top of that, we analyzed the failures a bit with @galargh, and realized that tests now are printed as failures when their subtests fail, even if the test itself passes. That behavior should be re-evaluated, as the subtest failure was already reported, so it looks weird and unnecessary.

alcuadrado:
> The `node-test-reporter` integration tests can't run with multiple versions of Node, as they depend on the behavior of it. The reason is that they are snapshot tests, and they will fail as soon as there's a version change. We should either make them support multiple snapshots, or run them only on a single version.
> 
> On top of that, we analyzed the failures a bit with @galargh, and realized that tests now are printed as failures when their subtests fail, even if the test itself passes. That behavior should be re-evaluated, as the subtest failure was already reported, so it looks weird and unnecessary.

@iosh this looks like the kind of issue that you may be interested in

kanej (approved):

Issue #6613, opened 2025-04-22
`getAllFullyQualifiedNames` is not selective enough


ItsNickBarry:
### Version of Hardhat

3.0.0-next.4

### What happened?

I'm working on migrating a project that adds a `package.json` to the `artifacts/` directory.  I don't know why they do this.  The result is that `getAllFullyQualifiedNames` includes this entry: `".:package"`.  This is not a real fully qualified name, and breaks my plugins when they try to retrieve artifacts for all fully qualified names.

Possible solutions:
1. `getAllFullyQualifiedNames` skips files in `artifacts/` root directory.  I believe Hardhat nests all artifacts in subdirectories, like `artifacts/contracts/`.
2. `getAllFullyQualifiedNames` validates the contents of the `.json` files before returning them.  Maybe could check the `_format` property.
3. Tell people to stop putting things in `artifacts/`.

### Minimal reproduction steps

Add a `.json` file to `artifacts/`.  Call `getAllFullyQualifiedNames`.  The output will contain an invalid name.

### Search terms

fully qualified names artifacts

alcuadrado:
https://github.com/NomicFoundation/hardhat/pull/6711

PR #6684, opened 2025-05-09
fix:  getAllFullyQualifiedNames to exclude invalid JSON files

Ernesto-tha-great:
This fixes issue #6613  where getAllFullyQualifiedNames is not selective enough.

### DIFF
--- a/packages/hardhat-core/src/internal/artifacts.ts
+++ b/packages/hardhat-core/src/internal/artifacts.ts
@@ -158,7 +158,20 @@ export class Artifacts implements IArtifacts {
       this._isArtifactPath(f)
     );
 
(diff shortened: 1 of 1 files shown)


Ernesto-tha-great:
@kanej 

Hi, please take a look at this. let me know if you have any feedback.

alcuadrado:
Thanks for submitting this PR!

I think this fix may be a bit slow, as it will read everything. What do you think about checking that it's a `.json` file immediately inside a `.sol` directory?

Do you have any examples of `.json` files that were problematic?

alcuadrado:
We picked another approach here: https://github.com/NomicFoundation/hardhat/pull/6711

PR #6686, opened 2025-05-11
clarify telemetry status message

ItsNickBarry:
I ran the `telemetry` task without any options, and, based on the output message, I thought it had toggled the consent status.  The word "now" in this context implies that the state is not the same as the previous state.


kanej (approved):
Good catch, thanks for the improved text @ItsNickBarry

PR #6692, opened 2025-05-14
build: upgrade to edr-0.13.0-alpha.2

Wodann:
I upgraded to `@ignored/edr-optimism` v0.13.0-alpha.2, which contains the following changes:
- Added hardfork activations for Prague
- Changed the `ChainConfig` to include a chain name and allow timestamp-based hardfork activations
- Changed the instrumentation API to require a coverage library path


schaable (approved):
LGTM

PR #6698, opened 2025-05-14
Fix sample projects and add tests to avoid them breaking again

alcuadrado:
This PR introduces a few changes:

- It adds tests that compile the sample projects, and fail if they fail to compile
- It fixes all the compilation failures (which were related to the change in the network.connect API)
- It improves the ethers template a bit, so that it looks nicer
- It fixes the dependencies in the templates. `hardhat-ignition` and `hardhat-ethers` need to be installed explicitly.

### DIFF
--- a/.changeset/bright-planes-talk.md
+++ b/.changeset/bright-planes-talk.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix sample projects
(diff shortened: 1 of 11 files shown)


.changeset/bright-planes-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix sample projects
```
kanej, line 5:
```suggestion
Fix to the template project scripts to match the new signature of `network.connect(...)` ([#6691](https://github.com/NomicFoundation/hardhat/issues/6691))
```

kanej (approved):

PR #6703, opened 2025-05-15
Fix typechain factories for abstract contracts

alcuadrado:
This PR fixes the issue that was leading to errors when running `pnpm build` in the `example-project` after compiling the contracts.

The root of the problem is that typechain generates contract factories for abstract contracts, but they don't extend `ContractFactory`. That made the `attach` fix to generate invalid code.

This PR restricts where we implement the attach fix, skipping abstract contracts.

### DIFF
--- a/v-next/hardhat-typechain/src/internal/generate-types.ts
+++ b/v-next/hardhat-typechain/src/internal/generate-types.ts
@@ -154,7 +154,7 @@ export function addJsExtensionsIfNeeded(content: string): string {
 function addSupportForAttachMethod(modifiedContent: string): string {
   // Pattern to find the contract name in factory files
   // Factories of abstract contracts don't extend ContractFactory, so we skip
-  // them by adding "extends" in the
+  // them by adding "extends"
   const pattern = /class\s+(\w+)__factory extends/;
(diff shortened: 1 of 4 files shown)


v-next/hardhat-typechain/src/internal/generate-types.ts
```diff
@@ -152,7 +152,10 @@ export function addJsExtensionsIfNeeded(content: string): string {
 //   ...
 // }
 function addSupportForAttachMethod(modifiedContent: string): string {
-  const pattern = /class\s+(\w+)__factory/; // Pattern to find the contract name in factory files
+  // Pattern to find the contract name in factory files
+  // Factories of abstract contracts don't extend ContractFactory, so we skip
+  // them by adding "extends" in the
```
ChristopherDedominici, line 157: I think this comment is missing the final part:
`them by adding "extends" in the .. (?)`

ChristopherDedominici (approved):

v-next/hardhat-typechain/src/internal/generate-types.ts
alcuadrado: Thanks!

PR #6727, opened 2025-05-22
Support Etherscan API v2

### LINKED ISSUE #6716: Support Etherscan's new API with `hardhat-verify`
For Hardhat 2 the `hardhat-verify` plugin has to be updated to use the new Etherscan API.
The API has changed to allow one API key to be used against multiple chains with the chainId passed as an additional parameter.

antico5:
Closes #6716

This PR means to support the api v2 of Etherscan.io. The main difference is that with v2 you can use only one api key (from etherscan.io), to access the explorer services on all supported chains,  in contrast with having to register and have multiple api keys and explorers.

There are three main scenarios in which the plugin is used right now, related to this change, and we intend to provide support for all of them.

1- The user specifies a single api key in the config. This is assumed to be an etherscan.io key and it will be used to verify contracts on all the supported chains, using the api v2.
2- The user specifies a map with network -> api key combination. This is now deprecated and we show a warning, but we still attempt to use the v1 api using these keys and endpoints until they discontinue support from Etherscan. We cannot use v2 in this case because api keys of different explorers (e.g. arbiscan) are not valid for etherscan.io v2 api.
3- The user specifies custom chains. In this case we keep the existing behavior, treating the api as an etherscan v1 api.

The actual implementation is simple, when we want to use the v2 api, we use the updated endpoint (which is etherscan.io for all chains) and pass a query parameter with the chain id.

### DIFF
--- a/.changeset/nasty-readers-explode.md
+++ b/.changeset/nasty-readers-explode.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/hardhat-verify": patch
 ---
 
-Support Etherscan API v2
+Support Etherscan API v2 (#6716)
--- a/.changeset/nasty-readers-explode.md
+++ b/.changeset/nasty-readers-explode.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
+Support Etherscan API v2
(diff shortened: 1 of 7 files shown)


.changeset/nasty-readers-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
+Support Etherscan API v2
```
ChristopherDedominici, line 5: I noticed that when we release a new hardhat version, we also add a link to the original issue issue. See here: https://github.com/NomicFoundation/hardhat/releases.
So maybe it can be added here too

ChristopherDedominici (approved):
I left a minor comment but apart from that LGTM

.changeset/nasty-readers-explode.md
antico5: Done, I appended the issue number to the changeset

PR #6741, opened 2025-05-23
feat: update hardhat-gas-reporter peer dependency to ^2.0.0

itasli:

---

Updates the peer dependency for `hardhat-gas-reporter` from `^1.0.8` to `^2.0.0` to support the latest versions and resolve dependency conflicts.

### DIFF
--- a/packages/hardhat-toolbox-viem/package.json
+++ b/packages/hardhat-toolbox-viem/package.json
@@ -62,13 +62,13 @@
     "eslint-plugin-mocha": "10.4.1",
     "eslint-plugin-prettier": "3.4.0",
     "hardhat": "workspace:^2.11.0",
-    "hardhat-gas-reporter": "^1.0.8",
+    "hardhat-gas-reporter": "^2.0.0",
     "mocha": "^10.0.0",
     "prettier": "2.4.1",
     "rimraf": "^3.0.2",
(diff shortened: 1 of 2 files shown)


fvictorio:
Hi @itasli, thanks for this PR, but I think this is a bit more involved than just bumping the dependency. Also, it's a significant change, so we need to do some other things around it; I opened [an issue](https://github.com/NomicFoundation/hardhat/issues/6794) with the details.

I'm tentatively closing this because I believe this is something we'll need to do ourselves. We'll try to prioritize it, because I understand this means that the toolbox and the v2 version of the gas reporter plugin can't be used together without some annoying workarounds (like telling npm to ignore the toolbox peer dependency version range).

itasli:
Hi @fvictorio 

Thanks for the detailed feedback! I totally understand your approach this kind of significant dependency update requires more careful consideration than just a simple version bump.

Before submitting the PR, I did run through the available tests to make sure there weren't any breaking changes, and everything seemed to pass without issues.

I appreciate you opening an issue to track this properly and prioritizing it.

Thanks for taking the time to explain the reasoning.

fvictorio:
Good to know in advance that there won't be any surprises when testing this manually, that should help us prioritize this!

PR #6742, opened 2025-05-24
Fix node test reporter not stripping some diffs because of coloring

### LINKED ISSUE #6688: Reporter bug: error info are duplicated
### Version of Hardhat

3.0.0-next.7

### What happened?

Issue checked in node version:
- 22.14.0
- 23.10.0

When an assertion fails, I see an error message where the same information is displayed twice, though worded differently.
This only seems to happen with more 'complex' types, numbers don't show this issue, but strings, for example, do.

## Repro steps:

I created a new `hardhat` project with `TS` and `Viem`.
And I added this test:

```

describe("Test", async function () {
  it("should fail", async function () {
    assert.equal(1, 2);

    assert.equal("CustomError1", "CustomErrorWithInt1");
  });
}); 
```

When I run:
```
assert.equal(1, 2);
```
I get a "normal" error message:
```
  Test
    1) should fail


0 passing (146ms)
1 failing

  1) Test
       should fail:

   AssertionError: Expected values to be strictly equal
   - Expected
   + Received
   
   - 2
   + 1
```
But when I run:
```
assert.equal("CustomError1", "CustomErrorWithInt1");
```
I get a weird error message with "duplicated" info:
```
  Test
    1) should fail


0 passing (241ms)
1 failing

  1) Test
       should fail:

   AssertionError: Expected values to be strictly equal:
   actual expected
   
   'CustomErrorWithInt1'
   
   - Expected
   + Received
   
   - CustomErrorWithInt1
   + CustomError1
   
       at TestContext.<anonymous> (test/test.ts:8:12)
```
The error is also "weird" for other "complex" types like objects:
```
assert.deepEqual({ a: 1 }, { a: 2 });

Throws:

  1) Test
       should fail:

   AssertionError: Expected values to be strictly deep-equal:
   + actual - expected
   
     {
   +   a: 1
   -   a: 2
     }
   
   - Expected
   + Received
   
     Object {
   -   "a": 2,
   +   "a": 1,
     }
   
       at TestContext.<anonymous> (test/test.ts:10:12)
```

antico5:
The code inside error-formating#formatSingleError is removing the diff text generated by the node:test error, so we can append the one generated by us (using jest-diff).

It was failing to find some of the diffs when color was enabled on the terminal (e.g. on my system, the tests for the reporter don't pass on v-next branch). 

This PR disables color via env. variable , and restores it just after running the tests. This env. var doesn't affect our own diff coloring (for which we use jest-diff).

Also I set the env. var on the integration tests suite, so tests now pass on my system besides just the CI

Closes #6688

image

image


galargh (commented):
Were you able to find out why it was the issue on your system? In other words, can we create a test that would have failed if we disable code colours?

Do you know if there are any cases where we might want to keep the default node colouring? Nothing in particular pops into my head straight away.

alcuadrado:
Oh, i never realized that this was because of the colors!

I wonder if we can make something similar, but in the reporter, no the test run. So that the reporter logic is self-contained.

galargh:
Nice! Let's just add `process.env.FORCE_COLOR = 1` before https://github.com/NomicFoundation/hardhat/blob/v-next/v-next/hardhat-node-test-reporter/integration-tests/index.ts#L85. That way we'll see whether the colors are correctlys tripped in CI.

galargh (approved):

Issue #6726, opened 2025-05-21
Support grep on the `npx hardhat test` task


kanej:
We now support grep on all subtasks - but not on the overarching `npx hardhat test`.

We should add it in as an option on the top level task and have it delegate down.

PR #6745, opened 2025-05-26
Support grep on the npx hardhat test task

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/6726

Usage examples:
```
// no test runner specified, no file paths specified
pnpm hardhat test --grep "Other example"

// no test runner specified, file paths specified
pnpm hardhat test --grep "Other example"  test/node/*

// test runner specified, no file paths specified
pnpm hardhat test node --grep "Other example" 
```

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
@@ -11,7 +11,7 @@ import { HardhatRuntimeEnvironmentImplementation } from "../../core/hre.js";
 interface TestActionArguments {
   testFiles: string[];
   noCompile: boolean;
-  grep: string;
+  grep: string | undefined;
 }
 
 const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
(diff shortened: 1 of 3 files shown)


v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -50,9 +51,9 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
     }
 
     if (subtask.options.has("noCompile")) {
-      await subtask.run({ testFiles: files, noCompile: true });
+      await subtask.run({ testFiles: files, noCompile: true, grep });
     } else {
-      await subtask.run({ testFiles: files });
+      await subtask.run({ testFiles: files, grep });
```
antico5, line 56: To avoid duplication and potential future bug, I'd have outside of the conditional something like `arguments = {testFiles: files, grep}` Then `if(options.has('noCompile')) {arguments['noCompile'] = true} `

antico5:
Are we ok not having tests for this?

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -11,10 +11,11 @@ import { HardhatRuntimeEnvironmentImplementation } from "../../core/hre.js";
 interface TestActionArguments {
   testFiles: string[];
   noCompile: boolean;
+  grep: string;
```
galargh, line 14: Shouldn't that be a `string` or `undefined`?

ChristopherDedominici:
> Are we ok not having tests for this?

Currently, there are no tests for the `test` task. I believe this is because testing it would require a dependency on an external test runner (e.g., Mocha or the Node test runner). One option could be to use the `solidity test` plugin, which is included with Hardhat core.
However, given the simplicity of this task, I believe the risk is acceptable. If not, we can create a follow-up task to add proper tests.

antico5 (approved):

PR #6757, opened 2025-05-27
feat: replace @sentry/node with @sentry/core

galargh:
..., vendored modules and custom implementations

---

In this PR, I replace usage of @sentry/node with @sentry/core because the former comes with a lot of dependencies that we don't need.

To accomplish that I vendored `context` and `contextlines` integrations and `module` util from @sentry/node. By vendoring these, we will be able to easily update them when needed.

I also added our own implementation of `onuncaughtexception` and `onunhandledrejection` integrations since we only need a very simple versions of these and our own implementation of the `init` function, again, for the simplicity.

Finally, I created a fetch-based transport that is used to send events/messages to sentry.

I tested this by sending events to our test Sentry project and comparing their JSONs to the ones from events sent to our production project.

An example of a JSON for an event sent from this branch - https://us.sentry.io/api/0/projects/nomic-labs/hardhat-v3-testing/events/9aef4aeae293452e9e98112e8bdf31ff/json/
An example of a similar error event sent to our production project - https://us.sentry.io/api/0/projects/nomic-labs/hardhat3/events/2dfc81ad0ed344c6b323ee7eff2be4a1/json/


antico5:
Looks good to me. Just to double check, are we ok not having the following default integrations?
-  inboundFiltersIntegration
-  consoleIntegration
-  httpIntegration
-  nativeNodeFetchIntegration
-  localVariablesIntegration
-  childProcessIntegration
-  processSessionIntegration

fvictorio:
I trusted Piotr's assessment there because I don't know what each integration does.

@galargh up to you if you want to go ahead and merge, but maybe a quick description of each of the excluded integrations and the reason not to include it would be useful.

antico5:
I just approved this PR, my comment was more aimed towards double checking the excluded default integrations before merging. I took a quick look at it and it seems we were not using them

galargh:
I'll do a quick run-through of the integrations that we don't include anymore with really simple descriptions. I agree that it will be helpful when we find ourselves missing some info in the future.

I'll post it here and merge the PR once done.

galargh:
I added the inline notes about all the default integrations in https://github.com/NomicFoundation/hardhat/pull/6757/commits/3b7feb9bd0a058488b41c538b5191f06588ed4f1

Marking this as ready for merge now.

fvictorio:
Those explanations were super useful, thanks!

PR #6760, opened 2025-05-27
Support using format on config variables

### LINKED ISSUE #6162: Support templating capabilities in configuration variables
We should consider methods of supporting patterns for Config Variables that are the equivalent of:

```js
{
  url: `https://alchemy.com/api?key=${env.API_KEY}`
}
```

A suggestion was to support templating maybe with a template library:

```ts
{
  url: config.variable("https://alchemy.com/api?key={{API_KEY}}")
}
```

Though I would consider template literals:

```ts
{
  url: config.variable`https://alchemy.com/api?key=${"API_KEY"}`
}
```

```ts
function configurationVariableResolver(name: string): string | null {
  if (name === "API_KEY") {
    return "secret"
  }

  return null
}

const config = {
  variable: (strings: TemplateStringsArray, ...keys: string[]): string => {
    return strings.reduce((result, str, i) => {
        const key = keys[i - 1];
        
        return result + configurationVariableResolver(key) + str;
    });
  },
}

const example = {
  url: config.variable`https://alchemy.com/api?key=${"API_KEY"}`
}


console.log(example)
// {
//  "url": "https://alchemy.com/api?key=secret"
// } 
```
The template literal example here would need to be changed to return a ConfigVariable object so it could be resolved later, but the principle stands.

antico5:
Closes #6162

This allows for config like the following:
```
url: configVariable(
        "ALCHEMY_KEY",
        "https://eth-sepolia.g.alchemy.com/v2/{variable}"
      ),
```

When using `get` on the resolved variable, the returned value will be `https://eth-sepolia.g.alchemy.com/v2/my_api_key`

### DIFF
--- a/v-next/hardhat-errors/src/descriptors.ts
+++ b/v-next/hardhat-errors/src/descriptors.ts
@@ -427,8 +427,8 @@ Please add the property "type" with the value "module" in your package.json to e
       CONFIG_VARIABLE_FORMAT_MUST_INCLUDE_VARIABLE: {
         number: 19,
         messageTemplate: `The format string "{format}" must include {marker} marker`,
-        websiteTitle: "Config variable format must include {marker}",
-        websiteDescription: `The config variable format must include the string {marker}, which will be replaced with the actual value of the variable.`,
+        websiteTitle: "Config variable format must include {variable}",
+        websiteDescription: `The config variable format must include the string "{variable}", which will be replaced with the actual value of the variable.`,
       },
     },
     INTERNAL: {


v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -424,6 +424,12 @@ Please add the property "type" with the value "module" in your package.json to e
         websiteTitle: "Invalid hex string",
         websiteDescription: `Given value was not a valid hex string.`,
       },
+      CONFIG_VARIABLE_FORMAT_MUST_INCLUDE_VARIABLE: {
+        number: 19,
+        messageTemplate: `The format string "{format}" must include {marker} marker`,
+        websiteTitle: "Config variable format must include {marker}",
```
schaable, line 430: `websiteTitle` and `websiteDescription` can't include variables.

PR #6767, opened 2025-05-29
test: disable reporter test that fails on macos only

galargh:

---

This PR disables a test that started failing on macos with Node 22.16.0 only - see https://github.com/NomicFoundation/hardhat/actions/runs/15307515828/job/43103972244?pr=6759

I was able to replicate the issue locally. We need to disable it because now node produces different outputs for that test on different systems and our IT running infrastructure does not currently support such a case.


v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/errors-test/test.ts
```diff
@@ -13,8 +13,10 @@ test("aggregate error in top level test", async () => {
   return Promise.any([promise1, promise2]);
 });
 
-test("error with circular cause in top level test", async () => {
-  const error = new Error("circular error");
-  error.cause = error;
-  throw error;
-});
+// NOTE: We're commenting out this test case because https://nodejs.org/en/blog/release/v22.16.0
```
kanej, line 16:
```suggestion
// TODO: We're commenting out this test case because https://nodejs.org/en/blog/release/v22.16.0
```

kanej (approved):

PR #6777, opened 2025-05-30
fix: pass buildProfile to getCompilationJobs

schaable:
The build profile was not being passed to the `getCompilerInput`, making the results invalid.


kanej (approved):
Looks good. And not like an object.

PR #6779, opened 2025-05-30
Linting for template projects

### LINKED ISSUE #6715: Add linting to Hardhat 3 Template Projects
We have added guards to check that there are no TypeScript type errors in our Hardhat 3 template projects (i.e. `v-next/hardhat/templates/*`).

We should also add basic linting and prettier formatting checks.

antico5:
Enabled linting for template projects. Prettier formatting was already enabled.

I've overriden a few rules that made sense, I'm open to feedback on these.

Closes #6715

### DIFF
--- a/v-next/hardhat/eslint.config.js
+++ b/v-next/hardhat/eslint.config.js
@@ -1,3 +1,30 @@
 import { createConfig } from "../config/eslint.config.js";
 
-export default createConfig(import.meta.filename);
+const config = createConfig(import.meta.filename);
+
+// General rules for template projects
+const templatesConfig = {
+  ...config[0], // First item is general config for hh3 codebase
+  files: ["templates/**/*.ts"],
+  ignores: ["templates/**/types/**"],
+  rules: {
+    ...config[0].rules,
+    "import/no-extraneous-dependencies": "off", // hardhat is used as a dev dependency
+    "no-restricted-syntax": "off", // top-level await
+  },
+};
+
+// Rules for test files inside template projects
+const templatesTestConfig = {
+  ...templatesConfig,
+  files: ["templates/**/test/*.ts"],
+  rules: {
+    ...templatesConfig.rules,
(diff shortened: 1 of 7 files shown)


v-next/hardhat/eslint.config.js
```diff
@@ -1,3 +1,30 @@
...
+  },
+};
+
+// Rules for test files inside template projects
+const templatesTestConfig = {
+  ...templatesConfig,
+  files: ["templates/**/test/*.ts"],
+  rules: {
+    ...templatesConfig.rules,
+    "@typescript-eslint/no-floating-promises": "off", // 'describe' and 'it' functions from node:test
```
ChristopherDedominici, line 23: Just to confirm I understand correctly: without this rule, when using the nodeJS test runner, an error will occur indicating that `it` and `describe` are floating promises?

v-next/hardhat/eslint.config.js
```diff
@@ -1,3 +1,30 @@
...
+const config = createConfig(import.meta.filename);
+
+// General rules for template projects
+const templatesConfig = {
+  ...config[0], // First item is general config for hh3 codebase
+  files: ["templates/**/*.ts"],
+  ignores: ["templates/**/types/**"],
+  rules: {
+    ...config[0].rules,
+    "import/no-extraneous-dependencies": "off", // hardhat is used as a dev dependency
```
ChristopherDedominici, line 12: Is it possible to set it `on` and allow the import of only specific packages like hardhat? Or this will be too restrictive?

v-next/hardhat/eslint.config.js
antico5: Yes, that's correct

v-next/hardhat/eslint.config.js
antico5: Probably it will be too restrictive since I think people include the plugins also on devDependencies. I think this rule may be safe to add since it's different what a dependency/devDependency means in a solidity project context vs a javascript one.

PR #6784, opened 2025-06-02
fix: add retryCount: 0 for viem devnet

TateB:

---

re-added 0 retries when on developer networks, to match hardhat-viem v2 (see [original pr](https://github.com/NomicFoundation/hardhat/pull/5708) for more detail)

### DIFF
--- a/.changeset/two-dolls-pull.md
+++ b/.changeset/two-dolls-pull.md
@@ -1,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-viem": patch
+"hardhat": patch
+---
+
+Fix to set 0 retries on development networks for `hardhat-viem`, thanks @TateB ([#6784](https://github.com/NomicFoundation/hardhat/pull/6784))


v-next/hardhat-viem/src/internal/clients.ts
```diff
@@ -154,17 +154,23 @@ export async function getTestClient<ChainTypeT extends ChainType | string>(
...
 }
 
 const DEFAULT_CLIENT_PARAMETERS = { pollingInterval: 50, cacheTime: 0 };
+const DEFAULT_TRANSPORT_PARAMETERS = { retryCount: 0 };
 
-async function getDefaultClientParameters(provider: EthereumProvider) {
-  return (await isDevelopmentNetwork(provider))
-    ? DEFAULT_CLIENT_PARAMETERS
-    : {};
+async function getDefaultParameters<TConfig extends {} | undefined>(
```
schaable: I don't think this function needs to know about the `config` or handle merging the default client parameters with it, since its name suggests it only returns the default parameters. Also, you forgot to use a ternary to decide whether to return `DEFAULT_CLIENT_PARAMETERS`, as these should only be added to dev networks.

Suggestions:

- Roll back merging the config with the default parameters.
- Use a ternary to decide whether to return the default client parameters.
- If you're up for it, I think we should rename `DEFAULT_CLIENT_PARAMETERS` and `DEFAULT_TRANSPORT_PARAMETERS` to `DEFAULT_DEVELOPMENT_CLIENT_PARAMS` and `DEFAULT_DEVELOPMENT_TRANSPORT_PARAMS` as these names make their purpose clearer.

Suggested code (remember to update the calls as well, you should merge the client parameters with the `config`): 

```typescript
const DEFAULT_DEVELOPMENT_CLIENT_PARAMS = { pollingInterval: 50, cacheTime: 0 };
const DEFAULT_DEVELOPMENT_TRANSPORT_PARAMS = { retryCount: 0 };

async function getDefaultParams(provider: EthereumProvider) {
  const isDevelopment = await isDevelopmentNetwork(provider);
  return isDevelopment
    ? {
        defaultClientParams: DEFAULT_DEVELOPMENT_CLIENT_PARAMS,
        defaultTransportParams: DEFAULT_DEVELOPMENT_TRANSPORT_PARAMS,
      }
    : {};
}
```

schaable (changes requested):

TateB:
@schaable added your suggestion + a sanity check in the tests for non-dev poll/cache time

schaable (approved):

Issue #6771, opened 2025-05-29
Viem assertion `viem.assertions.emitWithArgs` only looks at the first event


alcuadrado:
### Feedback

If you have a contract that emits the same event multiple times, `await viem.assertions.emitWithArgs(...)` won't look at all of them to see if one matches the expected arguments, but only to the first one.

That's unexpected. We should check all of them.

PR #6787, opened 2025-06-02
Viem assertion viem.assertions.emitWithArgs only looks at the first event

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/6771

### DIFF
--- a/v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
+++ b/v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
@@ -27,23 +27,23 @@ export async function emitWithArgs<
   contractFn: Promise<ReadContractReturnType | WriteContractReturnType>,
   contract: ContractReturnType<ContractName>,
   eventName: EventName,
-  args: any[],
+  expectedArgs: any[],
 ): Promise<void> {
   const abiEvents: AbiEvent[] = contract.abi.filter(
     (item): item is AbiEvent =>
       item.type === "event" &&
       item.name === eventName &&
-      item.inputs.length === args.length,
+      item.inputs.length === expectedArgs.length,
   );
 
   assert.ok(
     abiEvents.length !== 0,
-    `Event "${eventName}" with argument count ${args.length} not found in the contract ABI`,
+    `Event "${eventName}" with argument count ${expectedArgs.length} not found in the contract ABI`,
   );
 
   assert.ok(
     abiEvents.length === 1,
-    `There are multiple events named "${eventName}" that accepts ${args.length} input arguments. This scenario is currently not supported.`,
+    `There are multiple events named "${eventName}" that accepts ${expectedArgs.length} input arguments. This scenario is currently not supported.`,
   );
 
   const parsedLogs = await handleEmit(viem, contractFn, contract, eventName);
@@ -54,7 +54,7 @@ export async function emitWithArgs<
     let emittedArgs: unknown[] = [];
 
     if (logArgs === undefined) {
-      if (args.length === 0) {
+      if (expectedArgs.length === 0) {
         // If the logs contain no arguments and none are expected, we can return, this is a valid match
         return;
       }
@@ -70,7 +70,7 @@ export async function emitWithArgs<
     }
 
     if (Array.isArray(logArgs)) {
-      // All the args are listed in an array, this happens when some of the event parameters do not have parameter names.
+      // All the expected args are listed in an array, this happens when some of the event parameters do not have parameter names.
       // Example: event EventX(uint u, uint) -> mapped to -> [bigint, bigint]
       emittedArgs = logArgs;
     } else {
@@ -93,7 +93,7 @@ export async function emitWithArgs<
         if (parsedLogs.length === 1) {
           // Provide additional error details only if a single event was emitted
           assert.fail(
-            `The provided event "${eventName}" expects ${args.length} arguments, but the emitted event contains ${Object.keys(logArgs).length}.`,
+            `The provided event "${eventName}" expects ${expectedArgs.length} arguments, but the emitted event contains ${Object.keys(logArgs).length}.`,
           );
         }
 
@@ -101,7 +101,7 @@ export async function emitWithArgs<
       }
     }
 
-    if ((await deepEqual(emittedArgs, args)) === true) {
+    if ((await deepEqual(emittedArgs, expectedArgs)) === true) {
       return;
     }
 
@@ -109,7 +109,7 @@ export async function emitWithArgs<
       // Provide additional error details only if a single event was emitted
       assert.deepEqual(
         emittedArgs,
-        args,
+        expectedArgs,
         "The event arguments do not match the expected ones.",
       );
     }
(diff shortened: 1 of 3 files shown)


v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
```diff
@@ -47,20 +48,31 @@ export async function emitWithArgs<
...
+    let emittedArgs: unknown[] = [];
 
-    if (Array.isArray(parsedLog)) {
+    if (logArgs === undefined) {
+      if (args.length === 0) {
+        // If the logs contain no arguments and none are expected, we can return, this is a valid match
+        return;
+      }
+
+      if (parsedLogs.length === 1) {
```
ChristopherDedominici: Here, and also later in the file, I provide more details about the error when only a single error is emitted. If there are too many events, it can be confusing to include too much information, such as all the arguments emitted by each event. So, I decided to set a cutoff: detailed information is only provided when there's a single event

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
```diff
@@ -47,20 +48,31 @@ export async function emitWithArgs<
 
   const parsedLogs = await handleEmit(viem, contractFn, contract, eventName);
 
-  let emittedArgs: any[] = [];
-  if (args.length > 0) {
```
antico5, line 51: Should we rename `args` to `expectedArgs`? otherwise it's difficult to track together with `emittedArgs` and `logArgs`

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
antico5: Can we skip the length==1 check here? so just fail if there is at least 1 emitted event with argument count == 0 but expected argument count != 0

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
```diff
@@ -72,22 +84,38 @@ export async function emitWithArgs<
           `The event parameter at index ${index} does not have a name`,
         );
 
-        emittedArgs.push(parsedLog[param.name]);
+        emittedArgs.push(logArgs[param.name]);
 
         parsedLogCount++;
```
antico5: Can we remove the `parsedLogCount` variable and instead just use `emittedArgs.length`?

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
```diff
@@ -72,22 +84,38 @@ export async function emitWithArgs<
           `The event parameter at index ${index} does not have a name`,
         );
 
-        emittedArgs.push(parsedLog[param.name]);
+        emittedArgs.push(logArgs[param.name]);
```
antico5, line 79: Should we check that `param.name` is a key of `logArgs`, or that would be redundant ?

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
```diff
@@ -72,22 +84,38 @@ export async function emitWithArgs<
...
 
-      if (parsedLogCount !== Object.keys(parsedLog).length) {
-        assert.fail(
-          `The provided event "${eventName}" expects ${args.length} arguments, but the emitted event contains ${Object.keys(parsedLog).length}.`,
-        );
+      if (parsedLogCount !== Object.keys(logArgs).length) {
+        if (parsedLogs.length === 1) {
+          // Provide additional error details only if a single event was emitted
+          assert.fail(
+            `The provided event "${eventName}" expects ${args.length} arguments, but the emitted event contains ${Object.keys(logArgs).length}.`,
```
antico5: Is this the correct place to check that an emitted event's arg count mismatches its signature? I'm not familiar with viem or its associated plugins but it makes some noise to me, since this code is unrelated to the assertion itself.
And if we keep this, then similarly to a previous comment, I don't understand the value on checking that length == 1

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
```diff
@@ -72,22 +84,38 @@ export async function emitWithArgs<
...
+
+        continue;
       }
     }
+
+    if ((await deepEqual(emittedArgs, args)) === true) {
+      return;
+    }
+
+    if (parsedLogs.length === 1) {
```
antico5, line 87: In this case I think it's ok for checking length == 1

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
ChristopherDedominici: No I'd say it's redundant. If it's not a key, it is because the event currently analyzed in the `logArgs` is not the expected one, and it will fail later in the `deepEqual`

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
ChristopherDedominici: Good point, it can be moved out.

The idea behind the `=== 1` check is that users will likely write tests with only one event emitted per test. In those cases, I think it's helpful to include additional information in the error message. However, if there are, say, 10 events, the error becomes too long and noisy. So, I only show the extra info when a single event is emitted. That said, I'm open to suggestions if you think it's useless or it can be improved

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6787/commits/4a2e0d5d0c4cac51d835a31709a0ee1f4bca02f6

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6787/commits/d0605606c622b1505921d8a1160fd2c2f2b9cb3e

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
ChristopherDedominici: Done

v-next/hardhat-viem-assertions/src/internal/assertions/emit/emit-with-args.ts
ChristopherDedominici: Can be removed, done here: https://github.com/NomicFoundation/hardhat/pull/6787/commits/8bd1227926e6075ae487317355b4adf3274af095

antico5 (approved):

PR #6788, opened 2025-06-02
fix: use production build profile by default

schaable:
Set "production" as the default build profile if `--build-profile` isn't specified.

### DIFF
--- a/v-next/hardhat-verify/src/internal/verification.ts
+++ b/v-next/hardhat-verify/src/internal/verification.ts
@@ -66,7 +66,7 @@ export async function verifyContract(
   const {
     artifacts,
     config,
-    globalOptions: { buildProfile: buildProfileName },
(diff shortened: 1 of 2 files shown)


v-next/hardhat-verify/src/internal/verification.ts
```diff
@@ -88,7 +88,9 @@ export async function verifyContract(
     // provider
   } = verifyContractArgs;
 
-  const buildProfile = config.solidity.profiles[buildProfileName ?? "default"];
+  const buildProfile = config.solidity.profiles[buildProfileName];
+  // The "production" build profile is always present by default.
+  // This check only fails if the user specifies a non-existent build profile name.
```
kanej, line 93: Literally the only question I had when thinking about this.

kanej (approved):
LGTM!

PR #6795, opened 2025-06-04
[hardhat-etehrs] Add implementation for `authorize` and `populateAuthorization`

ChristopherDedominici:


v-next/hardhat-ethers/src/internal/signers/signers.ts
```diff
@@ -47,43 +66,74 @@ export class HardhatEthersSigner implements HardhatEthersSignerI {
...
+
+      this.#privateKey = matchingKey;
+    }
+
+    if (this.#privateKey === undefined) {
+      // eslint-disable-next-line no-restricted-syntax -- TODO
+      throw new Error(`No private key found for address ${this.address}`);
+    }
+
+    const wallet = new Wallet(this.#privateKey, this.provider);
```
ChristopherDedominici: do we want to cache the wallet too?

v-next/hardhat-ethers/src/internal/signers/derive-private-key.ts
```diff
@@ -0,0 +1,85 @@
...
+  return derivePrivateKeysImpl(
+    mnemonic,
+    accounts.path,
+    accounts.initialIndex,
+    accounts.count,
+    passphrase,
+  );
+}
+
+async function derivePrivateKeysImpl(
```
ChristopherDedominici, line 26: I copied and pasted the same function we have in hardhat core. The error messages seem general enough to be reused here, so we might want to consider exposing this function directly from the core

v-next/hardhat-ethers/src/internal/signers/signers.ts
```diff
@@ -47,43 +66,74 @@ export class HardhatEthersSigner implements HardhatEthersSignerI {
...
+  public async populateAuthorization(
+    authReq: AuthorizationRequest,
   ): Promise<AuthorizationRequest> {
-    throw new HardhatError(
-      HardhatError.ERRORS.HARDHAT_ETHERS.GENERAL.METHOD_NOT_IMPLEMENTED,
-      {
-        method: "HardhatEthersSigner.populateAuthorization",
-      },
-    );
+    const auth = { ...authReq };
```
ChristopherDedominici: Do we want to use the method exposed by the `wallet` here as well? At the moment, I’ve copied the same logic used in ethers, with a few minor changes that utilize our own method

v-next/hardhat-ethers/package.json
```diff
@@ -59,6 +59,7 @@
     "@nomicfoundation/hardhat-errors": "workspace:^3.0.0-next.11",
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.0-next.11",
     "debug": "^4.3.2",
+    "ethereum-cryptography": "^2.2.1",
```
ChristopherDedominici, line 62: This dependency is required if we decide to duplicate the function `derivePrivateKeys`

fvictorio (approved):

PR #6797, opened 2025-06-04
[hardhat-ethers] Add implementation for authorize and populateAuthorization

ChristopherDedominici:


packages/hardhat-ethers/src/signers.ts
```diff
@@ -213,6 +260,34 @@ export class HardhatEthersSigner implements ethers.Signer {
...
+        return this.accounts as string[];
+      }
+
+      return (this.accounts as HardhatNetworkAccountConfig[]).map(
+        (acc) => acc.privateKey
+      );
+    }
+
+    if ("mnemonic" in this.accounts) {
+      return derivePrivateKeys(
```
ChristopherDedominici, line 288: The error messages in `derivePrivateKeys` seem general enough, so the function can be reused here without duplicating it, wdyt?

packages/hardhat-ethers/src/signers.ts
```diff
@@ -67,32 +80,66 @@ export class HardhatEthersSigner implements ethers.Signer {
...
+      );
+    }
+
+    if (this.privateKey === undefined) {
+      throw new HardhatEthersError(
+        `No private key found for address ${this.address}`
+      );
+    }
+
+    const wallet = new Wallet(this.privateKey, this.provider);
```
ChristopherDedominici: do we want to cache the wallet too?

packages/hardhat-ethers/src/signers.ts
```diff
@@ -67,32 +80,66 @@ export class HardhatEthersSigner implements ethers.Signer {
...
+
+    return wallet.authorize(auth);
   }
 
-  public populateAuthorization(
+  public async populateAuthorization(
     _auth: AuthorizationRequest
   ): Promise<AuthorizationRequest> {
-    throw new NotImplementedError("HardhatEthersSigner.populateAuthorization");
+    const auth = { ..._auth };
```
ChristopherDedominici, line 125: Do we want to use the method exposed by the wallet here as well? At the moment, I’ve copied the same logic used in ethers, with a few minor changes that utilize our own method

packages/hardhat-ethers/src/signers.ts
fvictorio: No, I don't think that's necessary.

packages/hardhat-ethers/src/signers.ts
fvictorio: Yes, I think that's a fine compromise.

packages/hardhat-ethers/src/signers.ts
fvictorio: Let's go with your approach.

Delegating to the wallet would mean that this method wouldn't work for JSON-RPC signers, but there's no reason to add that limitation given that this method doesn't sign anything. (Admittedly, it doesn't make sense to populate an authorization if you are not going to sign it later, but still.)

packages/hardhat-ethers/test/environment.ts
```diff
@@ -85,10 +92,13 @@ export function usePersistentEnvironment(
...
  */
-export function useGeneratedEnvironment(
+export function useGeneratedEnvironment<N extends "hardhat" | "localhost">(
   hardhatGasLimit: "default" | "auto" | number,
   localhostGasLimit: "default" | "auto" | number,
-  networkName: "hardhat" | "localhost"
+  networkName: N,
+  accounts?: N extends "hardhat"
+    ? HardhatNetworkAccountsUserConfig
+    : HttpNetworkAccountsUserConfig
```
fvictorio, line 99: Nice

fvictorio (approved):

PR #6798, opened 2025-06-04
feat: prepare regression tests for hardhat2 commands

galargh:

---

This PR prepares the regression tests framework for running hardhat2 commands. It also configured OpenZeppelin contracts as the first repository in which both hardhat 2 and hardhat 3 commands can be run.

After the changes, each repository configuration explicitly states which commands can be run in it. If one tries to run an unsupported command in a repository, the workflow will fail.

Here's an example run of `hardhat compile` and `hardhat2 compile` commands on Linux, Windows and macOS in `OpenZeppelin/openzeppelin-contracts` repository: https://github.com/NomicFoundation/hardhat/actions/runs/15401835601

Follow-up (not blocking; to be addressed when possible): make the regression tests workflow reusable (this will allow using it in EDR, as well as creating more specialised versions of the workflows with different defaults in hardhat, e.g. one for solidity tests and another for v2 vs c3 comparison).


kanej:
To run the workflow testing Hardhat 2 repos vs Hardhat 3, I would do the following:

```
repositories:
["OpenZeppelin/openzeppelin-contracts"]
runners:
["ubuntu-latest", "windows-latest", "macos-latest"]
commands:
["hardhat compile", "hardhat2 compile"]
```

galargh:
> To run the workflow testing Hardhat 2 repos vs Hardhat 3, I would do the following:
> 
> ```
> repositories:
> ["OpenZeppelin/openzeppelin-contracts"]
> runners:
> ["ubuntu-latest", "windows-latest", "macos-latest"]
> commands:
> ["hardhat compile", "hardhat2 compile"]
> ```

This is correct! We also have to remember to set the correct value for the `Use workflow from` input. It controls where the worfklow definition (`.github/workflows/regression-tests.yml`) and the configuration (`.github/config/regression-tests.yml`) are taken from. Until we merge this, it should be set to `hardhat2-regression-tests`. After, we will be able to use `v-next`.

This should not be confused with `hardhat-ref` input which controls what ref Hardhat 3 used for testing is build from (or downloaded in case the provided value starts with `next`).

kanej (approved):

Issue #6792, opened 2025-06-03
Node 24 Support


kanej:
Allow the Alpha to be run under Node 24. Users are likely to be running on the latest and greatest when interacting with the Hardhat 3 Alpha, so we should provide Node 24 support.

See this initial attempt for more context: https://github.com/NomicFoundation/hardhat/pull/6681

# TODO

- [x] keep node test reporter tests on 22
- [x] fix the snapshot strategy and the output for 24

PR #6804, opened 2025-06-05
Updated node test reporter with node v24 support

galargh:
Resolves https://github.com/NomicFoundation/hardhat/issues/6792

---

In this PR, I added expected result file versioning with the node major version to the selected note test reporter integration tests. After these changes are introduced, the `nested-test` fixture will have separate expected result snapshots for node v22 and v24. This will enable us to continue testing the reporter with both versions in CI.

For now, the versioning of result files for the `nested-test` is hardcoded. If we find more tests that require this in the future, we should either allow the tests to self-identify as such or simply version all the result files.

The other thing from https://github.com/NomicFoundation/hardhat/pull/6681#issuecomment-2859224923 that this PR was supposed to address was how the subtest failures are reported, but I'm not sure we should address this.

This is how they are being displayed in node v22 - https://raw.githubusercontent.com/NomicFoundation/hardhat/4093d9c10b59b8086824330900b25d0616369ef1/v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/nested-test/result.v22.svg

And this is how they are being displayed in node v24 - https://raw.githubusercontent.com/NomicFoundation/hardhat/4093d9c10b59b8086824330900b25d0616369ef1/v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/nested-test/result.v24.svg

Apart from fewer stack trace frames and the lack of a cancelled unawaited test, the outputs are very similar.
The question is whether we should report failure details 5), 6) and 7) at all. I'd argue yes because those failures are for failing `it`s rather than `describe`s (see https://github.com/NomicFoundation/hardhat/blob/4093d9c10b59b8086824330900b25d0616369ef1/v-next/hardhat-node-test-reporter/integration-tests/fixture-tests/nested-test/test.ts#L21) and since `it` represents a test case, we should have an indication of it failing when one of its "subtests" fails. The reason why we report on these failures after reporting on the subtests is that we display the results as soon as we get them.


v-next/hardhat-node-test-reporter/integration-tests/README.md
```diff
@@ -2,7 +2,7 @@
 
 This folder contains integration tests for the reporter. They don't use `node:test` as driver of the test runs, as you can't run `node:test` within `node:test`.
 
-Instead, the script `index.ts` runs all the tests in each fixture folder, comparing the reporter's results with `result.txt`.
+Instead, the script `index.ts` runs all the tests in each fixture folder, comparing the reporter's results with `result.$NODE_MAJOR_VERSION.txt`.
```
galargh:
```suggestion
Instead, the script `index.ts` runs all the tests in each fixture folder, comparing the reporter's results with `result.$NODE_MAJOR_VERSION.txt` (or `result.txt` as a fallback if a node-specific result file does not exist).
```

kanej (approved):
This looks good for supporting multiple node versions in our node test reporter suite.

I think we should separate out the question of reporting of subtests.

galargh:
> I think we should separate out the question of reporting of subtests.

Yes, totally! I just wanted to point out that it doesn't look like a regression between v22 and v24 after all.

PR #6825, opened 2025-06-07
Revert "chore(regression-tests): unignore testBumpSlot"

galargh:
Reverts NomicFoundation/hardhat#6800

It turns out that the test is just flaky. See this run for example: https://github.com/NomicFoundation/hardhat/actions/runs/15506491633

It passed on macos but failed on linux and windows.

### DIFF
--- a/.github/config/regression-tests.yml
+++ b/.github/config/regression-tests.yml
@@ -439,6 +439,10 @@ repositories:
       test/WebAuthn.t.sol
 
       # https://github.com/NomicFoundation/hardhat/issues/6509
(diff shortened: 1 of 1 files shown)


.github/config/regression-tests.yml
```diff
@@ -438,6 +438,10 @@ repositories:
       test/P256.t.sol
       test/WebAuthn.t.sol
 
+      # https://github.com/NomicFoundation/hardhat/issues/6509
```
galargh:
```suggestion
      # This test is flaky - no action required 
```

kanej (approved):

PR #6837, opened 2025-06-10
Use tilde version for solhint

fvictorio:
Solhint released a breaking change in v5.1.0 version. Adapting to that breaking change is not that hard, but just using `~5.0.0` as the dependency range is easier.


kanej (approved):
Looks good.

PR #6838, opened 2025-06-10
CI failing: fix console task logic that triggers asynchronous activity after test execution

ChristopherDedominici:


v-next/hardhat/src/internal/builtin-plugins/console/task-action.ts
```diff
@@ -51,6 +51,9 @@ const consoleAction: NewTaskActionFunction<ConsoleActionArguments> = async (
 
       // Resolve the task action promise only when the REPL server exits
       replServer.on("exit", () => {
+        // Suppress the REPL’s final prompt to avoid a "use after close" error.
+        // Prevent from queueing one last async prompt after `.exit`.
+        replServer.displayPrompt = () => {};
```
galargh: From what I gather from the title this only happens in tests, right? And it's not a persistent issue either as these tests on `v-next` do pass on `v-next` branch right now.

I'd think that any fix for that would be better applied inside the tests rather than here. Have you tried that? Also, what kind of "use after close" is happening in the test that was failing for you?

Finally, to verify a fix like this, we would likely want to run the flaky test a lot of times in the CI and seeing how it behaves. We can do that by modifying the CI workflow on some test branch.

ChristopherDedominici: This is a persistent issue in Node.js v24.2.0. I can reproduce it locally in every execution, it's not a flaky test.
If you check the [first commit](https://github.com/NomicFoundation/hardhat/pull/6838/commits/289ad746f17c32504f9766fa9a99f61b80105b8b) of this PR, you'll see that the CI failed even with the simple addition of a comment.

It's strange that the CI didn't fail earlier when updating to Node 24, could a different version have been used?

galargh: Oh, that's great news! It'll make testing the fix so much easier. 

Yes, as you can see in https://github.com/NomicFoundation/hardhat/actions/runs/15533095582/job/43726184029 that we were using `v24.1.0` there.

I still think it'll be good to try to understand what's the activity happening after close and, possibly, applying the fix in tests (unless, given the activity, it makes sense to surpress it in the task).

kanej: I get failing tests for `./v-next/hardhat` with Node `v24.2.0` across three runs. The number of fails varies. But representative examples include:

```sh
1) SolidityBuildSystemImplementation
       emitArtifacts:

   Error: test did not finish before its parent and was cancelled
```

```sh
20) analytics
        running in an interactive environment (simulated with ENV variables)
          should create the correct payload for the task analytics:

   Error: Subprocess was not executed in the expected time
```

ChristopherDedominici: We can move `replServer.displayPrompt = () => {};` directly into the test file.
What I’m having trouble understanding is whether the error is caused by the `PassThrough` assigned to the `input` and `output` in the test or by our actual implementation of the task action, which might be executing code after the resolve method.

Both these solutions fix the issue though. Wdyt?

galargh: Since we don't know the root cause, I'd bet on the safe side and apply the workaround on the test side only. This would ensure we don't hide any important information from the users. 

And if it turns out it is just noise, then hopefully we'll get a report about it at some point or, even better, we'll be able to invest some time into the investigation and discover it ourselves at some point.

ChristopherDedominici: Make sense, I did this change, the PR is ready for review @galargh

galargh (approved):
I dug a little deeper and I think I found an alternative solution that seems less intrusive (see https://github.com/NomicFoundation/hardhat/pull/6846) but it currently fails with seemingly unrelated errors.

I'd say, let's merge this to get v-next back to green, and we can follow up with any alternatives later.

kanej (approved):
LGTM!

PR #6840, opened 2025-06-10
fix: check the --no-commit flag is supported on installed forge version

### LINKED ISSUE #6824: Plugin hardhat-foundry run forge install failed
### Version of Hardhat

@nomicfoundation/hardhat-foundry 1.1.3

### What happened?

Run `npx hardhat init-foundry` got error `Command failed: forge install --no-commit foundry-rs/forge-std
error: unexpected argument '--no-commit' found`

Base on [forge reference](https://getfoundry.sh/forge/reference/forge-install#vcs-options), need remove the option `--no-commit`

Follow the doc [Adding Foundry to a Hardhat project](https://hardhat.org/hardhat-runner/docs/advanced/hardhat-and-foundry#adding-foundry-to-a-hardhat-project)

### Search terms

_No response_

antico5:
Closes #6824 

Up to version 1.0.0 of foundry, the `--no-commit` flag is mandatory on our usage of `forge install`. From `1.1.0` onwards, it's mandatory to not use it. The simplest way to detect this for me is to just run `forge install --help` and see if the `--no-commit` flag is there.

### DIFF
--- a/.changeset/fruity-brooms-stare.md
+++ b/.changeset/fruity-brooms-stare.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-foundry": patch
+---
+
+When running forge install, check the --no-commit flag is supported on installed forge version
--- a/.changeset/fruity-brooms-stare.md
+++ b/.changeset/fruity-brooms-stare.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/hardhat-foundry": patch
 ---
 
-When running forge install, check the --no-commit flag is supported on installed forge version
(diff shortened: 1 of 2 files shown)


.changeset/fruity-brooms-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-foundry": patch
+---
+
+When running forge install, check the --no-commit flag is supported on installed forge version
```
fvictorio: Changesets don't need to include implementation details. They should only tell the user what changed.

```suggestion
Fixed a bug in the `init-foundry` task that prevented it from being used with the latest versions of `forge`
```

PR #6842, opened 2025-06-10
Clear account code on forking

### LINKED ISSUE #6834: In Hardhat 3 support forking tests by undelegating test accounts
Pectra and EIP-7702 allowed EOAs to delegate to contracts. In Mainnet this has been done on all the Hardhat test accounts as the private keys are known.
In Hardhat 3, HH is responsible for setting initial state and passing it to EDR. It is therefore on the HH side to implement undelegation of the test accounts.

antico5:
Closes #6834 

Set empty code on EDR account genesis state


v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -421,6 +421,7 @@ async function getProviderConfig(
       const accountOverride: AccountOverride = {
         address,
         balance: BigInt(balance),
+        code: new Uint8Array(), // Empty account code, removing potential delegation code when forking
```
frangio, line 424: For Hardhat 2 we've only done this for ["leaked" accounts](https://github.com/NomicFoundation/edr/commit/65e8d2588077ea1a6d86676221de7991b7108c07#diff-03b2ca6fd82ba7931a70b3b1ef672aa25b18c86bf643d21a930efa2691e701cfR93) (so basically the default accounts). That way if the account is a private account of the user's we don't unexpectedly remove a delegation.

antico5: Makes sense, I'll add that filter

antico5: Added it but just rolled it back because of the final call about this

.changeset/green-tools-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Clear account code on forking
```
fvictorio, line 5: Fixed a problem related to configured accounts during forking. Some of them may be delegated in mainnet, making them unusable for local development. They are now undelegated after forking to turn them into normal EOAs.

fvictorio (approved):

.changeset/green-tools-trade.md
fvictorio:
```suggestion
Fixed a problem related to configured accounts during forking. Some of them may be delegated in mainnet, making them unusable for local development. They are now undelegated after forking to turn them into normal EOAs.
```

fvictorio: Ah, the changeset we used in HH2 is better than my propoed change:

> Fixed a bug when sending ETH to the testing accounts in forked networks. Now testing accounts are automatically undelegated.

.changeset/green-tools-trade.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed a problem related to configured accounts during forking. Some of them may be delegated in mainnet, making them unusable for local development. They are now undelegated after forking to turn them into normal EOAs.
```
kanej:
```suggestion
Fixed a bug when sending ETH to the testing accounts in forked networks. Now testing accounts are automatically undelegated ([#6834](https://github.com/NomicFoundation/hardhat/issues/6834))
```

Issue #6791, opened 2025-06-03
Bug: Native token transfer with call{value:} to test accounts on Hardhat fork does not increase recipient balance


maqicheng123:
### Version of Hardhat

2.24.0

### What happened?

When using Hardhat's mainnet fork feature, transferring native tokens (ETH/BNB) to a test account (the default accounts provided by Hardhat) using call{value:} does not increase the recipient's balance. The transfer appears to be "ineffective"—the recipient's balance does not increase, even though the transaction succeeds and gas is consumed. This is inconsistent with mainnet/testnet, where the recipient's balance should increase by the transferred amount.

### Minimal reproduction steps

1.Start Hardhat in fork mode (e.g. fork BSC mainnet).
2.Deploy the following contract:
    contract Test {
        function test() public payable {
            (bool sent, ) = payable(msg.sender).call{value: msg.value}("");
            require(sent, "Failed to send Ether");
        }
    }
3.Use a Hardhat test account (e.g. owner or user1) to call test() with 500 BNB.
4.Check the account's balance before and after the call.

5.Example Output
// test is hardhat test. user1 is random account
test balance1 10000000000000000000000n
test balance2 9499999950918999950919n
user1 balance1 1000000000000000000000n
user1 balance2 999999963382999963383n

### Search terms

test account balance bug

kanej:
Hey @maqicheng123, thanks for reporting.
Someone has setup [EIP-7702 delegations](https://etherscan.io/address/0xbe862ad9abfe6f22bcb087716c7d89a26051f74c#authlist7702) for the Hardhat testing accounts (as the private keys are public). When you fork, you are getting that delegation which is draining the transferred tokens.

We are working on a solution for forking where we will undelegate from the testing accounts in the local Hardhat node.

nebasuke:
@maqicheng123 
We are looking to address this. For now, you can get around this by using a specific block number before the delegations were set up. See https://github.com/NomicFoundation/edr/pull/922/files how to do this in the provider.

maqicheng123:
> Hey [@maqicheng123](https://github.com/maqicheng123), thanks for reporting. Someone has setup [EIP-7702 delegations](https://etherscan.io/address/0xbe862ad9abfe6f22bcb087716c7d89a26051f74c#authlist7702) for the Hardhat testing accounts (as the private keys are public). When you fork, you are getting that delegation which is draining the transferred tokens.
> 
[...]

PR #6845, opened 2025-06-11
build: upgrade EDR to v0.11.1

frangio:
Upgraded EDR to [v0.11.1](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.11.1):

- Fixed bug where native token sent to test account does not increase recipient balance.

Fixes https://github.com/NomicFoundation/hardhat/issues/6791

### DIFF
--- a/.changeset/chubby-flies-shine.md
+++ b/.changeset/chubby-flies-shine.md
@@ -1,0 +1,7 @@
+---
+"hardhat": patch
+---
+
+Upgraded EDR to [v0.11.1](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.11.1):
+
+- Fixed bug where native token sent to test account does not increase recipient balance.
--- a/.changeset/chubby-flies-shine.md
+++ b/.changeset/chubby-flies-shine.md
@@ -2,6 +2,4 @@
 "hardhat": patch
 ---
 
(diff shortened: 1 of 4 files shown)


.changeset/chubby-flies-shine.md
```diff
@@ -0,0 +1,7 @@
+---
+"hardhat": patch
+---
+
+Upgraded EDR to [v0.11.1](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.11.1):
+
+- Fixed bug where native token sent to test account does not increase recipient balance.
```
alcuadrado: This is surprising: apparently changesets doesn't work with `- `, as it uses them to create a changelog.

Also, should we explain the bug a little? What about this?

```suggestion
Upgraded EDR to [v0.11.1](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.11.1), which fixed a bug when sending ETH to the testing accounts in forked networks. Now testing accounts are automatically undelegated.
```

frangio: I essentially copied the format from [#6590](https://github.com/NomicFoundation/hardhat/pull/6590/files#diff-54eef00cd364f62dc648eb7c88e301b73f79b2443787bb24d3de34a32d064b91) but it looks like I added a newline and maybe that doesn't work.

Yes I think that is a much better description.

PR #6846, opened 2025-06-12
test: force console tests to operate in tty mode

galargh:
I browsed through the node changelog and I found an even less intrusive fix for our "use after close" issue. In fact, this could probably be considered a full fix unless we expect users to use the console tasks in a nonTTY environments.


galargh:
The fix for the failing WASM compilation is inspired by the conversation from https://github.com/nodejs/node/issues/58515 (and more specifically a similar fix that was applied here https://github.com/nodejs/node/pull/58598#issuecomment-2956151936 + https://github.com/nodejs/node/pull/58598/commits/4d6f69300c04d6fb5e20ed78cb8259f336737f9f).

I'm still not sure why we weren't hit by this earlier.

kanej (approved):
Looks good. Thanks for documenting our issues here in the code comment.

PR #6847, opened 2025-06-12
Communication via socket for test child processes' user interruptions

antico5:
This is a WIP/exploration PR. Some feedback/polishing may be needed.

This addresses the issue where node test child processes can't deal with user input, which e.g. prevents users from typing the keystore master password.

This proposed solution uses a file socket to communicate between child test processes and the main process runner, so when they need user input (`requestSecretInput` hook for now), it gets forwarded to the main process and the response is sent back. 

This implementation is somewhat naive but has room for extension if it's approved. For now I added caching based on interrupter plugin and prompt, so the master password isn't requested once for each child process.

### DIFF
--- a/v-next/hardhat-node-test-runner/src/protocol.ts
+++ b/v-next/hardhat-node-test-runner/src/protocol.ts
@@ -1,0 +1,148 @@
+import type { HardhatRuntimeEnvironment } from "hardhat/types/hre";
+
+import fs from "node:fs";
+import net from "node:net";
+import * as os from "node:os";
+
+import { assertHardhatInvariant } from "@nomicfoundation/hardhat-errors";
+
+function getSocketPath(): string {
+  if (os.platform() === "win32") {
+    return "\\\\.\\pipe\\hardhat_node_test";
+  } else {
+    return "/tmp/hardhat_node_test.sock";
+  }
+}
+
+export function isChildTestProcess(): boolean {
+  return process.env.NODE_TEST_CONTEXT !== undefined;
+}
+
+export class TestProtocolClient {
+  readonly #socketPath: string;
+
+  constructor() {
+    this.#socketPath = getSocketPath();
+  }
+
+  public async requestSecretInput(
+    interruptor: string,
+    inputDescription: string,
+  ): Promise<string> {
+    const response = await this.#request({
+      type: "requestSecretInput",
+      params: { interruptor, inputDescription },
+    });
+
+    assertHardhatInvariant(
+      response.type === "secretInputResponse",
+      "Expected response to be a secretInputResponse",
+    );
+
+    const { secretInput } = response.params;
+
+    return secretInput;
+  }
+
+  async #request(msg: Message): Promise<Message> {
+    return new Promise((resolve) => {
+      const socket = net.createConnection(this.#socketPath, () => {
+        socket.write(JSON.stringify(msg));
+      });
+
+      socket.on("data", (data) => {
+        const message = JSON.parse(data.toString());
+        socket.end();
+        resolve(message);
+      });
+    });
+  }
+}
+
+export class TestProtocolServer {
+  readonly #socketPath: string;
+  #server?: net.Server;
+  readonly #secretInputCache: Map<string, string> = new Map();
+  readonly #hre: HardhatRuntimeEnvironment;
+
+  constructor(hre: HardhatRuntimeEnvironment) {
+    this.#hre = hre;
+    this.#socketPath = getSocketPath();
+  }
+
+  public start(): void {
+    if (process.platform !== "win32" && fs.existsSync(this.#socketPath)) {
+      fs.unlinkSync(this.#socketPath); // remove stale socket
+    }
+
+    this.#server = net.createServer((socket) => {
+      socket.on("data", async (data) => {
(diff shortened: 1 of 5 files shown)


v-next/hardhat-node-test-runner/src/protocol.ts
```diff
@@ -0,0 +1,148 @@
...
+  readonly #secretInputCache: Map<string, string> = new Map();
+  readonly #hre: HardhatRuntimeEnvironment;
+
+  constructor(hre: HardhatRuntimeEnvironment) {
+    this.#hre = hre;
+    this.#socketPath = getSocketPath();
+  }
+
+  public start(): void {
+    if (process.platform !== "win32" && fs.existsSync(this.#socketPath)) {
```
galargh, line 74: Why don't we do the unlink on Windows?

v-next/hardhat-node-test-runner/src/protocol.ts
```diff
@@ -0,0 +1,148 @@
...
+      fs.unlinkSync(this.#socketPath); // remove stale socket
+    }
+
+    this.#server = net.createServer((socket) => {
+      socket.on("data", async (data) => {
+        await this.#handleData(socket, data);
+      });
+    });
+
+    this.#server.listen(this.#socketPath);
```
galargh, line 84: AI suggests restricting the permissions on the created socket file.

```suggestion
    this.#server.listen(this.#socketPath, () => {
        fs.chmodSync(this.#socketPath, 0o600);
    });
```

galargh (commented):
This seems reasonable. However, I think we should make it absolutely clear that this is how secrets are being handled. At least in the documentation, but maybe also in the help section of the task?

I want to think about how we could test it a little more, and if there are any other ways we should secure it.

v-next/hardhat-node-test-runner/src/protocol.ts
```diff
@@ -0,0 +1,148 @@
+import type { HardhatRuntimeEnvironment } from "hardhat/types/hre";
+
+import fs from "node:fs";
+import net from "node:net";
+import * as os from "node:os";
+
+import { assertHardhatInvariant } from "@nomicfoundation/hardhat-errors";
+
+function getSocketPath(): string {
```
alcuadrado, line 9: I think we need to generate a unique name, and pass it to the child processes somehow. Otherwise you wouldn't be able to run multiple instances of HH

v-next/hardhat-node-test-runner/src/protocol.ts
```diff
@@ -0,0 +1,148 @@
...
+  if (os.platform() === "win32") {
+    return "\\\\.\\pipe\\hardhat_node_test";
+  } else {
+    return "/tmp/hardhat_node_test.sock";
```
alcuadrado, line 13: Same about the random. We may want to place this under the cache, or use mktmp from the ustils.

v-next/hardhat-node-test-runner/src/protocol.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (message.type === "requestSecretInput") {
+      await this.#handleRequestSecretInput(socket, message);
+    }
+  }
+
+  async #handleRequestSecretInput(
+    socket: net.Socket,
+    message: RequestSecretInputMessage,
+  ): Promise<void> {
+    const { interruptor, inputDescription } = message.params;
```
alcuadrado, line 103: iI wouldn't add any cache here.

I think we should also override the config variables hooks, and that way we'd have the cache handled there.

v-next/hardhat-node-test-runner/src/protocol.ts
```diff
@@ -0,0 +1,148 @@
+import type { HardhatRuntimeEnvironment } from "hardhat/types/hre";
```
alcuadrado, line 1: This, and the associated handlers may be so common that we should place them in core. I guess mocha running in parallel would also need them.

v-next/hardhat-node-test-runner/src/protocol.ts
antico5: In what scenario wouldn't it work ? you mean running the tests with concurrency > 1 ?

v-next/hardhat-node-test-runner/src/protocol.ts
antico5: Should I consider switching to tcp socket instead of file ? I don't think there are any downsides besides choosing an available port. Looks like using file sockets have some overhead

v-next/hardhat-node-test-runner/src/protocol.ts
antico5: Should I override the config var hook alone, or also keep the current override of secret input (but without caching)?

v-next/hardhat-node-test-runner/src/protocol.ts
antico5: should I do this in the scope of this PR ? (moving to core + checking/adding support for mocha in parallel)

kanej:
After discussion, we are going to close this PR - it help us understand the problem, but we intend to take a simpler approach to the child process context for user interruptions (initially just throwing an error).

PR #6849, opened 2025-06-12
Hardhat verify tests

schaable:


v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -2573,7 +2573,7 @@ Please ensure each library address is correct. You can remove entries from your
 This can occur if a library is only used in the contract's constructor. The missing libraries are:
 {missingList}
 
-{solution}`,
+Visit https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-verify#libraries-with-undetectable-addresses to learn how to solve this.`,
```
kanej: Is this not likely to change with the new docs?

v-next/hardhat-verify/src/internal/verification.ts
```diff
@@ -62,6 +64,9 @@ export interface VerifyContractArgs {
 export async function verifyContract(
   verifyContractArgs: VerifyContractArgs,
   hre: HardhatRuntimeEnvironment,
+  consoleLog: (text: string) => void = console.log,
+  testDispatcher?: Dispatcher,
```
kanej: I think we should differentiate between the internal `verifyContract` and the external exported `verifyContract`. So the exported version has only the first two args, but delegates down to the internal version injecting implementations of logger, httpDispatcher and EthereumProvider.
The test suite can cut in based on the internal version and provide alternate implementations.
I would make the parameters required in the internal version, but I will leave that as a discretionary choice.

v-next/hardhat-errors/src/descriptors.ts
schaable: Possibly. Should we remove it from the descriptor for now?

kanej: Yes remove it. My ideal would be a placeholder that prompts us to update it, but we plan to do a complete review of errors before the Beta. So lets assume we pick it up then.

v-next/hardhat-verify/test/hook-handlers/config.ts
```diff
@@ -0,0 +1,263 @@
...
+
+      assertValidationErrors(validationErrors, [
+        {
+          path: ["verify"],
+          message: "Expected object, received string",
+        },
+      ]);
+    });
+
+    it("should throw if verify is not an object", async () => {
```
kanej: Is this a duplicate?

v-next/hardhat-verify/test/verification.ts
```diff
@@ -0,0 +1,309 @@
...
+          () => {},
+          testDispatcher.interceptable,
+          provider,
+        );
+
+        assert(result, "Verification should return true");
+      });
+    });
+
+    describe.todo("all cases", () => {
```
kanej, line 155:
```suggestion
    // TODO: Include remaining `hardhat-verify` verification test cases
    describe.todo("all cases", () => {
```

v-next/hardhat-verify/src/internal/etherscan.ts
```diff
@@ -41,18 +44,26 @@ export class Etherscan {
...
+  readonly #testDispatcher?: Dispatcher;
+  readonly #pollingIntervalMs: number;
+
   constructor(etherscanConfig: {
     chainId: number;
     name?: string;
     url: string;
-    apiUrl: string;
     apiKey: string;
+    testDispatcher?: Dispatcher;
```
kanej: Can we rename this to `dispatcher` or `givenDispatcher` or something else?
I know our only intended usage is in tests, but I think constructor injection is valid approach, so we don't need to shamefully call it `testDispatcher`.

v-next/hardhat-verify/src/internal/etherscan.ts
```diff
@@ -41,18 +44,26 @@ export class Etherscan {
...
   }) {
     this.chainId = String(etherscanConfig.chainId);
     this.name = etherscanConfig.name ?? "Etherscan";
     this.url = etherscanConfig.url;
     this.apiUrl = ETHERSCAN_API_URL;
     this.apiKey = etherscanConfig.apiKey;
+    this.#testDispatcher = etherscanConfig.testDispatcher;
+    this.#pollingIntervalMs =
+      etherscanConfig.testDispatcher !== undefined
+        ? 0
```
kanej, line 65: Nit: I would be tempted to inject the polling interval as well - but this is internal code, so I am happy to keep it.

v-next/hardhat-verify/src/internal/etherscan.ts
schaable: I've also renamed `testProvider` for consistency.

v-next/hardhat-verify/src/internal/etherscan.ts
schaable: Injecting the polling interval would require us to also pass it to `verifyContract` or resolve it there, which I'd rather avoid. Since we're not testing this directly, just using it to keep tests fast, I’d prefer to leave it as is.

kanej: Cool. Agreed.

PR #6857, opened 2025-06-16
Add configurable maxFeePerGas and default to it if set

zoeyTM:
We already have this field configurable for `maxPriorityFeePerGas` but not `maxFeePerGas`. By adding this and defaulting to it if its set by the user, we can eliminate the need for extending the hardcoded zero gas fee chains list such as in [this PR](https://github.com/NomicFoundation/hardhat-ignition/pull/866).

### DIFF
--- a/packages/hardhat-ignition-core/test-integrations/new-api/internal/new-execution/jsonrpc-client.ts
+++ b/packages/hardhat-ignition-core/test-integrations/new-api/internal/new-execution/jsonrpc-client.ts
@@ -126,15 +126,13 @@ describe("JSON-RPC client", function () {
             this.hre.network.provider,
             {
               maxFeePerGas: 1n,
-              maxPriorityFeePerGas: 0n,
             }
           );
           const fees = await maxFeeClient.getNetworkFees();
 
           assert("maxFeePerGas" in fees);
 
-          // internal logic multiplies the given maxFeePerGas by 2 and adds the maxPriorityFeePerGas
-          assert.equal(fees.maxFeePerGas, 2n);
+          assert.equal(fees.maxFeePerGas, 1n);
         });
 
         it("Should use return legacy fees when deploying to polygon network (chainId 137)", async function () {
--- a/.changeset/eighty-pianos-wait.md
+++ b/.changeset/eighty-pianos-wait.md
@@ -1,0 +1,6 @@
+---
+"@nomicfoundation/ignition-core": patch
+"@nomicfoundation/hardhat-ignition": patch
+---
+
+Add configurable maxFeePerGas and default to it if set


zoeyTM:
Note that there is one integration test failing, however, I confirmed that this test is also failing for me locally on an unmodified `main` branch. I'm guessing there was an EDR update that broke it, given the nature of the test, but I'm unsure.

packages/hardhat-ignition-core/src/internal/execution/jsonrpc-client.ts
```diff
@@ -687,7 +688,8 @@ export class EIP1193JsonRpcClient implements JsonRpcClient {
 
       // Logic copied from ethers v6
       const maxFeePerGas =
-        latestBlock.baseFeePerGas * 2n + maxPriorityFeePerGas;
+        (this._config?.maxFeePerGas ?? latestBlock.baseFeePerGas) * 2n +
+        maxPriorityFeePerGas;
```
fvictorio: This doesn't look correct to me, conceptually. I think it should be something like:

```typescript
      const maxFeePerGas =
        this._config?.maxFeePerGas ?? ((latestBlock.baseFeePerGas) * 2n +
        maxPriorityFeePerGas);
```

Otherwise the `maxFeePerGas` in the config is being interpreted as the base fee of the block, which doesn't make sense to me.

packages/hardhat-ignition-core/test-integrations/test-execution-result-fixtures.ts
```diff
@@ -112,6 +112,7 @@ describe("execution-result-fixture tests", function () {
           `Fixture for ${contractName}.${abiItem.name} missing`
         );
 
+        console.log(contractName, abiItem.name);
```
fvictorio: Should this be removed?

packages/hardhat-ignition-core/src/internal/execution/jsonrpc-client.ts
zoeyTM: Ah! You're absolutely right

packages/hardhat-ignition-core/test-integrations/new-api/internal/new-execution/jsonrpc-client.ts
```diff
@@ -121,6 +121,22 @@ describe("JSON-RPC client", function () {
...
+              maxFeePerGas: 1n,
+              maxPriorityFeePerGas: 0n,
+            }
+          );
+          const fees = await maxFeeClient.getNetworkFees();
+
+          assert("maxFeePerGas" in fees);
+
+          // internal logic multiplies the given maxFeePerGas by 2 and adds the maxPriorityFeePerGas
+          assert.equal(fees.maxFeePerGas, 2n);
```
fvictorio: Shouldn't this have been updated too? Tests seem to pass though, so I'm confused about the `2`.

zoeyTM: I think integration tests are disable right now on CI for some reason. This would also explain that one other test I mentioned that was failing for me locally despite CI passing.

zoeyTM: Good catch, I just pushed a fix for this test (though they still won't run on CI yet)

kanej:
@zoeyTM can you open a PR for the docs, to add the new config option?

PR #6868, opened 2025-06-16
chore: move coverage directory to the root of the project

galargh:

---

This PR addresses the suggestion from https://nomicfoundation.slack.com/archives/C081DQFBGP3/p1750081460946359

It updates the path under which we store coverage related files from `$ROOT_PATH/cache/coverage` to `$ROOT_PATH/coverage`. In particular, `lcov.info` will now be stored under `$ROOT_PATH/coverage/lcov.info`.

I also updated the `gitignore` files in the template projects and the `.gitignore` in the example project.

Finally, I added a new `onClean` hook to the built-in `clean` plugin and used it from the `coverage` plugin to clean up the `coverage` directory when `hardhat clean` task is called.

### DIFF
--- a/v-next/hardhat/test/internal/builtin-plugins/clean/task-action.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/clean/task-action.ts
@@ -2,7 +2,7 @@ import type { HardhatRuntimeEnvironment } from "../../../../src/types/hre.js";
 
 import assert from "node:assert/strict";
 import path from "node:path";
-import { before, beforeEach, describe, it } from "node:test";
+import { before, beforeEach, describe, it, mock } from "node:test";
 
 import { useFixtureProject } from "@nomicfoundation/hardhat-test-utils";
 import {
@@ -22,9 +22,13 @@ let globalCacheDir: string;
 let cacheDir: string;
 let artifactsDir: string;
 
+const onClean = mock.fn(async () => {});
+
 function assertCleanBehavior(global: boolean) {
   it("should clean the cache and artifacts directories", async () => {
     await cleanAction({ global }, hre);
+
+    assert.equal(onClean.mock.calls.length, 1, "onClean was called once");
 
     // If the cache dir exists, it should be empty
     if (await exists(cacheDir)) {
@@ -66,6 +70,13 @@ describe("clean/task-action", () => {
         // TODO remove this once cache and artifacts are resolved in the config
         paths: { cache: cacheDir, artifacts: artifactsDir },
       });
+      hre.hooks.registerHandlers("clean", {
+        onClean,
+      });
+    });
+
+    beforeEach(async () => {
+      onClean.mock.resetCalls();
     });
 
     describe("when cache and artifact dirs don't exist", async () => {


.changeset/friendly-moons-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Made lcov.info coverage report available under coverage directory in the root of the project
```
kanej:
```suggestion
Made lcov.info coverage report available under coverage directory in the root of the project ([#6891](https://github.com/NomicFoundation/hardhat/issues/6891))
```

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/clean.ts
```diff
@@ -0,0 +1,12 @@
+import type { CleanHooks } from "../../../../types/hooks.js";
+
+import { remove } from "@nomicfoundation/hardhat-utils/fs";
+
+import { getCoveragePath } from "../helpers.js";
+
+export default async (): Promise<Partial<CleanHooks>> => ({
+  onClean: async (context) => {
+    const coveragePath = getCoveragePath(context.config.paths.root);
```
kanej, line 9: Should the coverage path be set as path in config then used from here?

galargh: I understood from https://nomicfoundation.slack.com/archives/C081DQFBGP3/p1750084648063699?thread_ts=1750081460.946359&cid=C081DQFBGP3 that we don't want to add it to config yet.

v-next/hardhat/src/internal/builtin-plugins/clean/task-action.ts
```diff
@@ -9,8 +9,10 @@ interface CleanActionArguments {
 
 const cleanAction: NewTaskActionFunction<CleanActionArguments> = async (
   { global },
-  { config },
+  { config, hooks },
 ) => {
+  await hooks.runParallelHandlers("clean", "onClean", []);
```
kanej: I am tempted to say we move the hook to the end, that way our default cleanup happens even if a plugin fails. What do you think?

galargh: I put it before as I thought it was potentially more useful to be able to do something before files are cleaned up, but I don't mind moving it to after. For the use case we have for it now, it works either way.

kanej (changes requested):
This looks good, I left a few comments.
I would add a test to check that the new clean hook is run, but that might be me playing for a code coverage high score. Do you think a test would justify itself?

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/clean.ts
kanej: Cool. Lets leave it out of config then.

v-next/hardhat/src/internal/builtin-plugins/clean/task-action.ts
kanej: That strikes me as a `pre-clean` hook. I think we move it to the end and then we always have the option of adding an additional `pre-clean` hook.

PR #6889, opened 2025-06-17
Revamp dependency resolution

alcuadrado:
This PR introduces a new dependency resolver. The main changes are:

- The `remappings` field was removed from the config.
- `remappings.txt` files are loaded and processed, both from the Hardhat project and npm packages.
- Every source name has either an `npm/` or a `project/` prefix now.
- The dependency graph was modified so that every edge has the set of remappings that were used to resolve that dependency.
- The `dependenciesToCompile` now behave exactly the same as imports into npm, including being affected by user remappings.
- Improved the error handling by switching from exceptions into a `Result` type.


v-next/hardhat-ignition-ethers/test/strategies/helper-invocation.ts
```diff
@@ -35,7 +35,7 @@ describe("strategies - invocation via helper", () => {
 
       assert.equal(
         await result.foo.getAddress(),
-        "0x647fB9ef6cd97537C553f6cC3c7f60395f81b410",
```
kanej, line 38: I am happy with these changes to the Ignition tests.

I tested manually by:

* Using the `create2` strategy
  * altering the `from` account that these `create2` tests used for deployment - no change to the address - as expected
  * altering the source of the used contracts - changed address - as expected
* Using the `default` strategy
  * altering the `from` account tests used for deployment - change of address - as expected for standard deployment transactions
  * altering the source of the used contracts - no change of address - as expected for standard deployment transactions

The create2/default strategy are behaving as expected - the only change is the bytecode that the fixture projects is producing, hence this is commit represents safe updates to these expected addresses.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/dependency-graph-building.ts
```diff
@@ -1,41 +1,75 @@
...
+            error: formatProjectRootResolutionError(resolutionResult.error),
+          },
+        );
+      }
+
+      resolvedFile = resolutionResult.value;
+      dependencyGraph.addRootFile(
+        // We don't use the package's root source name in the local files public
+        // source name.
+        resolvedFile.sourceName.substring(
```
antico5, line 68: What do you think about encapsulating this logic in e.g. `ResolvedFile#publicSourceName()` ?

alcuadrado: I thought about it, but a single ResolvedFile could have multiple public source names. 

Imagine a project that has this `remappings.txt`:

```
foo/=node_modules/foo/
another-foo/=node_modules/foo/
```

and

```
dependenciesToCompile: ["foo/A.sol", "another-foo/A.sol"]
```

In that case we'll have a file `npm/foo@1.2.3/A.sol` with the public source names `foo/A.sol` and `another-foo/A.sol`.

The reason is that the "public" part of it is something that belongs to the graph and compilation job, and not to the file itself.

alcuadrado: I know this is a weird example, but it's intended to show how I realized that public source names can't belong to the file itself, as it's more of a way to refer to the file, rather than a property of it.

.changeset/hip-crabs-notice.md
```diff
@@ -0,0 +1,12 @@
...
+"@nomicfoundation/hardhat-errors": patch
+"@nomicfoundation/hardhat-ignition-ethers": patch
+"@nomicfoundation/hardhat-ignition-viem": patch
+"@nomicfoundation/hardhat-ignition": patch
+"@nomicfoundation/hardhat-node-test-reporter": patch
+"@nomicfoundation/hardhat-utils": patch
+"hardhat": patch
+---
+
+Revamp the dependency resolution system: adding support for `remappings.txt` files, and dropping support for remappings from the config.
```
alcuadrado:
```suggestion
Revamp the dependency resolution system: adding support for `remappings.txt` files, dropping support for `remappings` from the config, and changing how remappings into npm packages are written. 
```

PR #6899, opened 2025-06-18
Fix warning: passing args to `spawn` when shell is set to true

### LINKED ISSUE #6896: Avoid `[DEP0190] DeprecationWarning` during project init on Node 24
If you initialize a project in Node 24 you get `[DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.`. We want to remove this deprecation warning.

The following approach appears to work:

```
import { spawn } from "node:child_process";

// Prints de deprecation warning
spawn("node", ["--version"], { stdio: "inherit" });

// Doesn't print it
spawn("node --version", { shell: true, stdio: "inherit" });
```

antico5:
Closes #6896

### DIFF
--- a/v-next/hardhat/src/internal/cli/init/init.ts
+++ b/v-next/hardhat/src/internal/cli/init/init.ts
@@ -305,11 +305,15 @@ export async function validatePackageJson(
   // package managers support `<package manager> pkg set type=module`.
   const packageManagerToUse = packageManager === "pnpm" ? "pnpm" : "npm";
 
-  await spawn(packageManagerToUse, ["pkg", "set", "type=module"], {
-    cwd: workspace,
-    shell: true,
-    stdio: "inherit",
-  });
+  await spawn(
+    [packageManagerToUse, "pkg", "set", "type=module"].join(" "),
+    [],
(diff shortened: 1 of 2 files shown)


v-next/hardhat/src/internal/cli/init/init.ts
```diff
@@ -469,7 +473,7 @@ export async function installProjectDependencies(
       console.log();
       console.log(commandString);
 
-      await spawn(command[0], command.slice(1), {
+      await spawn([command[0], ...command.slice(1)].join(" "), [], {
```
schaable:
```suggestion
      await spawn(commandString, [], {
```

v-next/hardhat/src/internal/cli/init/init.ts
```diff
@@ -513,7 +517,7 @@ export async function installProjectDependencies(
...
```
schaable:
```suggestion
      await spawn(commandString, [], {
```

schaable (approved):

PR #6901, opened 2025-06-18
[WIP] Drop node 22

alcuadrado:
A PR to drop support for Node 22. Currently in WIP, as I'm waiting to see if I can get this https://github.com/DefinitelyTyped/DefinitelyTyped/pull/73062 in, so that we can enable `no-floating-promises` in our tests.


kanej:
I propose we close this PR as we are unlikely to drop Node 22 support for Hardhat 3 - vscode will be on Node 22 for the foreseeable future.

PR #6919, opened 2025-06-23
chore: bump the version of the viem toolbox

kanej:
We released a new major version of the viem toolbox. Hence the alpha version has to be incremented by a major version.

## Testing

I tested the version change manually using verdaccio, then installed both the viem and ethers toolbox template projects and ran through the standard compile/test/run/ignition smoke tests.

### DIFF
--- a/.changeset/pre.json
+++ b/.changeset/pre.json
@@ -17,7 +17,7 @@
     "@nomicfoundation/hardhat-node-test-reporter": "2.0.0",
     "@nomicfoundation/hardhat-node-test-runner": "2.0.0",
     "@nomicfoundation/hardhat-test-utils": "2.0.0",
-    "@nomicfoundation/hardhat-toolbox-viem": "3.0.0",
+    "@nomicfoundation/hardhat-toolbox-viem": "4.0.0",
     "@nomicfoundation/hardhat-typechain": "2.0.0",
     "@nomicfoundation/hardhat-utils": "2.0.0",
     "@nomicfoundation/hardhat-toolbox-mocha-ethers": "2.0.0",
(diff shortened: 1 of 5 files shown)


.changeset/pre.json
```diff
@@ -17,7 +17,7 @@
     "@nomicfoundation/hardhat-node-test-reporter": "2.0.0",
     "@nomicfoundation/hardhat-node-test-runner": "2.0.0",
     "@nomicfoundation/hardhat-test-utils": "2.0.0",
-    "@nomicfoundation/hardhat-toolbox-viem": "3.0.0",
```
kanej, line 20: The released version of `@nomicfoundation/hardhat-toolbox-viem` is now 4.0.0

v-next/hardhat-toolbox-viem/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nomicfoundation/hardhat-toolbox-viem",
-  "version": "4.0.0-next.17",
```
kanej, line 3: With the released version being `4.0.0`, we bump the pre-release version to `5.0.0-next.17`. So the next major version, but we keep the pre-release number in line with all other packages i.e. `-next.17`.

v-next/hardhat/templates/01-node-test-runner-viem/package.json
```diff
@@ -6,7 +6,7 @@
   "type": "module",
   "devDependencies": {
     "hardhat": "workspace:^3.0.0-next.17",
-    "@nomicfoundation/hardhat-toolbox-viem": "workspace:^4.0.0-next.17",
```
kanej, line 9: We bump the dependency in the template project.

Issue #6827, opened 2025-06-09
Blockscout support


kanej:
Replicate the Etherscan support in `hardhat-verify` for Hardhat 3 with Blockscout support.

PR #6920, opened 2025-06-23
hardhat-verify Blockscout support

schaable:
Added support for verifying contracts in Blockscout. Closes https://github.com/NomicFoundation/hardhat/issues/6827


v-next/hardhat-verify/src/internal/blockscout.ts
```diff
@@ -0,0 +1,359 @@
...
+        {
+          name: this.name,
+          url: this.apiUrl,
+          statusCode: response.statusCode,
+          errorMessage: responseBody.result,
+        },
+      );
+    }
+
+    const blockscoutResponse = new BlockscoutVerificationStatusResponse(
```
kanej, line 275: Nit: no need to change - or even agree.
I suspect I would have dodged the class here and done a straight transform to pure data, rather than wrapping and deferring the calculation of the results.
This status response class feels like a pure data carrier.

v-next/hardhat-verify/src/internal/hook-handlers/config.ts
```diff
@@ -62,13 +70,23 @@ export async function resolveUserConfig(
   };
 }
 
+function resolveBlockscoutConfig(
+  blockscoutConfig: BlockscoutUserConfig | undefined = {
+    enabled: false,
+  },
+): BlockscoutConfig {
+  return {
+    enabled: blockscoutConfig.enabled ?? true,
```
kanej, line 79: Does this mean we always try to verify with blockscout even if no blockscout config is set by the user? Is that expected?

schaable:
@kanej First one is a valid question. I did it this way for consistency between verification providers, but it's true it might be annoying.
Regarding the second issue, could you retry? I've seen that error from Blockscout when the contract hasn't yet been uploaded to their database. If you check the URL at that time, you'll notice it appears as a regular address instead of a contract. However, after waiting a couple of minutes, Blockscout updates the address to a contract, and then verification succeeds.
Maybe we could improve that error message by suggesting users wait a couple of minutes and retry?

v-next/hardhat-verify/src/internal/hook-handlers/config.ts
schaable: If no Blockscout config is provided at all, the provider is disabled by default (as we use `{ enabled: false }` as default for `blockscoutConfig`). If a Blockscout key is provided (`blockscout: {}`), we assume it's enabled. This behavior matches Etherscan, except Etherscan uses an `apiKey`, while Blockscout does not.

v-next/hardhat-verify/src/internal/blockscout.ts
schaable: I agree. The same applies to the `EtherscanVerificationStatusResponse` class. I intend to clean this up in the future.

kanej:
> @kanej First one is a valid question. I did it this way for consistency between verification providers, but it's true it might be annoying. Regarding the second issue, could you retry? I've seen that error from Blockscout when the contract hasn't yet been uploaded to their database. If you check the URL at that time, you'll notice it appears as a regular address instead of a contract. However, after waiting a couple of minutes, Blockscout updates the address to a contract, and then verification succeeds. Maybe we could improve that error message by suggesting users wait a couple of minutes and retry?

I am still seeing the error:

image

I think it makes sense to update the error message to indicate that timing might be an issue for blockscout.

kanej:
> > @kanej First one is a valid question. I did it this way for consistency between verification providers, but it's true it might be annoying. Regarding the second issue, could you retry? I've seen that error from Blockscout when the contract hasn't yet been uploaded to their database. If you check the URL at that time, you'll notice it appears as a regular address instead of a contract. However, after waiting a couple of minutes, Blockscout updates the address to a contract, and then verification succeeds. Maybe we could improve that error message by suggesting users wait a couple of minutes and retry?
> 
> I am still seeing the error:
> 
> image
> 
> I think it makes sense to update the error message to indicate that timing might be an issue for blockscout.

It was a timing issue - on a rerun I get a working outcome:

image

v-next/hardhat-verify/src/internal/hook-handlers/config.ts
kanej: Ah cool, okay.

schaable:
> Looks good. I am pre-approving. The only remaining question is the product one around requiring `blockscout: {enabled: true}` in config.

Update from Pato:
We're changing the config defaults in a follow-up PR. The idea is that all providers will be enabled by default. If no config is provided, the provider is assumed to be enabled.

Issue #6873, opened 2025-06-17
add task to change password in `hardhat-keystore`


kanej:
Support changing the password in `hardhat-keystore`.

PR #6922, opened 2025-06-24
Add task to change password in hardhat-keystore

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/6873

Internal link to design dock: https://www.notion.so/nomicfoundation/Change-keystore-password-21b578cdeaf580faa436d1fd322c8d21

### TO DISCUSS
Do we want to keep the task name as `change-password`?
Other proposed solutions:
- set-password
- password

My concern is that `set-password` (or simply `password`) could be misinterpreted as performing keystore initialization, which is currently done by the `set` task.
Using `change-password` as name makes the task’s purpose immediately obvious.

It's worth mentioning that since these commands are unlikely to be invoked programmatically, we can safely rename or adjust them later without affecting existing users.

### DIFF
--- a/v-next/hardhat-keystore/test/integration/tasks.ts
+++ b/v-next/hardhat-keystore/test/integration/tasks.ts
@@ -139,7 +139,7 @@ describe("integration tests for the keystore tasks", () => {
         "Password changed successfully!\n",
     );
 
-    tmpHre = tmpHre = await createHardhatRuntimeEnvironment({
+    tmpHre = await createHardhatRuntimeEnvironment({
       plugins: [
         hardhatKeystorePlugin,
         setupKeystoreFileLocationOverrideAt(keystoreFilePath),


v-next/hardhat-keystore/test/tasks/change-password.ts
```diff
@@ -0,0 +1,139 @@
...
+  let keystoreLoader: KeystoreLoader;
+  let keystoreLoaderTmp: KeystoreLoader;
+
+  beforeEach(() => {
+    mockFileManager = new MockFileManager();
+    mockConsoleLog = mock.fn();
+
+    // Mock this function because the `change-password` task renames the keystore file
+    // from its temporary name back to the original, so the file must exist on disk.
+    mockFileManager.writeJsonFile = mock.fn(
```
ChristopherDedominici, line 46: I’m open to alternative solutions. 

Explanation: 
currently, the `change-password` function ends by renaming the temporary keystore file to overwrite the original filename, which means the source path must exist (even as an empty placeholder file)

v-next/hardhat-keystore/test/integration/tasks.ts
```diff
@@ -118,6 +118,41 @@ describe("integration tests for the keystore tasks", () => {
       `${keystoreFilePath}\n`,
     );
   });
+
+  it("should change the password", async () => {
+    const NEW_PASSWORD = "newPassword";
+
+    let tmpHre = await createHardhatRuntimeEnvironment({
```
ChristopherDedominici, line 125: Create a new `hre` because the plugin `setupKeystorePassword` has different passwords compared to the `hre` used in the other tests

alcuadrado:
`change-password` looks good to me

v-next/hardhat-keystore/test/integration/tasks.ts
```diff
@@ -118,6 +118,41 @@ describe("integration tests for the keystore tasks", () => {
...
+    await _assertConsoleOutputMatchesFor(
+      () => tmpHre.tasks.getTask(["keystore", "change-password"]).run(),
+      "Unlock the keystore using your current password before proceeding with the password change.\n" +
+        "Change your password.\n" +
+        "The password must have at least 8 characters.\n" +
+        "\n" +
+        "Password changed successfully!\n",
+    );
+
+    tmpHre = tmpHre = await createHardhatRuntimeEnvironment({
```
schaable:
```suggestion
    tmpHre = await createHardhatRuntimeEnvironment({
```

PR #6928, opened 2025-06-25
feat: add watcher to hardhat node

galargh:

---

This replaces https://github.com/NomicFoundation/hardhat/pull/6524 

It is a simplified approach to watch for build completion events from `hardhat node` processes. Unlike the original approach, it relies on the fact that the build info output file and the build info file are created sequentially. This is achieved by first creating those files in parallel in the cache directory and moving them one by one to the artifacts directory only after both have been fully written to disk.


kanej:
We think we can land this after partial recompilation - we need to check that the changed locations for outputs are reviewed.

.changeset/old-bottles-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Add compilation results to a running node on recompilation
```
kanej:
```suggestion
Add compilation results to a running node on recompilation ([#6040](https://github.com/NomicFoundation/hardhat/issues/6040))
```

antico5:
@kanej this is ready for review. I changed it to use a new internal method vs an rpc method for the edr provider, and adjusted a couple lines related to recent build system changes.

.changeset/old-bottles-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Add compilation results to a running node on recompilation ([#6040](https://github.com/NomicFoundation/hardhat/issues/6040))
```
kanej:
```suggestion
Automatically add compilation results to a running node on recompilation ([#6040](https://github.com/NomicFoundation/hardhat/issues/6040))
```

v-next/hardhat/src/internal/builtin-plugins/node/artifacts/build-info-watcher.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  const isBuildInfoFile =
+    absolutePath.endsWith(".json") && !absolutePath.endsWith(".output.json");
+
+  if (!isBuildInfoFile) {
+    log(`File ${absolutePath} is not a build info file`);
+    return;
+  }
+
+  const buildId = path.basename(absolutePath).replace(".json", "");
```
kanej, line 42: Is there any implication from the change to include the solc version in the file name?
Do we consider the version to be part of the build id now?

v-next/hardhat/src/internal/builtin-plugins/node/artifacts/build-info-watcher.ts
```diff
@@ -0,0 +1,82 @@
...
+    ignoreInitial: true,
+    awaitWriteFinish: {
+      stabilityThreshold: STABILITY_THRESHOLD,
+      pollInterval: POLL_INTERVAL,
+    },
+  });
+
+  // NOTE: We listen only to the "add" event because the contents of the build info
+  // files identified by a build id should be considered immutable under usual circumstances.
+  watcher.on("add", listener.bind(null, handler));
```
kanej, line 79: Can I double check, we expect the update to happen when:
1. A new contract is added and then a compile done (leading to a completely new build info file)
2. An update on an existing contract happens, but this triggers a new build info file?
Or is it the case that we don't expect 2?

antico5: When a contract is modified and recompiled, a new build info file is always created. Since build info filenames are a hash of their own content, their content never changes, so no need to listen to update events

v-next/hardhat/src/internal/builtin-plugins/node/artifacts/build-info-watcher.ts
antico5: No implications. The build id in this case is used to get the build info filename. But even if it was used for something else, the whole concept of build id now includes the solc version on the id itself, besides the hash

PR #6936, opened 2025-06-26
Fix solidity tests selection and output

alcuadrado:


alcuadrado:
Merging without review as this includes an important quickfix

v-next/hardhat/src/internal/builtin-plugins/solidity-test/edr-artifacts.ts
```diff
@@ -0,0 +1,134 @@
...
+    // from an artifact id between the `run` call and the events emitted by the
+    // test runner.
+    return {
+      edrAtifact: {
+        id,
+        contract,
+      },
+      userSourceName: artifact.sourceName,
+    };
+  });
```
galargh, line 133: Note to anyone else reviewing this PR. When the contents of `edr-artifacts.ts` were in the `reporter.ts`, this return statement was:

```
return {
  id,
  contract,
};
```

There were no other meaningful changes to the functions now defined in this file.

PR #6941, opened 2025-06-27
feat: added support for short option names

galargh:

---

This PR adds support for short option names as per https://www.notion.so/nomicfoundation/Shorthand-flags-handling-20a578cdeaf580358bf0e4eba59324f1. It also disallows grouping and repetition of options. This PR does NOT add LEVEL nor FLAG argument types. They will be added in a separate PR to simplify the review process.


v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -344,7 +344,17 @@ For global options help run: hardhat --help`;
 
         const expected = `${chalk.bold("A task that uses arg1")}
 
-Usage: hardhat [GLOBAL OPTIONS] task
+Usage: hardhat [GLOBAL OPTIONS] task [--arg-1 <STRING>] [--arg-4] [--] arg2 arg3
+
+OPTIONS:
+
+  --arg-1       (default: <default-value1>)
+  --arg-4       (default: false)
```
fvictorio: Shouldn't this be:

```
  -o, --arg-1       (default: <default-value1>)
  -f, --arg-4       (default: false)
```

?

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -489,13 +499,32 @@ function parseOptions(
...
+    let optionDefinition: OptionDefinition | undefined;
+
+    const providedByName = arg.startsWith("--");
+    const providedByShortName = !providedByName && arg.startsWith("-");
+
+    if (providedByName) {
+      const name = kebabToCamelCase(arg.substring(2));
+      optionDefinition = optionDefinitions.get(name);
+    } else if (providedByShortName) {
+      const shortName = arg.substring(1, 2);
```
fvictorio: PP:

```suggestion
      const shortName = arg.substring[1];
```

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -511,6 +540,22 @@ function parseOptions(
       );
     }
 
+    const optionName = optionDefinition.name;
+
+    // Check if the short name is valid again now that we know its type
+    if (
+      providedArguments[optionName] !== undefined ||
+      (providedByShortName && arg.length > 2)
```
fvictorio, line 548: IMO this is too clever, but I see that it was improved in a follow-up PR, so I think we can ignore it here.

v-next/hardhat/src/internal/core/tasks/task-manager.ts
```diff
@@ -219,6 +219,35 @@ export class TaskManagerImplementation implements TaskManager {
...
+          globalOptionEntry,
+        );
+      }
+    }
+    const optionShortNames = Object.values(taskDefinition.options)
+      .map((option) => option.shortName)
+      .filter((shortName) => shortName !== undefined);
+
+    optionShortNames.forEach((argName) => {
+      const globalOptionEntry = globalOptionDefinitions.get(argName);
```
fvictorio: I think this is incorrect, because `argName` here is a short name, but the keys of `globalOptionDefinitions` are long names. I verified this with this example config:

```ts
const fooTask = task("foo")
  .addFlag({
    name: "fooOption",
    shortName: "m",
  })
  .setAction(() => {}).build()

const myPlugin: HardhatPlugin = {
  id: "my-plugin",
  globalOptions: [{
    name: "myPluginOption",
    shortName: "m",
    type: ArgumentType.BOOLEAN,
    description: "An example plugin option",
    defaultValue: false,
  }]
}

  plugins: [myPlugin],
  tasks: [fooTask],
```

This should've raised an error, but didn't.

galargh: Good catch! I added a test for this.

v-next/hardhat/test/internal/cli/main.ts
galargh: Good call, I put them in reversed order though. `--long-name, -s` as the first is more descriptive.

fvictorio:
Just one comment from my manual testing. If I do:

```
.addFlag({
    name: "fooOption",
    shortName: "foo",
})
```

the error message is:

```
Error HHE503: Argument name "foo" is invalid. It must consist only of alphanumeric characters and cannot start with a number.
```

Instead, it should explicitly say that short names can only be a single letter.

fvictorio (approved):

PR #6942, opened 2025-06-27
feat: added FLAG and LEVEL arguments types

galargh:

---

This PR adds FLAG and LEVEL argument types with their special behaviours as per https://www.notion.so/nomicfoundation/Shorthand-flags-handling-20a578cdeaf580358bf0e4eba59324f1. This PR targets https://github.com/NomicFoundation/hardhat/pull/6941, which introduces short option names.


v-next/hardhat/src/internal/core/arguments.ts
```diff
@@ -249,3 +273,23 @@ function validateAndParseBoolean(name: string, value: string): boolean {
 
   return normalizedValue === "true";
 }
+
+// NOTE: Even though flags do not accept values, we still provide a validation
+// funciton for consistency. Alternatively, we could always throw an error when
```
galargh:
```suggestion
// function for consistency. Alternatively, we could always throw an error when
```

.changeset/long-bats-nail.md
```diff
@@ -0,0 +1,6 @@
+---
+"hardhat": patch
+"@nomicfoundation/hardhat-typechain": patch
```
fvictorio, line 3: The typechain change doesn't seem to be user-facing.

v-next/hardhat/src/internal/core/arguments.ts
```diff
@@ -249,3 +273,23 @@ function validateAndParseBoolean(name: string, value: string): boolean {
 
   return normalizedValue === "true";
 }
+
+// NOTE: Even though flags do not accept values, we still provide a validation
+// function for consistency. Alternatively, we could always throw an error when
+// we receive a value for a flag.
```
fvictorio: I think I'd rather throw an assertion error. Otherwise this is both misleading and could potentially hide issues.

v-next/hardhat/src/internal/core/config-validation.ts
```diff
@@ -523,6 +536,20 @@ export function validatePositionalArguments(
 
           break;
         }
+        case ArgumentType.LEVEL:
```
fvictorio: It seems a bit weird to support FLAG and LEVEL as positional parameters, given that their difference with BOOLEAN and INT are only relevant when used as options.

I'm not against this if it simplifies things, but I wonder if we should forbid it, even if it's just at runtime. That is, doing:

```ts
        case ArgumentType.FLAG:
        case ArgumentType.LEVEL: {
          throw ...
        }
```

I don't have a strong opinion here, so up to you if you want to make this change.

v-next/hardhat/src/internal/core/config.ts
```diff
@@ -89,7 +89,22 @@ export function globalFlag(options: {
...
   return buildGlobalOptionDefinition({
     ...options,
-    type: ArgumentType.BOOLEAN,
+    type: ArgumentType.FLAG,
     defaultValue: false,
   });
 }
+
+/**
+ * Defines a global flag.
```
fvictorio:
```suggestion
 * Defines a global level.
```

.changeset/long-bats-nail.md
galargh: You're right. The behaviour of the flag there changes, but it changes in hardhat as a whole rather than in typechain specifically. I'm going to leave it as is though since we do release everything in v-next anyway.

v-next/hardhat/src/internal/core/arguments.ts
galargh: On it!

v-next/hardhat/src/internal/core/config-validation.ts
galargh: Done! I think it makes sense to throw for positional arguments. We can always revert it if we find a usecase for it.

PR #6945, opened 2025-06-30
feat: added a global verbosity flag

galargh:

---

This PR builds on top of https://github.com/NomicFoundation/hardhat/pull/6942 which added support for level argument types. It adds a global option - `verbosity` - of type level. It can be used to set the verbosity level of debug logging.

After the changes, hardhat debug logging will be enabled when either the `verbose` or `verbosity` flag is provided.

I also added `verbose` and `verbosity` to the global options interface, and simplified the logic of parsing builtin global options (now it uses the same parser all the other options go through).

### DIFF
--- a/v-next/hardhat/src/internal/builtin-global-options.ts
+++ b/v-next/hardhat/src/internal/builtin-global-options.ts
@@ -1,6 +1,6 @@
 import type { GlobalOptionDefinitions } from "../types/global-options.js";
 
-import { globalFlag, globalOption } from "../config.js";
+import { globalFlag, globalLevel, globalOption } from "../config.js";
 import { ArgumentType } from "../types/arguments.js";
 
 export const BUILTIN_GLOBAL_OPTIONS_DEFINITIONS: GlobalOptionDefinitions =
@@ -55,6 +55,17 @@ export const BUILTIN_GLOBAL_OPTIONS_DEFINITIONS: GlobalOptionDefinitions =
         option: globalFlag({
           name: "verbose",
           description: "Enables Hardhat verbose logging.",
+        }),
+      },
+    ],
+    [
+      "verbosity",
+      {
+        pluginId: "builtin",
+        option: globalLevel({
+          name: "verbosity",
+          shortName: "v",
+          description: "Sets the verbosity level.",
         }),
       },
     ],
--- a/v-next/hardhat/src/internal/cli/main.ts
+++ b/v-next/hardhat/src/internal/cli/main.ts
@@ -242,12 +242,16 @@ export async function parseBuiltinGlobalOptions(
   showStackTraces: boolean;
   help: boolean;
   version: boolean;
+  verbose: boolean;
+  verbosity: number;
 }> {
   let configPath: string | undefined;
   let showStackTraces: boolean = isCi();
   let help: boolean = false;
   let version: boolean = false;
   let init: boolean = false;
+  let verbose: boolean = false;
+  let verbosity: number = 0;
 
   const builtinGlobalOptions = await parseGlobalOptions(
     BUILTIN_GLOBAL_OPTIONS_DEFINITIONS,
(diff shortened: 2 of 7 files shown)


v-next/hardhat/src/internal/builtin-global-options.ts
```diff
@@ -58,6 +58,17 @@ export const BUILTIN_GLOBAL_OPTIONS_DEFINITIONS: GlobalOptionDefinitions =
         }),
       },
     ],
+    [
+      "verbosity",
+      {
+        pluginId: "builtin",
+        option: globalLevel({
+          name: "verbosity",
```
fvictorio, line 66: Sorry I didn't notice this before, but I think having both `verbose` and `verbosity` options might be confusing. What do you think of using a more (ahem) verbose name like `--trace-verbosity`, `--execution-trace-verbosity`, `--test-verbosity`, or something like that?

v-next/hardhat/src/internal/builtin-global-options.ts
```diff
@@ -58,6 +58,17 @@ export const BUILTIN_GLOBAL_OPTIONS_DEFINITIONS: GlobalOptionDefinitions =
...
+          shortName: "v",
+          description: "Sets the verbosity level.",
```
fvictorio, line 68: Even if we don't change the name, let's be more explicit here. Something like "Sets the verbosity level of the execution traces".

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -241,73 +242,58 @@ export async function parseBuiltinGlobalOptions(
...
+  }
 
-    if (arg === "--help") {
-      usedCliArguments[i] = true;
-      help = true;
-      continue;
-    }
+  if (builtinGlobalOptions.verbose === true) {
+    verbose = true;
+    verbosity = 1;
```
fvictorio, line 284: I'm not sure it's desirable to enable execution traces when the user does `--verbose`. One is about our internal logs and the other is about the test traces. @alcuadrado wdyt?

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -241,73 +242,58 @@ export async function parseBuiltinGlobalOptions(
...
-    if (arg === "--version") {
-      usedCliArguments[i] = true;
-      version = true;
-      continue;
-    }
+  if (
+    builtinGlobalOptions.verbosity !== undefined &&
+    builtinGlobalOptions.verbosity !== 0
+  ) {
+    verbose = true;
```
fvictorio, line 291: Same comment as above here. This one I'm even more confident it's something we shouldn't do.

v-next/hardhat/src/internal/builtin-global-options.ts
galargh: I intentionally kept `verbose` and `verbosity` similar. We briefly talked about it when we discussed how flags/options/etc should behave in general. Happy to change it if we don't want to do it that way though.

I included a short section about it in the newest traces doc - https://www.notion.so/nomicfoundation/Foundry-style-traces-in-Hardhat-3-229578cdeaf5804c906ef6ae77e7643e?source=copy_link#229578cdeaf580929b85ef66c3735932

This section of the doc is also relevant as it discusses the current behaviour of the `verbose` flag - https://www.notion.so/nomicfoundation/Foundry-style-traces-in-Hardhat-3-229578cdeaf5804c906ef6ae77e7643e?source=copy_link#229578cdeaf58023a00fea78c19b3072

Let me know how we should proceed here.

v-next/hardhat/src/internal/cli/main.ts
galargh: FYI, I included a callout for this here - https://www.notion.so/nomicfoundation/Foundry-style-traces-in-Hardhat-3-229578cdeaf5804c906ef6ae77e7643e?source=copy_link#229578cdeaf580929b85ef66c3735932 - and here - https://www.notion.so/nomicfoundation/Foundry-style-traces-in-Hardhat-3-229578cdeaf5804c906ef6ae77e7643e?source=copy_link#229578cdeaf58023a00fea78c19b3072

If we decide to make `verbosity` separate from `verbose`, I agree that we should change the name and probably make it a task level flag instead of a global one.

v-next/hardhat/src/internal/cli/main.ts
galargh: In my mind, it depends on whether we think of `verbosity`/`verbose` as a single global level concept or if we want to make one solidity test task specific and the other left as-is.

v-next/hardhat/src/internal/builtin-global-options.ts
galargh: I created a thread for this on Slack so we could discuss the options in one place - https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1752064116423929

PR #6954, opened 2025-07-01
Add 'test solidity' test verifying a failure

alcuadrado:
This PR adds a test to the `test solidity` task, verifying that the tests are actually run. This is done by introducing a failing test and asserting its failure.


alcuadrado:
The intention of this test is to caught the error in `-next.19` in the CI.

PR #6955, opened 2025-07-01
feat: added spellcheck workflow for checking spelling (main)

galargh:
This is an equivalent PR to https://github.com/NomicFoundation/hardhat/pull/6951 but for `main`. As per the suggestion from https://github.com/NomicFoundation/hardhat/pull/6951#discussion_r2177583916, I'm not creating a changelog entry here as it would result in a package release which should not be necessary.


packages/hardhat-ignition-core/test/helpers/execution-result-fixtures.ts
```diff
@@ -399,7 +399,7 @@ export const staticCallResultFixturesArtifacts: {
       },
     ],
     bytecode:
```
galargh, line 401: TODO: Regenerate and update the bytecode value here.

Issue #6866, opened 2025-06-16
Multi-chain support in Solidity Test


kanej:
EDR is being modified to support changing the chain type, e.g. setting it to Optimism, in Solidity Tests.
This functionality needs to be exposed on the Hardhat side, this will require a small amount of design work - however, for ETHGlobal Cannes we intend a smaller change, with further alterations in the API later.

This initial move is to pass `--chain-type` as an option to the Solidity test task, and apply that to the EDR setup.

kanej:
I think the next step is to review and integrate: https://github.com/NomicFoundation/hardhat/pull/6912

PR #6958, opened 2025-07-02
Enable multichain testing support in Solidity test runner

schaable:
This PR introduces support for multichain testing in the Solidity test runner.

#### Functional changes
- Configure predeploys for the solidity test runner.
- Allow passing `--chain-type` as an option to the `test solidity` task and validate it.
- Allow passing `--chain-type` as an option to the `test` task and forward it to `test solidity`.
- Add tests to ensure the `chainType` parameter is passed and predeploy contracts are working as expected.

#### Non-functional changes
- Refactor utilities: moved logic from `network-manager` to `internal/edr` and renamed.
- Fix ESLint `import/order` rule by adding the `"internal"` group so absolute imports from local packages are correctly ordered between external and relative imports.

Closes https://github.com/NomicFoundation/hardhat/issues/6866

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
@@ -14,7 +14,7 @@ import { getAllFilesMatching } from "@nomicfoundation/hardhat-utils/fs";
 import { resolveFromRoot } from "@nomicfoundation/hardhat-utils/path";
 import { createNonClosingWriter } from "@nomicfoundation/hardhat-utils/stream";
 
-import { isEdrSupportedChainType } from "../../edr/chain-type.js";
+import { isSupportedChainType } from "../../edr/chain-type.js";
 import { throwIfSolidityBuildFailed } from "../solidity/build-results.js";
 
 import { getEdrArtifacts, getBuildInfos } from "./edr-artifacts.js";
@@ -39,7 +39,7 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
 ) => {
   let rootFilePaths: string[];
 
-  if (!isEdrSupportedChainType(chainType)) {
+  if (!isSupportedChainType(chainType)) {
     // NOTE: We could make the error more specific here.
     throw new HardhatError(
       HardhatError.ERRORS.CORE.ARGUMENTS.INVALID_VALUE_FOR_TYPE,
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
@@ -113,6 +113,7 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
 
   const config: SolidityTestRunnerConfigArgs =
     solidityTestConfigToSolidityTestRunnerConfigArgs(
+      chainType,
       hre.config.paths.root,
       solidityTestConfig,
       grep,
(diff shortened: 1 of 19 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -38,14 +39,16 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
 ) => {
   let rootFilePaths: string[];
 
-  if (chainType !== "l1") {
-    console.log(
-      chalk.yellow(
-        `Chain type selection for tests will be implemented soon. Please check our communication channels for updates. For now, please run the task without the --chain-type option.`,
-      ),
+  if (!isSupportedChainType(chainType)) {
+    // NOTE: We could make the error more specific here.
```
alcuadrado: i think this error works

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -50,14 +55,14 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
       continue;
     }
 
-    const args = {
+    const args: TaskArguments = {
       testFiles: files,
-      noCompile: false,
       grep,
+      noCompile: subtask.options.has("noCompile"),
```
alcuadrado, line 61: I'm a bit confused about this logic. I thought we were checking if it has that option to avoid passing an unknown option. That used to throw, doesn't it?

Now, your change doesn't really change the previous logic.

alcuadrado: I double checked it. It works just by chance, because right now every subtask has that option.

alcuadrado (approved):
I left some minor comments, but it looks food

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
schaable: I just copied the error from a similar validation in the node plugin, along with the comment. I'll remove it from both.

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
schaable: Correct. Maybe we should evaluate which parameters should be considered "default" for all subtasks and which shouldn't. That said, the current setup (with `grep`, `noCompile`, and `testFiles` being required by all subtasks) seems reasonable. What I'm more curious about is how future subtasks defined in community plugins can extend the parent `test` task by adding new parameters, like we did with `chainType`, so they can be passed through properly, although that doesn't seem like a common use case.

fvictorio: But irrespective of that, we should change this, shouldn't we? I mean to something like:

```ts
    const args: TaskArguments = {
      testFiles: files,
      grep,
    };

    if (subtask.options.has("noCompile")) {
      args.noCompile = true;
    }
```

Otherwise the logic doesn't make sense to me.

fvictorio: I see now that that was the previous code :sweat_smile:

fvictorio (approved):
Pre-approving, but I think we need to fix the issue that Pato pointed out.

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
schaable: Just to clarify, the logic didn’t change compared to the previous version, it’s just written differently. I opened an issue to track the [expected parameters for test subtasks](https://github.com/NomicFoundation/hardhat/issues/6964), feel free to take a look and add any comments if needed.

Issue #6929, opened 2025-06-25
Sending more eth than account has crashes `npx hardhat node`


kanej:
If you send more eth from an account than it has, then not only does the transaction fail, but it takes down the hardhat node.
This looks like we are expecting `undefined` in the response from EDR but finding `null`.

## Reproduction steps

1. Create a new Hardhat 3 Alpha project with the viem stack

2. Add a `send-eth.ts` script:

```ts
import { network } from "hardhat";

const { viem } = await network.connect({
  chainType: "l1",
});

const publicClient = await viem.getPublicClient();
const [senderClient, recieverClient] = await viem.getWalletClients();

const TEN_THOUSAND_ETH = 10000000000000000000000n;

console.log(
  "Sending 10000 ETH from",
  senderClient.account.address,
  "to",
  recieverClient.account.address
);

const tx = await senderClient.sendTransaction({
  to: recieverClient.account.address,
  value: TEN_THOUSAND_ETH,
});

await publicClient.waitForTransactionReceipt({ hash: tx });

console.log("\nTransaction sent successfully");
```

3. Run `npx hardhat node` in one terminal
4. Run `npx hardhat run ./scripts/send-eth.ts --network localhost` in the other. You should get an error at for the cli command:

```sh
Compiling your Solidity contracts
Compiled 1 Solidity file with solc 0.8.28 (evm target: cancun)

Sending 10000 ETH from 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 to 0x70997970c51812dc3a010c7d01b50e0d17dc79c8
An unexpected error occurred:

TransactionExecutionError: An unknown RPC error occurred.

Request Arguments:
  from:   0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
  to:     0x70997970c51812dc3a010c7d01b50e0d17dc79c8
  value:  10000 ETH

Details: Failed to make POST request to http://localhost:8545
Version: viem@2.31.4
    at getTransactionError (<user-path>/node_modules/viem/utils/errors/getTransactionError.ts:44:10)
    at sendTransaction (<user-path>/node_modules/viem/actions/wallet/sendTransaction.ts:352:11)
    at async <anonymous> (<user-path>/scripts/send-eth.ts:19:12)

If you think this is a bug in Hardhat, please report it here: https://hardhat.org/report-bug
```

And the `npx hardhat node` process with crash with:

```sh
[...]

PR #6965, opened 2025-07-03
fix: node crash when sending tx with insufficient funds

schaable:
Fixes https://github.com/NomicFoundation/hardhat/issues/6929.


fvictorio:
This doesn't have a regression test. I'm ok with omitting it if it's extremely involved to do, but I don't see why that would be the case.

schaable:
> This doesn't have a regression test. I'm ok with omitting it if it's extremely involved to do, but I don't see why that would be the case.

@fvictorio added tests to cover all the scenarios in `_handleError`

fvictorio:
Using 8546 as a hardcoded port seems too brittle to me. I pushed a commit updating that test to use a random available port instead.

Issue #6875, opened 2025-06-17
Remove the `defaultNetwork` option and introduce a network called "default"


kanej:
Remove the `defaultNetwork` option from the Hardhat config. Instead we will have a new "default" network.

The default `--network` value should be `"default"`.

ChristopherDedominici:
Issue to re-enable the disabled ignition tests: https://github.com/NomicFoundation/hardhat/issues/6984

PR #6974, opened 2025-07-08
Remove the defaultNetwork option and introduce a network called "default"

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/6875


v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -53,14 +56,19 @@ export async function extendUserConfig(
...
       },
-      hardhat: {
-        chainId: 31337,
-        gas: "auto",
-        gasMultiplier: 1,
-        gasPrice: "auto",
-        ...hardhatConfig,
-        type: "edr",
-      },
+      // Only add the default network if the user hasn’t already provided one
```
fvictorio: This seems like a change in logic. IMO here you just need to change `hardhat:` to `[DEFAULT_NETWORK_NAME]:`. The rest should be kept as is.

v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -53,14 +56,19 @@ export async function extendUserConfig(
...
+      ...(DEFAULT_NETWORK_NAME in networks
+        ? {}
+        : {
+            [DEFAULT_NETWORK_NAME]: {
+              chainId: 31337,
+              gas: "auto",
+              gasMultiplier: 1,
+              gasPrice: "auto",
+              ...hardhatConfig,
```
fvictorio: Maybe we rename this variable to `defaultNetworkConfig`?

v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
ChristopherDedominici: But in doing so, we do not allow the user do define their own `default` network. If a user set a `default` network, it will be override by this one every time. Is this what we want?

fvictorio: Ah, the problem is that the `hardhatConfig` variable that is used here should come from `networks.default`, not `networks.hardhat`. This needs updating.

fvictorio (approved):
New changes LGTM. There are some CI checks failing, but I'm ok with merging if we are sure those are false negatives (let's re-run them though).

.changeset/tiny-apes-tell.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Remove the `defaultNetwork` option and introduce a network called `default`
```
kanej:
```suggestion
Remove the `defaultNetwork` config option and introduce a network called `default` ([#6875](https://github.com/NomicFoundation/hardhat/issues/6875))
```

Issue #6538, opened 2025-04-04
Verify migration path from Hardhat 2 Ignition


kanej:
We need to check the changed artifact formats.

PR #6976, opened 2025-07-09
Add migrate task to ignition

zoeyTM:
partially resolves https://github.com/NomicFoundation/hardhat/issues/6538

- adds `ignition migrate <deploymentId>` task to update artifacts & build-info to v3 formats
- added logic to the beginning of most ignition tasks to check if the artifacts for the given deployment are v2, and throw if they are


v-next/hardhat-ignition/src/index.ts
```diff
@@ -140,6 +140,14 @@ const hardhatIgnitionPlugin: HardhatPlugin = {
       })
       .setAction(import.meta.resolve("./internal/tasks/track-tx.js"))
       .build(),
+    task(["ignition", "migrate"], "Migrate artifacts to the new format")
```
kanej:
```suggestion
    task(["ignition", "migrate"], "Migrate artifacts to the new Hardhat 3 format")
```

v-next/hardhat-ignition-core/src/types/artifact.ts
```diff
@@ -29,6 +42,9 @@ export interface Artifact<AbiT extends Abi = Abi> {
   deployedBytecode: string;
   linkReferences: LinkReferences;
   deployedLinkReferences: LinkReferences;
+  immutableReferences?: ImmutableReferences;
+  buildInfoId?: string;
+  inputSourceName?: string;
```
kanej, line 47: Can I confirm my understanding here.
These properties are new in HH3. We represent them as optional so that we can represent both versions of the artifact with one type?
What do you think of having two separate types and detecting during verify/migrate, and going with the HH3 type in all other usage?

v-next/hardhat-ignition-core/src/types/artifact.ts
```diff
@@ -54,7 +70,7 @@ export interface BuildInfo {
   solcVersion: string;
   solcLongVersion: string;
   input: CompilerInput;
-  output: CompilerOutput;
+  userSourceNameMap: Record<string, string>;
```
kanej, line 73: Do we no longer have the output in the build info?

v-next/hardhat-ignition/src/internal/utils/verifyArtifactsVersion.ts
```diff
@@ -0,0 +1,58 @@
...
+import {
+  exists,
+  readdir,
+  readJsonFile,
+} from "@nomicfoundation/hardhat-utils/fs";
+
+export async function verifyArtifactsVersion(
+  deploymentDir: string | undefined,
+): Promise<void> {
+  const artifactsDir = `${deploymentDir}/artifacts`;
```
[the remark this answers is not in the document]
kanej: Sorry, ignore me, this is the ignition version of artifacts.

v-next/hardhat-ignition/test/migrate/verifyArtifactsVersion.ts
```diff
@@ -0,0 +1,32 @@
...
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
+import { assertRejectsWithHardhatError } from "@nomicfoundation/hardhat-test-utils";
+import { assert } from "chai";
+
+import { verifyArtifactsVersion } from "../../src/internal/utils/verifyArtifactsVersion.js";
+
+describe("verifyArtifactsVersion", function () {
+  it("should return when the deploymentDir is undefined or nonexistent", async function () {
+    assert.isFulfilled(verifyArtifactsVersion(undefined));
+    assert.isFulfilled(verifyArtifactsVersion("nonexistent"));
```
kanej:
```suggestion
    await assert.isFulfilled(verifyArtifactsVersion(undefined));
    await assert.isFulfilled(verifyArtifactsVersion("nonexistent"));
```
I don't know that this is strictly necessary - but lets prefer explicit.

v-next/hardhat-ignition/src/internal/tasks/migrate.ts
```diff
@@ -0,0 +1,92 @@
...
+        buildInfoId: buildInfo.id,
+      };
+
+      await writeJsonFile(artifactPath, newArtifact);
+    }
+
+    if (buildInfo._format === "hh-sol-build-info-1") {
+      const userSourceNameMap: Record<string, string> = {};
+      for (const key of Object.keys(buildInfo.input.sources)) {
+        userSourceNameMap[key] = key;
```
kanej, line 102: Can you walk me through this one.
Does this work because in v2 all sourcenames were just the filenames, so a one to one mapping just works?

v-next/hardhat-ignition/test/migrate/migrate.ts
```diff
@@ -0,0 +1,64 @@
...
+    );
+    const updatedBuildInfo = await readJsonFile(
+      path.join(buildInfoPath, "ad47b6fd82fea4f651540333f2a0887a.json"),
+    );
+
+    assert.deepEqual(updatedArtifact, v3Artifact);
+    assert.deepEqual(updatedBuildInfo, v3BuildInfo);
+  });
+
+  after(async () => {
```
kanej: Lets move the `after` up to beside the `before`.

v-next/hardhat-ignition/test/migrate/migrate.ts
```diff
@@ -0,0 +1,64 @@
...
+        recursive: true,
+        force: true,
+      },
+    );
+
+    prevWorkingDir = process.cwd();
+    process.chdir(path.join(basePath, tmpProjectPath));
+  });
+
+  it("should migrate old artifacts and build-info to the new format", async function () {
```
kanej: Can you add a test for not changing a v3 deployment? If I followed the logic you have already covered it.

kanej (changes requested):
This looks good. I left a few suggestions and questions.
I will do a manual run through with one of the more complicated examples from the old Ignition repo. We should probably check with a migration of the graph as well.

v-next/hardhat-ignition-core/src/types/artifact.ts
zoeyTM: The only reason these are optional is because they are optional in the Hardhat type as well. That being said, I'm not opposed to having two separate types if we decide to go that direction.

v-next/hardhat-ignition-core/src/types/artifact.ts
zoeyTM: Correct, this is the updated version of BuildInfo from Hardhat v3.

v-next/hardhat-ignition/src/internal/tasks/migrate.ts
zoeyTM: In all honesty, I don't fully understand the build-info changes myself. This was how Pato said it should be written.

.changeset/thirty-colts-ring.md
```diff
@@ -0,0 +1,6 @@
+---
+"@nomicfoundation/ignition-core": patch
+"@nomicfoundation/hardhat-ignition": patch
+---
+
+Add migration task for HHv2 Ignition deployments ([#6538](https://github.com/NomicFoundation/hardhat/issues/6538))
```
kanej, line 6:
```suggestion
Add migration task for Hardhat v2 Ignition deployments ([#6538](https://github.com/NomicFoundation/hardhat/issues/6538))
```

v-next/hardhat-ignition/src/internal/tasks/migrate.ts
```diff
@@ -0,0 +1,96 @@
...
+        _format: "hh3-artifact-1",
+        immutableReferences:
+          output.contracts[artifact.sourceName][artifact.contractName].evm
+            .deployedBytecode.immutableReferences ?? {},
+        inputSourceName: artifact.sourceName,
+        buildInfoId: buildInfo.id,
+      };
+
+      await writeJsonFile(artifactPath, newArtifact);
+      await remove(debugFilePath);
```
kanej: This is a problem, also the following buildinfo update.
More than one artifact can depend on the same build info. So buildinfo update (and debug file removal) can only happen once all artifacts are updated.

v-next/hardhat-ignition/src/internal/tasks/migrate.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    const artifactPath = path.join(artifactsDir, filename);
+    const artifact: Artifact = await readJsonFile(artifactPath);
+
+    const debugFilePath = path.join(
+      artifactsDir,
+      `${filename.split(".json")[0]}.dbg.json`,
+    );
+    const debugFile: { buildInfo: string } = await readJsonFile(debugFilePath);
```
kanej: I ran into a problem. In the complete example, libraries don't seem to generate `dbg.json` files.

PR #6982, opened 2025-07-10
chore: update minimal supported version of node to v20

galargh:
This PR updates the minimal supported version of node to v20.

Node v18 is now end of life and so support in Hardhat has been dropped inline with our stability guarantees: https://hardhat.org/hardhat-runner/docs/reference/stability-guarantees#node.js-versions-support


packages/hardhat-core/test/internal/hardhat-network/helpers/constants.ts
```diff
@@ -47,10 +46,3 @@ export const LAST_TX_HASH_OF_10496585 = toBuffer(
 );
 
 export const NULL_BYTES_32 = toBuffer(`0x${"0".repeat(64)}`);
-
-export const FORK_TESTS_CACHE_PATH = path.join(
```
kanej: This is just an unused const right?

.changeset/sharp-views-scream.md
```diff
@@ -0,0 +1,27 @@
...
+"@nomicfoundation/hardhat-ledger": minor
+"@nomicfoundation/hardhat-verify": minor
+"@nomiclabs/hardhat-solpp": minor
+"@nomiclabs/hardhat-vyper": minor
+"hardhat": minor
+"@nomicfoundation/hardhat-viem": minor
+"@nomiclabs/hardhat-web3": minor
+---
+
+Updated the minimal supported version of node to v20
```
kanej:
```suggestion
Updated the minimal supported version of Node to v20 ([#6982](https://github.com/NomicFoundation/hardhat/pull/6982))
```

packages/hardhat-core/src/internal/core/config/config-loading.ts
```diff
@@ -34,9 +34,13 @@ export function importCsjOrEsModule(filePath: string): any {
     // The kind of error we get in these cases depends on the Node.js version.
     const node20Heuristic = e.code === "ERR_REQUIRE_ESM";
     // Also works for node v24
+    const lines = e.message.split("\n");
+    const message = lines.length > 0 ? lines[0] : "";
```
kanej, line 38: Are all of the messages we chech against in the heuristics expected to be multiline (hence the `""` default)?

packages/hardhat-core/test/builtin-tasks/compile.ts
```diff
@@ -223,12 +223,18 @@ describe("compile task", function () {
...
+                path.dirname(absolutePath)
+              )}\\${path.basename(absolutePath)}`
+            : absolutePath;
 
         await expect(
           this.env.run(TASK_COMPILE_SOLIDITY_READ_FILE, { absolutePath })
         )
           .to.be.rejectedWith(
-            `ENOENT: no such file or directory, lstat '${absolutePath}'`
+            `ENOENT: no such file or directory, open '${expectedPath}'`
```
kanej, line 237: This is a change in the way the latest versions of the support Nodes versions throw when files don't exist?

packages/hardhat-core/test/internal/hardhat-network/helpers/constants.ts
galargh: It is, but I think we might be able to restore it now that I put other fixes in place.

packages/hardhat-core/src/internal/core/config/config-loading.ts
galargh: They used to be single line, but starting with 22.17, they're multiline.

packages/hardhat-core/test/builtin-tasks/compile.ts
galargh: Not quite, I've updated the error handling logic in `packages/hardhat-core/src/builtin-tasks/compile.ts`. Previously, we were covering the original error originating from the operation we actually wanted to perform with another one coming from the operations we performed inside the catch statement.

packages/hardhat-core/test/internal/hardhat-network/helpers/constants.ts
```diff
@@ -1,6 +1,5 @@
 // reused from ethers.js
 import { Address, toBytes } from "@ethereumjs/util";
```
galargh:
```suggestion
import { Address, toBytes } from "@ethereumjs/util";
import path from "path";
```

packages/hardhat-core/test/internal/hardhat-network/helpers/constants.ts
```diff
@@ -47,10 +46,3 @@ export const LAST_TX_HASH_OF_10496585 = toBuffer(
 );
 
 export const NULL_BYTES_32 = toBuffer(`0x${"0".repeat(64)}`);
```
galargh:
```suggestion
export const NULL_BYTES_32 = toBuffer(`0x${"0".repeat(64)}`);

export const FORK_TESTS_CACHE_PATH = path.join(
  __dirname,
  "..",
  "provider",
  ".hardhat_node_test_cache"
);
```

Issue #4059, opened 2023-06-22
[hardhat-ledger] Using both accounts and ledgerAccounts in an http network doesn't work


fvictorio:
This doesn't work at the moment:

```
mainnet: {
  accounts: [privateKey],
  ledgerAccoutns: [ledgerAddress]
}
```

The reason is that http networks use the `LocalAccountsProvider`, which intercepts `eth_accounts` calls and only returns the ones that were configured in `accounts`.

The obvious solution is for that provider to call `eth_accounts` in the wrapped provider, and add that to the returned list. I'm not sure if this would be a bug fix or a breaking change though.

A less risky alternative is to do that only when `providerExtenders.length > 0`, similar to what we do here:

https://github.com/NomicFoundation/hardhat/blob/8587456b744bc2f7a0e1ce1baff5b14db6a859a6/packages/hardhat-core/src/internal/core/providers/construction.ts#L246

koonopek:
Will be fixed? Hardhat ledger is unusable now

Kinco-dev:
Any updates ? Still unusable now

Jnrlouis:
Will this be fixed please

kanej:
We are addressing this issue as part of Hardhat 3. I will ping here when the capability becomes available in the Hardhat 3 Alpha:

https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%403.0.0-next.0

michalbrabec:
Review state in Hardhat 3. Close in case it is resolved by V3 or communicate state clearly (in docs) in case we will not resolve it.

ChristopherDedominici:
This will be fixed once PR [#6983](https://github.com/NomicFoundation/hardhat/pull/6983) is merged.

ChristopherDedominici:
Fixed in https://github.com/NomicFoundation/hardhat/pull/6983

PR #6983, opened 2025-07-10
Hardhat-ledger v3

ChristopherDedominici:
Ports `hardhat-ledger` to Hardhat 3.

`hardhat-ledger` will not be included in the toolboxes, but will provide a key deployment option when running in Hardhat scripts. Hardhat Ignition support is not included in this PR but is coming soon.

Link to [internal design doc](https://www.notion.so/nomicfoundation/hardhat-ledger-20e578cdeaf580d59402da8a38e1f9a2)

Fixes: https://github.com/NomicFoundation/hardhat/issues/5646

## TODO after MERGE
- [x] Write documentation: https://github.com/NomicFoundation/hardhat-website/issues/96
- [ ] Improve UI: https://github.com/NomicFoundation/hardhat/issues/7576

## TODO

- [x] Port code to v3
- [x] Port tests to v3
- [x] Resolve manual test failure
- [x] Add `README.md`
- [x] Resolve question of exporting core functions (are they going to utils and when?) - https://github.com/NomicFoundation/hardhat/pull/7281
- [x] Resolve `node-hid` overrides question: synch needed - need second eye opinion https://github.com/NomicFoundation/hardhat/pull/6983#discussion_r2225623969
- [x] Resolve `npx hardhat node` failure: https://github.com/NomicFoundation/hardhat/pull/7106
- [x] Remove this change (node-hid) before merge: https://github.com/NomicFoundation/hardhat/pull/6983/files#r2225623969
- [x] Check that v3 handles the scenarios addressed in [this PR](https://github.com/NomicFoundation/hardhat/pull/6398)
- [x] Manual testing with a `hardhat-ledger`

## Related V2 bugs fixed in V3
- [x] [Using both accounts and ledgerAccounts in an http network doesn't work](https://github.com/NomicFoundation/hardhat/issues/4059) is now fixed, as shown in [this test](https://github.com/NomicFoundation/hardhat/blob/0ceac9ec8718fb57a9b7c9776d89f63ce70827a0/v-next/hardhat-ledger/test/integration.ts#L42)
- [x] [Hardhat-ledger plugin error using blockchains that don't support EIP-1559 transaction format](https://github.com/NomicFoundation/hardhat/issues/5786) is now fixed, as shown in the [example project](https://github.com/NomicFoundation/hardhat/blob/0ceac9ec8718fb57a9b7c9776d89f63ce70827a0/v-next/example-project/scripts/ledger.ts#L42)

### DIFF
--- a/.changeset/purple-wasps-jump.md
+++ b/.changeset/purple-wasps-jump.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Validate initialBaseFeePerGas against hardfork only for L1 chain type
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
@@ -346,7 +346,7 @@ function refineEdrNetworkUserConfig(
     }
 
     const resolvedHardfork = hardfork ?? getCurrentHardfork(chainType);
-    if (chainType === L1_CHAIN_TYPE || chainType === GENERIC_CHAIN_TYPE) {
+    if (chainType === L1_CHAIN_TYPE) {
       if (hardforkGte(resolvedHardfork, L1HardforkName.LONDON, chainType)) {
         if (minGasPrice !== undefined) {
           ctx.addIssue({
--- a/v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
@@ -2190,13 +2190,21 @@ describe("NetworkManagerImplementation", () => {
           validationErrors = await validateNetworkUserConfig(
             edrConfig({ initialBaseFeePerGas: 123n }),
           );
+          assertValidationErrors(validationErrors, []);
 
+          validationErrors = await validateNetworkUserConfig(
+            edrConfig({ initialBaseFeePerGas: 123, hardfork: "berlin" }),
+          );
           assertValidationErrors(validationErrors, []);
         });
 
         it("should not validate an invalid network config", async () => {
           let validationErrors = await validateNetworkUserConfig(
-            edrConfig({ initialBaseFeePerGas: 123, hardfork: "berlin" }),
+            edrConfig({
+              chainType: L1_CHAIN_TYPE,
+              initialBaseFeePerGas: 123,
+              hardfork: "berlin",
+            }),
           );
 
           assertValidationErrors(validationErrors, [
@@ -2323,7 +2331,7 @@ describe("NetworkManagerImplementation", () => {
 
         it("should not validate an invalid network config", async () => {
           let validationErrors = await validateNetworkUserConfig(
-            edrConfig({ minGasPrice: 123 }),
+            edrConfig({ chainType: L1_CHAIN_TYPE, minGasPrice: 123 }),
           );
 
           assertValidationErrors(validationErrors, [


package.json
```diff
@@ -16,5 +16,10 @@
     "lint": "pnpm run --recursive --if-present lint",
     "prelint:fix": "pnpm build",
     "lint:fix": "pnpm run --recursive --if-present lint:fix"
+  },
```
ChristopherDedominici, line 21: I'm not entirely sure how to address this. This is a temporary workaround to allow me to use the plugin with a physical Ledger device.
Without this fix, I get the following error when trying to connect to the Ledger:

```
Error HHE90002 in plugin hardhat-ledger: There was an error trying to establish a connection to the Ledger wallet: Error: Could not locate the bindings file. Tried:
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/build/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/build/Debug/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/build/Release/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/out/Debug/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/Debug/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/out/Release/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/Release/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/build/default/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/compiled/24.2.0/linux/x64/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/addon-build/release/install-root/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/addon-build/debug/install-root/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/addon-build/default/install-root/HID_hidraw.node
 → /home/chris/repos/nomic-foundation/hardhat4/node_modules/.pnpm/node-hid@2.1.2/node_modules/node-hid/lib/binding/node-v137-linux-x64/HID_hidraw.node.

Make sure your Ledger is connected and unlocked, and that the Ethereum app is open.
```

v-next/hardhat/src/utils/rpc.ts
```diff
@@ -0,0 +1,14 @@
+export { rpcAddress } from "../internal/builtin-plugins/network-manager/rpc/types/address.js";
```
ChristopherDedominici: This is a temporary file I created to export the core logic methods, RPC types, and the mutex needed in the plugin.
Should we consider officially exporting these features from this package, or would it make more sense to move some of them to `hh-utils`?

v-next/hardhat-ledger/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { PLUGIN_NAME } from "./internal/plugin-name.js";
+
+const hardhatLedgerPlugin: HardhatPlugin = {
+  id: PLUGIN_NAME,
+  hookHandlers: {
+    config: import.meta.resolve("./internal/hook-handlers/config.js"),
+    network: import.meta.resolve("./internal/hook-handlers/network.js"),
+  },
+  tasks: [emptyTask("ledger", "Interact with Ledger hardware wallets").build()],
```
kanej: I think we can drop this? There are no tasks in the v2 version.

v-next/hardhat/src/utils/rpc.ts
kanej: @schaable what do you think of these exported functions - I don't have a strong sense. Should these be exported from `core` or moved to `hardhat-utils`?

v-next/hardhat-ledger/src/internal/config/resolution.ts
```diff
@@ -0,0 +1,23 @@
+import type { HardhatUserConfig } from "hardhat/config";
+import type { HardhatConfig } from "hardhat/types/config";
+
+export function resolveLedgerUserConfig(
```
kanej, line 4: Can we rename the file to `resolve-ledger-user-config.ts` to match the single function?
I would do the same for `validate-ledger-user-config.ts`.
A more generic name makes sense if there is a collection of functions, but I would be explicit if only one is ever going to be exported.

v-next/hardhat-ledger/src/internal/handler.ts
```diff
@@ -0,0 +1,607 @@
...
+        }),
+      );
+    }
+
+    assertHardhatInvariant(
+      this.#chainId !== undefined,
+      "chainId should be defined",
+    );
+
+    const unsignedTx = createTx(txRequest, this.#chainId);
```
ChristopherDedominici, line 562: Creation of TX + signing

v-next/hardhat-ledger/src/internal/handler.ts
```diff
@@ -0,0 +1,607 @@
...
+    displayMessage: (interruptor: string, message: string) => Promise<void>,
+    customConfig?: {
+      // Allows passing a custom config, primarily used for testing
+      ethConstructor?: typeof Eth.default;
+      transportNodeHid?: typeof TransportNodeHid.default;
+      cachePath?: string;
+    },
+  ) {
+    this.#ethConstructor = customConfig?.ethConstructor ?? Eth.default;
+    this.#transportNodeHid =
```
kanej, line 72: Can you talk me through this?

In the previous implementation the transportNodeHid was initialized with:

```ts
const transport = await TransportNodeHid.create(
  openTimeout,
  connectionTimeout
);
this._eth = wrapTransport(transport);
```
Now we are using `TransportNodeHid.default`. I looked into this because I was wondering if there was cleanup we should do in `closeConnection` of the network hook handler.

So my three questions are:

1. Have I understood correctly that we will use the default TransportNodeHid across all network connections?
2. It doesn't make sense to have one transport per network connection as there is only one ledger.
3. There is no cleanup that is needed, when the plugin is garbage collected that is it.

v-next/hardhat-ledger/src/internal/hook-handlers/network.ts
```diff
@@ -0,0 +1,120 @@
...
+        jsonRpcRequest.method === "eth_chainId" ||
+        jsonRpcRequest.method === "eth_getTransactionCount" ||
+        jsonRpcRequest.method === "eth_sendRawTransaction"
+      ) {
+        // Allow these methods to pass through untouched.
+        // The ledger handler calls them directly, so intercepting them here would lead to infinite recursion.
+        return next(context, networkConnection, jsonRpcRequest);
+      }
+
+      const ledgerHandler = await initializationMutex.exclusiveRun(async () => {
```
ChristopherDedominici, line 45: For every connection, a new ledgerHandler is created

v-next/hardhat-ledger/src/internal/handler.ts
ChristopherDedominici: See [here](https://github.com/NomicFoundation/hardhat/pull/6983/files#r2228634719) to better understand:

in the `network onRequest hook handler`, for every new connection, a new ledger class is instantiated. So every class will have its own `TransportNodeHid` (so you think one is enough and it can be shared?).

We still do `TransportNodeHid.create`. But since we are ESM the import has to be done via `TransportNodeHid.default`

In the `network closeConnection hook handlers` closed connection are cleaned. 

Btw, with these comments you made me realize I could add a test to cover multiple connections at the same time

v-next/hardhat-ledger/src/internal/handler.ts
```diff
@@ -0,0 +1,607 @@
...
+      );
+
+      try {
+        return await this.#eth.signEIP712Message(path, typedMessage);
+      } catch (_error) {
+        return this.#eth.signEIP712HashedMessage(
+          path,
+          domainHash,
+
+          structHash,
```
kanej:
```suggestion
          path,
          domainHash,
          structHash,
```

v-next/hardhat-ledger/src/internal/handler.ts
```diff
@@ -0,0 +1,606 @@
...
+
+      return this.#eth.signTransaction(path, txToSign, resolution);
+    });
+
+    const signedTx = new Transaction(unsignedTx.type, {
+      ...unsignedTx.raw,
+      r: toBigInt(normalizeHexString(signature.r)),
+      s: toBigInt(normalizeHexString(signature.s)),
+      yParity: hexStringToNumber(normalizeHexString(signature.v)) % 2,
+    }).toHex();
```
kanej, line 582: I don't know how to judge the changes required for the move to `micro-eth-signer`.

package.json
ChristopherDedominici: Link to `@ledgerhq/hw-transport-node-hid` dependencies: https://github.com/LedgerHQ/ledger-live/blob/3853940b4365c21a544e707313add106803111a5/libs/ledgerjs/packages/hw-transport-node-hid/package.json#L36C5-L36C25

v-next/hardhat-ledger/src/index.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6983/commits/b5b1eb1fd9bfc7c2af2c9c7197150d3b57f2d1b7

v-next/hardhat-ledger/src/internal/config/resolution.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6983/commits/b5b1eb1fd9bfc7c2af2c9c7197150d3b57f2d1b7

v-next/hardhat-ledger/src/internal/handler.ts
```diff
@@ -0,0 +1,603 @@
...
+
+    /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    -- A type assertion is necessary because there is no type overlap between the `domain` imported from `@ledgerhq`
+    and the parameter type expected by the function imported from `micro-eth-signer`. */
+    const enc = typed.encoder(types, domain as any);
+
+    /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    -- A type assertion is necessary because there is no type overlap between the `domain` imported from `@ledgerhq`
+    and the parameter type expected by the function imported from `micro-eth-signer`. */
+    const domainHash = enc.structHash("EIP712Domain", domain as any);
```
ChristopherDedominici, line 467: I need a double check on this logic. Is every hash struct a `EIP712Domain`?

v-next/hardhat/src/utils/rpc.ts
kanej: Pato weighed in, because any further changes would be breaking we should move these to `hardhat-utils`.

.changeset/shiny-pigs-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ledger": patch
```
fvictorio: Do we need to add `hardhat-errors` here too?

package.json
michalbrabec: We decided to:
- complete hardhat ledger port for HH3
- not include it by default with HH3 toolboxes
- document in the ledger docs that the option `--allow-build=node-hid` is necessary to use it
https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1757505683993459

.changeset/shiny-pigs-give.md
```diff
@@ -0,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-errors": patch
+"@nomicfoundation/hardhat-ledger": patch
+---
+
+Add the `@nomicfoundation/hardhat-ledger` plugin ([#5646](https://github.com/NomicFoundation/hardhat/issues/5646))
```
kanej:
```suggestion
Ported the `@nomicfoundation/hardhat-ledger` plugin to Hardhat 3 ([#5646](https://github.com/NomicFoundation/hardhat/issues/5646))
```

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -2782,4 +2800,72 @@ export default { lib1: "0x...", lib2: "0x...", ... };
...
+
+Make sure your Ledger is connected and unlocked, and that the Ethereum app is open.`,
+        websiteTitle: "Ledger connection error",
+        websiteDescription: `There was an error trying to establish a connection to the Ledger wallet.
+
+Make sure your Ledger is connected and unlocked, and that the Ethereum app is open.`,
+      },
+      ERROR_WHILE_DERIVING_PATH: {
+        number: 90003,
+        messageTemplate: `There was an error trying to derivate path "{path}": "{message}". The wallet might be connected but locked or in the wrong app.`,
```
kanej:
```suggestion
        messageTemplate: `There was an error trying to derive path "{path}": "{message}". The wallet might be connected but locked or in the wrong app.`,
```

v-next/hardhat-ledger/src/internal/get-y-parity.ts
```diff
@@ -0,0 +1,17 @@
+import { assertHardhatInvariant } from "@nomicfoundation/hardhat-errors";
+
+export function getYParity(v: number): number {
```
kanej, line 17: I think this could use a doc comment, just to help orientate someone reading the code.

v-next/hardhat-ledger/src/internal/get-y-parity.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (v === 27 || v === 28) {
+    return v - 27;
+  }
+
+  if (v === 0 || v === 1) {
+    return v;
+  }
+
+  if (v >= 35n) {
```
kanej:
```suggestion
  if (v >= 35) {
```
The comparison will work with a bigint, but I think given the signature works in terms of `number` we should keep the comparisons on the same types.

v-next/hardhat-ledger/src/internal/hook-handlers/network.ts
```diff
@@ -0,0 +1,117 @@
...
+export default async (): Promise<Partial<NetworkHooks>> => {
+  // This map is essential for managing multiple network connections in Hardhat V3.
+  // Since Hardhat V3 supports multiple connections, we use this map to track each one
+  // and associate it with the corresponding handlers array.
+  // When a connection is closed, its associated handler is removed from the map.
+  // See the "closeConnection" function at the end of the file for more details.
+  const ledgerHandlerPerConnection: WeakMap<
+    NetworkConnection<ChainType | string>,
+    LedgerHandler
+  > = new Map();
```
kanej:
```suggestion
  const ledgerHandlerPerConnection: Map<
    NetworkConnection<ChainType | string>,
    LedgerHandler
  > = new Map();
```
I think we are just using it as a map, as we need to do explicit resource cleanup. Or do you think both type and instance should be `WeakMap`?

v-next/hardhat-ledger/src/internal/create-tx.ts
```diff
@@ -0,0 +1,102 @@
...
+import { assertHardhatInvariant } from "@nomicfoundation/hardhat-errors";
+import { toBigInt } from "@nomicfoundation/hardhat-utils/bigint";
+import {
+  bytesToBigInt,
+  bytesToHexString,
+  bytesToNumber,
+} from "@nomicfoundation/hardhat-utils/bytes";
+import { addr, Transaction } from "micro-eth-signer";
+
+const STRICT_MODE = false;
```
kanej, line 15: What is the trade-off here? What was it that forced us into non-strict mode?

v-next/hardhat-ledger/src/internal/create-tx.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export function createTx(
+  txRequest: RpcTransactionRequest,
+  chainId: bigint,
+): Transaction<"eip7702" | "eip1559" | "eip2930" | "legacy"> {
+  const baseTxParams = {
+    to:
+      txRequest.to !== undefined && txRequest.to !== null
+        ? bytesToHexString(txRequest.to)
+        : "0x",
```
kanej: Is "0x" the best default here?

In the local accounts handler `to` is would be set something like:

```ts
const checksummedAddress = addr.addChecksum(
  bytesToHexString(txRequest.to ?? new Uint8Array()),
  true,
);

const baseTxParams = {
  to: checksummedAddress,
  nonce: txRequest.nonce !== undefined ? toBigInt(txRequest.nonce) : 0n,
  chainId,
  value: txRequest.value ?? 0n,
  data: bytesToHexString(txRequest.data ?? new Uint8Array()),
};
```

Is "0x" the equivalent?

v-next/hardhat-ledger/src/internal/create-tx.ts
```diff
@@ -0,0 +1,102 @@
...
+    nonce: txRequest.nonce !== undefined ? toBigInt(txRequest.nonce) : 0n,
```
kanej: In local accounts signer we have an assertion ensuring that the nonce is passed in.
I think we should do that here. Defaulting to the 0th nonce for signing doesn't seem correct.

v-next/hardhat-ledger/src/internal/hook-handlers/network.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        assertHardhatInvariant(
+          Array.isArray(accountsResponse.result) &&
+            accountsResponse.result.every((v) => typeof v === "string"),
+          "accountsResponse.result should be an array and every element should be a string",
+        );
+
+        otherAccounts = accountsResponse.result;
+      }
```
kanej: Can you talk me through this one?

My reading is that this can call `next()` twice, once immediately within the `if` and the second at the bottom of the overarching function?

Is the intent to get the `eth_accounts` and include them within within the signing before sending the request onto the next handler?

v-next/hardhat-ledger/src/internal/hook-handlers/network.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      const newRequestOrResponse = await ledgerHandler.handle(
```
kanej: Can I check, when would this be a response (rather than a request)?

v-next/hardhat-ledger/test/internal/handler.ts
```diff
@@ -0,0 +1,813 @@
...
+        const request = createJsonRpcRequest("eth_sendTransaction", [
+          {
+            from: account.address,
+            to: LEDGER_ADDRESSES[1],
+            value: numberToHexString(100),
+            gas: numberToHexString(1000000),
+            gasPrice: numberToHexString(100),
+            gasLimit: numberToHexString(1000000),
+          },
+        ]);
```
kanej: When I alter the `gas` field the test still passes. I am not sure that all of the gas related fields are going through.

We should double check that the signed transaction isn't just ending up with the defaults from `micro-eth-signer`.

Can we add another test here, so one for:
* pre-eip 1559 transactions with the appropriate gas fields
* post-eip 1559 transactions with teh appropriate gas fields

v-next/hardhat-ledger/src/internal/handler.ts
kanej: My reading is that the encoder gets setup with the domain struct and that each of the hashes created uses that including the hashing of the domain itself.

What do you think? Do we need to pull in more expertise here?

kanej (changes requested):
Hey, I have done another round of review. I think this PR is looking good, and has good test coverage.
I have left a bunch of questions, that I think we should work through. I particularly want to better understand the hook handler logic on request preparation and `eth_accounts`.
We may want to pull in some more expertise for some of the signing. My understanding here is we have ported across `hardhat-ledger` code by taking the local account signer as the base?
In particular I want us to double check (and add some more tests) to confirm that the signed transaction is reflecting the expected gas fields for eip-1559 style transactions and pre-eip-1559 style transactions.

So I think the remaining todos:

- [ ] Address each open conversation
- [ ] Add tests and confirm gas fields are coming through
- [ ] Manual testing [John]

v-next/hardhat-ledger/src/internal/create-tx.ts
ChristopherDedominici: Same approach as we do in the `local-accounts`: https://github.com/NomicFoundation/hardhat/pull/6644

v-next/hardhat-ledger/src/internal/get-y-parity.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6983/commits/6b5211cd3ec168a643841f34566d0a04a6370b97

v-next/hardhat-ledger/src/internal/create-tx.ts
ChristopherDedominici: Good question! For the `0x` address there is no difference, but for the other addresses is a good additional check. I updated the code to follow the same logic here: https://github.com/NomicFoundation/hardhat/pull/6983/commits/db90d3b185381e9d09f250480a3536d97ecd6411

v-next/hardhat-ledger/src/internal/create-tx.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6983/commits/5c7ded48057832db8cec4903ffc8a99593c4c1ed

v-next/hardhat-ledger/src/internal/hook-handlers/network.ts
ChristopherDedominici: Yes, this is a special case when `eth_accounts` is called.
If the request was handled only by the `ledger plugin`, it would return only the ledger accounts.
The idea, therefore, is to first gather all the other accounts and then pass them to the `ledger plugin`, so they can be combined with the ledger accounts.

Take a look at the `handle` function: when the method is `eth_accounts`, you can see that the `otherAccounts` from the `network.ts` file are combined with the ledger accounts.

image

v-next/hardhat-ledger/src/internal/hook-handlers/network.ts
ChristopherDedominici: Response:
- for example, when collecting accounts(`eth_accounts`), the handle function returns the accounts that can be immediately returned to the user

Request:
- when the requested method is not handled by the `ledger plugin`, the original request must still be propagated

- for example, with `eth_sendTransaction`: the payload is modified by the ledger plugin, but the request still needs to be sent to the network

v-next/hardhat-ledger/src/internal/hook-handlers/network.ts
ChristopherDedominici: We use the same logic in the `handlers`: https://github.com/NomicFoundation/hardhat/blob/d22f9dd3543a9762b8cb438fe436795267f7551a/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/network.ts#L23

I think that having a `WeakMap` is a safe net to clean up resources if for any reason `closeConnection` is not properly called. See old comment: https://github.com/NomicFoundation/hardhat/pull/5994#discussion_r1871977938

kanej: Cool, then we should update `new Map();` to be `new WeakMap();`. The type is `WeakMap` but the implementation is a standard map.

ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6983/commits/5f59a872da958a0f838cd46aca4a03f867b50aa3

v-next/hardhat-ledger/test/internal/handler.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6983/commits/f480b357204b7414e01c684770bb6e76c6b8f090

I also modified the code to throw an error for type 4 transactions (EIP-7702), since they are currently not supported by the Ledger library. Ledger does have a new library that supports them, but it is still under development and may introduce breaking changes.
See here: https://github.com/NomicFoundation/hardhat/issues/7273

v-next/hardhat-ledger/src/internal/create-tx.ts
```diff
@@ -0,0 +1,114 @@
...
+    "nonce should be defined",
+  );
+
+  const baseTxParams = {
+    to: checksummedAddress,
+    nonce: toBigInt(txRequest.nonce),
+    chainId,
+    value: txRequest.value ?? 0n,
+    data: bytesToHexString(txRequest.data ?? new Uint8Array()),
+    gasLimit: txRequest.gasLimit,
```
kanej: The [eth_sendtransaction](https://ethereum.org/developers/docs/apis/json-rpc/#eth_sendtransaction) spec suggests that the field should be `gas`, but `eth-micro-signer` expects `gasLimit`.
As the provider operates at the level of the RPC API, I think we should expect `txRequest.gas` to be coming through. Indeed we have a check in `handle` to ensure that `txRequest.gas` exists.
So I think we need to do something like this:
https://github.com/NomicFoundation/hardhat/blob/a1157d6fd2971a8a083d275b8a9f27efe3918269/v-next/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts#L294

I would suggest something like: `gasLimit: transactionRequest.gasLimit ??transactionRequest.gas,`

We should check the tests so that gas/gasLimit are set and the output asserted.

kanej: Based on `geth` the RPC expects `gas` but will ignore `gasLimit`.
So we should:
* only accept `gas` at the RPC level when passing in the request
* map `gas` to `gasLimit` when passing off to the signing library (as local accounts does)
* ignore `gasLimit` as passed in at the RPC level

ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6983/commits/69ef31b0a62a20f0a5e3f34df4208fb03b6ba2f1

v-next/hardhat-ledger/src/internal/hook-handlers/network.ts
kanej: Ah okay, can I confirm then.
In the handler, if it is a `eth_accounts` call, we make the full request then and there with `next`, we then pass off the returned results to the Ledger handler so they can be combined with the ledger accounts?

If that is the case, I think we should pull the ledger handle code you are showing above up so that the `eth_accounts` code in the onRequest handler can be self contained (everything in the `if` for `eth_accounts` including the return statement).

kanej:
I have been able to run manual tests on windows for the new plugin, including:

* signing in a script against an in-memory instance (the example project script)
* running signed transactions against a hardhat node
* deploying with ignition against a Hardhat node

I think we likely need to iterate on the UI, but I think we can do that in a separate PR:

```
Batch NomicFoundation/hardhat#1
  Executing Apollo#Rocket...

[hardhat-ledger] Connecting to Ledger...
[hardhat-ledger] Connection successful
[hardhat-ledger] Confirmation start
[hardhat-ledger] Confirmation success
```

v-next/hardhat-ledger/src/internal/hook-handlers/network.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/6983/commits/0ebd4b7d03232a782bec9504f3117a92795029bc

michalbrabec:
Confirmed by @fvictorio

PR #6987, opened 2025-07-11
Partial compilation

### LINKED ISSUE #6897: Partial recompilation
Port the partial recompilation feature from Hardhat 2.

antico5:
- Compile cache has been swapped with a single-file table with metadata about previous file compilation. Compile input/output is stored on build info files and is not required on an extra cache
- The cache strategy involves building a compilation job for each root file and hashing it. Across compilations, if anything that may change the compilation job (file contents, solc config, sourcename) changes, the file is recompiled.
- Emitted files are stored in the cache file as well, so in case any is removed, the file is also marked for compilation
- The 'mergeCompilationJobs' parameter has been changed to 'isolated', which has the opposite effect. It has been added as an argument to the compile task

Closes #6897

### DIFF
--- a/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-three-files-A-B,C.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-three-files-A-B,C.ts
@@ -141,7 +141,7 @@ describe("Partial compilation", () => {
         assert.ok(firstSnapshot.typeFiles["Bar.sol"] !== undefined);
 
         // Recompile
-        await project.compile();
+        await project.compile({ isolated: true });
         const secondSnapshot = await project.getSnapshot();
 
         // Nothing in the snapshot should have changed
--- a/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-three-files-A-B--C.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-three-files-A-B--C.ts
@@ -145,7 +145,7 @@ describe("Partial compilation", () => {
         assert.ok(firstSnapshot.typeFiles["FooFactory.sol"] !== undefined);
 
         // Recompile
-        await project.compile();
+        await project.compile({ isolated: true });
         const secondSnapshot = await project.getSnapshot();
 
         // Nothing in the snapshot should have changed
--- a/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-three-files-A-B-C.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-three-files-A-B-C.ts
@@ -141,7 +141,7 @@ describe("Partial compilation", () => {
         assert.ok(firstSnapshot.typeFiles["IFoo.sol"] !== undefined);
 
         // Recompile
-        await project.compile();
+        await project.compile({ isolated: true });
         const secondSnapshot = await project.getSnapshot();
 
         // Nothing in the snapshot should have changed
--- a/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-two-connected-files.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-two-connected-files.ts
@@ -120,7 +120,7 @@ describe("Partial compilation", () => {
         assert.ok(firstSnapshot.typeFiles["IFoo.sol"] !== undefined);
 
         // Recompile
-        await project.compile();
+        await project.compile({ isolated: true });
         const secondSnapshot = await project.getSnapshot();
 
         // Nothing in the snapshot should have changed
--- a/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-two-indept-files.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/from-scratch-two-indept-files.ts
@@ -119,7 +119,7 @@ describe("Partial compilation", () => {
         assert.ok(firstSnapshot.typeFiles["B.sol"] !== undefined);
 
         // Recompile
-        await project.compile();
+        await project.compile({ isolated: true });
 
         const secondSnapshot = await project.getSnapshot();
 


.changeset/unlucky-monkeys-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+"hardhat": patch
+---
+
+Minimize compilation jobs based on changes in files and config across runs
```
fvictorio, line 6:
```suggestion
Added support for partial recompilation
```

v-next/hardhat-verify/src/internal/artifacts.ts
```diff
@@ -76,14 +76,22 @@ export async function getCompilerInput(
   sourceName: string,
   buildProfileName: string,
 ): Promise<CompilerInput> {
-  const compilationJob = await solidity.getCompilationJobs(
+  const getCompilationJobsResult = await solidity.getCompilationJobs(
     [path.join(rootFilePath, sourceName)],
     {
       buildProfile: buildProfileName,
       quiet: true,
+      force: true,
```
fvictorio, line 84: What's the reason for this?

v-next/hardhat-verify/src/internal/artifacts.ts
```diff
@@ -76,14 +76,22 @@ export async function getCompilerInput(
...
     },
   );
 
+  assertHardhatInvariant(
+    !("reason" in getCompilationJobsResult),
+    "getCompilationJobs should not error",
+  );
```
fvictorio, line 91: My understanding of this line is: if a user has some configuration that results in a `CompilationJobCreationError` (for example, they changed the solidity version but forgot to update some pragma) and they run `verify`, this will be thrown.

If that's correct: I don't think we should do it that way. `assertHardhatInvariant` is used for things that "should never happen", not for user errors, no matter how unlikely they are.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -32,11 +35,13 @@ export class CompilationJobImplementation implements CompilationJob {
     solcConfig: SolcConfig,
     solcLongVersion: string,
     hooks: HookManager,
+    contentHashes: Map<string, string> = new Map(),
   ) {
     this.dependencyGraph = dependencyGraph;
     this.solcConfig = solcConfig;
     this.solcLongVersion = solcLongVersion;
     this.#hooks = hooks;
+    this.#contentHashes = contentHashes;
```
fvictorio: I'm slightly concerned by the fact that this is meant to be shared between jobs, but it's not super obvious that that's the case. It took me a little while to understand why the map was received instead of being created by the constructor, until I realized the same map was passed to each instance and I remembered why that was done that way.

As far as I can tell, this is not tested, and I understand is not an easy thing to test, because it's an internal implementation detail of the build system.

I thought of using a singleton static property as the content hashes cache, but I think that's a dangerous idea: someone could modify a file and re-compile in the same process, and the cached hash would be used, which would be wrong.

So this is a long-winded way of saying "let's at least document this". And maybe we can rename it to `this.#sharedContentHashes` or something like that?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -199,17 +204,36 @@ export class CompilationJobImplementation implements CompilationJob {
     // remove it once the code coverage was added because it simplified the
     // implementation and because we expect the caching logic to change.
     const solcInput = await this.getSolcInput();
+    const smallerSolcInput = { ...solcInput };
+
+    smallerSolcInput.sources = Object.fromEntries(
```
fvictorio, line 206:
```suggestion
    // we replace the solidity code of each source by its cached content hash
    // to minimize the amount of hashing we have to do
    smallerSolcInput.sources = Object.fromEntries(
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/cache.ts
```diff
@@ -1,92 +1,43 @@
...
-/**
- * ObjectCache is a simple cache that stores objects as JSON files. It uses the
- * stream versions of the file system access functions to support large objects,
- * i.e. files exceeding the maximum string length.
- */
-export class ObjectCache<T> {
-  readonly #path: string;
-  readonly #defaultMaxAgeMs: number;
-  readonly #defaultMaxSize: number;
+export type CompileCache = Record<string, CompileCacheEntry>;
```
fvictorio, line 15: Let's add a comment explaining what the keys of this map are. From looking at the code, I think they are user source names (not input source names, nor paths), which makes sense, but I'm not 100% sure that's the case.

v-next/hardhat/src/types/solidity/build-system.ts
```diff
@@ -165,6 +165,18 @@ export interface FailedFileBuildResult {
   errors: CompilerOutputError[];
 }
 
+export interface GetCompilationJobsResult {
+  compilationJobsPerFile: Map<string, CompilationJob>;
+  indexedIndividualJobs: Map<string, CompilationJob>;
+}
+
+export interface EmitArtifactsResult {
+  artifactsPerFile: ReadonlyMap<string, string[]>;
```
fvictorio, line 174: Why is this one `Readonly` and `typeFilePaths` is not? I don't really care either way, just curious about the discrepancy.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -344,10 +335,97 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
+      }),
+    );
+
+    // Load the cache
+    this.#compileCache = await loadCache(this.#options.cachePath);
+
+    // Select which files to compile
+    const rootFilesToCompile: Set<string> = new Set();
+
+    for (const [rootFile, compilationJob] of indexedIndividualJobs.entries()) {
```
fvictorio, line 391: I thought this could've been done inside a `Promise.all` instead of a for-loop, but I changed it and it actually made it slightly slower. Performance is hard.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -344,10 +335,97 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
+      const jobHash = await compilationJob.getBuildId();
+      const cacheResult = this.#compileCache[rootFile];
+
+      // If there's no cache for the root file, or the compilation job changed, or using force flag, compile it
+      if (
+        (options?.force ?? false) ||
```
fvictorio: Most of our code uses this idiom for `boolean | undefined` things. I also find it a bit more readable tbh.

```suggestion
        (options?.force === true) ||
```

fvictorio (commented):
I reviewed the code, not yet the tests. I also want to:

- Think if we should we have a `--isolate` flag in the compile task. I'm pretty sure that we shouldn't. Instead,  we should add `isolate` as a build profile property and remove the `shouldMergeCompilationJobs` function. But I need to look into this a bit more to be sure.
- Run some manual tests to see if I can break the proposed approach. I hope I can't, because I really like how simple it is.

fvictorio:
Ah, one general comment more.

I think we need to be extra clear and tell apart individual compilation jobs vs "actual" compilation jobs, everywhere. In general I'd rather have more verbose names here if they make it clear what kind of entity it's being manipulated. It's the same approach we took with user/input source names, and I think it works.

(Alternatively, we could try to do this at the type level, but that's probably overkill.)

I'm not sure what's the right word for this though:

- **merged** is misleading because there is no actual merging if you run in isolate mode.
- **actual** is a meaningless word, and it only makes sense during this transition period from the previous approach.
- **solc** compilation job might work, because the defining characteristic of non-individual compilation jobs is that they are executed. But I don't quite like it.

But I think any prefix is going to be better than having non-qualified `compilationJob` names.

v-next/hardhat-verify/src/internal/artifacts.ts
antico5: The issue was that when the file was already compiled, then getCompilationJobs() wasn't returning a job for the file, so it wasn't able to get the solc input for verification purposes. I think we can either leave it like this, or access the compilation job from the other return value (indexedIndividualJobs), which is always populated.

v-next/hardhat/src/types/solidity/build-system.ts
antico5: The `artifactsPerFile` was pre-existent with the `ReadonlyMap` type. I didn't even notice to be honest, on the implementation they are both `Map`. I changed the type so now both are `ReadonlyMap` for consistency

v-next/hardhat-verify/src/internal/artifacts.ts
fvictorio: Ah, no, this makes sense, I didn't realize this was about getting the compilation jobs, not actually compiling.

fvictorio:
I found a problem with the current approach. It's very easy to reproduce:

1. Create a project with two contracts
2. Compile them in non-isolated mode
3. Compile them again in isolated mode

In step 2, you get a single build info that emits artifacts for both files. After step 3, you'd expect that build info to be deleted, and two separate new build infos to be created and used to emit the artifacts. Instead, nothing happens.

This makes sense: the build id of the individual compilation jobs didn't change, because they don't have any information of having been merged or not (by design).

I'm not sure what's the right way to fix this. One simple idea is to include the `isolated` flag as part of the config, so that it's used to invalidate the cache. But maybe there are better approaches.

---

Another thing I noticed, way less important but still something we should fix: when you run the compile task and there's nothing to compile, the output says "Compiling your Solidity contracts" instead of something like HH2's "Nothing to compile".

fvictorio:
Btw, notice that the problem I mentioned is only important when you go from non-isolated mode to isolated mode. The same thing happens in the other direction, but that's fine (and actually desirable, because it reduces the number of unnecessary compilations). That being said, if the fix causes a recompilation when going from isolated to non-isolated mode, but it's easy to implement, I'd be fine with it because I don't expect that scenario to happen often.

Going from non-isolated mode to isolated mode, though also unlikely to happen, has a way bigger impact: you are running in isolated mode for reproducibility, so ignoring it is very bad.

v-next/hardhat-verify/src/internal/artifacts.ts
antico5: I agree, I just left it how it was already implemented (just modified it since the return type on getCompilationJobs changed). Should I change it to just throw the error ?

antico5:
> ybe th

Huge catch. I think adding the isolated flag to the metadata should fix it

v-next/hardhat-verify/src/internal/artifacts.ts
fvictorio: Ah, sorry, my bad, I didn't realize this was the existing behavior. Out of scope for this PR then.

PR #7006, opened 2025-07-16
skip bytecode reconciliation for successfully deployed contract futures

zoeyTM:
second PR that resolves https://github.com/NomicFoundation/hardhat/issues/6538

partner PR is https://github.com/NomicFoundation/hardhat/pull/6976, but they don't depend on each other and can be merged in any order


alcuadrado:
I think we should also backport this to HH2, so that the semantics of both versions are aligned.

kanej (approved):
Looks good. I did a local test and confirmed that reconciliation has been loosened for successfully deployed contracts.

PR #7022, opened 2025-07-18
Fix to update ignition network connection

kanej:
Invocations of `npx hardhat ignition deploy ./modules/Counter.ts --network localhost` was not passing the network through to the default connection. The same was happening when running scripts.

This PR re-enables the `--network` global option.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
@@ -1,6 +1,8 @@
 import type { HardhatRuntimeEnvironmentHooks } from "../../../../types/hooks.js";
 import type { NetworkManager } from "../../../../types/network.js";
 
+import { DEFAULT_NETWORK_NAME } from "../../../constants.js";
+
 export default async (): Promise<Partial<HardhatRuntimeEnvironmentHooks>> => ({
   created: async (context, hre) => {
     let networkManager: NetworkManager | undefined;
@@ -15,6 +17,9 @@ export default async (): Promise<Partial<HardhatRuntimeEnvironmentHooks>> => ({
 
         if (networkManager === undefined) {
           networkManager = new NetworkManagerImplementation(
+            hre.globalOptions.network !== undefined
+              ? hre.globalOptions.network
+              : DEFAULT_NETWORK_NAME,
             hre.config.defaultChainType,
             hre.config.networks,
             context.hooks,
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
@@ -24,7 +24,6 @@ import { HardhatError } from "@nomicfoundation/hardhat-errors";
 import { readBinaryFile } from "@nomicfoundation/hardhat-utils/fs";
 import { deepMerge } from "@nomicfoundation/hardhat-utils/lang";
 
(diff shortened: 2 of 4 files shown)


v-next/hardhat/package.json
```diff
@@ -85,7 +85,7 @@
     "typescript": "~5.8.0"
   },
   "dependencies": {
-    "@ignored/edr-optimism": "0.13.0-alpha.7",
```
kanej, line 88: This was a last minute addition to the Alpha that I am reverting as it breaks one of the default scripts when running under optimism.

v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
```diff
@@ -15,6 +17,9 @@ export default async (): Promise<Partial<HardhatRuntimeEnvironmentHooks>> => ({
 
         if (networkManager === undefined) {
           networkManager = new NetworkManagerImplementation(
+            hre.globalOptions.network !== undefined
+              ? hre.globalOptions.network
+              : DEFAULT_NETWORK_NAME,
```
kanej, line 22: Before the changes to the default network this was:
```
hre.globalOptions.network !== undefined
              ? hre.globalOptions.network
              : hre.config.defaultNetwork,
```
The default network PR removed the globalOptions along with the `hre.config.defaultNetwork`, which is the root cause of the bug.

https://github.com/NomicFoundation/hardhat/pull/6974/files#diff-03173af1ae628b27bdff9793dc3e766d0a0a537dd3cd74c19c00f5934261e5bc

v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -157,9 +144,9 @@ describe("NetworkManagerImplementation", () => {
   describe("connect", () => {
     it("should connect to the default network and chain type if none are provided", async () => {
       const networkConnection = await networkManager.connect();
-      assert.equal(networkConnection.networkName, "default");
+      assert.equal(networkConnection.networkName, "localhost");
       assert.equal(networkConnection.chainType, GENERIC_CHAIN_TYPE);
-      assert.deepEqual(networkConnection.networkConfig, networks.default);
+      assert.deepEqual(networkConnection.networkConfig, networks.localhost);
```
kanej, line 149: These changes are a straight revert of the default network PR now that we are passing the network in as a parameter as before.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -100,7 +102,7 @@ export class NetworkManagerImplementation implements NetworkManager {
     chainType?: ChainTypeT,
     networkConfigOverride?: NetworkConfigOverride,
   ): Promise<NetworkConnection<ChainTypeT>> {
-    const resolvedNetworkName = networkName ?? DEFAULT_NETWORK_NAME;
+    const resolvedNetworkName = networkName ?? this.#defaultNetwork;
```
kanej, line 105: All of the changes in this file are a reversion to before the default network PR.

The default network PR moved the default fallback from the higher level `hre` handler that calls here and moved it to down.

Instead we re-introduce the `hre` handler as the place where the default network is determined. If it is passed at the cli as `--network`, that is used otherwise the default network is set to the `DEFAULT_NETWORK_NAME` constant. The default network can then be passed into the constructor of the network manager as before.

schaable (approved):

PR #7029, opened 2025-07-21
chore: update CI to test against Node 20.0.0

kanej:
Revert the `micro-eth-signer` dependency bump to support Node 20 versions before Node 20.19.

This PR also updates our CI to test against both Node 20.0.0 and Node 20 latest.

### DIFF
--- a/.changeset/honest-steaks-fail.md
+++ b/.changeset/honest-steaks-fail.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Revert bump of `micro-eth-signer` version back to `^0.14.0` ([#7026](https://github.com/NomicFoundation/hardhat/issues/7026))


.github/workflows/hardhat-ci.yml
```diff
@@ -71,16 +71,16 @@ jobs:
       matrix:
         package: ${{ fromJson(needs.list-packages.outputs.packages) }}
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node: [20, 22, 24]
+        node: [20.0.0, 20, 22, 24]
```
kanej: I am unsure whether we should go for the minimum version of each major node version as well?

```suggestion
        node: [20.0.0, 20, 22.0.0, 22, 24.0.0, 24]
```

packages/hardhat-core/package.json
```diff
@@ -120,7 +120,7 @@
     "json-stream-stringify": "^3.1.4",
     "keccak": "^3.0.2",
     "lodash": "^4.17.11",
-    "micro-eth-signer": "^0.16.0",
```
kanej, line 123: This is the reversion that gets us back to Node `20.0.0` support.

Issue #7028, opened 2025-07-21
Error on `--coverage` with Node Test runner


kanej:
On initializing a new viem based Hardhat 3 Alpha (`v3.0.0-next.23`), running the test command for js/ts tests with coverage fails:

```sh
> npx hardhat test nodejs --coverage

Compiling your Solidity contracts
Compiled 1 Solidity file with solc 0.8.28 (evm target: cancun)

Running node:test tests

node:internal/test_runner/harness:117
      throw err;
      ^

HardhatError: HHE100: An internal invariant was violated: Flags should never accept values
    at parseArgumentValue (/Users/userx/repos/nomic/samples/next-viem/node_modules/hardhat/src/internal/core/arguments.ts:188:13)
    at resolveGlobalOptions (/Users/userx/repos/nomic/samples/next-viem/node_modules/hardhat/src/internal/core/global-options.ts:174:23)
    at HardhatRuntimeEnvironmentImplementation.create (/Users/userx/repos/nomic/samples/next-viem/node_modules/hardhat/src/internal/core/hre.ts:118:27)
    at async getOrCreateGlobalHardhatRuntimeEnvironment (/Users/userx/repos/nomic/samples/next-viem/node_modules/hardhat/src/internal/hre-initialization.ts:92:15)
    at async <anonymous> (/Users/userx/repos/nomic/samples/next-viem/node_modules/hardhat/src/index.ts:19:3)

Node.js v24.4.1

  1) /Users/userx/repos/nomic/samples/next-viem/test/Counter.ts (454ms)


0 passing (458ms)
1 failing

  1) /Users/userx/repos/nomic/samples/next-viem/test/Counter.ts:

   Test file execution failed (exit code 7).

| File Path 📦          | Line % 📈 | Statement % 📈 | Uncovered Lines 🔍 | Partially Covered Lines 🔍 |
| --------------------- | --------- | -------------- | ------------------ | -------------------------- |
| contracts/Counter.sol | 0.00      | 0.00           | 10-11, 15-17       | -                          |
| --------------------- | --------- | -------------- | ------------------ | -------------------------- |
| Total                 | 0.00      | 0.00           |                    |                            |
```

PR #7030, opened 2025-07-21
fix: global flag option parsing

galargh:

---

This is the most straighforward fix for https://github.com/NomicFoundation/hardhat/issues/7028

In an usual case, we do not expect flag options to be associated with values. However, that's not always the case in our codebase. In particular, we're parsing global options differently to task options. This is because global option values can be also provided via the environmental variables. In particular, `--coverage` global flag option can be provided via `HARDHAT_COVERAGE` environmental variable. When that happens, we should know how to parse its' value (exactly the same as we do for boolean options). This is why we have to remove the invariant of never expecting flags to be associated with values from the `parseArgumentValue` function.

### DIFF
--- a/.changeset/spicy-ravens-type.md
+++ b/.changeset/spicy-ravens-type.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Fixed global flag option parsing
+Fixed global flag option parsing ([#7028](https://github.com/NomicFoundation/hardhat/issues/7028))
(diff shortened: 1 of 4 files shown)


.changeset/spicy-ravens-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed global flag option parsing
```
kanej:
```suggestion
Fixed global flag option parsing ([#7028](https://github.com/NomicFoundation/hardhat/issues/7028))
```

kanej (approved):
Looks good - I did a manual test against the example project.

PR #7036, opened 2025-07-21
Rename dependencies to compile and build task

### LINKED ISSUE #6877: Rename `dependenciesToCompile`
We need a better name for the config option under the build system. We are often asked about this feature, suggesting that `dependenciesToCompile` is not self-documenting.

antico5:
This PR renames the `dependenciesToCompile` config property to `npmFilesToBuild`.

Also it introduces a new task, `build`, that is just an alias for `compile`. It uses the same arguments and task action as `compile`

Closes #6877

### DIFF
--- a/.changeset/cold-oranges-guess.md
+++ b/.changeset/cold-oranges-guess.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Rename dependenciesToCompile to npmFilesToBuild. Add `build` task that is an alias to `compile`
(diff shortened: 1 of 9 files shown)


.changeset/cold-oranges-guess.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Rename dependenciesToCompile to npmFilesToBuild. Add `build` task that is an alias to `compile`
```
kanej:
```suggestion
Rename `dependenciesToCompile` to `npmFilesToBuild` in Hardhat config ([#6877](https://github.com/NomicFoundation/hardhat/issues/6877))
Add `build` task that replaces `compile`, but keep the `compile` task as an alias to `build` ([#6877](https://github.com/NomicFoundation/hardhat/issues/6877))
```

PR #7037, opened 2025-07-21
Support verifying contracts on all enabled providers from the main verify task

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7007.

This PR also updates how the enabled field in the config is handled, based on [this Slack conversation](https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1752771501519749). In short: the enabled flag is now only used by the main verify task. It is ignored when a provider is invoked directly, either via its specific task or programmatically.

Additionally, this fixes a small type issue we encountered during ETHGlobal Cannes, where the `apiKey` was incorrectly marked as required when Etherscan verification was disabled. The `apiKey` is still validated if Etherscan is invoked directly, so calling it without an API key will correctly throw a "missing API key" error.

Example screenshots:
verify-success-error
verify-success-success
verify-error-success
verify-warning


.changeset/old-wolves-rhyme.md
```diff
@@ -0,0 +1,7 @@
+---
+"@nomicfoundation/hardhat-errors": patch
+"@nomicfoundation/hardhat-verify": patch
+"hardhat": patch
+---
+
+Support verifying contracts on all enabled providers from the main verify task
```
kanej:
```suggestion
Support verifying contracts on all enabled providers from the main verify task ([#7007](https://github.com/NomicFoundation/hardhat/issues/7007))
```

fvictorio (approved):

PR #7045, opened 2025-07-23
fix: change deployment folder rule for ignition

### LINKED ISSUE #6999: Update deployment directory rules based on network type
In v2 the deployment is recorded to files on disk if:

* explicitly set with a flag
* or if the network being connected to is not named hardhat

We should (though confirm before making the change) take into account that different network configs can now fulfill the role of the one in-memory connection we had previously.

The suggestion is that we should not write to disk if the connection type is `EDR` - so in-memory. We should also print the warning that the user is running against an in-memory connection:

```
You are running Hardhat Ignition against an in-process instance of Hardhat Network.
This will execute the deployment, but the results will be lost.
You can use --network <network-name> to deploy to a different network.
```

kanej:
In Hardhat 2, the deployment folder is not generated if running against an in-memory network or `writeLocalhostDeployment` is enabled. The rule for in-memory network is detected was based on the name (i.e. "hardhat").

The rule is now changing to be based on whether the network running against is of type `edr`, which indicates that the network is running in-memory in Hardhat 3.

The rule is updated in the `deploy` task, as well of the helper objects in the viem and ethers Ignition plugins.

Resolves #6999.

### DIFF
--- a/.changeset/sour-pandas-attend.md
+++ b/.changeset/sour-pandas-attend.md
@@ -1,0 +1,7 @@
+---
+"@nomicfoundation/hardhat-ignition-ethers": patch
+"@nomicfoundation/hardhat-ignition-viem": patch
+"@nomicfoundation/hardhat-ignition": patch
+---
+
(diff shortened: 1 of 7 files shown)


zoeyTM (approved):
LGTM

Issue #7010, opened 2025-07-16
Report EDR stack trace generation errors to Sentry


kanej:
Currently EDR stack trace generation errors are not reported to Sentry, only logged to the user. We want these to flow through to our Sentry exception reporting.

See the design doc: https://www.notion.so/nomicfoundation/Report-stack-trace-generation-errors-to-Sentry-22c578cdeaf5803a9655e889f99a0322

PR #7046, opened 2025-07-23
feat: report stack trace generation errors to sentry

galargh:

---

This resolves https://github.com/NomicFoundation/hardhat/issues/7010

It adds a new error type for reporting stack trace generation errors.

I tested it by forcing https://github.com/NomicFoundation/hardhat/blob/v-next/v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts#L183 to be of kind `UnexpectedError`, enabling telemetry in my local environment, and hardcoding the response from https://github.com/NomicFoundation/hardhat/blob/v-next/v-next/hardhat/src/internal/cli/telemetry/sentry/subprocess.ts#L50 to be `true`. 

This resulted in the following events to be reported to Sentry: https://nomic-labs.sentry.io/issues/6765223811/?project=4508780138856448&referrer=project-issue-stream

### DIFF
--- a/.changeset/fifty-roses-refuse.md
+++ b/.changeset/fifty-roses-refuse.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Report stack trace generation errors to Sentry
+Report stack trace generation errors to Sentry ([#7010](https://github.com/NomicFoundation/hardhat/issues/7010))


.changeset/fifty-roses-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Report stack trace generation errors to Sentry
```
kanej:
```suggestion
Report stack trace generation errors to Sentry ([#7010](https://github.com/NomicFoundation/hardhat/issues/7010))
```

agostbiro:
Thanks for this PR @galargh ! It looks good overall, I just have one question:

Is it guaranteed that the new error type will get reported to Sentry? There seems to be two checks that it needs to pass:

https://github.com/NomicFoundation/hardhat/blob/6ebaa3a441759a295eeaf6ce520d018c71bb6531/v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts#L124

https://github.com/NomicFoundation/hardhat/blob/10f554d5fa45643eda3e7cee7060e57a8019bb92/v-next/hardhat/src/internal/cli/telemetry/sentry/subprocess.ts#L50

(For the latter you mentioned that you hardcoded the response which I'm not sure how to interpret)

galargh:
@agostbiro Both are fine. `shouldBeReported` only blocks a particular subset of errors that this new one does not fall under, and `raisedByHardhat` should be truthy as well since we create these new errors in Hardhat.

agostbiro (approved):
Awesome, thanks!

PR #7054, opened 2025-07-23
fix: correctly await unsubscribe requests when WebSocket closes

reallesee:
Replace forEach with Promise.all + map pattern when unsubscribing from subscriptions on WebSocket close. This ensures that all unsubscribe operations are properly awaited before the event handler completes, preventing potential race conditions and resource leaks

Using forEach with an async callback doesn't properly wait for the promises to resolve, which could lead to incomplete cleanup of WebSocket subscriptions


kanej:

PR #7061, opened 2025-07-24
chore: remove `enableRip7212` from EDR config

### LINKED ISSUE #6182: Remove fields from EDR network config
Remove the `enableRip7212` and `enableTransientStorage` fields from the EDR network config, as these were introduced for specific needs in the past and are no longer required. Additionally, remove all related logic from the network manager.

kanej:
This was needed in the past but is no longer required.

Resolves #6182.


v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -491,8 +490,6 @@ async function getProviderConfig(
       codeCoverage: coverageConfig,
     },
     ownedAccounts: ownedAccounts.map((account) => account.secretKey),
-    precompileOverrides: networkConfig.enableRip7212
-      ? [precompileP256Verify()]
-      : [],
```
kanej, line 496: My assumption is that the precompile is now handled with the hardfork.

v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
```diff
@@ -367,25 +364,6 @@ function refineEdrNetworkUserConfig(
...
-      }
-      if (
-        hardforkGte(resolvedHardfork, L1HardforkName.CANCUN, chainType) &&
-        enableTransientStorage === false
-      ) {
-        ctx.addIssue({
-          code: z.ZodIssueCode.custom,
-          message: `'enableTransientStorage' must be enabled for hardforks 'cancun' or later. To disable this feature, use a hardfork before 'cancun'.`,
-        });
-      }
```
kanej, line 388: These checks are no longer necessary because transient storage is handled at the hardfork level within EDR.

v-next/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts
```diff
@@ -276,7 +271,6 @@ export async function resolveChainDescriptors(
 export function resolveHardfork(
   hardfork: string | undefined,
   chainType: ChainType | undefined = GENERIC_CHAIN_TYPE,
-  _enableTransientStorage: boolean | undefined,
```
kanej, line 279: This is a safe non-breaking change as we are internal here.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
Wodann: I think this is scheduled for inclusion in the next hardfork (Fusaka). [[source](https://ethereum-magicians.org/t/eip-7212-precompiled-for-secp256r1-curve-support/14789/123)]

schaable (approved):

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
Wodann: Since we don't support Fusaka yet in EDR, is there an alternative mechanism for Hardhat users to opt in to "custom precompiles"?

PR #7064, opened 2025-07-24
Support predicates on viem-assertions

### LINKED ISSUE #7038: Support predicates in `hardhat-viem-matchers`
This is a follow on from the discussion in #6761.

In `hardhat-viem-matchers` we want to add something similar to our ethers chai matchers:

* Things like emitWithArgs and revertWithCustomErrorWithArgs accept either a value or a predicate, where the predicate returns a boolean indicating if the argument matches what is expected.
* We also offer some built-in matchers, like anyValue, whose implementation is just () => true.

antico5:
Closes #7038 

Predicates support has been added to `emitWithArgs` and `revertWithCustomErrorWithArgs`.

The assertion failures when using predicates to match are not displayed the same way as when passing primitive values. This is because for the latter we leverage assert.deepEqual, which shows the diff in the following way:
```
   AssertionError: The event arguments do not match the expected ones.
   - Expected
   + Received
   
     Array [
   -   2n,
   +   1n,
     ]
```

When using predicates and the assertion fails, we have to build a custom message and it looks like this:
```
   AssertionError: The event arguments do not match the expected ones:
   Expected: ["<predicate>"]
   Emitted: ["1"]
```

### DIFF
--- a/.changeset/lovely-dodos-tell.md
+++ b/.changeset/lovely-dodos-tell.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-viem-assertions": patch
+---
+
+Support predicates on viem-assertions ([#7038](https://github.com/NomicFoundation/hardhat/issues/7038))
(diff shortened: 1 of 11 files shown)


fvictorio (approved):
I manually tested this a bit and LGTM. I haven't reviewed the code, so let's wait for @ChristopherDedominici's review to merge.

kanej:
Can we add minimal docs on this to the readme?

PR #7073, opened 2025-07-28
Add null check to _isInsideSameDir method

strmfos:
Fix a potential issue in the resolver's _isInsideSameDir method by adding a null check for sourceNameToTest parameter. This prevents potential TypeErrors when the method is called with undefined or null values, improving the robustness of the code without changing existing behavior

### DIFF
--- a/.changeset/ten-gorillas-bathe.md
+++ b/.changeset/ten-gorillas-bathe.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Make the resolver more robust


packages/hardhat-core/src/internal/solidity/resolver.ts
```diff
@@ -474,7 +474,7 @@ export class Resolver {
   }
 
   private _getUriScheme(s: string): string | undefined {
-    const re = /([a-zA-Z]+):\/\//;
+    const re = /^([a-zA-Z]+):\/\//;
```
kanej: Can you explain to me this change? It seems unrelated.

strmfos: You're right, this change to the regex was unnecessary. I'll remove it and keep only the null check, thank you

Issue #7065, opened 2025-07-24
MultipleMatchingNetworksError (Base): Multiple networks with chain id 8453 found.


nenadmitt:
### Version of Hardhat

2.26.0

### What happened?

Failed to run a deployment script on base chain with hardhat, got an error:

```
MultipleMatchingNetworksError: Multiple networks with chain id 8453 found. You can override the chain by passing it as a parameter to the client getter:
```

Dependencies:

```{
  "name": "hardhat-project",
  "devDependencies": {
    "@nomicfoundation/hardhat-toolbox-viem": "^4.1.0",
    "hardhat": "^2.26.0"
  },
  "dependencies": {
    "@openzeppelin/contracts": "^5.4.0",
    "@uniswap/v4-core": "^1.0.2",
    "@uniswap/v4-periphery": "^1.0.2",
    "viem": "^2.33.0"
  }
}
```

### Minimal reproduction steps

Configured a base chain in my hardhat.config.ts

```
  networks: {
    sepolia: {
      url: RPC_URL,
      chainId: sepolia.id,
      accounts: [WALLET_KEY],
    },
    based: {
      url: RPC_URL,
      chainId: base.id,
      accounts: [WALLET_KEY],
    }
  },
```

Tried to run a deploy contract via script

`
npx hardhat run scripts/deploy.ts --network base
`

The script is simple:

```
  const [walletClient] = await hre.viem.getWalletClients();
  const publicClient = await hre.viem.getPublicClient();

  const permit2 = "0x000000000022D473030F116dDEE9F6B43aC78BA3"; address
  const positionManager = "0x7c5f5a4bbd8fd63184577525326123b519429bdc";  

  const v4Deployer = await hre.viem.deployContract("UniswapV4Deployer", [
    permit2,
    positionManager,
  ],
  });
```

Got an error:

```
MultipleMatchingNetworksError: Multiple networks with chain id 8453 found. You can override the chain by passing it as a parameter to the client getter:

import { someChain } from "viem/chains";
const client = await hre.viem.getPublicClient({
  chain: someChain,
  ...
});

You can find a list of supported networks here: https://github.com/wevm/viem/blob/main/src/chains/index.ts
```

The workaround is to manually add chain in the script:

```
  const [walletClient] = await hre.viem.getWalletClients({chain: base});
  const publicClient = await hre.viem.getPublicClient({chain: base});

  const permit2 = "0x000000000022D473030F116dDEE9F6B43aC78BA3"; address
[...]

PR #7076, opened 2025-07-28
Fixed "Multiple networks with chain id found" error when multiple chains share the same chain id

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7065

### DIFF
--- a/.changeset/fuzzy-peas-occur.md
+++ b/.changeset/fuzzy-peas-occur.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/hardhat-viem": patch
 ---
 
-Fixed "Multiple networks with chain id found" error when multiple chains share the same chain id
+Fixed "Multiple networks with chain id found" error when multiple chains share the same chain id ([#7065](https://github.com/NomicFoundation/hardhat/issues/7065))
--- a/.changeset/fuzzy-peas-occur.md
+++ b/.changeset/fuzzy-peas-occur.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-viem": patch
+---
+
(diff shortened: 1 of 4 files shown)


packages/hardhat-viem/test/chains.ts
```diff
@@ -52,41 +52,41 @@ describe("chains", () => {
       expect(chain).to.deep.equal(chains.foundry);
     });
 
-    it("should throw if the chain id is 31337 and the network is neither hardhat nor foundry", async () => {
```
schaable, line 55: This diff is hard to read. For reference, I changed the last test from `"should throw if the chain id is not 31337 and there are multiple chains with that id"` to `"should return the first matching chain if the chain id is not 31337 and there are multiple chains with that id"` and moved it to the top to keep the error tests grouped at the bottom.

.changeset/fuzzy-peas-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-viem": patch
+---
+
+Fixed "Multiple networks with chain id found" error when multiple chains share the same chain id
```
kanej:
```suggestion
Fixed "Multiple networks with chain id found" error when multiple chains share the same chain id ([#7065](https://github.com/NomicFoundation/hardhat/issues/7065))
```

.changeset/fuzzy-peas-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-viem": patch
```
kanej, line 2: This change means that where previously we would error now we provide a value based on `viem`'s own selection logic.
Therefore this is not a breaking change.

kanej (approved):
Looks good.

PR #7077, opened 2025-07-28
New solidity tests reporter

fvictorio:


v-next/hardhat/src/internal/builtin-plugins/solidity-test/formatters.ts
```diff
@@ -19,11 +29,15 @@ export function formatArtifactId(
   const sourceName =
     sourceNameToUserSourceName.get(artifactId.source) ?? artifactId.source;
 
-  return `${chalk.bold(`${sourceName}:${artifactId.name}`)} (v${artifactId.solcVersion})`;
+  return `${sourceName}:${artifactId.name}`;
```
galargh, line 32: Is it intentional to drop solc version information? I remember it being useful in debugging but might not be needed for everyday usecases.

galargh (approved):
🥇

v-next/hardhat/src/internal/builtin-plugins/solidity-test/formatters.ts
fvictorio: It felt too verbose to me, but I wouldn't be against adding it again if there's a good reason.

galargh: OK, let's remove it then. We can always add it back when we find a good reason for it.

Issue #7039, opened 2025-07-22
Sending with chaintype optimism fails


kanej:
With the upgrade to EDR Optimism alpha 7: "@ignored/edr-optimism": "0.13.0-alpha.7", we see a failure in the `./scripts/send-op-tx.ts` script included in the `viem` template project.

```ts
import { network } from "hardhat";

const { viem } = await network.connect({
  network: "hardhatOp",
  chainType: "optimism",
});

console.log("Sending transaction using the OP chain type");

const publicClient = await viem.getPublicClient();
const [senderClient] = await viem.getWalletClients();

// console.log("Sending 1 wei from", senderClient.account.address, "to itself");

// const l1Gas = await publicClient.estimateL1Gas({
//   account: senderClient.account.address,
//   to: senderClient.account.address,
//   value: 1n,
// });

// console.log("Estimated L1 gas:", l1Gas);

console.log("Sending L2 transaction");
const tx = await senderClient.sendTransaction({
  to: senderClient.account.address,
  value: 1n,
});

await publicClient.waitForTransactionReceipt({ hash: tx });

console.log("Transaction sent successfully");
```

This leads to an error on running `pnpm hardhat run ./scripts/send-op-tx.ts` of:

```
Compiling your Solidity contracts
Compiled 1 Solidity file with solc 0.8.28 (evm target: cancun)

Sending transaction using the OP chain type
Sending L2 transaction

thread 'tokio-runtime-worker' panicked at /Users/runner/work/edr/edr/crates/edr_op/src/block/builder.rs:113:30:
Unsupported base fee params version: 18. Expected 0.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command was killed with SIGABRT (Aborted): hardhat run ./scripts/send-op-tx.ts
```

The bump to edr in the Hardhat 3 Alpha was reverted - so this is not reproducible with the current Hardhat 3 Alpha

kanej:
We should add an integration test that covers this case so we don't regress.

PR #7019, opened 2025-07-18
[WIP] feat: enable code coverage for solidity tests

galargh:
- [ ] Because this PR includes a **bug fix**, relevant tests have been included.
- [ ] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.
- [x] I didn't do anything of this.

---

Should providing the `onCollectedCoverageData` callback through the observability config be enough to enable code coverage for solidity tests? The callback doesn't seem to be called in the current state of this branch.

```
{
  ...,
  observability: {
    codeCoverage: {
      onCollectedCoverageCallback: [AsyncFunction: onCollectedCoverageCallback]
    }
  },
  ...
}
```

PR #7080, opened 2025-07-29
build: upgrade to @nomicfoundation/edr v0.12.0-next.1

Wodann:

---

I upgrade the dependency for EDR to `@nomicfoundation/edr v0.12.0-next.1`, with the following changes:

### Minor Changes

- ffd2deb: Added value snapshots to Solidity test runner using gas & value cheatcodes

### Patch Changes

- adbba3d: Fixed mining of local blocks for OP

See [release notes](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.1).

### DIFF
--- a/v-next/hardhat/package.json
+++ b/v-next/hardhat/package.json
@@ -85,7 +85,7 @@
     "typescript": "~5.8.0"
   },
   "dependencies": {
-    "@ignored/edr-optimism": "0.13.0-alpha.6",
+    "@nomicfoundation/edr": "0.12.0-next.1",
     "@nomicfoundation/hardhat-errors": "workspace:^3.0.0-next.25",
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.0-next.25",
     "@nomicfoundation/hardhat-zod-utils": "workspace:^3.0.0-next.25",
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/types/coverage.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/types/coverage.ts
@@ -1,3 +1,3 @@
 export interface CoverageConfig {
-  onCollectedCoverageCallback: (coverageData: Uint8Array[]) => void;
+  onCollectedCoverageCallback: (coverageData: Uint8Array[]) => Promise<void>;
 }
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
@@ -226,13 +226,13 @@ export class NetworkManagerImplementation implements NetworkManager {
         );
         if (shouldEnableCoverage) {
           coverageConfig = {
-            onCollectedCoverageCallback: (coverageData: Uint8Array[]) => {
+            onCollectedCoverageCallback: async (coverageData: Uint8Array[]) => {
               // NOTE: We cast the tag we receive from EDR to a hex string to
               // make it easier to debug.
               const tags = coverageData.map((tag) =>
                 Buffer.from(tag).toString("hex"),
               );
-              void hookManager.runParallelHandlers(
+              await hookManager.runParallelHandlers(
                 "network",
                 "onCoverageData",
                 [tags],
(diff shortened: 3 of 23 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/types/coverage.ts
```diff
@@ -1,3 +1,3 @@
 export interface CoverageConfig {
-  onCollectedCoverageCallback: (coverageData: Uint8Array[]) => void;
+  onCollectedCoverageCallback: (coverageData: Uint8Array[]) => Promise<void>;
```
Wodann, line 2: We switched to `async` callbacks, in line with @galargh's  draft PR: https://github.com/NomicFoundation/hardhat/pull/7019/files#diff-bfee97552980f539bd76afaa6d6007702c031e1123e065d4649f7413316383acR136

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -226,13 +226,13 @@ export class NetworkManagerImplementation implements NetworkManager {
...
           coverageConfig = {
-            onCollectedCoverageCallback: (coverageData: Uint8Array[]) => {
+            onCollectedCoverageCallback: async (coverageData: Uint8Array[]) => {
               // NOTE: We cast the tag we receive from EDR to a hex string to
               // make it easier to debug.
               const tags = coverageData.map((tag) =>
                 Buffer.from(tag).toString("hex"),
               );
-              void hookManager.runParallelHandlers(
+              await hookManager.runParallelHandlers(
```
Wodann, line 235: This function returns a `Promise`, so I figured it should be `await`ed

v-next/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts
```diff
@@ -62,7 +62,7 @@ export async function* testReporter(
           const name = testResult.name;
           const status: TestStatus = testResult.status;
           const details = [
-            ["duration", `${testResult.durationMs} ms`],
+            ["duration", `${testResult.durationNs} ns`],
```
Wodann, line 65: We now return `ns` instead of `ms`. I replaced all usages of `ms` that I found with `ns`.

v-next/hardhat/package.json
```diff
@@ -85,7 +85,7 @@
     "typescript": "~5.8.0"
   },
   "dependencies": {
-    "@ignored/edr-optimism": "0.13.0-alpha.6",
+    "@nomicfoundation/edr": "0.12.0-next.1",
```
Wodann, line 88: We released a pre-releaes under the `@nomicfoundation/edr` package

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
galargh: If we're awaiting this, we should probably add a try catch so that errors are not thrown (as per the relevant comment in the config declaration). I'll add a try-catch here if you don't mind.

galargh (approved):
I verified that https://github.com/NomicFoundation/hardhat/issues/7039 is resolved here. Thank you 🙇

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
Wodann: Are you referring to EDR's config? I updated EDR to bubble up errors returned in this config.

Just to make sure, is this a long background process? This will block the EDR thread from progressing as we make it a blocking call internally.

galargh: Yes, exactly! Oh, I see now! I must have been looking at the outdated version.

Running parallel handlers should be relatively quick. We just add the data we receive to a collection and that's it. We defer any processing to a later stage.

I'm going to revert my commit then. The reason being, when someone enables coverage, it would be better for the issues with coverage collection to be obvious rather than hidden.

Wodann: Sounds good; thanks!

PR #7094, opened 2025-07-30
Add a development keystore

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/7003


v-next/hardhat-keystore/src/index.ts
```diff
@@ -31,10 +31,20 @@ const hardhatKeystorePlugin: HardhatPlugin = {
         name: "force",
         description: "Forces overwrite if the key already exists.",
       })
+      .addFlag({
```
ChristopherDedominici, line 34: I added a flag to every single task. Is there a better way to define the flag just once for the entire plugin?
I’d prefer to avoid hacks like object spread (...{ }).
If that’s the only alternative, I’d rather be a bit more verbose but stay consistent with the plugin guidelines.

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -2035,6 +2035,23 @@ Please check Hardhat's output for more details.`,
...
       },
+      CANNOT_CHANGED_PASSWORD_FOR_DEV_KEYSTORE: {
+        number: 50001,
+        messageTemplate:
+          "Cannot change password for dev keystore, as it is auto generated.",
+        websiteTitle: "Cannot change password for dev keystore",
+        websiteDescription:
+          "Cannot change password for dev keystore, as it is auto generated.",
+      },
+      KEY_NOT_FOUND_DURING_TESTS_WITH_DEV_KEYSTORE: {
```
ChristopherDedominici, line 2057: I'd like a double check on this message? Is it too long or it is descriptive enough?

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -24,6 +24,9 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
   { testFiles, chainType, grep, noCompile, verbosity },
   hre,
 ) => {
+  // Set an environment variable that plugins can use to detect when a process is running tests
+  process.env.HH_TEST = "true";
```
ChristopherDedominici: The idea is that by setting this environment variable, the keystore plugin can easily detect when Hardhat is running in test mode, allowing me to enforce the exclusive use of the development keystore

v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
```diff
@@ -29,34 +32,91 @@ export default async (): Promise<Partial<ConfigurationVariableHooks>> => {
...
 
-      if (!(await keystore.hasKey(variable.name, masterKey))) {
-        return next(context, variable);
+    if (!(await keystoreLoader.isKeystoreInitialized())) {
+      return undefined;
+    }
+
+    const keystore = await keystoreLoader.loadKeystore();
+
+    if (masterKey === undefined) {
```
ChristopherDedominici, line 83: I'm wondering if a mutex is required here. The code when executed manually it does work, but if multiple requests arrive in parallel?

v-next/hardhat-keystore/src/index.ts
fvictorio: This makes sense to me. I wish there was some way to just set it once at the level of the main `keystore` task, but I don't think that's possible right now.

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -2043,6 +2043,23 @@ Please check Hardhat's output for more details.`,
         websiteDescription:
           "The password you provided is incorrect or the keystore file is corrupted.",
       },
+      CANNOT_CHANGED_PASSWORD_FOR_DEV_KEYSTORE: {
+        number: 50001,
+        messageTemplate:
+          "Cannot change password for dev keystore, as it is auto generated.",
```
fvictorio: Mentioning that it's auto-generated means exposing an implementation detail, and the user probably doesn't care about this. Maybe let's just say:

```
The `keystore change-password` task cannot be used with the development keystore.
```

v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
fvictorio: I don't think so, because this uses the `UserInterruptionManagerImplementation`, which is already behind a mutex.

v-next/hardhat-keystore/src/internal/tasks/set.ts
```diff
@@ -63,14 +72,14 @@ export const set = async (
...
-    consoleLog(UserDisplayMessages.displayKeyAlreadyExistsWarning(key));
+    consoleLog(UserDisplayMessages.displayKeyAlreadyExistsWarning(key, dev));
     process.exitCode = 1;
     return;
   }
 
   const secret = await requestSecretInput(
-    PLUGIN_ID,
-    UserDisplayMessages.enterSecretMessage(),
+    dev ? PLUGIN_ID_DEV : PLUGIN_ID,
```
fvictorio: I'm not sure we should do this. The plugin is the same, it's just a different feature.

fvictorio (commented):
This still doesn't seem to have the right behavior if you use a configuration variable in a test and it doesn't exist in the development keystore: I still get prompet for the production keystore password.

Maybe I'm doing something wrong? I pnpm linked `hardhat`, `hardhat-keystore`, `hardhat-errors` and `hardhat-node-test-runner`, added this to the HH config:

```
  networks: {
    prod: {
        type: "http",
        url: configVariable("PROD"),
    },
  }
```

and then ran this test with `hh test nodejs`:

```js
import { it } from "node:test";

it("test", async function () {
    await network.connect("prod");
});
```

v-next/hardhat-errors/src/descriptors.ts
fvictorio: I pushed a commit changing the wording of this message.

fvictorio:
My bad, I hadn't pulled the latest changes. This is what I get for trusting vscode.

Refers to this comment: https://github.com/NomicFoundation/hardhat/pull/7094#pullrequestreview-3083152835

v-next/hardhat-errors/src/descriptors.ts
ChristopherDedominici: In the UI messages we have this: `npx hardhat keystore set {key} --dev`. We should probably keep the same wording for consistency

v-next/hardhat-errors/src/descriptors.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7094/commits/cb0a52ef895442caa4ad5f43c1229f7a7605dece

v-next/hardhat-keystore/src/internal/tasks/set.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7094/commits/39bf544ad314ee6b449797545ff5a4b5fcdbab04

v-next/hardhat-errors/src/descriptors.ts
fvictorio: Sure

ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7094/commits/9365a886f1a4f37cbaf7d9885fda4b0a704fb7d9

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -2048,6 +2048,20 @@ Please check Hardhat's output for more details.`,
...
+        websiteTitle: "Cannot change password for dev keystore",
+        websiteDescription: `The keystore "change-password" task cannot be used with the development keystore`,
+      },
+      KEY_NOT_FOUND_DURING_TESTS_WITH_DEV_KEYSTORE: {
+        number: 50002,
+        messageTemplate: `Key "{key}" not found in the development keystore. Run "npx hardhat keystore set {key} --dev" to set it.`,
+        websiteTitle: "Key not found in the development keystore during tests",
+        websiteDescription: `Key not found in the development keystore. During tests, configuration variables can only be accessed through the development keystore.
+
+Run "npx hardhat keystore set {key} --dev" to set it.`,
```
fvictorio: The reason I used `KEY` instead of `{key}` in the suggestion is that this won't be replaced by the actual key value; it's only used in the website content. And I think in that context `{key}` looks kind of weird compared to a more normal `keystore set KEY --dev` or `keystore set <KEY> --dev`

v-next/hardhat-keystore/src/internal/utils/get-keystore-file-path.ts
```diff
@@ -9,3 +9,19 @@ export async function getKeystoreFilePath(): Promise<string> {
...
+  const configDirPath = await getConfigDir();
+  return path.join(configDirPath, "dev.keystore.json");
+}
+
+/**
+ * Get the path for the file containing the unencrypted password for the development keystore in the Hardhat global config directory.
+ */
+export async function getDevKeystorePasswordFilePath(): Promise<string> {
+  const configDirPath = await getConfigDir();
+  return path.join(configDirPath, "dev.keystore.checksum.txt");
```
fvictorio:
```suggestion
  return path.join(configDirPath, "hardhat.checksum");
```

v-next/hardhat-errors/src/descriptors.ts
ChristopherDedominici: Fai point, done here with also the `hardhat.checksum` changes: https://github.com/NomicFoundation/hardhat/pull/7094/commits/ca7e50b592ff1f425f558f0531a3091041700562

v-next/hardhat-keystore/src/internal/utils/get-keystore-file-path.ts
ChristopherDedominici: I also modified the tests accordingly with file extension here: [https://github.com/NomicFoundation/hardhat/pull/7094/commits/b5f56461ed05aa4cd586b9634be030ed32b553a1
and here](https://github.com/NomicFoundation/hardhat/pull/7094/commits/ca7e50b592ff1f425f558f0531a3091041700562)

v-next/hardhat-keystore/src/internal/loaders/keystore-file-loader.ts
```diff
@@ -21,6 +27,10 @@ export class KeystoreFileLoader implements KeystoreLoader {
     return this.#keystoreFilePath;
   }
 
+  public getKeystoreDevPasswordFilePath(): string {
+    return this.#keystoreDevPasswordFilePath;
+  }
```
kanej, line 32: Nit: I think I would go with a public property rather than a getter.
It isn't adding anything, so it prompts the questions from a consumer, oh they haven't gone for a public property here there must be something more going on.

v-next/hardhat-keystore/src/internal/tasks/delete.ts
```diff
@@ -1,48 +1,57 @@
...
-    interruptor: string,
-    inputDescription: string,
-  ) => Promise<string>,
-  consoleLog: (text: string) => void = console.log,
+  requestSecretInput: KeystoreRequestSecretInput,
+  consoleLog: KeystoreConsoleLog = console.log,
 ): Promise<void> => {
   if (!(await keystoreLoader.isKeystoreInitialized())) {
-    consoleLog(UserDisplayMessages.displayNoKeystoreSetErrorMessage());
+    consoleLog(UserDisplayMessages.displayNoKeystoreSetErrorMessage(dev));
```
kanej, line 40: I am not a fan of the boolean here, so maybe `UserDisplayMessages` gets dynamically generated based on `dev`?
But I don't think the tradeoff is quite worth it. So lets leave it.

Issue #7093, opened 2025-07-30
Send OP script throws error `Chain "Hardhat" does not support contract "gasPriceOracle"`


kanej:
In `v-next` if your run `hardhat run ./scripts/send-op-tx.ts` from within the template for viem you get:

```sh
Compiling your Solidity contracts...

Nothing to compile

Sending transaction using the OP chain type
Sending 1 wei from 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266 to itself
An unexpected error occurred:

ChainDoesNotSupportContract: Chain "Hardhat" does not support contract "gasPriceOracle".

This could be due to any of the following:
- The chain does not have the contract "gasPriceOracle" configured.

Version: viem@2.30.0
    at getChainContractAddress (/Users/user-x/repos/nomic/hardhat/node_modules/.pnpm/viem@2.30.0_typescript@5.8.3_zod@3.24.1/node_modules/viem/utils/chain/getChainContractAddress.ts:21:11)
    at <anonymous> (/Users/user-x/repos/nomic/hardhat/node_modules/.pnpm/viem@2.30.0_typescript@5.8.3_zod@3.24.1/node_modules/viem/op-stack/actions/estimateL1Gas.ts:85:14)
    at estimateL1Gas (/Users/user-x/repos/nomic/hardhat/node_modules/.pnpm/viem@2.30.0_typescript@5.8.3_zod@3.24.1/node_modules/viem/op-stack/actions/estimateL1Gas.ts:90:5)
    at Object.estimateL1Gas (/Users/user-x/repos/nomic/hardhat/node_modules/.pnpm/viem@2.30.0_typescript@5.8.3_zod@3.24.1/node_modules/viem/op-stack/decorators/publicL2.ts:511:32)
    at <anonymous> (/Users/user-x/repos/nomic/hardhat/v-next/hardhat/templates/01-node-test-runner-viem/scripts/send-op-tx.ts:15:34)

If you think this is a bug in Hardhat, please report it here: https://hardhat.org/report-bug
```

PR #7096, opened 2025-07-30
Fix hh-viem l2 actions when using the optimism chain type

fvictorio:
Fixed viem L2 actions, which @schaable broke by making the mistake of trusting my judgement.

Notice that I'm only adding the contracts. Even this is conceptually wrong because EDR doesn't add all the contracts in viem's `optimism` chain, but I think this is a good middle ground.

I also removed a skipped test that didn't make sense anyway.


schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7093

PR #7101, opened 2025-07-31
regroup test configs under a test object

kanej:
Move the Mocha and Solidity test configurations under a `test` property of the Hardhat config.

Also rename `solidityTest` to just `solidity`.

So from top level properties:

```js
mocha: {
  color: true,
},
solidityTest: {
  timeout: 1000,
},
```

They are now grouped under a `test` property:

```js
test: {
    mocha: {
      color: true,
    },
    solidity: {
      timeout: 1000,
    },
}
```

# Approach

1. In `test` built-in plugin extend `HardhatConfig` and `HardhatUserConfig` with `HardhatTestConfig` and `HardhatTestUserConfig`
2. Export `HardhatTestConfig` and `HardhatTestUserConfig` as types under `types/test`
  - should this be under `types/config`?
4. Add config hook to `test` built-in plugin that adds an empty object for `test` in the returned `HardhatConfig` instance
  - Note: no validation was added for this empty config object
5. Rewire the `mocha` plugin and the `solidity-test` built-in plugin to read and write from the new `test` grouping property

### DIFF
--- a/v-next/hardhat-mocha/src/type-extensions.ts
+++ b/v-next/hardhat-mocha/src/type-extensions.ts
@@ -3,19 +3,22 @@ import "hardhat/types/config";
 import type { MochaOptions } from "mocha";
 
 declare module "hardhat/types/config" {
-  export interface HardhatUserConfig {
-    mocha?: MochaOptions;
-  }
-
   export interface TestPathsUserConfig {
     mocha?: string;
   }
 
-  export interface HardhatConfig {
-    mocha: MochaOptions;
-  }
-
   export interface TestPathsConfig {
     mocha: string;
+  }
+}
+
+import "hardhat/types/test";
+declare module "hardhat/types/test" {
+  export interface HardhatTestUserConfig {
+    mocha?: MochaOptions;
+  }
+
+  export interface HardhatTestConfig {
+    mocha: MochaOptions;
   }
 }
--- a/v-next/hardhat/src/internal/builtin-plugins/test/config.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/test/config.ts
@@ -1,0 +1,17 @@
+import type {
+  HardhatConfig,
+  HardhatUserConfig,
+} from "../../../types/config.js";
+
+export async function resolveTestUserConfig(
+  userConfig: HardhatUserConfig,
+  resolvedConfig: HardhatConfig,
+): Promise<HardhatConfig> {
+  return {
+    ...resolvedConfig,
+    /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- the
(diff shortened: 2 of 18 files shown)


v-next/hardhat-mocha/src/type-extensions.ts
```diff
@@ -3,19 +3,22 @@ import "hardhat/types/config";
...
-  export interface HardhatConfig {
-    mocha: MochaOptions;
-  }
-
   export interface TestPathsConfig {
     mocha: string;
   }
 }
+
+import "hardhat/types/test";
```
kanej, line 15: I was unsure whether these type should be under `hardhat/types/config` or `hardhat/types/test`

v-next/hardhat/src/internal/builtin-plugins/test/hook-handlers/config.ts
```diff
@@ -0,0 +1,22 @@
+import type { ConfigHooks } from "../../../../types/hooks.js";
+
+import { resolveTestUserConfig } from "../config.js";
+
+export default async (): Promise<Partial<ConfigHooks>> => {
```
kanej, line 5: This hook exists solely to set an empty `test` object.
This could have been done directly in the built-in plugin for Solidity Test, but that felt like a hack.

v-next/hardhat/src/internal/builtin-plugins/test/config.ts
```diff
@@ -0,0 +1,17 @@
...
+export async function resolveTestUserConfig(
+  userConfig: HardhatUserConfig,
+  resolvedConfig: HardhatConfig,
+): Promise<HardhatConfig> {
+  return {
+    ...resolvedConfig,
+    /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- the
+    empty object is typed correctly by core type, but not when the solidity test
+    plugins extensions are included  */
+    test: userConfig.test ?? ({} as any),
```
kanej, line 15: I couldn't think of a better pattern here given the automatic type extension from the Solidity test plugin.

v-next/hardhat/src/types/test.ts
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable-next-line @typescript-eslint/no-empty-interface -- Empty
+interface allow plugins to extend the Test user configuration for Hardhat. */
+export interface HardhatTestUserConfig {}
+
+/* eslint-disable-next-line @typescript-eslint/no-empty-interface -- Empty
+interface allow plugins to extend the Test configuration for Hardhat. */
+export interface HardhatTestConfig {}
```
kanej, line 7: As I said, maybe these move to config.
Can I confirm that two empty interfaces are the expectation here?

fvictorio (approved):

PR #7102, opened 2025-07-31
Upgrade to @nomicfoundation/edr v0.12.0-next.3 and add created address in trace

frangio:
Release notes:

### 0.12.0-next.2

- bf9b55b: Improved parallelism in test suite execution for better error reporting and improved performance
- 0e0619e: Added support for function-level gas-tracking in Solidity tests.
- 3f822d8: Fixed panic when using the `pauseGasMetering` cheatcode
- 74b1f05: Made the `contract` field in `CallTrace` optional, and added a separate `address` field that is always present. (Breaking change)

### 0.12.0-next.3

- 6ea800c: Removed deprecated JSON-RPC methods: `eth_mining`, `net_listening`, `net_peerCount`, `hardhat_addCompilationResult`, `hardhat_intervalMine`, and `hardhat_reset`.
- a5cc346: Added cheatcode error stack trace entry. This fixes stack traces for errors from expect revert cheatcodes and improves stack traces for other cheatcode errors.

---

In addition to adapting to the breaking change to call traces, I've also added the created contract address in the printed call trace, which was the original request that motivated that change.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/solidity-stack-trace.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/solidity-stack-trace.ts
@@ -24,6 +24,7 @@ import type {
   ContractTooLargeErrorStackTraceEntry,
   InternalFunctionCallStackEntry,
   ContractCallRunOutOfGasError,
+  CheatcodeErrorStackTraceEntry,
 } from "@nomicfoundation/edr";
 
 import {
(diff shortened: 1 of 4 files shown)


fvictorio:
This bump adds a new stack trace entry type, but we don't get a compilation error anywhere for not handling it. I think the reason is that `SolidityStackTraceEntry` is just a big union type that has to be manually updated.

There's probably a better way to handle this, but as a quick workaround I just added a type-level test to ensure that the `type` field of that union type is equal to EDR's exported `StackTraceEntryType` enum.

frangio:
Thanks @fvictorio!

> There's probably a better way to handle this

The way EDR exports these types makes this hard, but EDR is constrained by limitations of NAPI-RS. I think this will be fixable when EDR migrates to NAPI-RS v3 which was just released, as it enables discriminated unions natively.

PR #7106, opened 2025-08-01
fix: copy across network config extensions

kanej:
This PR fixes an issue that was showing up when running `npx hardhat node` while using a plugin that extends network config.

Fundamentally, if the user provides network config overrides, the extension from the plugin where not being copied across. `npx hardhat node` always adds overrides, hence why it showed there.

This PR resolves the problem with two moves:

1. Build up the correct NetworkConfig (including plugin extensions) by re-calling the two relevant hooks on a merged version of the User provided original network config with the particular overrides at the point that `network.create()` is called.
2. Loosen the validation performed on Network objects to no longer be strict, so that new properties can be defined in plugins.

> [!note]
> Review commit at a time

This work is part of #6983.

## Archive

Plugins can extend network config, but the config generated for a new network connection was missing those extensions (they where not being copied across).~~

This is in response to a bug booting `npx hardhat node` from the `v3-hh-ledger` branch within the example-project. If you run `pnpm hardhat node` you get an error:

```shell
> pnpm hardhat node                        
An unexpected error occurred:

TypeError: Cannot read properties of undefined (reading 'map')
    at new LedgerHandler (/Users/kanej/repos/nomic/hardhat/v-next/hardhat-ledger/src/internal/handler.ts:83:46)
    at <anonymous> (/Users/kanej/repos/nomic/hardhat/v-next/hardhat-ledger/src/internal/hook-handlers/network.ts:53:34)
    at AsyncMutex.exclusiveRun (/Users/kanej/repos/nomic/hardhat/v-next/hardhat/src/internal/core/async-mutex.ts:24:20)
    at async Array.onRequest (/Users/kanej/repos/nomic/hardhat/v-next/hardhat-ledger/src/internal/hook-handlers/network.ts:48:29)
    at async next (/Users/kanej/repos/nomic/hardhat/v-next/hardhat/src/internal/core/hook-manager.ts:124:13)

If you think this is a bug in Hardhat, please report it here: https://hardhat.org/report-bug
```

The issue is that the Ledger handler assumes that `networkConnection.networkConfig.ledgerAccounts` is populated:

```ts
// v-next/hardhat-ledger/src/internal/hook-handlers/network.ts line 48
handlerPerConnection = new LedgerHandler(
  networkConnection.provider,
  {
    accounts: networkConnection.networkConfig.ledgerAccounts,
  },
  context.interruptions.displayMessage.bind(context.interruptions),
);
```

The version of `networkConfig` associated with the connection is a clone, but the bug is that the `ledgerAccounts` are not being copied across.

This PR updates the config creation during network connection creation to merge the user provided network config (extended by plugins) and the overrides.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
@@ -48,5 +48,7 @@ async function createNetworkManager(
     context.artifacts,
     hre.userConfig,
     hre.config.chainDescriptors,
+    hre.globalOptions.config,
+    hre.config.paths.root,
   );
 }
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
@@ -2,7 +2,6 @@ import type { CoverageConfig } from "./edr/types/coverage.js";
 import type { ArtifactManager } from "../../../types/artifacts.js";
 import type {
   ChainDescriptorsConfig,
-  HardhatConfig,
   HardhatUserConfig,
   NetworkConfig,
   NetworkConfigOverride,
@@ -22,7 +21,6 @@ import type {
   JsonRpcRequest,
   JsonRpcResponse,
 } from "../../../types/providers.js";
-import type { TaskDefinition } from "../../../types/tasks.js";
 
 import {
   HardhatError,
@@ -31,14 +29,13 @@ import {
 import { exists, readBinaryFile } from "@nomicfoundation/hardhat-utils/fs";
 import { deepMerge } from "@nomicfoundation/hardhat-utils/lang";
 
-import { resolveConfigurationVariable } from "../../core/configuration-variables.js";
+import { resolveUserConfigToHardhatConfig } from "../../core/hre.js";
 import { isSupportedChainType } from "../../edr/chain-type.js";
 import { JsonRpcServerImplementation } from "../node/json-rpc/server.js";
 
 import { EdrProvider } from "./edr/edr-provider.js";
 import { HttpProvider } from "./http-provider.js";
 import { NetworkConnectionImplementation } from "./network-connection.js";
-import { validateNetworkConfigOverride } from "./type-validation.js";
 
 export type JsonRpcRequestWrapperFunction = (
   request: JsonRpcRequest,
@@ -53,6 +50,8 @@ export class NetworkManagerImplementation implements NetworkManager {
   readonly #artifactsManager: Readonly<ArtifactManager>;
   readonly #userConfig: Readonly<HardhatUserConfig>;
   readonly #chainDescriptors: Readonly<ChainDescriptorsConfig>;
+  readonly #userProvidedConfigPath: Readonly<string | undefined>;
+  readonly #projectRoot: string;
 
   #nextConnectionId = 0;
 
@@ -64,6 +63,8 @@ export class NetworkManagerImplementation implements NetworkManager {
     artifactsManager: ArtifactManager,
     userConfig: HardhatUserConfig,
     chainDescriptors: ChainDescriptorsConfig,
+    userProvidedConfigPath: string | undefined,
+    projectRoot: string,
   ) {
     this.#defaultNetwork = defaultNetwork;
     this.#defaultChainType = defaultChainType;
@@ -72,6 +73,8 @@ export class NetworkManagerImplementation implements NetworkManager {
     this.#artifactsManager = artifactsManager;
     this.#userConfig = userConfig;
     this.#chainDescriptors = chainDescriptors;
+    this.#userProvidedConfigPath = userProvidedConfigPath;
+    this.#projectRoot = projectRoot;
   }
 
   public async connect<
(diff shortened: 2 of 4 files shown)


schaable (approved):
I wonder if we should merge these recursively, but this seems good enough for now.

kanej:
I have spoken with Pato on this to check the logic.
We could merge this PR and it would resolve the `hardhat-ledger` problem with `npx hardhat node`.
However, we should fix the underlying issue, which is that we are not extending config here in the same way we do within the main config resolution.
We need to look at pull across some of that machinery to network config resolution. Right now that resolution does not have extension points.
This PR is not merge-able until then.

v-next/hardhat/src/internal/builtin-plugins/network-manager/type-validation.ts
```diff
@@ -404,8 +404,8 @@ const userConfigSchema = z.object({
 
 const networkConfigOverrideSchema = z
   .discriminatedUnion("type", [
-    httpNetworkUserConfigSchema.strict(),
-    edrNetworkUserConfigSchema.strict(),
```
kanej, line 408: Change the rules for validating the network config. We are going to allow arbitrary properties to be added to the network config as we intend to do that from plugins (i.e. `hardhat-ledger`).

v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -302,23 +302,6 @@ describe("NetworkManagerImplementation", () => {
       );
     });
 
-    it("should throw an error if the specified network config override has mixed properties from http and edr networks", async () => {
```
kanej, line 305: Because we now allow arbitrary properties, we can no longer provide this level of helpful error message.

v-next/hardhat/test/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -359,6 +361,165 @@ describe("NetworkManagerImplementation", () => {
     });
   });
 
+  describe("connect when config has been extended by plugins", () => {
+    const networkConfigAddingPlugin: HardhatPlugin = {
+      id: "network-config-adding-plugin",
```
kanej, line 368: This plugin simulates the equivalent of `hardhat-ledger` extending the Network config with additional properties.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -132,57 +132,10 @@ export class NetworkManagerImplementation implements NetworkManager {
...
-            );
-    }
-
-    const resolvedNetworkConfig =
-      resolvedNetworkConfigOverride ??
-      this.#networkConfigs[resolvedNetworkName];
+    const resolvedNetworkConfig = await this.#resolveNetworkConfig(
+      resolvedNetworkName,
+      networkConfigOverride,
+    );
```
kanej, line 146: This is a pure move refactoring. One commit moves the existing code into a subfunction, the next replaces the logic.

v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
```diff
@@ -47,7 +46,7 @@ async function createNetworkManager(
     hre.config.networks,
     context.hooks,
     context.artifacts,
-    userConfigNetworks,
```
kanej, line 50: To fully run the config resolution hooks that will extend the config for NetworkConfig, we need to run all the plugins across all of the config.
Hence we now pass in the the original `userConfig` and not just its networks.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -279,59 +284,91 @@ export class NetworkManagerImplementation implements NetworkManager {
...
-        throw new HardhatError(
-          HardhatError.ERRORS.CORE.NETWORK.INVALID_CONFIG_OVERRIDE,
-          {
-            errors: `\t* The type of the network cannot be changed.`,
-          },
-        );
-      }
+    if (networkConfigOverride === undefined) {
+      return this.#networkConfigs[resolvedNetworkName];
+    }
```
kanej, line 292: If no override is provided by the user on `hre.network.connect({...})` then we can use the existing pre-resolved network config and return immediately.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -279,59 +284,91 @@ export class NetworkManagerImplementation implements NetworkManager {
...
-              )
-              .join("\n\t")}`,
-          },
-        );
-      }
+    const newConfig = deepMerge(this.#userConfig, {
+      networks: {
+        [resolvedNetworkName]: networkConfigOverride,
+      },
+    });
```
kanej, line 311: Otherwise we merge in the user provided network config override into the user's `HardhatUserConfig`.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -279,59 +284,91 @@ export class NetworkManagerImplementation implements NetworkManager {
...
+              error.path.length > 0
+                ? `* Error in ${error.path.join(".")}: ${error.message}`
+                : `* ${error.message}`,
             )
-          : resolveEdrNetwork(newConfig, "", (strOrConfigVar) =>
-              resolveConfigurationVariable(this.#hookManager, strOrConfigVar),
-            );
+            .join("\n\t")}`,
+        },
+      );
```
kanej: We then validate that combination, but constrain the validation to only the relevant network config.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -279,59 +284,91 @@ export class NetworkManagerImplementation implements NetworkManager {
...
-      resolvedNetworkConfigOverride ??
-      this.#networkConfigs[resolvedNetworkName];
+    const updatedConfig = await this.#hookManager.runHandlerChain(
+      "config",
+      "extendUserConfig",
+      [newConfig],
+      async (c) => {
+        return c;
+      },
+    );
```
kanej: Finally we run two hooks to resolve the orginal user config + passed in overrides in to a full network config:

1. extendUserConfig
2. resolveUserConfig

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -279,59 +284,91 @@ export class NetworkManagerImplementation implements NetworkManager {
...
+    const initialResolvedConfig = {
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- kill me
+      plugins: [] as HardhatPlugin[],
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- kill me
+      tasks: [] as TaskDefinition[],
+      paths: {
+        root: "",
+        cache: "",
+      },
+    } as HardhatConfig;
```
kanej: This is the minimal config I was able to come up with to let me rerun the config resolution.
My justification is that we are really only interested in the network config that results, everything else (i.e. plugins/tasks/other top level properties) can be ignored.
But I would like to double check that logic.

alcuadrado: This will probably work, but i'm not entirely sure. One thing that we could do to ensure it does is refactoring `v-next/hardhat/src/internal/core/hre.ts` a bit so that it exposes a helper:

```ts
async function resolveConfig(
      userConfig: HardhatUserConfig,
      configPath: string | undefined,
      resolvedProjectRoot: string,
      resolvedPlugins: HardhatPlugin[],
      hooks: HookManager,
    ): Promise<HardhatConfig | HardhatUserConfigValidationError[]> {
  // ...      
}
```

And use it here and in `v-next/hardhat/src/internal/core/hre.ts`.

This would assume that any `HardhatUserConfigValidationError` found here is due to the override, which I think it's a safe assumption, as the original config was already validated.

alcuadrado: This comment may seem to imply a major change, but it's mostly moving these lines into a function: https://github.com/NomicFoundation/hardhat/blob/main/v-next/hardhat/src/internal/core/hre.ts#L70-L113

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -334,4 +377,19 @@ export class NetworkManagerImplementation implements NetworkManager {
...
+    if (path[0] !== undefined && path[0] === "networks") {
+      path = path.slice(1);
+    }
+
+    if (path[0] !== undefined && path[0] === resolvedNetworkName) {
+      path = path.slice(1);
+    }
+
+    return path;
+  }
```
kanej, line 394: This allows us to scope the validation errors to the network config of the named network.

v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
```diff
@@ -48,5 +48,7 @@ async function createNetworkManager(
     context.artifacts,
     hre.userConfig,
     hre.config.chainDescriptors,
+    hre.globalOptions.config,
+    hre.config.paths.root,
```
kanej, line 52: These two properties are required as input to the general `resolveUserConfigToHardhatConfig` function.
Let me know if there is a way of getting them from the existing passed in properties, I was unable to find them.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
kanej: I have refactored out the resolution code as a function and exported it: https://github.com/NomicFoundation/hardhat/pull/7106/commits/5a9d9c7c8603a01b9b170a778b9e240c27a284f1#diff-03173af1ae628b27bdff9793dc3e766d0a0a537dd3cd74c19c00f5934261e5bcR51-R52

The only only contentious part I could see was the wiring through of two additional properties from the `hre` to the resolution via the NetworkManager constructor.

alcuadrado (approved):

PR #7107, opened 2025-08-01
Compiler selection changes

### LINKED ISSUE #5993: Use wasm for all production build profile builds

antico5:
This PR introduces a new flag for solidity build profiles, `preferWasm`. When set to true, the wasm compiler will be used over the native one. All profiles have this set to `false` as default, except `production` which have this set to `true` by default.

Additionally, support for linux-aarch64 native compiler has been added. Some changes to the downloaded have been introduced since the source for the binaries doesn't behave like the official mirror for solc binaries. For instance, `keccak` checksum has been replaced with `sha256` to support both sources.

Since this is the only special case added, no design changes  have been made to the downloaded, although if we had more special cases in the future, we may need to think about using a strategy pattern or something similar.

Closes #5993

### DIFF
--- a/.changeset/nine-lies-press.md
+++ b/.changeset/nine-lies-press.md
@@ -1,0 +1,7 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+"@nomicfoundation/hardhat-utils": patch
+"hardhat": patch
+---
+
+Add preferWasm flag to config. Default to wasm on production profile. Support linux-aarch64 native compiler ([#5993](https://github.com/NomicFoundation/hardhat/issues/5993))
--- a/.changeset/nine-lies-press.md
+++ b/.changeset/nine-lies-press.md
@@ -1,5 +1,4 @@
 ---
-"@nomicfoundation/hardhat-verify": patch
 "@nomicfoundation/hardhat-utils": patch
 "hardhat": patch
 ---
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -152,15 +152,12 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       ),
     );
 
-    const runCompilationJobOptions: RunCompilationJobOptions = {
-      quiet: options?.quiet,
-    };
     const results: CompilationResult[] = await pMap(
       runnableCompilationJobs,
       async (runnableCompilationJob) => {
         const compilerOutput = await this.runCompilationJob(
           runnableCompilationJob,
-          runCompilationJobOptions,
+          options,
         );
 
         return {
@@ -301,22 +298,15 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       readSourceFileFactory(this.#hooks),
     );
 
-    const buildProfileName = options?.buildProfile ?? DEFAULT_BUILD_PROFILE;
-
-    if (this.#options.solidityConfig.profiles[buildProfileName] === undefined) {
-      throw new HardhatError(
-        HardhatError.ERRORS.CORE.SOLIDITY.BUILD_PROFILE_NOT_FOUND,
-        {
-          buildProfileName,
-        },
-      );
-    }
+    const { buildProfileName, buildProfile } = this.#getBuildProfile(
+      options?.buildProfile,
+    );
 
     log(`Using build profile ${buildProfileName}`);
 
     const solcConfigSelector = new SolcConfigSelector(
       buildProfileName,
-      this.#options.solidityConfig.profiles[buildProfileName],
+      buildProfile,
       dependencyGraph,
     );
 
@@ -346,7 +336,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       let solcLongVersion = solcVersionToLongVersion.get(solcConfig.version);
 
       if (solcLongVersion === undefined) {
-        const compiler = await getCompiler(solcConfig.version);
+        const compiler = await getCompiler(solcConfig.version, false);
         solcLongVersion = compiler.longVersion;
         solcVersionToLongVersion.set(solcConfig.version, solcLongVersion);
       }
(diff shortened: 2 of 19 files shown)


.changeset/nine-lies-press.md
```diff
@@ -0,0 +1,7 @@
+---
+"@nomicfoundation/hardhat-verify": patch
```
fvictorio: Only the tests were changed in `hardhat-verify`, it shouldn't be included in the changeset.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -506,6 +496,23 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     return { compilationJobsPerFile, indexedIndividualJobs };
   }
 
+  #getBuildProfile(_buildProfileName?: string) {
+    const buildProfileName = _buildProfileName ?? DEFAULT_BUILD_PROFILE;
```
fvictorio: Personal preference, I would replace these two lines with:

```typescript
  #getBuildProfile(buildProfileName: string = DEFAULT_BUILD_PROFILE) {
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/index.ts
```diff
@@ -77,17 +77,22 @@ export async function downloadConfiguredCompilers(
...
 
-export async function getCompiler(version: string): Promise<Compiler> {
-  const platform = CompilerDownloaderImplementation.getCompilerPlatform();
-  const compilerDownloader = new CompilerDownloaderImplementation(
-    platform,
-    await getGlobalCompilersCacheDir(),
-  );
+export async function getCompiler(
+  version: string,
+  preferWasm: boolean,
```
fvictorio: Can we make this an options object? Otherwise the calls to it look like this:

```ts
getCompiler(version, false)
```

instead of the much clearer:

```ts
getCompiler(version, { preferWasm: false })
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/index.ts
```diff
@@ -77,17 +77,22 @@ export async function downloadConfiguredCompilers(
...
+): Promise<Compiler> {
+  if (preferWasm === false) {
```
fvictorio: `preferWasm` is not nullable nor optional, so `!preferWasm` should be fine here.

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -300,12 +309,17 @@ function resolveSolidityConfig(
...
+      preferWasm,
     };
   }
 
+  // This will generate default build profiles (e.g. production) when they are not specified in the config, cloning from 'default', which is always present
   for (const profile of DEFAULT_BUILD_PROFILES) {
     if (!(profile in profiles)) {
-      profiles[profile] = profiles.default;
+      profiles[profile] = {
+        ...profiles.default,
```
fvictorio, line 320: Unrelated to this particular PR, but we should probably do a deep copy here. Otherwise this happens:

Image

It seems pedantic, but we've been bitten by this before.

But let's not do it in this PR. We should have a separate issue to make sure that both the user and the resolved config are made up of independent objects.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -354,12 +389,12 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
...
   ): Promise<boolean> {
-    const expectedKeccak256 = build.keccak256;
+    const expectedSha = getPrefixedHexString(build.sha256);
     const compiler = await readBinaryFile(downloadPath);
 
-    const compilerKeccak256 = bytesToHexString(await keccak256(compiler));
+    const compilerSha = bytesToHexString(await sha256(compiler));
 
-    if (expectedKeccak256 !== compilerKeccak256) {
+    if (expectedSha !== compilerSha) {
```
alcuadrado, line 397: A few lines below, you can find this

```ts
    if (
      this.#platform === CompilerPlatform.LINUX ||
      this.#platform === CompilerPlatform.MACOS
    ) {
      await chmod(downloadPath, 0o755);
    }
```

but that condition should also include LINUX_ARM64. Otherwise we `chmod` it and it's never used.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -343,7 +333,9 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       let solcLongVersion = solcVersionToLongVersion.get(solcConfig.version);
 
       if (solcLongVersion === undefined) {
-        const compiler = await getCompiler(solcConfig.version);
+        const compiler = await getCompiler(solcConfig.version, {
+          preferWasm: false,
```
alcuadrado: Why does this use `preferWasm: false`? It prevents wasm from working on linux arm64, as it will get the long version from the other compiler list, and it doesn't use long versions.

alcuadrado: `{ preferWasm: buildProfile.preferWasm }`

This fixes it

alcuadrado: It also was messing up with the cache system, as switching between wasm and this other build didn't generate different build ids.

alcuadrado (changes requested):
I left a few comments about two bugs I found when testing this on Linux aarch64.

I think we should include a `wasm: boolean` to `CompileCacheEntry`. This way we ensure that production builds don't get messed with the rest.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
antico5: good catch

antico5:
> I left a few comments about two bugs I found when testing this on Linux aarch64.
> 
> I think we should include a `wasm: boolean` to `CompileCacheEntry`. This way we ensure that production builds don't get messed with the rest.

Great catches. I will add wasm flag to compile cache entry. I didn't expect them to have the same long version as native compilers, but just checked and that is the case.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
antico5: Huge find! thanks for spotting it

antico5:
@alcuadrado Addressed the feedback, although I found that wasm is not working on windows (unrelated to these changes), but I'll fix it on this same PR

alcuadrado (approved):

Issue #7058, opened 2025-07-24
Change type of packageName


kanej:
We want to adapt Hardhat 3 plugins so: 
 
* Change `npmPackage?: string;` to `npmPackage?: string | null;`
* Semantics
    - `undefined` ⇒ use the id as package name
    - `string` ⇒ use that as package name to validate dependencies
    - `null` ⇒ don’t run dependencies validation
* We can add a test to the plugin boilerplate checking that the id is the package name

PR #7115, opened 2025-08-04
Change type of packageName

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/7058

I want to double check that this logic (described in the issue description) is correct: `undefined ⇒ use the id as package name`

If `npmPackage` is undefined, we're gonna fallback on the `id`, but what if the `id` does not match any package?
An error will be thrown. Is that ok? Check this test as example: https://github.com/NomicFoundation/hardhat/pull/7115/files#r2254358608


v-next/hardhat/test/internal/core/plugins/detect-plugin-npm-dependency-problems.ts
```diff
@@ -53,6 +69,24 @@ describe("Plugins - detect npm dependency problems", () => {
         },
       );
     });
+
+    it("should fail validation if the package has been installed and the npmPackage property is undefined but the id has a value that does not match the installed package", async () => {
```
ChristopherDedominici, line 73: I need a double check: is this a real scenario that could happen?

kanej: This is correct.
We fallback to the `id`, if package has not been installed, we error as this plugin will not be able to run.

.changeset/wicked-lies-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Change the type of the `npmPackage` plugin property [7058](https://github.com/NomicFoundation/hardhat/issues/7058)
```
kanej:
```suggestion
Change the type of the `npmPackage` plugin property ([7058](https://github.com/NomicFoundation/hardhat/issues/7058))
```

kanej (approved):

PR #7120, opened 2025-08-04
Set proper fromBlock value to avoid tests hanging on fork mode

### LINKED ISSUE #6967: Attempting to read events from forked network hangs tests
### Version of Hardhat

3.0.0-next.20

### What happened?

Attempting to fetch events from a forked edr network causes the tests to hang.

Maybe related: #1940

1. Initialize a new hardhat project with the default template:
   ```sh
   pnpm dlx hardhat@next --init
   ```

2. Setup a test
   file: `test/Counter.ts`
   ```ts
   
   describe("Counter", async function () {
     const { viem } = await network.connect("hardhatMainnet");
     const publicClient = await viem.getPublicClient();
   
     it("The sum of the Increment events should match the current value", async function () {
       const counter = await viem.deployContract("Counter");
   
       // run a series of increments
       for (let i = 1n; i <= 10n; i++) {
         await counter.write.incBy([i]);
       }
   
       console.log("fetching events");
       const events = await publicClient.getContractEvents({
         address: counter.address,
         abi: counter.abi,
         eventName: "Increment",
         fromBlock: 0n,
         strict: true,
       });
       console.log(`fetched ${events.length} events`);
   
       // check that the aggregated events match the current value
       let total = 0n;
       for (const event of events) {
         total += event.args.by;
       }
   
       assert.equal(total, await counter.read.x());
     });
   ```

3. Run test
   ```console
   $ pnpm hardhat test test/Counter.ts

   Compiling your Solidity contracts
   Compiled 1 Solidity file with solc 0.8.28 (evm target: cancun)
   
   Running node:test tests
   
   fetching events
   fetched 10 events
     Counter
       ✔ The sum of the Increment events should match the current value
   
   
   1 passing (1345ms)
   ```
   (test passes)

4. Setup forking
   ```diff
   diff --git a/hardhat.config.ts b/hardhat.config.ts
   index 2c515fa..69fd13c 100644
   --- a/hardhat.config.ts
   +++ b/hardhat.config.ts
   @@ -67,6 +67,11 @@ const config: HardhatUserConfig = {
        hardhatMainnet: {
          type: "edr",
          chainType: "l1",
   +      gasPrice: 10,
   +      initialBaseFeePerGas: 10,
   +      forking: {
   +        url: "https://eth.blockrazor.xyz",
   +      },
        },
        hardhatOp: {
          type: "edr",
   ```

5. Run test on forked network
   ```console
   $ pnpm hardhat test test/Counter.ts
   
   Compiling your Solidity contracts
   Compiled 1 Solidity file with solc 0.8.28 (evm target: cancun)
   
   Running node:test tests
   
   fetching events
   ```
   The command hangs after this, with no output.

### Search terms

forking, hang, stuck, local network, events

antico5:
Closes #6967


v-next/hardhat/templates/01-node-test-runner-viem/test/Counter.ts
```diff
@@ -71,6 +71,7 @@ describe("Counter", async function () {
   });
 
   it("The sum of the Increment events should match the current value", async function () {
+    const startBlockNumber = await publicClient.getBlockNumber();
```
fvictorio: I think a more common pattern is to get the block number after deploying the contract, because conceptually you want to use the range of blocks since the contract was deployed. It doesn't make a huge difference in practice, but I think the sample projects should be as accurate as possible.

This would also make the code a bit more self-explanatory, because you can call the variable `deploymentBlockNumber`, and it's more obvious why you are using it.

PR #7127, opened 2025-08-05
chore: include release info in Sentry

### LINKED ISSUE #7011: Include release information in Sentry
Sentry exceptions and transactions can be tagged with the current Hardhat release to allow for better metadata in the Sentry UI. This is especially useful when tracking down when a bug started to occur.

We leverage this feature in the VSCode extension and should take advantage of it in Hardhat 3.

kanej:
We include the hardhat version already, but by passing it as a release we leverage Sentries release tracking.

Resolves #7011

## Manual Testing

I manually tested this by forcing an error within the example project. This generated a new release under `hardhat3` in Sentry.

### DIFF
--- a/v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
+++ b/v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
@@ -55,6 +55,7 @@ export class Anonymizer {
     const result: Event = {
       event_id: event.event_id,
       platform: event.platform,
+      release: event.release,
       timestamp: event.timestamp,
       extra: event.extra,
     };
--- a/v-next/hardhat/src/internal/cli/telemetry/sentry/init.ts
+++ b/v-next/hardhat/src/internal/cli/telemetry/sentry/init.ts
@@ -30,6 +30,7 @@ import { createGetModuleFromFilename } from "./vendor/utils/module.js";
 interface InitOptions {
   dsn: string;
   transport: (transportOptions: BaseTransportOptions) => Transport;
+  release?: string;
   serverName?: string;
   integrations?: (integrations: Integration[]) => Integration[];
 }
--- a/v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
+++ b/v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
@@ -68,6 +68,7 @@ class Reporter {
       "./vendor/integrations/contextlines.js"
     );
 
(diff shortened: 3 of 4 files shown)


v-next/hardhat/src/internal/cli/telemetry/sentry/init.ts
```diff
@@ -30,6 +30,7 @@ import { createGetModuleFromFilename } from "./vendor/utils/module.js";
 interface InitOptions {
   dsn: string;
   transport: (transportOptions: BaseTransportOptions) => Transport;
+  release?: string;
```
kanej, line 33: It is optional here to allow the subprocess version to be booted without this information.

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -55,6 +55,7 @@ export class Anonymizer {
     const result: Event = {
       event_id: event.event_id,
       platform: event.platform,
+      release: event.release,
```
kanej, line 58: The anonymizer has to copy the release information across from the original event from the main process.

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -79,14 +80,15 @@ class Reporter {
     await init({
       dsn: SENTRY_DSN,
       transport: makeSubprocessTransport,
+      release: `hardhat@${hardhatVersion}`,
```
kanej, line 83: I don't think pulling in the plugin name is worth it.

v-next/hardhat/test/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -17,6 +19,29 @@ class MockedAnonymizer extends Anonymizer {
...
+    const anonymizer = new Anonymizer();
+
+    const result = await anonymizer.anonymize(originalEvent);
+
+    if (!result.success) {
+      assert.fail("The event should anonymize without issue");
+      return;
+    }
+
+    assert.deepEqual(result.event, originalEvent);
```
kanej, line 42: Any suggestions of a better pattern here would be welcome. Or would we use an invariant?

alcuadrado: i think it looks good

alcuadrado (approved):

PR #7133, opened 2025-08-06
fix: pass error as parent parameter in HardhatLedgerConnectionError constructor

MozirDmitriy:
Pass the original error as parent parameter to maintain proper error chaining.


fvictorio (approved):

fvictorio:
Thanks!

PR #7134, opened 2025-08-06
Refactor sentry reporter

alcuadrado:
This PR refactors how we use sentry, aligning it with the POC that I prepare. The key to understand the reason is that:

- Sentry internally has a buffer/queue of async actions
- We can't control that buffer without introducing a delay
- Most of the time, we want the process to die right after reporting (hence we don't want a delay)
- If sentry controls that buffer, we can't guarantee that the subprocesses are created in time
- Using the base transports and/or transport factories forces you into that buffered behavior
- Using a hand-rolled Transport doesn't. We keep control
- By doing that, we create the subprocesses synchronously, so don't even `return` to sentry before creating the subprocess
- A Transport works at the level of `Envelope`. The factories we were using use `Event`. So I needed to adapt the anonymizer

I removed some tests, as they weren't working. We should think how to test this.

I tested it manually tons of times.


v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -254,6 +287,8 @@ export class Anonymizer {
       result.stacktrace = await this.#anonymizeStacktrace(value.stacktrace);
     }
 
+    result.mechanism = value.mechanism;
```
kanej: Nit: Could this have been done in `results` literal initializer?

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -39,25 +50,47 @@ export class Anonymizer {
...
       event_id: event.event_id,
       platform: event.platform,
-      release: event.release,
       timestamp: event.timestamp,
       extra: event.extra,
+      release: event.release,
+      contexts: event.contexts,
+      sdk: event.sdk,
+      level: event.level,
+      server_name: event.server_name,
```
kanej: Server name likely requires additional anonymization?

v-next/hardhat/src/internal/cli/telemetry/sentry/init.ts
```diff
@@ -1,118 +1,149 @@
...
-  // at all, and they're of limited use in the context of a reporter subprocess.
-  const integrationOptions = {
-    defaultIntegrations: [
-      // Inbound filters integration filters out events (errors and transactions) mainly based on init inputs we never use
-      // Import from @sentry/core if needed
-      // inboundFiltersIntegration(),
-
-      functionToStringIntegration(),
-      linkedErrorsIntegration(),
-      requestDataIntegration(),
```
kanej, line 57: This one is intentionally skipped?

v-next/hardhat/src/internal/cli/telemetry/sentry/init.ts
```diff
@@ -1,118 +1,149 @@
...
-export async function init(options: InitOptions): Promise<void> {
-  const stackParser = stackParserFromStackParserOptions(
-    createStackParser(nodeStackLineParser(createGetModuleFromFilename())),
+export function init(options: GlobalCustomSentryReporterOptions): void {
+  initAndBind<ServerRuntimeClient, ServerRuntimeClientOptions>(
+    ServerRuntimeClient,
+    {
+      dsn: options.dsn,
+      environment: options.environment,
+      serverName: GENERIC_SERVER_NAME,
```
kanej, line 85: Ah, maybe anonymization isn't needed!

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -63,28 +68,24 @@ class Reporter {
...
+    const { setExtra } = await import("@sentry/core");
 
     const hardhatVersion = await getHardhatVersion();
     const { init } = await import("./init.js");
 
-    const linkedErrorsIntegrationInstance = linkedErrorsIntegration({
-      key: "cause",
-    });
+    const release = `hardhat@${hardhatVersion}`;
+    const environment = "production";
```
kanej, line 84: We can look at this later, it has been useful in vscode to test sentry changes against the `development` environment.

v-next/hardhat/src/internal/cli/telemetry/sentry/transport.ts
```diff
@@ -0,0 +1,204 @@
...
+  environment: string,
+  configPath?: string,
+): Transport {
+  return {
+    send: (envelope) => {
+      const verbose = log.enabled;
+
+      // **Synchronously** spawn a detached subprocess here
+
+      const out = verbose ? process.stdout : "ignore";
```
kanej, line 43: I think this is a potential epipe if `DEBUG=*` is on.

I suspect the way of dealing with this is within the subprocess, so a function that only logs if stdout is open, or does `debug` take care of that?

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
kanej: Ignore, we set it to a known value.

v-next/hardhat/src/internal/cli/telemetry/sentry/transport.ts
kanej: To be clear, we can do this later as a follow up PR.

alcuadrado: It may happen in an extreme case. But given that you need to be running in debug mode, i wouldn't care much about it.

kanej: Agreed, I couldn't trigger it manually

v-next/hardhat/src/internal/cli/telemetry/sentry/init.ts
alcuadrado: Yes, it's for services: https://docs.sentry.io/platforms/javascript/guides/connect/configuration/integrations/requestdata/

v-next/hardhat/src/internal/cli/telemetry/sentry/init.ts
alcuadrado: it wouldn't hurt being cautions. I changed it.

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
alcuadrado: TBH I don't really know how this should be used. If you explain it to me, I can switch it.

v-next/hardhat/src/internal/cli/telemetry/sentry/init.ts
alcuadrado: Addressed

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
alcuadrado: Done

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
```diff
@@ -63,28 +68,24 @@ class Reporter {
...
 
-    const contextLinesIntegrationInstance = contextLinesIntegration();
-
-    await init({
+    init({
```
kanej, line 86: I think this is a problem.
`main` sets `setCliHardhatConfigPath`, but doesn't instantiate an instance.
Instantiating an instance calls `init`, which then registers the global exception handlers.
However, if I am reading correctly, it is only on the first `sendErrorTelemetry` that the we instantiate the reporter.

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
kanej: When doing local dev, you could set this as `development`. Sentry lets you switch between environments in its views. That way the dashboard isn't polluted with weird errors during development.
But I don't think we address it here.

v-next/hardhat/src/internal/cli/telemetry/sentry/reporter.ts
alcuadrado: Take a look at my new commits

kanej: I have retested manually, I am able to trigger unexpected exceptions in the Hardhat code and see them come through.

kanej (approved):
I have done a fresh round of manual tests against:

1. Failures in user land e.g. Hardhat scripts
2. Unhandled exceptions within the (hacked) Hardhat code

Both exceptions come through cleanly to Sentry. Their logging locally shows successfully as well.

PR #7135, opened 2025-08-06
Avoid cleaning up solidity test artifacts on compile task

### LINKED ISSUE #7116: Incorrect caching in build for Solidity Test
The Solidity Test pipeline is not leveraging the build systems caching correctly.

antico5:
I found the issue with solidity tests being recompiled. The test task runs compile 2 times, one for the contracts and one for the tests. The first run, which compiles all contracts (empty files list), triggers the cleanup. Then this cleanup would remove the test artifacts, since they won't match any file from the full compilation (because full compilation doesn't compile test files).

The proposed fix is skipping test artifacts from cleanup, that would mean their source files end with `.t.sol` or they are `.sol` but inside the solidity tests folder. The only downside is that some artifact folders won't be automatically cleaned up if e.g. a test file is deleted or renamed. Build infos are still cleaned up correctly.

Closes #7116

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -84,6 +84,7 @@ export interface SolidityBuildSystemOptions {
   readonly soliditySourcesPaths: string[];
   readonly artifactsPath: string;
   readonly cachePath: string;
+  readonly solidityTestsPath: string;
 }
 
 export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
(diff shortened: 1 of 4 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -702,6 +702,20 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     )) {
       const relativePath = path.relative(this.#options.artifactsPath, file);
 
+      const testDirectorySubpath = path.relative(
+        this.#options.projectRoot,
+        this.#options.solidityTestsPath,
+      );
+      const hasTestFileExtension = file.endsWith(".t.sol");
+      const isInsideTestFolder = relativePath.startsWith(
+        testDirectorySubpath + path.sep,
```
fvictorio, line 711: Slightly safer, although probably unnecessary:

```suggestion
        path.join(testDirectorySubpath, path.sep),
```

fvictorio (approved):
Very minor comment, feel free to merge.

PR #7136, opened 2025-08-07
7114 global options arent displayed in the help message of a task

michalbrabec:
`GLOBAL OPTIONS:` section is displayed in the help message in all cases where the help message specifies that global options can be used (message contains `GLOBAL OPTIONS`)
For example: `Usage: hardhat [GLOBAL OPTIONS] <TASK> [SUBTASK] [TASK OPTIONS] [--] [TASK ARGUMENTS]`

This applies to all tasks, subtasks and global help
[Design doc](https://www.notion.so/nomicfoundation/Hardhat-CLI-help-247578cdeaf580b59adedc18f1d21af4?source=copy_link)

### DIFF
--- a/.changeset/neat-apes-beam.md
+++ b/.changeset/neat-apes-beam.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Display **GLOBAL OPTIONS** in help messages of all tasks that support them ([#7114](https://github.com/NomicFoundation/hardhat/issues/7114))
--- a/v-next/hardhat/test/internal/cli/help/get-global-help-string.ts
+++ b/v-next/hardhat/test/internal/cli/help/get-global-help-string.ts
@@ -66,7 +66,34 @@ To get help for a specific task run: npx hardhat <TASK> [SUBTASK] --help`;
           },
         ],
       ]);
-      const globalOptionDefinitions = new Map();
+
+      const globalOptionDefinitions = new Map([
+        [
+          "userOption1",
+          {
+            pluginId: "builtin",
+            option: globalOption({
+              name: "userOption1",
+              description: "userOption1 description.",
+              type: ArgumentType.STRING,
+              defaultValue: "default",
+            }),
+          },
+        ],
+        [
+          "userOption2",
+          {
+            pluginId: "builtin",
+            option: globalOption({
+              name: "userOption2",
+              shortName: "o",
+              description: "userOption2 description.",
+              type: ArgumentType.STRING,
+              defaultValue: "default",
+            }),
+          },
+        ],
+      ]);
 
       const help = await getGlobalHelpString(tasks, globalOptionDefinitions);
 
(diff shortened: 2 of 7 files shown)


.changeset/neat-apes-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+GLOBAL OPTIONS: section is displayed in the help message in all cases where the help message specifies that global options can be used (message contains GLOBAL OPTIONS)
```
kanej:
```suggestion
Display **GLOBAL OPTIONS** in help messages of all tasks that use them ([#7114](https://github.com/NomicFoundation/hardhat/issues/7114))
```
This is the text that will appear in the changelogs, and likely the release notes.

v-next/hardhat/test/internal/cli/help/get-global-help-string.ts
```diff
@@ -66,22 +70,34 @@ To get help for a specific task run: npx hardhat <TASK> [SUBTASK] --help`;
...
 
       const expected = `Hardhat version ${packageJson.version}
 
 Usage: hardhat [GLOBAL OPTIONS] <TASK> [SUBTASK] [TASK OPTIONS] [--] [TASK ARGUMENTS]
 
 AVAILABLE TASKS:
 
-  task1      task1 description
-  task2      task2 description
+  task1                    task1 description
```
kanej, line 106: My reading of this additional space, is that the addition of global options bumps the formatting algorithm? Is that right?

v-next/hardhat/src/internal/cli/help/utils.ts
```diff
@@ -9,7 +9,7 @@ import { camelToKebabCase } from "@nomicfoundation/hardhat-utils/string";
 
 export const GLOBAL_NAME_PADDING = 6;
 
-interface ArgumentDescriptor {
+export interface ArgumentDescriptor {
```
kanej: We are always careful about the types we export. We make a distinction between those that are available to consumers of the package (which are available at the top level) and those under `./internal`. Those under `./internal` are not exposed and we can manipulate and move without concern for the external API.
In this case I would be tempted to move this interface as its now shared among files to `./src/internal/cli/utils.ts` - a new file, and update the imports.
But I think it is borderline, so I will leave it to your judgement.

v-next/hardhat/test/internal/cli/help/get-global-help-string.ts
michalbrabec: Exactly so

v-next/hardhat/src/internal/cli/help/utils.ts
michalbrabec: I have removed the export and parse the parameters either in the `getHelpString` and `getGlobalHelpString` instead. These two are never called at the same time so there will be no lost perf.

.changeset/neat-apes-beam.md
michalbrabec: Updated

Issue #7132, opened 2025-08-06
Add GA event for project pick in `--init`


kanej:
Add a GA record for the project type picked during `--init`. This should follow the pattern of task invocation.

PR #7141, opened 2025-08-07
Send selected project type to GA

fvictorio:
Closes https://github.com/NomicFoundation/hardhat/issues/7132

The previous code assumed that all hits were task hits. I refactored it a bit to allow for different types of GA events with different payloads.


v-next/hardhat/src/internal/cli/telemetry/analytics/types.ts
```diff
@@ -54,10 +63,10 @@ export interface Payload extends BasePayload {
     };
   };
   events: Array<{
-    name: EventNames;
+    name: AnalyticsEvent["name"];
     params: {
       engagement_time_msec: string;
       session_id: string;
-    } & TaskParams;
+    } & AnalyticsEvent["params"];
```
fvictorio, line 70: This type _technically_ allows you to have an event that has `task` name but `init` params. I'm pretty sure there's a type-safe way to do this, but I didn't want to spend more time on this.

Issue #7117, opened 2025-08-04
Replace `import.meta.resolve`


kanej:
Don’t use `import.meta.resolve` because it’s not type-safe, and it doesn’t fail in compilation time.

## `HookHandlerCategoryFactories`:
Current:
            
```tsx
[HookCategoryNameT in keyof HardhatHooks]?:
    | HookHandlerCategoryFactory<HookCategoryNameT>
    | string;
```
            
New:
            
```tsx
[HookCategoryNameT in keyof HardhatHooks]?:
    | HookHandlerCategoryFactory<HookCategoryNameT>
    | () => Promise<{default: HookHandlerCategoryFactory<HookCategoryNameT>}>;
```
            
## `NewTaskDefinitionBuilder` and `TaskOverrideDefinitionBuilder`
Current:
            
```tsx
setAction(action: NewTaskActionFunction<TaskArgumentsT> | string): this;
```
            
New:
            
```tsx
interface LazyActionObject<T>{
  action: () => Promise<{ default: T }>;
};
```

```tsx
setAction(action: NewTaskActionFunction<TaskArgumentsT> | LazyActionObject<NewTaskActionFunction<TaskArgumentsT>>): this; 
```
            
## `HardhatPlugin`:
Current:
            
```tsx
dependencies?: Array<() => Promise<HardhatPlugin>>;
```
            
New:

```
dependencies?: () => Array<Promise<{default: HardhatPlugin}>>;
```

alcuadrado:
~Note that in some places we were warning when we were getting inline hooks and actions, that should still work. We made sure the proposed changes don't break that detection.~ We decided to remove the inline variants. All of them would be of the shape `() => import(...)`.

~Also, the way we load the plugin dependencies and run the detection of installation errors need to be updated. It used to be done one at the time, now it should be all at once.~ Done in a commit I pushed.

PR #7147, opened 2025-08-07
Replace import.meta.resolve

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/7117

### Questions:

1. Question related to [this comment](https://github.com/NomicFoundation/hardhat/issues/7117#issuecomment-3151803732) about `load the plugin dependencies`:

Are you referring to the function `reverseTopologicalSort`? Specifically to this (updated in this PR) code:

```
if (plugin.dependencies !== undefined) {
  for (const dependencyPromise of plugin.dependencies()) {
    const resolvedPlugin = await loadDependency(
      projectRoot,
      plugin,
      async () => (await dependencyPromise).default,
    );

    await dfs(resolvedPlugin);
  }
}
```

If yes, I do not think this can be parallelized, as the `dfs` need to build the dependency graph in a sync way

2. Now that `actions` are no longer imported via `string`, a common error that could happen is to do something like this:
```
.setAction(
  () => import("./task-action.js"))
```
instead of using the correct approach:
```
.setAction({
  action: () => import("./task-action.js"),
})
```
Should we try to prevent these accidental mistakes, if possible, to improve the user experience?


v-next/hardhat-ignition/src/internal/tasks/deploy.ts
```diff
@@ -30,9 +30,9 @@ import { verifyArtifactsVersion } from "../utils/verifyArtifactsVersion.js";
 
 interface TaskDeployArguments {
   modulePath: string;
-  parameters: string;
-  deploymentId: string;
-  defaultSender: string;
+  parameters?: string;
```
ChristopherDedominici, line 33: Marking these variables as `undefined` is necessary because stronger type enforcement is now in place. The import of `action` will detect any type mismatches.

For example, `parameters` has a default value of `undefined`, but in the deploy file, `parameters` is only marked as a `string`.


image

ChristopherDedominici: Note: the same logic is applied in subsequent files within this PR

v-next/hardhat/src/internal/core/config-validation.ts
```diff
@@ -272,10 +272,18 @@ export function validateNewTask(
...
+      typeof task.action === "object" &&
+      task.action !== null &&
+      "action" in task.action &&
+      typeof task.action.action === "function"
+    )
+  ) {
     validationErrors.push({
       path: [...path, "action"],
-      message: "task action must be a function or a string",
+      message: "task action must be a function or a lazy action object",
```
ChristopherDedominici: I need a double check: is this message ok?

v-next/hardhat/src/types/tasks.ts
```diff
@@ -136,9 +138,17 @@ export type TaskDefinition =
  **/
 export type ExtendTaskArguments<
   NameT extends string,
-  TypeT extends ArgumentType,
+  TypeT extends ArgumentType | ArgumentType[],
```
ChristopherDedominici, line 139: Why this type was changed?

The reason can be understood in this example:

image

As you can see, the expected argument type for `commands` in the code on the RIGHT is an array of strings.
The problem is that the resolved type in the `action` (code on the left) could not be determined correctly due to an error in our types, which is fixed in this change.
This error is surfacing now because the import for `actions` is now typed, whereas previously it was not

ChristopherDedominici: Note: addign
```
type: ArgumentType.STRING
```

will still not solve the type issue, so this change in the `type` is needed

v-next/hardhat/test/internal/core/hook-manager.ts
```diff
@@ -453,37 +453,6 @@ describe("HookManager", () => {
 
         assert.fail("Expected an error, but none was thrown");
       });
-
```
ChristopherDedominici, line 456: This test is no longer needed because a new logic is now in place, and we no longer check for non file paths

v-next/hardhat/test/internal/core/tasks/builders.ts
```diff
@@ -193,23 +203,6 @@ describe("Task builders", () => {
           },
         );
       });
-
```
ChristopherDedominici, line 196: This test is no longer needed because a new logic is now in place, and we no longer check for non file paths

v-next/hardhat/test/internal/core/tasks/builders.ts
```diff
@@ -1117,23 +1115,6 @@ describe("Task builders", () => {
...
-      it("should throw if the task action is not a valid file URL", () => {
```
ChristopherDedominici, line 1121: This test is no longer needed because a new logic is now in place, and we no longer check for non file paths

v-next/hardhat/test/internal/core/tasks/task-manager.ts
```diff
@@ -1046,61 +1046,50 @@ describe("TaskManagerImplementation", () => {
         );
       });
 
-      it("should throw if the task definition object has an invalid action file URL", async () => {
```
ChristopherDedominici, line 1049: This test is no longer needed because a new logic is now in place, and we no longer check for non file paths

v-next/hardhat/src/internal/core/tasks/validations.ts
```diff
@@ -23,6 +23,9 @@ export function validateId(id: string | string[]): void {
   }
 }
 
+//
+// TODO
+//
 export function validateAction(action: unknown): void {
```
ChristopherDedominici: Probably this method can be deleted now, since `strings` are no longer a valid type?

v-next/hardhat-ignition/src/index.ts
```diff
@@ -70,18 +63,24 @@ const hardhatIgnitionPlugin: HardhatPlugin = {
         description:
           "Write deployment information to disk when deploying to the in-memory network",
       })
-      .setAction(import.meta.resolve("./internal/tasks/deploy.js"))
+      .setAction({
+        action: () => import("./internal/tasks/deploy.js"),
```
alcuadrado: Let's only accept `() => import(...)`

alcuadrado: If needed, we can add `setInlineAction` in the future

v-next/hardhat/src/internal/core/tasks/resolved-task.ts
```diff
@@ -152,14 +153,14 @@ export class ResolvedTask implements Task {
         SHOULD_WARN_ABOUT_INLINE_TASK_ACTIONS_AND_HOOK_HANDLERS
       ) {
         console.warn(
-          `WARNING: Inline task action found in plugin "${pluginId}" for task "${formatTaskId(this.id)}". Use file:// URLs in production.`,
+          `WARNING: Inline task action found in plugin "${pluginId}" for task "${formatTaskId(this.id)}". Use the lazy import syntax in production.`,
```
alcuadrado: This warning is going away because we'll only accept one variant: `() => import(...)`

v-next/hardhat/src/internal/core/tasks/validations.ts
alcuadrado: Also goes away

v-next/hardhat/src/internal/core/config-validation.ts
alcuadrado: same

v-next/hardhat/src/internal/core/config-validation.ts
```diff
@@ -613,20 +629,39 @@ export function validatePluginsConfig(
...
         validationErrors.push({
           path: [...path, "plugins", index, "dependencies"],
-          message: "plugin dependencies must be an array",
+          message: "plugin dependencies must be a function returning an array",
         });
+      } else {
+        let result: unknown;
+
+        try {
+          result = plugin.dependencies();
```
alcuadrado: This also goes away, because otherwise we'll be loading the dependencies all the time. Same for actions and hook handlers. All we can validate is that they are functions.

v-next/hardhat/src/internal/core/hook-manager.ts
```diff
@@ -299,20 +299,19 @@ export class HookManagerImplementation implements HookManager {
 
           let hookCategory: Partial<HardhatHooks[HookCategoryNameT]>;
 
-          if (typeof hookHandlerCategoryFactory === "string") {
-            hookCategory = await this.#loadHookCategoryFactory(
-              plugin,
-              hookCategoryName,
-              hookHandlerCategoryFactory,
-            );
+          const maybeLazyLoad = await hookHandlerCategoryFactory();
```
alcuadrado: Let's also remove the inline hook handler. So you always do `() => import(...)`

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -617,18 +617,18 @@ Please ensure that an action is defined for each task.`,
         websiteDescription:
           "The task is empty. Please ensure that tasks have at least one action.",
       },
-      INVALID_ACTION_URL: {
+      INVALID_ACTION_IMPORT: {
         number: 411,
         messageTemplate:
-          'Unable to import the action specified by task "{task}" from the module "{action}"',
-        websiteTitle: "Invalid action URL",
+          'Unable to import the action for task "{task}" because the specified module could not be found.',
```
alcuadrado:
```suggestion
          'Unable to import the action for task "{task}".',
```

And the rest of the fields.

The reason is that the file may be present, but throw.

v-next/hardhat/src/internal/core/tasks/resolved-task.ts
ChristopherDedominici: Fixed here in PR on top of this one: https://github.com/NomicFoundation/hardhat/pull/7166/files#r2266672768

v-next/hardhat/src/internal/core/config-validation.ts
ChristopherDedominici: Fixed here in PR on top of this one: https://github.com/NomicFoundation/hardhat/pull/7166#discussion_r2266656136

v-next/hardhat/src/internal/core/hook-manager.ts
ChristopherDedominici: Fixed here in PR on top of this one: https://github.com/NomicFoundation/hardhat/pull/7166#discussion_r2266658007

alcuadrado (approved):

PR #7152, opened 2025-08-07
Hardhat local installation error

### LINKED ISSUE #5362: Hardhat local installation error
Hardhat can be installed as a global cli:

```sh
npm install -g hardhat
```

Running hardhat as if locally without a local installation should error:

```sh
> rm -rf ./node_modules
> npx hardhat --version

Error HHxxx: Trying to use a non-local installation of Hardhat, which is not supported.
Please install Hardhat locally using npm or Yarn, and try again.

For more info go to https://hardhat.org/HHxxx or run Hardhat with --show-stack-traces
```

antico5:
Closes #5362


v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -682,3 +702,52 @@ npm pkg set type="module"
...
+
+    if (!isLocalOrLinked) {
+      log("Determed that Hardhat is not installed locally/linked");
+      log(`  resolved package.json: ${resolvedPackageJson}`);
+      log(`  current package.json: ${thisPackageJson}`);
+    }
+
+    return isLocalOrLinked;
+  } catch (error) {
+    log("Error during installed locally/linked test", error);
```
kanej, line 740: Is there a better pattern for debug logging errors?

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -682,3 +702,52 @@ npm pkg set type="module"
...
+ * we use the current working directory.
+ */
+async function isHardhatInstalledLocallyOrLinked(
+  configPath: string,
+  log: debug.Debugger,
+) {
+  try {
+    // Setup a require to let us Node resolve based on the
+    // user's repo
+    const require = createRequire(configPath ?? process.cwd());
```
kanej: This appears to be the closest ESM equivalent to:

https://github.com/NomicFoundation/hardhat/blob/0a145098cd1f50f6b46fc17ae9a19d393c56d9d3/packages/hardhat-core/src/internal/core/execution-mode.ts#L11-L27

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -682,3 +696,52 @@ npm pkg set type="module"
...
+    // running with `node --preserve-symlinks`
+    const isLocalOrLinked =
+      (await getRealPath(resolvedPackageJson)) ===
+      (await getRealPath(thisPackageJson));
+
+    if (!isLocalOrLinked) {
+      log("Determed that Hardhat is not installed locally/linked");
+      log(`  resolved package.json: ${resolvedPackageJson}`);
+      log(`  current package.json: ${thisPackageJson}`);
+    }
```
kanej: I found the debug logging useful when trying to diagnose issues during testing.
There should be no debug clutter if the test passes.

v-next/hardhat/src/internal/cli/main.ts
alcuadrado: We have `findDependencyPackageJson` in `hardhat-utils`. Shouldn't that be enough?

kanej: I switched to `findDependencyPackageJson` and retested.

kanej:
We are going to hold on this for the moment. We want to leverage the version in utils - but need to check how we do that with the Hardhat tests.

PR #7154, opened 2025-08-08
Solidity Config Resolution Enhancements

schaable:
This PR improves how build profiles are created. Key changes:

- Removed unused profiles.  
- Added `outputSelection` to all `SolcConfig` settings to ensure consistency.  
- For the `production` profile, added optimization defaults to the `SolcConfig`.  
- Updated how production profiles are derived from the default profile, ensuring only the `version` is carried over.

### DIFF
--- a/v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
@@ -4,6 +4,7 @@ import { describe, it } from "node:test";
 
 import {
   resolveSolidityUserConfig,
+  shouldUseWasm,
   validateSolidityUserConfig,
 } from "../../../../src/internal/builtin-plugins/solidity/config.js";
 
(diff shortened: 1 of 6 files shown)


v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -482,11 +483,11 @@ describe("solidity plugin config resolution", () => {
 
       assert.equal(
         resolvedConfig.solidity.profiles.production.preferWasm,
-        true,
+        true && shouldUseWasm(),
```
kanej:
```suggestion
        shouldUseWasm(),
```
I can't think of a gain here from `true && shouldUseWasm()`

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -570,7 +571,7 @@ describe("solidity plugin config resolution", () => {
       );
       assert.equal(
         resolvedConfig.solidity.profiles.production.preferWasm,
-        true,
+        true && shouldUseWasm(),
       );
```
kanej:
```suggestion
```

I think this assertion and the reference to `production` above should be deleted - or we should reword the name of the test.

kanej (approved):

Issue #7161, opened 2025-08-08
`hh compile` fails without detailed error message if `/tmp/hardhat-solc` is not writable


KartikSoneji:
### Version of Hardhat

3.0.0-next.29

### What happened?

`npx hardhat compile` fails with a nondescriptive error message if `/tmp/hardhat-solc` is not writable.

The `spawnCompile()` function creates the folder, resulting it being owned by the current user.
If another user attempts compilation while the folder exists, it throws an error.
https://github.com/NomicFoundation/hardhat/blob/58a959853cccc037b3184d24d75525d5b4a58784/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/compiler.ts#L50-L58

`mkdir()` by default does create folders with mode `777`, but since the default umask is `022`, it results in a folder with `755`
(https://github.com/nodejs/node/issues/15092)
This can be fixed by explicitly using `chmod()` to set the permission after creation.

### Minimal reproduction steps

On a system that has more than one user:
1. compile project as user a
   ```sh
   sudo su user-a
   npx hardhat compile
   ```
2. switch to user b
   ```sh
   sudo su user-b
   ```
3. attempt to compile project as user b
   ```console
   $ npx hardhat compile
   Compiling your Solidity contracts...
   Error HHE907: A native version of solc failed to run.

   If you are running MacOS, try installing Apple Rosetta.

   If this error persists, run "npx hardhat clean --global".
   ```

### Search terms

compile error, /tmp/hardhat-solc not writable

fvictorio:
Hey @KartikSoneji, thanks a lot for the detailed issue description, this was super interesting.

The way I see it, the fix is to either `chmod` the created directory to have `777` permissions, or to use `/tmp/hardhat-solc-{uuid}` instead of `/tmp/hardhat-solc/{uuid}`.

michalbrabec:
Possible solutions:
- Create /tmp/hardhat-solc/ with writing permissions for everyone. I'm pretty sure this works and shouldn't have any security implications, but it makes me feel a bit nervous anyway.
- Create /tmp/hardhat-solc-{uuid} instead of /tmp/hardhat-solc/{uuid}. That way each workdir is independent and this problem doesn't happen.
- Stop using /tmp altogether. Using that directory has given us headaches in the past, so maybe we should always use the project's /cache directory (or /cache/tmp maybe?)

We selected option 2
We will migrate to option 3 as part of #7219

[...]

PR #7162, opened 2025-08-08
fix: `/tmp/hardhat/-solc` permissions to be world writable

KartikSoneji:
---
Fixes: #7161 

Use `chmod()` to set permissions to `777` on the `/tmp/hardhat-solc` folder, if possible, to ensure it's accessible.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/compiler.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/compiler.ts
@@ -16,6 +16,7 @@ import {
 } from "@nomicfoundation/hardhat-errors";
 import { ensureError } from "@nomicfoundation/hardhat-utils/error";
 import {
+  chmod,
   mkdir,
   readJsonFileAsStream,
   remove,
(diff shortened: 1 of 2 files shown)


alcuadrado:
Hey, thanks for this PR. I'll take a look soon.

Another alternative would be to stop using tmp dirs and use `<project>/cache/tmp` or something like that. So that we avoid these kind of problems altogether. WDYT?

KartikSoneji:
Hmm the issue is it would force everyone to update their `.gitignore` files.

alcuadrado:
but everyone has `cache/` in their gitignore files, we add it during initialization, and hardhat would be really annoying without it

KartikSoneji:
Oh ok.
Then that might be a better solution.
Should I make the changes?

alcuadrado:
if you want, that'd be cool.

I think in our own tests it's still fin to use tmp dirs. But in the actual code we should probably have a helper that creates them inside the cache dir. We have this:

```
export async function mkdtemp(prefix: string): Promise<string>;
```

Maybe it could be 

```
export async function mkdtemp(cointainer: string, nameHint: string): Promise<string> 
```

so that we call it with `hre.config.paths.cache` as container, and it creates a new folder base on the `nameHint`.

KartikSoneji:
Should it be `tmpDir` instead?

So instead of:
```ts
export async function mkdtemp(prefix: string): Promise<string> {
  try {
    return await getRealPath(
      await fsPromises.mkdtemp(path.join(tmpdir(), prefix)),
    );
  } catch (e) {
    ensureNodeErrnoExceptionError(e);
    throw new FileSystemAccessError(e.message, e);
  }
}
```

It can be:
```ts
export async function mkdtemp(
  prefix: string,
  tmpDir: string = tmpdir(),
): Promise<string> {
  try {
    return await getRealPath(
      await fsPromises.mkdtemp(path.join(tmpDir, prefix)),
    );
  } catch (e) {
    ensureNodeErrnoExceptionError(e);
    throw new FileSystemAccessError(e.message, e);
  }
}
```

kanej:
Hey @KartikSoneji, thanks for all the help here.
I am closing this PR as we tackled https://github.com/NomicFoundation/hardhat/issues/7161 with this PR: https://github.com/NomicFoundation/hardhat/pull/7220

PR #7188, opened 2025-08-12
ci: update regression tests

galargh:

---

Barring the OOM in `vectorized/solady`, this PR establishes the new baseline for regression tests where everything's passing - https://github.com/NomicFoundation/hardhat/actions/runs/16910142358

Unfortunately, we did have to exclude a lot more tests. The majority of errors in the tests that we had to exclude show up as reverts of a kind.

The other 2 groups that we should look into:
1. OOM in `vectorized/solady`
2. `the path * is not allowed to be accessed for read/write operations` which might have been broken in https://github.com/NomicFoundation/hardhat/pull/7119

Finally, we had to exclude `sablier-labs/lockup` from the `solidity-test` preset as the test execution there now hangs on, seemingly, forever. This likely has to do with the remappings that we had to introduce there for the compilation to pass `src/=src/` and `tests/=tests/`. Maybe this creates some sort of circular, infinite loop somewhere.

### DIFF
--- a/v-next/hardhat/src/internal/cli/init/prompt.ts
+++ b/v-next/hardhat/src/internal/cli/init/prompt.ts
@@ -164,4 +164,7 @@ function ensureTTY(): void {
       HardhatError.ERRORS.CORE.GENERAL.NOT_IN_INTERACTIVE_SHELL,
     );
   }
+
+  let a = "";
+  a = a.replaceAll(`\${${i}}`, i)
 }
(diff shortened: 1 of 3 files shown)


v-next/hardhat/src/internal/cli/init/prompt.ts
```diff
@@ -164,4 +164,7 @@ function ensureTTY(): void {
       HardhatError.ERRORS.CORE.GENERAL.NOT_IN_INTERACTIVE_SHELL,
     );
   }
+
+  let a = "";
+  a = a.replaceAll(`\${${i}}`, i)
```
kanej: Was this a dev test and left in?

galargh: Oh my 🤦

kanej (approved):

PR #7194, opened 2025-08-13
Do not throw HHE201 for optional peer dependencies

ItsNickBarry:

---

This plugin specifies `ethers` and `viem` as optional peer dependencies using the `peerDependenciesMeta` field: https://github.com/solidstate-network/hardhat-accounts/blob/hh3/package.json

Running Hardhat in a repository with the plugin installed threw this error: 

```
Error HHE201: Plugin "@solidstate/hardhat-accounts" is missing a peer dependency "viem".
```

This shouldn't happen because `viem` is optional.  I believe using `peerDependenciesMeta` is the correct way to specify optional dependencies (at least with pnpm), though I also noticed the `optionalDependencies` field in the `PackageJson` interface.  Maybe both should be supported?

### DIFF
--- a/v-next/hardhat-utils/src/package.ts
+++ b/v-next/hardhat-utils/src/package.ts
@@ -52,6 +52,7 @@ export interface PackageJson {
   devDependencies?: Record<string, string>;
   peerDependencies?: Record<string, string>;
   optionalDependencies?: Record<string, string>;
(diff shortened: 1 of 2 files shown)


ItsNickBarry:
Actually this was caused by something else, unrelated to the error message.

Issue #7185, opened 2025-08-12
cli help should consistently end with full stop or not


kanej:
In our task help and global options we are inconsistent with the use of a full stop.

<img width="698" height="185" alt="Image" src="https://github.com/user-attachments/assets/50ba90d6-da85-4e15-8452-3e15f0b09ba5" />

We should remove the full stops.

PR #7195, opened 2025-08-13
Remove full stops from help message descriptions

zoeyTM:
closes https://github.com/NomicFoundation/hardhat/issues/7185

### DIFF
--- a/v-next/hardhat/test/internal/cli/help/utils.ts
+++ b/v-next/hardhat/test/internal/cli/help/utils.ts
@@ -92,6 +92,44 @@ describe("utils", function () {
         ],
       });
     });
+
+    it("should remove a trailing full stop from tasks and subtasks descriptions", function () {
+      const task: Task = {
+        id: ["task"],
+        description: "task description.",
+        actions: [
+          {
+            pluginId: "task-plugin-id",
+            action: async () => ({
+              default: () => {},
+            }),
+          },
+        ],
+        options: new Map(),
+        positionalArguments: [],
+        pluginId: "task-plugin-id",
+        subtasks: new Map().set("subtask", {
+          id: ["task", "subtask"],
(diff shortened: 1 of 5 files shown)


antico5:
My first approach would have been to just modify the description strings, as I think that'd be less complexity. I think your approach can have some upsides though, like covering us or third party developers in the case we introduce full stops on the descriptions. Not sure the tradeoff is worth it or not, but I'll not make any call on it, just giving another perspective. 

On the current approach I'll leave just a small comment

v-next/hardhat/test/internal/cli/help/utils.ts
```diff
@@ -209,6 +282,84 @@ describe("utils", function () {
...
+            action: async () => ({
+              default: () => {},
+            }),
+          },
+        ],
+        options: new Map()
+          .set("option", {
+            name: "option",
+            shortName: "o",
+            description: "An example option",
```
antico5: Should these example options and arguments have a trailing dot?

zoeyTM:
> My first approach would have been to just modify the description strings, as I think that'd be less complexity. I think your approach can have some upsides though, like covering us or third party developers in the case we introduce full stops on the descriptions. Not sure the tradeoff is worth it or not, but I'll not make any call on it, just giving another perspective.
> 
> On the current approach I'll leave just a small comment

I wrote it this way because anyone can write their own plugins which include their own tasks with their own descriptions, so this lets us maintain a uniform help text for any of those community plugins as well as our own.

v-next/hardhat/test/internal/cli/help/utils.ts
zoeyTM: Whoops, good catch!

antico5 (approved):

PR #7205, opened 2025-08-13
fix: update the validation to reflect the new property

kanej:
This property was renamed from `solidityTest` to `test`, but this change was missed.

See https://github.com/NomicFoundation/hardhat/pull/7101 for context.


fvictorio:
Shouldn't we have a test for this kind of thing?

PR #7216, opened 2025-08-14
Agostbiro/fix soltest config options

agostbiro:

---

Added missing Solidity test config option `allowInternalExpectRevert` and fixed `timeout` config option not being passed in to EDR. 

Note that EDR supports the following config options that are not exposed by Hardhat, but I'm not sure if these are omitted on purpose, so I didn't add them:

1. `blockNumber` - The value of block.number in tests. Defaults to 1.
1. `chainId` - The value of the chainid opcode in tests. Defaults to 31337.
1. `gasLimit` - The gas limit for each test case. Defaults to 9_223_372_036_854_775_807 (i64::MAX). (The docs have blockGasLimit but it's described differently)
1. `gasPrice` - The price of gas (in wei) in tests. Defaults to 0.
1. `blockDifficulty` - The value of block.difficulty in tests. Defaults to 0.
1. `memoryLimit` - The memory limit of the EVM in bytes. Defaults to 33_554_432 (2^25 = 32MiB).

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
@@ -98,6 +98,7 @@ export function solidityTestConfigToSolidityTestRunnerConfigArgs(
   const ethRpcUrl = config.forking?.url;
   const forkBlockNumber = config.forking?.blockNumber;
   const rpcEndpoints = config.forking?.rpcEndpoints;
+  const promptTimeout = config.timeout;
 
   return {
     projectRoot,
@@ -116,6 +117,7 @@ export function solidityTestConfigToSolidityTestRunnerConfigArgs(
     ethRpcUrl,
     forkBlockNumber,
     rpcEndpoints,
+    promptTimeout
   };
 }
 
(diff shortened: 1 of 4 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
```diff
@@ -98,6 +98,7 @@ export function solidityTestConfigToSolidityTestRunnerConfigArgs(
   const ethRpcUrl = config.forking?.url;
   const forkBlockNumber = config.forking?.blockNumber;
   const rpcEndpoints = config.forking?.rpcEndpoints;
+  const promptTimeout = config.timeout;
```
galargh: `config.timeout` refers to the timeout for the entire run, see https://nomicfoundation.slack.com/archives/C03P6B72ZHU/p1755079564328649

We might want to add `promptTimeout` to the config if we want to support it.

agostbiro: Thanks, fixed in https://github.com/NomicFoundation/hardhat/pull/7216/commits/7cfc025e1b269b713090293d5c5851105a44b229

PR #7220, opened 2025-08-14
Use random non-nested temp dirs during compilation

### LINKED ISSUE #7161: `hh compile` fails without detailed error message if `/tmp/hardhat-solc` is not writable

fvictorio:
Closes #7161.

See comments inline.

### DIFF
--- a/.changeset/six-rivers-flow.md
+++ b/.changeset/six-rivers-flow.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed an issue with compiling a Hardhat project after switching OS users (#7161)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/compiler.ts
```diff
@@ -54,8 +54,7 @@ async function spawnCompile(
 ): Promise<CompilerOutput> {
   // We create a temporary folder to store the output of the compiler in
   // We use a random UUID to avoid collisions with other compilations
-  const tmpFolder = path.join(os.tmpdir(), "hardhat-solc", crypto.randomUUID());
-  await mkdir(tmpFolder);
+  const tmpFolder = await mkdtemp("hardhat-solc-");
```
fvictorio, line 57: This technically increases the chances of collision, but 6 alphanumerical characters still mean 56_800_235_584 combinations. I think it's fine for the use case.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/compiler.ts
```diff
@@ -177,8 +176,7 @@ export class NativeCompiler implements Compiler {
         args.push("--no-import-callback");
       } else if (semver.gte(this.version, "0.6.9")) {
         // version >= 0.6.9
-        const tmpFolder = path.join(os.tmpdir(), "hardhat-solc");
-        await mkdir(tmpFolder);
+        const tmpFolder = await mkdtemp("hardhat-solc-");
```
fvictorio, line 179: I also made this change because the assumption here is that `tmpFolder` is empty, but the other part of the code was creating directories under `/tmp/hardhat-solc`. While unlikely to be a problem in practice, I still think it's better to just use a unique random empty directory each time.

alcuadrado:
Adding this to the merge queue as franco is OOO

PR #7225, opened 2025-08-14
Spawn hardhat node programatically + hardhat-ethers TODO cleanup

### LINKED ISSUE #6472: Option to create custom node startup
### Feedback

For my project, I want to have a custom way to create a Hardhat node that is instrumented with my own logic instead of the Hardhat one (specifically, we want opentelemetry support)

# What worked in Hardhat v2

In hardhat v2, this was possible with either:
- `TASK_NODE_CREATE_SERVER` (as mentioned [here](https://github.com/NomicFoundation/hardhat/issues/872#issuecomment-859139329)
- By simply doing `import { JsonRpcServer } from "hardhat/internal/hardhat-network/jsonrpc/server.js";`

# Why it doesn't work in Hardhat v3

In Hardhat v3, this is not possible:
1. There is not only a monlithic `node` which not only does too much for my use-case, it also has no return type
2. Doing `import { JsonRpcServer } from "hardhat/internal/builtin-plugins/node/json-rpc/server.ts";` doesn't work because the path is not publicly exported

I can copy-paste the contents as-is, but that's not ideal since this folder also depends on `builtin-plugins/network-manager` as well

### LINKED ISSUE #6819: A helper to create a `localhost` endpoint must be implemented to achieve 100% coverage of the account private keys. Link

### LINKED ISSUE #6882: `hardhat-ethers` cleanup
Resolve the inline todos across the code and tests in `hardhat-ethers`.

antico5:
- Added `createServer` to NetworkManager for creating a JsonRpcServer
- This allows many skipped tests on the hardhat-ethers package to be enabled, because they required an http node

Closes #6819
Closes #6882
Closes #6472

### DIFF
--- a/v-next/hardhat/src/types/network.ts
+++ b/v-next/hardhat/src/types/network.ts
@@ -54,11 +54,13 @@ export interface NetworkManager {
   ): Promise<NetworkConnection<ChainTypeT>>;
 
   /**
-   * Spawn an Ethereum JSON-RPC server listening on HTTP and Websocket.
+   * Spawns an Ethereum JSON-RPC server listening on HTTP and Websocket.
    *
    * @param networkOrParams The network name or connection parameters.
    * @param hostname Hostname to bind the server to. Defaults to localhost or 0.0.0.0 on docker.
    * @param port Port to listen on. Defaults to a random available port.
+   *
+   * @return A `JsonRpcServer` instance that can be started with {@link JsonRpcServer.listen}.
    */
   createServer(
     networkOrParams?: NetworkConnectionParams | string,
@@ -86,7 +88,7 @@ export interface JsonRpcServer {
   /**
    * Starts the Ethereum JSON-RPC server.
    *
-   * @returns returns the address and port the server is listening on
+   * @returns returns the address and port the server is listening on.
    */
   listen(): Promise<{ address: string; port: number }>;
 


alcuadrado:
This unfortunately got lost in the mess getting to the release. We were discussing solutions here: https://github.com/NomicFoundation/hardhat/issues/6472

And during a call we discussed that this one in particular would be good:

> Expose an API that initializes a JsonRpcServerImplementation as part of NetworkManager


Imagine `const server = await network.createServer(....)` that has all the same params of `connect()`. Same behavior, but returns a server object.

I think your code still works, and it's a matter of moving the apis around.

.changeset/hot-turkeys-pull.md
```diff
@@ -0,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-ethers": patch
+"hardhat": patch
+---
+
+Enable to spawn hardhat node programmatically and integrate with hardhat-ethers tests ([#6819](https://github.com/NomicFoundation/hardhat/issues/6819)) ([#6882](https://github.com/NomicFoundation/hardhat/issues/6882))
```
kanej:
```suggestion
Added `network.createServer(...)` to spawn a node programmatically ([#6819](https://github.com/NomicFoundation/hardhat/issues/6819)) ([#6472](https://github.com/NomicFoundation/hardhat/issues/6472))
```

v-next/hardhat/src/types/network.ts
```diff
@@ -65,3 +71,9 @@ export interface NetworkConnection<
 
   close(): Promise<void>;
 }
+
+export interface JsonRpcServer {
+  listen(): Promise<{ address: string; port: number }>;
+  waitUntilClosed(): Promise<void>;
+  close(): Promise<void>;
+}
```
kanej: Can I check that it is our intention to expose `waitUntilClosed`? We use it in one place the `npx hardhat node` task to ensure everything is fully closed before finishing the task.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -97,6 +99,20 @@ export class NetworkManagerImplementation implements NetworkManager {
     return networkConnection as NetworkConnection<ChainTypeT>;
   }
 
+  public async createServer(
+    networkOrParams: NetworkConnectionParams | string = "default",
+    hostname: string = "127.0.0.1",
+    port?: number,
```
kanej, line 105: We are intentionally choosing not to set a default port right?
My reading of this is: allow undefined as port all the way to Node's server.listen where it will choose a random ephemeral port.

antico5: That's right. It came handy because e.g. on parallel tests, it will use different ports per process. If we wanted to use a default port, we could move the responsibility of choosing a random free port to the caller, depending on the use case. I think this way might be a bit simpler

v-next/hardhat/src/types/network.ts
antico5: Good point, I'll remove it from the interface, I don't think it's necessary

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
kanej: That makes sense on the port. We need to tweak the hostname so that it defaults to "0.0.0.0" in docker:

https://github.com/NomicFoundation/hardhat/blob/8b187472e4d16e2a1a795bdc098c5195b7d835e4/packages/hardhat-core/src/builtin-tasks/node.ts#L308-L320

v-next/hardhat/src/types/network.ts
kanej: I asked Pato on this one. `close` instructs the two sub-servers (http and websocket) to close. `waitUntilClosed` returns a promise, you can await that will only resolve when someone else closes the server.
`waitUntilClosed` is only used in the `npx hardhat node` task, as that is what we want to do here, kick off a server but only finish up the task when something triggers a close on the daemon.

Basically the problem is naming - none of that is obvious to me in reading the interface. I assumed that `waitUntilClose` triggered a closing of the server - it doesn't.

Chatgpt proposes an alternative interface: Use a closed Promise property (like WHATWG streams) plus the familiar listen/close:

```ts
export interface JsonRpcServer {
  listen(): Promise<{ address: string; port: number }>;
  close(): Promise<void>;

  /** Resolves once the server has closed. */
  readonly closed: Promise<void>;
}
```

Usage:

```ts
const { port } = await server.listen();
console.log(`listening on :${port}`);
await server.closed; // waits without implying it triggers close
```

What do you think?

At a minimum we need a better name for `waitUntilClose`, and JSDoc comments on each method of the interface.

alcuadrado: nitpick: but I'd rather do `closed()` than having a "floating promise" assigned to an object. 

What happens if it rejects and nobody awaits it? I should crash the system. 

So as an user of that api I'd be somewhat scared/confused about it. I guess we could promise that it never rejects, but that just adds confusion over creating a new promise every time.

kanej: To be explicit, @alcuadrado your proposal is: 

```
export interface JsonRpcServer {
  listen(): Promise<{ address: string; port: number }>;
  close(): Promise<void>;
  closed(): Promise<void>;
}
```

I think that is too alike (I almost said **close**). Maybe `afterClosed()`?

alcuadrado: > I think that is too alike (I almost said close)

Agreed.

TBH, I have no strong preference on the actual name. `afterClosed()` works for me too. `waitUntilClosed()` with good jsdocs too.

kanej: Cool. Let's go with `afterClosed` and JSDocs on the JsonRpcServer and the new `createServer` method.

PR #7236, opened 2025-08-15
chore: add testFail to supported solidity test config options

galargh:

---

The projects currently tested in the regression test suite require `testFail` option.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/type-extensions.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/type-extensions.ts
@@ -24,6 +24,7 @@ declare module "../../../types/test.js" {
     isolate?: boolean;
     ffi?: boolean;
     allowInternalExpectRevert?: boolean;
+    testFail?: boolean;
     from?: string; // 0x-prefixed hex string
     txOrigin?: string; // 0x-prefixed hex string
     initialBalance?: bigint;
--- a/.changeset/dirty-experts-march.md
+++ b/.changeset/dirty-experts-march.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Added `testFail` to supported solidity test config options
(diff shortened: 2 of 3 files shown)


agostbiro:
The `testFail` option was removed in Foundry 1.0, so if we don't support it already, I don't think we should add it now. Wdyt @fvictorio ?

v-next/hardhat/src/internal/builtin-plugins/solidity-test/type-extensions.ts
```diff
@@ -24,6 +24,7 @@ declare module "../../../types/test.js" {
     isolate?: boolean;
     ffi?: boolean;
     allowInternalExpectRevert?: boolean;
+    testFail?: boolean;
```
michalbrabec, line 27: Would be possible to add tests for the change?

galargh: Definitely! Any preference for what you'd like tested? The existence of the option, its E2E behaviour?

We shoudld resolve this comment first though - https://github.com/NomicFoundation/hardhat/pull/7236#issuecomment-3191406227

michalbrabec: I would test the existence of the option and then a behaviour with the option set to `true` and `false`, basically capture the difference.

michalbrabec: Agreed, https://github.com/NomicFoundation/hardhat/pull/7236#issuecomment-3191406227 should go first

galargh:
> The `testFail` option was removed in Foundry 1.0, so if we don't support it already, I don't think we should add it now. Wdyt @fvictorio ?

Following up on that in HH sync, we decided not to add `testFail` to the config. I'm going to close this down and update the regression test suite with an explicit reason for excluding `testFail` cases. I also created an issue for providing a better warning when we find testFail test cases - https://github.com/NomicFoundation/hardhat/issues/7253

PR #7238, opened 2025-08-15
docs(errors): align applyErrorMessageTemplate JSDoc with current implementation

Forostovec:
- Updated v-next/hardhat-errors/src/errors.ts JSDoc to describe {variable} placeholders instead of %variable%.
- Documented exact replacement rules and value conversions:
- undefined → "undefined", null → "null", bigint → "<value>n", arrays → JSON.stringify(value), others → value.toString().
- Clarified that extra keys in values are ignored and that %% escape is not supported.
- Noted that tags inside value strings are not re-interpreted.


kanej:

Issue #7235, opened 2025-08-15
`loggingEnabled` is ignored by EDR


michalbrabec:
### Version of Hardhat

3.0.0

### What happened?

logging enabled, a setting in edr networks, doesn’t seem to work

### Minimal reproduction steps

Test EDR network with logging enabled by
```typescript
...
networks: {
    hardhatMainnet: {
      type: "edr-simulated",
      chainType: "l1",
      loggingEnabled: true, // enable EDR logs
    },
...
```
Can be triggered by `pnpm hardhat ignition deploy ignition/modules/Counter.ts --network hardhatMainnet`

ChristopherDedominici:
Task created to track the lack of tests for EDR configuration: https://github.com/NomicFoundation/hardhat/issues/7250

PR #7239, opened 2025-08-15
loggingEnabled is ignored by EDR

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/7235

Task created to track the lack of tests for EDR configuration: https://github.com/NomicFoundation/hardhat/issues/7250

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
@@ -179,7 +179,7 @@ export class EdrProvider extends BaseProvider {
         hardhatChainTypeToEdrChainType(networkConfig.chainType),
         providerConfig,
         {
-          enable: loggerConfig.enabled,
+          enable: loggerConfig.enabled || networkConfig.loggingEnabled,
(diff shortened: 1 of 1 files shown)


michalbrabec:
Tested by `pnpm hardhat run scripts/send-op-tx.ts` with the `send-op-tx.ts` changed to network `edrOp` defined in`hardhat.config.json` as:
```typescript
    edrOp: {
      type: "edr-simulated",
      chainType: "op",
      chainId: 10,
      loggingEnabled: true,
    },
```

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -179,7 +179,7 @@ export class EdrProvider extends BaseProvider {
         hardhatChainTypeToEdrChainType(networkConfig.chainType),
         providerConfig,
         {
-          enable: loggerConfig.enabled,
+          enable: loggerConfig.enabled || networkConfig.loggingEnabled,
```
michalbrabec, line 182: Can we add a test for the flag so we guard agains regressions?

ChristopherDedominici: Added comment in PR description to track this

PR #7242, opened 2025-08-16
fix: broken links in plugin READMEs

cypherpepe:
Hi! While reading through the docs for hardhat-verify, hardhat-viem and hardhat-keystore I noticed that some links still point to the old hardhat3-alpha pages. Those URLs are no longer valid and give 404s.


kanej (approved):
Great catch, thanks!

Issue #7232, opened 2025-08-15
Compiler tests fail on ARM64 Linux


michalbrabec:
### Version of Hardhat

3.0.0

### What happened?

compiler related tests fail in Linux arm64 because of the new binaries that we started downloading from our mirror

### Minimal reproduction steps

Using ARM64 Linux, run the following commands in the hardhat repo root 
- `pnpm build`
- `pnpm test`

PR #7247, opened 2025-08-18
ci: test on Linux ARM and deprecate Intel MacOS

galargh:

---

Resolves https://github.com/NomicFoundation/hardhat/issues/7232#event-19198518774

Linux ARM machines are now free for public repos - https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

While Intel MacOS machines are getting deprecated in September - https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/

### DIFF
--- a/.github/workflows/v-next-ci.yml
+++ b/.github/workflows/v-next-ci.yml
@@ -290,7 +290,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
         # We run the tests for `hardhat-node-test-reporter` with the latest
         # versions of Node (the testing output relying on recent changes)
         node: [22, 24]
--- a/.github/workflows/v-next-ci.yml
+++ b/.github/workflows/v-next-ci.yml
@@ -239,7 +239,7 @@ jobs:
       fail-fast: false
       matrix:
         package: ${{ fromJson(needs.list-packages.outputs.packages) }}
-        os: [ubuntu-latest, macos-13, macos-latest, windows-latest]
+        os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
         node: [22.0.0, 24.0.0]
         exclude:
           - package: hardhat-node-test-reporter
@@ -290,7 +290,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-13, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
         # We run the tests for `hardhat-node-test-reporter` with the latest
         # versions of Node (the testing output relying on recent changes)
         node: [22, 24]
--- a/v-next/hardhat-ignition-core/test/helpers/execution-result-fixtures.ts
+++ b/v-next/hardhat-ignition-core/test/helpers/execution-result-fixtures.ts
@@ -4,9 +4,19 @@
  * @file
  */
 
+import os from "node:os";
+
 import { Artifact } from "../../src/types/artifact.js";
 
 import { RawStaticCallResult } from "../../src/internal/execution/types/jsonrpc.js";
+
+// NOTE: Compiler's long version string is an input to the build info ID calculation.
+// Linux ARM64 is a special case because the mirror we use doesn't advertise long
+// version information and we fall back to the short version string.
+const buildInfoId =
+  os.platform() === "linux" && os.arch() === "arm64"
+    ? "solc-0_8_19-212d1ae643d27f15734ceaa4108859c2f104c4ca"
+    : "solc-0_8_19-2fe1cba8ef218c727cd1ce3da3aaa170f8016338";
 
 export const staticCallResultFixtures: {
   [contractName: string]: { [functionName: string]: RawStaticCallResult };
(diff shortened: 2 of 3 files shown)


v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -39,29 +39,29 @@ describe(
...
-            version: "0.4.12a",
+            version: "0.5.0a",
           },
         );
       });
 
       it("should return false when the compiler isn't downloaded yet", async function () {
         assert.ok(
-          !(await wasmDownloader.isCompilerDownloaded("0.4.12")),
+          !(await wasmDownloader.isCompilerDownloaded("0.5.0")),
```
galargh, line 52: 0.5.0 is the earliest version supported by the mirror we use for Linux ARM64 builds. See https://solc-linux-arm64-mirror.hardhat.org/linux/aarch64

.github/workflows/v-next-ci.yml
```diff
@@ -290,7 +290,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-13, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
galargh: We should add Linux ARM64 here, too.

```suggestion
        os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
```

v-next/hardhat-ignition-core/test/helpers/execution-result-fixtures.ts
```diff
@@ -4,10 +4,20 @@
...
 import { Artifact } from "../../src/types/artifact.js";
 
 import { RawStaticCallResult } from "../../src/internal/execution/types/jsonrpc.js";
 
+// NOTE: Compiler's long version string is an input to the build info ID calculation.
+// Linux ARM64 is a special case because the mirror we use doesn't advertise long
+// version information and we fall back to the short version string.
+const buildInfoId =
+  os.platform() === "linux" && os.arch() === "arm64"
+    ? "solc-0_8_19-212d1ae643d27f15734ceaa4108859c2f104c4ca"
```
michalbrabec, line 17: Can we unify on the min version that supports all platforms?

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -435,31 +435,35 @@ describe(
           },
         );
 
-        await mockDownloader.updateCompilerListIfNeeded(new Set(["0.4.12"]));
+        await mockDownloader.updateCompilerListIfNeeded(new Set(["0.5.0"]));
 
+        const remoteVersion =
+          platform === CompilerPlatform.LINUX_ARM64
+            ? "0.5.0"
+            : "0.5.0+commit.1d4f565a";
```
michalbrabec, line 443: Why is a more detailed version used outside of ARM64 Linux?
Can we use a single version that supports all platforms (0.5.0)?

galargh: We cannot because of https://github.com/NomicFoundation/hardhat/blob/998165ac300cf7c7c142b30f32c00e4ec69416e0/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts#L357

v-next/hardhat-ignition-core/test/helpers/execution-result-fixtures.ts
galargh: We're already using a single version - 0.8.19 - here. The difference comes from https://github.com/NomicFoundation/hardhat/blob/998165ac300cf7c7c142b30f32c00e4ec69416e0/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts#L357 (as per the comment).

v-next/hardhat-ignition-core/test/helpers/execution-result-fixtures.ts
```diff
@@ -4,10 +4,20 @@
...
 
+import os from "node:os";
+
 import { Artifact } from "../../src/types/artifact.js";
 
 import { RawStaticCallResult } from "../../src/internal/execution/types/jsonrpc.js";
 
+// NOTE: Compiler's long version string is an input to the build info ID calculation.
+// Linux ARM64 is a special case because the mirror we use doesn't advertise long
+// version information and we fall back to the short version string.
```
kanej:
```suggestion
// NOTE: We are testing against a known build info id. As input to the build info // ID calculation we use the Compiler's long version string if available, and the // short version string if it is not.
// On Linux ARM64 we download from a mirror that does not provide the long version, hence the known build info id has to change when running the test suite on that platform.
```

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
michalbrabec: Agreed

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
michalbrabec: Could we use a higher version that is common to all platforms, let's say `0.5.1` or something like that?

galargh: I do not understand the question. Could you try rephrasing? AFAICT, we're already using a single version.

PR #7251, opened 2025-08-19
improve exception filtering before sending to Sentry

### LINKED ISSUE #7246: Better filtering for Sentry exceptions
Re-enable a filter to ensure only Hardhat errors are reported:

Image

## TODO

- [x] Fix bug with fetching hardhat config lazily
- [ ] Add back in exception filter, but do it before anonymization
- [ ] Add filter for require exceptions
- [ ] Review other heuristics

kanej:
Our Sentry exception reporting is including exceptions coming from within user scripts and tests. We need to improve our filtering to reduce the number of exceptions being reported that are really user land issues rather than Hardhat issues.

We intend to do this over a serious of tightening iterations. At the first pass we re-enable a filter and explicitly exclude:

* require in ESM exceptions
* cannot find package exceptions
* cannot find modules exceptions

We also broadly exclude exceptions originating in a `hardhat.config.{js,ts}` or that originate in a `node_modules` package that is not a hardhat package.

We attempt to exclude errors originating in user project files e.g. `./scripts/run.js` or `./ignition/modules/my-module.ts`.

Once we have established this baseline we will tighten up the definition until Sentry is only getting exceptions that are a strong signal that something unexpected has happened in Hardhat code.

Resolves #7246.

## Reviewer Note

> [!IMPORTANT]
> This PR is easier to review commit at a time

## TODO

- [x] Fix bug with fetching hardhat config lazily
- [x] Provide a custom replace for config path to help debugging
- [x] Add back in exception filter, but do it before anonymization
- [x] Review other heuristics
- [x] Add filter for require in ESM exceptions
- [x] Add filter for "Cannot find package" errors
- [x] Exemption for error messages starting with “Cannot find module” based on the examples enough
- [x] Re-introduce an equivalent of frame.filename === ANONYMIZED

### DIFF
--- a/v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
+++ b/v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
@@ -37,7 +37,6 @@ export type AnonymizeEventResult =
   | { success: true; event: Event }
   | { success: false; error: string };
 
-const ANONYMIZED_FILE = ANONYMIZED_PATH;
 const ANONYMIZED_MNEMONIC = "<mnemonic>";
 const MNEMONIC_PHRASE_LENGTH_THRESHOLD = 7;
 const MINIMUM_AMOUNT_OF_WORDS_TO_ANONYMIZE = 4;
@@ -57,7 +56,7 @@ export class Anonymizer {
   ): Promise<AnonymizeEnvelopeResult> {
     for (const item of envelope[1]) {
       if (item[0].type === "event") {
-        /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions 
+        /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
           -- We know that the item is an event item */
         const eventItem = item as EventItem;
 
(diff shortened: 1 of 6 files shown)


v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -166,38 +215,47 @@ export class Anonymizer {
...
 
+      // We don't report errors originating from non-Hardhat packages
       if (
-        this.#configPath !== undefined &&
-        this.#configPath.includes(frame.filename)
+        this.#isPackageFile(frame.filename) &&
+        !this.#isHardhatFile(frame.filename)
       ) {
         return false;
       }
```
kanej: There is an argument to say this is too aggresive at the first pass, and we should disable this heuristic for the moment.

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -166,38 +215,47 @@ export class Anonymizer {
...
+
+      // TODO: For the moment we are loose and treat as Hardhat errors
+      // anything not caught as by the explicit opt-outs above.
+      // Right now this will include files from the user's project
+      // (i.e. js/ts files not under `./node_modules`)
+      // We will tighten these up based on Sentry examples.
+      return true;
```
kanej: Right now we are not filtering out errors where the originating file is outside of `node_modules` i.e. a script or ignition module or such.

v-next/hardhat/src/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -212,15 +270,20 @@ export class Anonymizer {
...
+
     filename = filename.toLowerCase();
 
     return (
-      filename.startsWith(nomicFoundationPath) ||
-      filename.startsWith(ignoredOrgPath) ||
-      filename.startsWith(hardhatPath)
+      filename.includes(nomicFoundationPath) ||
+      filename.includes(ignoredOrgPath) ||
+      filename.includes(hardhatPath)
```
kanej, line 310: I changed this to includes as we are operating against the absolute path, rather than the anonymized version of the path that removed everything before the project directory.

v-next/hardhat/src/internal/cli/telemetry/sentry/subprocess.ts
```diff
@@ -24,7 +24,9 @@ if (process.argv.length !== 7) {
   process.exit(1);
 }
 
+log("Config path:", configPath);
 log("Received envelope to be sent to Sentry from a subprocess");
+log(JSON.stringify(JSON.parse(serializedEnvelope), null, 2));
```
kanej: I have kept this in - arguably it is quite noisy for users if we ask them to enable debugging on their failing error. What do you think?

v-next/hardhat/test/internal/cli/telemetry/sentry/anonymizer.ts
```diff
@@ -399,25 +417,50 @@ describe("Anonymizer", () => {
...
 
   describe("raisedByHardhat", () => {
-    function createTestEvent(filePath: string) {
+    const projectRoot = path.resolve(
+      path.dirname(fileURLToPath(import.meta.url)),
+      "../../../../../../../my-project",
+    );
+
+    const hardhatFile = path.join(
+      projectRoot,
```
kanej, line 426: I am adding the project root in here to simulate a full absolute path - previously the test was using examples that assumed an anonymized version.

v-next/hardhat/src/internal/cli/telemetry/sentry/subprocess.ts
alcuadrado: Oops, that was added for debugging 😅 I think they can be removed.

kanej: Will do, handy but not for standard users trying to debug an issue.

Issue #7233, opened 2025-08-15
The inter process mutex hangs


michalbrabec:
### Version of Hardhat

3.0.0

### What happened?

the inter process mutex hangs for too long after the process that created it died. That situation could be handled better by storing the pid of the process that created it and checking if it’s still running

When we download the `solc` binaries and binary list (and index of the existing compilers) we use a lock file as a mutex and if you kill the process in the middle, the lock file is left there and takes about 1min to get released while it could be almost instant

### Minimal reproduction steps

TBD

ChristopherDedominici:
While fixing this issue, I noticed a potential race condition. I’d classify it as low risk, so it should be addressed but isn’t very urgent: https://github.com/NomicFoundation/hardhat/issues/7270

PR #7263, opened 2025-08-20
The inter process mutex hangs

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/7233

### DIFF
--- a/v-next/hardhat-utils/src/synchronization.ts
+++ b/v-next/hardhat-utils/src/synchronization.ts
@@ -13,8 +13,9 @@ import path from "node:path";
 
 import debug from "debug";
 
-import { ensureNodeErrnoExceptionError } from "./error.js";
+import { ensureError, ensureNodeErrnoExceptionError } from "./error.js";
 import { FileSystemAccessError } from "./errors/fs.js";
+import { readUtf8File } from "./fs.js";
 import { sleep } from "./lang.js";
 
 const log = debug("hardhat:util:multi-process-mutex");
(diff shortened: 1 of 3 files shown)


v-next/hardhat-utils/src/synchronization.ts
```diff
@@ -123,4 +136,28 @@ export class MultiProcessMutex {
...
+      // If the file doesn't exist, it means the owning process deleted it
+      mutexPid = await readUtf8File(this.#mutexFilePath);
+    } catch (_e) {
+      return false;
+    }
+
+    try {
+      process.kill(parseInt(mutexPid, 10), 0);
+    } catch (e) {
+      ensureError<Error & { code: string }>(e);
```
schaable: nit: you can use `ensureNodeErrnoExceptionError` instead of passing `Error & { code: string }` as a type parameter.

v-next/hardhat-utils/test/synchronization.ts
```diff
@@ -100,4 +104,59 @@ describe("multi-process-mutex", () => {
...
+    // since it will detect that the process holding the lock is no longer running.
+
+    const child = spawn(process.execPath, [
+      "--import",
+      "tsx/esm",
+      path.resolve("test/helpers/synchronization.ts"),
+    ]);
+
+    // Wait until the process PID is available and written to the mutex file
+    await new Promise((resolve) => {
```
schaable: nit: this `setInterval` may run indefinitely if the process fails and the file is never written. We could add a timeout to clear the interval, or set a timeout on the test itself. No action needed now, but let’s watch CI for flakes.

v-next/hardhat-utils/test/helpers/synchronization.ts
```diff
@@ -0,0 +1,8 @@
+import { MultiProcessMutex } from "../../src/synchronization.js";
+
+const m = new MultiProcessMutex("test-mutex", 20_000);
```
schaable: This will break if the mutex name in `v-next/hardhat-utils/test/synchronization.ts` changes, and it’s easy to miss. Could we define a const in that file and import it here?

schaable (approved):
Looks good, pre-approving.

PR #7281, opened 2025-08-22
Move methods and types for hh-ledger

ChristopherDedominici:
Related to the `hardhat-ledger` PR: https://github.com/NomicFoundation/hardhat/pull/6983


v-next/hardhat-utils/src/synchronization.ts
```diff
@@ -124,3 +124,55 @@ export class MultiProcessMutex {
...
   }
 }
+
+/**
+ * A class that implements an asynchronous mutex (mutual exclusion) lock.
+ *
+ * The mutex ensures that only one asynchronous operation can be executed at a time,
+ * providing exclusive access to a shared resource.
+ */
+export class AsyncMutex {
```
michalbrabec, line 134: I'd be great to retain history for this code, but it is farily stable so it should be ok.

kanej (approved):

v-next/hardhat-zod-utils/src/errors/rpc.ts
```diff
@@ -0,0 +1,9 @@
+import { CustomError } from "@nomicfoundation/hardhat-utils/error";
+
+export class RpcValidationError extends CustomError {
```
alcuadrado: I think it would be better to:

* Add `hardhat-errors` as a dependency of this package, and use it (I checked, and this doesn't introduce any circular dependency.
* Make sure that the error descriptor is one of Hardhat core. This won't be 100% correct, but the "less wrong" option.

This way, we reuse all the error handling infra, and don't add any complexity to things like error reporting, filtering, etc.

ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7281/commits/9d37b8be8a4567b6ca329eef7ccc07f1ba54c848

kanej (approved):
Looks good. Lets get this merged!

Issue #7248, opened 2025-08-18
CLI network setting does not work for Node test runner


michalbrabec:
### Version of Hardhat

3.0.0

### What happened?

The `--network` switch does not work for the Node test runner sample project.

### Minimal reproduction steps

- create an empty directory and `cd` into it
- `pnpm dlx hardhat --init` with default settings
  - Hardhat V3
  - directory `.`
  - `A TypeScript Hardhat project using Node Test Runner and Viem`
  - Install dependencies `true`
- `pnpm hardhat test nodejs --network NON_EXISTENT_NET`

alcuadrado:
I believe this is a problem we had at one point in Hardhat 2.

The issue is that when we run a task that uses subprocesses, those subprocesses don't use the [global instance of the HRE](https://github.com/NomicFoundation/hardhat/blob/main/v-next/hardhat/src/internal/global-hre-instance.ts). Instead, they initialize their own. 

This happens with `node:test`, but could also happen with Mocha running in `parallel` mode.

Getting the subprocesses to use the same HRE instance is not possible, but we can get them to initialize an equivalent one, by providing the same set of global options. This can be done by providing the global options as env variables, as they can be set as `HARDHAT_<GLOBAL_OPTION_NAME>`, as shown in this screenshot:

<img width="933" height="495" alt="Image" src="https://github.com/user-attachments/assets/bbe28339-1e55-4abf-98bf-c6e4473d744d" />

An initial fix could be just setting all the env variables before running `mocha`/`node:test`. 

We can analyze later if there's something more systematic that we can do, like always setting them, or letting a task declare that they are needed, for example.

PR #7255, opened 2025-08-19
fix: passing network option to tests and mocha parallel execution

galargh:
- [ ] Because this PR includes a **bug fix**, relevant tests have been included.
- [ ] Because this PR includes a **new feature**, the change was previously discussed on an Issue or with someone from the team.
- [x] I didn't do anything of this.

---

Resolves https://github.com/NomicFoundation/hardhat/issues/7248 and fixes coverage for parallel Mocha test runs

This pull request introduces improvements to how global options and coverage are handled in Hardhat's testing infrastructure, especially for parallel test runs with Mocha. The main focus is on ensuring that global network options are correctly propagated to subprocesses and that coverage is accurately collected in parallel test environments.

**Key changes:**

### Global options propagation

* Ensured that the `HARDHAT_NETWORK` environment variable is set when running tests in subprocesses for both `hardhat-mocha` and `hardhat-node-test-runner`, so global network options are correctly passed to child processes.

### Coverage collection improvements

* Added a new `mochaHooks` export in `hardhat-mocha/coverage` to mark test workers as done after all tests finish, enabling accurate coverage collection in parallel Mocha test runs.
* Updated the test action logic to correctly require the new coverage hook when coverage is enabled and tests are run in parallel.

### Refactoring and minor code improvements

* Added a TODO comment suggesting a more systematic approach to passing global options as environment variables.

PR #7298, opened 2025-08-25
CLI network setting does not work for Node test runner

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/7248

### DIFF
--- a/v-next/hardhat-mocha/src/task-action.ts
+++ b/v-next/hardhat-mocha/src/task-action.ts
@@ -72,6 +72,15 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
 
   const mochaConfig: MochaOptions = { ...hre.config.test.mocha };
 
+  if (hre.config.test.mocha.parallel === true) {
+    // NOTE: We set the global variables as environment variables here because, as of now,
+    // the global options are not automatically passed to the child processes when the parallel
+    // execution is enabled.
+    if (hre.globalOptions.network !== undefined) {
+      process.env.HARDHAT_NETWORK = hre.globalOptions.network;
+    }
+  }
+
   if (grep !== undefined && grep !== "") {
     mochaConfig.grep = grep;
   }


v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts
```diff
@@ -15,11 +15,14 @@ export default async (): Promise<Partial<HardhatRuntimeEnvironmentHooks>> => ({
           "../network-manager.js"
         );
 
+        const networkName =
+          hre.globalOptions.network ??
+          process.env.HARDHAT_NETWORK ??
```
alcuadrado, line 20: I believe this could be `hre.globalOptions.network ?? DEFAULT_NETWORK_NAME`, as the globalOptions already use the env variables.

v-next/hardhat-mocha/src/task-action.ts
```diff
@@ -72,6 +72,15 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
 
   const mochaConfig: MochaOptions = { ...hre.config.test.mocha };
 
+  if (hre.config.test.mocha.parallel === true) {
+    // NOTE: We set the global variables as environment variables here because, as of now,
+    // the global options are not automatically passed to the child processes when the parallel
+    // execution is enabled.
+    if (hre.globalOptions.network !== undefined) {
```
alcuadrado, line 79: What if we set all the globalOptions? I think this error would reappear otherwise.

Maybe the confusion comes from the network manager explictly using `process.env`, but every globalOption can be read from the env. See [here](https://github.com/NomicFoundation/hardhat/blob/main/v-next/hardhat/src/internal/core/global-options.ts#L171).

alcuadrado: Maybe we can have a `getGlobalOptionEnvVariableName` helper, to keep these things in sync. Maybe even a `setGlobalOptionsAsEnvVariables`.

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -86,6 +86,10 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
     imports.push(coverage.href);
   }
 
+  if (hre.globalOptions.network !== undefined) {
+    process.env.HARDHAT_NETWORK = hre.globalOptions.network;
```
alcuadrado, line 90: same here, we could set all of them

kanej:
I think this PR can be closed - the narrow fix is already merged with https://github.com/NomicFoundation/hardhat/pull/7255

The larger env var fix mentioned in the comments is being addressed in https://github.com/NomicFoundation/hardhat/pull/7305

PR #7305, opened 2025-08-26
Load global options in env variables

ChristopherDedominici:
Fixes the issue mentioned in this comment: https://github.com/NomicFoundation/hardhat/pull/7298/files#r2298397675

### DIFF
--- a/v-next/hardhat/src/internal/core/hre.ts
+++ b/v-next/hardhat/src/internal/core/hre.ts
@@ -30,6 +30,7 @@ import { resolveConfigurationVariable } from "./configuration-variables.js";
 import {
   buildGlobalOptionDefinitions,
   resolveGlobalOptions,
+  setGlobalOptionsAsEnvVariables,
 } from "./global-options.js";
 import { HookManagerImplementation } from "./hook-manager.js";
 import { resolvePluginList } from "./plugins/resolve-plugin-list.js";
@@ -119,6 +120,8 @@ export class HardhatRuntimeEnvironmentImplementation
       userProvidedGlobalOptions,
       globalOptionDefinitions,
     );
+
+    setGlobalOptionsAsEnvVariables(globalOptions);
 
     // Set the HookContext in the hook manager so that non-config hooks can
     // use it
--- a/v-next/hardhat/src/internal/core/global-options.ts
+++ b/v-next/hardhat/src/internal/core/global-options.ts
@@ -168,7 +168,7 @@ export function resolveGlobalOptions(
     if (value !== undefined) {
       parsedValue = value;
     } else {
-      value = process.env[`HARDHAT_${camelToSnakeCase(name).toUpperCase()}`];
+      value = process.env[getEnvVariableNameFromGlobalOption(name)];
       if (value !== undefined) {
         // if the value is provided via an env var, it needs to be parsed
         parsedValue = parseArgumentValue(value, option.type, name);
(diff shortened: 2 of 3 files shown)


v-next/hardhat/src/internal/core/hre.ts
```diff
@@ -120,6 +121,8 @@ export class HardhatRuntimeEnvironmentImplementation
       globalOptionDefinitions,
     );
 
+    setGlobalOptionsAsEnvVariables(globalOptions);
```
kanej: I like explicitly separating out the state side-effect here, rather than meshing it with the global option resolution.

kanej: Okay, I have thought about this more after looking at the test.
What happens if I have a script that creates two hre instances with different `userProvidedGlobalOptions` options definitions passed in. Both instances will overwrite the ENV vars.
I think out intent is: at the top level of invocation of a task(?), convert global options to env vars, on HRE initialization it should pick up those env vars whether in a process/subprocess.

alcuadrado: I think I would do it:

- [In the mocha plugin](https://github.com/NomicFoundation/hardhat/blob/main/v-next/hardhat-mocha/src/task-action.ts#L86-L88)
- [In the node:test plugin](https://github.com/NomicFoundation/hardhat/blob/main/v-next/hardhat-node-test-runner/src/task-action.ts#L89-L91)

The reason being that this creates side effects and race conditions (if there are multiple instances of the HRE), as @kanej explained, so doing it there tries to minimize its impact.

v-next/hardhat-utils/src/env.ts
```diff
@@ -0,0 +1,33 @@
+import { camelToSnakeCase } from "./string.js";
+
+/**
+ * Sets the resolved global options as environment variables.
+ *
+ * @param globalOptions An object containing the resolved global options,
+ * with each option adhering to its definition in the globalOptionDefinitions.
+ */
+export function setGlobalOptionsAsEnvVariables<
```
ChristopherDedominici, line 9: Do we want to keep this method specific to global options, hence the name `setGlobalOptionsAsEnvVariables`, or do we want to keep it general and rename it something like `setEnvVariablesFromObject`?

v-next/hardhat-utils/src/env.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Converts a global option name to its corresponding environment variable name.
+ * The conversion involves transforming the option name from camelCase to
+ * SNAKE_CASE and prefixing it with "HARDHAT_".
+ *
+ * @param globalOptionName The name of the global option in camelCase.
+ *
+ * @returns The corresponding environment variable name in the format
+ * "HARDHAT_<OPTION_NAME_IN_SNAKE_CASE>".
+ */
+export function getEnvVariableNameFromGlobalOption(globalOptionName: string) {
```
ChristopherDedominici, line 31: Same question as the one in the previous comment

v-next/hardhat-utils/src/env.ts
kanej: Lets keep it specific for now. We can generalise later, and it will make it easier to understand what we are actually using it for right now.

.changeset/kind-avocados-compete.md
```diff
@@ -0,0 +1,8 @@
+---
+"@nomicfoundation/hardhat-mocha": patch
+"@nomicfoundation/hardhat-node-test-runner": patch
+"@nomicfoundation/hardhat-utils": patch
+"hardhat": patch
+---
+
+Load resolved global options into environment variables during tests
```
kanej:
```suggestion
Load resolved global options into environment variables during tests ([#7305](https://github.com/NomicFoundation/hardhat/pull/7305))
```

kanej (approved):
This looks great.
I did a manual test in the example project and `network` and `coverage` are passed to sub-processes cleanly in both `mocha` and `nodejs`.

PR #7307, opened 2025-08-26
chore: set maximum parallelism for package CI jobs

kanej:
We are seeing npm rate limiting on installing for each of our package jobs (and linting). This may be an issue with out github action caching, but for the moment I am limiting the number of jobs starting at once.

Next steps are an investigation into caching, bearing in mind we moved to `actions/checkout@v5` with its support for newer runners in GitHub Actions (changing the IP range as seen from NPM).

### DIFF
--- a/.github/workflows/v-next-ci.yml
+++ b/.github/workflows/v-next-ci.yml
@@ -192,6 +192,7 @@ jobs:
 
     strategy:
       fail-fast: false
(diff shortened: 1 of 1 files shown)


kanej:
It looks like npm is having issues - I am backing out of this change and waiting for npm to come back up fully.

Issue #7308, opened 2025-08-26
Hardhat 3 import resolution with remappings.txt and aliased npm packages


andrejrakic:
### Version of Hardhat

3.0.1

### What happened?

After upgrading to Hardhat 3, we are experiencing issues with how imports are resolved when using `remappings.txt` in combination with aliased npm packages.  

Our setup requires multiple versions of OpenZeppelin contracts to coexist in the same project. To achieve this, we install different versions under aliases in `package.json`, for example:

```json
"@openzeppelin/contracts@4.7.3/": "npm:@openzeppelin/contracts@4.7.3",
"@openzeppelin/contracts@4.8.3/": "npm:@openzeppelin/contracts@4.8.3",
"@openzeppelin/contracts@4.9.6/": "npm:@openzeppelin/contracts@4.9.6",
"@openzeppelin/contracts@5.0.2/": "npm:@openzeppelin/contracts@5.0.2",
"@openzeppelin/contracts@5.1.0/": "npm:@openzeppelin/contracts@5.1.0"
```

The `node_modules` folder contains the expected directories (e.g. `contracts@5.1.0`, `contracts@5.0.2`, etc.).

Our `remappings.txt` initially contained:

```txt
@openzeppelin/=node_modules/@openzeppelin/
```

We also tried more explicit remappings:

```txt
@openzeppelin/contracts=node_modules/@openzeppelin/contracts@4.7.3
@openzeppelin/contracts=node_modules/@openzeppelin/contracts@4.8.3
@openzeppelin/contracts=node_modules/@openzeppelin/contracts@4.9.6
@openzeppelin/contracts=node_modules/@openzeppelin/contracts@5.0.2
@openzeppelin/contracts=node_modules/@openzeppelin/contracts@5.1.0
```

### Error

However, when compiling, Hardhat fails to resolve the import with the following error message:

```shell
Error HHE902: There was an error while resolving the import "@openzeppelin/contracts@5.0.2/utils/Strings.sol" from "./src/v0.8/workflow/v1/WorkflowRegistry.sol":

You are trying to import an npm file but its syntax is invalid.

For more info go to https://hardhat.org/HHE902 or run Hardhat with --show-stack-traces
```

### Expected Behavior

Imports such as:

```solidity
import "@openzeppelin/contracts@5.0.2/utils/Strings.sol";
```

should resolve correctly using `remappings.txt`, even when multiple aliased versions of the same package are present.

[...]

PR #7309, opened 2025-08-27
Update npm module regex to include versions

zoeyTM:
resolves https://github.com/NomicFoundation/hardhat/issues/7308

### DIFF
--- a/.changeset/tidy-beds-dance.md
+++ b/.changeset/tidy-beds-dance.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Update npm module regex to include versions in solidity imports ([#7308](https://github.com/NomicFoundation/hardhat/issues/7308))


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/npm-module-parsing.ts
```diff
@@ -1,9 +1,9 @@
 import { assertHardhatInvariant } from "@nomicfoundation/hardhat-errors";
 
 const NPM_MODULE_PATTERN =
-  /^(?<package>(?:@[a-z0-9]+(?:[._-][a-z0-9]+)*\/)?[a-z0-9]+(?:[._-][a-z0-9]+)*)\/(?<path>.+)$/;
+  /^(?<package>(?:@[a-z0-9]+(?:[._-][a-z0-9]+)*\/)?[a-z0-9]+(?:[._-][a-z0-9]+)*(?:@(?:\d+\.)?(?:\d+\.)?(?:\*|\d+))?)\/(?<path>.+)$/;
```
michalbrabec, line 4: Fixes the user issue, but it supports only packages with version `major.minor.patch` (up to 3 version components).
Would it make sense to add support for more components like `major.minor.patch.build` just in case any package would use the notation?
TS/JS uses generally semantic versioning, but C# uses 4 component version

fvictorio: I don't think npm supports that, but it does support prereleases (which this regex doesn't handle).

That being said, using package names like this is already quite niche and surprising. Using it with pre-releases seems very unlikely to me, so I'd rather only add support for that if someone asks for it.

fvictorio:
Just for the record/future reference: I tried to split these regular expressions into fragments to make them more readable and to be able to better document them, but I didn't think the result was better enough than what we currently have: you have to either use strings and escape the backslashes, or use regular expressions and get their `.source`, which IMO gets unnecessarily complex.

fvictorio (approved):

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/npm-module-parsing.ts
michalbrabec: Agreed, we can always add it later if we get a request.

PR #7318, opened 2025-08-27
chore(ignition-core): return artifact id in docs and simplify verify yield

Galoretka:
- Align docs/comments with tested behavior by clarifying that the fallback string from getVerificationInformation is the artifact id (e.g., Module#Name), not the contract name. This matches tests and avoids ambiguity when multiple contracts share a name.
- Simplify generator loop in verify.ts by removing the redundant conditional yield/continue; now a single yield handles both VerifyInfo and string cases without behavior change.

Tests depend on artifact id being returned. Single-yield refactor is no-op behaviorally and aligns with project style of minimal branching when union types suffice.


kanej:

PR #7323, opened 2025-08-28
Feature: Add support for optional dependencies

Amxx:

---

Some plugins may want to provide additional (optional) features if another plugin is available. For example, a plugin that affect the connection may also perform additional actions on the `connection.ethers` or `connection.viem` objects if the corresponding plugins are installed. If either plugin is not installed, then it should not cause an error.

This PR changes the type of `HardhatPlugin['dependencies']` from
```typescript
Array<Promise<{ default: HardhatPlugin }>>
```
to
```typescript
Array<Promise<{ default: HardhatPlugin } | undefined>>
```

This allows the plugin to make a dependency optional by doing
```typescript
import("optional-dependency").catch(() => undefined);
```

This change is backward compatible: all existing `dependencies` arrays remain valid, and are not treated as optional.

### DIFF
--- a/v-next/hardhat/test/internal/core/plugins/resolve-plugin-list.ts
+++ b/v-next/hardhat/test/internal/core/plugins/resolve-plugin-list.ts
@@ -139,6 +139,44 @@ describe("Plugins - resolve plugin list", () => {
     );
   });
 
+  it("should support optional (resolved to undefined) dependencies", async () => {
+    function makeOptional(
+      dep: Promise<{ default: HardhatPlugin }>,
+    ): Promise<{ default: HardhatPlugin } | undefined> {
+      return dep.catch(() => undefined);
+    }
+
+    // C has no dependencies
+    // B depends on C (mandatory) and "not-installed-package" (optional, not found)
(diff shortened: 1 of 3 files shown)


v-next/hardhat/test/internal/core/plugins/resolve-plugin-list.ts
```diff
@@ -139,6 +139,44 @@ describe("Plugins - resolve plugin list", () => {
...
+      dep: Promise<{ default: HardhatPlugin }>,
+    ): Promise<{ default: HardhatPlugin } | undefined> {
+      return dep.catch(() => undefined);
+    }
+
+    // C has no dependencies
+    // B depends on C (mandatory) and "not-installed-package" (optional, not found)
+    // A depends on B (optional found) and "not-installed-package" (optional, not found)
+    //
+    // Resolution of A includes B (optional found) and C (transitive from C found)
```
Amxx:
```suggestion
    // Resolution of A includes B (optional found) and C (transitive from B)
```

michalbrabec:
We are building a comprehensive solution as part of #7424 
The proposal document is attached in the issue

Issue #7326, opened 2025-08-28
Error `cause` chains are truncated too aggressively.


alcuadrado:
## Context

In our custom `node:test` reporter, we truncate the amount of `cause`s that we show. That is, if an error has a `.cause` property, we show it. That's normally also an error, which may have a `.cause`, and so on. We truncate it after three iterations.

## Proposal

I think we should either: (1) not truncate it at all on CIs, (2) have a much much higher limit on CIs.

The only reason I propose (2) is that maybe, just maybe, we can find a loop. It would be hard to create one, but you never know.

### Subproposal

We could also bump the default a bit. From `3` to maybe `5`?

## Why am I proposing this?

We get CI failures like this one, that don't get to show the original error and hence are really hard to debug, especially if they are flaky: https://github.com/NomicFoundation/hardhat/actions/runs/17294799176/job/49090291482?pr=7324#step:6:4231

## Scope

You could argue that this fix should only be done for our own CI, but I think it should be applied to every project using our `node:test` reporter, as a similar situation can happen to our users. i.e. Hardhat users.

## Technical context

This is handled here: https://github.com/NomicFoundation/hardhat/blob/main/v-next/hardhat-node-test-reporter/src/error-formatting.ts#L17

Note that `node-test-reporter` is our "root" package, so it can't use `hardhat-utils`'s `isCI()`. But maybe worth duplicating it with a big explanation.

You could also consider passing it as a flag from `hardhat-node-test-runner`, but that wouldn't work with the tests or our own packages, where we load the reporter with a `node` flag (i.e. `--test-reporter=@nomicfoundation/hardhat-node-test-reporter`)

alcuadrado:
Changing the priority as this was reported by users.

PR #7327, opened 2025-08-28
test-reporter: Improve stack trace readability

iosh:
- [x] Because this PR includes a **bug fix**, relevant tests have been included.  #7326
- [  ] I didn't do anything of this.

---

- Increasing the default limit to 5.
- Setting the limit to 100 in CI environments for better debugging.
I copied the `isCI` function to avoid a "circular dependency" problem. I added comments to both files so we remember to update them together.

### DIFF
--- a/.changeset/polite-mails-grab.md
+++ b/.changeset/polite-mails-grab.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/hardhat-node-test-reporter": patch
 ---
 
-Fix test error logs being cut off. The default is now 5 lines (up from 3). In CI environments, it shows 100 lines for more detail.
+Fix test error `cause` chains being cut off. The default is now 10 `cause`s (up from 3). In CI environments, it's 100.
--- a/v-next/hardhat-node-test-reporter/src/error-formatting.ts
+++ b/v-next/hardhat-node-test-reporter/src/error-formatting.ts
@@ -15,7 +15,7 @@ import {
 const AGGREGATE_ERROR_INNER_ERROR_INDENT = 2;
 const ERROR_CAUSE_INDENT = 2;
 const ERROR_STACK_INDENT = 4;
-const MAX_ERROR_CHAIN_LENGTH = isCi() ? 100 : 5;
+const MAX_ERROR_CHAIN_LENGTH = isCi() ? 100 : 10;
 
 /**
  * Represents the result of successful parsing of a stack trace line.


v-next/hardhat-node-test-reporter/src/error-formatting.ts
```diff
@@ -14,7 +15,7 @@ import {
 const AGGREGATE_ERROR_INNER_ERROR_INDENT = 2;
 const ERROR_CAUSE_INDENT = 2;
 const ERROR_STACK_INDENT = 4;
-const MAX_ERROR_CHAIN_LENGTH = 3;
+const MAX_ERROR_CHAIN_LENGTH = isCi() ? 100 : 5;
```
alcuadrado: What if we do a bit more by default? 10?

alcuadrado: I know I said 5, but now I'm having second thoughts

v-next/hardhat-node-test-reporter/src/ci.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Copied from `v-next/hardhat-utils/src/ci.ts` to prevent a circular dependency.
+ * IMPORTANT: If you change this function, you must also change the original.
+ */
```
alcuadrado, line 4: Agreed with this approach

v-next/hardhat-utils/src/ci.ts
```diff
@@ -1,3 +1,8 @@
+/**
+ * IMPORTANT: A copy of this function exists in `v-next/hardhat-node-test-reporter/src/ci.ts`.
+ * If you change this function, you must also change the copy.
+ */
+
```
alcuadrado, line 5: 👌

.changeset/polite-mails-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-node-test-reporter": patch
+---
+
+Fix test error logs being cut off. The default is now 5 lines (up from 3). In CI environments, it shows 100 lines for more detail.
```
alcuadrado:
```suggestion
Fix test error `cause` chains being cut off. The default is now 10 `cause`s (up from 3). In CI environments, it's 100.
```

alcuadrado: It's not lines, but `cause`s.

That means `error.cause.cause.cause...`.

Note that I already switched the `5` for `10`.

alcuadrado (approved):

PR #7328, opened 2025-08-28
Custom compiler support

### LINKED ISSUE #7130: Using custom compiler in Hardhat v3
### Discussed in https://github.com/NomicFoundation/hardhat/discussions/7112

<div type='discussions-op-text'>

<sup>Originally posted by **MrKampla** August  3, 2025</sup>
Hey, I couldn't find docs about how to setup custom solidity compiler in Hardhat V3, there's one for Hardhat V2: https://hardhat.org/hardhat-runner/docs/other-guides/using-custom-solc.

Does Hardhat v3 support custom solidity compilers? I wanted to use the new solx compiler from the ZKSync team https://solx.zksync.io/#quickstart</div>

Basically what is the equivalent in Hardhat 3 for overriding the TASK_COMPILE_SOLIDITY_GET_SOLC_BUILD to support using an alternate solc compiler: https://hardhat.org/hardhat-runner/docs/other-guides/using-custom-solc

antico5:
This PR allows the user to specify a custom compiler for each solidity version.

Only the path needs to be specified, as the other required files (long version and wasm/native) are inferred.

Closes #7130


v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -1134,6 +1134,18 @@ Please check Hardhat's output for more details.`,
...
       },
+      COMPILER_PATH_DOES_NOT_EXIST: {
+        number: 913,
+        messageTemplate: `The compiler path "{compilerPath}" specified in your config for version {version} does not exist.`,
+        websiteTitle: "Compiler path does not exist",
+        websiteDescription: `The compiler path you are trying to use does not exist. Ensure you specified the correct path in the config file and that the file exists.`,
+      },
+      PARSING_VERSION_STRING_FAILED: {
+        number: 914,
+        messageTemplate: `Could not parse the full solc version from "{versionString}".`,
```
fvictorio: I think this error should include the compiler path too.

v-next/hardhat-utils/src/fs.ts
```diff
@@ -834,6 +834,43 @@ export async function findUp(
   }
 }
 
+/**
+ * This function uses some heuristics to check if a file is binary by reading the first bytesToCheck bytes from the file.
+ *
+ * @param filePath
+ * @param bytesToCheck
+ * @returns
```
fvictorio: I'd rather remove these jsdoc fields if they don't have anything.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/index.ts
```diff
@@ -79,8 +86,71 @@ export async function downloadConfiguredCompilers(
...
+  const match = stdout.match(/(?<longVersion>\d+\.\d+\.\d+\+commit\.\w+)/);
+
+  if (match === null || match.groups === undefined) {
+    throw new HardhatError(
+      HardhatError.ERRORS.CORE.SOLIDITY.PARSING_VERSION_STRING_FAILED,
+      { versionString: stdout },
+    );
+  }
+
+  const { longVersion } = match.groups;
```
fvictorio, line 139: I'm not convinced about this. Some questions:

- How sure are we that the output of solc's `--version` won't change?
- Does this work with solx?
- Do we actually need this? What happens if we just use `compilerVersion` as the long version in the `SolcJsCompiler`/`NativeCompiler` constructors? Would that be more or less risky than relying on the `--version` output?

.changeset/lazy-buttons-burn.md
```diff
@@ -0,0 +1,7 @@
+---
+"@nomicfoundation/hardhat-errors": patch
+"@nomicfoundation/hardhat-utils": patch
+"hardhat": patch
+---
+
+Support custom compilers ([#7130](https://github.com/NomicFoundation/hardhat/issues/7130))
```
michalbrabec: Nit: please change to a full sentence since it will end up in HH changeset
Like `Support for custom ...`.

v-next/hardhat-errors/src/descriptors.ts
michalbrabec: The description specifically mentions `solc`, do we support other EVM compilers with this change?
I would rephrase it from solc to just compiler in case we do.

v-next/hardhat-utils/src/fs.ts
```diff
@@ -834,6 +834,43 @@ export async function findUp(
...
+ */
+export async function isBinaryFile(
+  filePath: string,
+  bytesToCheck = 8000,
+): Promise<boolean> {
+  const fd = await fsPromises.open(filePath, "r");
+
+  const buffer = Buffer.alloc(bytesToCheck);
```
michalbrabec, line 846: It might be safer to load and process the file as a stream or check the size to avoid loading potentially huge file in memory.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/index.ts
```diff
@@ -1,24 +1,31 @@
...
 } from "./downloader.js";
+import wrapper from "./solcjs-wrapper.js";
 
 async function getGlobalCompilersCacheDir(): Promise<string> {
   const globalCompilersCacheDir = await getCacheDir();
 
   return path.join(globalCompilersCacheDir, "compilers-v3");
 }
 
+const log = debug("hardhat:core:solidity:build-system:compiler");
```
michalbrabec, line 27: Does this log init only when the debug logging is enabled or is it used always?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/index.ts
michalbrabec: Agreed, it is designed for a very specific version scheme that might not apply to other compilers.

antico5: - The --version output format has been always the same since the first version I could check (0.4.11). It could change but seems unlikely
- It works with solx and solcjs
- We'd need to ask pato on the implications on not having the proper long version

v-next/hardhat-errors/src/descriptors.ts
antico5: So far I tested with solx which uses the same frontend as solc, not sure if there are other compatible compilers available to test. In any case I'll change the wording, I think it makes sense

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/index.ts
fvictorio: I'm ok with keeping this approach if necessary (I agree with Armando that it seems unlikely to change, and if it does it will fail in a easy to detect way). But right now I don't think I can make a call on the trade-off because I don't know how important it is to have the proper long version.

v-next/hardhat-utils/src/fs.ts
antico5: I agree, but since this function just reads a small number of bytes to apply the binary/non binary heuristic, I think we can use the simpler approach

fvictorio: This only reads the first 8000 bytes of the file, not the whole thing.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/index.ts
antico5: This kind of `log` pattern works by only printing the output when passing a `DEBUG` env var. So in this case it would be `DEBUG=hardhat:core:solidity:build-system:compiler npx hardhat compile`

fvictorio: It's always created, and then calls to `log` have an effect or not depending on whether debug logging is enabled. This is a common pattern when using the `debug` module.

v-next/hardhat-utils/src/fs.ts
fvictorio: (I sent my comment at the same time as Armando, sorry for the duplication. Same thing below.)

Issue #7303, opened 2025-08-16
Frequent "all the transactions of its network interaction 1 were dropped" error.


yoshiso:
### What happened?

I have encounter this error quite oftern in hyperEVM testnet environment:
`all the transactions of its network interaction 1 were dropped. Please try rerunning Hardhat Ignition.`

And realized this issue was mitigated by this PR before:
https://github.com/NomicFoundation/hardhat-ignition/issues/665

But this fix immediately removed wit this PR.
https://github.com/NomicFoundation/hardhat-ignition/pull/667
https://github.com/NomicFoundation/hardhat-ignition/commit/6476b089de095f6d41c61f33ab0d910a705231ec#diff-45138ac0caea70558c5383e59e28d59e5f9cefcf9dfe2e4a75b394b74b75acf8R211-R216

Is this intentionally removed? I’d like to have this feature back to mitigate issue.

### Minimal reproduction steps

Deploy arbitrary contract to hyperEVM testnet.

E-r-i-o-n-d:
Hi,

I have the same issue with my ignition script for HyperEVM and big blocks enabled for the deployer account.
Is there a way to handle it or tell the ignition plugin to wait longer?

kanej:
Can I confirm you are running on Hardhat 2 + Ignition?

E-r-i-o-n-d:
I use:
"hardhat": "2.26.3",
"@nomicfoundation/hardhat-ignition": "0.15.13"

yoshiso:
Here is mine:
"hardhat": "2.26.1", 
"@nomicfoundation/hardhat-ignition": "0.15.13"

zoeyTM:
Hi! 

I've been attempting to reproduce this for a bit now, and I've been running into difficulties finding working faucets for HyperEVM to test with. A more specific solution for HyperEVM is still on our radar, but we will need to figure out a way to consistently test on this network first.

In the meantime, we're going to move forward with implementing a middle-ground solution that will give you the ability to set custom values for the retry loop we use in the PR you linked. This should hopefully allow you to find values that work for your specific HyperEVM environments.

yoshiso:
One thing I forgot to mention about big block feature. This issue happens when enabled big block which executed every 60seconds.

https://hyperliquid.gitbook.io/hyperliquid-docs/for-developers/hyperevm/dual-block-architecture

And here is community built frontend enable bug block
https://hyperevm-block-toggle.vercel.app/

[...]

PR #7331, opened 2025-08-29
Expose ignition retry loop variables in user config

zoeyTM:
fix for https://github.com/NomicFoundation/hardhat/issues/7303

### DIFF
--- a/v-next/hardhat-ignition-core/src/internal/defaultConfig.ts
+++ b/v-next/hardhat-ignition-core/src/internal/defaultConfig.ts
@@ -9,6 +9,8 @@ export const defaultConfig: DeployConfig = {
   maxFeeBumps: 4,
   requiredConfirmations: 5,
   disableFeeBumping: false,
+  maxRetries: 10,
+  retryInterval: 1_000,
 };
 
 /**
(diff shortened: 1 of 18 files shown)


v-next/hardhat-ignition-core/src/internal/defaultConfig.ts
```diff
@@ -9,6 +9,8 @@ export const defaultConfig: DeployConfig = {
   maxFeeBumps: 4,
   requiredConfirmations: 5,
   disableFeeBumping: false,
+  maxRetries: 10,
```
schaable, line 12: Can we centralize these values? They are hardcoded here and in `v-next/hardhat-ignition-core/src/internal/execution/future-processor/handlers/monitor-onchain-interaction.ts`. In the future we might update one and miss the other.

schaable (approved):
Looks good, but let’s wait for @kanej to give a second review since I’m not familiar with the ignition codebase.

v-next/hardhat-ignition-core/src/internal/defaultConfig.ts
zoeyTM: I think that's a good idea

zoeyTM:
closed in favor of https://github.com/NomicFoundation/hardhat/pull/7525 and https://github.com/NomicFoundation/hardhat/pull/7447

PR #7353, opened 2025-09-01
fix: correct handling of optimize flag for Vyper compiler versions

lyubovgrechiha98:

---

### What I did
- Refactored `getOptimize` function to simplify conditions.
- Fixed incorrect handling of the `optimize` setting for Vyper compiler:
  - For versions <= 0.3.0, any `optimize` value now correctly throws.
  - For versions 0.3.1–0.3.9:
    - `optimize: true` → no flag (enabled by default).
    - `optimize: false` → `--no-optimize`.
  - For versions >= 0.3.10:
    - `optimize: true` → throws.
    - `optimize: false` → `--optimize none`.

### Why
The previous implementation had duplicated conditions and was harder to follow.  
This change makes the logic more explicit, less error-prone, and covers all edge cases consistently.

### How I did it
- Removed nested `if/else` blocks.
- Added early guard for `<= 0.3.0`.
- Unified return of flags into a single ternary expression.

### DIFF
--- a/archive/hardhat-vyper/src/compiler.ts
+++ b/archive/hardhat-vyper/src/compiler.ts
@@ -71,29 +71,18 @@ function getOptimize(
   }
 
   if (typeof optimize === "boolean") {
-    if (optimize) {
-      if (
-        semver.gte(compilerVersion, "0.3.10") ||
-        semver.lte(compilerVersion, "0.3.0")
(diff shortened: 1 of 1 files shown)


kanej:

Vyper is under the archive folder and has not currently been ported to Hardhat 3.

Issue #7288, opened 2025-08-23
viem assertions: revert does not handle Panic errors


mdesousa:
### Version of Hardhat

3.0.1

### What happened?

an assertion like `await viem.assertions.revert(token.write.transfer([address, amount]));` should succeed if a Panic (e.g. math overflow) error occurs. instead it's interpreting as a custom error and failing:

```
AssertionError: Expected default error revert, but got a custom error selector "0x4e487b71" with data "0x4e487b710000000000000000000000000000000000000000000000000000000000000011"
```

### Minimal reproduction steps

invoice a contract function with a math overflow

michalbrabec:
@mdesousa can you please check if I reproduce the same issue as you, following this guide
Repro guide:

- generate a new HH3 project `pnpm dlx hardhat --init`
- update `Counter.sol` to use `uint8` instead of `uint` for `x` (make overflow simple)
- test the functionality by:
  ```typescript
  it.only("Should throw a panic revert", async function () {
    const counter = await viem.deployContract("Counter");
  
    assert(await counter.write.incBy([200]));
    await viem.assertions.revert(counter.write.incBy([200])); // should revert with uint8 overflow
  });
  ```
  - throws `AssertionError: Expected default error revert, but got a custom error selector "0x4e487b71" with data "0x4e487b710000000000000000000000000000000000000000000000000000000000000011"`

mdesousa:
hi @michalbrabec , yes that looks right to me 👍 thanks

PR #7354, opened 2025-09-02
Viem assertions: revert does not handle Panic errors

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/7288


fvictorio (commented):
Shared some comments via DM. I'm just commenting here so that it doesn't show up in my list of PRs awaiting for review from me.

fvictorio (approved):

PR #7358, opened 2025-09-02
Update the CI pipelines for new GH issues

michalbrabec:
- include the entire team in triage
- remove the `status:ready` label
- include @michalbrabec in automatic PR assignment for external contributors
- eliminate `v-next` action since it is no longer needed
- eliminate `Close stale issues` action since the robotic approach was prompting negative sentiment from users

### DIFF
--- a/.github/workflows/autoassign-issues.yml
+++ b/.github/workflows/autoassign-issues.yml
@@ -15,11 +15,11 @@ jobs:
           script: |
             // each user has a chance of (p - (previousP ?? 0)) to be assigned
             const potentialAssignees = [
-              ["kanej", 1.0],
-              ["michalbrabec", 1.0],
-              ["schaable", 1.0],
-              ["ChristopherDedominici", 1.0],
-              ["antico5", 1.0],
+              ["kanej", 1/6],
+              ["michalbrabec", 2/6],
+              ["schaable", 3/6],
+              ["ChristopherDedominici", 4/6],
+              ["antico5", 5/6],
               ["zoeyTM", 1.0],
             ];
 
--- a/.github/workflows/autoassign-prs.yml
+++ b/.github/workflows/autoassign-prs.yml
@@ -13,7 +13,25 @@ jobs:
       - uses: actions/github-script@v7
         with:
           script: |
-            const externalPrTriagers = ["kanej", "michalbrabec"];
+            // each user has a chance of (p - (previousP ?? 0)) to be assigned
+            const potentialAssignees = [
+              ["kanej", 1/2],
+              ["michalbrabec", 1.0]
+            ];
+
+            let assignee;
+            const r = Math.random();
+            console.log("r:", r);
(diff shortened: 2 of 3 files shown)


.github/workflows/autoassign-issues.yml
```diff
@@ -15,9 +15,12 @@ jobs:
...
             const potentialAssignees = [
-              ["kanej", 1/3],
-              ["schaable", 2/3],
+              ["kanej", 1.0],
+              ["michalbrabec", 1.0],
+              ["schaable", 1.0],
               ["ChristopherDedominici", 1.0],
+              ["antico5", 1.0],
+              ["zoeyTM", 1.0],
             ];
```
kanej:
```suggestion
           const potentialAssignees = [
              ["kanej", 1/6],
              ["michalbrabec", 2/6],
              ["schaable", 3/6],
              ["ChristopherDedominici", 4/6],
              ["antico5", 5/6],
              ["zoeyTM", 1.0],
           ];
```
I think the randomization assumes the structure above - rather than the number indicating the weighting.

.github/workflows/autoassign-prs.yml
```diff
@@ -24,13 +24,12 @@ jobs:
...
 
-            if (isCollaborator) {
-              return
+            // we only assign PRs from external users
+            if (!isCollaborator) {
+              await github.rest.issues.addAssignees({
+                owner: context.issue.owner,
+                repo: context.issue.repo,
+                issue_number: context.issue.number,
+                assignees: externalPrTriagers,
```
kanej: The plan here is that we are both assigned?

.github/workflows/v-next-ci.yml
```diff
@@ -1,315 +0,0 @@
-name: vNext CI
```
kanej: This file shouldn't be deleted, while it is named v-next-ci.yml it is the main ci flow. The name wasn't updated with the branch migration.

kanej (changes requested):
Looks good. I left a few comments and requested a file be brought back.

.github/workflows/autoassign-issues.yml
michalbrabec: Ah, I get it.
It was not clear from the short comment. 
Updated

.github/workflows/autoassign-prs.yml
michalbrabec: Yep, so it is clear that one of us should take a look, my plan was that I'll un-assign you from those I review and you can do the same.
We can also randomise it.

.github/workflows/v-next-ci.yml
michalbrabec: Missed that, I thought it assigns the `v-next` label.
I'll rename it and the action them to make it clear.

.github/workflows/autoassign-prs.yml
kanej: Sounds good.

PR #6886, opened 2025-06-17
Upgrade hardhat-gas-reporter to v2 on toolboxes

antico5:
The package hardhat-gas-reporter was upgraded from 1.0.10 to ^2.3.0.

Toolboxes' major versions have been bumped, and hardhat minor version as well. The new project template has been edited to reflect the new versions.

Since the hardhat docs don't mention any particular feature/config that's affected by the version change, no changes are required on them.

Existing users that bump their toolbox versions would need to keep in mind the reporter's config changes, detailed next. New projects generated with the latest hardhat version shouldn't have any issue.

The following config variables are no longer supported by the plugin. Note that no warnings are issued by the plugin, only typescript errors if using typescript for config file:
  - onlyCalledMethods: reworked into "showUncalledMethods" (default false)
  - showTimeSpent
  - url
  - maxMethodDiff: support for Codechecks has been removed
  - maxDeploymentDiff

The following config variables are new:
  - etherscan (api key)
  - excludeAutoGeneratedGetters
  - includeIntrinsicGas
  - L1
  - L2
  - offline
  - optimismHardfork
  - reportPureAndViewMethods
  - trackGasDeltas
  - currencyDisplayPrecision
  - darkMode
  - forceTerminalOutput
  - forceTerminalOutputFormat
  - reportFormat
  - showUncalledMethods
  - suppressTerminalOutput
  - includeBytecodeInJSON
  - outputJSONFile
  - outputJSON
  - baseFee
  - baseFeePerByte
  - blobBaseFee
  - blobBaseFeeApi
  - gasPriceApi
  - getBlockApi
  - opStackBaseFeeScalar
  - opStackBlobBaseFeeScalar
  - tokenPrice

PR #7367, opened 2025-09-03
Fix REPORT_GAS envvar in toolboxes

fvictorio:
When we [upgraded](https://github.com/NomicFoundation/hardhat/pull/6886) `hardhat-gas-reporter` to v2, the `REPORT_GAS` variable stopped working.

The toolboxes READMEs have this snippet:

```
REPORT_GAS=true npx hardhat test
```

But now gas is reported whether you set the variable or not. The reason is that v2 of `hardhat-gas-reporter` now uses `extendConfig` and sets its options to some default values. `gasReport.enabled` defaults to true, so our code never changes it.

I changed the code so that `enabled` is set using the `REPORT_GAS` envvar only if the **user** config of gas reporter doesn't have an `enabled` field. This also means that gas is **not** reported by default, unlike what happens if you install the plugin manually. This was a conscious decision, both to have the same behavior we used to have, and because I don't think it makes sense to show gas usage by default.

I manually tested this locally with verdaccio and confirmed it works fine.

### DIFF
--- a/packages/hardhat-toolbox-viem/src/index.ts
+++ b/packages/hardhat-toolbox-viem/src/index.ts
@@ -20,6 +20,7 @@ extendConfig((config, userConfig) => {
   const configAsAny = config as any;
 
   if (userConfig.gasReporter?.enabled === undefined) {
-    configAsAny.gasReporter.enabled = process.env.REPORT_GAS ? true : false;
+    configAsAny.gasReporter.enabled =
+      process.env.REPORT_GAS === "true" ? true : false;
   }
 });
--- a/packages/hardhat-toolbox-viem/src/index.ts
+++ b/packages/hardhat-toolbox-viem/src/index.ts
@@ -19,21 +19,7 @@ import { extendConfig } from "hardhat/config";
 extendConfig((config, userConfig) => {
   const configAsAny = config as any;
 
-  // hardhat-gas-reporter doesn't use extendConfig, so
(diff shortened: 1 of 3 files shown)


packages/hardhat-toolbox-viem/src/index.ts
```diff
@@ -19,21 +19,7 @@ import { extendConfig } from "hardhat/config";
...
-
-  if (gasReporterConfig?.enabled === undefined) {
-    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+  if (userConfig.gasReporter?.enabled === undefined) {
     configAsAny.gasReporter.enabled = process.env.REPORT_GAS ? true : false;
   }
-
-  if (gasReporterConfig?.currency === undefined) {
-    configAsAny.gasReporter.currency = "USD";
-  }
```
fvictorio, line 38: `currency` already defaults to USD in hardhat-gas-reporter v2.

packages/hardhat-toolbox-viem/src/index.ts
```diff
@@ -19,21 +19,7 @@ import { extendConfig } from "hardhat/config";
 extendConfig((config, userConfig) => {
   const configAsAny = config as any;
 
-  // hardhat-gas-reporter doesn't use extendConfig, so
```
fvictorio, line 22: This comment was outdated: hardhat-gas-reported does use `extendConfig` now

Issue #7052, opened 2025-07-23
Return types for `getBuildInfoPath` and `getBuildInfoOutputPath`


kanej:
Some issues with getBuildInfoPath and getBuildInfoOutputPath.  ArtifactManager indicates that they can both return undefined if build id doesn't exist.  ArtifactManagerImplementation narrows the type of one of them (but not the other) to return only string, but the functions themselves both return only string.I tried calling with a contract name and got a nonexistent path like this: /home/nickbarry/workspace/hardhat-plugins/hardhat-contract-sizer/artifacts/build-info/Test.json.

kanej:
This has been released as part of `hardhat@3.0.5`: https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%403.0.5

PR #7373, opened 2025-09-03
fix: correctly return undefined when build info files are missing

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7052

I think it would be cleaner to always return a `string` and let the caller handle validation, since the file could still be deleted between calling `getBuildInfoPath`/`getBuildInfoOutputPath` and reading it. In practice, this seems unlikely, so the current approach should be safe.


v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
```diff
@@ -117,22 +118,30 @@ export class ArtifactManagerImplementation implements ArtifactManager {
...
+    const buildInfoPath = path.join(
       this.#artifactsPath,
       BUILD_INFO_DIR_NAME,
       buildInfoId + ".json",
     );
+
+    return (await exists(buildInfoPath)) ? buildInfoPath : undefined;
   }
 
   public async getBuildInfoOutputPath(
```
michalbrabec, line 133: Can you please add info into the comment of [`getBuildInfoOutputPath `](https://github.com/NomicFoundation/hardhat/blob/0a60e34cf6b514e5b06aef7012a6b8451009847e/v-next/hardhat/src/types/artifacts.ts#L114) explaining that undefined is returned in case the file does not exist?
All the functions calling this one expect that `undefined` might be returned, but it is not clearly stated in the interface.

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
```diff
@@ -117,22 +118,30 @@ export class ArtifactManagerImplementation implements ArtifactManager {
     return new Set(paths.map((p) => path.basename(p, ".json")));
   }
 
-  public async getBuildInfoPath(buildInfoId: string): Promise<string> {
-    return path.join(
+  public async getBuildInfoPath(
```
michalbrabec, line 121: This is not an API change, it is a fix for the API making sure that the `ArtifactManagerImplementation` complies with the [`ArtifactManager`](https://github.com/NomicFoundation/hardhat/blob/0a60e34cf6b514e5b06aef7012a6b8451009847e/v-next/hardhat/src/types/artifacts.ts#L104) interface.

michalbrabec:
Please consider adding / expanding tests to cover the `undefined` results.

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/7373/commits/effa8839347aed46f6ca71f72d1845ad1a2f0353

schaable:
> Please consider adding / expanding tests to cover the `undefined` results.

@michalbrabec the `ArtifactManagerImplementation` previously had no tests. I added a test structure and implemented tests for `getBuildInfoPath` and `getBuildInfoOutputPath`. I also fixed some `hardhat-utils` tests that were failing under WSL2.

michalbrabec:
> > Please consider adding / expanding tests to cover the `undefined` results.
> 
> @michalbrabec the `ArtifactManagerImplementation` previously had no tests. I added a test structure and implemented tests for `getBuildInfoPath` and `getBuildInfoOutputPath`. I also fixed some `hardhat-utils` tests that were failing under WSL2.

I know there we no tests, thanks for adding them, another point for our test coverage!

PR #7378, opened 2025-09-04
Build system support for compiling tests separately from contracts

### LINKED ISSUE #6474: hardhat compile and test contracts
### Feedback

Currently it seems that hardhat compile do not compile test contract

but it compiles them when running test unless `--no-compile` is specified in which case it won't compile them

this means that if you want to control compilation and use `--no-compile` the only way to compile the test contract is to use 

`hardhat compile src/**/*.sol` (assuming contract source is src)

but this behaviot depends on the shell and can cause problem where this command is specified in npm/package.json scripts

it would be great to have an option to compile all contracts

antico5:
- A new option has been added to the `build` task, named `targetSources` (name can be changed), that means whether you want to compile contracts or tests.
- The general idea is that now the build system understands what type of solidity files (contracts/tests) it's compiling and it can take some decisions based on that.
- Logic for determining what solidity files are tests has been deduplicated
- Artifacts and build infos for tests now reside in `cache/test-artifacts`, since they are considered to be an implementation detail unlike contract artifacts
- Open for feedback on the workaround on solidity-tests task for using a different artifacts folder than the one in the single `ArtifactManager` instance. Likely a bigger refactoring related to artifacts management is needed but out of scope.
- Artifact cleanup is now scoped to each targetSource. I.e. when compiling contracts, tests artifacts are not cleaned up and vice versa.
- The `onCleanupArtifact` hook, used for generating typechain types, is not fired on tests compilation
- The `targetSources` option was added to the build task instead of the `BuildSystemImplementation`, because there was top-level logic on the task action that we wanted to keep, so a side effect is now the `--target-sources` is exposed in the cli. There's a discussion that arises from this about supporting task args that are not exposed in the cli.

Closes #6474


fvictorio:
Does this really close #6474? Users still don't have a way to compile all contracts (or to compile test contracts), which is the spirit of the issue. (I'm assuming we are going to hide the `targetSources` option when we have a mechanism to do it.)

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/integration/target-sources.ts
```diff
@@ -0,0 +1,258 @@
...
+        path.join(contractsArtifactsPath, "contracts", `Foo.sol`, `Foo.json`),
+      );
+
+      // Assert build info file exists
+      const buildInfoPath = path.join(
+        contractsArtifactsPath,
+        "build-info",
+        `${artifact.buildInfoId}.json`,
+      );
+      assert.equal(await exists(buildInfoPath), true);
```
fvictorio: This test doesn't seem complete to me. For example: maybe the test in the `test/` directory is not being compiled, but the `.t.sol` one in `contracts` is.

Easy fix: load the build info and check that it only has a single source.

antico5:
> Does this really close #6474? Users still don't have a way to compile all contracts (or to compile test contracts), which is the spirit of the issue. (I'm assuming we are going to hide the `targetSources` option when we have a mechanism to do it.)

I know it can be a bit misleading because the issue was the starting point, but the actual requirements were specified in an call with Pato and refined in a design doc. I'm not sure we will remove the `targetSources` option, and in case it stays, users will be able to compile tests. For compiling all, there's no single command but it can be achieved by running once for contracts and once for tests. I can remove the reference to the issue after we clarify what we want to do with that.

This sounds like a product decision for me, so once it's been settled I can make the required changes. Maybe Pato has some input.

v-next/hardhat/src/internal/builtin-plugins/solidity/tasks/build.ts
```diff
@@ -1,46 +1,77 @@
...
 import { throwIfSolidityBuildFailed } from "../build-results.js";
 import { isNpmRootPath } from "../build-system/root-paths-utils.js";
 
-interface CompileActionArguments {
+interface BuildActionArguments {
   force: boolean;
   files: string[];
   quiet: boolean;
   defaultBuildProfile: string | undefined;
+  targetSources: string;
```
alcuadrado: Let's replace with `--no-tests`, defaults to `false`

v-next/hardhat/src/types/solidity/build-system.ts
```diff
@@ -301,4 +305,11 @@ export interface SolidityBuildSystem {
...
     options?: CompileBuildInfoOptions,
   ): Promise<CompilerOutput>;
+
+  /**
+   * Gets the artifacts directory for a given target (contracts/tests)
+   */
+  getArtifactsDirectory(targetSources: TargetSources): Promise<string>;
 }
+
+export type TargetSources = "contracts" | "tests";
```
alcuadrado: let's name the type `BuildScope`.

v-next/hardhat/src/types/solidity/build-system.ts
```diff
@@ -196,7 +196,7 @@ export interface SolidityBuildSystem {
    *
    * @returns An array of root file paths.
    */
-  getRootFilePaths(): Promise<string[]>;
+  getRootFilePaths(targetSources?: TargetSources): Promise<string[]>;
```
alcuadrado: In the methods that receive the this new param, let's use an options object (for the actual options — for consistency). Note that this value (the BuildScope) is always an option, so it could be `{scope?: BuildScope}`.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -68,45 +64,28 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
...
-  throwIfSolidityBuildFailed(results);
+  // Run the compile task for test files
+  const { rootPaths }: { rootPaths: string[] } = await hre.tasks
+    .getTask("compile")
+    .run({
+      targetSources: "tests",
+      quiet: true,
+      force: false,
+      files: testFiles,
+    });
```
alcuadrado: As we discussed, `build`/`compile` will run with both build scopes by default, so we only need to call it once, and without any special behavior, other than quiet: true.

v-next/hardhat/src/internal/builtin-plugins/solidity/tasks/build.ts
```diff
@@ -1,46 +1,77 @@
...
 }
 
-const buildAction: NewTaskActionFunction<CompileActionArguments> = async (
-  { force, files, quiet, defaultBuildProfile },
+const buildAction: NewTaskActionFunction<BuildActionArguments> = async (
```
alcuadrado, line 19: This will build with both build scopes by default.

How should this work?

- First compile the `contracts` scope.
- If it fails, the task should fail.
- If it succeeds, we emit artifacts, complete that side of the build process (i.e. for the `contracts scope`)
- We then run the build for `tests` scope.
- If it fails, the task should fail.
- If it succeeds, we emit the test artifacts.

The output that we print, should explain that these are two processes.

alcuadrado: If you provide a list of files:

* You get a list of files
* You split up into two lists, depending on their scopes
* Then we run each compilation phase, with each list of files (post-categorization).
* If a list is empty, we skip the phase completely.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -603,18 +635,21 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
   public async emitArtifacts(
     runnableCompilationJob: CompilationJob,
     compilerOutput: CompilerOutput,
+    targetSources: TargetSources,
```
alcuadrado: Let's create an issue so that in the future we don't even generate .d.ts files for solidity test artifacts.

@antico5 feel free to add it to this PR if it's small enough.

v-next/hardhat/src/internal/builtin-plugins/solidity/tasks/build.ts
```diff
@@ -1,46 +1,89 @@
...
-        });
+  if (args.noContracts === false) {
+    contractRootPaths.push(...(await buildForScope("contracts", args, hre)));
+  }
+
+  if (args.noTests === false) {
+    testRootPaths.push(...(await buildForScope("tests", args, hre)));
+  }
+
+  return { contractRootPaths, testRootPaths };
```
fvictorio, line 34: FYI we consider anything returned by a task as public API, meaning we need to keep it backwards compatible once we release it.

I don't mind this structure (although I'd probably prefer something like `{ rootPaths: { contracts, tests } }`), but just want to make sure you know that anything we do here is "forever" (kind of; we are still in beta and no one is going to rely on this any time soon).

PR #7384, opened 2025-09-05
Improve error messages for viem assertions

ChristopherDedominici:


v-next/hardhat-viem-assertions/src/internal/assertions/revert/handle-revert-with-custom-error.ts
```diff
@@ -42,15 +39,33 @@ export async function handleRevertWithCustomError<
         abi: contract.abi,
       });
 
+      if (isKnownErrorSelector(rawError.data)) {
+        if (isPanicErrorSelector(rawError.data)) {
+          assert.fail(
+            `The function was expected to revert with custom error "${customErrorName}", but it reverted with a panic error: ${rawError.message}`,
```
fvictorio: This results in an error message like this, which I don't quite like:

> but it reverted with a panic error: VM Exception while processing transaction: reverted with panic code

We should have everything we need to print a better error message:

- `args[0]` has the panic code
- We have [this logic](https://github.com/NomicFoundation/hardhat/blob/9d4b8948d2a8e41917c3882338761ed5af561ab8/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/panic-errors.ts#L13) to "format" a panic code. (We might need to move that function to hardhat-utils, but I think that's fine).

ChristopherDedominici: Commits to address this comment:
- moved panic utils to hh-utils: https://github.com/NomicFoundation/hardhat/pull/7384/commits/ff13aefc882e346273a37beafc6f9d427762d1b5
- improved error messages: https://github.com/NomicFoundation/hardhat/pull/7384/commits/73532a10174b7e38302e56e0ed4506342ddb6ba9

Example for new message:
The function was expected to revert with custom error "CustomError", but it reverted with panic code 0x11 (Arithmetic operation overflowed outside of an unchecked block)

fvictorio (approved):

PR #7390, opened 2025-09-06
feat(hardhat-verify): add automatic proxy detection for verification

shwndi:
Hardhat Proxy Detection Solution
Problem
Hardhat 3 verification fails in restricted networks despite successful contract deployment. Contract verification bypasses proxy settings, causing ECONNREFUSED errors when accessing Etherscan API.
Root Cause

createVerificationProviderInstance function receives undefined dispatcher
No automatic proxy detection when dispatcher is not explicitly provided
Verification requests proceed without proxy configuration


schaable:
Adding this logic to the passed dispatcher doesn’t make sense, since that dispatcher is mainly intended for testing. I’ve opened another PR that handles proxies in verification, so I’ll close this one in favor of that. Thanks a lot for submitting it though!

PR #7393, opened 2025-09-07
Fix command injection vulnerability in file opening utility

sashaodessa:
**Security Fix**: Replaced unsafe `execSync` with shell command concatenation with secure `execFileSync`/`spawnSync` calls using argument arrays.

### What Changed

- **Before**: `execSync(\`${command} ${filePath}\`)` - vulnerable to command injection via `filePath` parameter
- **After**: Platform-specific secure execution:
  - **Windows**: `spawnSync("cmd", ["/c", "start", "", filePath])`
  - **macOS**: `execFileSync("open", [filePath])`  
  - **Linux**: `execFileSync("xdg-open", [filePath])`

### Why This Matters

The previous implementation concatenated the file path directly into a shell command string, allowing potential command injection if the `filePath` parameter contained shell metacharacters (`;`, `&&`, `|`, `$()`, etc.).

### DIFF
--- a/v-next/hardhat-ignition/src/internal/utils/open.ts
+++ b/v-next/hardhat-ignition/src/internal/utils/open.ts
@@ -1,21 +1,18 @@
-import { execSync } from "node:child_process";
+import { execFileSync, spawnSync } from "node:child_process";
 import os from "node:os";
 
 export function open(filePath: string): void {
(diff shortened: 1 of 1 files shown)


michalbrabec:
Hi,

PR #7395, opened 2025-09-08
refactor(network-helpers): make toBigInt synchronous and update call sites

Galoretka:
- Convert internal toBigInt(NumberLike) to a synchronous function and remove unnecessary awaits at all call sites and tests.
- This change eliminates a misleading async signature with no awaits, reduces overhead, and aligns toBigInt with other conversion helpers (toNumber, toRpcQuantity) and with @nomicfoundation/hardhat-utils/bigint which is synchronous.
- Impact: internal-only API change; updated all internal usages and tests accordingly. No external breaking change.


michalbrabec:
Hi,

PR #7405, opened 2025-09-09
Issue templates updated to match the GH structure

michalbrabec:
Updated the GH issue templates to match the new GH issue structure (feature, task, bug differences)
- Added automatic issue type setting based on the template
- Added a template for a Task
- Added an automatic lable `status:needs-priority` for features to minimise scope creep


.github/ISSUE_TEMPLATE/task-definition.yml
```diff
@@ -0,0 +1,11 @@
...
+description: Define a new task
+type: "Task"
+body:
+  - type: textarea
+    id: task-description
+    attributes:
+      label: "Describe the task"
+      description: "A description of an actionable tasks with a single deliverable and a limited scope"
+    validations:
+      required: true
```
kanej: I am not sure about adding a `task` definition.
It it what we want to use internally, but I think it adds friction to external users (not a huge amount)

michalbrabec: How about the Other issue coming with the type `Task` so we can use it to quickly file them without adding friction for external users?
This would be mostly invisible for them since they do not have to worry about issue types as that is purely our classification

kanej: Sounds good.

kanej (approved):

Issue #7349, opened 2025-09-01
Support fixture sharing between files


michalbrabec:
### Describe the feature

Now you are forced to have it in the same test file and can't share it between test files.
For example, this works:
```
const { ethers, networkHelpers } = await network.connect();

async function myFixture() {
  const foo = await ethers.deployContract("Foo")

  return { foo }
}

it("test", async function () {
  const { foo } = await networkHelpers.loadFixture(myFixture)
})
```
But that won't work if you have the fixture in a separate file.
The obvious solution is for fixtures to automatically receive the network connection. That way you can do this:
```
async function myFixture({ ethers }) {
  const foo = await ethers.deployContract("Foo")

  return { foo }
}

const { ethers, networkHelpers } = await network.connect();

it("test", async function () {
  const { foo } = await networkHelpers.loadFixture(myFixture)
})
```
And the fixture can be located anywhere.

PR #7406, opened 2025-09-09
Support fixture sharing between files

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/7349

Example of fixture function with the `connection` parameter:
```
export async function myFixture(connection: NetworkConnection) {
  const foo = await connection.ethers.deployContract("Foo");

  return { foo };
}

// OR, for non generic connections:

export async function myFixture(connection: NetworkConnection<"l1">) { // Specify the chain type
  const foo = await connection.ethers.deployContract("Foo");

  return { foo };
}
```

The website docs have been adjusted in this PR: https://github.com/NomicFoundation/hardhat-website/pull/75


v-next/hardhat-network-helpers/src/internal/hook-handlers/network.ts
```diff
@@ -11,12 +11,9 @@ export default async (): Promise<Partial<NetworkHooks>> => {
         nextContext: HookContext,
       ) => Promise<NetworkConnection<ChainTypeT>>,
     ) {
-      const connection: NetworkConnection<ChainTypeT> = await next(context);
```
michalbrabec, line 14: Why was the explicit type removed?

v-next/hardhat-network-helpers/src/types.ts
```diff
@@ -236,7 +244,10 @@ export interface NetworkHelpers {
   takeSnapshot(): Promise<SnapshotRestorer>;
 }
 
-export interface Time {
+export interface Time<
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars -- ChainTypeT is used in the class implementing the interface
+  ChainTypeT extends ChainType | string = DefaultChainType,
```
michalbrabec: Why must be the type parameter declared in the interface when it is not using it for anything?
It can be declared only in the implementation.
Usually, ifaces come with type params when some of their methods need the types as parameters or return values.

v-next/hardhat-network-helpers/src/internal/hook-handlers/network.ts
ChristopherDedominici: Because it is inferred automatically, it was redundant

v-next/hardhat-network-helpers/src/types.ts
ChristopherDedominici: It's needed because in the `NetworkHelpers` interface, `Time` must have the same `ChainTypeT` as the `NetworkHelpers`. So the `Time` interface without the `ChianTypeT`, will trigger a error:
image

v-next/hardhat-network-helpers/src/internal/hook-handlers/network.ts
michalbrabec: I understand, I'm just a fan of explicit typing wherever possible since it quickly indicates when things change unexpectedly.
Just a nit :D

v-next/hardhat-network-helpers/src/types.ts
michalbrabec: `Time` in this context is the interface of the same name, which would be non-generic on this context, while the implementation would be generic in `v-next/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts`

michalbrabec: Removed from type parameter for `interface Time` as discussed.

fvictorio (commented):
Code looks fine and I confirmed manually that it works. But as mentioned in Slack, this needs a companion PR in https://github.com/NomicFoundation/hardhat-website

Issue #7403, opened 2025-09-09
forking with a `configVariable`


olehmisar:
### Describe the feature

I want to use a `configVariable` RPC url for forking

### Repro guide

```typescript
const config: HardhatUserConfig = {
    test: {
        solidity: {
            forking: {
                rpcEndpoints: {
                    arbitrum: configVariable("ARBITRUM_RPC_URL")
                }
            }
        }
    }
};
```

### Solution

Change all URLs in the [`HardhatUserConfig`](https://github.com/NomicFoundation/hardhat/blob/c2efd230fe013c6165d7209d56458c5dfcabdc69/v-next/hardhat/src/types/config.ts#L68) and its extensions from `string` to [`SensitiveString`](https://github.com/NomicFoundation/hardhat/blob/c2efd230fe013c6165d7209d56458c5dfcabdc69/v-next/hardhat/src/types/config.ts#L62).

The issue for Solidity test forking is caused by the type of [`rpcEndpoints`](https://github.com/NomicFoundation/hardhat/blob/c2efd230fe013c6165d7209d56458c5dfcabdc69/v-next/hardhat/src/internal/builtin-plugins/solidity-test/type-extensions.ts#L39), it should be `Record<string, SensitiveString>`

Document the Solidity test configuration https://github.com/NomicFoundation/hardhat-website/issues/79

PR #7432, opened 2025-09-13
add configVariable support for test.solidity.forking config

iosh:

---

#7403 
This PR adds support for `configVariable` to the `test.solidity.forking` configuration.

Key changes:
- Updated the forking configuration types:
  - `url?: string;` is now `url?: SensitiveString;`
  - `rpcEndpoints?: Record<string, string>;` is now `rpcEndpoints?: Record<string, SensitiveString>;`
  
- Updated the Zod validation to support the new `SensitiveString` type.
- Change the `solidityTestConfigToSolidityTestRunnerConfigArgs` function to `async` to support resolving `configVariable` values inside it.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
@@ -1,6 +1,14 @@
 import type { HardhatUserConfig } from "../../../config.js";
-import type { HardhatConfig } from "../../../types/config.js";
+import type {
+  ConfigurationVariableResolver,
+  HardhatConfig,
+  ResolvedConfigurationVariable,
+} from "../../../types/config.js";
 import type { HardhatUserConfigValidationError } from "../../../types/hooks.js";
+import type {
+  SolidityTestForkingConfig,
+  SolidityTestUserConfig,
+} from "../../../types/test.js";
 
 import path from "node:path";
 
@@ -8,6 +16,8 @@ import { isObject } from "@nomicfoundation/hardhat-utils/lang";
 import { resolveFromRoot } from "@nomicfoundation/hardhat-utils/path";
 import {
   conditionalUnionType,
+  sensitiveStringSchema,
+  sensitiveUrlSchema,
   validateUserConfigZodType,
 } from "@nomicfoundation/hardhat-zod-utils";
 import { z } from "zod";
@@ -49,9 +59,9 @@ const solidityTestUserConfigType = z.object({
     .optional(),
   forking: z
     .object({
-      url: z.string().optional(),
+      url: z.optional(sensitiveUrlSchema),
       blockNumber: z.bigint().optional(),
-      rpcEndpoints: z.record(z.string()).optional(),
+      rpcEndpoints: z.record(sensitiveStringSchema).optional(),
     })
     .optional(),
   invariant: z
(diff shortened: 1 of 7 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -133,7 +133,7 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
   }
 
   const config: SolidityTestRunnerConfigArgs =
-    solidityTestConfigToSolidityTestRunnerConfigArgs(
+    await solidityTestConfigToSolidityTestRunnerConfigArgs(
```
michalbrabec, line 136: Would it be possible resolving the URLs for `resolveSolidityTestUserConfig` at this point?
It would allow us to avoid making `solidityTestConfigToSolidityTestRunnerConfigArgs` async, but I would avoid making the change, since it would make the code more messy.

iosh: I tried fetching the URLs inside the resolveSolidityTestUserConfig function to avoid making solidityTestConfigToSolidityTestRunnerConfigArgs async, but I got an error: "Context must be set before running non-config hooks."

michalbrabec: Let's leave it as is, since this is more logical structure

v-next/hardhat/src/internal/builtin-plugins/solidity-test/type-extensions.ts
```diff
@@ -66,9 +70,16 @@ declare module "../../../types/test.js" {
...
-  // eslint-disable-next-line @typescript-eslint/no-empty-interface -- This could be an extension point
-  export interface SolidityTestConfig extends SolidityTestUserConfig {}
+  export interface SolidityTestForkingConfig {
+    url?: ResolvedConfigurationVariable;
+    blockNumber?: bigint;
+    rpcEndpoints?: Record<string, ResolvedConfigurationVariable>;
+  }
 
+  export interface SolidityTestConfig
+    extends Omit<SolidityTestUserConfig, "forking"> {
```
michalbrabec: Why is `SolidityTestUserConfig` inherited this way?
I would prefer to create a new interface without `forking` (for example `SolidityTestConfigBase`. Define different interfaces for the user and resolved forking and then compose `SolidityTestUserConfig` and `SolidityTestConfig` as `SolidityTestConfig  extends SolidityTestConfigBase, SolidityTestConfigForking {...}`.

Another example you might follow is  the property `mnemonic` in `EdrNetworkHDAccountsConfig` and `EdrNetworkHDAccountsUserConfig` the interfaces are defined independently.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
```diff
@@ -98,9 +98,22 @@ export function solidityTestConfigToSolidityTestRunnerConfigArgs(
 
   const blockDifficulty = config.prevRandao;
 
-  const ethRpcUrl = config.forking?.url;
+  let ethRpcUrl: string | undefined;
+  if (config.forking?.url !== undefined) {
```
michalbrabec, line 102: Nit: Could be shortened as `let ethRpcUrl: string | undefined  =  config.forking?.url && await config.forking.url.get();`, guard operator.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -88,6 +98,32 @@ const userConfigType = z.object({
...
+  if (forkingUserConfig.rpcEndpoints !== undefined) {
+    for (const [name, url] of Object.entries(forkingUserConfig.rpcEndpoints)) {
+      resolvedRpcEndpoints[name] = resolveConfigurationVariable(url);
+    }
+  }
+
+  return {
+    ...forkingUserConfig,
+    url:
+      forkingUserConfig.url !== undefined
```
michalbrabec, line 120: Nit: Can be shortened to `forkingUserConfig.url && resolveConfigurationVariable(forkingUserConfig.url)`

iosh: ESLint requires null check.

michalbrabec: Agreed, stricter type control disables this.

PR #7438, opened 2025-09-15
fix: prevent production keystore access during tests in configuration-variables

Fibonacci747:
When running tests (HH_TEST === "true"), we must avoid any interaction with the production keystore to prevent password prompts. This change:
- Skips the production keystore fallback in fetchValue when in test mode.
- Adds an early return in getValue for the prod path when tests are running, preventing loader initialization and password requests.

This aligns behavior with the in-file comment (“only dev keystore to avoid prompting for the production keystore password”) and avoids unintended prompts. CI bypass remains intact; non-test behavior is unchanged.

### DIFF
--- a/v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
+++ b/v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
@@ -45,10 +45,12 @@ export default async (): Promise<Partial<ConfigurationVariableHooks>> => {
 
       // Then, if the development keystore does not have the key and `fetchValue` is not called from a test,
       // attempt to retrieve the value from the production keystore.
-      value = await getValue(context, variable, false, onlyAllowDevKeystore);
+      if (!onlyAllowDevKeystore) {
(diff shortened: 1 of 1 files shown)


v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
```diff
@@ -45,10 +45,12 @@ export default async (): Promise<Partial<ConfigurationVariableHooks>> => {
 
       // Then, if the development keystore does not have the key and `fetchValue` is not called from a test,
       // attempt to retrieve the value from the production keystore.
-      value = await getValue(context, variable, false, onlyAllowDevKeystore);
```
michalbrabec, line 48: The logic should be unified, at this point `getValue` has sufficient information to enforce the behaviour
It can check for the actual value only `if (isDevKeystore || !onlyAllowDevKeystore)`

michalbrabec:
Please add tests for this behavior so we do not regress in the future.

michalbrabec:
Hi,

Issue #7290, opened 2025-08-24
HHE900: error resolving project file referenced in `npmFilesToBuild`


mdesousa:
### Version of Hardhat

3.0.1

### What happened?

we have proxy contracts and are importing openzeppelin files as shown below.
during ignition deployment get an HHE900 error (see details with stack trace at the bottom). the project file mentioned in the error `./@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol` exists under `/artifacts` but it's a folder, not a file.

```
const config: HardhatUserConfig = {
  plugins: [hardhatToolboxViem],
  solidity: {
    version: '0.8.30',
    npmFilesToBuild: [
      '@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol',
      '@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol',
    ],
  },
  networks,
  verify: { etherscan: { apiKey: ETHERSCAN_API_KEY } },
};
```

error details and stack trace
```
const config: HardhatUserConfig = {
  plugins: [hardhatToolboxViem],
  solidity: {
    version: '0.8.30',
    npmFilesToBuild: [
      '@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol',
      '@openzeppelin/contracts/proxy/utils/UUPSUpgradeable.sol',
    ],
  },
  networks,
  verify: { etherscan: { apiKey: ETHERSCAN_API_KEY } },
};
```

### Minimal reproduction steps

Deploy a UUPSUpgradable contract with ignition... see module code below.

```
import { buildModule } from '@nomicfoundation/hardhat-ignition/modules';

const ERC20CreateImplModule = buildModule('ERC20CreateImplModule', (m) => {
  const impl = m.contract('ERC20', []);
  return { impl };
});

export const ERC20CreateModule = buildModule('ERC20CreateModule', (m) => {
  const { impl } = m.useModule(ERC20CreateImplModule);
  const init = m.encodeFunctionCall(impl, 'initialize', []);
  const id = { id: 'ERC20Proxy' };
  const proxy_ = m.contract('ERC1967Proxy', [impl, init], id);
  const proxy = m.contractAt('UUPSUpgradeable', proxy_);
  return { proxy, impl };
});

export default ERC20CreateModule;
```

michalbrabec:
@mdesousa can you help me to reproduce your issue please?
I was not able to reproduce the issue.

My approach is:
[...]

PR #7442, opened 2025-09-16
HHE900: error resolving project file referenced in npmFilesToBuild

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/7290


v-next/hardhat-verify/src/internal/artifacts.ts
```diff
@@ -72,12 +72,18 @@ export async function getBuildInfoAndOutput(
 // TODO: consider moving this to the solidity build system as a helper function
 export async function getCompilerInput(
   solidity: SolidityBuildSystem,
-  rootFilePath: string,
+  rootFilePath: string | undefined, // Undefined when the target source is in an npm package
   sourceName: string,
   buildProfileName: string,
 ): Promise<CompilerInput> {
+  const rootFilePaths = [
```
michalbrabec: I would prefer the caller to make the decision what should be the `rootFilePath` - move the condition to the code calling this function. To align with a single-purpose principle, the decision should not be the responsibility of this function, it should just get the compiler input.

v-next/hardhat-verify/src/internal/verification.ts
```diff
@@ -193,10 +193,19 @@ Explorer: ${instance.getContractUrl(address)}`);
     contractInformation.userFqn,
   );
 
+  let root: string | undefined = config.paths.root;
+  let sourceName = contractInformation.sourceName;
+  if (contractInformation.inputFqn.startsWith("npm/")) {
```
michalbrabec: I would expect that prefixing the path would be a behaviour of the `ContractInformation`, could be an NPM-specific implementation of the interface that understands that it must prefix the path with `npm:`.

How is it possible that HH is able to verify the `ERC20` but it fails to verify the `ERC1967Proxy` (https://github.com/NomicFoundation/hardhat/issues/7290#issuecomment-3244536003)?

ChristopherDedominici: I believe the reason is that the `ERC20` contract is defined within the hardhat project and imports openzeppelin packages installed via npm. As a result, during verification, these contracts are treated as dependencies and processed accordingly (hardhat-verify knows how to handle them).

However, the issue happens when a contract is imported directly from openzeppelin (not created within the hardhat project, e.g.: not located in the contracts folder), deployed independently, and then verified. 
In this case, the problem occurs during the verification process: if the contract isn't part of the hardhat project, `hardhat-verify` struggles to correctly resolve its npm based dependencies. This appears to be a bug in `hardhat-verify`.

This would explain why the `Sivo` contract was deployed and verified successfully, while the `ERC1967Proxy` contract was not.

fvictorio:
Removing myself as a reviewer until we confirm that the issue is still happening. Feel free to re-add me later if needed.

v-next/hardhat-verify/src/internal/artifacts.ts
```diff
@@ -72,12 +72,17 @@ export async function getBuildInfoAndOutput(
 // TODO: consider moving this to the solidity build system as a helper function
 export async function getCompilerInput(
   solidity: SolidityBuildSystem,
-  rootFilePath: string,
+  root: string,
   sourceName: string,
+  isNpmModule: boolean,
   buildProfileName: string,
 ): Promise<CompilerInput> {
+  const rootFilePath = isNpmModule
```
ChristopherDedominici, line 80: Here I am using the name `rootFilePath` because the param name for the function `getCompilationJobs` is `rootFilePaths: string[]`

v-next/hardhat-verify/src/internal/verification.ts
```diff
@@ -32,7 +32,7 @@ export interface VerifyContractArgs {
   address: string;
   constructorArgs?: unknown[];
   libraries?: LibraryAddresses;
-  contract?: string;
+  contract?: string; // format: <source-name>:<contract-name>
```
ChristopherDedominici: It would be better to name this property `fullyQualifiedName`, but that would be a breaking change. It would also require renaming parameters in other places (e.g., Ignition), which is where the bug addressed by this PR is generated.

Do we want to do it in this PR, or open another one to better investigate better names? I'd open another one

.changeset/itchy-sheep-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
+Fixed a bug that prevented verification of contracts imported from npm modules [7442](https://github.com/NomicFoundation/hardhat/pull/7442)
```
fvictorio:
```suggestion
Fixed a bug that prevented verification of contracts imported from npm modules ([#7442](https://github.com/NomicFoundation/hardhat/pull/7442))
```

v-next/hardhat-verify/src/internal/verification.ts
fvictorio: Ah, makes sense. I think just documenting it is fine. But let's use a proper tsdoc so that it shows up in IDEs:

```
  /** The fully qualified name of the contract, in the format: `<source-name>:<contract-name>` */
  contract?: string;
```

Which looks nice enough:

Image

fvictorio (approved):
Minor comments, pre-approving assuming that this was confirmed manually to work (I didn't test it).

v-next/hardhat-verify/src/internal/verification.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7442/commits/d60c7c5c47d57764731ece9856b8a2ecd8002870

PR #7453, opened 2025-09-18
chore: revert the release commit

kanej:
Due to issues on npm we cancelled the release after the release branch was merged but before the publish happened in the CI.

This commit undoes the release commit, reverting version numbers and changelogs back.


.github/workflows/add-issue-to-project.yml
michalbrabec, line 1: Can we disable the action without deleting the definition?
Screenshot 2025-09-18 at 13 57 22

kanej: I have enabled the automatic workflow under the Hardhat project for adding issues to projects. So I don't think we need this CI workflow anymore? Is there a reason to keep it?

alcuadrado: I don't think we'll ever enable a workflow that requires a PAT again, so I don't see a reason to keep it.

PR #7458, opened 2025-09-20
Expose  with Hardhat and EDR package versions

### LINKED ISSUE #7322: Add a `versions` object to the HRE
### Describe the feature

Add a `versions` field to the HRE. This is an object with the versions of Hardhat and EDR. It could have other things in the future if it makes sense, but that seems unlikely.

Example of trying this in `hh console`:

```
> hre.versions
{ hardhat: "3.1.0", edr: "0.12.0" }
```

Implementation tip: we don't want to load EDR for this, because it would make the HRE creation slower. We can do `await import("@nomicfoundation/edr/package.json", { with: { type: "json" } })` instead. This will stop working if EDR becomes ESM, but we can think about that when/if it happens.

iosh:

---

close #7322

### DIFF
--- a/v-next/hardhat/src/types/hre.ts
+++ b/v-next/hardhat/src/types/hre.ts
@@ -2,6 +2,11 @@ import type { GlobalOptions } from "./global-options.js";
 import type { UserInterruptionManager } from "./user-interruptions.js";
 import type { HardhatConfig, HardhatUserConfig } from "../types/config.js";
 
+export interface HardhatRuntimeEnvironmentVersions {
+  readonly hardhat: string;
+  readonly edr: string;
+}
+
 /**
  * The Hardhat Runtime Environment (HRE) is an object that exposes
  * all the functionality available through Hardhat.
@@ -11,6 +16,7 @@ export interface HardhatRuntimeEnvironment {
   readonly userConfig: HardhatUserConfig;
   readonly globalOptions: GlobalOptions;
   readonly interruptions: UserInterruptionManager;
+  readonly versions: HardhatRuntimeEnvironmentVersions;
   // These fields are defined using module agumentation despite being part of
   // Hardhat's core:
   // readonly hooks: HookManager;
--- a/.changeset/early-hairs-buy.md
+++ b/.changeset/early-hairs-buy.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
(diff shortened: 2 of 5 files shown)


v-next/hardhat/src/types/hre.ts
```diff
@@ -2,6 +2,11 @@ import type { GlobalOptions } from "./global-options.js";
 import type { UserInterruptionManager } from "./user-interruptions.js";
 import type { HardhatConfig, HardhatUserConfig } from "../types/config.js";
 
+export interface HardhatRuntimeEnvironmentVersions {
+  readonly hardhat: string;
```
michalbrabec: Nit: I would prefer to name the properties `hardhatVersion` and `edrVersion`, even though it is a bit redundant

v-next/hardhat/src/types/hre.ts
```diff
@@ -2,6 +2,11 @@ import type { GlobalOptions } from "./global-options.js";
 import type { UserInterruptionManager } from "./user-interruptions.js";
 import type { HardhatConfig, HardhatUserConfig } from "../types/config.js";
 
+export interface HardhatRuntimeEnvironmentVersions {
+  readonly hardhatVersion: string;
+  readonly edrVersion: string;
```
fvictorio: I think these should be just `hardhat` and `edr`, adding `Version` seems redundant to me.

v-next/hardhat/src/types/hre.ts
```diff
@@ -2,6 +2,11 @@ import type { GlobalOptions } from "./global-options.js";
 import type { UserInterruptionManager } from "./user-interruptions.js";
 import type { HardhatConfig, HardhatUserConfig } from "../types/config.js";
 
+export interface HardhatRuntimeEnvironmentVersions {
```
fvictorio: I'm not convinced about the name of this type, and given this is public API, we can't change it later.

On the other hand, I don't have a good alternative (perhaps `HardhatVersions`?). But I don't think we need to export this. We can just remove the `export`, and use an inferred type in `src/internal/core/hre.ts`

michalbrabec: I’d be for an inferred type in this case since it is very simple.

v-next/hardhat/src/types/hre.ts
michalbrabec: I requested the change to this, sorry for the confusion @iosh

PR #7460, opened 2025-09-21
add compilation progress spinner to show build progress

iosh:

---

Add compilation progress spinner to improve user experience during Solidity builds.

**Before**
```bash
hardhat compile
```
image

**After**
image
image

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -23,6 +23,7 @@ import type {
   CompilerOutputError,
 } from "../../../../types/solidity/compiler-io.js";
 import type { SolidityBuildInfo } from "../../../../types/solidity.js";
+import type { ISpinner } from "../../../utils/spinner.js";
 
 import os from "node:os";
 import path from "node:path";
@@ -50,6 +51,7 @@ import debug from "debug";
 import pMap from "p-map";
 
 import { FileBuildResultType } from "../../../../types/solidity/build-system.js";
+import { createSpinner } from "../../../utils/spinner.js";
 import { DEFAULT_BUILD_PROFILE } from "../build-profiles.js";
 
 import {
@@ -175,11 +177,32 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       ..._options,
     };
 
-    if (!options.quiet) {
-      console.log(`Compiling your Solidity ${options.scope}...`);
-    }
+    const spinner: ISpinner | undefined = options.quiet
+      ? undefined
+      : createSpinner({ text: `Compiling your Solidity ${options.scope}...` });
 
     await this.#downloadConfiguredCompilers(options.quiet);
+
+    spinner?.start();
+
+    let spinnerFinalized = false;
+
+    const finalizeSpinner = (
+      onSpinner: (active: ISpinner) => void,
+      onNoSpinner?: () => void,
+    ) => {
+      if (spinnerFinalized) {
+        return;
+      }
+
+      if (spinner !== undefined) {
+        onSpinner(spinner);
+      } else {
+        onNoSpinner?.();
+      }
+
+      spinnerFinalized = true;
+    };
 
     const { buildProfile } = this.#getBuildProfile(options.buildProfile);
 
@@ -189,6 +212,9 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     );
 
     if ("reason" in compilationJobsResult) {
+      finalizeSpinner((s) =>
+        s.fail(`Failed to compile Solidity ${options.scope}`),
+      );
       return compilationJobsResult;
     }
 
@@ -199,6 +225,23 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       ...new Set(compilationJobsPerFile.values()),
     ];
 
+    const totalJobs = runnableCompilationJobs.length;
+    let completedJobs = 0;
+
+    const updateProgress = () => {
+      if (spinner === undefined || spinnerFinalized) {
+        return;
+      }
+      if (totalJobs === 0) {
+        spinner.update(`Compiling your Solidity ${options.scope}...`);
+        return;
+      }
+      spinner.update(
+        `Compiling your Solidity ${options.scope} (${completedJobs}/${totalJobs})...`,
+      );
+    };
+
+    updateProgress();
     // NOTE: We precompute the build ids in parallel here, which are cached
     // internally in each compilation job
     await Promise.all(
@@ -214,7 +257,8 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
           runnableCompilationJob,
           options,
         );
-
+        completedJobs += 1;
+        updateProgress();
         return {
           compilationJob: runnableCompilationJob,
           compilerOutput: output,
@@ -238,6 +282,11 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     const isSuccessfulBuild =
       uncachedResults.length === uncachedSuccessfulResults.length;
 
+    if (!isSuccessfulBuild) {
+      finalizeSpinner((s) =>
+        s.fail(`Failed to compile Solidity ${options.scope}`),
+      );
+    }
     const contractArtifactsGeneratedByCompilationJob: Map<
       CompilationJob,
       ReadonlyMap<string, string[]>
@@ -330,6 +379,12 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
             contractArtifactsGenerated.get(userSourceName) ?? [],
           warnings: errors,
         });
+      }
+    }
+    if (!spinnerFinalized) {
+      finalizeSpinner((s) => s.stop());
+      if (totalJobs === 0 && !options.quiet) {
+        console.log("Nothing to compile");
       }
     }
 
@@ -1074,10 +1129,6 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       string,
       Map<string, CompilationJob[]>
     >();
-
-    if (runnableCompilationJobs.length === 0) {
-      console.log("Nothing to compile");
-    }
 
     for (const job of runnableCompilationJobs) {
       const solcVersion = job.solcConfig.version;
--- a/.changeset/two-lions-judge.md
+++ b/.changeset/two-lions-judge.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Add compilation progress spinner to show build progress
--- a/v-next/hardhat/src/internal/utils/spinner.ts
+++ b/v-next/hardhat/src/internal/utils/spinner.ts
@@ -1,0 +1,163 @@
+import chalk from "chalk";
+
+const FRAMES = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"];
+const FRAME_INTERVAL_MS = 80;
+
+export interface SpinnerOptions {
+  text?: string;
+  stream?: NodeJS.WriteStream;
+  enabled?: boolean;
+}
+
+export interface ISpinner {
+  readonly isEnabled: boolean;
+  start(text?: string): void;
+  update(text: string): void;
+  stop(text?: string): void;
+  stopAndPersist(text: string, symbol?: string): void;
+  succeed(text?: string): void;
+  fail(text?: string): void;
+}
+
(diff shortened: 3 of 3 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -1069,16 +1085,14 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     }
   }
 
-  async #printCompilationResult(runnableCompilationJobs: CompilationJob[]) {
+  async #printCompilationResult(
```
michalbrabec: Nit: why was the return type added, when the implementation barely changed?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -333,8 +345,12 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       }
     }
 
-    if (!options.quiet) {
-      if (isSuccessfulBuild) {
+    if (totalJobs === 0 && !options.quiet) {
```
michalbrabec: Nit: The condition structure could be hard to follow, I would prefer a clearer structure where the conditions do not repeat, for example:
```typescript
if(!quiet) {
  if(totalJobs === 0) {
  }
  else {
    if (isSuccessfulBuild) { ...}
  }
}
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -199,6 +206,8 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       ...new Set(compilationJobsPerFile.values()),
     ];
 
+    const totalJobs = runnableCompilationJobs.length;
```
michalbrabec: Nit: the variable is not needed, length can be accessed when needed later.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -333,8 +345,12 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       }
     }
 
-    if (!options.quiet) {
-      if (isSuccessfulBuild) {
+    if (totalJobs === 0 && !options.quiet) {
+      spinner.stopAndPersist("Nothing to compile");
```
michalbrabec: I would prefer to keep this as part of `#printCompilationResult`, purely close the spinner here and then print whatever is needed later.
This would also help simplify the spinner, which would no longer need multiple ways to stop.
It would just clear it's text, stop interval and leave the rest of the code to print as needed

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,272 @@
+const FRAMES = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"];
+const FRAME_INTERVAL_MS = 80;
+
+export interface ISpinner {
```
michalbrabec, line 4: I'd love to simplify the class as much as possible, I have dropped some ideas bellow.
I would prefer to start with a bare-bones implementation and expand later if needed.

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,272 @@
...
+
+    this.#text = initialText;
+    this.#silent = silent;
+  }
+
+  public get isSpinning(): boolean {
+    return this.#isSpinning;
+  }
+
+  public runWithPause(action: () => void): void {
```
michalbrabec: I'd prefer to drop the method and instead call explicitely
```typescript
spinner.stop();
action(...);
spinner.start();
```

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,272 @@
...
+ * Falls back to console logs when output is disabled.
+ */
+class Spinner implements ISpinner {
+  public readonly isEnabled: boolean;
+
+  #text: string;
+  #frameIndex = 0;
+
+  #interval: NodeJS.Timeout | undefined;
+  #isSpinning = false;
```
michalbrabec: This logic could be removed and instead spin after `start` and stop following `stop`, `finalize` ...

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,272 @@
...
+    this.#render(false);
+    this.#interval = setInterval(() => this.#render(true), FRAME_INTERVAL_MS);
+  }
+
+  /**
+   * Replace the current message while the spinner is running.
+   *
+   * @param nextText New message to display.
+   */
+  public update(nextText: string): void {
```
michalbrabec: Not used anywhere, can be dropped

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,272 @@
...
+    this.#stopInternal();
+  }
+
+  /**
+   * Stop the spinner and leave the last line in the terminal.
+   *
+   * @param text Text to print on the final line.
+   * @param symbol Optional prefix placed before the text.
+   */
+  public stopAndPersist(text: string, symbol?: string): void {
```
michalbrabec: I would drop this method too, let `stop` clean the spinner output and the calling code print whatever is needed.

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,272 @@
...
+      this.start(this.#text);
+    }
+  }
+
+  /**
+   * Begin rendering frames or log the first message when disabled.
+   *
+   * @param nextText Optional text to show while running.
+   */
+  public start(nextText?: string): void {
```
michalbrabec: The `nextText` is duplicating the constructor parameter, `start` can use the value provided and rely on `update` for change of the value.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -292,12 +298,19 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
         ),
       );
 
-      this.#printSolcErrorsAndWarnings(errors);
+      if (!options.quiet) {
```
michalbrabec: Why is the check for `options.quiet` needed here?
The spinner is created regardless of the option and started at the 187, we can just stop and start again regardless.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -292,12 +298,19 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
+        spinner.stop();
+      }
 
+      this.#printSolcErrorsAndWarnings(errors);
       const successfulResult = !this.#hasCompilationErrors(
         result.compilerOutput,
       );
 
+      if (!options.quiet) {
```
michalbrabec: Same as above, why is the check needed?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -333,10 +346,14 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       }
     }
 
-    if (!options.quiet) {
-      if (isSuccessfulBuild) {
+    if (isSuccessfulBuild) {
+      spinner.stop();
```
michalbrabec: `spinner.stop();` is called in both branches of the if, we can call it just before the if and simplify it that way

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,180 @@
...
+ *
+ * Falls back to console logs when output is disabled.
+ */
+class Spinner implements ISpinner {
+  public readonly isEnabled: boolean;
+  readonly #text: string;
+  #interval: NodeJS.Timeout | null = null;
+  #loggedFallback = false;
+  readonly #stream: NodeJS.WriteStream;
+  readonly #silent: boolean;
```
michalbrabec: The logic of this variable almost completely the same as `isEnabled`, would it be possible to drop it and rely purely on `isEnabled`?
In the `build-system.ts` they are even set to the same value `enabled: !options.quiet, silent: options.quiet,`. Complementary value to be more precise since one is in positive logic and the other in negative.

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,180 @@
...
+        console.log(this.#text);
+        this.#loggedFallback = true;
+      }
+      return;
+    }
+
+    this.#stopAnimation();
+    let frameIndex = 0;
+
+    const tick = () => {
```
michalbrabec: This variable is not necessary, the arrow function could be directly used as the param for `setInterval`

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,180 @@
...
+      this.#render(FRAMES[frameIndex]);
+      frameIndex = (frameIndex + 1) % FRAMES.length;
+    };
+
+    this.#interval = setInterval(tick, FRAME_INTERVAL_MS);
+    tick();
```
michalbrabec: It should be sufficient to rely purely on `setInterval` to render the content, the timeout is 80ms which is almost imperceptible.

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  /**
+   * Stop the spinner without printing a final line.
+   */
+  public stop(): void {
+    this.#stopAnimation();
+  }
+
+  #clearLine(): void {
+    if (
```
michalbrabec: Why is the type check needed here, TypeScript should help us make sure that the two are functions.
Plus there is no check for the `write` function.

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,180 @@
...
+    ) {
+      this.#stream.clearLine(0);
+      this.#stream.cursorTo(0);
+    } else {
+      this.#stream.write("\r");
+    }
+  }
+
+  #render(frame: string): void {
+    if (!this.isEnabled || this.#interval === null) {
```
michalbrabec: Why is there a check for the `#interval`?
Render should basically do what the name implies, render the text (when enabled).

v-next/hardhat-utils/src/spinner.ts
```diff
@@ -0,0 +1,180 @@
...
+ */
+class Spinner implements ISpinner {
+  public readonly isEnabled: boolean;
+  readonly #text: string;
+  #interval: NodeJS.Timeout | null = null;
+  #loggedFallback = false;
+  readonly #stream: NodeJS.WriteStream;
+  readonly #silent: boolean;
+
+  constructor(
```
michalbrabec: Why are the `SpinnerOptions` used in the `createSpinner` function, but not here?
I would prefer to use them in both cases or none, the second option would be even cleaner (using the same parameters for the function as the constructor).

v-next/hardhat-utils/src/spinner.ts
iosh: This check is necessary because cursor control only exists in a TTY environment. The `write` function is a basic method so we don't neec check.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -333,6 +341,8 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       }
     }
 
+    spinner.stop();
```
fvictorio: I'd remove this second pair of start/stop calls. What happens in between is really fast so this seems unnecessary. Besides, even if it were very slow, the result would be that you get this sequence:

- The spinner is shown
- The spinner is hidden
- Warnings and errors are (perhaps) logged
- The spinner is shown again (why?)
- The spinner is hidden

v-next/hardhat-utils/src/spinner.ts
fvictorio: This makes me realize that maybe we should always set `this.isEnabled` to false when `this.#stream.isTTY` is false. That would also make these checks unnecessary.

@iosh wdyt?

fvictorio: This makes me realize that maybe we should always set `this.isEnabled` to false when `this.#stream.isTTY` is false. That would also make these checks unnecessary.

@iosh wdyt?

PR #7464, opened 2025-09-22
Add `defineConfig` in `hardhat/config`

alcuadrado:
This is an idea that @fvictorio had, based on `eslint`, but also used by `astro`.

Instead of letting the user do:

```ts
import type { HardhatUserConfig } from "hardhat/config";

const config: HardhatUserConfig = {};

export default config;
```

we can have a `defineConfig` function so it's just:

```ts

export default defineConfig({});
```

which is shorter, type-safe, and has autocomplete on javascript too (!)

Note: we can't use `defineConfig` in our own tests, because we use `--isolatedDeclarations`. I added a comment about it to the tsdoc of the function.

Documentation update: https://github.com/NomicFoundation/hardhat-website/pull/125

### DIFF
--- a/.changeset/changeset_pass-config-to-ignition-helpers.md
+++ b/.changeset/changeset_pass-config-to-ignition-helpers.md
@@ -1,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-ignition-ethers": patch
+"@nomicfoundation/hardhat-ignition-viem": patch
+---
+
+Fixed a bug preventing Ignition from using the `hre.config.ignition` settings when deploying via script ([#7641](https://github.com/NomicFoundation/hardhat/issues/7641))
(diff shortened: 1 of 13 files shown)


alcuadrado:
I initially thought this wasn't going to work, as the template packages needs to point to the newest version of hardhat. But turns out that changesets also bumps the version range in the template packages :)

I added a PR to the website that needs to be merged once this is released.

Also, I updated the readmes of all the packages. I don't think this requires bumping their peerDependency on hardhat, as it's just a documentation doc. /cc @kanej

alcuadrado:
@michalbrabec this can be re-reviewed now.

alcuadrado:
Also, after releasing this we should update the branches `main` and `tutorial` of https://github.com/NomicFoundation/hardhat3-plugin-template, but it can't be done yet.

fvictorio (approved):

PR #7466, opened 2025-09-23
feat: conditional dependencies

### LINKED ISSUE #7424: Add support for conditional dependencies to V3
### Describe the feature

Some plugins may want to provide additional (optional) features if another plugin is available. For example, a plugin that affects the connection may also perform additional actions on the `connection.ethers` or `connection.viem` objects if the corresponding plugins are installed. If either plugin is not installed, then it should not cause an error.

an example of plugin that would use that feature can be found here: https://github.com/Amxx/hardhat-predeploy/tree/optional-dependencies

The context is outlined in [[Public] Conditional plugins](https://www.notion.so/Public-Conditional-plugins-high-level-idea-and-type-extensions-validation-270578cdeaf5804fb947f33daea71bf3)

antico5:
Closes #7424

### DIFF
--- a/v-next/hardhat/src/internal/core/plugins/resolve-plugin-list.ts
+++ b/v-next/hardhat/src/internal/core/plugins/resolve-plugin-list.ts
@@ -88,27 +88,44 @@ async function reverseTopologicalSort(
       }
 
       for (const conditionalDependency of plugin.conditionalDependencies) {
+        // Check all condition plugins are installed
+        let conditionModules;
         try {
-          // Check all condition plugins are installed
-          const conditionModules = await Promise.all(
+          conditionModules = await Promise.all(
             conditionalDependency.condition(),
           );
+        } catch (_error) {
+          continue;
+        }
 
-          // Check all condition plugins are loaded
-          if (
-            conditionModules.some(
-              (conditionPlugin) =>
-                !resolvedPlugins.includes(conditionPlugin.default),
-            )
-          ) {
-            continue;
-          }
+        // Check all condition plugins are loaded
+        if (
+          conditionModules.some(
+            (conditionPlugin) =>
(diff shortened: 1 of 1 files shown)


v-next/hardhat/src/internal/core/plugins/resolve-plugin-list.ts
```diff
@@ -69,12 +69,49 @@ async function reverseTopologicalSort(
...
+      if (plugin.conditionalDependencies === undefined) {
+        continue;
+      }
+
+      for (const conditionalDependency of plugin.conditionalDependencies) {
+        try {
+          // Check all condition plugins are installed
+          const conditionModules = await Promise.all(
+            conditionalDependency.condition(),
+          );
```
alcuadrado, line 96:
```suggestion
        let conditionModules;
        try {
          // Check all condition plugins are installed
          const conditionModules = await Promise.all(
            conditionalDependency.condition(),
          );
        } catch {
           continue;
        }
```

alcuadrado: This way this try/catch only catches errors from the `Promise.all`.

v-next/hardhat/src/internal/core/plugins/resolve-plugin-list.ts
```diff
@@ -69,12 +69,49 @@ async function reverseTopologicalSort(
...
+            conditionModules.some(
+              (conditionPlugin) =>
+                !resolvedPlugins.includes(conditionPlugin.default),
+            )
+          ) {
+            continue;
+          }
+
+          // Load the conditional dependency
+          const pluginModule = await conditionalDependency.plugin();
```
alcuadrado: This one can fail like in line 53, so it should have a similar try/catch

Issue #7428, opened 2025-09-11
Implement gas statistics for js/ts tests


schaable:
### Describe the issue

Add gas usage statistics collection and reporting during test execution as specified in the [design doc](https://www.notion.so/nomicfoundation/Gas-statistics-263578cdeaf580e2907cd714fb1814c0).

Tasks:

-  [x] Create `gas-analytics` plugin with `--gas-stats` global flag
-  [x] Add `GasAnalyticsManager` to HRE
 - [x] Implement `onGasMeasurement` hook handler in `gas-analytics` plugin
-  [x] Add `onCollectedGasReportCallback` callback support in `network-manager`
-  [x] Implement EDR gas report conversion utilities
 - [x] Add statistics calculation (min, avg, median, max, count)
-  [x] Integrate with test nodejs test subtask
 - [x] Implement table output formatting
 - [x] Handle overloaded function signatures
- [x] Add tests

PR #7472, opened 2025-09-25
Collect and display gas statistics on JavaScript/TypeScript tests

schaable:
Implements gas usage statistics collection and reporting during JavaScript/TypeScript test execution.

### Changes
 - Added `gas-analytics` built-in plugin with `--gas-stats` global flag.
 - Integrated gas data collection from EDR via `onCollectedGasReportCallback` callback.
 - Implemented statistics calculation (min, avg, median, max, call count) for contract functions.
 - Added formatted table output displayed after test completion. 
 - Supported JavaScript/TypeScript test runners.

### Technical Details

 - Gas measurements are collected via the `onGasMeasurement` hook in the `network-manager` plugin.
 - Statistics are aggregated by contract and function, with proper handling of overloaded functions. For deployments, only the last deployment is saved.

### Usage
```
hardhat test nodejs --gas-stats
```

### Output Example
```
// TODO
```

Closes https://github.com/NomicFoundation/hardhat/issues/7428

### DIFF
--- a/v-next/hardhat-utils/src/internal/format.ts
+++ b/v-next/hardhat-utils/src/internal/format.ts
@@ -1,0 +1,5 @@
+export function getStringWidth(str: string): number {
+  // Remove ANSI escape codes if present
+  const stripped = str.replace(/\u001b\[[0-9;]*m/g, "");
+  return stripped.length;
+}
--- a/v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-analytics-manager.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-analytics-manager.ts
@@ -1,9 +1,61 @@
 import type { GasAnalyticsManager, GasMeasurement } from "./types.js";
 
+import crypto from "node:crypto";
+import path from "node:path";
+
+import {
+  ensureDir,
+  remove,
+  writeJsonFile,
+} from "@nomicfoundation/hardhat-utils/fs";
+import debug from "debug";
+
+const gasStatsLog = debug(
+  "hardhat:core:gas-analytics:gas-analytics-manager:gas-stats",
+);
+
 export class GasAnalyticsManagerImplementation implements GasAnalyticsManager {
-  public async addGasMeasurement(
-    _gasMeasurement: GasMeasurement,
-  ): Promise<void> {
+  public gasMeasurements: GasMeasurement[] = [];
+  readonly #gasStatsPath: string;
+
+  constructor(gasStatsPath: string) {
+    this.#gasStatsPath = gasStatsPath;
+  }
+
+  public addGasMeasurement(gasMeasurement: GasMeasurement): void {
+    this.gasMeasurements.push(gasMeasurement);
+
+    gasStatsLog(
+      "Added gas measurement",
+      JSON.stringify(
+        gasMeasurement,
+        (_, val) => (typeof val === "bigint" ? val.toString() : val),
+        2,
+      ),
+    );
+  }
+
+  public async clearGasStats(id: string): Promise<void> {
+    const dataPath = await this.#getGasStatsPath(id);
+    await remove(dataPath);
+    this.gasMeasurements = [];
+    gasStatsLog("Cleared gas stats from disk and memory");
+  }
+
+  public async saveGasStats(id: string): Promise<void> {
+    const dataPath = await this.#getGasStatsPath(id);
+    const filePath = path.join(dataPath, `${crypto.randomUUID()}.json`);
+    await writeJsonFile(filePath, this.gasMeasurements);
+    gasStatsLog("Saved gas stats", id, filePath);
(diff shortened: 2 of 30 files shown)


v-next/hardhat-utils/src/internal/format.ts
```diff
@@ -0,0 +1,5 @@
+export function getStringWidth(str: string): number {
+  // Remove ANSI escape codes if present
+  const stripped = str.replace(/\u001b\[[0-9;]*m/g, "");
```
fvictorio, line 10: I'm pretty sure this is good enough for what we have now, and it will probably never cause issues, but a proper ANSI strip looks like this: https://github.com/chalk/ansi-regex/blob/main/index.js

I'd update the comment saying that it's (intentionally) incomplete, just for future reference.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * NOTE: The following helpers interact with the global HRE instance only;
+ * This is OK because:
+ * - They are intended for the internal use only. They are exposed via the
+ *   internal public API only.
+ * - We know the HRE has been initialized by the time they are used.
+ */
+
+export function getGasStatsPath(rootPath: string): string {
+  return path.join(rootPath, "coverage");
```
fvictorio: We shouldn't use the `coverage` path to save the gas measurements. I think I'd use `hre.config.paths.cache` for this.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-analytics-manager.ts
```diff
@@ -0,0 +1,315 @@
...
+    this.#gasStatsPath = gasStatsPath;
+  }
+
+  public addGasMeasurement(gasMeasurement: GasMeasurement): void {
+    this.gasMeasurements.push(gasMeasurement);
+
+    gasStatsLog(
+      "Added gas measurement",
+      JSON.stringify(gasMeasurement, null, 2),
+    );
```
fvictorio: I don't think this is going to be useful debug information: it's too granular, and it could mostly be obtained from disk if necessary. I would remove it.

v-next/hardhat/test/internal/builtin-plugins/gas-analytics/gas-analytics-manager.ts
```diff
@@ -0,0 +1,1001 @@
...
+      });
+    });
+
+    describe("saveGasMeasurements", () => {
+      afterEach(async () => {
+        await remove(path.join(process.cwd(), "gas-stats"));
+      });
+
+      it("should save gas measurements in memory", async () => {
+        const manager = new GasAnalyticsManagerImplementation(process.cwd());
```
fvictorio: I think you should use a tmp dir for this. Otherwise, if any of these tests fail after a measurement has been written, you'll have to manually delete the gas-stats directory wherever you run the tests.

With that approach you can also drop the `afterEach` cleanup.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/helpers.ts
schaable: That's an unintentional leftover from migrating logic from the coverage plugin. Good catch!

schaable: Done in https://github.com/NomicFoundation/hardhat/pull/7472/commits/1c5ebdbbc68787b715e80d26026ce1a6d9de1c14

v-next/hardhat/test/internal/builtin-plugins/gas-analytics/gas-analytics-manager.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/7472/commits/f15c8e0c4d7c5dbb5d606caf92cd1000fabda906

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-analytics-manager.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/7472/commits/5f69b05fdb81aa2f6157fe297a2db6db1a69d500

v-next/hardhat-utils/src/internal/format.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/7472/commits/5f96b70ca6cf056ac8c460518e2aa791a64610bb

Ideally, we should switch to something like the `string-width` package, but I didn't want to add a new dependency due to the recent npm events.

PR #7476, opened 2025-09-26
Fix userSourceName generation from relative paths on artifact cleanup

### LINKED ISSUE #7459: Ignition upgradable contract case in Hardhat 3 does not work
### Version of Hardhat

3.0.3

### What happened?

I added `npmFilesToBuild ` to `hardhat.config.ts` under `solidity`:

> solidity: {
>     // ...
>     npmFilesToBuild: [
>       "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol",
>       "@openzeppelin/contracts/proxy/transparent/TransparentUpgradeableProxy.sol",
>     ],
>   }

run test `npx hardhat test .\test\ProxyDemo.ts`
Then, after running, the following error was obtained:
`HardhatError: HHE1000: Artifact for contract "ProxyAdmin" not found.`

Tried Node version v22.10.0 and 24.6.0.

[Part 1: Deploying our proxies](https://hardhat.org/ignition/docs/guides/upgradeable-proxies#part-1-deploying-our-proxies)
[Part 2: Interacting with our proxy](https://hardhat.org/ignition/docs/guides/upgradeable-proxies#part-2-interacting-with-our-proxy)
[Part 3: Upgrading our proxy with an initialization function](https://hardhat.org/ignition/docs/guides/upgradeable-proxies#part-3-upgrading-our-proxy-with-an-initialization-function)
[Part 4: Interacting with our upgraded proxy](https://hardhat.org/ignition/docs/guides/upgradeable-proxies#part-4-interacting-with-our-upgraded-proxy)
[Testing our Ignition modules](https://hardhat.org/ignition/docs/guides/upgradeable-proxies#testing-our-ignition-modules)

### Search terms

_No response_

antico5:
Closes #7459

### DIFF
--- a/v-next/hardhat-utils/src/path.ts
+++ b/v-next/hardhat-utils/src/path.ts
@@ -53,3 +53,19 @@ export function shortenPath(absolutePath: string): string {
 
   return absolutePath;
 }
+
+/**
+ *  Ensures the given relative path is in posix format
+ *  Throws if the string argument is not a relative path
+ */
+export function toPosixRelativePath(thePath: string): string {
+  if (path.isAbsolute(thePath)) {
+    throw new Error(`Expected a relative path, got absolute: "${thePath}"`);
+  }
+
+  // Normalize first so things like "foo/../bar" collapse consistently
+  const normalized = path.normalize(thePath);
+
+  // Convert backslashes (Windows) to forward slashes
+  return normalized.split(/[\\\/]/).join(path.posix.sep);
+}
--- a/.changeset/fast-turkeys-applaud.md
+++ b/.changeset/fast-turkeys-applaud.md
@@ -1,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-utils": patch
+"hardhat": patch
+---
+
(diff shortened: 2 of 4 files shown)


v-next/hardhat-utils/src/path.ts
```diff
@@ -53,3 +53,19 @@ export function shortenPath(absolutePath: string): string {
 
   return absolutePath;
 }
+
+/**
+ *  Ensures the given relative path is in posix format
+ *  Throws if the string argument is not a relative path
+ */
+export function toPosixRelativePath(thePath: string): string {
+  if (path.isAbsolute(thePath)) {
```
michalbrabec: I would prefer a more robust check for the value of `thePath`, right now the function rejects absolute paths, but accepts random strings (and replaces `\` there).
This function is in utils and is shared so it should be more hardened against unexpected input.

antico5: @michalbrabec done, reduced the scope to avoid touching hardhat-utils. This is a very specific place where it was needed and we didn't require a full blown relative path conversion function

michalbrabec: Makes sense, approved ✅

PR #7492, opened 2025-09-30
chore: include minimum release age

kanej:
Use the new minimum release age feature of pnpm to reduce local development install risks.

Also set the expected pnpm version to the most recent version - which includes this capability.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,2 @@
 hoist-pattern[]=!@types/*
-minimum-release-age=50240
+minimum-release-age=10080


.npmrc
```diff
@@ -1 +1,2 @@
 hoist-pattern[]=!@types/*
+minimum-release-age=50240
```
michalbrabec: How was this value selected? 
is it the maximum?
It is about a month.

alcuadrado: I think i'd do 7days, which is 10080

.github/actions/setup-env/action.yml
```diff
@@ -5,7 +5,7 @@ inputs:
   pnpm-version:
     description: Version of pnpm to install
     required: false
-    default: "9"
+    default: "10.17.1"
```
alcuadrado, line 8: note: i don't think this need to be aligned with the vesion in the package.json, but it has to be a recent (10.something).

pnpm will see the package.json and download the right version automatically

michalbrabec: I like the explicit version, but I;m not sure if there are any down sides to this approach.
Is there a reason to not be specific in this case?

.npmrc
michalbrabec: 7 days should be enough

kanej:
```suggestion
minimum-release-age=10080
```

.github/actions/setup-env/action.yml
kanej: Apparently, we can just omit the version from the `pnpm/action-setup` and it will take the version specified in the `packageManager` field: https://github.com/pnpm/action-setup?tab=readme-ov-file#install-only-pnpm-with-packagemanager

I have applied that change.

.npmrc
kanej: I have updated to the 7 day constraint.

PR #7493, opened 2025-09-30
Improve error test rejected promise

### LINKED ISSUE #7321: Print a better error in our unhandled promise rejection handler if an `AssertionError` is detected
### Describe the feature

_Opening this issue, as I think we can improve the situation described here: https://github.com/NomicFoundation/hardhat/issues/7317_

## Context

`chai` was designed before promises where a thing, `chai-as-promised` helps, but it's also confusing, because sometimes you have to `await` your calls to `expect`/`assert`s, and sometimes you don't. This means that the `await`s are easy to forget.

When you forget an `await` of an `async` `expect`, you get an undhandled promise rejection with an `AssertionError`. 

We should detect that situation and print a better error, possibly with a link to the docs explaining how to know when you should `await`.

### Search terms

_No response_

antico5:
Closes #7321 

This PR handles the issue where a user forgets to `await` async matchers (e.g.. `to.revert()`). In practice when this happens, an unhandled rejection is emitted and bubbles up  to the process level.  The implementation handles this in two ways, one for when mocha is running in parallel mode, and other in non parallel mode.

When running in non parallel mode, the tests are ran in the same process, so I just import a file that sets up the handler. No "grace period" is needed in tihs case as the process wont exit until floating promises are finished.

When running in parallel mode, mocha spawns multiple processes, probably one per test suite, so we use the `require` mocha option to load the handler as a hook. The problem with these processes is that they exit as soon as the tests finish, and unresolved promises are discarded, so the handler never gets called. The only workaround I found was to add a "grace period" of 100 ms per process, that they will wait before exiting. That'll give time for the rejected promise to float up. I assume that 100ms is more than enough for EDR which I assume is what they are always running in mocha tests.

In any case I didn't find a way to link to the line/test case  where the promise is rejected, so I just show a general error.

### DIFF
--- a/v-next/hardhat-mocha/src/unhandled-rejection-mocha-hook.ts
+++ b/v-next/hardhat-mocha/src/unhandled-rejection-mocha-hook.ts
@@ -18,9 +18,9 @@ process.on("exit", () => {
     console.log(
       chalk.red(
         [
-          'Error: Missing "await" on async expectation.',
+          'Error: Missing "await" on async assert/expect.',
           "",
-          'You called "expect(...)" on a value that returns a Promise, but did not "await" it or return it from the test.',
+          'You called "expect(...)" or an assertion function on a value that returns a Promise, but did not "await" it or return it from the test.',
           "",
           "This means the assertion ran asynchronously and Mocha may finish the test before the assertion actually fails.",
         ].join("\n"),
@@ -31,8 +31,9 @@ process.on("exit", () => {
 
 // Only used in parallel mode. Mocha will load and execute this hook
 // This grace period is required otherwise mocha just kills the child processes before they get notice of the unhandled rejections
-// The value is what we think is appropriate to wait for e.g. an EDR reverted transaction
-const GRACE_TIME_MS = 100;
+// The value is what we think is appropriate to wait for e.g. an EDR reverted transaction.
+// It could be a border case where this value is too low, but we prefer to eventually lose a warning than to significantly slow down all test suites
+const GRACE_TIME_MS = 10;
 
 export const mochaHooks = {
   async afterAll(): Promise<void> {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -892,7 +892,7 @@ importers:
         version: link:../hardhat-zod-utils
       chalk:
         specifier: ^5.3.0
-        version: 5.4.1
+        version: 5.6.2
       hardhat:
         specifier: workspace:^3.0.0
         version: link:../hardhat


v-next/hardhat-mocha/src/unhandled-rejection-mocha-hook.ts
```diff
@@ -0,0 +1,41 @@
+import chalk from "chalk";
+
+// This file is loaded directly by our mocha runner task, as well as by mocha when ran in parallel mode, via the `require` arg.
+// The purpose of this patch is to detect unhandled rejected promises that are caused by not awaiting async expectations
+
+let showNotAwaitedError = false;
+
+process.on("unhandledRejection", (e: Error) => {
```
alcuadrado, line 8: Does this work when telemetry is enabled? because it sets a handler that does `process.exit(1)`.

You can run with `DEBUG=hardhat* pnpm hardhat test`, and a it will print a log about telemetry enabledness.

alcuadrado: If the telemetry handler runs first, it may kill the process, and this one may never be run.

v-next/hardhat-mocha/src/unhandled-rejection-mocha-hook.ts
```diff
@@ -0,0 +1,41 @@
...
+    showNotAwaitedError = true;
+  }
+});
+
+process.on("exit", () => {
+  if (showNotAwaitedError) {
+    console.log(
+      chalk.red(
+        [
+          'Error: Missing "await" on async expectation.',
```
alcuadrado:
```suggestion
          'Error: Missing "await" on async assert/expect.',
```

v-next/hardhat-mocha/src/unhandled-rejection-mocha-hook.ts
```diff
@@ -0,0 +1,41 @@
...
+          "",
+          'You called "expect(...)" on a value that returns a Promise, but did not "await" it or return it from the test.',
```
alcuadrado:
```suggestion
          'You called "expect(...)" or an assertion function on a value that returns a Promise, but did not "await" it or return it from the test.',
```

v-next/hardhat-mocha/src/unhandled-rejection-mocha-hook.ts
alcuadrado: We validated this

antico5:
@alcuadrado ready for re-review

alcuadrado (approved):

PR #7504, opened 2025-10-02
use package.json files instead of .npmignore for ignition packages

### LINKED ISSUE #7446: Use package.files instead of `.npmignore` in ignition packages
### Describe the issue

The Ignition packages use an `.npmignore` file instead of explicitly listing the files to publish in the `package.json`, as the rest of the packages do. This is probably an inconsistency left from what Ignition had its own repo.

I noticed this because the installed Ignition package doesn't have the `src` directory, while the rest of the packages do.

We should:

- Remove all `.npmignore` files
- Replace them with an explicit list of `files` to publish, which includes `src`

iosh:

---

 - replace .npmignore with package.json files arrays for all ignition packages
 - keep the current publish behavior and only add the `src/` directory to the files lists

## Steps

Example for `@nomicfoundation/ignition-ui`:

```bash
    cd v-next/hardhat-ignition-ui
    mkdir -p ../tmp-packs
    pnpm pack --pack-destination ../tmp-packs
    tar -tf ../tmp-packs/nomicfoundation-ignition-ui-3.0.3.tgz > ../tmp-packs/ignition-ui-before.txt
```

Then remove .npmignore and add this to package.json:

```
    "files": [
      "dist/",
      "src/",
      "CHANGELOG.md",
      "README.md",
      "eslint.config.js",
      "vite.config.js",
      "vite.config.d.ts",
      "tsconfig.node.tsbuildinfo",
      "tsconfig.tsbuildinfo"
    ]
```
After that:

```bash
   pnpm pack --pack-destination ../tmp-packs
    tar -tf ../tmp-packs/nomicfoundation-ignition-ui-3.0.3.tgz > ../tmp-packs/ignition-ui-after.txt
    diff -u ../tmp-packs/ignition-ui-before.txt ../tmp-packs/ignition-ui-after.txt
```

The diff shows only the new src/* files and nothing removed.

Questions:

1. pnpm copies the repo root LICENSE into these tarballs because the ignition packages don’t have their own LICENSE file. Do we want to add a package-level LICENSE?
2. These packages still include config files like vite.config.js and the tsconfig.* files. Should we remove them?

resolve #7446


fvictorio:
> pnpm copies the repo root LICENSE into these tarballs because the ignition packages don’t have their own LICENSE file. > Do we want to add a package-level LICENSE?

Yes, I guess it should have the same license that the other plugins do.

> These packages still include config files like vite.config.js and the tsconfig.* files. Should we remove them?

I don't think so. Those are only relevant during development, so I don't see any value in including them.

Issue #7053, opened 2025-07-23
Integrated test summary between different test tasks


kanej:
Show the combined results of a test run at the end when invoking `npx hardhat test`.

PR #7514, opened 2025-10-06
Add summary outputs when multiple test runners are invoked

zoeyTM:
closes https://github.com/NomicFoundation/hardhat/issues/7053

I'm not entirely sure this is exactly what we want. In my opinion, it looks a bit sloppier than just the runners on their own, but i'm happy to workshop the output if we decide to.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
@@ -47,6 +47,8 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
     hre._coverage.disableReport();
   }
 
+  const testSummaries: Record<string, string> = {};
+
   for (const subtask of thisTask.subtasks.values()) {
     const files = getTestFilesForSubtask(subtask, testFiles, subtasksToFiles);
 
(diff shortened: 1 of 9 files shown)


v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -70,9 +83,69 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
       args.verbosity = verbosity;
     }
 
-    await subtask.run(args);
+    if (subtask.options.has("testSummaryIndex")) {
+      args.testSummaryIndex = failureIndex;
+
+      const summaryId = subtask.id.join("-");
```
fvictorio: I would use this instead:

```suggestion
      const summaryId = subtask.id[subtask.id.length - 1];
```

Otherwise the output looks like this:

```
5 passing (3 test-solidity, 2 test-nodejs)
```

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -89,6 +162,22 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
   }
 };
 
+function logSummaryLine(
+  label: string,
+  items: Array<[string, number]>,
+  color: ChalkInstance = chalk.white,
+): void {
+  let total = 0;
+  const todoStr = items
```
fvictorio: `todoStr`?

fvictorio:
Spacing is slightly broken.

With only a nodejs failure, an empty line is needed:

image

With only a solidity failure, it's fine:

image

When there are failures in both nodejs and solidity, there's a double space between the first and the second error:

image

The last example also shows an inconsistency between the nodejs and solidity outputs that looks a bit ugly to me, but I would consider it out of the scope for this PR. The fix IMO is to change the solidity test output to be `1) contracts/Counter.t.sol:CounterTest:test_initialValue()` (maybe using color for contrast)`.

fvictorio:
And the CI failures don't look like false negatives to me, unless I'm missing something?

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
zoeyTM: i truly don't know how i typed "todo" here

fvictorio (approved):

PR #7525, opened 2025-10-08
Expose ignition retry loop variables in user config (Hardhat v3)

zoeyTM:
closes https://github.com/NomicFoundation/hardhat/issues/7303

docs: https://github.com/NomicFoundation/hardhat-website/pull/190

### DIFF
--- a/v-next/hardhat-ignition/src/internal/tasks/deploy.ts
+++ b/v-next/hardhat-ignition/src/internal/tasks/deploy.ts
@@ -190,14 +190,26 @@ const taskDeploy: NewTaskActionFunction<TaskDeployArguments> = async (
 
   const strategyConfig = hre.config.ignition.strategyConfig?.[strategyName];
 
+  if (
+    hre.config.ignition.maxRetries === undefined &&
+    hre.config.networks[connection.networkName]?.ignition.maxRetries !==
+      undefined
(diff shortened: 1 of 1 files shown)


v-next/hardhat-ignition-core/src/deploy.ts
```diff
@@ -74,6 +76,8 @@ export async function deploy<
   maxPriorityFeePerGas?: bigint;
   gasPrice?: bigint;
   disableFeeBumping?: boolean;
+  maxRetries?: number;
```
michalbrabec: Why are we adding the property here, when the `config` property contains the value as well?
The final value is calculated as follows, which can be done by the `deploy` function.
```typescript
    maxRetries:
      hre.config.ignition.maxRetries ??
      hre.config.networks[connection.networkName]?.ignition.maxRetries,
```
This could also concentrate the logic in one place instead of spreading it through all the `ignition-helper`s.

zoeyTM: The first part is a good catch, but as far as I know, I don't think we can remove the logic from the ignition-helpers as they don't deploy via the `deploy` function inside of `hardhat-ignition`. They bypass it and go straight to `hardhat-ignition-core`

PR #7528, opened 2025-10-08
[html coverage] Add vendored package

ChristopherDedominici:
Link to internal design doc: https://www.notion.so/nomicfoundation/Code-coverage-HTML-report-286578cdeaf580bcad22db58ce54fb47?source=copy_link

This PR only adds the `vendored package`. 
The PR was supposed to also add the HTML report, but after discovering that the coverage data has several bugs, we decided to include only the `vendored package` for now and add the HTML report after fixing the coverage data bug.

### TODOs before merge
- [x] add mention to vendored code 

### NOTEs for review
- All the code inside the folder `v-next/hardhat-vendored/src/coverage-module` has been copied from vendored packages, so there’s no need for a deep review, a quick scan will be sufficient

### DIFF
--- a/scripts/lib/packages.mjs
+++ b/scripts/lib/packages.mjs
@@ -8,14 +8,22 @@ const packagesDir = "v-next";
  * Read all the package.json files of the packages that we release to npm.
  */
 export async function readAllReleasablePackages() {
-  const allPackageNames = (await readdir(packagesDir))
-    .filter(file => !['config', 'example-project', 'template-package', 'hardhat-test-utils'].includes(file));
+  const allPackageNames = (await readdir(packagesDir)).filter(
+    (file) =>
+      ![
+        "config",
+        "example-project",
+        "template-package",
+        "hardhat-test-utils",
+        "hardhat-vendored",
+      ].includes(file)
+  );
 
   return Promise.all(allPackageNames.map(readPackage));
 }
 
 export async function readPackage(name) {
-  return JSON.parse(await readFile(`./v-next/${name}/package.json`, 'utf-8'));
+  return JSON.parse(await readFile(`./v-next/${name}/package.json`, "utf-8"));
 }
 
 export async function getLatestPackageVersionFromNpm(name) {


v-next/hardhat/assets/index.js
```diff
@@ -0,0 +1,2 @@
+import libCoverage from "./lib-coverage/index.cjs";
```
ChristopherDedominici: This is a "bridge" to expose the common js code into ESM js code, so that the ts file can then import it

v-next/hardhat-vendored/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "lint:fix": "pnpm prettier --write && pnpm eslint --fix",
+    "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
+    "prettier": "prettier \"**/*.{ts,js,md,json}\"",
+    "test": "node --import tsx/esm --test --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/*.ts\" \"test/!(fixture-projects|helpers)/**/*.ts\"",
+    "test:only": "node --import tsx/esm --test --test-only --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/*.ts\" \"test/!(fixture-projects|helpers)/**/*.ts\"",
+    "test:coverage": "c8 --reporter html --reporter text --all --exclude test --exclude \"src/**/{types,type-extensions}.ts\" --src src node --import tsx/esm --test --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/*.ts\" \"test/!(fixture-projects|helpers)/**/*.ts\"",
+    "pretest": "pnpm build",
+    "pretest:only": "pnpm build",
+    "build": "tsc --build .",
+    "postbuild": "node dist/src/copy-assets.js",
```
ChristopherDedominici, line 33: This step is required to copy the `assets` folder, which contains the CSS and JS files used to generate the report, into the `dist` folder after compilation

v-next/hardhat-vendored/src/copy-assets.ts
```diff
@@ -0,0 +1,33 @@
+import { cp } from "node:fs/promises";
```
ChristopherDedominici, line 1: This function is invoked by the `postbuild` in the `package.json` file.

Used to copy the assets folder, which contains the CSS and JS files used to generate the report, into the dist folder after compilation

ChristopherDedominici: I'm open to better ideas to solve this issue

v-next/hardhat-vendored/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
```
ChristopherDedominici, line 1: I need a double-check here. I copied the configuration from `tsconfig.base.json`, but removed all the properties that would cause errors during compilation, since this folder contains non-ESM files

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -548,4 +559,54 @@ export class CoverageManagerImplementation implements CoverageManager {
 
     return rows.map((row) => `| ${row.join(" | ")} |`).join("\n");
   }
+
+  async #writeHtmlReport(
+    report: Report,
+    htmlReportPath: string,
+  ): Promise<void> {
+    const baseDir = process.cwd();
```
ChristopherDedominici: Since there is a bug in how we generate coverage data, we currently only display executed and non-executed lines. The line counter is either 1 for executed lines or 0 for non-executed lines:

Image

This will be improved after fixing how we collect coverage data

v-next/hardhat-vendored/test/index.ts
ChristopherDedominici, line 1: This file is intentionally empty to prevent build errors and avoid changing the default scripts in `package.json`

.changeset/gorgeous-swans-try.md
```diff
@@ -0,0 +1,6 @@
+---
+"@nomicfoundation/hardhat": minor
+"@nomicfoundation/hardhat-vendored": patch
```
fvictorio: A couple of things:

- `@nomicfoundation/hardhat` should be `hardhat`
- The changeset for Hardhat should be patch. I don't think this merits a minor bump.
- `@nomicfoundation/hardhat-vendored` is a new package, so it shouldn't be included here. Otherwise its first version will be 3.0.1, not 3.0.0. I checked locally that `changeset publish` will publish `hardhat-vendored` even if it doesn't have a changeset, because it detects that it's not published yet. (I'm not entirely sure that the CI will have the same behavior, but I don't see why it wouldn't.)

```suggestion
"hardhat": patch
```

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -119,6 +126,10 @@ export class CoverageManagerImplementation implements CoverageManager {
     await writeUtf8File(lcovReportPath, lcovReport);
     log(`Saved lcov report to ${lcovReportPath}`);
 
+    const htmlReportPath = path.join(this.#coveragePath, "html");
```
fvictorio: Nyc, jest and vitest all dump the html inside the coverage directory, even when both the HTML and lcov reporters are used together. I personally assume that `coverage/index.html` will exist and was a bit surprised when it didn't.

I'm not against using a subdirectory for this (in fact, it would be my preference if we were doing this from scratch), but I think it's better to be consistent with other tools here.

```suggestion
    const htmlReportPath = path.join(this.#coveragePath);
```

.changeset/gorgeous-swans-try.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Added HTML coverage reporting and a new internal `hardhat-vendored` package ([7528](https://github.com/NomicFoundation/hardhat/pull/7528)).
```
fvictorio: Changesets shouldn't include implementation details.

```suggestion
Added HTML coverage reporting ([7528](https://github.com/NomicFoundation/hardhat/pull/7528)).
```

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -119,6 +126,10 @@ export class CoverageManagerImplementation implements CoverageManager {
...
+    await this.#writeHtmlReport(report, htmlReportPath);
+    console.log(`Saved html report to ${htmlReportPath}`);
```
fvictorio: I think you meant to use a debug log here.

```suggestion
    log(`Saved html report to ${htmlReportPath}`);
```

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -557,4 +568,54 @@ export class CoverageManagerImplementation implements CoverageManager {
...
+    report: Report,
+    htmlReportPath: string,
+  ): Promise<void> {
+    const baseDir = process.cwd();
+    const coverageMap = libCoverage.createCoverageMap({});
+
+    // Construct coverage data for each tested file,
+    // detailing whether each line was executed or not.
+    for (const [p, fileCoverageInput] of Object.entries(report)) {
+      const testedFilePath = path.isAbsolute(p) ? p : path.join(baseDir, p);
```
fvictorio: Not a blocker, but: I am surprised about this. We generate the report, so we should know if the keys are relative or absolute paths. (The `Report` interface names the key `relativePath`, so I guess they are meant to be relative).

Another concern is that, if they are relative, it's unclear whether they are relative to the cwd or the project's root (IMO it should be the latter).

Again: I wouldn't do anything in this PR, but this seems like technical debt. Not sure how/if we are capturing those.

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -557,4 +568,54 @@ export class CoverageManagerImplementation implements CoverageManager {
...
+    }
+
+    await mkdir(htmlReportPath);
+
+    const context = libReport.createContext({
+      dir: htmlReportPath,
+      coverageMap,
+    });
+
+    reports.create("html", undefined).execute(context);
```
fvictorio: Optional parameter with default value.

```suggestion
    reports.create("html").execute(context);
```

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -557,4 +568,54 @@ export class CoverageManagerImplementation implements CoverageManager {
...
+          start: { line, column: 0 },
+          end: { line, column: 0 },
+        };
+
+        if (fileCoverageInput.partiallyExecutedLines.has(line)) {
+          fc.s[line] = 0; // mark as not covered
+          continue;
+        }
+
+        // TODO: bug line count
```
fvictorio: What does this mean?

v-next/hardhat-vendored/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "@nomicfoundation/hardhat-vendored",
+  "version": "3.0.0",
+  "description": "Internal dependencies used by Hardhat that have been vendored to prevent bloating the main package",
+  "homepage": "https://github.com/nomicfoundation/hardhat/tree/v-next/v-next/hardhat-vendored",
```
fvictorio: We are not using the `v-next` branch anymore.

```suggestion
  "homepage": "https://github.com/nomicfoundation/hardhat/tree/main/v-next/hardhat-vendored",
```

v-next/hardhat-vendored/test/index.ts
fvictorio: This should be a comment inside the file.

v-next/hardhat-vendored/src/copy-assets.ts
fvictorio: I tested this PR with verdaccio and there were no issues, so this seems to work fine and it's simple enough.

fvictorio:
For the record: the way uncovered code looks is not great:

image

I think this is caused by us using lines instead of statements, or something along those lines. My guess is that we can work our way around this, but I wouldn't do it yet; I'd rather release this sooner.

v-next/hardhat-vendored/test/index.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7528/commits/392f8f90cc6160f6574926c84cbbc59519a3f2aa

.changeset/gorgeous-swans-try.md
ChristopherDedominici: Since this PR only adds the vendored package, I removed the changeset file: https://github.com/NomicFoundation/hardhat/pull/7528/commits/fe70838193923e1a7a6743e65625448e010e53e9

.changeset/gorgeous-swans-try.md
ChristopherDedominici: Since this PR only adds the vendored package, I removed the changeset file: https://github.com/NomicFoundation/hardhat/pull/7528/commits/fe70838193923e1a7a6743e65625448e010e53e9

PR #7567, opened 2025-10-09
feat(ignition-core): add tests for using ReadEventArgumentFuture values

Galoretka:
- Add “using the values” tests in v-next/hardhat-ignition-core/test/readEventArgument.ts
- Verify ReadEventArgumentFuture can be:
  - used as constructor argument (dependency wired)
  - used as function argument (dependency wired)
  - used as value for contract deployment (value + dependency)
  - used as value for contract call (value + dependency)
  - used as address in send (to + dependency)
  - used as address in contractAt (address + dependency)


michalbrabec:
Hi,

PR #7577, opened 2025-10-14
Fix Etherscan Verification Failing to Retry with Full Solc Input

SyedAsadKazmi:
## Summary

This PR fixes a critical bug in Etherscan contract verification where detailed failure messages prevented the automatic retry with full solc input, causing verification to fail even though it would succeed with the full compiler input.

## Problem

When verifying contracts on Etherscan, users were experiencing failures with the error:

```
HHE80024: The contract verification status polling encountered an error: 
"Fail - Unable to verify. Compiled contract deployment bytecode does NOT match 
the transaction deployment bytecode.". Verification may still succeed.
```

However, the same verification would succeed on Blockscout, which properly retried with the full solc input after the minimal input failed.

### Root Cause

There were **two bugs** in the Etherscan verification status polling logic:

1. **Bug 1 - Exact String Matching**: The `isFailure()` method used exact string matching (`===`) instead of prefix matching (`startsWith()`), so it couldn't recognize failure messages that included additional details.

   ```typescript
   // Before (line 386)
   public isFailure(): boolean {
     return this.message === "Fail - Unable to verify";  // ❌ Exact match only
   }
   ```

2. **Bug 2 - Incorrect Logic Order**: The polling logic checked `isOk()` before checking `isFailure()`, causing it to throw an error before recognizing legitimate failures. When Etherscan returns a failure, the status code is `0`, making `isOk()` return `false` and immediately throwing an error.

   ```typescript
   // Before (lines 323-328)
   if (!etherscanResponse.isOk()) {  // ❌ Throws error before checking isFailure()
     throw new HardhatError(
       HardhatError.ERRORS.HARDHAT_VERIFY.GENERAL.CONTRACT_VERIFICATION_STATUS_POLLING_FAILED,
       { message: etherscanResponse.message },
     );
   }
   ```

This prevented the automatic retry mechanism from kicking in, even though the verification logic was designed to retry with full solc input on failure.

### Change 1: Use Prefix Matching for Failure Detection

Changed the `isFailure()` method to use `startsWith()` to properly detect failure messages with additional details:

```typescript
// After (line 386)
public isFailure(): boolean {
  return this.message.startsWith("Fail - Unable to verify");  // ✅ Prefix match
}
```

### Change 2: Reorder Logic to Check Success/Failure First

Reordered the polling logic to check for recognized success/failure states before throwing errors:

```typescript
// After (lines 323-328)
if (etherscanResponse.isFailure() || etherscanResponse.isSuccess()) {
  return {
    success: etherscanResponse.isSuccess(),
    message: etherscanResponse.message,
  };
}

if (!etherscanResponse.isOk()) {
  throw new HardhatError(
    HardhatError.ERRORS.HARDHAT_VERIFY.GENERAL.CONTRACT_VERIFICATION_STATUS_POLLING_FAILED,
    { message: etherscanResponse.message },
  );
}
```

### Before

```
=== Etherscan ===
📤 Submitted source code for verification...
⏳ Waiting for verification result...
❌ HHE80024: The contract verification status polling encountered an error...
(Verification stops here, never retries with full input)

=== Blockscout ===
📤 Submitted source code for verification...
⏳ Waiting for verification result...
The initial verification attempt failed using the minimal compiler input.
Trying again with the full solc input...
✅ Contract verified successfully on Blockscout!
```

### After

```
=== Etherscan ===
📤 Submitted source code for verification...
⏳ Waiting for verification result...
The initial verification attempt failed using the minimal compiler input.
Trying again with the full solc input...
📤 Submitted source code for verification...
⏳ Waiting for verification result...
✅ Contract verified successfully on Etherscan!

=== Blockscout ===
(Already verified)
```

## Testing

Tested with a Counter contract deployed to Sepolia:
- ✅ Verification now properly retries with full solc input on Etherscan
- ✅ Successfully verifies contracts that previously failed
- ✅ Behavior now matches Blockscout verification flow
- ✅ Verified contract: https://sepolia.etherscan.io/address/0x0d1Bc9C51556586246aE16733B49C54F4f95BbFd#code

## Files Changed

- `v-next/hardhat-verify/src/internal/etherscan.ts`
  - Fixed `isFailure()` method to use prefix matching
  - Reordered `pollVerificationStatus()` logic to check success/failure before throwing errors

## Breaking Changes

None. This is a bug fix that makes Etherscan verification work as originally intended.

## Related Issues

Fixes the issue where Etherscan verification fails with:
- `"Compiled contract deployment bytecode does NOT match the transaction deployment bytecode"`
- While the same contract successfully verifies on Blockscout using full solc input

This aligns Etherscan verification behavior with Blockscout, ensuring both providers properly utilize the two-step verification process (minimal input first, then full input on failure).

### DIFF
--- a/v-next/hardhat-verify/test/etherscan.ts
+++ b/v-next/hardhat-verify/test/etherscan.ts
@@ -599,6 +599,71 @@ describe("etherscan", () => {
         assert.equal(response.message, "Fail - Unable to verify");
       });
 
+      it("should handle detailed failure messages with additional information", async () => {
+        const etherscan = new Etherscan({
+          ...etherscanConfig,
+          dispatcher: testDispatcher.interceptable,
+        });
+
+        const detailedErrorMessage =
+          "Fail - Unable to verify. Compiled contract deployment bytecode does NOT match the transaction deployment bytecode.";
+
+        pollVerificationStatusInterceptor.reply(200, {
+          status: "0",
+          result: detailedErrorMessage,
+        });
+
+        let response: { success: boolean; message: string } | undefined;
+        try {
+          response = await etherscan.pollVerificationStatus(
+            guid,
+            address,
+            contract,
+          );
+        } catch {
+          assert.fail("Expected pollVerificationStatus to not throw an error");
+        }
+
+        assert.equal(response.success, false);
+        assert.equal(response.message, detailedErrorMessage);
+      });
+
+      it("should handle various detailed failure messages starting with 'Fail - Unable to verify'", async () => {
+        const etherscan = new Etherscan({
+          ...etherscanConfig,
+          dispatcher: testDispatcher.interceptable,
+        });
+
+        const failureMessages = [
+          "Fail - Unable to verify. Some other reason.",
+          "Fail - Unable to verify. Constructor arguments do not match.",
+          "Fail - Unable to verify. Invalid compiler version specified.",
+        ];
+
+        for (const failureMessage of failureMessages) {
+          pollVerificationStatusInterceptor.reply(200, {
+            status: "0",
+            result: failureMessage,
+          });
+
+          let response: { success: boolean; message: string } | undefined;
+          try {
+            response = await etherscan.pollVerificationStatus(
+              guid,
+              address,
+              contract,
+            );
+          } catch {
+            assert.fail(
+              "Expected pollVerificationStatus to not throw an error",
+            );
+          }
+
+          assert.equal(response.success, false);
+          assert.equal(response.message, failureMessage);
+        }
+      });
+
       it("should poll the verification status until it is successful or fails", async () => {
         const etherscan = new Etherscan({
           ...etherscanConfig,


SyedAsadKazmi:
@michalbrabec, @alcuadrado, any updates regarding this PR?

alcuadrado:
Hey @SyedAsadKazmi, sorry for the delay. We'll release some etherscan related fixes next, and will take a look at it after. Thanks for sending the PR!

SyedAsadKazmi:
Noted, @alcuadrado. Thanks!

schaable:
Hi @SyedAsadKazmi, could you share the steps to reproduce that specific error?
If your local contracts don’t match the deployed bytecode, it’s usually caught by:

> HHE80009: The address contains a contract whose bytecode does not match any of your local contracts.

The order of these checks in the `verify` and `pollVerificationStatus` functions is intentional, it helps ensure any change in the Etherscan API doesn’t go unnoticed. That might actually be what’s happening here.

I’ll need the sample contracts you used to trigger this scenario. I tested with the Counter contract you shared and was able to verify it successfully, so it could be related to your compiler settings or the way you’re verifying (for example, via the programmatic API). Could you share the exact contract and detailed steps to reproduce the issue?

SyedAsadKazmi:
Hello @schaable,

It's reproducible unless we use the exact same `settings` in `default` and `production` profiles, like:

```ts
solidity: {
  profiles: {
    default: {
      version: "0.8.24",
      settings: {
        optimizer: {
          enabled: true,
          runs: 1_000,
        },
      }
    },
    production: {
      version: "0.8.24",
      settings: {
        optimizer: {
          enabled: true,
          runs: 1_000,
        },
      }
    },
  },
},
```

When you run `npx hardhat compile` without specifying the profile, then the compilation is done using the `default`, but during the verification process, it's picking the `production` profile by default.

So, generally speaking, the problem is more related to `verify` picking the `production` profile by default, as ideally it should pick the `default` profile. But, if it's supposed to be designed this way only, then using the full solc input on first failure (i.e., minimal input first, then full input on failure), solves the problem and also aligns the Etherscan verification behavior with that of the Blockscout.

schaable:
Hi @SyedAsadKazmi! Thanks for the follow-up, that helps clarify things. I’m still unable to reproduce the issue though. Here are the steps I’m using:

```
pnpm hardhat compile
pnpm hardhat ignition deploy ./ignition/modules/Counter.ts --network sepolia
pnpm hardhat verify --network sepolia ADDRESS
```

Am I missing something? Are you deploying your contract with Ignition? If you can share a small repro repo, that would be great!
I’ll likely take another look at this during or after Devconnect, as we’re currently getting ready for it. Sorry for the delay!

SyedAsadKazmi:
Hello @schaable,

No worries at all!

Here’s the [minimal repo](https://github.com/SyedAsadKazmi/hh3-example). You can follow the steps provided in the [README](https://github.com/SyedAsadKazmi/hh3-example?tab=readme-ov-file#steps-to-redproduce-the-verification-issue) to reproduce the issue.

No rush, enjoy Devconnect! Please take a look whenever you get a chance.

Thanks!

schaable:
Hey @SyedAsadKazmi, we're back from Devconnect and I'm finally able to follow up on this issue 😅

I think there are two separate parts to this:

**1. `hardhat-verify` using `production` as the default build profile**
This is intentional, since `production` is usually what people use when deploying and verifying contracts. That said, the error message here is definitely confusing. The issue happens because the contract passes our internal bytecode comparison (which only checks the executable section, excluding metadata), but fails Etherscan’s verification since they compare the full bytecode. I’ve opened a separate [issue](https://github.com/NomicFoundation/hardhat/issues/7730) so we can detect this case and improve the error message.

**2. `hardhat-verify` not retrying with the full compiler input**
Your fix is correct and solves the immediate problem, but it’s incomplete. We don’t want to retry in every failure scenario. For example, if the bytecode doesn’t match or constructor arguments are wrong, retrying with the full compiler input will almost certainly fail again.

According to https://docs.etherscan.io/api-reference/endpoint/checkverifystatus, there are specific errors where retry won't help. 
I'm going to approve this PR as it fixes the immediate problem, but I've also created an [issue](https://github.com/NomicFoundation/hardhat/issues/7731) to:
 - Add logic to distinguish between retryable and non-retryable errors.
 - Extend this behavior to the other verification providers (Blockscout and Sourcify).
 
Before merging, please add tests to `v-next/hardhat-verify/test/etherscan.ts`:
1. A test for detailed error messages (e.g., "Fail - Unable to verify. Compiled contract deployment bytecode does NOT match...") to ensure they’re treated as failures.
2. Verify that existing tests still pass with the updated polling logic.

Thanks again for the contribution, and your patience throughout this process!

schaable (approved):
Pre-approving, but needs tests (read my earlier comment).

SyedAsadKazmi:
Hey @schaable,

I hope you enjoyed Devconnect! 😄

Thanks for the approval!

I've added the tests and run the full test suite to confirm everything is working correctly. I'll also see if I can contribute toward resolving the issues (#7730 and #7731).

Thanks!

PR #7583, opened 2025-10-16
Set NODE_ENV for hardhat test tasks

### LINKED ISSUE #7511: hardhat test should set NODE_ENV to "test"
### Describe the feature

I expect `process.env.NODE_ENV` to be `"test"` when i run `pnpm hardhat test`

### Search terms

_No response_

michalbrabec:
The vitest, jest and other testing tools automatically set `NODE_ENV` to be `"test"` if it is not set when the tests are running. The variable is se only if it was not set before the tests started, this is the same for vitest and the other environments.
This enables the code under test and the tests themselves to react to the fact that it is running as part of a test

We agreed with Pato that all test runners should set the environment variable so their behaviour is consistent. Community test runners are free to set the variables as they see fit.

Documentation in https://github.com/NomicFoundation/hardhat-website/pull/103

Closes #7511

### DIFF
--- a/.changeset/purple-wasps-jump.md
+++ b/.changeset/purple-wasps-jump.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Validate initialBaseFeePerGas against hardfork only for L1 chain type
--- a/.changeset/shiny-pigs-give.md
+++ b/.changeset/shiny-pigs-give.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-errors": patch
+---
+
+Ported the `@nomicfoundation/hardhat-ledger` plugin to Hardhat 3 ([#5646](https://github.com/NomicFoundation/hardhat/issues/5646))
--- a/package.json
+++ b/package.json
@@ -18,5 +18,16 @@
     "prelint:fix": "pnpm build",
     "lint:fix": "pnpm run --recursive --if-present lint:fix",
     "version-for-release": "pnpm changeset version && pnpm -w install --lockfile-only"
+  },
+  "pnpm": {
+    "onlyBuiltDependencies": [
+      "node-hid"
+    ],
+    "ignoredBuiltDependencies": [
+      "esbuild",
+      "keccak",
+      "unrs-resolver",
+      "usb"
+    ]
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -71,6 +71,9 @@ importers:
       '@nomicfoundation/hardhat-keystore':
         specifier: workspace:^3.0.0
         version: link:../hardhat-keystore
+      '@nomicfoundation/hardhat-ledger':
+        specifier: workspace:^1.2.0
+        version: link:../hardhat-ledger
       '@nomicfoundation/hardhat-mocha':
         specifier: workspace:^3.0.0
         version: link:../hardhat-mocha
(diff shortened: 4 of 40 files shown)


v-next/hardhat-mocha/src/task-action.ts
```diff
@@ -54,6 +54,10 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
   // Set an environment variable that plugins can use to detect when a process is running tests
   process.env.HH_TEST = "true";
 
+  // Sets the NODE_ENV environment variable to "test" so the code can detect that tests are running
+  // This is done by other JS/TS test frameworks like vitest
+  process.env.NODE_ENV ??= "test";
```
ChristopherDedominici, line 59: Just to confirm: don't we want to always overwrite this? If not, this could break the tests logic. Are we ok with the "risk"?

v-next/hardhat-mocha/src/task-action.ts
```diff
@@ -54,6 +54,10 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
   // Set an environment variable that plugins can use to detect when a process is running tests
   process.env.HH_TEST = "true";
```
ChristopherDedominici, line 55: Should we replace HH_TEST and use NODE_ENV instead?

v-next/hardhat-mocha/test/env.ts
```diff
@@ -0,0 +1,30 @@
+import assert from "node:assert/strict";
+import { describe, it } from "node:test";
+
+import { useFixtureProject } from "@nomicfoundation/hardhat-test-utils";
+
+describe("Hardhat Mocha env variables", () => {
+  useFixtureProject("test-project");
+
+  it("should set the NODE_ENV variable if undefined and HH_TEST always", async () => {
```
ChristopherDedominici, line 9: In case we decide to not always overwrite `NODE_ENV`, I think that here we are missing the test that is present in the nodejs test runner:
```
it("should not set the NODE_ENV variable if defined before", async () => {
```

v-next/hardhat-mocha/src/task-action.ts
michalbrabec: This is a behaviour taken over from vitest, it is designed to set it only in case it is not set already so the users have a way to test production code without hacking the tests. They can either leave the var undefined and let the tests set it or set it to a value they need.

We do not use the `NODE_ENV` in any way and we do not plan to use it, it is purely for the user tests.

v-next/hardhat-mocha/src/task-action.ts
michalbrabec: I was considering it, but the two are not identical and we cannot rely on `NODE_ENV` since it can be overridden by users.

v-next/hardhat-mocha/test/env.ts
michalbrabec: The test exists in Node test runner, but it is missing in Mocha, because `harhat-mocha` does not allow multiple invocations of the task (there is a global var enforcing this [testsAlreadyRun](https://github.com/NomicFoundation/hardhat/blob/58fcc838e0b58e65f4aa4f0c4e87a26425a3771b/v-next/hardhat-mocha/src/task-action.ts#L49)). The only way I could find to avoid this lock was to split the test to a separate file, which felt wasteful considering the change, but I can add it.

ChristopherDedominici (approved):

PR #7584, opened 2025-10-16
Update release workflow to use trusted publishing

antico5:
The release workflow is now split into 5 jobs:
1. PR: Execute the changesets action. This will either create a PR bumping packages versions, or enable the rest of the workflow for publishing packages.
2. Pack: Build packages,execute a dry run, and pack them into tarballs to be uploaded as artifacts of the workflow run. The pack/publish steps are separated to minimize attack surface, since only the publish job runs in an environment whitelisted for publishing. Only packages with changes are affected (by leveraging dry-run output report).
3. Review: A diff is shown between proposed packages and latest published versions of those packages
4. Publish: This action is the only one that runs in an environment allowed to publish. Also at repository level this should be configured to require approvers. These should double check the output of the previous job (diff) before approving the release. The action itself downloads the previously generated tarballs and pushes them to npm.
5. Tag: Tags are created using changesets and pushed to the repo.

Preparation:
- Set up the github environment that will be used for publishing. Set up approvers and any other additional protection rule.
- Set up trusted publishing on npm.js both for the organization and the `hardhat` package. The workflow file name and the github environment id must be specified.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -167,7 +167,8 @@ jobs:
           cd tarballs
           for filename in *.tgz; do
             echo "Publishing $filename"
-            npm publish $filename --tag latest
+            # We can switch to pnpm publish when pnpm supports Trusted publishing
+            npm publish $filename --tag latest --access public
           done
 
   tag:


.github/workflows/release.yml
```diff
@@ -43,26 +44,155 @@ jobs:
...
+      - name: Set up the environment
+        uses: ./.github/actions/setup-env
+
+      - name: Install Dependencies
+        run: pnpm install --frozen-lockfile --prefer-offline
+
+      # This step is needed otherwise tag creation fails
+      - run: |
+          git config user.name  "github-actions"
+          git config user.email "github-actions@users.noreply.github.com"
```
kanej, line 195: What is it that allowed us to get away without setting them previously?

antico5: I'm not sure to be honest. This was the only way I could get it working with my set of monorepo + packages I used for  testing. Maybe we could do a test run on the hardhat CI without this and see if tagging works, and add it later otherwise. Worth mentioning, `changeset tag` will claim that tags are created even if they are not, and no git failure message is shown.

.github/workflows/release.yml
```diff
@@ -43,26 +44,155 @@ jobs:
...
+      contents: write
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v5
+
+      - name: Set up the environment
+        uses: ./.github/actions/setup-env
+
+      - name: Install Dependencies
+        run: pnpm install --frozen-lockfile --prefer-offline
```
alcuadrado: I think `--prefer-offline` could be removed from this workflow. It's a bit faster, because it removes some integrity validations. While I think it's a good trade off for most of our CI workflows, for the release one, i think it isn't.

.github/workflows/release.yml
```diff
@@ -43,26 +44,157 @@ jobs:
...
         env:
-          NODE_AUTH_TOKEN: ${{ secrets.NODE_AUTH_TOKEN }}
-          NPM_CONFIG_PROVENANCE: true
-        run: pnpm publish --filter "./v-next/**" -r --no-git-checks --access public
+          NPM_CONFIG_PROVENANCE: true # explicitly enable NPM provenance
+        run: |
+          cd tarballs
+          for filename in *.tgz; do
+            echo "Publishing $filename"
+            npm publish $filename --tag latest
```
kanej:
```suggestion
            # We can switch to pnpm publish when pnpm supports Trusted publishing
            npm publish $filename --tag latest --access public
```

Added `--access public` which helps for new package publishing.

.github/workflows/release.yml
```diff
@@ -43,26 +44,157 @@ jobs:
...
+      - name: Download tarballs
+        uses: actions/download-artifact@v5
+        with:
+          name: tarballs
+          path: tarballs
+
+      - name: Setup node to be able to update npm
+        uses: actions/setup-node@v5
+        with:
+          node-version: lts/*
```
kanej:
```suggestion
          node-version: 24
```
This is just to match the convention we use in the CI more generally. Our alternative would be to replace this step with: `./.github/actions/setup-env` and rely on the default node version.

.github/workflows/release.yml
kanej: I have removed all of the `--prefer-offline` flags.

PR #7590, opened 2025-10-19
fix: Guard v-next script validation against missing package scripts

strmfos:
This makes the validation script resilient when a package lacks a scripts section or its clean entry, so the tooling reports actionable errors instead of crashing. It’s important because a single incomplete package.json could previously halt the check and hide other issues across the workspace.


kanej:

Issue #7426, opened 2025-09-11
Implement the support for the programatic-only options


michalbrabec:
### Describe the issue

Implement the support for the options based on the design doc built for the feature

PR #7591, opened 2025-10-20
Add ability for task options to be hidden from the CLI

zoeyTM:
closes https://github.com/NomicFoundation/hardhat/issues/7426

### DIFF
--- a/v-next/hardhat/src/internal/cli/help/utils.ts
+++ b/v-next/hardhat/src/internal/cli/help/utils.ts
@@ -68,9 +68,7 @@ export function parseOptions(task: Task): {
   const options = [];
   const positionalArguments = [];
 
-  for (const [optionName, option] of task.options
-    .entries()
-    .filter(([, o]) => o.hidden !== true)) {
+  for (const [optionName, option] of task.options) {
     if (option.hidden === true) {
       continue;
     }


v-next/hardhat/src/internal/cli/help/utils.ts
```diff
@@ -69,6 +69,10 @@ export function parseOptions(task: Task): {
   const positionalArguments = [];
 
   for (const [optionName, option] of task.options) {
+    if (option.hidden === true) {
```
michalbrabec, line 72: Nit: `for (const [optionName, option] of task.options.filter(o => !o.hidden))` might be more elegant.

v-next/hardhat/src/types/tasks.ts
```diff
@@ -200,6 +200,7 @@ export interface NewTaskDefinitionBuilder<
     description?: string;
     type?: TypeT;
     defaultValue: ArgumentTypeToValueType<TypeT>;
+    hidden?: boolean;
```
fvictorio, line 203: We need to add this to the `addFlag` types too, right?

I wonder if we can have some type-level tests to ensure these types and the actual functions are in sync. It seems like it should be doable.

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -228,6 +228,22 @@ describe("main", function () {
       });
     });
 
+    describe("task with hidden option", function () {
+      useFixtureProject("cli/parsing/hidden-option");
+
+      it("should throw when passing a hidden option from the CLI", async function () {
```
fvictorio, line 234: Let's add a test for `addFlag` too.

v-next/hardhat/src/internal/core/global-options.ts
```diff
@@ -69,6 +69,17 @@ export function buildGlobalOptionDefinitions(
         }
       }
 
+      // @ts-expect-error -- validation against js users
```
fvictorio: This is a bit surprising. I'm not sure what we should do here.

On one hand, it makes sense to check this. It certainly doesn't hurt.

On the other hand, I don't think we do validation for js users anywhere else, do we? That, combined with `ts-expect-error` being something I think we should avoid unless absolutely necessary, makes me think we should just remove this check.

Not a super strong preference though, I wouldn't mind keeping it if there's a good reason.

v-next/hardhat/test/internal/core/tasks/builders.ts
```diff
@@ -380,6 +383,40 @@ describe("Task builders", () => {
...
+        const taskAction = async () => ({
+          default: () => {},
+        });
+        const taskDefinition = builder
+          .setAction(taskAction)
+          .addOption({
+            name: "arg",
+            defaultValue: "default",
+            hidden: true,
+          })
```
fvictorio, line 404: We should use `.addFlag` here too.

zoeyTM: Does the test on line 515 work, or did you mean something else?

fvictorio: I missed that, sorry. Yes, I think that works.

Btw, I was a bit confused at first because it looked like you were removing an existing test, but it turns out that test is duplicated for some reason?

And it seems like there's another duplication in this file: the tests in lines 1391 and 1420. I know that's unrelated to this PR, but (assuming I'm not wrong) we should delete one of those two. Or replace the second one with a test for the hidden option, like you did with the one in line 515, if it makes sense there too; up to you.

v-next/hardhat/src/internal/cli/help/utils.ts
zoeyTM: `options` is a Map not an array, so we could do something like this:

```ts
for (const [optionName, option] of task.options
    .entries()
    .filter(([, o]) => o.hidden !== true)) {...}
```

but at that point, that seems quite a bit less elegant to me

fvictorio: I think the code as it is is idiomatic and clear.

fvictorio: Hmm, now the code has both the `.entries().filter()` expression and the `if/continue` statements. One of those should be removed (I would remove the filter).

fvictorio:
Technically this needs docs, but I'm marking it as "no docs needed" because we don't have a tasks reference in the plugin authors docs yet, and I don't think we should mention it in a guide.

v-next/hardhat/src/internal/cli/help/utils.ts
zoeyTM: Whoops, accidentally committed it. I'll remove that.

Issue #7579, opened 2025-10-15
Running forked tests with hardhat keystore doesn't work with node test runner


GlitchicaL:
### Version of Hardhat

3.0.7

### What happened?

Running TypeScript tests against a forked network using Node.js test runner & Viem results in the prompt for Hardhat's keystore password to not hide the password, and seems to hang with no error:
 
<img width="920" height="356" alt="Image" src="https://github.com/user-attachments/assets/62dfc5a2-7dcd-4dbf-9adf-2861f2c13393" />

When testing using Mocha & Ethers.js, it seems to work as expected and doesn't hang:

<img width="930" height="327" alt="Image" src="https://github.com/user-attachments/assets/e200963d-66d4-49d5-840c-33819e60f741" />

Note both examples above were done with a new Hardhat project and only the *hardhat.config.ts* was edited with a remote network.

### Minimal reproduction steps

1. Create a new Hardhat project with `npx hardhat --init`.
2. Upon asking what type of project to initialize, select *"A TypeScript Hardhat project using Node Test Runner and Viem"*.
3. Install the necessary dependencies. For background, this is the command that is ran: 

```bash
npm install --save-dev "hardhat@^3.0.7" "@nomicfoundation/hardhat-toolbox-viem@^5.0.0" "@nomicfoundation/hardhat-ignition@^3.0.0" "@types/node@^22.8.5" "forge-std@foundry-rs/forge-std#v1.9.4" "typescript@~5.8.0" "viem@^2.30.0"
```

4. Edit *hardhat.config.ts* with a remote network:

```typescript
    hardhatMainnet: {
      type: "edr-simulated",
      chainType: "l1",
      forking: {
        url: configVariable("ALCHEMY_RPC_URL"),
        blockNumber: 23284000,
        enabled: true,
      }
    },
```

6. Run tests `npx hardhat test --network hardhatMainnet`

michalbrabec:
The error is exactly as described, public node can be used for testing as well https://ethereum-sepolia.publicnode.com/

michalbrabec:
Keystore is not supported in subprocesses like `node:test` and `mocha:parallel`.
Hardhat should display a clear error instructing users to use the dev keystore instead.

michalbrabec:
Will be resolved by #7583

PR #7592, opened 2025-10-20
Keystore Bug: dev keystore logic skipped if dev keystore is missing during tests

ChristopherDedominici:
Original issue: https://github.com/NomicFoundation/hardhat/issues/7579

CAUSE
During tests, if the development keystore does not exist, the logic incorrectly attempts to access the production keystore.
This should not happen, only the development keystore should be used during tests.
Instead, the code should fail with an error indicating that the key is missing and must be set in the development keystore.

This is the error that should be show during tests if a key is not set:
```
HardhatError: HHE50002: Key <keyName> not found in the development keystore. Run "npx hardhat keystore set <keyName> --dev" to set it.
```

### DIFF
--- a/.changeset/orange-pumpkins-watch.md
+++ b/.changeset/orange-pumpkins-watch.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/hardhat-keystore": patch
 ---
 
-Fixed a bug where the development keystore logic was skipped during tests if the development keystore was missing [7592](https://github.com/NomicFoundation/hardhat/pull/7592)
+Fixed a bug where the development keystore logic was skipped during tests if the development keystore was missing ([7592](https://github.com/NomicFoundation/hardhat/pull/7592))


v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts
```diff
@@ -64,6 +64,16 @@ export default async (): Promise<Partial<ConfigurationVariableHooks>> => {
     let keystoreLoader = isDevKeystore ? keystoreLoaderDev : keystoreLoaderProd;
     let masterKey = isDevKeystore ? masterKeyDev : masterKeyProd;
 
+    if (!isDevKeystore && onlyAllowDevKeystore) {
```
michalbrabec: I agree with the change, but I'm curious, were you able to detect why you got the `KEY_NOT_FOUND_DURING_TESTS_WITH_DEV_KEYSTORE` error while I reached the get password section prior to showing the error?

ChristopherDedominici: Yes, the issue was that the development store was not defined, so the code returned early ([see here](https://github.com/NomicFoundation/hardhat/blob/58fcc838e0b58e65f4aa4f0c4e87a26425a3771b/v-next/hardhat-keystore/src/internal/hook-handlers/configuration-variables.ts#L77)) instead of continuing with the function execution. However, it should have proceeded and thrown an error due to the missing key

michalbrabec: Confirmed - you can hit this error only if you do not have the entire dev keystore set up at all.
It does no longer repro once you set up any key in the dev keystore, as long as the keystore remains defined, even if empty.

.changeset/orange-pumpkins-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-keystore": patch
+---
+
+Fixed a bug where the development keystore logic was skipped during tests if the development keystore was missing [7592](https://github.com/NomicFoundation/hardhat/pull/7592)
```
kanej:
```suggestion
Fixed a bug where the development keystore logic was skipped during tests if the development keystore was missing ([7592](https://github.com/NomicFoundation/hardhat/pull/7592))
```

PR #7599, opened 2025-10-21
Solidity gas statistics

schaable:
This PR adds gas analytics reporting support for Solidity tests in Hardhat. Gas analytics was previously available for JavaScript/TypeScript tests, and this PR extends that functionality to Solidity tests by integrating the necessary glue code and refactoring the internal test event system.

Docs: https://github.com/NomicFoundation/hardhat-website/pull/108

### DIFF
--- a/.changeset/giant-radios-swim.md
+++ b/.changeset/giant-radios-swim.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Added support for showing gas statistics after running solidity tests ([#7503](https://github.com/NomicFoundation/hardhat/issues/7503)).


v-next/hardhat/src/internal/builtin-plugins/solidity-test/runner.ts
```diff
@@ -80,26 +81,38 @@ export function run(
...
             });
             remainingSuites.delete(
               formatArtifactId(suiteResult.id, sourceNameToUserSourceName),
             );
             if (remainingSuites.size === 0) {
               clearTimeout(timeout);
-              controller.close();
+
+              if (runCompleted) {
+                controller.close();
```
fvictorio, line 102: This code is slightly worrying because it means that a `suite:done` event could be emitted after `run:done` was emitted. I think this is not problematic right now, but it could be down the road.

No need to do anything about it though (except maybe adding a comment somewhere?).

schaable: I'll add a comment and also check with the EDR team. I was also surprised to see tests failing because of this.

PR #7600, opened 2025-10-21
Support Sourcify in `hardhat-verify`

### LINKED ISSUE #6885: Add sourcify support to `hardhat-verify`
Add sourcify support to `hardhat-verify`.

manuelwedler:

---

Closes #6885

Related documentation change: https://github.com/NomicFoundation/hardhat-website/pull/116

### Changes

- Adds the `Sourcify` class which implements the `VerificationProvider`
- Sourcify is enabled by default, and called from the main verify task
- Also adds the Sourcify subtask
- Adds an additional `--creation-tx-hash` option to the verify task, which can optionally be provided by the user to more easily get a creation match
- Sourcify URLs can be configured via the user config of the plugin. Note that Etherscan urls are configured via the global chain descriptors, but I think Sourcify doesn't fit into being added as an explorer.
- I had to change how the compiler input is passed to the `verify` function of the `VerificationProvider` interface, because the Sourcify API expects the non-stringified version of the compiler input.

### Example

Screenshot from 2025-10-21 15-10-00_0

### DIFF
--- a/.changeset/short-zoos-fold.md
+++ b/.changeset/short-zoos-fold.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/hardhat-verify": patch
 ---
 
-Added support for verifying contracts via Sourcify ([#6885](https://github.com/NomicFoundation/hardhat/issues/6885)).
+Added support for verifying contracts via Sourcify, thanks @manuelwedler ([#6885](https://github.com/NomicFoundation/hardhat/issues/6885)).


v-next/hardhat-verify/src/type-extensions.ts
```diff
@@ -23,13 +24,20 @@ declare module "hardhat/types/config" {
         enabled?: false;
       };
 
+  export interface SourcifyUserConfig {
+    repoUrl?: string;
```
schaable: For consistency with `chainDescriptors`, we should name this `url` instead.

v-next/hardhat-verify/src/internal/sourcify.ts
```diff
@@ -0,0 +1,376 @@
...
+      compilerVersion: string;
+      creationTransactionHash?: string;
+    } = {
+      stdJsonInput: compilerInput,
+      contractIdentifier: contractName,
+      compilerVersion,
+    };
+
+    if (creationTxHash !== undefined) {
+      body.creationTransactionHash = creationTxHash;
```
schaable, line 160: Can you expand on how `creationTransactionHash` is used and what its purpose is? Since it’s optional, I just want to confirm that we really need it before adding it.
I couldn't find this in your [docs](https://docs.sourcify.dev/docs/api/#/Verify%20Contracts/verify), so it would be helpful to understand the expected usage and whether we should document it.

v-next/hardhat-verify/src/internal/tasks/verify/types.ts
```diff
@@ -2,11 +2,12 @@ import type { LibraryAddresses } from "../../libraries.js";
 
 export interface VerifyActionArgs {
   address: string;
-  constructorArgs: string[];
+  constructorArgs?: string[];
```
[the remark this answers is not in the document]
schaable: Let's keep `constructorArgs` required for now, and Sourcify can just ignore them. We can update this later if users request it.

v-next/hardhat-utils/src/internal/request.ts
```diff
@@ -30,14 +30,14 @@ export async function generateTempFilePath(filePath: string): Promise<string> {
 
 export async function getBaseRequestOptions(
   requestUrl: string,
-  { extraHeaders, abortSignal, queryParams }: RequestOptions = {},
+  { extraHeaders, abortSignal, queryParams, throwOnError }: RequestOptions = {},
```
schaable: If I understood your code correctly, this shouldn't be necessary. The body of 4xx and 5xx responses can be accessed in the thrown error, for example:

```typescript
const url =
  "https://sourcify.dev/server/v2/contract/11155111/0xdadB0d80178819F2319190D340ce9A924f783711";

try {
  const response = await getRequest(url);
  // ...
} catch (error: any) {
  console.error("Caught error!");
  console.error("Error type:", error.constructor.name);
  console.error("Error message:", error.message);
  console.error("Error cause:", error.cause);
  console.error("Response body:", JSON.stringify(error.body, null, 2));
}
```

will output:
```
Caught error!
Error type: ResponseStatusCodeError
Error message: Received an unexpected status code from https://sourcify.dev
Error cause: ResponseStatusCodeError: Response status code 404: Not Found
    at /workspaces/hardhat/node_modules/.pnpm/undici@6.21.3/node_modules/undici/lib/api/util.js:53:33
    at node:internal/process/task_queues:151:7
    at AsyncResource.runInAsyncScope (node:async_hooks:214:14)
    at AsyncResource.runMicrotask (node:internal/process/task_queues:148:8)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
  code: 'UND_ERR_RESPONSE_STATUS_CODE',
  body: {
    match: null,
    creationMatch: null,
    runtimeMatch: null,
    chainId: '11155111',
    address: '0xdadB0d80178819F2319190D340ce9A924f783711'
  },
  status: 404,
  statusCode: 404,
  headers: {
    'x-powered-by': 'Express',
    'access-control-allow-origin': '*',
    'content-type': 'application/json; charset=utf-8',
    etag: 'W/"83-Hbj8EZFj77oB2gS6tHOdfSdpeB0"',
    'x-cloud-trace-context': 'f510ffd0b064d06a2eeebe4892aa9885',
    date: 'Tue, 28 Oct 2025 15:51:20 GMT',
    server: 'Google Frontend',
    'content-length': '131',
    via: '1.1 google',
    'alt-svc': 'h3=":443"; ma=2592000,h3-29=":443"; ma=2592000'
  }
}
Response body: {
  "match": null,
  "creationMatch": null,
  "runtimeMatch": null,
  "chainId": "11155111",
  "address": "0xdadB0d80178819F2319190D340ce9A924f783711"
}
```

Let me know if I'm missing something or if there's a different reason this needs to be changed!

v-next/hardhat-verify/src/internal/sourcify.ts
```diff
@@ -0,0 +1,376 @@
...
+  shouldUseProxy,
+} from "@nomicfoundation/hardhat-utils/request";
+
+export const SOURCIFY_PROVIDER_NAME: keyof VerificationProvidersConfig =
+  "sourcify";
+
+const VERIFICATION_STATUS_POLLING_SECONDS = 3;
+
+export const SOURCIFY_REPO_URL = "https://repo.sourcify.dev";
+export const SOURCIFY_API_URL = "https://sourcify.dev/server";
```
schaable, line 36: Are the staging and local urls commonly used? How often do users use a different repo or api url?

v-next/hardhat-verify/src/type-extensions.ts
```diff
@@ -40,4 +48,10 @@ declare module "hardhat/types/config" {
     apiKey: ResolvedConfigurationVariable;
     enabled: boolean;
   }
+
+  export interface SourcifyConfig {
+    repoUrl: string | undefined;
+    apiUrl: string | undefined;
```
schaable: For consistency with existing types:

```suggestion
    repoUrl?: string;
    apiUrl?: string;
```

v-next/hardhat-verify/src/internal/sourcify.ts
manuelwedler: It's mainly important to have these configurable to have support for custom Sourcify instances. A couple of chains run their own Sourcify server. For example there is: https://sourcify-api-monad.blockvision.org/health or https://verify-api.vechainstats.com/health

So any users wanting to verify on a custom Sourcify instance would use a different URL.

v-next/hardhat-verify/src/internal/tasks/verify/types.ts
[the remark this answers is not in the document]
manuelwedler: sounds good

v-next/hardhat-utils/src/internal/request.ts
manuelwedler: Okay, I see. I can make the logic work without it.

v-next/hardhat-verify/src/internal/sourcify.ts
manuelwedler: The creation transaction hash is needed to get creation bytecode from the rpc. If it is not provided by the user, Sourcify relies on external services like Etherscan and Blockscout to get the creation tx hash. If these fail, Sourcify isn't able to get a creation code match. Explorers don't need this option since they index the chain and have the creation bytecode available anyway.

I'd say it's not super important to have on the cli. It just makes the verification process more reliable and a little bit quicker. 

There is a reason to have it at least on the Typescript interface. We didn't ask about that yet, but if you would add Sourcify support to `hardhat ignition deploy --verify`, it would make a lot of sense to pass the creation transaction hash, since you should have it available anyway.

Btw, thanks for pointing out that it's not really documented. We will update the docs.

v-next/hardhat-utils/src/internal/request.ts
manuelwedler: @schaable I'm getting caught on this one currently. I moved the logic inside the error handlers, tested it manually, and it seems to work fine. But the tests are failing. The issue is that the error object passed to the error handlers does not contain a body. It looks like this:

```
ResponseStatusCodeError: Received an unexpected status code from http://localhost
    at handleError (/home/manuel/Projects/hardhat/hardhat/v-next/hardhat-utils/dist/src/internal/request.js:98:15)
    at getRequest (/home/manuel/Projects/hardhat/hardhat/v-next/hardhat-utils/dist/src/request.js:35:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Sourcify.isVerified (/home/manuel/Projects/hardhat/hardhat/v-next/hardhat-verify/src/internal/sourcify.ts:83:18)
    at async TestContext.<anonymous> (/home/manuel/Projects/hardhat/hardhat/v-next/hardhat-verify/test/sourcify.ts:199:28)
    at async Test.run (node:internal/test_runner/test:1054:7)
    at async Suite.processPendingSubtests (node:internal/test_runner/test:744:7) {
  statusCode: 404,
  headers: {},
  body: null,
  [cause]: ResponseStatusCodeError: Response status code 404: Not Found
      at /home/manuel/Projects/hardhat/hardhat/node_modules/.pnpm/undici@6.21.3/node_modules/undici/lib/api/util.js:34:35
      at node:internal/process/task_queues:151:7
      at AsyncResource.runInAsyncScope (node:async_hooks:214:14)
      at AsyncResource.runMicrotask (node:internal/process/task_queues:148:8)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5) {
    code: 'UND_ERR_RESPONSE_STATUS_CODE',
    body: undefined,
    status: 404,
    statusCode: 404,
    headers: {}
  }
}
```

To me it looks like there could be something wrong with the interceptor or the parsing of the response in the test setup. How am I supposed to pass an error body via the interceptor? I cannot find another test where a json body is passed via the interceptor.

The change is in the last commit.

schaable: I think you need to specify the response body in the interceptor. Once everything else is ready, push the code and I’ll take a look at the tests.

manuelwedler: Thanks. The pushed commits should already solve everything else.

manuelwedler:
Btw, I added two additional changes:

With a recent update of the Sourcify server, it is now possible to reach the repo ui via redirect from the API. Therefore, I added a change to remove the repo URL from the configuration. This reduces the configuration overhead a bit.
If you disagree with the change, we can move back to have two configuration variables - `url` and `apiUrl`.

I also added a commit to show a link to the job info page of the Sourcify UI in case the verification fails. The page provides more info to the user for debugging why the verification fails. See for example: https://verify.sourcify.dev/jobs/75d1a316-dd30-4fb1-a889-e6427e3d3211

schaable:
Thanks, @manuelwedler! I’ll do another pass to fix any minor details I find along with the tests. After that, we’re good to go.

Regarding:

> With a recent update of the Sourcify server, it is now possible to reach the repo ui via redirect from the API. Therefore, I added a change to remove the repo URL from the configuration. This reduces the configuration overhead a bit.
> If you disagree with the change, we can move back to have two configuration variables - url and apiUrl.

Is this also available for custom Sourcify instances like https://sourcify-api-monad.blockvision.org/health
 or https://verify-api.vechainstats.com/health ?

alcuadrado:
I updated the issue description to point to the new website's PR.

manuelwedler:
> Is this also available for custom Sourcify instances like https://sourcify-api-monad.blockvision.org/health
> or https://verify-api.vechainstats.com/health ?

Not yet I suppose. We are currently in contact with all custom instances we know of because some still only run the legacy API. Eventually, they should have these endpoints. The redirects are implemented inside the server code, such that it should only be a matter of upgrading to the newest server package/image.

manuelwedler:
Thanks for finalizing the PR!

PR #7608, opened 2025-10-22
[HHV2] handle invalid Solidity cache JSON

### LINKED ISSUE #6489: Remove `solidity-files-cache.json` file if it's broken
Reported by @shark0der.

If the `cache/solidity-files-cache.json` file is broken (in his case it was because it was empty, which is an invalid JSON), then compilation fails.

When reading that file, we should detect if it's broken and remove it, which I think can always be done without issues.

iosh:

---

Ignore parsing errors in solidity-files-cache.json so compilation no longer stalls; fall  back to the default cache whenever the file is invalid.
resolve #6489

### DIFF
--- a/.changeset/stupid-lions-know.md
+++ b/.changeset/stupid-lions-know.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Ignore malformed solidity-files-cache so compilation keeps running while logging the issue in debug.
+Ignore malformed solidity-files-cache so compilation keeps running while logging the issue in debug ([#6489](https://github.com/NomicFoundation/hardhat/issues/6489))


packages/hardhat-core/src/builtin-tasks/utils/solidity-files-cache.ts
```diff
@@ -53,12 +53,17 @@ export class SolidityFilesCache {
...
   ): Promise<SolidityFilesCache> {
-    let cacheRaw: Cache = {
+    const defaultCache: Cache = {
       _format: FORMAT_VERSION,
       files: {},
     };
+    let cacheRaw = defaultCache;
     if (await fsExtra.pathExists(solidityFilesCachePath)) {
-      cacheRaw = await fsExtra.readJson(solidityFilesCachePath);
+      try {
```
michalbrabec, line 63: `fsExtra.readJson` can throw multiple different types of errors (file not found, path is a directory, syntax error ...), we should not flatly absorb all of them and quietly ignore them.
I would propose:
- add a check to the `if` that `solidityFilesCachePath` is a file
-  catch all exceptions
    - do `cacheRaw = defaultCache;`
    - report any error to the terminal, except the syntax error (broken JSON or empty file)
      > not rethrow, just print the message

packages/hardhat-core/src/builtin-tasks/utils/solidity-files-cache.ts
```diff
@@ -53,12 +53,34 @@ export class SolidityFilesCache {
...
+
+        if (!cacheStats.isFile()) {
+          console.warn(
+            `Hardhat couldn't read the solidity files cache at ${solidityFilesCachePath} because it's not a file. Please remove it.`
+          );
+        } else {
+          cacheRaw = await fsExtra.readJson(solidityFilesCachePath);
+        }
+      } catch (error) {
+        cacheRaw = defaultCache;
```
michalbrabec: Nit: this is not necessary since the default cache is already assigned on the line 60 and any failure in `fsExtra.readJson` prevents assignment.

packages/hardhat-core/src/builtin-tasks/utils/solidity-files-cache.ts
```diff
@@ -53,12 +53,34 @@ export class SolidityFilesCache {
...
+        if (!(error instanceof SyntaxError)) {
+          const errorMessage =
+            error instanceof Error ? error.message : String(error);
+
+          console.warn(
+            `Hardhat couldn't read the solidity files cache at ${solidityFilesCachePath}: ${errorMessage}`
```
michalbrabec: Nit: Could be shorten as `Hardhat couldn't read the solidity files cache at ${solidityFilesCachePath}: ${error.message ?? String(error)}`. The `errorMessage ` could be then removed. Tested with:
```typescript
try {
  throw Error("test");
} catch (error: any) {
  console.log(error.message ?? String(error))
}
```

packages/hardhat-core/src/builtin-tasks/utils/solidity-files-cache.ts
```diff
@@ -53,12 +53,34 @@ export class SolidityFilesCache {
...
       files: {},
     };
+    let cacheRaw = defaultCache;
+
     if (await fsExtra.pathExists(solidityFilesCachePath)) {
-      cacheRaw = await fsExtra.readJson(solidityFilesCachePath);
+      try {
+        const cacheStats = await fsExtra.stat(solidityFilesCachePath);
+
+        if (!cacheStats.isFile()) {
```
michalbrabec: Nit: Could be shortened as `if ((await fsExtra.stat(solidityFilesCachePath)).isFile()) {`

michalbrabec (approved):

packages/hardhat-core/src/builtin-tasks/utils/solidity-files-cache.ts
```diff
@@ -53,12 +54,34 @@ export class SolidityFilesCache {
...
     };
+    let cacheRaw = defaultCache;
+
     if (await fsExtra.pathExists(solidityFilesCachePath)) {
-      cacheRaw = await fsExtra.readJson(solidityFilesCachePath);
+      try {
+        if ((await fsExtra.stat(solidityFilesCachePath)).isFile()) {
+          cacheRaw = await fsExtra.readJson(solidityFilesCachePath);
+        } else {
+          console.error(
```
michalbrabec: Delete the directory, without throwing errors, in the catch block.

packages/hardhat-core/src/builtin-tasks/utils/solidity-files-cache.ts
```diff
@@ -53,12 +54,34 @@ export class SolidityFilesCache {
...
+            picocolors.red(
+              `Hardhat couldn't read the solidity files cache at ${solidityFilesCachePath} because it's not a file. Please remove it.`
+            )
+          );
+        }
+      } catch (error) {
+        if (!(error instanceof SyntaxError)) {
```
michalbrabec: Do not show any error messages, print a debug message instead, for all errors (no if needed).
Debug message for inspiration [config-loading.ts](https://github.com/NomicFoundation/hardhat/blob/ab394ff047caa23a11af07aa886023d788deb3e7/v-next/hardhat/src/internal/config-loading.ts#L41)
Delete the file if it is malformed (parsing fails) or directory (since it is a directory and not a cache file).

.changeset/stupid-lions-know.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Ignore malformed solidity-files-cache so compilation keeps running while logging the issue in debug.
```
kanej:
```suggestion
Ignore malformed solidity-files-cache so compilation keeps running while logging the issue in debug ([#6489](https://github.com/NomicFoundation/hardhat/issues/6489))
```

kanej (approved):

PR #7610, opened 2025-10-22
Remove unused category for `hh-ledger` from `hh-errors`

ChristopherDedominici:
Fix for an issue that also impact `hardhat-website`: https://github.com/NomicFoundation/hardhat-website/pull/105#issuecomment-3431561468

### DIFF
--- a/.changeset/young-penguins-think.md
+++ b/.changeset/young-penguins-think.md
@@ -1,5 +1,0 @@
----
-"@nomicfoundation/hardhat-errors": patch
----
-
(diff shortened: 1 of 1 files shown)


fvictorio:
> I wouldn't include a changeset here because this doesn't need to be released (the website clones the repos). But I don't mind if it's merged as is.

Mea culpa: we should've included a changeset because the `prepare-error-list.ts` script in the `hardhat-website` repo downloads this from unpkg (that is, from the published package), it doesn't clone the repo :man_facepalming:

I'll just fix this on the script for now.

PR #7626, opened 2025-10-27
feat: don't compile solidity test contracts from js/ts tests

kanej:
Previously both core contracts and tests were compiled for the mocha/nodejs tests. Now only the core contracts are compiled as these are the only contracts that are needed for JS/TS tests.

Both js/ts tasks now don't compile tests. Note: the `npx hardhat test` overrarching task no longer compiles Solidity tests either - this is handled by the Solidity test task - it basically always compiles the tests.

## Manual testing

To check that this works I ran manual tests against the example project. Specifically running `npx hardhat clean` then:

* `npx hardhat test`
* `npx hardhat test mocha`
* `npx hardhat test nodejs`
* `npx hardhat test solidity`

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
@@ -35,7 +35,9 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
   const thisTask = hre.tasks.getTask("test");
 
   if (!noCompile) {
-    await hre.tasks.getTask("compile").run({});
+    await hre.tasks.getTask("compile").run({
+      noTests: true,
+    });
     console.log();
   }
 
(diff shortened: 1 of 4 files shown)


v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -35,7 +35,9 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
   const thisTask = hre.tasks.getTask("test");
 
   if (!noCompile) {
-    await hre.tasks.getTask("compile").run({});
```
kanej, line 38: Note: this should be safe (based on manual testing). The `npx hardhat test solidity` task always runs compile on test contracts.

fvictorio (approved):

PR #7627, opened 2025-10-27
Require peer dependency review on release PR

antico5:
This PR introduces a workflow that checks for the presence of the "peer dependencies reviewed" label on the release PR. Since that PR is created by a workflow itself, it can't trigger other ones. For this, a step was added after the changesets/action to mark the PR as draft. Then, when an user manually marks it for ready for review, the label check workflow is ran.

The goal of this workflow is to ensure the person doing the release manually checks for peer dependency version range bumps.

### DIFF
--- a/.github/workflows/ensure-peer-deps-review.yml
+++ b/.github/workflows/ensure-peer-deps-review.yml
@@ -5,8 +5,12 @@ name: Check peer dependency versions have been reviewed
 
 on:
   pull_request:
+    branches:
+      - changeset-release/main
     types:
+      - opened
       - synchronize
+      - reopened
       - labeled
       - unlabeled
       - edited
@@ -14,7 +18,6 @@ on:
 
 jobs:
   check-label:
-    if: github.head_ref == 'changeset-release/main'
     name: Check peer dependency versions have been reviewed
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/ensure-peer-deps-review.yml
```diff
@@ -0,0 +1,27 @@
...
+    types:
+      - synchronize
+      - labeled
+      - unlabeled
+      - edited
+      - ready_for_review
+
+jobs:
+  check-label:
+    if: github.head_ref == 'changeset-release/main'
```
kanej: What is the difference between this check and stating the branch in the `on` section?

.github/workflows/ensure-peer-deps-review.yml
```diff
@@ -0,0 +1,27 @@
...
+name: Check peer dependency versions have been reviewed
+
+on:
+  pull_request:
+    types:
+      - synchronize
+      - labeled
+      - unlabeled
+      - edited
+      - ready_for_review
```
kanej, line 17: Will this set the PR back to reviewed if any edit happens?

.github/workflows/ensure-peer-deps-review.yml
michalbrabec: There should be no difference, moved the check to the `on` section as the other workflows.

.github/workflows/ensure-peer-deps-review.yml
michalbrabec: This should only specify when the workflow runs and for this one it should basically mean "run this thing when the PR changes from draft to a full PR"

PR #7630, opened 2025-10-28
chore: v2 trusted publishing based release process

kanej:
Add the trusted publishing release process to Hardhat 2.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,11 @@ on:
     branches:
       - v2
 
+env:
+  # This is the tag the release will use 
+  # (and the review step will use for comparisons)
+  NPM_TAG: hh2
+
 jobs:
   release:
     name: Release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -30,3 +30,239 @@ jobs:
...
+          cd tarballs
+          jq -r '.publishedPackages[] | "\(.name)|\(.version)"' ../pnpm-publish-summary.json |
+          while IFS='|' read -r name version; do
+            NORMALIZED_NAME="${name//\//-}"
+            NORMALIZED_NAME="${NORMALIZED_NAME//@/}"
+            filename=$NORMALIZED_NAME-$version.tgz
+            
+            echo "Publishing $filename"
+            # We can switch to pnpm publish when pnpm supports Trusted publishing
+            npm publish $filename --tag v2 --access public
```
kanej: Note the `v2` flag rather than latest.

.github/workflows/release.yml
```diff
@@ -30,3 +30,239 @@ jobs:
...
+      - name: Upload packed tarballs
+        uses: actions/upload-artifact@v4
+        if: steps.detect.outputs.hasPackages == 'true'
+        with:
+          name: tarballs
+          path: tarballs
+          overwrite: true
+          if-no-files-found: error
+
+  review:
```
alcuadrado, line 154: I'm pretty sure this is incorrect, as it seems to be comparing with latest.

In the example repo i created the tag is a variable, so that would make it simpler to keep in sync.

kanej: Variable added and comparison updated.

PR #7641, opened 2025-10-30
fix(ignition): Pass global config to Ignition helper implementations

### LINKED ISSUE #7640: hre.config.ignition settings not propagated to ignition.deploy() when called from scripts
### Version of Hardhat

3.0.9

### What happened?

When calling ignition.deploy() from a Hardhat script or task, the global hre.config.ignition configuration (like requiredConfirmations, blockPollingInterval, etc.) is not being used . Instead, the deployment falls back to default values. [code](https://github.com/NomicFoundation/hardhat/blob/3a397fd0ad6c3bc2c2a321f8d57c01f9b7f48af4/v-next/hardhat-ignition-core/src/internal/defaultConfig.ts#L6-L12)

**Expected behavior:** Setting requiredConfirmations: 1 in hardhat.config.ts should apply to all deployments via ignition.deploy().

**Actual behavior:** The deployment uses the default value of 5 confirmations, causing IGN403: WAITING_FOR_CONFIRMATIONS errors on networks with lazy mining.

**Root cause:** The Ignition helper implementations (EthersIgnitionHelperImpl and ViemIgnitionHelperImpl) are instantiated without passing the global config. The constructor accepts an optional 4th parameter config?: Partial<DeployConfig> , but it's never passed, leaving this.#config as undefined. When the deploy method merges configs, only the per-deployment config is used .

1. Create a hardhat.config.ts with:
```
export default {  
  ignition: {  
    requiredConfirmations: 1,  
  },  
};
```
2. Create a deployment script:
`const result = await hre.ignition.deploy(moduleDefinition);`

3. Add logging to verify config:
`console.log('HRE config:', hre.config.ignition.requiredConfirmations); // Shows: 1`

4. Run the deployment and observe the error message still references "5 confirmations"

### Search terms

hardhat ignition requiredConfirmations not working, IGN403 WAITING_FOR_CONFIRMATIONS, hre.config.ignition ignored, hardhat ignition config not applied, lazy mining confirmations error

siljapetasch:
## Summary  
  
Fixes a bug where global `hre.config.ignition` settings (like `requiredConfirmations`, `blockPollingInterval`, etc.) are not propagated to `ignition.deploy()` when called from scripts.  
  
## Problem  
  
The Ignition helper implementations (`EthersIgnitionHelperImpl` and `ViemIgnitionHelperImpl`) are instantiated without passing the global configuration from `hre.config.ignition`. This causes deployments to fall back to default values instead of respecting user configuration.  
  
For example, setting `requiredConfirmations: 1` in `hardhat.config.ts` has no effect, and deployments use the default value of `5`, causing `IGN403: WAITING_FOR_CONFIRMATIONS` errors on networks with lazy mining.  
  
## Root Cause  
  
The network hook handlers instantiate the helper implementations with only 3 arguments, omitting the optional 4th parameter `config?: Partial<DeployConfig>`:  
  
- `v-next/hardhat-ignition-ethers/src/internal/hook-handlers/network.ts:24-27`  
- `v-next/hardhat-ignition-viem/src/internal/hook-handlers/network.ts:24-27`  
  
This leaves `this.#config` as `undefined` in the helper instances, so when the `deploy()` method merges configurations, only per-deployment config and defaults are used.  
  
## Solution  
  
Pass `context.config.ignition` as the 4th parameter when instantiating both helper implementations. This ensures the global configuration is properly merged with per-deployment config, establishing the correct precedence:  
  
`defaultConfig` → `hre.config.ignition` → per-deployment config  
  
## Changes  
  
- Modified `v-next/hardhat-ignition-ethers/src/internal/hook-handlers/network.ts` to pass `context.config.ignition` to `EthersIgnitionHelperImpl` constructor  
- Modified `v-next/hardhat-ignition-viem/src/internal/hook-handlers/network.ts` to pass `context.config.ignition` to `ViemIgnitionHelperImpl` constructor  
  
## Testing  
  
Verified that:  
1. Global `hre.config.ignition` settings are now respected in `ignition.deploy()` calls  
2. Per-deployment config still takes precedence over global config  
3. The `deploy` task continues to work correctly (it was already working)  
  
## Related Issues  
  
Fixes #7640

### DIFF
--- a/.changeset/hungry-lies-look.md
+++ b/.changeset/hungry-lies-look.md
@@ -1,0 +1,7 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+"@nomicfoundation/hardhat-utils": patch
+"hardhat": patch
+---
+
+Fixed Etherscan verification failures by removing hardcoded v1 API URLs from chain descriptors ([#7623](https://github.com/NomicFoundation/hardhat/issues/7623)). Also enhanced config resolution to support partial overrides in block explorer configurations for future extensibility.
--- a/v-next/hardhat-utils/src/internal/lang.ts
+++ b/v-next/hardhat-utils/src/internal/lang.ts
@@ -13,36 +13,41 @@ export async function getDeepCloneFunction(): Promise<<T>(input: T) => T> {
   return clone;
 }
 
-export function deepMergeImpl<T extends object, U extends object>(
+export function deepMergeImpl<T extends object, S extends object>(
   target: T,
-  source: U,
-): T & U {
+  source: S,
+  shouldOverwriteUndefined: boolean,
+): T & S {
   /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-  -- The result is expected to include properties from both target and source,
-  but initially only target is spread in, so a cast is needed. */
-  const result = { ...target } as T & U;
+  -- Result will include properties from both T and S, but starts with only T */
+  const result = { ...target } as T & S;
 
-  /*  eslint-disable-next-line @typescript-eslint/consistent-type-assertions
-  -- TypeScript cannot infer the correct union of string and symbol keys, but all keys come from U */
+  /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  -- All keys come from S, TypeScript can't infer the union of string and symbol keys */
   const keys = [
     ...Object.keys(source),
     ...Object.getOwnPropertySymbols(source),
-  ] as Array<keyof U>;
+  ] as Array<keyof S>;
 
   for (const key of keys) {
     if (
       isObject(source[key]) &&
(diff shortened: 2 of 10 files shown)


v-next/hardhat-ignition-ethers/src/internal/hook-handlers/network.ts
```diff
@@ -25,6 +25,7 @@ export default async (): Promise<Partial<NetworkHooks>> => {
         context.config,
         context.artifacts,
         connection,
+        context.config.ignition,
```
michalbrabec, line 28: @siljapetasch Can you please add a changeset outlining the change and update the tests to verify that the config gets correctly propagated?

zoeyTM (approved):

michalbrabec (approved):

PR #7655, opened 2025-11-02
Fix package manager detection

alcuadrado:
I noticed that I was getting the wrong package manager during the initialization of HH3, so I explored the situation.

I run all the possible combinations of package managers, and noticed that all of them provide the necessary information in the same env variable, so I fixed it.

I also added Bun and Deno as a best effort, as it was fairly cheap to do it.

### DIFF
--- a/v-next/hardhat/src/internal/cli/init/package-manager.ts
+++ b/v-next/hardhat/src/internal/cli/init/package-manager.ts
@@ -55,7 +55,7 @@ type PackageManager = "npm" | "yarn" | "pnpm" | "bun" | "deno";
  *
  * deno:
  *
- *   uninstall: deno run -A npm:print-environment
+ *   uninstalled: deno run -A npm:print-environment
  *     "deno/2.5.6 npm/? deno/2.5.6 linux aarch64"
  *
  *   installed: deno init && deno add npm:print-environment && deno --allow-env print-environment


v-next/hardhat/test/internal/cli/init/package-manager.ts
```diff
@@ -1,50 +1,59 @@
...
 
   afterEach(() => {
     process.env.npm_config_user_agent = originalUserAgent;
   });
 
-  it("should return pnpm if pnpm-lock.yaml exists", async () => {
-    await writeUtf8File("pnpm-lock.yaml", "");
-    assert.equal(await getPackageManager(process.cwd()), "pnpm");
-  });
+  const fixtureValues = [
```
michalbrabec, line 20: Do we have end-to-end tests to alert us in case any of the package managers change this behaviour?
They could change the the variable, format of the message or stop reporting.

v-next/hardhat/src/internal/cli/init/package-manager.ts
```diff
@@ -1,35 +1,98 @@
...
   }
-  if ((await exists(pathToPnpmLock)) || invokedFromPnpm) {
-    return "pnpm";
+
+  const firstSlashIndex = userAgent.indexOf("/");
+  if (firstSlashIndex === -1) {
+    return DEFAULT;
+  }
+
+  const packageManager = userAgent.substring(0, firstSlashIndex);
```
michalbrabec, line 80: Do the package managers provide any guarantees on the message they use?

v-next/hardhat/test/internal/cli/init/package-manager.ts
alcuadrado: No, but I think the chances of that changing are remote, as they would break the entire ecosystem

v-next/hardhat/src/internal/cli/init/package-manager.ts
alcuadrado: Same as above

PR #7656, opened 2025-11-03
exclude Core and its testnet from zero fee config

gr-akshaya:
This PR replicates the change introduced in [NomicFoundation/hardhat-ignition#755](https://github.com/NomicFoundation/hardhat-ignition/pull/755), where the BNB network was excluded from the baseFee configuration.
Similarly, the Core network requires exclusion from the baseFee logic to support successful deployments via Hardhat Ignition.

Context

- An earlier [PR for Hardhat Ignition](https://github.com/NomicFoundation/hardhat-ignition/pull/866) implementing the same fix for Core was closed before merging.

- Mentioning, [NomicFoundation/hardhat#6857](https://github.com/NomicFoundation/hardhat/pull/6857) introduced support for maxFeePerGas configuration in Ignition.

- However, the corresponding changes do not appear to be included in the current Hardhat v3 release, which results in failed deployments for Core due to incorrect baseFee handling.

Impact

- Without this fix, contract deployments on the Core network via Hardhat Ignition fail during gas estimation due to incompatible baseFee logic.

Request

- Please review and merge this PR to restore deployment compatibility for the Core network.
- If there’s an alternative or preferred approach to handle this configuration, kindly share feedback so we can align and update accordingly.

### DIFF
--- a/v-next/hardhat-ignition-core/src/internal/execution/jsonrpc-client.ts
+++ b/v-next/hardhat-ignition-core/src/internal/execution/jsonrpc-client.ts
@@ -670,6 +670,7 @@ export class EIP1193JsonRpcClient implements JsonRpcClient {
       // of blockchains using Besu. We explicitly exclude BNB
       // Smartchain (chainId 56) and its testnet (chainId 97)
       // as well as opBNB (chainId 204) and its testnet (chainId 5611)
+      // as well as Core (chainId 1116) and its testnet (chainId 1114)
       // from this logic as it is EIP-1559 compliant but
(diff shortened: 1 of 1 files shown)


zoeyTM:
Hi @gr-akshaya! 

You are absolutely correct that this fee configuration was accidentally lost during our update to v3, thanks for alerting us to that! I've opened a [new PR](https://github.com/NomicFoundation/hardhat/pull/7666) that restores this functionality for v3.

I'm going to close this PR in favor of the one that restores the full feature, so as soon as that is merged and released you will be able to use it again exactly like in v2. 

Thanks!

Issue #7609, opened 2025-10-22
unable to deploy upgradeable smart contract


Shahzaibnm:
### Version of Hardhat

3.0.8

### What happened?

i was trying to deploy my smart contract using OpenZeppelin's packages, but the issue im facing is that, when i deploy the code it throws the error build is not found, and compilation  gets done fine, creates the artifacts and build, but in cache it does not.

error is mentioned below:

$ npx hardhat ignition deploy ignition/modules/proxyModule.ts
An unexpected error occurred:

FileNotFoundError: File D:\hardhat-example\cache\build-info\solc-0_8_29-9fd9249cc37933dfb6bb6367749729dc924579fc.output.json not found
    at move (D:\hardhat-example\node_modules\@nomicfoundation\hardhat-utils\src\fs.ts:695:15)
    at async SolidityBuildSystemImplementation.emitArtifacts (D:\hardhat-example\node_modules\hardhat\src\internal\builtin-plugins\solidity\build-system\build-system.ts:806:5)

### Minimal reproduction steps

use this hardhat version and try upgradeable contract deploying .

michalbrabec:
@Shahzaibnm are you including the NPM-acquired contracts in the `config.solidity.npmFilesToBuild` array?

I am able to compile and deploy an ERC20-based token with the following code and config

```typescript
import type { HardhatUserConfig } from "hardhat/config";
import hardhatToolboxViemPlugin from "@nomicfoundation/hardhat-toolbox-viem";

const config: HardhatUserConfig = {
  plugins: [hardhatToolboxViemPlugin],
  solidity: {
    profiles: {
      default: {
        version: "0.8.28",
      },
    },
    npmFilesToBuild: ["@openzeppelin/contracts/token/ERC20/ERC20.sol"],
  },
  networks: {
    hardhat: {
      type: "edr-simulated",
      chainType: "l1",
    },
  },
};
export default config;
```

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.28;

import "@openzeppelin/contracts/token/ERC20/ERC20.sol";

contract MyToken is ERC20 {
    constructor(string memory name, string memory symbol) ERC20(name, symbol) {
        _mint(msg.sender, 1_000_000 ether);
    }
}
```

Shahzaibnm:
yes i did add
    npmFilesToBuild: [
      "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol",
[...]

PR #7660, opened 2025-11-04
Fix bug when `isolated` property is true in build profile

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/7609

Attention: this is not the final code, but a temporary solution to validate that the fix works. The code will be cleaned up and improved once the solution is confirmed to be working.


alcuadrado:
Replaced by https://github.com/NomicFoundation/hardhat/pull/7667

PR #7668, opened 2025-11-05
Make the `hardhat-ignition` `PrettyEventHandler` class work well with User Interruptions

alcuadrado:
This PR makes `PrettyEventHandler`'s handlers uninterruptible by user interruptions.


kanej (approved):
We have agreed to add this improvement. But we will do a further round of refinement to get the ignition/ledger UI integration working well.

PR #7670, opened 2025-11-06
feat: implement maxBatchSize support for Ignition

### LINKED ISSUE #7648: Implement support for `maxBatchSize`
### Describe the feature

Add the `maxBatchSize` property to the Ignition config, which sets maximum size of the batch in Ignition.

The batches are currently created based on the dependencies among the transactions. With the `maxBatchSize` in place, any batch that would contain more transactions than `maxBatchSize` would be divided into multiple batches of up-to `maxBatchSize` (minimum number of batches of maximum size, up to `maxBatchSize`). The sub-divided block would come with an artificial dependency so Ignition does not send any transaction until the previous block is fully processed and thus avoid straining the chain memory pool.

### Search terms

_No response_

DeVikingMark:
Implements support for `maxBatchSize` configuration option in Ignition.

When a batch exceeds `maxBatchSize`, it is automatically split into multiple smaller batches. Artificial dependencies are added between split batches to ensure sequential execution and prevent mempool strain.

Closes #7648

### DIFF
--- a/v-next/hardhat-ignition-core/test/batcher.ts
+++ b/v-next/hardhat-ignition-core/test/batcher.ts
@@ -179,21 +179,100 @@ describe("batcher", () => {
       [["Module1#Contract1"], ["Module1#Contract3"]],
     );
   });
+
+  it("should split batches exceeding maxBatchSize", () => {
+    const ignitionModule = buildModule("Module1", (m) => {
+      const contract1 = m.contract("Contract1");
+      const contract2 = m.contract("Contract2");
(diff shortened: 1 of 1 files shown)


michalbrabec (commented):
@DeVikingMark thanks for this PR, the changes look good.
Could you please add a bit more tests, include some more complex scenarios to make sure the batching works as expected (multiple initially too big batches or similar)?

I ran all the CIs.
Could you please resolve conflicts, fux linter errors and failing tests?

PR #7672, opened 2025-11-06
chore: update version of ledger in prep for release

kanej:
Add a changeset that sets the `hardhat-ledger` version to 3.0.0.

Merging this setup will lead to a release of the Hardhat 3 version of `hardhat-ledger`.

### DIFF
--- a/v-next/example-project/package.json
+++ b/v-next/example-project/package.json
@@ -31,7 +31,7 @@
     "@nomicfoundation/ignition-core": "workspace:^3.0.4",
     "@nomicfoundation/hardhat-ignition-viem": "workspace:^3.0.4",
     "@nomicfoundation/hardhat-keystore": "workspace:^3.0.3",
-    "@nomicfoundation/hardhat-ledger": "workspace:^1.2.1",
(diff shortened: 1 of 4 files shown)


v-next/example-project/package.json
```diff
@@ -31,7 +31,7 @@
     "@nomicfoundation/ignition-core": "workspace:^3.0.4",
     "@nomicfoundation/hardhat-ignition-viem": "workspace:^3.0.4",
     "@nomicfoundation/hardhat-keystore": "workspace:^3.0.3",
-    "@nomicfoundation/hardhat-ledger": "workspace:^1.2.1",
+    "@nomicfoundation/hardhat-ledger": "workspace:^2.0.0",
```
ChristopherDedominici, line 34: Just to confirm: why 2 and not 3 to be aligned with the other plugins?

kanej: There is a trick. I bumped the version here - but also add a major update. So when this merges, `changeset version` will set both this ref and the version of the ledger package to `3.0.0`.

PR #7678, opened 2025-11-08
Fix/network without extra base fee per gas value

wighawag:
This PR fix issues with some network that do not follow the spec and do not add an extra value in the response of `eth_feeHistory` for the baseFeePerGas array

Somnia Testnet for example

It does not change the behavior for other

### DIFF
--- a/.changeset/forty-rice-clap.md
+++ b/.changeset/forty-rice-clap.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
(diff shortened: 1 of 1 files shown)


fvictorio:
Hey Ronan, thanks for this. I guess it makes sense. Technically the `maxFeePerGas` value for these networks will be computed differently. A better fix would be to actually compute the proper base fee for the next block ourselves. But I don't think it makes sense to do that just for some edge cases.

This PR has a couple of unrelated commits though. Could you rebase it so that there's only one commit with this change?

wighawag:
just did it, thanks!

fvictorio:
Thanks!

PR #7686, opened 2025-11-11
Fix `test solidity` and `build` issues

alcuadrado:
This PR implements a few small changes to the `build` and `test solidity` tasks:

- It always clears the compilation progress spinner. It wasn't cleared on compilation errors.
- It fixes how `test solidity` builds contracts and tests. Previously it wasn't building the contracts in some cases. They are now treated as two separate calls to `build`.
- It moves the default `verbosity` to a contstant. 
- It enables `CollectStackTraces.Always` when running solidity tests with verbosity over the default.
- Adds a log to the test tracing errors recommending increasing the verbosity (this is a temporary fix, we should have an explicit flag for that)
- It throws an error if a file provided to build isn't neither a test or a contract. This used to make `pnpm hardhat test solidity non-existing.sol` fail silently.

### DIFF
--- a/v-next/hardhat/test/internal/builtin-plugins/solidity-test/helpers.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/solidity-test/helpers.ts
@@ -67,7 +67,7 @@ describe("solidityTestConfigToSolidityTestRunnerConfigArgs", () => {
     }
   });
 
-  it("should enable always tracinng for verbosities above the default one", async () => {
+  it("should enable always tracing for verbosities above the default one", async () => {
     for (const verbosity of [1, 2, 3, 4, 5, 6, 7]) {
       const args = await solidityTestConfigToSolidityTestRunnerConfigArgs({
         chainType: GENERIC_CHAIN_TYPE,


alcuadrado:
~We should have follow up tasks to test these changes, @michalbrabec~

~> - It enables CollectStackTraces.Always when running solidity tests with verbosity over the default.~
~> - It throws an error if a file provided to build isn't neither a test or a contract. This used to make pnpm hardhat test solidity non-existing.sol fail silently.~

~But if this looks good to @schaable let's release it.~

PS: I added them

alcuadrado:
Well, this uncovered some broken tests that are now failing. They were passing but not actually checking what they are doing

alcuadrado:
Or maybe we should consider every .t.sol file a test

schaable (approved):
Reviewed in a meeting. LGTM

alcuadrado:
> Or maybe we should consider every .t.sol file a test

Chatted with @fvictorio and @schaable and decided to leave it as is and fix the test until we have more feedback

PR #7690, opened 2025-11-12
feat: add support for Osaka hardfork

Wodann:

---

This PR introduces the Osaka hardfork in Hardhat 2 by bumping EDR to [v0.12.0-next.16](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.16).

Only a small change was required in the configuration logic. Since the Osaka hardfork introduces the `P256Verify` precompile, we no longer need to add the precompile manually when the `enableRip7212` option is `true`.

### DIFF
--- a/.github/workflows/hardhat-core-ci.yml
+++ b/.github/workflows/hardhat-core-ci.yml
@@ -1,7 +1,5 @@
 name: hardhat-core CI
 
-# Note: this workflow uses `macos-13` to use x86
-
 on:
   push:
     branches:
@@ -40,7 +38,7 @@ jobs:
       fail-fast: false
       matrix:
         node: [20.0.0]
-        os: ["macos-13", "ubuntu-latest", "windows-latest"]
+        os: ["macos-latest", "ubuntu-latest", "windows-latest"]
     steps:
       - uses: actions/checkout@v4
 
@@ -78,7 +76,7 @@ jobs:
       fail-fast: false
       matrix:
         node: [20.0.0]
-        os: ["macos-13", "ubuntu-latest", "windows-latest"]
+        os: ["macos-latest", "ubuntu-latest", "windows-latest"]
     steps:
       - uses: actions/checkout@v4
 


packages/hardhat-core/src/internal/core/config/config-resolution.ts
```diff
@@ -179,7 +186,9 @@ function resolveHardhatNetworkConfig(
 
   const blockGasLimit =
     hardhatNetworkConfig.blockGasLimit ??
-    clonedDefaultHardhatNetworkParams.blockGasLimit;
+    (hardforkGte(getHardforkName(hardfork), HardforkName.OSAKA)
```
Wodann: The Osaka hardfork introduced a transaction gas cap. This makes the default block gas limit dependant on the hardfork. When the hardfork is post-Osaka, we use the transaction gas cap.

packages/hardhat-core/src/internal/core/config/default-config.ts
```diff
@@ -62,63 +62,6 @@ export const defaultHardhatNetworkParams: Omit<
    * https://api-TESTNET.etherscan.io/api?module=block&action=getblocknobytime&timestamp=TIMESTAMP&closest=before&apikey=APIKEY
    */
   chains: new Map([
-    [
```
Wodann, line 65: Instead of using chain overrides, we can depend on EDR's chain configurations for the removed maps

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -236,6 +240,12 @@ export class EdrProviderWrapper
       return account.privateKey;
     });
 
+    const precompileOverrides = config.enableRip7212
```
Wodann, line 243: Osaka introduced its own version of of P256Verify, so we don't need to override it

packages/hardhat-core/test/internal/core/config/config-resolution.ts
```diff
@@ -761,19 +761,6 @@ describe("Config resolution", () => {
           });
         });
       });
-
-      it("should default to the latest hardfork", async function () {
```
Wodann, line 765: Once the Osaka hardfork is activated in mainnet, EDR will do a new release to switch the "latest" hardfork to Osaka.

packages/hardhat-core/test/internal/core/config/default-config.ts
```diff
@@ -1,24 +0,0 @@
-import { assert } from "chai";
-
-import { defaultHardhatNetworkParams } from "../../../../src/internal/core/config/default-config";
-import { HardforkName } from "../../../../src/internal/util/hardforks";
-
-describe("Default config", function () {
-  it("should include block numbers for all hardforks", async function () {
```
Wodann, line 7: We now depend on EDR's chain configs

Wodann:
I have no idea why [this](https://github.com/NomicFoundation/hardhat/actions/runs/19385333002/job/55471502783?pr=7690) CI job is failing. Any ideas?

packages/hardhat-core/src/internal/core/config/config-resolution.ts
michalbrabec: - disconnect default transaction gas limit from block gas limit (in ETH call and send transaction)
- introduce default transaction gas limit network configuration
    - set to 2^24 by default
    - enable the config option only for EDR-based networks
    - supported values are (0, 2^24], throw an error otherwise
    - document the behaviour (post Osaka) and link the docs in the error message
    
geth resolved the issue by completely disabling the check for the transaction gas limit

Wodann: For reference, geth provides a config option to disable the check for the transaction gas limit. By default, geth still runs the checks, but a user can override it.

michalbrabec: we agreed to keep the block gas limit as the default transaction gas limit, but replace the block gas limit value by 16M for transactions post Osaka, prototype `defaultTxGasLimit = (osaka ? 16M : blockGasLimit)`

Wodann: Handled by EDR in https://github.com/NomicFoundation/edr/pull/1200

Wodann: Custom code removed in commit https://github.com/NomicFoundation/hardhat/pull/7690/commits/f6284d422b9b5417933cee5a8d33f77d418621c0

packages/hardhat-core/package.json
```diff
@@ -103,7 +103,7 @@
   "dependencies": {
     "@ethereumjs/util": "^9.1.0",
     "@ethersproject/abi": "^5.1.2",
-    "@nomicfoundation/edr": "^0.12.0-next.7",
+    "@nomicfoundation/edr": "0.12.0-next.16",
```
Wodann, line 106: Using `pnpm add`, this was converted to no longer include the `^`  at the start. Is that expected?

Claude tells me it's a result of it being a pre-release (due to the `-next.16` postfix)

pnpm-lock.yaml
```diff
@@ -43,7 +43,7 @@ importers:
     dependencies:
       '@typescript-eslint/utils':
         specifier: ^5.61.0
-        version: 5.62.0(eslint@8.57.1)(typescript@5.3.3)
+        version: 5.62.0(eslint@8.57.1)(typescript@5.0.4)
```
Wodann, line 46: The typescript version was downgraded after applying `pnpm add` for EDR's version bump

Wodann:
> I have no idea why [this](https://github.com/NomicFoundation/hardhat/actions/runs/19385333002/job/55471502783?pr=7690) CI job is failing. Any ideas?

This is [still failing](https://github.com/NomicFoundation/hardhat/actions/runs/19545173608/job/55962029111?pr=7690) in the latest version.

packages/hardhat-core/package.json
michalbrabec: Same as HH3

PR #7691, opened 2025-11-12
feat: add support for Osaka hardfork

Wodann:

---

This PR introduces the Osaka hardfork in Hardhat 3.

Two changes were required on the Hardhat side:
- When the Osaka hardfork is used, we use a default block gas limit of 16,777,216 instead of 30,000,000. This is the transaction gas cap (introduced in Osaka)
- Previously Hardhat was always defaulting to the latest hardfork, even if not enabled. I changed the logic to use the latest hardfork reported by EDR. Once the Osaka hardfork is activated on mainnet, we'll do a release to change this value. This avoids both EDR and Hardhat having to maintain this logic.

### DIFF
--- a/v-next/hardhat/package.json
+++ b/v-next/hardhat/package.json
@@ -87,7 +87,7 @@
     "typescript": "~5.8.0"
   },
   "dependencies": {
-    "@nomicfoundation/edr": "0.12.0-next.15",
+    "@nomicfoundation/edr": "0.12.0-next.16",
     "@nomicfoundation/hardhat-errors": "workspace:^3.0.4",
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.5",
     "@nomicfoundation/hardhat-zod-utils": "workspace:^3.0.1",
--- a/.changeset/dull-beers-learn.md
+++ b/.changeset/dull-beers-learn.md
@@ -2,4 +2,6 @@
 "hardhat": patch
 ---
 
-Added support for Osaka hardfork. Bumped EDR version to [`0.12.0-next.15`](https://www.npmjs.com/package/@nomicfoundation/edr/v/0.12.0-next.15)
+Bumped EDR version to [`0.12.0-next.16`](https://www.npmjs.com/package/@nomicfoundation/edr/v/0.12.0-next.16)
+- Added support for Osaka hardfork
+- Added full support for OP stack Isthmus hardfork
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -147,8 +147,8 @@ importers:
   v-next/hardhat:
     dependencies:
       '@nomicfoundation/edr':
-        specifier: 0.12.0-next.15
-        version: 0.12.0-next.15
+        specifier: 0.12.0-next.16
+        version: 0.12.0-next.16
       '@nomicfoundation/hardhat-errors':
         specifier: workspace:^3.0.4
         version: link:../hardhat-errors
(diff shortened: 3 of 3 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts
```diff
@@ -291,3 +301,15 @@ export function resolveInitialBaseFeePerGas(
...
+function resolveDefaultBlockGasLimit(
+  hardfork: string,
+  chainType: ChainType | undefined = GENERIC_CHAIN_TYPE,
+): bigint {
+  const isPostOsaka =
+    chainType === OPTIMISM_CHAIN_TYPE
+      ? false // TODO: implement OP hardfork comparison once we have more OP hardforks
+      : hardforkGte(hardfork, "osaka", chainType);
+
+  return isPostOsaka ? 16_777_216n : 30_000_000n;
```
Wodann: The Osaka hardfork introduced a transaction gas cap. This makes the default block gas limit dependant on the hardfork. When the hardfork is post-Osaka, we use the transaction gas cap.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/types/hardfork.ts
```diff
@@ -43,15 +50,10 @@ export function getHardforks(chainType: ChainType): string[] {
...
-const L1_LATEST_HARDFORK: L1HardforkName =
-  L1_HARDFORK_ORDER[L1_HARDFORK_ORDER.length - 1];
-const OP_LATEST_HARDFORK: OpHardforkName =
-  OP_HARDFORK_ORDER[OP_HARDFORK_ORDER.length - 1];
-
 export function getCurrentHardfork(chainType: ChainType): string {
   return chainType === OPTIMISM_CHAIN_TYPE
-    ? OP_LATEST_HARDFORK
-    : L1_LATEST_HARDFORK;
+    ? edrOpHardforkToHardhatOpHardforkName(opLatestHardfork())
```
Wodann, line 55: Instead of using the last hardfork in the list, we use EDR's reported "latest" hardfork.

Once the Osaka hardfork is activated in mainnet, EDR will do a new release to switch the "latest" hardfork to Osaka.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
```diff
@@ -66,6 +69,90 @@ import { L1HardforkName, OpHardforkName } from "../types/hardfork.js";
 
 import { getL1HardforkName, getOpHardforkName } from "./hardfork.js";
 
+export function edrL1HardforkToHardhatL1HardforkName(
```
Wodann, line 72: Both functions are needed for conversion from EDR hardfork types to Hardhat hardfork name types.

v-next/hardhat/test/internal/builtin-plugins/network-manager/config-resolution.ts
```diff
@@ -208,6 +208,20 @@ describe("config-resolution", () => {
       assert.equal(edrNetworkConfig.throwOnTransactionFailures, true);
     });
 
+    it("should use the transaction gas cap as default block gas limit for post-Osaka hardforks", () => {
```
Wodann: Added a test to validate that we use the transaction gas cap as the default block gas limit in post-Osaka hardforks.

Wodann: Removed in https://github.com/NomicFoundation/hardhat/pull/7691/commits/1f15b34beccc29115022346046caea5108ee5de5, as the custom logic was removed. EDR implemented a different solution in https://github.com/NomicFoundation/edr/pull/1200

v-next/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts
Wodann: Removed in https://github.com/NomicFoundation/hardhat/pull/7691/commits/1f15b34beccc29115022346046caea5108ee5de5, as the custom logic was removed. EDR implemented a different solution in https://github.com/NomicFoundation/edr/pull/1200

v-next/hardhat/package.json
```diff
@@ -87,7 +87,7 @@
     "typescript": "~5.8.0"
   },
   "dependencies": {
-    "@nomicfoundation/edr": "0.12.0-next.14",
+    "@nomicfoundation/edr": "0.12.0-next.16",
```
Wodann, line 90: Using `pnpm add`, this was converted to no longer include the `^`  at the start. Is that expected?

Claude tells me it's a result of it being a pre-release (due to the `-next.16` postfix)

michalbrabec: It follows the same convention as the previous releases

PR #7705, opened 2025-11-19
Improve hardhat-viem's README

fvictorio:


v-next/hardhat-viem/README.md
```diff
@@ -41,3 +41,140 @@ console.log(await counter.read.x());
...
+The configuration object supports the following properties:
+
+- `confirmations`: the number of confirmations to wait after the deployment transaction is mined. Default is `1`.
+- `libraries`: an object specifying the libraries to link in the contract.
+
+### `sendDeploymentTransaction(contractName, constructorArgs?, sendDeploymentContractConfig?)`
+
+Same as `deployContract`, but doesn't wait for the deployment to be mined, and returns an object with two properties:
+
+- `contract`: the contract instance, which is available even before the deployment is mined.
```
schaable:
```suggestion
- `contract`: the contract instance, which is available even before the transaction is mined.
```

schaable: Alternatively, deployment transaction. But I feel "deployment" alone is slightly off. It’s phrased the same way in the previous sentence as well.

schaable:
`deployContract` and `sendDeploymentTransaction` can also take any of these properties through the options object. The most relevant one is probably `client`, since users have asked about this before:

```
export interface SendTransactionConfig {
  client?: KeyedClient;
  gas?: bigint;
  gasPrice?: bigint;
  maxFeePerGas?: bigint;
  maxPriorityFeePerGas?: bigint;
  value?: bigint;
}
```

fvictorio:
> deployContract and sendDeploymentTransaction can also take any of these properties through the options object

Ah, good catch, I missed the `extends` part of the type. Fixed.

PR #7707, opened 2025-11-19
Generate typechain types on partial compilation

### LINKED ISSUE #6311: [Typechain] modify the plugin to generate types for a subset of contracts

antico5:
A new hook was added to the solidity plugin, `onBuildComplete`. The type generation is now fired off this hook instead of the `onCleanupArtifacts`, because this last one is only triggered when a full compilation is performed. The artifact paths, needed by typechain, are retrieved through the artifact manager. So typechain can have access to all artifact paths even if a partial compilation was performed.

Closes #6311

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
@@ -109,6 +109,11 @@ export class ArtifactManagerImplementation implements ArtifactManager {
     return allFullyQualifiedNames;
   }
 
+  public async getAllArtifactPaths(): Promise<ReadonlySet<string>> {
+    const { allArtifactPaths } = await this.#getFsData();
+    return allArtifactPaths;
+  }
+
   public async getAllBuildInfoIds(): Promise<ReadonlySet<string>> {
     const paths = await getAllFilesMatching(
       path.join(this.#artifactsPath, BUILD_INFO_DIR_NAME),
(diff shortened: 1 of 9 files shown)


schaable (approved):
LGTM

kanej:
I am moving this to draft as I think it overlaps with the open zeppelin work I am doing.
My plan is to look this over, then rebase my changes against it - rather than making changes here. But lets not merge a new hook that we immediately change.

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
```diff
@@ -109,6 +109,11 @@ export class ArtifactManagerImplementation implements ArtifactManager {
     return allFullyQualifiedNames;
   }
 
+  public async getAllArtifactPaths(): Promise<ReadonlySet<string>> {
+    const { allArtifactPaths } = await this.#getFsData();
+    return allArtifactPaths;
+  }
```
kanej, line 115: This is an addition to the artifacts API. It seems reasonable as the default implementation will have this info from `FsData`.

Looking at the `ArtifactManager` API, is an alternative to use `getAllFullyQualifiedNames` then `getArtifactPath(contractNameOrFullyQualifiedName: string): Promise<string>`? @alcuadrado

kanej:
The next steps on this PR are getting a product sign off for:
1. The addition to the `ArtifactsManager` API mentioned above: https://github.com/NomicFoundation/hardhat/pull/7707#discussion_r2669173616
2. The name of the new build system hook

v-next/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
kanej: We are happy to add this method to the artifact manager.

PR #7717, opened 2025-11-24
fix: handle WebSocket unsubscribe cleanup promises properly

maradini77:
When a WebSocket connection closes, the code attempts to unsubscribe from all active `eth_subscribe` subscriptions. However, the cleanup uses `forEach` with an `async` callback, which creates promises that are never awaited. This leads to:

- **Unhandled promise rejections**: If `eth_unsubscribe` fails (e.g., provider already closed), it can crash the Node.js process in Node 16+ due to unhandled promise rejections
- **Resource leaks**: Failed unsubscriptions may leave active subscriptions on the provider side, consuming resources unnecessarily

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
@@ -120,12 +120,22 @@ export class JsonRpcHandler {
 
       // Clear any active subscriptions for the closed websocket connection.
       isClosed = true;
-      subscriptions.forEach(async (subscriptionId) => {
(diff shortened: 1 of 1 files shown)


schaable:
Hi, thanks for the PR. This change is minor but still requires review and adjustments on our side, and before moving forward we’d need to understand its impact and prioritize it accordingly. Please check our [policy](https://github.com/NomicFoundation/hardhat/blob/main/CONTRIBUTING.md#note-about-small-prs-and-airdrop-farming) on small/inconsequential PRs.

If this is affecting you, feel free to open an issue so we can better understand it. Thanks!

PR #7719, opened 2025-11-25
fix(hardhat): add missing 'F' character to hex validation patterns

Bilogweb3:
The hex validation regex in argument parsing was missing the letter `F` (and `f`).

The pattern `[\dABCDEabcde]` doesn't include `F`, so values like `0xFF`, `0xCAFE`, or `0xDEADBEEF` would incorrectly fail validation.

Changed to `[\dA-Fa-f]` which is also consistent with how hex is validated elsewhere in the codebase (see `anonymizer.ts`).

### DIFF
--- a/.changeset/fix-hex-validation-patterns.md
+++ b/.changeset/fix-hex-validation-patterns.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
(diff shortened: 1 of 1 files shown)


michalbrabec:
@Bilogweb3 Please include a changeset to ensure that the change is properly released.

Bilogweb3:
Ready, thanks.

PR #7726, opened 2025-11-27
chore: add reproducible example of mocha missing async warning

### LINKED ISSUE #7722: Improve error message when no await in chai matchers
A user lost time on their mocha test suite using chai matchers, as they where missing an await on an `expect`. The error shown was:

```shell
HardhatError: HHE100: An internal invariant was violated
```

We should try and detect for this case and show a clear error message suggesting that the likely cause is a lack of an await.

## Example 

```ts
it('Should transfer money to another wallet with extra value', async function () {
    const { accounts: { alice }, contracts: { drop }, others: { hashedElements, leaves, proofs, signature } } = await loadFixture(deployContractsFixture);
    const txn = await drop.claim(alice, 1, proofs[leaves.indexOf(hashedElements[0])], signature, { value: 10 });
    
    expect(txn).to.changeEtherBalance(ethers, alice, 10); // <-- missing await expect(txn)
});
```

## Investigation

An initial look suggests something like the following is happening

1. By missing out the `await` on `expect(txn).to.changeEtherBalance(ethers, alice, 10);`, the test completes and passes
2. The test assertion is then run (in this case `changeEtherBalance`) because it was not awaited
3. The assertion uses `ethers` underlying provider to make a `getBlock` request. The provider can't connect and `ethers` returns null 
4. The null block number trips the internal variant exception.

kanej:
This fix for #7722 is still in the investigation stage. The issue is that a missing async await against our mocha assertions can lead to an invariant error - we would prefer a clear message suggesting that an `await` is missing.

This PR adds an example project that replicates the error, to run it:

```shell
cd v-next/example-project-assertion
pnpm install
pnpm hardhat test mocha

# No contracts to compile

# Running Mocha tests


#   Hardhat3 test
#     ✔ Should transfer money to another wallet with extra value (331ms)

# Unhandled promise rejection:

# HardhatError: HHE100: An internal invariant was violated: The block doesn't exist
#     at assertHardhatInvariant (/workspaces/hardhat/v-next/hardhat-errors/src/errors.ts:237:11)
#     at getBalanceChange (/workspaces/hardhat/v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeEtherBalance.ts:100:3)
#     at async Promise.all (index 0)
```

## Investigation status

My current hypothesis is that the lack of an await, means that some lower layer is closed or cleaned up. When the assertion `expect(txn).to.changeEtherBalance(ethers, alice, 10)` is run, the test has already completed, so the calls to the in-memory EDR network connection fail, `ethers` returns null for the block number and so we get the invariant exception above (but also a passing test).

My guess was that the EDR connection was giving some odd error rather than a connection closed when `ethers` called `getBlock`, and was then falling back on `null`. However, closing the connection leads to a connection closed error, see:

https://github.com/NomicFoundation/hardhat/blob/7b335f9a64b46715b9db9082ff16bbe2cf95fc81/v-next/example-project-assertion/scripts/test-closed-connection.ts

## Next steps

Resolve the mechanism that leads `ethers` to return null for `getBlock`. Has the underlying connection been cleaned up with the completion of the test or is it some other reason? Setting up debugging is my next move.


zoeyTM:
After further investigation, I have found that this issue appears to be due to the garbage collector cleaning up the in-use `NetworkConnectionImplementation` instance inside hardhat's network manager, specifically this section:

https://github.com/NomicFoundation/hardhat/blob/7b335f9a64b46715b9db9082ff16bbe2cf95fc81/v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts#L176-L193

The local variable on line 179 does appear to serve its intended purpose according to the comment above it, however when the handler chain is run on line 188, inside the chains delayed execution is where the strange error is being thrown. 

Removing the local variable on line 179 or changing line 188 to invoke `this.#hookManager.runHandlerChain` directly instead of via the local variable; either of these options will independently fix the problem and the error will not be shown when running the tests in the reproducible example. I'm unsure if that is a solution we are happy with as we would lose the garbage collector inefficiencies mentioned in the comment, but those solutions do at least serve as evidence of where exactly the problem lies.

alcuadrado:
After the changes I pushed, the error now looks like this, which is what I expected.

image

Issue #7700, opened 2025-11-16
Changing the default chain type doesn't work


alcuadrado:
### Version of Hardhat

3.0.15

### What happened?

When you change the default Chain Type by doing this

```ts
import "hardhat/types/network";
declare module "hardhat/types/network" {
  interface ChainTypeConfig {
    defaultChainType: "op";
  }
}

export default defineConfig({
  defaultChainType: "op",
  // ...
});
```

This doesn't work, because the file `v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts` does `defaultChainType: resolvedConfig.defaultChainType ?? GENERIC_CHAIN_TYPE,` instead of `defaultChainType: userConfig.defaultChainType ?? GENERIC_CHAIN_TYPE,`.

Note the `userConfig`.

Moreover, I think we should move the default resolution of the `hardfork` field to the point where we execute `connect()`, because the resolution right now ignores `userConfig.defaultChainType`, and taking that into account wouldn't be enough.

For example, a project with this empty config:

```ts
import { defineConfig } from "hardhat/config";

export default defineConfig({});
```

And this `scripts/repro.ts`:

```ts
import { network } from "hardhat";

await network.connect({ chainType: "op" });
```

Fails when running the script:

```
$ pnpm hardhat run --no-compile scripts/repro.ts --show-stack-traces
Error HHE100: An internal invariant was violated: Invalid hardfork name prague

HardhatError: HHE100: An internal invariant was violated: Invalid hardfork name prague
    at assertHardhatInvariant (/workspaces/npmprojects/npmproject-34/node_modules/.pnpm/@nomicfoundation+hardhat-errors@3.0.5/node_modules/@nomicfoundation/hardhat-errors/src/errors.ts:237:11)
    at getOpHardforkName (/workspaces/npmprojects/npmproject-34/node_modules/.pnpm/hardhat@3.0.15/node_modules/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/hardfork.ts:25:3)
    at hardhatOpHardforkToEdrSpecId (/workspaces/npmprojects/npmproject-34/node_modules/.pnpm/hardhat@3.0.15/node_modules/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts:79:24)
[...]

PR #7728, opened 2025-11-27
Re-validate and re-resolve chainType-related fields on connect

schaable:
This PR makes sure all `chainType`-related fields (like `hardfork`) are re-validated and re-resolved during network connection. The `resolvedChainType` is now passed to `resolveUserConfigToHardhatConfig` as part of the `userConfig`, so everything gets validated and resolved again consistently. It also updates the `override` object type so `chainType` can only be provided as a top-level property in `connect()`, avoiding confusion. The PR also removes the error for mismatched `chainType`, since that case is no longer valid, and adds tests covering all scenarios.

Closes https://github.com/NomicFoundation/hardhat/issues/7700

### DIFF
--- a/.changeset/loose-peaches-swim.md
+++ b/.changeset/loose-peaches-swim.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-The network config is now correctly re-validated and re-resolved on `connect()` based on `chainType` ([#7700](https://github.com/NomicFoundation/hardhat/issues/7700)).
+Fixed the network config to re-validate and re-resolve `chainType` on `connect()` ([#7700](https://github.com/NomicFoundation/hardhat/issues/7700)).


v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -307,12 +292,16 @@ export class NetworkManagerImplementation implements NetworkManager {
...
     resolvedNetworkName: string,
     networkConfigOverride: NetworkConfigOverride | undefined = {},
+    resolvedChainType: ChainTypeT,
   ): Promise<NetworkConfig> {
     const existingNetworkConfig = this.#networkConfigs[resolvedNetworkName];
-    if (Object.keys(networkConfigOverride).length === 0) {
+    if (
+      Object.keys(networkConfigOverride).length === 0 &&
+      resolvedChainType === existingNetworkConfig.chainType
+    ) {
```
kanej, line 304: Can I check I have followed the change here.
In the previous commit it only checked `networkConfigOverride` was empty i.e. no overrides. This was wrong - because the chainType is now provided as well as the overrides. So this check need to be updated, it should only short circuit if there are no overrides and the chainType remains the same as the original network config.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -142,37 +144,20 @@ export class NetworkManagerImplementation implements NetworkManager {
       );
     }
 
+    /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions --
+     * Type assertion is safe: defaultChainType ensures non-undefined, and the
+     * resolved value will be ChainTypeT (if provided) or a fallback that
+     * satisfies the ChainType | string constraint */
+    const resolvedChainType = (chainType ??
+      existingNetworkConfig.chainType ??
```
kanej, line 152: Can I just double check you are happy that `resolvedNetworkConfig` is no longer used and `existingNetworkConfig` is used?
Is this because resolution won't affect the chainType?

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -340,9 +329,12 @@ export class NetworkManagerImplementation implements NetworkManager {
       );
     }
 
-    const newConfig = deepMerge(this.#userConfig, {
+    const userConfigWithOverrides = deepMerge(this.#userConfig, {
```
kanej, line 332: Good call.

kanej (commented):
I have left a couple of questions. But the tests and changes look good.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
schaable: That is correct 💪

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
schaable: This should be fine since the `resolvedChainType` value is not changed by the `#resolveNetworkConfig` function.

kanej (approved):
I have done a small round of manual tests - no issues.

Specifically within the example project, recreating the original issue: https://github.com/NomicFoundation/hardhat/issues/7700


```ts
// hardhat.config.ts

import "hardhat/types/network";

declare module "hardhat/types/network" {
  interface ChainTypeConfig {
    defaultChainType: "op";
  }
}

export default defineConfig({
  solidity: "0.8.22",
  defaultChainType: "op",
});
```

```ts
./scripts/chain-type-change.ts

const connection = await network.connect({
  chainType: "generic",
});

console.log("network connection created with chaintype: ", connection.chainType);
```

I tested changing the leaving out and changing the `hardhat.config.ts` `defaultChainType`, and how that interacts with the `chainType` passed in `network.connect({...})`.

The `network.connect` takes precedence over `defaultChainType`. `generic` is used if none are specified.

The test suite was thorough and easy to understand. Nice work. LGTM!

.changeset/loose-peaches-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+The network config is now correctly re-validated and re-resolved on `connect()` based on `chainType` ([#7700](https://github.com/NomicFoundation/hardhat/issues/7700)).
```
kanej:
```suggestion
Fixed the network config to re-validate and re-resolve `chainType` on `connect()` ([#7700](https://github.com/NomicFoundation/hardhat/issues/7700)).
```

PR #7744, opened 2025-12-01
feat: add new invokeSolc hook

### LINKED ISSUE #7646: Add build system hooks for responding to compilations
With the Hardhat 3 build system we intentionally added only the plugins needed for the core functionality, with the intention of adding further hooks as real use cases were reported. The OpenZeppelin team have, to enable the upgrade contract plugin, the following:

- `onCompile` which is run once a compilation has been complete

The new hook will be added to `SolidityHooks`. Note the `onCompile` hook will run even if there are compilation errors.

kanej:
Add a `invokeSolc` hook that is called to run the `solc` compiler that Hardhat has selected for a given compilation job. The hook passes in:

* the Compiler instance that Hardhat has selected for the compilation job
* the Solc input generated from the compilation job - presenting an extension point for transforming the input
* the Solc config that was used by Hardhat to load the particular Solc compiler

It returns the Solc output from running `solc`.

The `invokeSolc` provides an extension point that can be used to transform or manipulate the Solc input before the running of `solc`, or similarly to manipulate the output from the compiler.

## invokeSolc Hook

```ts
/**
 * Hook triggered to invoke a passed in Solc compiler on the
 * Solc input generated for a given compilation job.
 * This hook allows for manipulating the Solc input passed into the Solc
 * compiler Hardhat has selected for the compilation job, and similarly to
 * manipulate the Solc output.
 *
 * @param context The hook context.
 * @param compile The Solc compiler selected by Hardhat for this compilation
 * job.
 * @param solcInput The solc input json constructed from the compilation
 * job.
 * @param solcConfig The configuration used to setup solc e.g. version.
 * @param next A function to call the next handler for this hook, or the
 * default implementation if no more handlers exist.
 */
invokeSolc(
  context: HookContext,
  compiler: Compiler,
  solcInput: CompilerInput,
  solcConfig: SolcConfig,
  next: (
    nextContext: HookContext,
    nextCompiler: Compiler,
    nextSolcInput: CompilerInput,
    nextSolcConfig: SolcConfig,
  ) => Promise<CompilerOutput>,
): Promise<CompilerOutput>;
```

Resolve #7646


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -633,9 +633,17 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
-      await runnableCompilationJob.getSolcInput(),
+    const input = await runnableCompilationJob.getSolcInput();
+
+    const output = await compiler.compile(input);
+
+    await this.#hooks.runHandlerChain(
+      "solidity",
+      "onSolcCompileComplete",
+      [runnableCompilationJob, input, output],
+      async () => {},
```
kanej: I have opted to not return anything - an alternative would be to let plugins fail the build at this point by returning a `bool` or a structured result object. This can be added later if it proves needed.

v-next/hardhat/test/fixture-projects/solidity/simple-project/contracts/A.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: UNLICENSED
```
kanej, line 1: The new fixture project is a minimal project for compilation. The solidity example project has contracts that won't compile.

afa7789:
up

kanej:
@fvictorio queried the hook, arguing it should be something more like:
```
solcCompile(
  context: HookContext,
  solcConfig: SolcConfig,
  solcInput: CompilerInput,
  next: (
    nextContext: HookContext,
    nextSolcConfig: SolcConfig,
    nextSolcInput: CompilerInput,
  ) => Promise<CompilerOutput>,
): Promise<CompilerOutput>;
```
So taking the input and giving the output.
This would work if we could do this if we include the `solcConfig` settings that OpenZeppelin required as well.

When I looked to implement this though I ran into the question of how to deal with the compiler. Is it passed in also? It seems like the hook should encapsulate the selection and retrieval of the compiler. But then should the hook pass the compiler back as a result, to allow it to be propagated up the call chain as `runCompilationJob` requires?

v-next/hardhat/test/internal/builtin-plugins/solidity/hooks.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import { useFixtureProject } from "@nomicfoundation/hardhat-test-utils";
+
+import { createHardhatRuntimeEnvironment } from "../../../../src/hre.js";
+
+describe("solidity - hooks", () => {
+  describe("invokeSolc", () => {
+    useFixtureProject("solidity/simple-project");
+
+    const expectedSolidityVersion = "0.8.23";
```
kanej, line 25: I am controlling the Solidity version here to let me check for it against the passed in Solc compiler instance.

kanej:
> @fvictorio queried the hook, arguing it should be something more like:
> 
> ```
> solcCompile(
>   context: HookContext,
>   solcConfig: SolcConfig,
>   solcInput: CompilerInput,
>   next: (
>     nextContext: HookContext,
>     nextSolcConfig: SolcConfig,
>     nextSolcInput: CompilerInput,
>   ) => Promise<CompilerOutput>,
> ): Promise<CompilerOutput>;
> ```
> 
> So taking the input and giving the output. This would work if we could do this if we include the `solcConfig` settings that OpenZeppelin required as well.
> 
> When I looked to implement this though I ran into the question of how to deal with the compiler. Is it passed in also? It seems like the hook should encapsulate the selection and retrieval of the compiler. But then should the hook pass the compiler back as a result, to allow it to be propagated up the call chain as `runCompilationJob` requires?

After discussion we opted to follow this path, but to go with a `invokeSolc` hook that is passed the Hardhat selected solc compiler; rather than a hook that encapsulated the solc compiler selection and loading as well.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -633,9 +633,17 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
 
-    const output = await compiler.compile(
-      await runnableCompilationJob.getSolcInput(),
+    const input = await runnableCompilationJob.getSolcInput();
+
+    const output = await this.#hooks.runHandlerChain(
+      "solidity",
+      "invokeSolc",
+      [compiler, input, runnableCompilationJob.solcConfig],
+      async () => {
```
alcuadrado: you should use the compiler and input that you receive here

alcuadrado: otherwise they won't be overridable

alcuadrado (approved):
pre-approving, but take a look at the comment i left

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
kanej: 🤦 I have updated the hook call, and updated the test as well to cover that overriding is happening.

Amxx:
This is awesome. Can't wait for it to be released !

PR #7755, opened 2025-12-04
fix: Osaka support for Ethers

Wodann:

This PR contains a potential fix for Ethers after the Osaka hardfork

### DIFF
--- a/packages/hardhat-core/src/internal/core/config/config-resolution.ts
+++ b/packages/hardhat-core/src/internal/core/config/config-resolution.ts
@@ -181,7 +181,7 @@ function resolveHardhatNetworkConfig(
     hardhatNetworkConfig.blockGasLimit ??
     clonedDefaultHardhatNetworkParams.blockGasLimit;
 
-  const gas = hardhatNetworkConfig.gas ?? blockGasLimit;
+  const gas = hardhatNetworkConfig.gas ?? "auto";
   const gasPrice =
     hardhatNetworkConfig.gasPrice ?? clonedDefaultHardhatNetworkParams.gasPrice;
   const initialBaseFeePerGas =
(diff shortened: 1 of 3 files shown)


packages/hardhat-ethers/src/signers.ts
```diff
@@ -59,24 +59,6 @@ export class HardhatEthersSigner implements ethers.Signer {
       if (configuredGasLimit !== "auto") {
         // if the resolved gas config is a number, we use that
         gasLimit = configuredGasLimit;
-      } else {
```
Wodann, line 63: On the backend, EDR will already use the `blockGasLimit` provided by Hardhat for pre-Osaka hardforks. For post-Osaka hardforks, we use the transaction gas cap

packages/hardhat-core/src/internal/core/config/config-resolution.ts
```diff
@@ -181,7 +181,7 @@ function resolveHardhatNetworkConfig(
     hardhatNetworkConfig.blockGasLimit ??
     clonedDefaultHardhatNetworkParams.blockGasLimit;
 
-  const gas = hardhatNetworkConfig.gas ?? blockGasLimit;
+  const gas = hardhatNetworkConfig.gas ?? "auto";
```
Wodann, line 184: Instead of always using the `blockGasLimit` to determine the `gas` limit, we set the `undefined` case to `"auto"`, allowing different call sites to calculate the gas limit.

**⚠️ NOTE**: This would be a breaking change, as previously we'd always use the `blockGasLimit` and now we might use the gas limit determined by the `"auto"` logic of different call sites.

kanej:
This PR became: https://github.com/NomicFoundation/hardhat/pull/7756

PR #7756, opened 2025-12-05
fix for Hardhat 2 osaka gas limit issues

kanej:
Osaka enabled EIP-7825, which sets a per transaction gas limit of 16777216 (2^24). This limit is enforced in EDR when the hardfork is set to Osaka, which is set as the default.

Hardhat 2 sets gas limits on transactions in multiple places including:

* Hardhat config network defaults for gas limits
* within network stack providers that calculate gas before sending on the transaction
* within the Ethers providers of `hardhat-ethers`

All of these gas settings interact to produce failures when running against an in-memory or standalone Hardhat node with a `osaka` hardfork set. There may also be some edge cases where running against `Mainnet`/`Sepolia` fails.

Hardhat 2 defaults the gas limit in config to the BlockGasLimit (30million) - which will fail the new transaction limit. However this value is only sometime used. Both `viem` and `ethers` via their plugins will tend to estimate gas rather than relying on the default. Similarly the `http` provider in Hardhat overwrites the gas default to be `auto` - leading to estimation of an acceptable value. Code paths going through the EDR provider or that are connecting to a node identified as a Hardhat node will set the gas either based on config (default 30m based on BlockGasLimit) or set it directly to the block gas limit.

This PR is designed to make the minimal intervention necessary to allow an upgraded Hardhat 2 project to continue to:

1. Run an existing test suite with the assumption of Osaka as the Hardfork
2. Deploy with Ignition
3. Run against Mainnet/Sepolia

The minimal intervention is to:

- change the default gas limit in config to the minimum of the BlockGasLimit or the Fusaka transaction limit
- change the gas providers gas setting behaviour to similarly take the minimum of the BlockGasLimit or transaction gas limit
- change the Ethers providers behaviour when connecting to a Hardhat node to take the minimum of BlockGasLimit or transaction gas limit

This approach does not take into account the hardfork. In effect it shifts the default from 30million to 16777216. This is a change, but for any project running against `osaka` it is strictly a logical improvement - 30 million will straightforwardly fail.

## Alternatives

Set the default gas limit to `auto` and remove the special cases of gas limit setting in the providers and Ethers. This was deemed a larger breaking change than keeping a known numerical default but reducing it to the transaction limit.

Another alternative is to keep to this minimal approach but try to vary based on default hardfork - this would requiring wiring that information through to the key usage points. This was determined to be more complicated than the additional refinement would warrant.

## Manual tests

Invoked `npx changeset version` then publish to a local verdaccio. Install `hardhat@hh2` via verdaccio then initialize new projects from template to test:

- [x] Ethers based Hardhat 2 template:
  - [x] unit tests pass (i.e. `npx hardhat test`)
  - [x] local ignition deployment passes
  - [x] A send transaction via `ethers` works:
    - [x] against in-memory
    - [x] against `npx hardhat node` over an http connection
  - [x] A send transaction via the Hardhat provider works:
    - [x] against in-memory
    - [x] against `npx hardhat node` over an http connection
  - [x] Sepolia
    - [x] Send a transaction via provider
    - [x] Deploy a contract via Ignition
    - [x] Deploy a contract via ethers and run a function
- [x] Viem based Hardhat 2 template:
  - [x] unit tests pass (i.e. `npx hardhat test`)
  - [x] local ignition deployment passes
  - [x] A send transaction via `viem` works:
    - [x] against in-memory
    - [x] against `npx hardhat node` over an http connection
  - [x] A send transaction via the Hardhat provider works:
    - [x] against in-memory
    - [x] against `npx hardhat node` over an http connection
  - [x] Sepolia
    - [x] Send a transaction via provider
    - [x] Deploy a contract via Ignition
    - [x] Deploy a contract via viem and run a function
- [x] External repositories
  - [x] OpenZeppelin (test fails due to [ModEXP](https://eips.ethereum.org/EIPS/eip-7823) changes affecting RSA tests that leverage it)
  - [x] Uniswap (test fails due to [gas cost changes](https://eips.ethereum.org/EIPS/eip-7883) )
  - [x] Rocketpool

### DIFF
--- a/.changeset/puny-mangos-open-1.md
+++ b/.changeset/puny-mangos-open-1.md
@@ -1,0 +1,5 @@
+---
+"hardhat": minor
+---
+
+Changed default L1 hardfork to `osaka` ([@nomicfoundation/edr@0.12.0-next.17](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.17))
--- a/.changeset/puny-mangos-open-2.md
+++ b/.changeset/puny-mangos-open-2.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Changed default OP stack hardfork to `isthmus` ([@nomicfoundation/edr@0.12.0-next.17](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.17))
--- a/.changeset/puny-mangos-open-3.md
+++ b/.changeset/puny-mangos-open-3.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed default transaction gas limit for post-Osaka hardforks in OP stack and generic chains ([@nomicfoundation/edr@0.12.0-next.17](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.17))
--- a/.changeset/puny-mangos-open.md
+++ b/.changeset/puny-mangos-open.md
@@ -1,8 +1,0 @@
----
-"hardhat": minor
----
-
-Bumped EDR version to [`0.12.0-next.17`](https://www.npmjs.com/package/@nomicfoundation/edr/v/0.12.0-next.17)
-- Changed default L1 hardfork to Osaka
-- Changed default OP stack hardfork to Isthmus
-- Fixed default transaction gas limit for post-Osaka hardforks in OP stack and generic chains


nebasuke:
Had another look. While I agree with having the max transaction gas cap, we should additionally make sure that it's not easy to accidentally set the transaction gas cap net below the transaction gas limit, as that is the error EDR was throwing. 

`ProviderError: Transaction gas limit is 30000000 and exceeds transaction gas cap of 16777216`.

Relevant error:
https://github.com/NomicFoundation/edr/blob/10f25fdb2c2a244e4960cac298f893ecde227b79/crates/mem_pool/src/lib.rs#L113

Relevant EDR line: https://github.com/NomicFoundation/edr/blob/10f25fdb2c2a244e4960cac298f893ecde227b79/crates/mem_pool/src/lib.rs#L460

packages/hardhat-core/src/internal/core/config/default-config.ts
```diff
@@ -36,7 +36,7 @@ export const defaultHardhatNetworkParams: Omit<
   HardhatNetworkConfig,
   "gas" | "initialDate"
 > = {
-  hardfork: HardforkName.PRAGUE,
+  hardfork: HardforkName.OSAKA,
   blockGasLimit: 30_000_000,
```
Amxx, line 40:
```suggestion
  blockGasLimit: 60_000_000,
```

In Osaka, the max gas per block has been raised from 30m to 60m.
- [EIP-7935: Set default gas limit to 60M](https://eips.ethereum.org/EIPS/eip-7935) (that is included in the "other EIPs" section of [EIP-7607](https://eips.ethereum.org/EIPS/eip-7607))
- [example of post osaka "big block"](https://etherscan.io/block/23967189)

kanej: Thanks @Amxx, that is an excellent point. I have opened a new issue for this here: https://github.com/NomicFoundation/hardhat/issues/7759

For the moment I want to minimize changes to get the Osaka as default hardfork release out.

Amxx: LEt me know when this is something I could create a PR for :)

pcaversaccio:
@kanej the PR doesn't fix it for me. Uncomment the following line in my repo here: https://github.com/pcaversaccio/batch-distributor/blob/de1f77b8db45b82b632d9d7f8ecd9c578af0b061/hardhat.config.ts#L103

and run `pnpm test` and you will get:

```
  1) Distributor Contract
       "before each" hook for "Transfers ETH to the given address":
     ProviderError: Transaction gas limit is 30000000 and exceeds transaction gas cap of 16777216
```

PR #7757, opened 2025-12-07
Improve HHE909 error message to show incompatible files Fixes #7748

### LINKED ISSUE #7748: Improve error message when no solc version is compatible with a file
### Describe the issue

When a file can't be compiled because no solc version in the config is compatible with its version pragma and those of its dependencies, you get an error like this:

```
Error HHE909: Failed to create compilation job for file "<file>" using the build profile "default".

No solc version enabled in this profile is compatible with this file and all of its dependencies.
```

This is fine when `<file>` has the problematic pragma, but it's not as helpful if the problem is in one of its dependencies. The error would be much better if it included more information about what exactly is going on, because it can happen that `<file>` has a big dependency tree and the incompatible pragma is in some random file down that tree.

### Reproduction steps

These reproduction steps are contrived, but the idea is to have a file that depends on another file, and where only the dependent can't be compiled. This illustrates that the source of the problem (the pragma of the dependency) cannot be inferred from the error.

```solidity
// IFoo.sol
pragma solidity 0.8.28;

interface IFoo {}
```

```solidity
// Foo.sol
pragma solidity ^0.8.28;

import { IFoo } from "./IFoo.sol";

contract Foo is IFoo {}
```

```typescript
// hardhat.config.ts
import { configVariable, defineConfig } from "hardhat/config";

export default defineConfig({
  solidity: {
    compilers: [
      {
        version: "0.8.28",
      },
      {
        version: "0.8.29",
      },
    ],
    overrides: {
      "contracts/Foo.sol": {
        version: "0.8.29",
      },
    },
  },
});
```

## Other comments

Some things worth noting:

- Since this is an error condition, performance doesn't matter that much. We can do whatever we need to build the best possible error.
- We should avoid printing a huge error if there are a lot of incompatible files in the dependency tree. We could show only the first clearly incompatible file, or just the first N and mention that there are M more. We could also always print the whole list in the debug output, so that it's available if someone needs it.

## Priority

I'm marking this as P1, because this happened to me during a migration. Since our compilation pipeline was significantly rewritten, errors related to it are likely, and they should be as helpful as possible.

sanskar717:
## Description

Improves the HHE909 error message to show which files have incompatible Solidity version requirements, making it easier to diagnose version compatibility issues in projects with large dependency trees.

## Problem

When a file can't be compiled due to version pragma incompatibilities deep in its dependency tree, the error message only says "No solc version enabled in this profile is compatible with this file and all of its dependencies" - without telling you *which* files are actually causing the problem.

This makes debugging extremely difficult, especially in projects with large dependency trees where the incompatible pragma could be buried several levels down.

## Solution

The improved error message now displays:
- ✅ Which specific files have incompatible version requirements
- ✅ The exact version pragma for each incompatible file
- ✅ Available compiler versions from the config
- ✅ A count of remaining files if there are more than 5 (to keep output manageable)

## Changes Made

- Added detailed file information to HHE909 error output
- Shows version pragma for each incompatible file
- Displays available compiler versions for reference
- Limits display to first 5 files with indicator of total count
- Improves developer experience when troubleshooting Solidity version issues

## Related Issue

Fixes #7748

## Testing

Tested with projects having:
- Single incompatible file in dependency tree
- Multiple incompatible files
- Large dependency trees (verified output is bounded to 5 files)

## Checklist

- [x] Changes follow the existing code style
- [x] Tests added/updated (if applicable)
- [x] Documentation updated (if applicable)
- [x] No breaking changes

### DIFF
--- a/v-next/example-project/hardhat.config.ts.save
+++ b/v-next/example-project/hardhat.config.ts.save
@@ -1,5 +1,3 @@
-port { defineConfig } from "hardhat/config";
-
 import { HardhatPluginError } from "hardhat/plugins";
 
 import util from "node:util";
--- a/v-next/example-project/hardhat.config.ts.save
+++ b/v-next/example-project/hardhat.config.ts.save
@@ -1,0 +1,178 @@
+port { defineConfig } from "hardhat/config";
+
+import { HardhatPluginError } from "hardhat/plugins";
+
+import util from "node:util";
+import { task, emptyTask, globalOption, configVariable } from "hardhat/config";
+import HardhatNodeTestRunner from "@nomicfoundation/hardhat-node-test-runner";
+import HardhatMochaTestRunner from "@nomicfoundation/hardhat-mocha";
+import HardhatKeystore from "@nomicfoundation/hardhat-keystore";
+import HardhatViem from "@nomicfoundation/hardhat-viem";
+import HardhatViemAssertions import hardhatNetworkHelpersPlugin from "@nomicfoundation/hardhat-network-helpers";
+import hardhatEthersPlugin from "@nomicfoundation/hardhat-ethers";
+import hardhatChaiMatchersPlugin from "@nomicfoundation/hardhat-ethers-chai-matchers";
+import hardhatTypechain from "@nomicfoundation/hardhat-typechain";
+import hardhatIgnitionViem from "@nomicfoundation/hardhat-ignition-viem";
+import hardhatVerify from "@nomicfoundation/hardhat-verify";
+import hardhatLedger from "@nomicfoundation/hardhat-ledger";
+import { ArgumentType } from "hardhat/types/arguments";
+
+util.inspect.defaultOptions.depth = null;
+
+const exampleEmptyTask = emptyTask("empty", "An example empty task").build();
+
+const exampleEmptySubtask = task(["empty", "task"])
+  .setDescription("An example empty subtask task")
+  .setAction(async () => ({
+    default: async (_, _hre) => {
+      console.log("empty task");
+    },
+  }))
+  .build();
+
+const exampleTaskOverride = task("example2")
+  .setAction(async () => ({
+    default: async (_, _hre) => {
+      console.log("from an override");
+    },
+  }))
+  .setDescription("An example task")
+  .addVariadicArgument({
+    name: "testFiles",
+    description: "An optional list of files to test",
+    defaultValue: [],
+  })
+  .addFlag({
+    name: "noCompile",
+    description: "Don't compile before running this task",
+  })
+  .addFlag({
+    name: "parallel",
+    description: "Run tests in parallel",
+  })
+  .addFlag({
+    name: "bail",
+    description: "Stop running tests after the first test failure",
+  })
+  .addOption({
+    name: "grep",
+    description: "Only run tests matching the given string or regexp",
+    type: ArgumentType.STRING_WITHOUT_DEFAULT,
+    defaultValue: undefined,
+  })
+  .build();
+
+const greeting = task("hello", "Print a greeting")
+  .addOption({
+    name: "greeting",
+    description: "The greeting to print",
+    defaultValue: "Hello, World!",
+  })
+  .addOption({
+    name: "programmatic",
+    description: "An example to show a hidden option",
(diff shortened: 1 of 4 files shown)


sanskar717:
## Summary

This PR improves the HHE909 error message by showing which specific files have incompatible Solidity version requirements, making debugging much easier.

## Testing Evidence

Here's the improved error output in action:

[
image
]

### Before vs After

**Before:** ❌ Only showed "No solc version enabled" without details
**After:** ✅ Shows exactly which file (`./contracts/IFoo.sol`) requires what version (`0.8.28`)

### Why This Helps

- Developers immediately see the problematic file
- No need to manually trace dependency trees
- Available versions clearly listed for reference
- Perfect for large projects with complex dependencies

## Testing Verified

- ✅ Single incompatible file in dependency tree
- ✅ Multiple incompatible files  
- ✅ Large dependency trees (output capped to 5 files)

v-next/example-project/hardhat.config.ts.save
```diff
@@ -0,0 +1,178 @@
+port { defineConfig } from "hardhat/config";
```
kanej: I think this file was included by accident.

kanej:
I was able to run the changes locally, based on the repro steps from #7748:

image

v-next/example-project/hardhat.config.ts.save
sanskar717: @kanej Yes, you're correct - that was included by mistake. Should I remove it from this PR?

kanej: Yes, I don't seem to have push permissions to your branch.

kanej: Can you add a changeset as well?

sanskar717: @kanej Sure, I'll add a changeset. Just to confirm - should I run `npx changeset` in the root directory and document the improved error message changes?

kanej: That would be great, thanks!

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
```diff
@@ -154,12 +154,66 @@ export class SolcConfigSelector {
       }
     }
 
+    //  Collect incompatible files information
+    const incompatibleFiles: Array<{ path: string; pragma: string }> = [];
+
+    // check root file
+    if (maxSatisfying(compilerVersions, rootVersionRange) == null) {
```
kanej:
```suggestion
    if (maxSatisfying(compilerVersions, rootVersionRange) === null) {
```
This is to satisfy our linter.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
```diff
@@ -154,12 +154,66 @@ export class SolcConfigSelector {
...
+
+    // Check all dependencies
+    for (const transitiveDependency of this.#getTransitiveDependencies(
+      root,
+      dependencyGraph,
+    )) {
+      const depVersionRange = transitiveDependency.versionPragmasPath
+        .map((pragmas) => pragmas.join(" "))
+        .join(" ");
+      if (maxSatisfying(compilerVersions, depVersionRange) == null) {
```
kanej:
```suggestion
      if (maxSatisfying(compilerVersions, depVersionRange) === null) {
```
This is to satisfy our liner.

v-next/example-project/hardhat.config.ts.save
sanskar717: @kanej Changeset added! Thank you for your guidance and patience throughout this process. Looking forward to your review!

sanskar717: @kanej Hi! Just a quick follow-up — would you mind taking another look at this PR when you have time?
If everything looks correct and helpful on your end, I’d really appreciate a merge. Thanks again for your guidance!

alcuadrado:
Hey, sorry we didn't get this past the finish line, @sanskar717. Unfortunately we were also working on a slightly different approach, which you can take a look at https://github.com/NomicFoundation/hardhat/pull/7988

PR #7767, opened 2025-12-09
Coverage fix

### LINKED ISSUE #7632: Code coverage numbers do not match test execution number
### Version of Hardhat

3.0.9

### What happened?

Described in #7629

Described in #7629

### Search terms

_No response_

ChristopherDedominici:
Fixes [https://github.com/NomicFoundation/hardhat/issues/7632](https://github.com/NomicFoundation/hardhat/issues/7632)

Related PR (on top of this one): https://github.com/NomicFoundation/hardhat/pull/7787
This PR adds a HTML report.

Not addressed in this PR, but this bug affects code coverage: https://github.com/NomicFoundation/hardhat/issues/7788

### Notes for the review

The PR is quite long, but around 40 files are related to tests, and the source code changes are limited to a few files.

#### How to run
In the `example-project` just run `pnpm hardhat test solidity --coverage`.
Just modify the `.sol` file to add an additional scenario and test different coverage cases.

#### Keep in mind
Currently, the counters for how many times a statement is executed are not implemented in EDR, so the only information available is whether a statement was executed, not how many times it was executed. Also, branch coverage is not available. In addition, partially executed lines (for example, ternary operators) cannot be determined, as this information is missing in EDR, since only whole lines can be registered as executed or not.
This can be addressed in later PR.

#### How the tests are organized

There are several `fixture projects`, each containing a Solidity contract with the code from which the coverage will be calculated. The coverage result is then compared against the expected result, which is stored in the same folder as each solidity contract.


v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -2,49 +2,54 @@ import type {
...
 
 /**
  * @private exposed for testing purposes only
  */
+export interface FileReport {
+  // NOTE: currently, the counters for how many times a statement is executed are not implemented in EDR,
+  // so the only information available is whether a statement was executed, not how many times it was executed.
+  // Also, branch coverage is not available.
+  // In addition, partially executed lines (for example, ternary operators) cannot be determined, as this information is missing in EDR,
+  // since only whole lines can be registered as executed or not.
```
Wodann:
```suggestion
  // since only whole statements can be registered as executed or not.
```

Suggestion: EDR implements instrumentation for statement-coverage. If a line contains multiple statements, EDR would be able to report it

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -153,138 +158,83 @@ export class CoverageManagerImplementation implements CoverageManager {
...
-
-          const branchExecutionCount =
-            branchExecutionCounts.get([line, tag]) ?? 0;
-          branchExecutionCounts.set(
-            [line, tag],
-            branchExecutionCount + executionCount,
-          );
+      // Create a map that tracks how many times each line was executed.
+      // Map: line number -> execution count
+      // Currently, from EDR we only know whether a line was executed or not,
```
Wodann:
```suggestion
      // Currently, from EDR we only know whether a statement was executed or not,
```
Suggestion: EDR generates statement-based coverage instrumentation

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,349 @@
...
+// pragma solidity ^0.8.0;
+function createMarkedFile(
+  fileContent: string,
+  minCharI: number,
+  maxCharI: number,
+  rawNotExecutedStatements: Statement[],
+): Array<null | boolean> {
+  const markedFile: Array<null | boolean> = new Array(fileContent.length).fill(
+    null,
+  );
```
Wodann: Question: Do you need to be able to represent `null`, or could you get away by defaulting to `true` or `false` and the algorithm would still work?

I ask because depending on how smart the TS compiler and JS interpreter is, this could affect the type's size:

`null | boolean`, naively requires a flag to mark which type and then the largest size of the two types: 1 byte + 1 byte 2 bytes
`boolean`: 1 byte

This means that - given a compiler that isn't allowed a smart optimisation - using `null | boolean` uses twice as much memory as `boolean`. This does not just impact memory usage but also runtime performance due to data locality (L1/L2/L3 caches will not be able to fit all data to be able to quickly provide the CPU with the data the algorithm needs)

If you do need three possible states, implementing an `enum` with three variants might be worth it.

If you're noticing performance slowdowns for large Solidity files and this of interest to you, I'd recommend checking out the memory sizes of an object using something like [this](https://stackoverflow.com/a/12408593) before deciding on a course of action.

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,349 @@
...
+    }
+  }
+
+  return markedFile;
+}
+
+// The following characters are not relevant for the code coverage, so they must be ignored
+// when marking characters as executed (true) or not executed (false)
+function charMustBeIgnored(c: string): boolean {
+  return /\s/.test(c) || c === "{" || c === "}";
```
Wodann: Question: What does `/\s/.test(c)` do?

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,349 @@
...
+// ->
+// [
+//   { start: 12, end: 20, executed: true  }
+// ]
+function getProcessedExecutedStatements(markedFile: Array<boolean | null>): {
+  executed: ReportCoverageStatement[];
+  notExecuted: ReportCoverageStatement[];
+} {
+  const executed = generateProcessedStatements(markedFile, true);
+  const notExecuted = generateProcessedStatements(markedFile, false);
```
Wodann: Suggestion(non-blocking): This could be executed as a single pass through the `markedFile` adding ranges to an array of `executed` or  `notExecuted` depending on the value of `markedFile[i]`.

The main reason for doing this would be performance. Iterating of a large range twice would require loading the memory twice.

Wodann (commented):
Nice algorithm! Nothing much to add in that regard.

I did leave some comments exploring potential performance considerations that might be of use if you're seeing slowdowns for large Solidity files.

These relate to expected real-world performance - due to the way computer's are architected - rather than algorithmic performance.

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: Oh wow, very interesting point! Thanks for the detailed explanation 😎
I will take a look and give a more precise answer, but I think it is needed. `null` is used to mark something that should be ignored in code coverage, so it is neither marked as green (executed, `true`) nor as not executed (`false`).
For example: comments.
But again, let me double check to confirm 100%

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: It checks if a char is a whitespace char.
Examples:
* space `" "`
* tab `\t`
* newline `\n`
* carriage return `\r`

All these chars, plus `{` and `}` can be ignored during the code coverage. Currently they are marked as `null`. See other comment to understand how `null` is used in this coverage logic

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: Example of graphical representation:
image

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
Wodann: Thanks for the explanation!

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
Wodann: As I mentioned:
1. The compiler might be able to optimise this to an enum with 3 values
2. If the compiler doesn't optimise, you could introduce an enum yourself with three variants: `NotTracked`, `Hit`, `NotHit`

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/solidity.ts
```diff
@@ -124,9 +122,20 @@ export default async (): Promise<Partial<SolidityHooks>> => ({
...
+
+      if ((await exists(coverageSolPath)) === false) {
+        // Fallback for when running tests
+        coverageSolPath = path.join(
+          import.meta.dirname,
+          "../../../../../coverage.sol",
+        );
+      }
+
+      const content = await readUtf8File(coverageSolPath);
```
schaable: Minor suggestion: the relative `..` path + fallback looks a bit fragile (especially since we have one path for the built output and another for tests).
Could we use `findClosestPackageRoot` from `hardhat-utils` instead? Something like:
```
const packageRoot = await findClosestPackageRoot(import.meta.url);
const coverageSolPath = path.join(packageRoot, "coverage.sol");
```
That should be more robust and avoids the dual-path logic. Wdyt?

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -153,138 +158,83 @@ export class CoverageManagerImplementation implements CoverageManager {
...
-  public getReport(): Report {
+  public async getReport(): Promise<Report> {
     const report: Report = {};
 
-    const relativePaths = this.metadata.map(({ relativePath }) => relativePath);
-
-    const allStatements = this.metadata;
+    const seenMetadata = new Set<string>();
+    const uniqueMetadata: CoverageMetadata = [];
+    for (const m of this.metadata) {
```
schaable: Suggestion (not a blocker, can be addressed later): since `metadata` is deduplicated on every report generation and `data` is converted to a `Set` for lookups, using these structures directly could simplify things:

```
public metadata = new Map<string, Statement>();
public data = new Set<Tag>();

public async getReport(): Promise<Report> {
  const uniqueMetadata = Array.from(this.metadata.values());
  ...
}
```

Edit: If we worry about performance, this could be even better:

```
public metadata = new Map<string, Map<string, Statement>>();
// Map<relativePath, Map<compositeKey, Statement>>

public async getReport(): Promise<Report> {
  ...
  const fileRelativePaths = this.metadata.keys();
  for (const fileRelativePath of fileRelativePaths) {
    const statementsForFile = Array.from(
      this.metadata.get(fileRelativePath)!.values()
    );
  ...
}
```

This would eliminate the following from `getReport()`:
 - The deduplication loop.
 - The Set creation for file paths.
 - The filter operation for each file.

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -153,138 +158,83 @@ export class CoverageManagerImplementation implements CoverageManager {
...
+        ),
+      ]);
+
+      let executedStatementsCount = 0;
+      let unexecutedStatementsCount = 0;
+      for (const s of tagsForFile) {
+        if (allExecutedTags.has(s)) {
+          executedStatementsCount++;
+        } else {
+          unexecutedStatementsCount++;
```
schaable: This could be unified with the `executedTagsForFile` logic to avoid the `filter`.

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -153,138 +158,83 @@ export class CoverageManagerImplementation implements CoverageManager {
...
+      // Create a map that tracks how many times each line was executed.
+      // Map: line number -> execution count
+      // Currently, from EDR we only know whether a statement was executed or not,
+      // so the execution count is either 0 or 1.
+      const lineExecutionCounts = new Map<number, number>([
+        ...[...coverageInfo.lines.executed.keys()].map((k) => [k, 1] as const),
+        ...[...coverageInfo.lines.notExecuted.keys()].map(
+          (k) => [k, 0] as const,
+        ),
+      ]);
```
schaable: I don’t think you need to spread `keys()` here, since it’s already iterable. That said, this feels a bit compressed and hard to read. I think this is clearer and easier to follow:

```typescript
const lineExecutionCounts = new Map<number, number>();
coverageInfo.lines.executed.forEach((_, line) => lineExecutionCounts.set(line, 1));
coverageInfo.lines.notExecuted.forEach((_, line) => lineExecutionCounts.set(line, 0));
```

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -487,48 +433,44 @@ export class CoverageManagerImplementation implements CoverageManager {
...
           lineExecutionCounts.size === 0
             ? 0
             : (executedLinesCount * 100.0) / lineExecutionCounts.size;
         const statementCoverage =
-          tagExecutionCounts.size === 0
+          executedStatementsCount === 0
             ? 0
-            : (executedTagsCount * 100.0) / tagExecutionCounts.size;
+            : (executedStatementsCount * 100.0) /
+              (executedStatementsCount + unexecutedStatementsCount);
```
schaable, line 459: Are `unexecutedStatementsCount` only used to compute the total statement count? If so, could we use `allExecutedTags` instead? If you switch `this.data` to a `Set` as suggested in my other comment, it should be available in `formatMarkdownReport`.

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -153,138 +158,83 @@ export class CoverageManagerImplementation implements CoverageManager {
...
-      const executedTagsCount = tagExecutionCounts.size;
-      const executedLinesCount = lineExecutionCounts.size;
-      const executedBranchesCount = branchExecutionCounts.size;
+      const executedLinesCount = coverageInfo.lines.executed.size;
 
-      const partiallyExecutedLines = new Set<number>();
-      const unexecutedLines = new Set<number>();
+      const unexecutedLines = new Set([
+        ...coverageInfo.lines.notExecuted.keys(),
+      ]);
```
schaable: This information could be derived from the `lineExecutionCounts` map:

```
let executedLinesCount = 0;
const unexecutedLines = new Set<number>();

for (const [line, count] of lineExecutionCounts) {
  if (count === 0) {
    unexecutedLines.add(line);
  } else {
    executedLinesCount++;
  }
}
```

I haven’t looked closely at `getProcessedCoverageInfo` yet, but since its return value is later used to compute `lineExecutionCounts`, `executedLinesCount`, and `unexecutedLines`, it might make sense for it to return `lineExecutionCounts` directly and derive the other two from it as shown above.

schaable (commented):
I’ve made an initial pass and left some comments. I still need to review `getProcessedCoverageInfo` and the tests, but I wanted to leave some early feedback. Feel free to take a look at the comments and decide whether you want to apply any of the suggestions first. If you do, I’m happy to wait for the PR to be updated before continuing the review.

v-next/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/solidity.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/5872730ab8c4deeabad2599d5080d17d29ac0ad2

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/a23f904e9ea48dd7d67f61865b8d39223b08bee9

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
ChristopherDedominici: True, I just wanted to be more explicit and provide direct access to the info about not executed lines, instead of deriving the info. I'll wait for the review of `getProcessedCoverageInfo`, then we can sync on what do you think is best

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
ChristopherDedominici: As mentioned above, in this PR I would prefer not to modify `this.data`

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
ChristopherDedominici: I did it for metadata, but for data it affects the original logic. It requires a bit more investigation, so I'd prefer doing it in a separate PR. Commit: https://github.com/NomicFoundation/hardhat/pull/7767/commits/bab618eceb3273ad031aa06fa66d5c5a8a46173c

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
ChristopherDedominici: Fixed as consequence of the `metadata` commit: https://github.com/NomicFoundation/hardhat/commit/bab618eceb3273ad031aa06fa66d5c5a8a46173c

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,366 @@
...
+  notExecutedStatements: Statement[];
+} {
+  const hitTagsSet = new Set(hitTags);
+
+  const executedStatements = metadata.filter((node) =>
+    hitTagsSet.has(node.tag),
+  );
+  const notExecutedStatements = metadata.filter(
+    (node) => !hitTagsSet.has(node.tag),
+  );
```
schaable: Replace two loops with a single loop:

```suggestion
  const executedStatements: Statement[] = [];
  const notExecutedStatements: Statement[] = [];

  for (const node of metadata) {
    if (hitTagsSet.has(node.tag)) {
      executedStatements.push(node);
    } else {
      notExecutedStatements.push(node);
    }
  }
```

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,366 @@
...
+  }
+
+  return markedFile;
+}
+
+// The following characters are not relevant for the code coverage, so they must be ignored
+// when marking characters as executed (true) or not executed (false)
+function charMustBeIgnored(c: string): boolean {
+  return /\s/.test(c) || c === "{" || c === "}";
+}
```
schaable: Since this function is called for every character in the file (potentially twice per character), using direct character comparisons should be more performant than a regex. It also avoids creating a regex on each call, which adds extra overhead.

```suggestion
function charMustBeIgnored(c: string): boolean {
  return c === " " || c === "\t" || c === "\n" || c === "\r" || c === "{" || c === "}";
}
```

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,366 @@
...
+
+  markAsNonExecutableSpecialSubstrings(fileContent, markedFile);
+
+  // Enable this function while debugging to view the coverage for a file
+  // printFileCoverageForDebugging(fileContent, markedFile);
+
+  return {
+    statements: getProcessedExecutedStatements(markedFile),
+    lines: getLinesInfo(fileContent, markedFile),
+  };
```
schaable, line 59: I’d simplify the return value since we only use the line information for now. We can always add `statements` later if needed. Along the same lines, I’d rename the function to `getLineCoverageInfo`.
```suggestion
  return getLinesInfo(fileContent, markedFile)
```

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,366 @@
...
+    fileContent,
+    minCharI,
+    maxCharI,
+    rawStatements.notExecutedStatements,
+  );
+
+  markAsNonExecutableSpecialSubstrings(fileContent, markedFile);
+
+  // Enable this function while debugging to view the coverage for a file
+  // printFileCoverageForDebugging(fileContent, markedFile);
```
schaable: Let's remove this. The function docs already explain how to use it.

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,366 @@
...
+  );
+
+  const markedFile = createMarkedFile(
+    fileContent,
+    minCharI,
+    maxCharI,
+    rawStatements.notExecutedStatements,
+  );
+
+  markAsNonExecutableSpecialSubstrings(fileContent, markedFile);
```
schaable: These two functions are both building the marked file, so it doesn’t really make sense for them to be separate steps. I’d call `markAsNonExecutableSpecialSubstrings` from within `createMarkedFile`.

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -153,143 +182,63 @@ export class CoverageManagerImplementation implements CoverageManager {
...
-    }
+        const fileContent = await readUtf8File(
+          path.join(process.cwd(), fileRelativePath),
+        );
 
-    const uniqueExecutedTags = new Set(allExecutedTags);
-    const uniqueUnexecutedTags = Array.from(statementsByTag.keys()).filter(
-      (tag) => !uniqueExecutedTags.has(tag),
-    );
+        const tags: string[] = [];
```
schaable: Since this gets converted to a `Set` inside `getProcessedCoverageInfo` and isn’t used anywhere else, we could just use a Set from the start and avoid the extra conversion.

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -336,11 +280,12 @@ export class CoverageManagerImplementation implements CoverageManager {
...
-        lcov += `BRDA:${line},0,${tag},${executionCount === 0 ? "-" : executionCount}\n`;
-      }
-      lcov += `BRH:${executedBranchesCount}\n`;
-      lcov += `BRF:${branchExecutionCounts.size}\n`;
+      // TODO: currently EDR does not provide branch coverage information.
+      // for (const [[line, tag], executionCount] of branchExecutionCounts) {
+      //   lcov += `BRDA:${line},0,${tag},${executionCount === 0 ? "-" : executionCount}\n`;
+      // }
+      // lcov += `BRH:${executedBranchesCount}\n`;
+      // lcov += `BRF:${branchExecutionCounts.size}\n`;
```
schaable, line 288: Let’s just remove this. We can always recover it from the git history if needed.

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,366 @@
...
+    markedFile[i] = STATUS_EXECUTED;
+  }
+
+  for (const n of rawNotExecutedStatements) {
+    for (let i = n.startUtf16; i < n.endUtf16; i++) {
+      if (charMustBeIgnored(fileContent[i])) {
+        continue;
+      }
+
+      markedFile[i] = STATUS_NOT_EXECUTED;
```
schaable: We can improve performance a bit by skipping lines with `STATUS_IGNORED`, since `charMustBeIgnored` already walks the full range in the first loop.

```suggestion
      if (markedFile[i] === STATUS_EXECUTED) {
        markedFile[i] = STATUS_NOT_EXECUTED;
      }
```

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
```diff
@@ -0,0 +1,366 @@
...
+      executed: targetStatus === STATUS_EXECUTED,
+    });
+  }
+
+  return ranges;
+}
+
+// Return the executed and non executed lines based on the marked file.
+// Some lines are excluded from the execution count, for example, comments.
+function getLinesInfo(
```
schaable: There’s no need for an intermediate step here. We could build the maps directly and avoid the double loop. Also, there’s a bug: if the file doesn’t end with a newline, the last line is lost.

schaable (commented):
I’ve reviewed `v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts` and left some comments. If you want, we can sync tomorrow and go through them.

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/7fefeeee7b0717eea29a30246edce9f55de39f65

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/865bb275fc20967d9152faa0d6a330a89724ceff

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: Partially done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/70b31953cf4b4471b4d2e9a6ba5a0d5f3e214d2a

I have removed the statement results to optimize the computation, but I kept the function due to its complexity. It should be useful soon if we decide to further enhance the HTML report

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/ab7e65b7e09855f69d69ae050d60ac91252d5052

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/b7605ecb0fbe2516d07cdde29abc4a4d8db51e4d

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/b1f4e48a22c177c3f190bfc04708d6f9abac1a4a

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
ChristopherDedominici: I would keep it here, because if someone works on this file, they will instantly see the TODO. They probably will not think to check the git history, so this info might get lost.

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/966428512b4ef64fd9921e540188137baf475587

v-next/hardhat/src/internal/builtin-plugins/coverage/process-coverage.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7767/commits/11b0be7a53fbc507f37a22a916e05b7af42f4fdb

ChristopherDedominici:
@schaable I also cleaned the code:
- reduced line length for comments: https://github.com/NomicFoundation/hardhat/pull/7767/commits/8b1ecf699fc6f713c4baa10f6e7704391893444a
- renamed variables and functions: https://github.com/NomicFoundation/hardhat/pull/7767/commits/e8786b9fb480f4debce3f38df356e65fa8246e2f

v-next/hardhat/test/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -383,3 +368,89 @@ describe("CoverageManagerImplementation", () => {
...
+    it(testScenrario.description, async () => {
+      await hre.tasks.getTask(["test", "solidity"]).run({
+        noCompile: true,
+        testFiles: [testScenrario.testFilePath],
+      });
+
+      const res = await coverageManagerTmp.getReport();
+
+      assert.deepEqual(
+        res[testScenrario.sourceFilePath],
```
schaable: typo 😅

schaable (approved):
Left one tiny comment about a typo, but otherwise it looks good! Nicely done 💯

v-next/hardhat/test/internal/builtin-plugins/coverage/coverage-manager.ts
ChristopherDedominici: Done: https://github.com/NomicFoundation/hardhat/pull/7767/commits/b92f41907187110f537199f97d2cc1f979f8b5cb

PR #7768, opened 2025-12-09
Remove solidity timeout

### LINKED ISSUE #7750: Remove `timeout` option for solidity tests
### Describe the issue

This option is useless and misleading. It's a _whole run_ timeout, not a per-test timeout  like the one most test runners have.

Making it work in a per-test basis seems hard and unnecessary, so we should just remove it.

sanskar717:
## Description

This PR removes the misleading `timeout` option from the solidity test configuration in `hardhat.config.ts`.

## Problem

The `timeout` option was confusing because it applies to the **entire test run**, not to individual tests (per-test timeout) as most developers would expect from their experience with other test runners like Mocha or Jest.

## Solution

Removed the `timeout: 1000` configuration from the `solidity` test settings as:
- Making it work on a per-test basis would be complex and unnecessary
- The option in its current form provides little value and causes confusion
- Users can rely on their test runner's native timeout mechanisms instead

## Related Issue

Fixes #7750

---

### DIFF
--- a/v-next/example-project/hardhat.config.ts
+++ b/v-next/example-project/hardhat.config.ts
@@ -237,7 +237,6 @@ export default defineConfig({
       color: true,
     },
     solidity: {
-      timeout: 1000,
     },
   },
 });


v-next/example-project/hardhat.config.ts
```diff
@@ -237,7 +237,6 @@ export default defineConfig({
       color: true,
     },
     solidity: {
-      timeout: 1000,
```
kanej, line 240: This is a configuration in an example project. It will not change the behaviour of Hardhat itself.

kanej:
I am going to close this PR. The timeout removed here is in an example project, it does not remove the timeout for solidity tests.

sanskar717:
Hi @kanej,
Thank you for reviewing my PR. I apologize for the mistakes - this is my first time contributing to open source and I'm still learning the process. I understand now that this change was only in the example project and doesn't affect the actual Hardhat behavior.
I really appreciate your time and feedback. I'll be more careful to understand the codebase better before submitting PRs in the future.
Thanks again for your patience!

Issue #7761, opened 2025-12-08
Configured solc `0.8.31` downloads the pre-release version


pcaversaccio:
### Version of Hardhat

`2.27.1`

### What happened?

If you configure `version: "0.8.31",` in the hardhat config, it downloads the prelease version:

<img width="1040" height="185" alt="Image" src="https://github.com/user-attachments/assets/a431008f-9285-4fab-9b0b-8d72aac289f4" />

### Minimal reproduction steps

You can use my template here: https://github.com/pcaversaccio/hardhat-project-template-ts

set in the following line `version: "0.8.31",` https://github.com/pcaversaccio/hardhat-project-template-ts/blob/560edb6aa22bb0e7599020c3bd5ef192244b3426/hardhat.config.ts#L77

and invoke `pnpm test:hh`

### Search terms

0.8.31, solc

nebasuke:
@ChristopherDedominici 

This likely due to solc populating multiple downloads for a single version. see https://binaries.soliditylang.org/linux-amd64/list.json and check for 0.8.31. 

EDR fixed this by bumping `svm-rs`, see https://github.com/NomicFoundation/edr/pull/1209 (and the fixed issue https://github.com/alloy-rs/svm-rs/issues/180).

pcaversaccio:
@ChristopherDedominici can we please get this fixed sooner than later because it's a hard blocker for me right now.

pcaversaccio:
https://binaries.soliditylang.org/linux-amd64/list.json

<img width="1033" height="550" alt="Image" src="https://github.com/user-attachments/assets/5152cd07-c3c1-4bcf-b634-c9385217097b" />

so the `version` isn't enough anymore, needs additional logic with `prerelease`.

michalbrabec:
Hi folks, the situation is even more complex, because solc was not available for ARM64 Linux until `0.8.31` and we had to produce our own version. We already have an issue open for testing and integration the official version as #7752 and we will keep the issue in mind when resolving that one.

At this point, we did not consider pre-releases because they did not appear among the solc releases.

The solc download logic is in [downloader.ts](https://github.com/NomicFoundation/hardhat/blob/65de9a4eeacf6c259cfd29e361a662dc67641f77/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts#L344).
[...]

PR #7774, opened 2025-12-09
Fix solc download when there are prereleases in the list

fvictorio:
Closes https://github.com/NomicFoundation/hardhat/issues/7761.

### DIFF
--- a/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
@@ -469,6 +469,14 @@ describe(
       });
 
       it("should work if there are prereleases in the list", async () => {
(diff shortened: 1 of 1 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -273,7 +274,9 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
 
     const list = await this.#readCompilerList(listPath);
 
-    const build = list.builds.find((b) => b.version === version);
+    const build = list.builds.find(
```
michalbrabec, line 277: Are there any changes needed for ARM Linux version?
We distribute our own build as it was not available officially until now.

michalbrabec (approved):

PR #7786, opened 2025-12-11
replace memory-safe-assembly natspec comments with memory-safe assemb…

ItsNickBarry:
This should fix a compiler warning about the deprecation of the "memory-safe-assembly" comment.

### DIFF
--- a/v-next/hardhat/console.sol
+++ b/v-next/hardhat/console.sol
@@ -7,8 +7,7 @@ library console {
 
     function _sendLogPayloadImplementation(bytes memory payload) internal view {
         address consoleAddress = CONSOLE_ADDRESS;
-        /// @solidity memory-safe-assembly
-        assembly {
+        assembly ('memory-safe') {
             pop(
                 staticcall(
                     gas(),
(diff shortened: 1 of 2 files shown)


kanej:
Hey @ItsNickBarry, thanks for this.
This warning is coming from using solc `0.8.31` with its deprecation notice for `Removal of memory-safe-assembly Special Comment`: https://www.soliditylang.org/blog/2025/12/03/solidity-0.8.31-release-announcement/

I think it would be good to fix this. The `assembly ("memory-safe")` syntax was introduced in `0.8.13`, I am worried we will break older versions of `solc` with this change.

Maybe we can use different versions or modify based on solc version?

ItsNickBarry:
That's inconvenient.  Best approach is probably to close this PR and suppress the warning somehow.  I think some compiler warnings from imports should be suppressed by default.

v-next/hardhat/console.sol
```diff
@@ -7,8 +7,7 @@ library console {
 
     function _sendLogPayloadImplementation(bytes memory payload) internal view {
         address consoleAddress = CONSOLE_ADDRESS;
-        /// @solidity memory-safe-assembly
-        assembly {
+        assembly ('memory-safe') {
```
alcuadrado, line 10: I think this changes make the libraries compatible with less versions of solc.

I guess instead of doing that, we can filter out these warnings.

michalbrabec:
Opened the issue #7842 to implement the suppression of the warning.
The warning displayed in case both the assembly annotation and the natspec comment will not be impacted
@ItsNickBarry please feel free to add more details to the new issue as appropriate.

PR #7787, opened 2025-12-11
[Code coverage] Add HTML coverage report

ChristopherDedominici:
Related to: https://github.com/NomicFoundation/hardhat/pull/7767

Docs updated in this PR: https://github.com/NomicFoundation/hardhat-website/pull/198

#### How to run
In the `example-project` just run `pnpm hardhat test solidity --coverage`.
Just modify the `.sol` file to add an additional scenario and test different coverage cases.

The HTML report will be generated in the `example-project` in the file at the path: `./coverage/html/index.html`

### DIFF
--- a/.changeset/strange-games-dance.md
+++ b/.changeset/strange-games-dance.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Added HTML coverage report for solidity tests ([7787](https://github.com/NomicFoundation/hardhat/pull/7787)).
--- a/v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
@@ -21,6 +21,7 @@ import chalk from "chalk";
 import debug from "debug";
 
 import { getProcessedCoverageInfo } from "./process-coverage.js";
+import { generateHtmlReport } from "./reports/html.js";
 
 const log = debug("hardhat:core:coverage:coverage-manager");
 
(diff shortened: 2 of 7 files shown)


.changeset/strange-games-dance.md
michalbrabec, line 1: Do we need a changeset for the vendored package as well?
It is not yet published and the changesets drive publishing

v-next/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -124,6 +125,10 @@ export class CoverageManagerImplementation implements CoverageManager {
     await writeUtf8File(lcovReportPath, lcovReport);
     log(`Saved lcov report to ${lcovReportPath}`);
 
+    const htmlReportPath = path.join(this.#coveragePath, "html");
```
michalbrabec, line 128: Do we always generate the HTML report or is there an option that enables / disables it?
I could not find any and the design doc does not specify

ChristopherDedominici: Currently yes, it is always generated like all the other coverage reports. See also this comment in the design doc (internal link): https://www.notion.so/Code-coverage-HTML-report-286578cdeaf580bcad22db58ce54fb47?d=287578cdeaf580759bd9001c093dc31c&source=copy_link#286578cdeaf58000858ffd467737729e

.changeset/strange-games-dance.md
ChristopherDedominici: Probably yes, good point

schaable (approved):
LGTM!

PR #7791, opened 2025-12-12
Add mocha results to test summary numbers

zoeyTM:
Screen Shot 2025-12-12 at 11 28 00 AM

### DIFF
--- a/v-next/hardhat-ethers-chai-matchers/test/index.ts
+++ b/v-next/hardhat-ethers-chai-matchers/test/index.ts
@@ -22,6 +22,6 @@ describe("hardhat-ethers-chai-matchers plugin correctly initialized", () => {
       noCompile: true,
     });
 
-    assert.equal(result, 0);
+    assert.deepEqual(result, { failed: 0, passed: 1 });
   });
 });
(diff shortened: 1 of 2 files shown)


v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -52,20 +52,16 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
   const testSummaries: Record<
     string,
     {
-      failed: number;
-      passed: number;
-      skipped: number;
-      todo: number;
-      failureOutput: string;
+      failed?: number;
```
ChristopherDedominici, line 55: Just a question to ensure I understand correctly: why is it necessary to add `?`. Is it possible that some of these properties could be undefined in certain situations, for example, when there are no failed tests but only passed ones? If so, why use undefined instead of 0? Is it simply a convention?

zoeyTM: Mocha only has the `failed` and `passed` fields for its summary, so I think I just added the optional to them all to handle them all the same way for consistency.

PR #7792, opened 2025-12-12
Write snapshots to .gas-snapshot file with --snapshot flag

### LINKED ISSUE #7770: Write snapshots to `.gas-snapshot` file with `--snapshot` flag
### Describe the issue

**Tasks:**
- Parse gas usage from test results (unit tests: single value, fuzz tests: average & median, invariant: skip)
- Generate `.gas-snapshot` file
- Add `--snapshot` flag to `test` and `test:solidity` tasks
- Only write file when tests pass
- Handle new and removed tests

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7770


v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-analytics-manager.ts
```diff
@@ -140,8 +140,8 @@ export class GasAnalyticsManagerImplementation implements GasAnalyticsManager {
 
         const isOverloaded = overloadedFnNames.has(functionName);
         const stats: GasStats = {
-          min: Math.min(...gasValues),
-          max: Math.max(...gasValues),
+          min: gasValues.reduce((a, b) => Math.min(a, b), Infinity),
+          max: gasValues.reduce((a, b) => Math.max(a, b), -Infinity),
```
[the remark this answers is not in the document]
schaable: This change was done because the spread operator was causing an overflow with fuzz tests.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -0,0 +1,32 @@
...
+interface GasAnalyticsTestActionArguments {
+  snapshot: boolean;
+}
+
+const runSolidityTests: TaskOverrideActionFunction<
+  GasAnalyticsTestActionArguments
+> = async (args, hre, runSuper) => {
+  const suiteResults: SuiteResult[] = await runSuper(args);
+
+  if (args.snapshot && process.exitCode !== 1) {
```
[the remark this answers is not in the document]
schaable: This will make eslint complain with `Unexpected value in conditional. A boolean expression is required.`. Additionally, we only set the `exitCode` to 1 when tests fail, so there's no need to check other values.

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -88,12 +88,18 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
       args.verbosity = verbosity;
     }
 
+    for (const [key, value] of Object.entries(otherArgs)) {
+      if (subtask.options.has(key)) {
+        args[key] = value;
+      }
+    }
```
kanej, line 95: Can I confirm I follow this - we will delegate all arguments passed to the top level test down to each sub-test task - and we rely on each test plugin to add the args to the top level test task?

schaable: That’s correct. The test task delegates all options to its subtasks. If a subtask uses an option, it should declare it in its task definition. In this particular case, that’s done via an override in `v-next/hardhat/src/internal/builtin-plugins/gas-analytics/index.ts`

schaable: A similar approach is used in the `hardhat-verify` plugin, where the `verify` task accepts arguments that are specific to its subtasks (for example, `creationTxHash`, which is only used by Sourcify), although the way arguments are forwarded is a bit different.

v-next/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -113,11 +113,13 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
...
-    skipped: number;
-    todo: number;
-    failureOutput: string;
+    testSummary: {
+      failed: number;
+      passed: number;
+      skipped: number;
+      todo: number;
+      failureOutput: string;
+    };
```
kanej, line 122: This is a change in the API. It makes sense and lets test tasks return additional scoped information.
To check, this change is so we can also have `suiteResults` which is how the gas information is being passed?

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/index.ts
kanej, line 16: Is this a still be implemented feature? Or a hangover from a previous iteration?

v-next/hardhat-node-test-runner/src/task-action.ts
schaable: That’s correct. To give you more context, the task initially had no return value, but we later added the test summary information as part of this [issue](https://github.com/NomicFoundation/hardhat/pull/7514), which was worked on in parallel with this one.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/index.ts
kanej: This is the gas statistics feature.

v-next/hardhat/test/internal/builtin-plugins/gas-analytics/gas-snapshots.ts
```diff
@@ -0,0 +1,430 @@
...
+ContractA:testA2 (gas: 12000)`;
+      assert.equal(result, expected);
+    });
+  });
+
+  describe("saveGasFunctionSnapshots", () => {
+    let tmpDir: string;
+
+    before(async () => {
+      tmpDir = await mkdtemp("gas-snapshots-test-");
```
kanej, line 310:
```suggestion
      tmpDir = await mkdtemp("gas-snapshots-test");
```
Is the dash at the end needed?

schaable: I think that, while it’s technically not required, we add it because the argument is used as a prefix. You can see this in the implementation:

```
export async function mkdtemp(prefix: string): Promise<string> {
  try {
    return await getRealPath(
      await fsPromises.mkdtemp(path.join(tmpdir(), prefix)),
    );
  } catch (e) {
    ensureNodeErrnoExceptionError(e);
    throw new FileSystemAccessError(e.message, e);
  }
}
```

`fsPromises.mkdtemp()` appends six random characters to the end, so if the prefix ends with a dash, the resulting path will look like `/tmp/prefix-XXXXXX`.

schaable: I’m fine with normalizing this and adding the dash inside the function, but that’s probably better handled in a separate PR.

kanej: No need. That all makes sense, I just didn't realise the mkdtemp api appended after the prefix.

schaable: I actually had to double-check. We should mention it in the jsdoc 😅

kanej (commented):
I have completed a round of manual tests. This is looking really good.

## Manual tests

- [x] Check that standard test comes through with `gas`
- [x] Check that invariant tests are excluded.
- [x] Check that invariant test comes through with runs/average/median in right order
- [x] Check duplicate contract names in different files to force FQN
- [x] Check `--snapshot` is delegated when run with `npx hardhat test --snapshot`
- [ ] Rerunning the snapshot is deterministic
- [ ] hardhat test --snapshot outputs `Gas snapshots written successfully`

There were two failures:

1. based on the spec `Gas snapshots written successfully` should be output on writing the `.gas-snapshots` file.
2. Fuzz tests are non-deterministic. A repeated run will produce different values for median and average gas usage. This is different from foundry.

kanej:
We discussed out of band:

1. `Gas snapshots written successfully` is addressed in the follow up PR
2. Fuzz tests will vary, but you can pin with a seed.

I think we should discuss having a default seed so that the `.gas-snapshot` is deterministic, but I consider this PR good to go.

kanej (approved):

PR #7796, opened 2025-12-14
fix: remove redundant id assignment in getJsonRpcRequest

Bilogweb3:
Remove dead code in `getJsonRpcRequest` function. The `id` parameter has type `number | string` (not optional), so the check `if (id !== undefined)` is always true. Additionally, `id` was already assigned during object creation, making the subsequent conditional assignment redundant.


michalbrabec:
Hi,

PR #7804, opened 2025-12-17
Use block annotation instead of natspec comment for memory safe assembly

arr00:

---

This PR moves away from the deprecated natspec comment flagging a block of assembly memory safe. This causes a warning due to deprecation when compiling with solc 0.8.31.


kanej:
Hey @arr00, thanks for this. I am going to close this PR in favour of: https://github.com/NomicFoundation/hardhat/pull/7786

However, we are not sure we can make the change being suggested without breaking compilation under older Solidity versions.

PR #7805, opened 2025-12-17
Allow overriding the type of the default network

### LINKED ISSUE #7795: Configuration for default network is ignored
### Describe the issue

Using a config that sets the default network to point to `localhost:8545` doesn't work. Take this config:

```js

export default defineConfig({
  solidity: "0.8.31",
  networks: {
    default: {
      type: "http",
      url: "http://localhost:8545"
    },
  },
});
```

and run `hh test nodejs`. It should fail (assuming no node is running), but it works. Worse, running `hh test nodejs --network default` also doesn't seem to work as expected. Renaming the network to `default2` and running `hh test nodejs --network default2` does produce the expected behavior (tests fail because it cannot connect to the network).

fvictorio:
This fixes https://github.com/NomicFoundation/hardhat/issues/7795, but I'm not fully convinced about the change.

The problem is that we are always setting the type of default network to be a simulated network, and we also set some other default values. This makes sense, but it does seem to assume that the network is not of type `http`. My concern is that a user's settings could interact in the wrong way with those default values.

On the other hand, without this change there's no way to have an http network as the default network, something that was possible in HH2.

One alternative here is to re-introduce the `defaultNetwork` config, and default it to "default" (yes).

Another alternative is to only use those default values if the user doesn't change the type, but that starts feeling a bit too complex.

### DIFF
--- a/v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
+++ b/v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
@@ -41,6 +41,43 @@ describe("network-manager/hook-handlers/config", () => {
       });
     });
 
+    it("should extend the localhost network when its type is undefined", async () => {
+      const config = {
+        networks: {
+          localhost: {},
+        },
+      };
+      const next = async (nextConfig: HardhatUserConfig) => nextConfig;
+
+      /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      -- testing incomplete config for js users */
+      const extendedConfig = await extendUserConfig(config as any, next);
+      assert.deepEqual(extendedConfig.networks?.localhost, {
+        type: "http",
+        url: "http://localhost:8545",
+      });
+    });
+
+    it("should extend the default network when its type is undefined", async () => {
+      const config = {
+        networks: {
+          default: {},
+        },
+      };
+      const next = async (nextConfig: HardhatUserConfig) => nextConfig;
+
+      /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      -- testing incomplete config for js users */
+      const extendedConfig = await extendUserConfig(config as any, next);
+      assert.deepEqual(extendedConfig.networks?.default, {
+        chainId: 31337,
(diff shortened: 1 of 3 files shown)


kanej:
We should double check running tests against a `npx hardhat node` instance, and add a changeset.
But in principle we are happy with this approach to let users set default to an `http` network.

michalbrabec (commented):
Overriding the default network creates the need to setup a new network for `npx hardhat node` and then used it with the `--network` parameter. 
https://github.com/NomicFoundation/hardhat/issues/7795#issuecomment-3777253615

alcuadrado:
I went for an option where we only set the defaults if the network is `undefined`, it has no `type`, or the `type` matches the expected one. Otherwise I just forward whatever the user provided.

I also implemented the same logic for `localhost`, so that no network in this PR has any unexpected behavior.

v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -79,11 +116,11 @@ describe("network-manager/hook-handlers/config", () => {
       });
     });
 
-    it("should not allow overriding the type of the localhost network", async () => {
+    it("should allow overriding the type of the localhost network", async () => {
```
schaable, line 119: the type assertion is no longer needed

v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -93,10 +130,30 @@ describe("network-manager/hook-handlers/config", () => {
       -- testing invalid network type for js users */
       const extendedConfig = await extendUserConfig(config as any, next);
       assert.deepEqual(extendedConfig.networks?.localhost, {
-        url: "http://localhost:8545",
-        type: "http",
+        type: "edr-simulated",
       });
     });
+
+    it("should allow overriding the type of the default network", async () => {
```
schaable, line 135: the type assertion is not needed

schaable (approved):
LGTM

kanej:

PR #7808, opened 2025-12-18
test: update test to reflect viem change

kanej:
`viem` added the hyper liquid entry into its chain. The chain id is 999 which broke an existing test. I have bumped the test expectations.
I have updated the pnpm-lock file as well, attempting to keep it constrained to the latest version of `viem` and its deps.


packages/hardhat-viem/test/chains.ts
kanej:
```suggestion
      // chain id 999 corresponds to hyperEvm, wanchainTestnet and zoraTestnet
```

schaable (approved):

PR #7809, opened 2025-12-18
test: fix test that dependended on specific chain

kanej:
`viem` added another chain with chain id 999, breaking a test. The test has been updated, along with a minimal update to `pnpm-lock.json`.


schaable (approved):
Looks good. It would be good to update the comment at the top of the test to reflect the change and avoid confusion. Pre-approving.

v-next/hardhat-viem/test/chains.ts
kanej:
```suggestion
      // chain id 999 corresponds to hyperEvm, wanchainTestnet and also zoraTestnet
```

PR #7814, opened 2025-12-20
fix(hardhat-shorthand): handle signal terminations correctly

anim001k:
The `hh` wrapper was incorrectly treating signal-terminated Hardhat processes as successful (exit code 0). When Hardhat was terminated by a signal (e.g., SIGINT, SIGTERM), `spawnSync` returns `status === null`, and the wrapper would default to exit code 0, causing CI/CD pipelines and shell scripts to incorrectly report success.

### DIFF
--- a/archive/hardhat-shorthand/src/index.ts
+++ b/archive/hardhat-shorthand/src/index.ts
@@ -18,7 +18,7 @@ export async function main() {
     process.exit(1);
   }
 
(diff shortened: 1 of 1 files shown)


michalbrabec:
This code is archived as shorthand is not yet ported to Hardhat 3.
We plan a redesign of the plugin during the port.

PR #7815, opened 2025-12-20
chore(scripts): dedupe Ignition skip list in v-next npm scripts check

sashaodessa:
## Summary
- replace five repeated Ignition skip checks with a single Set lookup in `scripts/check-v-next-npm-scripts.js`
- keep the same skip list while reducing duplication and maintenance risk


michalbrabec:
Hi,

PR #7835, opened 2026-01-04
Official Linux Arm64 compiler support added

michalbrabec:
The official solc Linux ARM64 compiler support added.

The official list supports only version `0.8.31+` so it is necessary to merge the official compiler list and our own build list to maintain support for older versions (our own builds), while adding new versions (from the official list).

The builds can now have an optional URL property that is used for the older ARM64 Linux builds which are downloaded from a different URL than the official ones. The property can be also leveraged later down the line for other compiler.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
@@ -40,7 +40,7 @@ const LINUX_ARM64_REPOSITORY_URL =
 
 export enum CompilerPlatform {
   LINUX = "linux-amd64",
-  LINUX_ARM64 = "linux-aarch64",
+  LINUX_ARM64 = "linux-arm64",
   WINDOWS = "windows-amd64",
   MACOS = "macosx-amd64",
   WASM = "wasm",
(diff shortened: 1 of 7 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -341,46 +342,59 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
...
-    await this.#downloadFunction(url, downloadPath);
+    // download hte official solc compiler list (now that ARM64 Linus is supported)
+    await this.#downloadFunction(
+      `${COMPILER_REPOSITORY_URL}/${this.#platform}/list.json`,
+      downloadPath,
+    );
+
+    // for Linux ARM64, we need to merge the official list with our custom builds
+    if (this.#platform === CompilerPlatform.LINUX_ARM64) {
+      // cache the official list since the file will be overwritten bellow
```
ChristopherDedominici: Super minor: spelling error for `bellow`

michalbrabec: Fixed.

PR #7840, opened 2026-01-05
Check `.gas-snapshot` with `--snapshot-check` flag

### LINKED ISSUE #7771: Check `.gas-snapshot` with `--snapshot-check` flag
### Describe the issue

**Tasks:**
- Add `--snapshot-check` flag to `test` and `test solidity` tasks
- Compare current gas usage with stored `.gas-snapshot` (unit tests: single value, fuzz tests: median, invariant: skip)
- Return non-zero exit code on mismatch
- Create `.gas-snapshot` if it doesn't exist (don't fail)
- Handle new tests (add entries without failing)

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7771

Should be reviewed after https://github.com/NomicFoundation/hardhat/pull/7792


v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-snapshots.ts
```diff
@@ -26,46 +140,266 @@ export function extractFunctionGasSnapshots(
...
-        ? `gas: ${gasUsage.consumedGas}`
+      gasUsage.kind === "standard"
+        ? `gas: ${gasUsage.gas}`
         : `runs: ${gasUsage.runs}, μ: ${gasUsage.meanGas}, ~: ${gasUsage.medianGas}`;
 
-    lines.push(`${contractNameOrFqn}:${functionName} (${gasDetails})`);
+    lines.push(`${contractNameOrFqn}:${functionSig} (${gasDetails})`);
+  }
+
+  return lines.sort((a, b) => a.localeCompare(b)).join("\n");
```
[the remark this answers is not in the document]
schaable: The string format is the output and what we want to sort in the `.gas-snapshots` file. The stringification and sorting are coupled, so separating them wouldn’t make much sense and would also make the sorting more complex.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -1,30 +1,27 @@
...
-    await saveGasFunctionSnapshots(
-      hre.config.paths.root,
-      stringifiedFunctionGasSnapshots,
-    );
+  if (testsPassed) {
+    if (args.snapshot) {
+      await handleSnapshot(rootPath, suiteResults);
+    } else if (args.snapshotCheck) {
+      await handleSnapshotCheck(rootPath, suiteResults);
+    }
```
kanej, line 24: Should we throw an error if both `--snapshot` and `--snapshot-check` are set?

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-snapshots.ts
```diff
@@ -1,22 +1,136 @@
...
+
+  if (changed.length > 0) {
+    console.log();
+    console.log(
+      `${chalk.red("Gas snapshot check failed:")} ${chalk.grey(`${changed.length} function(s) changed`)}`,
+    );
+    console.log();
+
+    printFunctionGasSnapshotChanges(changed);
+    process.exitCode = 1;
```
kanej, line 94: I think I would move this to the bottom of the block and add explicit returns for each of these if branches.
This is a stylistic preference on my part, but I like being explicit about call paths and making it obvious that this one means the check failed and it has these side-effects and consequences.
I might be tempted to pull the exit code and logging up and out of this function (everyone loves a results object!), but I think you are better placed to make the call on effort to value - I am happy with either way.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
schaable: I'm up for being explicit. I'll add this to https://github.com/NomicFoundation/hardhat/pull/7854 to avoid rebasing.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-snapshots.ts
schaable: I agree. I’ll have `handleSnapshotCheck` return a boolean, and then set `process.exitCode` based on that.

Moving the logs is a bit trickier since a big part of what this function does is logging. I also prefer keeping the task action simple and clean.

kanej: Cool. That sounds like a good tradeoff.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-snapshots.ts
```diff
@@ -1,22 +1,136 @@
...
+): Promise<void> {
+  const functionGasSnapshots = extractFunctionGasSnapshots(suiteResults);
+
+  let previousFunctionGasSnapshots: FunctionGasSnapshot[];
+  try {
+    previousFunctionGasSnapshots = await readFunctionGasSnapshots(basePath);
+  } catch (error) {
+    if (error instanceof FileNotFoundError) {
+      return handleSnapshot(basePath, suiteResults);
+    }
```
kanej, line 76: I would be tempted to put in a guard at the start of this function (or pull it up to the task), to check for the `.gas-snapshot` file and then change the control flow. This works and is more compact, so I think this is a tradeoff decision that I will leave to you.
I mention it as the flow of your functions so cleanly express the algorithm, this was the only place I missed that there was more going on.

kanej:
I am running through manual tests. One of the things I think we should add for snapshot check failures is a link to the test that triggers the fail. A test failure generates a clickable link (that in vscode will jump to the test):

image

Currently a fail looks like:

```shell
Gas snapshot check failed: 4 function(s) changed

  CalculatorTest#testFuzzSubtraction(uint256,uint256)
    Runs: 256
    Expected (~): 58729
    Actual (~):   58755 (+0.04%, Δ+26)

  CalculatorTest#testSnapshotGasMultipleOperations()
    Expected (gas): 74661
    Actual (gas):   74687 (+0.03%, Δ+26)

  CalculatorTest#testSnapshotGasSubtraction()
    Expected (gas): 57557
    Actual (gas):   57583 (+0.05%, Δ+26)

  CalculatorTest#testSubtraction()
    Expected (gas): 59962
    Actual (gas):   59988 (+0.04%, Δ+26)

To update snapshots, run your tests with --snapshot
```

Maybe something like:

```shell
  CalculatorTest#testFuzzSubtraction(uint256,uint256)
    (in test/contracts/Calculator.t.sol:34)
    Runs: 256
    Expected (~): 58729
    Actual (~):   58755 (+0.04%, Δ+26)
```

kanej (approved):
I have been through the code and left a few comments - @schaable is planning on putting in those changes in a follow up PR.
I did a round of manual tests and ran into no failures - however, there is an issue with our current spec.
If tests are added or removed, the spec calls for displaying a list of additions and a list of removals - but only if there are no gas changes.
However, any addition or removal I try leads to trivial gas changes. For instance, in the example projects `calculator.t.sol`, if you comment out the first test `testAddition` instead of seeing one removed test you will get gas report failures in the other tests:

```shell
Gas snapshot check failed: 9 function(s) changed

  CalculatorTest#testDivision()
    Expected (gas): 60060
    Actual (gas):   59994 (-0.11%, Δ-66)

  CalculatorTest#testDivisionByZero()
    Expected (gas): 9923
    Actual (gas):   9945 (+0.22%, Δ+22)

  CalculatorTest#testFuzzAddition(uint128,uint128)
    Runs: 256
    Expected (~): 58490
    Actual (~):   58423 (-0.11%, Δ-67)

  CalculatorTest#testFuzzMultiplication(uint128,uint128)
    Runs: 256
    Expected (~): 58599
    Actual (~):   58532 (-0.11%, Δ-67)

  CalculatorTest#testReset()
    Expected (gas): 45401
    Actual (gas):   45419 (+0.04%, Δ+18)

  CalculatorTest#testSnapshotGasAddition()
    Expected (gas): 57485
    Actual (gas):   57507 (+0.04%, Δ+22)

  CalculatorTest#testSnapshotGasDivision()
    Expected (gas): 57589
    Actual (gas):   57611 (+0.04%, Δ+22)

  CalculatorTest#testSnapshotGasMultipleOperations()
    Expected (gas): 74661
    Actual (gas):   74574 (-0.12%, Δ-87)

  CalculatorTest#testSubtraction()
    Expected (gas): 59962
    Actual (gas):   59984 (+0.04%, Δ+22)

To update snapshots, run your tests with --snapshot
```

I am approving this PR as this is an issue in the spec rather than the code. We will revisit the display or added/removed tests before the feature is considered complete.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/7854/changes/c86a6e106f240df658f510b5aab7ee25e018c701).

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/gas-snapshots.ts
schaable: Ok, so I also ended up extracting the logging into its own functions. Done [here](https://github.com/NomicFoundation/hardhat/pull/7854/changes/6aa26ec4e7db78c2d42327fa48536f9c0a11534f).

Issue #7834, opened 2026-01-02
hardhat_reset behavior in edr@0.12.0


keithchew:
### Version of Hardhat

2.26.5

### What happened?

Hi

First a background on the working setup, hardhat@2.26.5 which uses edr@0.11.3.

I call hardhat_reset:

                await this.providerSendAsync(web3Hre, 'hardhat_reset', [
                    {
                        forking: {
                            jsonRpcUrl: this.provider,
                            blockNumber: blockNumberReal,
                        },
                    },
                ]);

where blockNumberReal is the blockNumber from the real chain. Then I call this to make sure the fork worked:

                const blockNumberResults = await this.providerSendAsync(web3Hre, "eth_blockNumber", []);
                const blockNumber = blockNumberResults.result;
                if (blockNumber === '0') {
                    throw new Error(`Internal error: hardhat did not fork correctly`);
                }
This setup works beautifully.

But in the next version, hardhat@2.27.0, it uses edr@0.12.0-next.7. With this version, I am getting 0 for the blockNumber after hardhat_reset. 

Looking at the edr's repo history, I can see that hardhat_reset was removed in 0.12.0-next.3.

Can someone please advise how can I achieve the same behavior as my working setup in edr@0.12.x?

### Minimal reproduction steps

Steps outlined above.

Wodann:
See https://github.com/NomicFoundation/edr/issues/1244#issuecomment-3710838640.

I'm happy to look into this, as I rewrote the Hardhat 2 logic that caused this bug.

> ### Version of Hardhat
>
> 0.26.5

@keithchew do you mean version 2.26.5?

kanej:
Thanks @Wodann, I took a quick look and was able to `hardhat_reset` with `2.26.5` and the reset worked. I haven't tried with the forking settings above - so I concur it is likely the reconstruction logic when working with forking.

keithchew:
Hi @Wodann, thank you so much for looking into this, much appreciated.

Yes my apologies, it is 2.26.5 and 2.27.0, which I have now corrected in the post (as well as https://github.com/NomicFoundation/edr/issues/1244). I am currently pinning to 2.26.5 to retain the feature.

Please let me know if I can be of any further assistance, also happy to help test any fixes or a branch when that is ready.

Wodann:
@keithchew, I added a test to reproduce your issue but it's working as expected.

Could you please elaborate what `this.provider` is in this context:

```ts
[...]

PR #7850, opened 2026-01-08
fix: switching between local and fork mode using

Wodann:

---

Fixes https://github.com/NomicFoundation/hardhat/issues/7834

After running the user's reproduction script, I was able to observe a reported error from EDR:

```
{"jsonrpc":"2.0","id":1767895347701,"error":{"code":-32603,"message":"Error: Storage overrides are not supported for forked blocks yet. See https://github.com/NomicFoundation/edr/issues/911","data":{"message":"Error: Storage overrides are not supported for forked blocks yet. See https://github.com/NomicFoundation/edr/issues/911"}}}
```

Fork mode does not support storage overrides yet. When using `hardhat_reset` to switch between local and fork mode, we are reusing the same `ProviderConfig`, which contained storage overrides for the L1 genesis state.

This PR fixes that by reconstructing the desired genesis state, depending on whether the `hardhat_reset` request requests local or fork mode.

I also made sure that the `cacheDir` and `chainOverrides` are set correctly, depending on whether we're dealing with local or fork mode after the `hardhat_reset`.

I added a reproduction test in EDR: https://github.com/NomicFoundation/edr/pull/1257

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
@@ -530,8 +530,6 @@ export class EdrProviderWrapper
           ? this._originalChainOverrides
           : this._providerConfig.fork?.chainOverrides;
 
-      this._providerConfig;
-
       this._providerConfig.fork = {
         blockNumber:
           forkConfig.blockNumber !== undefined
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
@@ -161,7 +161,7 @@ export class EdrProviderWrapper
     // calls, in case there is switching between local and fork configurations.
     private readonly _originalGenesisAccounts: GenesisAccount[],
     private readonly _originalCacheDir: string | undefined,
-    private readonly _originalChainOverrides: Array<ChainOverride> | undefined
+    private readonly _originalChainOverrides: ChainOverride[] | undefined
   ) {
     super();
   }
@@ -681,7 +681,7 @@ function _genesisStateAndOwnedAccounts(
   isForked: boolean,
   hardfork: string,
   genesisAccounts: GenesisAccount[]
-): [Array<AccountOverride>, string[]] {
+): [AccountOverride[], string[]] {
   const { l1GenesisState, l1HardforkFromString } = requireNapiRsModule(
     "@nomicfoundation/edr"
   ) as typeof import("@nomicfoundation/edr");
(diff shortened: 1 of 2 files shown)


packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -512,14 +510,33 @@ export class EdrProviderWrapper
...
     const forkConfig = networkConfig?.forking;
 
+    const [genesisState, ownedAccounts] = _genesisStateAndOwnedAccounts(
+      forkConfig !== undefined,
+      this._providerConfig.hardfork,
+      this._originalGenesisAccounts
+    );
+
+    this._providerConfig.genesisState = genesisState;
+    this._providerConfig.ownedAccounts = ownedAccounts;
```
kanej, line 520: Can I check I have followed. This is the heart of the fix, we are pulling the genesis state transform across from `create` and using it in `reset` as well?

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
```diff
@@ -512,14 +510,33 @@ export class EdrProviderWrapper
...
     if (forkConfig !== undefined) {
+      const cacheDir =
+        this._providerConfig.fork === undefined
+          ? this._originalCacheDir
+          : this._providerConfig.fork?.cacheDir;
+
+      const chainOverrides =
+        this._providerConfig.fork === undefined
+          ? this._originalChainOverrides
+          : this._providerConfig.fork?.chainOverrides;
```
kanej, line 531: Can you talk me through this change from only relying on the `_providerConfig.fork?`? What case triggers this?

kanej (commented):
Thanks @Wodann for putting this together. I left a few comments that are for my understanding, this is not an area I am familiar with.

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
Wodann: Correct

packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
Wodann: The original config contains a `cacheDir` and `chainOverrides` configurations.
The EDR `ProviderConfig` does not include these fields itself, but rather treats them as part of the `ForkConfig`, since they are only used for the forking feature.

If the previous provider's configuration was set for local mode, the `ProviderConfig.fork === undefined` and we won't know the `cacheDir` or `chainOverrides` to use anymore.

If the reset request is for fork mode, we would be missing the `cacheDir` and `chainOverrides`. This change ensures that we can fallback to the `cacheDir` and `chainOverrides` specified in the original values of `Provider.create`'s `ProviderConfig`.

kanej (approved):

PR #7851, opened 2026-01-08
Inline task actions - strong types check

### LINKED ISSUE #7560: Inline task actions
## Context

One of the last-minute breaking changes that @ChristopherDedominici and I decided to implement before the beta, was simplifying `setAction` in task builders, and remove the possibility of adding an inline action. 

The reason behind that was that it having an union type there was extremely error prone and hard to debug.

Then cons of having done that is that users defining their own tasks have to split the task action to a different file, which is annoying and extra boilerplate. Note that extra boilerplate is ok for plugins, as plugin authors are more invested already. 

To be clear, when I refer to user tasks I refer to the ones defined in `HardhatUserConfig#tasks`. Not the ones coming from plugins.

## Proposed change

Add a `setInlineAction` function that can only be used in user tasks.

## Potential designs

1. We add it to the task builders and add config validation logic to forbid plugins to use it.
2. We create different builders for users and plugins.

In both cases we should validate that the user doesn't use both `setAction` and `setInlineAction`.

(1) is simpler to implement. (2) is more involved, but also harder to enforce, as plugin authors may accidentally still use the user-level builders.

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/7560

Same as [7950](https://github.com/NomicFoundation/hardhat/pull/7950), but here checks are mostly executed at type level

### DIFF
--- a/v-next/hardhat/src/internal/core/tasks/resolved-task.ts
+++ b/v-next/hardhat/src/internal/core/tasks/resolved-task.ts
@@ -8,6 +8,7 @@ import type {
   LazyActionObject,
   NewTaskActionFunction,
   Task,
+  TaskAction,
   TaskActions,
   TaskArguments,
   TaskOverrideActionFunction,
@@ -49,7 +50,7 @@ export class ResolvedTask implements Task {
     hre: HardhatRuntimeEnvironment,
     id: string[],
     description: string,
-    action: LazyActionObject<NewTaskActionFunction>,
+    taskAction: TaskAction,
     options: Record<string, OptionDefinition>,
     positionalArguments: PositionalArgumentDefinition[],
     pluginId?: string,
@@ -57,7 +58,12 @@ export class ResolvedTask implements Task {
     return new ResolvedTask(
       id,
       description,
-      [{ pluginId, action }],
+      [
+        {
+          pluginId,
+          ...taskAction,
+        },
+      ],
       new Map(Object.entries(options)),
       positionalArguments,
       pluginId,
@@ -142,17 +148,25 @@ export class ResolvedTask implements Task {
       nextTaskArguments: TaskArguments,
       currentIndex = this.actions.length - 1,
     ): Promise<any> => {
-      // The first action may be empty if the task was originally an empty task
-      const currentAction =
-        this.actions[currentIndex].action ??
-        (async () => ({
-          default: () => {},
-        }));
+      const currentTaskAction = this.actions[currentIndex];
 
-      const actionFn = await this.#resolveImportAction(
-        currentAction,
-        this.actions[currentIndex].pluginId,
-      );
+      let actionFn: NewTaskActionFunction | TaskOverrideActionFunction;
+
+      // The first action may be empty if the task was originally an empty task
+      if (currentTaskAction.inlineAction !== undefined) {
+        actionFn = currentTaskAction.inlineAction;
+      } else {
+        const lazyAction =
+          currentTaskAction.action ??
+          (async () => ({
+            default: () => {},
+          }));
+
+        actionFn = await this.#resolveImportAction(
+          lazyAction,
+          currentTaskAction.pluginId,
+        );
+      }
 
       if (currentIndex === 0) {
         /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions --
--- a/v-next/hardhat/src/internal/core/tasks/resolved-task.ts
+++ b/v-next/hardhat/src/internal/core/tasks/resolved-task.ts
@@ -86,7 +86,11 @@ export class ResolvedTask implements Task {
   }
 
   public get isEmpty(): boolean {
-    return this.actions.length === 1 && this.actions[0].action === undefined;
+    return (
+      this.actions.length === 1 &&
+      this.actions[0].action === undefined &&
+      this.actions[0].inlineAction === undefined
+    );
   }
 
   /**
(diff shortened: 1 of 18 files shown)


v-next/hardhat/src/internal/core/tasks/resolved-task.ts
```diff
@@ -142,17 +148,25 @@ export class ResolvedTask implements Task {
       nextTaskArguments: TaskArguments,
       currentIndex = this.actions.length - 1,
     ): Promise<any> => {
+      const currentTaskAction = this.actions[currentIndex];
```
ChristopherDedominici, line 155: This is where the `run` command decides whether to run the `action` or the `inlineAction`, depending on which one is defined.

v-next/hardhat/src/types/plugins.ts
```diff
@@ -20,6 +24,14 @@ declare module "./config.js" {
   }
 }
 
+/**
+ * A helper type to strictly enforce that plugins only use file-based actions.
+ */
+export type PluginSafeTaskDefinition =
```
alcuadrado, line 30: Maybe let's call it just `PluginTaskDefinition`. The `Safe` part makes it a bit confusing.

v-next/hardhat/src/types/tasks.ts
```diff
@@ -75,6 +75,17 @@ export enum TaskDefinitionType {
   TASK_OVERRIDE = "TASK_OVERRIDE",
 }
 
+export type TaskAction =
+  | { action: LazyActionObject<NewTaskActionFunction>; inlineAction?: never }
```
alcuadrado, line 79: What about renaming this to:

```suggestion
  | { lazyAction: LazyActionObject<NewTaskActionFunction>; inlineAction?: never }
```

this way any error message will be more clear.

v-next/hardhat/src/types/tasks.ts
```diff
@@ -160,10 +196,28 @@ export interface EmptyTaskDefinitionBuilder {
...
+ * and can strictly be called only once per task definition chain.
+ * @template ActionTypeT Tracks if the action is "FILE" (Plugin Safe) or "INLINE".
  */
 export interface NewTaskDefinitionBuilder<
   TaskArgumentsT extends TaskArguments = TaskArguments,
+  ActionStateT extends
+    | MissingActionState
+    | ActionDefinedState
+    | DuplicateActionError = MissingActionState,
+  ActionTypeT extends "FILE" | "INLINE" | "NONE" = "NONE",
```
alcuadrado, line 212: what about something easier to read, like this? Again, because this may leak to type errors

```suggestion
  ActionTypeT extends "LAZY_ACTION" | "INLINE_ACTION" | "MISSING_ACTION" = "MISSING_ACTION",
```

v-next/hardhat/src/types/tasks.ts
```diff
@@ -160,10 +196,28 @@ export interface EmptyTaskDefinitionBuilder {
...
 > {
+  /**
+   * Technical property needed to enforce the `ActionStateT` state check at compile time.
```
alcuadrado, line 215: Is this really needed? What happens if you omit it? I think a common name, at least in rust, would be Phantom: https://doc.rust-lang.org/std/marker/struct.PhantomData.html

v-next/hardhat/src/types/tasks.ts
```diff
@@ -172,15 +226,70 @@ export interface NewTaskDefinitionBuilder<
...
+      ActionDefinedState,
+      "FILE" | "INLINE"
+    >,
     action: LazyActionObject<NewTaskActionFunction<TaskArgumentsT>>,
-  ): this;
+  ): NewTaskDefinitionBuilder<
+    TaskArgumentsT,
+    DuplicateActionError,
+    "FILE" | "INLINE"
+  >;
```
alcuadrado, line 260: I understand this, but maybe it's too much? I think if the duplicated action doesn't fail at type-level, but fails immediately in the builder that's fine. It's a bug that you would catch in the very first run.

Same with missing action.

Removing them would simplify the code, at the expenses of only finding the error in the first run (I think that's fine), and adding some runtime validation in the builders and config validation.

Note that the validation is still needed, as the user may be using js.

alcuadrado: Would we loose anything else?

alcuadrado: This also applies for the other kind of task definitions

alcuadrado (changes requested):
I left some comments.

I think the main one is weather if the type complexity is worth it or not.

Also, this PR will need to update the config validation. The reason is that a user can provide a task definition that wasn't created with the builders, so we need to validate them.

v-next/hardhat/src/types/tasks.ts
ChristopherDedominici: I agree, but this will break current plugins, is this ok?

ChristopherDedominici:
@alcuadrado I created a separate PR where most of the checks have been modified to be runtime checks, see here ([see here](https://github.com/NomicFoundation/hardhat/pull/7950)), so we can have a direct comparison between the two PRs.

v-next/hardhat/src/types/tasks.ts
ChristopherDedominici: This will be a breaking change, so are we okay with that?

PR #7861, opened 2026-01-13
[CI] Fix bug on workflow `check-v3-with-latest-dependencies`

ChristopherDedominici:


.changeset/polite-stingrays-develop.md
```diff
@@ -0,0 +1,13 @@
...
+"@nomicfoundation/hardhat-ignition": patch
+"@nomicfoundation/hardhat-toolbox-viem": patch
+"@nomicfoundation/hardhat-utils": patch
+"@nomicfoundation/hardhat-viem-assertions": patch
+"@nomicfoundation/hardhat-viem": patch
+"hardhat": patch
+"template-node-test-runner-viem": patch
+---
+
+Bumped `viem` version across all packages [7861](https://github.com/NomicFoundation/hardhat/pull/7861).
```
ChristopherDedominici, line 11: This is the only impact on users, all the other changes are internal

v-next/hardhat-utils/src/internal/lang.ts
```diff
@@ -53,3 +53,39 @@ export function deepMergeImpl<T extends object, S extends object>(
...
+let cachedCustomEqual: ((a: unknown, b: unknown) => boolean) | undefined;
+
+/**
+ * Performs a custom deep equality check using `fast-equals` with specific overrides.
+ *
+ * @param x The first value to compare.
+ * @param y The second value to compare.
+ * @returns A promise that resolves to true if the values are deeply equal, false otherwise.
+ */
+export async function customFastEqual<T>(x: T, y: T): Promise<boolean> {
```
ChristopherDedominici, line 66: `fast-equals` version `5.4.0` introduced strict byteOffset checks for TypedArray equality, [see this PR](https://github.com/planttheidea/fast-equals/releases/tag/v5.4.0). This breaks comparison for Node.js Buffers because Node uses a shared memory pool for small buffers. Consequently, two identical buffers (e.g., Buffer.from('test')) will share the same underlying ArrayBuffer but have different byteOffsets, causing deepEqual to return false.

Example:
```
const buf1 = Buffer.from('test');
const buf2 = Buffer.from('test');

// Why it fails in v5.4.0+:
// Node.js allocates small buffers from a shared memory pool.
console.log(buf1.equals(buf2));           // true (Same content)
console.log(buf1.buffer === buf2.buffer); // true (Same underlying ArrayBuffer)
console.log(buf1.byteOffset);             // 0
console.log(buf2.byteOffset);             // 8  <-- strict check fails here!
```

v-next/hardhat/templates/hardhat-2/04-mocha-viem-ts/package.json
```diff
@@ -20,6 +20,6 @@
     "solidity-coverage": "^0.8.0",
     "ts-node": ">=8.0.0",
     "typescript": "~5.0.4",
-    "viem": "^2.7.6"
```
ChristopherDedominici, line 23: All other packages have `viem: ^2.30.0`, but here is `2.7.6`. Do you think it is safe to align this too with the others, and update it to `2.43.0`?

schaable (approved):
💯

v-next/hardhat/templates/hardhat-2/04-mocha-viem-ts/package.json
schaable: Yes, we probably just forgot to update it last time.

v-next/hardhat-utils/src/errors/fs.ts
```diff
@@ -35,7 +35,7 @@ export class NotADirectoryError extends CustomError {
 }
 
 export class IsDirectoryError extends CustomError {
-  constructor(filePath: string, cause: Error) {
+  constructor(filePath: string, cause: Error | undefined) {
```
kanej, line 38:
```suggestion
  constructor(filePath: string, cause?: Error) {
```
This would match `CustomError`.

v-next/hardhat-utils/src/internal/lang.ts
kanej: Should we bump the minimum `fast-equals` dependency to `"fast-equals": "^5.4.0",` in `hardhat-utils`? Maybe the API used here works, but I would bump it just to make sure.

.changeset/polite-stingrays-develop.md
kanej, line 11:
```suggestion
---
"@nomicfoundation/hardhat-ignition-viem": patch
"@nomicfoundation/hardhat-ignition": patch
"@nomicfoundation/hardhat-toolbox-viem": patch
"@nomicfoundation/hardhat-utils": patch
"@nomicfoundation/hardhat-viem-assertions": patch
"@nomicfoundation/hardhat-viem": patch
"hardhat": patch
---

Bumped `viem` version across all packages [7861](https://github.com/NomicFoundation/hardhat/pull/7861).
```
I think the example project and the template project should be excluded and would cause issues with the release PR.

kanej (approved):
This looks great. Thanks for clearing down the CI breaks.
I looked through and all the changes to utils look sensible to me, though I think @schaable is a better judge.
I am pre-approving, but I think we should consider:
* bumping up the `fast-equals` min supported, unless your happy the new API usage works on the current minimal version
* bumping the `viem` version for the other package now Luis has confirmed.

v-next/hardhat-utils/src/errors/fs.ts
schaable: I thought about this, and I believe the original code suits the usage better, since in the copy call we’re explicitly passing undefined, not omitting the parameter. Both ways are equivalent though, this is mostly semantic, so I’m good with either approach.

kanej: Cool. Lets leave it.

v-next/hardhat-utils/src/internal/lang.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7861/commits/d0e4245b96ffb96c9ecac2b4e3b44c58643a2295

PR #7862, opened 2026-01-14
Suppress the Natspec memory-safe-assembly special comment warning 

### LINKED ISSUE #7842: Suppress the Natspec `memory-safe-assembly` special comment warning
### Describe the issue

The natspec annotation for the inline assembly (`/// @solidity memory-safe-assembly`) was deprecated in Solidity 0.8.31 and will be removed in Solidity 0.9.0. Since then, solc 0.8.31+ started showing the following warning, Suppress the warning.

```
```

The warning displayed in case both the annotation and natspec comment will not be impacted.

```
Warning: Inline assembly marked as memory safe using both a NatSpec tag and an assembly block annotation. If you are not concerned with backwards compatibility, only use the assembly block annotation, otherwise only use the NatSpec tag.
```

For context both options in Solidity code
```solidity
/// @solidity memory-safe-assembly <- natspec comment
assembly("memory-safe") // <- block annotation
{ ... }
```

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/7842

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -1055,9 +1055,17 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       return;
     }
 
+    // Filter out deprecated natspec memory-safe-assembly warnings
+    const filteredErrors = errors.filter((error) => {
+      const msg = error.formattedMessage ?? error.message;
+      return msg.includes(
+        "Warning: Natspec memory-safe-assembly special comment for inline assembly is deprecated and scheduled for removal. Use the memory-safe block annotation instead.",
+      );
+    });
+
     console.log();
 
-    for (const error of errors) {
+    for (const error of filteredErrors) {
       if (error.severity === "error") {
         const errorMessage: string =
           this.#getFormattedInternalCompilerErrorMessage(error) ??
@@ -1077,7 +1085,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       }
     }
 
-    const hasConsoleErrors: boolean = errors.some((e) =>
+    const hasConsoleErrors: boolean = filteredErrors.some((e) =>
       this.#isConsoleLogError(e),
     );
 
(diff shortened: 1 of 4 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -75,6 +75,11 @@ import { SolcConfigSelector } from "./solc-config-selection.js";
 
 const log = debug("hardhat:core:solidity:build-system");
 
+// Warnings that should be suppressed from the build output.
+export const SUPPRESSED_WARNINGS: string[] = [
```
ChristopherDedominici: If in the future we want suppress more warnings, we just need to add them here

schaable (approved):
Looks good to me!

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -75,6 +75,21 @@ import { SolcConfigSelector } from "./solc-config-selection.js";
...
+// This allows suppressing known warnings from internal files (e.g., console.sol)
+// while still showing the same warning type from user code.
+export const SUPPRESSED_WARNINGS: Array<{
+  message: string;
+  sourceFile: string;
+}> = [
+  {
+    message:
+      "Natspec memory-safe-assembly special comment for inline assembly is deprecated and scheduled for removal. Use the memory-safe block annotation instead.",
+    sourceFile: path.normalize("hardhat/console.sol"),
```
kanej, line 89: What is the `path.normalize` buying us here?

kanej (approved):
This looks good. The tests in particular.
I did a manual test based on the example project, and the code suppresses warnings coming from Hardhat's console.
Anything pulling in `forge-std` is still going to get inundated, this PR gives us a mechanism for suppressing that but we can do that as a follow up PR.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
ChristopherDedominici: On windows the path becomes `hardhat\console.sol`. So it is ignored if I hardcode `hardhat/console.sol`

PR #7868, opened 2026-01-14
fix: isolate clean task tests to prevent deleting global solc cache

alcuadrado:
The clean task tests were explicitly deleting the real global cache directory, causing the user's solc binaries to be deleted after running tests. This fix uses an isolated temporary directory via XDG_CACHE_HOME environment variable, ensuring the real global cache is preserved.

### DIFF
--- a/v-next/hardhat-utils/src/global-dir.ts
+++ b/v-next/hardhat-utils/src/global-dir.ts
@@ -21,8 +21,9 @@ export async function getConfigDir(
  * Returns the cache directory path for a given package (defaults to "hardhat").
  * Ensures that the directory exists before returning the path.
  *
- * The cache directory can be overridden by setting the `HARDHAT_CACHE_DIR`
- * environment variable.
+ * For testing purposes only, the cache directory can be overridden by setting
+ * the `HARDHAT_TEST_CACHE_DIR` environment variable. This is intended to
+ * isolate tests from the real global cache.
  *
  * @param packageName The name of the package for which to generate paths. Defaults to "hardhat" if no package name is provided.
  * @returns The path to the hardhat cache directory.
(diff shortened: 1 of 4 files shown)


v-next/hardhat-utils/src/global-dir.ts
```diff
@@ -21,13 +21,24 @@ export async function getConfigDir(
...
+ *
  * @param packageName The name of the package for which to generate paths. Defaults to "hardhat" if no package name is provided.
  * @returns The path to the hardhat cache directory.
  * @throws FileSystemAccessError for any error.
  */
 export async function getCacheDir(
   packageName: string = HARDHAT_PACKAGE_NAME,
 ): Promise<string> {
+  // Allow override for testing purposes only
+  const testOverrideDir = process.env.HARDHAT_TEST_CACHE_DIR;
```
michalbrabec: There is nothing enforcing that the override is used only for testing.
I am fine with that but the comment is not accurate in that case.

v-next/hardhat-utils/test/global-dir.ts
```diff
@@ -18,37 +21,77 @@ describe("global-dir", () => {
...
+        (await getExpectedPath(CUSTOM_PACKAGE_NAME)).cache,
+      );
+    });
+
+    it("should use HARDHAT_TEST_CACHE_DIR when set (for testing purposes only)", async () => {
+      const customPath = path.join(
+        os.tmpdir(),
+        `hardhat-test-cache-override-${Date.now()}`,
+      );
+      await mkdir(customPath);
```
michalbrabec, line 69: `getCacheDir()` must ensure that the directory exists before returning.
We should add a test that verifies that for the override directory

v-next/hardhat/test/internal/builtin-plugins/clean/task-action.ts
```diff
@@ -63,6 +70,11 @@ describe("clean/task-action", () => {
     useFixtureProject("loaded-config");
 
     before(async function () {
+      // Set up isolated cache directory to avoid deleting the real global cache
+      originalHardhatTestCacheDir = process.env.HARDHAT_TEST_CACHE_DIR;
+      testGlobalCacheRoot = await getTmpDir("clean-task-global-cache");
+      process.env.HARDHAT_TEST_CACHE_DIR = testGlobalCacheRoot;
```
michalbrabec: Are there other tests/test suites that should leverage this?
The tests might not clean the directory completely like the clean task, but they likely still modify it, which could lead to conflict in case the user creates contracts that conflict with some tests.

v-next/hardhat-utils/src/global-dir.ts
alcuadrado: good point, i switched to an explicit mocking api

v-next/hardhat/test/internal/builtin-plugins/clean/task-action.ts
alcuadrado: another good point, i looked for tests that could use the cache and updated them

ChristopherDedominici (approved):

PR #7871, opened 2026-01-16
Improve ignition viem build

kanej:
We check the types of the generated output of `hardhat-ignition-viem` using ts expected error lines within the tests.

Previously this was checked on build, but that required a prebuild step - which leads to a mirroring hierarchy of prebuild steps matching with the tsc references.

Now build of the package will exclude the tests. Triggering the tests will lead to a prestep that runs the viem type generation for the test fixture projects via a Hardhat build. It will then run a tsc build on the test files including the ts expected error lines, before finally running the tests.

### DIFF
--- a/v-next/example-project/package.json
+++ b/v-next/example-project/package.json
@@ -16,7 +16,6 @@
     "lint": "pnpm prettier --check",
     "lint:fix": "pnpm prettier --write",
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
-    "prebuild": "pnpm run --dir ../hardhat-ignition-viem build",
     "build": "tsc --build .",
     "clean": "rimraf dist artifacts cache ignition/deployments types",
     "pretest": "pnpm build && pnpm install",
--- a/v-next/hardhat-ignition-ui/package.json
+++ b/v-next/hardhat-ignition-ui/package.json
@@ -20,7 +20,7 @@
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
     "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
     "preview": "vite preview",
-    "clean": "rimraf dist",
+    "clean": "rimraf ./dist ./tsconfig.node.tsbuildinfo ./tsconfig.tsbuildinfo",
     "prepack": "pnpm build && vite build"
   },
   "files": [
--- a/v-next/hardhat-ignition-viem/eslint.config.js
+++ b/v-next/hardhat-ignition-viem/eslint.config.js
@@ -1,3 +1,15 @@
+import path from "node:path";
 import { createConfig } from "../config/eslint.config.js";
 
-export default createConfig(import.meta.filename);
+const config = createConfig(import.meta.filename);
+
+// Use the tsconfig.test.json file for linting as it
+// includes the test files
+const parserOptions = config[0].languageOptions.parserOptions;
+parserOptions.project = path.join(
+  path.dirname(import.meta.filename),
+  "tsconfig.test.json",
+);
+parserOptions.projectService = undefined;
+
+export default config;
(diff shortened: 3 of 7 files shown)


v-next/example-project/package.json
```diff
@@ -16,7 +16,6 @@
     "lint": "pnpm prettier --check",
     "lint:fix": "pnpm prettier --write",
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
-    "prebuild": "pnpm run --dir ../hardhat-ignition-viem build",
```
kanej, line 19: The prebuild is no longer needed, the build references will build the dependencies via `tsc` now.

v-next/hardhat-ignition-ui/package.json
```diff
@@ -20,7 +20,7 @@
     "prettier": "prettier \"**/*.{ts,js,md,json}\"",
     "eslint": "eslint \"src/**/*.ts\" \"test/**/*.ts\"",
     "preview": "vite preview",
-    "clean": "rimraf dist",
```
kanej, line 23: I updated `clean` to follow the principle that `pnpm clean` should be the equivalent of `git clean -fdx .` but excluding `node_modules`.

v-next/hardhat-ignition-viem/eslint.config.js
```diff
@@ -1,3 +1,15 @@
...
+const config = createConfig(import.meta.filename);
+
+// Use the tsconfig.test.json file for linting as it
+// includes the test files
+const parserOptions = config[0].languageOptions.parserOptions;
+parserOptions.project = path.join(
+  path.dirname(import.meta.filename),
+  "tsconfig.test.json",
+);
+parserOptions.projectService = undefined;
```
kanej, line 13: The default `tsconfig.json` now intentionally excludes `test` as this won't typecheck until the fixture projects have been hardhat compiled to generate the viem type outputs.
So eslint needs to be redirected to an alternate `"tsconfig.test.json"`. The project service if not undefined attempts its own resolution - so I am turning it off.

v-next/hardhat-ignition-viem/tsconfig.json
```diff
@@ -8,5 +8,5 @@
     { "path": "../hardhat-viem" },
     { "path": "../hardhat-test-utils" }
   ],
-  "exclude": ["${configDir}/dist", "${configDir}/node_modules"]
+  "exclude": ["test", "${configDir}/dist", "${configDir}/node_modules"]
```
kanej, line 11: Allow build without having to `hardhat compile` the test fixture projects to get the viem contract types.
The tests have their own `tsconfig.test.json`

v-next/hardhat-ignition-viem/package.json
```diff
@@ -25,19 +25,20 @@
...
     "test:only": "node --import tsx/esm --test --test-only --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/*.ts\" \"test/!(fixture-projects|helpers)/**/*.ts\"",
     "test:coverage": "c8 --reporter html --reporter text --all --exclude test --exclude \"src/**/{types,type-extensions}.ts\" --src src node --import tsx/esm --test --test-reporter=@nomicfoundation/hardhat-node-test-reporter \"test/*.ts\" \"test/!(fixture-projects|helpers)/**/*.ts\"",
-    "clean": "rimraf .nyc_output coverage dist tsconfig.tsbuildinfo ./test/fixture-projects/**/deployments ./test/fixture-projects/**/artifacts",
+    "clean": "rimraf .nyc_output coverage dist artifacts cache tsconfig.tsbuildinfo ./test/fixture-projects/**/deployments ./test/fixture-projects/**/artifacts ./test/fixture-projects/**/cache ./test/fixture-projects/tmp",
     "prepack": "pnpm build",
-    "pretest": "pnpm build",
-    "pretest:only": "pnpm build"
+    "test-build": "tsc --build tsconfig.json && node ./scripts/compile-test-fixture-project.js create2 && node ./scripts/compile-test-fixture-project.js minimal && tsc --build tsconfig.test.json",
```
kanej, line 39: Linting and testing require a standard build, as that will build the dependencies -including Hardhat.
They then need a hardhat compile (captured in the script calls) in the text fixtures to create the viem contract types.
Finally we check that the expected type errors in the tests work correctly with the generated viem contract types, the final `tsc --build tsconfig.test.json`.

zoeyTM: should we add a `pretest` script that runs `test-build`?

zoeyTM (approved):
one question but approved whichever decision you go with 👍

v-next/hardhat-ignition-viem/package.json
kanej: Yes, I have added both `pretest` and `pretest:only`, all they do is invoke `test-build`.

PR #7876, opened 2026-01-16
Define Compiler type in hardhat/types/solidity

### LINKED ISSUE #7875: Expose Compiler type used by the `invokeSolc` hook
### Describe the issue

#7744 introduces a new hook. One of the arguments for this hook it `Compiler`, which is defined in `src/internal/builtin-plugins/solidity/build-system/compiler/compiler.ts`

This is visible in the test file that imports it like this:
```typescript
import type { Compiler } from "../../../../src/internal/builtin-plugins/solidity/build-system/compiler/compiler.js";
```

---

I've been getting issues importing this type. It might be my IDE's config, but I get error when trying to import anything from `src/internal`. So far my solution has been to import `RunCompilationJobResult` from `hardhat/types/solidity` and then get the type as `RunCompilationJobResult["compiler"]`.

I believe it should be possible to get the `Compiler` type dirrectly. Either by having it defined in the `hardhat/types/solidity` (and imported where its used) or by having `hardhat/types/solidity` export the type that it imports from the internal location that defines it.

Amxx:
- [x] Other: this is a proposal to improve experience. Doesn't strictly qualify as a bug.

---

Fixes #7875

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -1,5 +1,4 @@
 import type { CompileCache } from "./cache.js";
-import type { Compiler } from "./compiler/compiler.js";
 import type { DependencyGraphImplementation } from "./dependency-graph.js";
 import type { Artifact } from "../../../../types/artifacts.js";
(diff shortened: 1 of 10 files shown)


kanej (approved):
Thanks @Amxx, this is great.
I did a manual check that the Compiler type is now coming through for the `invokeSolc` hook.

PR #7877, opened 2026-01-16
Compiler downloader retry

### LINKED ISSUE #7031: Flaky test: Compiler downloader downloadCompiler multiple downloads should download multiple different compilers
[[hardhat] ci on windows-latest (Node 22)](https://github.com/NomicFoundation/hardhat/actions/runs/16416075715/job/46382247057#logs)

```
1) Compiler downloader
       downloadCompiler
         multiple downloads
           should download multiple different compilers:

   HardhatError: HHE904: Couldn't download compiler version 0.5.5+commit.47a71e8f. Please check your internet connection and try again.
       at <anonymous> (src\internal\builtin-plugins\solidity\build-system\compiler\downloader.ts:209:15)
       at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
       at async MultiProcessMutex.#executeFunctionAndReleaseMutex (D:\a\hardhat\hardhat\v-next\hardhat-utils\src\synchronization.ts:80:14)
       at async Promise.all (index 4)
       at async TestContext.<anonymous> (test\internal\builtin-plugins\solidity\build-system\compiler\downloader.ts:340:11)
     [cause]: DownloadError: Failed to download file from https://binaries.soliditylang.org/
         at download (D:\a\hardhat\hardhat\v-next\hardhat-utils\dist\src\request.js:139:15)
         at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
         at async CompilerDownloaderImplementation.#downloadCompiler (src\internal\builtin-plugins\solidity\build-system\compiler\downloader.ts:348:5)
         at async <anonymous> (src\internal\builtin-plugins\solidity\build-system\compiler\downloader.ts:205:24)
         at async MultiProcessMutex.#executeFunctionAndReleaseMutex (D:\a\hardhat\hardhat\v-next\hardhat-utils\src\synchronization.ts:80:14)
         at async Promise.all (index 4)
         at async TestContext.<anonymous> (test\internal\builtin-plugins\solidity\build-system\compiler\downloader.ts:340:11)
       [cause]: RequestError: Failed to make GET request to https://binaries.soliditylang.org/
           at getRequest (D:\a\hardhat\hardhat\v-next\hardhat-utils\dist\src\request.js:36:15)
           at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
           at async download (D:\a\hardhat\hardhat\v-next\hardhat-utils\dist\src\request.js:125:27)
           at async CompilerDownloaderImplementation.#downloadCompiler (src\internal\builtin-plugins\solidity\build-system\compiler\downloader.ts:348:5)
           at async <anonymous> (src\internal\builtin-plugins\solidity\build-system\compiler\downloader.ts:205:24)
           at async MultiProcessMutex.#executeFunctionAndReleaseMutex (D:\a\hardhat\hardhat\v-next\hardhat-utils\src\synchronization.ts:80:14)
           at async Promise.all (index 4)
           at async TestContext.<anonymous> (test\internal\builtin-plugins\solidity\build-system\compiler\downloader.ts:340:11)
         [cause]: The error chain has been truncated because it's too long (limit: 3)
```

michalbrabec:
Compiler downloader retries the download in case it fails or the file verification does.
The retry mitigates cases where download fails due to the server limitation or timeouts, but it does not completely eliminates the issue.

The option `retryCount` was added to the compile task to control the number of retries, defaulting to 3.

### DIFF
--- a/.changeset/tasty-animals-scream.md
+++ b/.changeset/tasty-animals-scream.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Compiler downloader retires download in case of failure [7031](https://github.com/NomicFoundation/hardhat/issues/7031).
+Compiler downloader retries download in case of failure [7031](https://github.com/NomicFoundation/hardhat/issues/7031).
--- a/.changeset/tasty-animals-scream.md
+++ b/.changeset/tasty-animals-scream.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Compiler downloader retires download in case of failure [7031](https://github.com/NomicFoundation/hardhat/issues/7031).
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
@@ -32,6 +32,8 @@ import { NativeCompiler, SolcJsCompiler } from "./compiler.js";
 const log = debug("hardhat:solidity:downloader");
 
 const COMPILER_REPOSITORY_URL = "https://binaries.soliditylang.org";
+const DEFAULT_COMPILER_DOWNLOAD_RETRY_COUNT = 3;
+const DEFAULT_COMPILER_DOWNLOAD_RETRY_DELAY_MS = 2000;
 
 // We use a mirror of nikitastupin/solc because downloading directly from
 // github has rate limiting issues
@@ -150,7 +152,6 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
   readonly #platform: CompilerPlatform;
   readonly #compilersDir: string;
   readonly #downloadFunction: typeof download;
-  readonly #retryCount: number;
 
   readonly #mutexCompiler = new MultiProcessMutex("compiler-download");
   readonly #mutexCompilerList = new MultiProcessMutex("compiler-download-list");
@@ -161,13 +162,11 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
   constructor(
     platform: CompilerPlatform,
     compilersDir: string,
-    retryCount: number = 3,
     downloadFunction: typeof download = download,
   ) {
     this.#platform = platform;
     this.#compilersDir = compilersDir;
     this.#downloadFunction = downloadFunction;
-    this.#retryCount = retryCount;
   }
 
   public async updateCompilerListIfNeeded(
@@ -215,17 +214,23 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
       const build = await this.#getCompilerBuild(version);
 
       let downloadPath: string = "";
-      for (let i = 0; i <= this.#retryCount; i++) {
+      for (let i = 0; i <= DEFAULT_COMPILER_DOWNLOAD_RETRY_COUNT; i++) {
         try {
           downloadPath = await this.#downloadAndVerifyCompiler(build);
           break;
         } catch (e) {
-          if (i === this.#retryCount) {
+          if (i === DEFAULT_COMPILER_DOWNLOAD_RETRY_COUNT) {
             ensureError(e);
             throw e;
           } else {
+            const attempt = i + 1;
+
             log(
-              `Download or verification failed for solc ${version}, retrying`,
+              `Download or verification failed for solc ${version}, retrying (attempt ${attempt} of ${DEFAULT_COMPILER_DOWNLOAD_RETRY_COUNT})`,
+            );
+
+            await new Promise((resolve) =>
+              setTimeout(resolve, DEFAULT_COMPILER_DOWNLOAD_RETRY_DELAY_MS),
             );
           }
         }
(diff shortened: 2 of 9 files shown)


.changeset/tasty-animals-scream.md
ChristopherDedominici, line 1: Typo `retires` -> `retries `

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -211,29 +214,21 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
...
-      const verified = await this.#verifyCompilerDownload(build, downloadPath);
-      if (!verified) {
-        throw new HardhatError(
-          HardhatError.ERRORS.CORE.SOLIDITY.INVALID_DOWNLOAD,
-          {
-            remoteVersion: build.longVersion,
-          },
-        );
+      let downloadPath: string = "";
+      for (let i = 0; i <= this.#retryCount; i++) {
```
ChristopherDedominici: maybe adding a small delay between retires could be useful? Maybe with exponential backoff

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -211,29 +214,21 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
...
+        try {
+          downloadPath = await this.#downloadAndVerifyCompiler(build);
+          break;
+        } catch (e) {
+          if (i === this.#retryCount) {
+            ensureError(e);
+            throw e;
+          } else {
+            log(
+              `Download or verification failed for solc ${version}, retrying`,
```
ChristopherDedominici: minor: maybe mention the current attempt number for the retry

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -160,11 +161,13 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
   constructor(
     platform: CompilerPlatform,
     compilersDir: string,
+    retryCount: number = 3,
```
ChristopherDedominici: I noticed that 3 is repeated several times, maybe it can stored in a const? 
Example: DEFAULT_COMPILER_DOWNLOAD_RETRY_COUNT

ChristopherDedominici: Adding `retryCount` before `downloadFunction` forces all test mocks to explicitly pass a retry count value. Looking at the test changes, every mock now needs to pass 0 or 3 as the third argument. Can we move it after?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -150,6 +150,7 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
   readonly #platform: CompilerPlatform;
```
ChristopherDedominici, line 152: I'm wondering if it makes sense to expose `retryCount` and not just use a predefined value. I don't think it's very important across the code to decide how many times to retry a download, I think it's ok if it is a decision that is centralized. Unless we have specific reasons I missed. Wdyt?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
michalbrabec: I've noticed the same, but the other options do the same.
I'll explore if adding a constant would be possible across all places, without adding too much additional complexity

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
michalbrabec: That would simplify the code for sure
One consideration was test performance - in case the solc server is down or generally unavailable, then many tests will be retrying multiple times before failing, which could block them for significant time.
I am happy to change the option into an env variable or a global option in the hardhat config or drop it completely and hardcode the number in the downloader.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
michalbrabec: That was my initial intention, but that forced me to pass the downloader function in cases where I wanted the count explicitly set, which was similarly annoying.
Let's settle how the value will be set before deciding, if it stays as an option or will change into something else. The solution might invalidate this concern.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
kanej: I think we should add a small delay but keep it constant time.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
kanej: Agreed.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
kanej: Out of band we discussed this and opted to not expose the option.
We will default it to 3 with the likely gain in stability our users will see, but we will not expose the retry to user manipulation.

ChristopherDedominici (approved):

Issue #7865, opened 2026-01-14
gas snapshots should show added and removed tests


kanej:
### Describe the issue

Whether a snapshot check passes or fails, the added and removed tests should be shown:

```shell
Gas snapshot check failed: N function(s) changed

  ContractName#standardTestFunctionSig
    (in test/contracts/ContractName.t.sol:34)
    Expected (gas): <value>
    Actual (gas):   <value> (+<percent>%, Δ+<diff>)
    
Added M function(s):
  + ContractName#standardTestFunctionSig (gas: <gas>)
  + ContractName#fuzzTestFunctionSig (runs: <runs>, μ: <avg>, ~: <median>)

Removed N function(s):
  - ContractName#standardTestFunctionSig (gas: <gas>)
  - ContractName#fuzzTestFunctionSig (runs: <runs>, μ: <avg>, ~: <median>)
    
To update snapshots, run your tests with --snapshot
```

The problem is that the removal or addition of a test to a contract will change the gas usage of the other tests. So gas changes will always happen.

Issue #7773, opened 2025-12-09
Check snapshot cheatcodes with `--snapshot-check` flag


schaable:
### Describe the issue

**Tasks:**
- Make `--snapshot-check` validate `snapshots/` files
- Return non-zero exit code on mismatch
- Create missing `snapshots/*.json` files instead of failing

PR #7854, opened 2026-01-09
Write snapshot cheatcodes to `snapshots/` with `--snapshot` flag

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7772

Should be reviewed after https://github.com/NomicFoundation/hardhat/pull/7840

PR #7880, opened 2026-01-19
Check snapshot cheatcodes with `--snapshot-check` flag

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7773

Should be reviewed after https://github.com/NomicFoundation/hardhat/pull/7854

Documentation PR: https://github.com/NomicFoundation/hardhat-website/pull/201

This PR also resolves https://github.com/NomicFoundation/hardhat/issues/7865.


v-next/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -164,250 +170,1650 @@ describe("solidity-test/task-action (override in gas-analytics/index)", () => {
...
+
+          const text = getLoggerOutput();
+          const expected = `
+Snapshot check passed
+
+Snapshot cheatcodes: 1 added, 2 removed
+
+  Added 1 snapshot(s):
+    + TestGroup#test-entry: 42
+
```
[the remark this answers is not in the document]
schaable: These tests validate the full output and, as such, do not modify the expected output. The empty line is intentional (and expected).

.changeset/brown-pandas-film.md
```diff
@@ -0,0 +1,7 @@
+---
+"@nomicfoundation/example-project": patch
```
kanej:
```suggestion
```
We should avoid the example project in changesets, even if it has changed. It is private and can't be released - so better its version remains forever the same.

v-next/example-project/.gas-snapshot
```diff
@@ -2,8 +2,8 @@ CalculatorTest#testAddition() (gas: 62074)
 CalculatorTest#testDivision() (gas: 60060)
 CalculatorTest#testDivisionByZero() (gas: 9923)
 CalculatorTest#testFuzzAddition(uint128,uint128) (runs: 256, μ: 58490, ~: 58490)
-CalculatorTest#testFuzzDivision(uint256,uint256) (runs: 256, μ: 37453, ~: 18953)
-CalculatorTest#testFuzzMultiplication(uint128,uint128) (runs: 256, μ: 57199, ~: 58599)
+CalculatorTest#testFuzzDivision(uint256,uint256) (runs: 256, μ: 37298, ~: 18953)
```
kanej, line 5: I noticed this in the last PR - was there a change to the tests that tweaked the gas usage (rather than a change to the calculation logic)?

.changeset/brown-pandas-film.md
schaable: I've missed that line. Good catch.

v-next/example-project/.gas-snapshot
schaable: I think there was, but I’m not completely sure. We should keep an eye on it, since it should stay the same after setting the fuzz seed in the config.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/snapshot-cheatcodes.ts
```diff
@@ -72,3 +105,271 @@ export async function writeSnapshotCheatcodes(
...
+      return {
+        passed: true,
+        comparison: {
+          added: [],
+          removed: [],
+          changed: [],
+        },
+        written: true,
+      };
+    }
```
kanej, line 247: If I follow the code, this can only happen if the snapshots directory doesn't exist. What do you think about pre-checking for the directory and handling this case rather than using exception handling?

v-next/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -131,21 +131,27 @@ describe("solidity-test/task-action (override in gas-analytics/index)", () => {
   });
 
   describe("logSnapshotResult", () => {
+    let output: string[] = [];
+    const logger = (...args: any[]) => output.push(args.join(""));
+    const getLoggerOutput = (): string =>
+      // Remove ANSI escape codes for color and formatting
+      output.join("\n").replace(/\x1b\[[0-9;]*m/g, "");
```
kanej, line 138: Nice trick. We could use that in the Ignition UI tests.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/snapshot-cheatcodes.ts
schaable: Mainly to mimic the function gas snapshots logic. I don’t mind either way and I’m happy to change it if you believe pre-checking is better.

v-next/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -164,250 +170,1650 @@ describe("solidity-test/task-action (override in gas-analytics/index)", () => {
...
+
+        await handleSnapshot(tmpDir, initialResults, true);
+
+        const { snapshotCheatcodesCheck } = await handleSnapshotCheck(
+          tmpDir,
+          changedResults,
+        );
+
+        assert.equal(snapshotCheatcodesCheck.passed, false);
+        assert.equal(snapshotCheatcodesCheck.written, false);
```
kanej, line 394: I hadn't thought about this case, but reading through the design doc I think this is what the spec says.

v-next/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
schaable: I’d like to move this to `hardhat-utils` or `hardhat-test-utils`, but we can do that later.

v-next/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -164,250 +170,1650 @@ describe("solidity-test/task-action (override in gas-analytics/index)", () => {
...
+      logSnapshotCheckResult(result, logger);
+
+      const text = getLoggerOutput();
+      assert.match(text, /Snapshot check passed/);
+      assert.match(text, /Snapshot cheatcodes: 1 removed/);
+      assert.match(text, /Removed 1 snapshot\(s\):/);
+      assert.match(text, /- TestGroup#test-entry: 42/);
+    });
+
+    describe("full output", () => {
```
kanej, line 795: Well now you are just showing off 😄

v-next/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
schaable: Yeah, if there’s a failure, there’s no easy way to update the snapshots for added or removed entries. It would be nice, but it doesn’t add much value here, since the user will likely rerun the tests with `--snapshot` or fix the tests so the snapshots match. In that case, added or removed snapshots will be updated on the next `--snapshot-check` run.

kanej: Sounds good.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/snapshot-cheatcodes.ts
kanej: I'll leave it up to you - I prefer to avoid flow control via exception handling, but I think it is well handled here.

kanej (approved):
Looks good. The output tests are a thing of beauty.

I did a quick round of manual tests and the output all looks good. We can do a final round with the docs PR.

v-next/hardhat/src/internal/builtin-plugins/gas-analytics/snapshot-cheatcodes.ts
schaable: Makes sense. I'll merge it as-is for now to unblock this, and we can adjust the approach later if needed. Thanks for the input ❤️

Issue #7864, opened 2026-01-14
gas snapshots should inlcude a link to the test on failure


kanej:
### Describe the issue

If the `--snapshot-check` gas check shows failures per test, but it is not easy to navigate to the test:

```shell
CalculatorTest#testFuzzSubtraction(uint256,uint256)
    Runs: 256
    Expected (~): 58729
    Actual (~):   58755 (+0.04%, Δ+26)
```

Instead we should add a link:

```shell
CalculatorTest#testFuzzSubtraction(uint256,uint256)
    (in test/contracts/Calculator.t.sol:34) # <-- additional line
    Runs: 256
    Expected (~): 58729
    Actual (~):   58755 (+0.04%, Δ+26)
```

The line should be gray to de-emphasize it relative to the title.

PR #7883, opened 2026-01-21
Gas snapshots - Show a link to the test file on failures

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7864.

Since EDR doesn’t provide the function line, we agreed to show only the source.


ChristopherDedominici (approved):
LGTM!

Issue #7863, opened 2026-01-14
gas snapshot should use a default seed for fuzz tests


kanej:
### Describe the issue

Enhance the `--snapshot` option to use a default seed for fuzz tests if one is not provided in the `hardhat.config.ts` file.

This means that a user can run `npx hardhat test solidity --snapshot` twice and get the same `.gas-snapshot` file with no further modification of `hardhat.config.ts`.

PR #7886, opened 2026-01-22
feat: add a default fuzz test seed when none is provided

schaable:
Closes https://github.com/NomicFoundation/hardhat/issues/7863


v-next/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -169,3 +175,12 @@ export async function resolveSolidityTestUserConfig(
     },
   };
 }
+
+export function resolveFuzzConfig(
```
ChristopherDedominici, line 179: Not a blocker, but I’m wondering about the following:

the type `SolidityTestConfig["fuzz"]` defines `seed` as `string | undefined`. Since `seed` is always defined here, I’m wondering whether enforcing it as defined would improve the developer experience.

I don’t have full knowledge of the gas feature, so this may not be very impactful. Feel free to disregard this comment in that case

ChristopherDedominici (approved):
I left a comment, but other than that LGTM

v-next/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
schaable: I agree, good catch!

schaable: Done [here](https://github.com/NomicFoundation/hardhat/pull/7886/commits/b191531ee42a810b41ccd3b6c50bb27d2d513976). @kanej this is the change we discussed on Slack. It ended up being a bit more complex than expected, since the fuzz config is now always resolved to an object containing a seed string. I’m going to merge this into the feature branch. Let me know if you want me to tweak the commit in any way.

PR #7889, opened 2026-01-23
Error when testing coverage: html folders flaky creation

### LINKED ISSUE #7887: Error when testing coverage
### Version of Hardhat

3.1.5

### What happened?

When I execute `npx hardhat test --coverage` all tests are passing but when it tries to show the Coverage Report if fails with error
`Error: ENOENT: no such file or directory, open '.../coverage/html/contracts/payment/index.html'`

Version 3.1.3 is working fine.  But if I install version 3.1.5 is failing with this error.

Execute `npx hardhat test --coverage`

### Search terms

hardhat coverage report

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/7887#issuecomment-3785495439.

Reason: the `mkdir` function was not awaited in our `vendored-package`, so nested paths were not created in time, causing the error. The root path was created automatically, which hid the bug and caused it to surface only for nested folders.

Solution: await the creation of all folders before proceeding.

### DIFF
--- a/v-next/hardhat-vendored/src/coverage-module/istanbul-lib-report/lib/file-writer.cjs
+++ b/v-next/hardhat-vendored/src/coverage-module/istanbul-lib-report/lib/file-writer.cjs
@@ -8,7 +8,6 @@
  */
 const path = require("node:path");
 const fs = require("node:fs");
-const { mkdir } = require("node:fs/promises");
 
 /**
  * Base class for writing content
@@ -155,7 +154,7 @@ class FileWriter {
       throw new Error(`Cannot write to absolute path: ${dest}`);
     }
     dest = path.resolve(this.baseDir, dest);
-    mkdir(path.dirname(dest), { recursive: true });
+    fs.mkdirSync(path.dirname(dest), { recursive: true });
     let contents;
     if (header) {
       contents = header + fs.readFileSync(source, "utf8");
@@ -179,7 +178,7 @@ class FileWriter {
       throw new Error(`Cannot write to absolute path: ${file}`);
     }
     file = path.resolve(this.baseDir, file);
-    mkdir(path.dirname(file), { recursive: true });
+    fs.mkdirSync(path.dirname(file), { recursive: true });
     return new FileContentWriter(fs.openSync(file, "w"));
   }
 }
--- a/v-next/hardhat/src/internal/builtin-plugins/coverage/reports/html.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/coverage/reports/html.ts
@@ -3,7 +3,6 @@ import type { FileCoverageData } from "@nomicfoundation/hardhat-vendored/coverag
 
 import path from "node:path";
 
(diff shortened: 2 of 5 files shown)


v-next/hardhat/src/internal/builtin-plugins/coverage/reports/html.ts
```diff
@@ -17,8 +17,6 @@ export async function generateHtmlReport(
   const baseDir = process.cwd();
   const coverageMap = istanbulLibCoverage.createCoverageMap({});
 
-  await mkdir(htmlReportPath);
```
ChristopherDedominici, line 20: all folders now are created thanks to the fix in the above file `v-next/hardhat-vendored/src/coverage-module/istanbul-lib-report/lib/file-writer.cjs` so there is no need for this line anymore

v-next/hardhat/test/internal/builtin-plugins/coverage/coverage-manager.ts
```diff
@@ -490,3 +494,45 @@ describe("CoverageManagerImplementation - report data processing", () => {
...
+    const coveragePath = path.join(process.cwd(), "coverage");
+    const htmlIndexPath = path.join(coveragePath, "html", "index.html");
+
+    await remove(coveragePath);
+
+    await coverageManagerTmp.report();
+
+    const htmlContent = await readUtf8File(htmlIndexPath);
+    assert.ok(htmlContent.length > 0, "HTML report should have content");
+  });
```
kanej, line 552: Should this be expanded to check that a representative sample of the indexes and standard html files in subfolders are also created?

v-next/hardhat-vendored/src/coverage-module/istanbul-lib-report/lib/file-writer.cjs
```diff
@@ -155,7 +154,7 @@ class FileWriter {
       throw new Error(`Cannot write to absolute path: ${dest}`);
     }
     dest = path.resolve(this.baseDir, dest);
-    mkdir(path.dirname(dest), { recursive: true });
+    fs.mkdirSync(path.dirname(dest), { recursive: true });
```
kanej, line 157: If I understand correctly. There is a bug here in the vendor file as it mixes sync and async folder/file manipulation methods.
We overcame that in our code by forcing the root folder to exist before using `copyFile` from the vendored code.
This failed when there were subfolders beneath the root folder.
The fix is to fix the vendor file to using sync methods (which it already did) for directory creation, and remove the now unneeded guard code on our side?
I think this approach works.

kanej (changes requested):
Can you double check the test, maybe I missed the check for subfolders.
I think this PR looks good. I did a manual test that subfolders are created based on Luis reproduction steps.

ChristopherDedominici:
Test improved here: https://github.com/NomicFoundation/hardhat/pull/7889/commits/8760aa0a3db0b6a20133629bb7207aa6eabda434

@kanej

kanej (approved):

PR #7897, opened 2026-01-26
fix(hardhat-ledger): Reconnect when device is unplugged

### LINKED ISSUE #7896: [hardhat-ledger] Verify if the underlying connection is still active before failing, and reconnect if needed
### Version of Hardhat

3.1.5

### What happened?

Issue reported by @zappso on telegram. hardhat-ledger does not recover in cases where the ledger was locked, or where the ledger was unplugged and plugged back in.

From the user report: 

Some more info... it seems Hardhat only connects to the Ledger when signing the first transaction.

If the Ledger is then physically disconnected/reconnected, Hardhat just tries to have the transaction signed without reconnecting.

<ignition deploy>

[hardhat-ledger] Connecting to Ledger...
[hardhat-ledger] Connection successful
eth_chainId
[hardhat-ledger] Confirmation start
[hardhat-ledger] Confirmation success
eth_sendRawTransaction
  Contract deployment: MyContract
  Contract address:    0x88a5cf2849efdce0830966f63eb362107aad8e1f
  Transaction:         0x5f12c939a089e7f482615632c5dc1ab32b1f81a5cee5918b78a195bbb4a0c486
  From:                0xdfdfe09aa2a7c33aecf65098f69ef819222562d5
  Value:               0 ETH
  Gas used:            3035878 of 3035878
  Block #2:            0xa33fa90157191c958431ab54740445a6495098265e97d07aa1f33ee5dc699807

eth_getTransactionByHash
eth_getBlockByNumber
eth_getTransactionReceipt

<physically disconnect / reconnect the Ledger>

<send a mint transaction>

eth_accounts
eth_chainId
eth_accounts
eth_blockNumber
eth_estimateGas
eth_getBlockByNumber
eth_feeHistory
eth_getTransactionCount
[hardhat-ledger] Confirmation start
[hardhat-ledger] Confirmation failure

User reported: 

If the Ledger is disconnected and reconnected... then signing no longer works properly, Hardhat Node shows Confirmation Failure, and my hh task throws an Internal Error.

Sometimes after the error, the Ledger goes through its sign data dialogs.. and sometimes Hardhat Node shows Confirmation Success even though the tx has not yet been signed.

This behaviour also sometimes happens if the Ledger times out from inactivity and locks, requiring an unlock.

Also, sometimes a signature will work again.

I don't currently have the time to draft up minimal steps to reproduce, but you should be able to hit the error by disconnecting and reconnecting, then trying to sign a tx.

### Search terms

_No response_

marianfe:
Fixes #7896

## Summary

When a Ledger device is unplugged/replugged during a Hardhat session, subsequent signing operations would fail because the plugin kept trying to use a stale connection.

This PR adds automatic reconnection logic that:
- Detects `DisconnectedDevice` and `DisconnectedDeviceDuringOperation` errors from the transport layer
- Closes the stale transport and clears the cached connection
- Reconnects and retries the operation (once)

Added unit tests. This required updating the mock to enable introducing connection failures after a given number of calls. Also manually tested to ensure the fix works as expected. 

Note: `LockedDeviceError` cases are for now handled by existing code, which shows a user-friendly error asking to unlock the device and try again. Will address handling this in a subsequent PR.

### DIFF
--- a/.changeset/bhham-rsdem-drzyn.md
+++ b/.changeset/bhham-rsdem-drzyn.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ledger": patch
+---
+
+Fixed reconnection when Ledger device is unplugged/replugged during operations ([#7896](https://github.com/NomicFoundation/hardhat/issues/7896)).
--- a/v-next/hardhat-ledger/src/internal/handler.ts
+++ b/v-next/hardhat-ledger/src/internal/handler.ts
@@ -7,6 +7,10 @@ import type {
   JsonRpcResponse,
 } from "hardhat/types/providers";
 
+import {
+  DisconnectedDevice,
+  DisconnectedDeviceDuringOperation,
+} from "@ledgerhq/errors";
 import { isEIP712Message } from "@ledgerhq/evm-tools/lib/index";
 import Eth, { ledgerService } from "@ledgerhq/hw-app-eth";
 import TransportNodeHid from "@ledgerhq/hw-transport-node-hid";
@@ -30,6 +34,7 @@ import {
   rpcTransactionRequest,
   validateParams,
 } from "@nomicfoundation/hardhat-zod-utils/rpc";
+import debug from "debug";
 import { Transaction } from "micro-eth-signer";
 import * as typed from "micro-eth-signer/typed-data";
 import { add0x, initSig } from "micro-eth-signer/utils";
@@ -40,9 +45,12 @@ import { getYParity } from "./get-y-parity.js";
 import { PLUGIN_NAME } from "./plugin-name.js";
 import { getRequestParams } from "./rpc-helpers.js";
 
+const log = debug("hardhat:hardhat-ledger:handler");
+
 export class LedgerHandler {
   public static readonly MAX_DERIVATION_ACCOUNTS = 20;
   public static readonly DEFAULT_TIMEOUT = 3000;
+  public static readonly MAX_RECONNECTION_ATTEMPTS = 1;
 
   readonly #provider: EthereumProvider;
   readonly #displayMessage: (message: string) => Promise<void>;
@@ -269,6 +277,7 @@ export class LedgerHandler {
 
   async #derivePath(
     addressToFindAsBuffer: Uint8Array<ArrayBufferLike>,
+    reconnectionAttempts: number = 0,
   ): Promise<string> {
     const addressToFind = bytesToHexString(addressToFindAsBuffer).toLowerCase();
 
@@ -312,6 +321,25 @@ export class LedgerHandler {
     } catch (error) {
       ensureError(error);
 
+      // Check if we should attempt reconnection
+      if (
+        this.#isReconnectableError(error) &&
+        reconnectionAttempts < LedgerHandler.MAX_RECONNECTION_ATTEMPTS
+      ) {
+        log("Reconnectable error during path derivation, attempting reconnect");
+        log(error);
+
+        await this.#displayMessage("Reconnecting to Ledger...");
+        await this.#resetConnection();
+        await this.init();
+
+        // Retry derivation
+        return this.#derivePath(
+          addressToFindAsBuffer,
+          reconnectionAttempts + 1,
+        );
+      }
+
       await this.#displayMessage("Derivation failure");
 
       throw new HardhatError(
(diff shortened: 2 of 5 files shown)


v-next/hardhat-ledger/src/internal/handler.ts
```diff
@@ -343,8 +366,40 @@ export class LedgerHandler {
...
+  /**
+   * Resets the Ledger connection by closing the transport and clearing the eth instance.
+   * This allows the next init() call to create a fresh connection.
+   */
+  async #resetConnection(): Promise<void> {
+    if (this.#eth !== undefined) {
+      try {
+        await this.#eth.transport.close();
+      } catch (_error) {
+        // Transport may already be closed/disconnected, ignore
```
alcuadrado: Here I would use `debug` to log the error. Here's an example of how to do it: https://github.com/NomicFoundation/hardhat/blob/12340060382ff7ddfd93390b0bcfc9abc4311360/v-next/hardhat-verify/src/internal/etherscan.ts#L36

alcuadrado: And how to see the debug logs: https://hardhat.org/docs/cookbook/debug-logs

v-next/hardhat-ledger/test/helpers/eth-mocked.ts
```diff
@@ -103,14 +133,30 @@ export function getEthMocked(
...
-        assert.equal(
-          path,
-          this.#methodsConfig.signPersonalMessage.expectedParams.path,
-        );
-        assert.equal(
-          messageHex,
-          this.#methodsConfig.signPersonalMessage.expectedParams.data,
-        );
+        const c = calls.get("signPersonalMessage");
+        assertHardhatInvariant(c !== undefined, "c should be defined");
```
alcuadrado: While `assertHardhatInvariant` works for tests, we normally only use them for production code, to check true invariants, and not as a test assertion.

`assert.equal` and throw better errors, that are displayed in a nicer way by test reporters.

`assertHardhatInvariant`'s errors are optimized for the Hardhat CLI and we report them through Sentry

v-next/hardhat-ledger/test/internal/handler.ts
```diff
@@ -1090,4 +1098,311 @@ describe("LedgerHandler", () => {
...
+
+        // Verify initial state
+        assert.equal(
+          transportState.createCount,
+          1,
+          "Transport should be created once initially",
+        );
+        const signCalls = calls.get("signPersonalMessage");
+        assert.equal(
+          signCalls?.totalCalls,
```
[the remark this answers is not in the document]
alcuadrado: Ignore this message. We don't use `!` in the codebase actually. If we find the need to use it, we use `assertHardhatInvariante`, so that we get a better error message instead of an undefined dereference error.

In this case in particular, `?` is the right call, as the test won't pass if the expression results in `undefined`.

v-next/hardhat-ledger/test/helpers/eth-mocked.ts
```diff
@@ -54,21 +60,33 @@ export interface MethodsConfig {
...
     };
+    /** If set, throws errorToThrow when call count equals this value */
+    throwOnCall?: number;
+    errorToThrow?: Error;
   };
 }
 
+export interface MockCallState {
+  totalCalls: number;
+  args: any[];
```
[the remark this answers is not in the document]
marianfe: fixed.

v-next/hardhat-ledger/test/internal/handler.ts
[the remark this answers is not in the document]
marianfe: thanks, agree, i've already ignored this one. fixed only the above other one.

v-next/hardhat-ledger/src/internal/handler.ts
marianfe: agree, fixing now.

v-next/hardhat-ledger/test/helpers/eth-mocked.ts
marianfe: agree, thanks for spotting this, fixing now.

.changeset/bhham-rsdem-drzyn.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ledger": patch
+---
+
+Fixed reconnection when Ledger device is unplugged/replugged or auto-locks during operations ([#7896](https://github.com/NomicFoundation/hardhat/issues/7896)).
```
alcuadrado:
```suggestion
Fixed reconnection when Ledger device is unplugged/replugged ([#7896](https://github.com/NomicFoundation/hardhat/issues/7896)).
```

I think this would be the updated changeset now that you are splitting your pr into two

alcuadrado (approved):

.changeset/bhham-rsdem-drzyn.md
marianfe: agree, fixing now.

PR #7898, opened 2026-01-27
chore: add "no-downgrade" as pnpm trust policy

kanej:
Update to the latest pnpm to take advantage of the new trust policy setting.

This guards against malicious npm releases by insisting that a package that was previously published with trusted publising can't be installed if it reverts to standard publishing (an indication of stolen publishing credentials).

See https://pnpm.io/settings#trustpolicy

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -2,3 +2,4 @@ hoist-pattern[]=!@types/*
 minimum-release-age=10080
 minimum-release-age-exclude[]="hardhat"
 minimum-release-age-exclude[]="@nomicfoundation/*"
+trust-policy=no-downgrade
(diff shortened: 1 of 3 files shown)


.npmrc
```diff
@@ -2,3 +2,4 @@ hoist-pattern[]=!@types/*
 minimum-release-age=10080
 minimum-release-age-exclude[]="hardhat"
 minimum-release-age-exclude[]="@nomicfoundation/*"
+trust-policy=no-downgrade
```
kanej, line 5: I initially added it to the `pnpm-workspace.yaml` file but wanted to keep the security related pnpm settings together.

alcuadrado (approved):

.npmrc
alcuadrado: makes sense

PR #7900, opened 2026-01-27
Build system fixes

alcuadrado:
This PR introduces three fixes to the build system, one per commit:

1. It fixes some inconsistencies in the SolidityBuildSystem API and its usage. We were mixing userSourceNames and rootFilepaths. This requires updating `hardhat-verify` in tandem, but I think we should fix this while we are still in a beta. The code was pretty hard to reason about with the inconsistencies in place.

2. It restructures how we keep track of cache hits, and returns them in `SolidityBuildSystem#build`. According to its type, we were returning the hits, but in practice that never happened. I had to update the type we use to represent cache hits to make it feasible and performant, which in theory is a breaking change, but again, this type was never used in practice.

3. Updated how we clean up artifacts after a complete build in the build task, as it was preventing the `Solidity#build` hook from adding/removing root files.

### DIFF
--- a/v-next/hardhat-verify/src/internal/artifacts.ts
+++ b/v-next/hardhat-verify/src/internal/artifacts.ts
@@ -103,11 +103,7 @@ export async function getCompilerInput(
     "The compilation job for the contract source was not found.",
   );
 
-  const compilationJobKey = isNpmModule ? `npm:${sourceName}` : sourceName;
-
-  const compilerInput = await compilationJob
-    .get(compilationJobKey)
-    ?.getSolcInput();
+  const compilerInput = await compilationJob.get(rootFilePath)?.getSolcInput();
 
   // TODO: should this be an error instead?
   assertHardhatInvariant(
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -15,6 +15,7 @@ import type {
   EmitArtifactsResult,
   RunCompilationJobResult,
   BuildScope,
+  CacheHitInfo,
 } from "../../../../types/solidity/build-system.js";
 import type {
   CompilationJob,
@@ -93,7 +94,6 @@ export const SUPPRESSED_WARNINGS: Array<{
 interface CompilationResult {
   compilationJob: CompilationJob;
   compilerOutput: CompilerOutput;
-  cached: boolean;
   compiler: Compiler;
 }
 
@@ -224,7 +224,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     spinner.start();
 
     try {
-      const { compilationJobsPerFile, indexedIndividualJobs } =
+      const { compilationJobsPerFile, indexedIndividualJobs, cacheHits } =
         compilationJobsResult;
 
       const runnableCompilationJobs = [
@@ -250,7 +250,6 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
           return {
             compilationJob: runnableCompilationJob,
             compilerOutput: output,
-            cached: false,
             compiler,
           };
         },
@@ -262,13 +261,11 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
         },
       );
 
-      const uncachedResults = results.filter((result) => !result.cached);
-      const uncachedSuccessfulResults = uncachedResults.filter(
+      const successfulResults = results.filter(
         (result) => !this.#hasCompilationErrors(result.compilerOutput),
       );
 
-      const isSuccessfulBuild =
-        uncachedResults.length === uncachedSuccessfulResults.length;
+      const isSuccessfulBuild = results.length === successfulResults.length;
 
       const contractArtifactsGeneratedByCompilationJob: Map<
         CompilationJob,
@@ -347,18 +344,6 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
             continue;
           }
 
-          if (result.cached) {
-            resultsMap.set(formatRootPath(userSourceName, root), {
-              type: FileBuildResultType.CACHE_HIT,
-              compilationJob: result.compilationJob,
-              contractArtifactsGenerated:
-                contractArtifactsGenerated.get(userSourceName) ?? [],
-              warnings: errors,
-            });
-
-            continue;
-          }
-
           resultsMap.set(formatRootPath(userSourceName, root), {
             type: FileBuildResultType.BUILD_SUCCESS,
             compilationJob: result.compilationJob,
@@ -367,6 +352,15 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
             warnings: errors,
           });
         }
+      }
+
+      // Add cache hits to the results map
+      for (const [rootFilePath, cacheHitInfo] of cacheHits.entries()) {
+        resultsMap.set(rootFilePath, {
+          type: FileBuildResultType.CACHE_HIT,
+          buildId: cacheHitInfo.buildId,
+          contractArtifactsGenerated: cacheHitInfo.artifactPaths,
+        });
       }
 
       if (!options.quiet) {
@@ -485,6 +479,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
 
     // Select which files to compile
     const rootFilesToCompile: Set<string> = new Set();
+    const cacheHits: Map<string, CacheHitInfo> = new Map();
 
     const isolated = buildProfile.isolated;
 
@@ -533,6 +528,19 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
           rootFilesToCompile.add(rootFile);
           break;
         }
+      }
+
+      // If file was not added to rootFilesToCompile, it's a cache hit
+      if (!rootFilesToCompile.has(rootFile)) {
+        const [userSourceName, root] = [
+          ...compilationJob.dependencyGraph.getRoots().entries(),
+        ][0];
+        // Extract buildId from buildInfoPath (format: <dir>/<buildId>.json)
+        const buildId = path.basename(cacheResult.buildInfoPath, ".json");
+        cacheHits.set(formatRootPath(userSourceName, root), {
+          buildId,
+          artifactPaths,
+        });
       }
     }
 
@@ -614,7 +622,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       }
     }
 
-    return { compilationJobsPerFile, indexedIndividualJobs };
+    return { compilationJobsPerFile, indexedIndividualJobs, cacheHits };
   }
 
   #getBuildProfile(buildProfileName: string = DEFAULT_BUILD_PROFILE) {
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -22,6 +22,7 @@ import type {
   Compiler,
   CompilerOutput,
   CompilerOutputError,
+  DependencyGraph,
   SolidityBuildInfo,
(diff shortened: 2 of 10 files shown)


alcuadrado:
This can be reviewed one commit at the time

v-next/hardhat-verify/src/internal/artifacts.ts
```diff
@@ -103,11 +103,7 @@ export async function getCompilerInput(
     "The compilation job for the contract source was not found.",
   );
 
-  const compilationJobKey = isNpmModule ? `npm:${sourceName}` : sourceName;
-
-  const compilerInput = await compilationJob
-    .get(compilationJobKey)
-    ?.getSolcInput();
+  const compilerInput = await compilationJob.get(rootFilePath)?.getSolcInput();
```
alcuadrado, line 106: I used claude to analyze all the places that needed to be updated

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -250,7 +250,6 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
           return {
             compilationJob: runnableCompilationJob,
             compilerOutput: output,
-            cached: false,
```
alcuadrado, line 253: This was always false, so unnecessary

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -262,13 +261,11 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
         },
       );
 
-      const uncachedResults = results.filter((result) => !result.cached);
-      const uncachedSuccessfulResults = uncachedResults.filter(
+      const successfulResults = results.filter(
```
alcuadrado, line 283: Same here, `.cached` was always false

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -577,7 +581,8 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
             "there should be only 1 root file on subgraph",
           );
 
-          const rootFile = Array.from(subgraph.getRoots().keys())[0];
+          const [userSourceName, root] = [...subgraph.getRoots().entries()][0];
+          const rootFile = formatRootPath(userSourceName, root);
```
kanej: These two lines should be pulled out as a helper and used to dry up the other examples above.

v-next/hardhat/src/internal/builtin-plugins/solidity/tasks/build.ts
```diff
@@ -114,7 +114,10 @@ async function buildForScope(
 
   // If we recompiled the entire project we cleanup the artifacts
   if (isFullCompilation) {
-    await solidity.cleanupArtifacts(rootPaths, { scope });
+    // Use the root files from the build results, which may include
+    // additional files added by plugins hooking into solidity#build
+    const builtRootPaths = [...results.keys()];
+    await solidity.cleanupArtifacts(builtRootPaths, { scope });
```
kanej, line 120: Not related to this change - can I check my understanding. A root file path can be a NPM uri because of `npmFilesToBuild`.
Artifacts will be generated for `npmFilesToBuild` like a more standard project file, and artifact cleanup should be applied to them as well.

v-next/hardhat/test/internal/builtin-plugins/solidity/tasks/build-cleanup-artifacts.ts
```diff
@@ -0,0 +1,420 @@
...
+                      project.path,
+                      "extra/AddedByHook.sol",
+                    );
+
+                    // Add extra file to the build
+                    const extendedPaths = rootFilePaths.some(
+                      (p) => p === extraFile,
+                    )
+                      ? rootFilePaths
+                      : [...rootFilePaths, extraFile];
```
kanej: Would it not be better here to assert that the extra file should not exist in the `rootFilePaths`, and afterwards add it?

Can I check that the trick here is that the contract under `./extra` won't be included in the build by default?

v-next/hardhat/test/internal/builtin-plugins/solidity/tasks/build-cleanup-artifacts.ts
```diff
@@ -0,0 +1,420 @@
...
+          "Second.sol artifact should exist",
+        );
+        assert.ok(
+          await exists(path.join(artifactsDir, "Third.sol", "Third.json")),
+          "Third.sol artifact should exist",
+        );
+      });
+
+      it("should not run cleanupArtifacts for partial compilation", async () => {
+        await using project = await useTestProjectTemplate({
```
kanej, line 285: I think that is the first use of `using` I have seen!

v-next/hardhat/src/internal/builtin-plugins/solidity/tasks/build.ts
alcuadrado: correct

v-next/hardhat/test/internal/builtin-plugins/solidity/tasks/build-cleanup-artifacts.ts
alcuadrado: > Can I check that the trick here is that the contract under ./extra won't be included in the build by default?

Correct. I can explain that in a comment.

I'll take a look at the other point in a few minutes.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: Agreed re the helper, but I don't understand what "dry up" means in this context.

kanej: Ha, I meant `DRY` as in don't repeat yourself.

v-next/hardhat/src/types/solidity/build-system.ts
```diff
@@ -146,9 +146,8 @@ export type FileBuildResult =
 
 export interface CacheHitFileBuildResult {
   type: FileBuildResultType.CACHE_HIT;
-  compilationJob: CompilationJob;
+  buildId: string;
   contractArtifactsGenerated: string[];
-  warnings: CompilerOutputError[];
 }
```
kanej, line 151: If I have followed, running the build task programmatically will give you a set of root file results.
Previously we returned a cache hit result that was similar to a success build result; it had warnings + a compilation job.
Now we provide only the buildId - how do we expect these cache results to be used? Is it really mainly for our use, and we can handle the reduced returned properties? Or is there a larger use case, but that is satisfied by `buildId` + using compileBuildInfo?

alcuadrado: The main reason for the change is that the compilation job that gets created for each file doesn't necessarily match the one stored in the build info. That lead to a situation where you could use `compilationJob.getBuildId()` of a cache hit, and end up with an non-existing build.

Instead, what we really know, is which is the buildId that was used to compile that root the last time. That info is in the cache storage, so we know it's correct, and that its corresponding buildinfo exists (otherwise we don't consider it a hit).

So by returning the build id, we give the user the capability of reading the actual build info. It's not exactly the same as a compilation job (different api), but conceptually equivalent.

Why don't we return the warnings? Just a performance optimization. If we did, we'd be force to read every build info output of the cache hits.

alcuadrado: We could wrap the build info in a CompilationJob-like API, but that forces us to read it and parse it.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/get-compilation-jobs-cache-hits.ts
```diff
@@ -0,0 +1,255 @@
...
+    assert(!("reason" in result), "getCompilationJobs should succeed");
+    assert.equal(result.cacheHits.size, 1, "Should have one cache hit");
+    assert.equal(
+      result.compilationJobsPerFile.size,
+      0,
+      "Should have no compilation jobs",
+    );
+    assert(
+      result.cacheHits.has(filePath),
+      "Cache hit should be keyed by absolute path",
```
kanej, line 43: Is there another test here for NPM URI paths coming through the cache?

kanej: If the `useTestProjectTemplate` does not support faking `node_modules` I am happy to leave the case out.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/get-compilation-jobs-cache-hits.ts
```diff
@@ -0,0 +1,255 @@
...
+      result.cacheHits.size,
+      0,
+      "Should have no cache hits when dependency changed",
+    );
+    assert.equal(
+      result.compilationJobsPerFile.size,
+      1,
+      "Should need recompilation",
+    );
+  });
```
kanej, line 254: To check why this test passes. The caching is based on the buildId, which is derived from among other things the solc input.
Changing the contents of one root file changes its `buildId`. But it will also change the `buildId` of the Child.sol root file as the contents of its dependency changed, which will still be part of its solc input.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/get-compilation-jobs-cache-hits.ts
kanej: Okay, well it turns out there is a whole file of npm cache tests, so let me sheepishly close this comment.

kanej (approved):
I have left some comments - mainly questions clarifying my own understanding.
There are no blockers so I am pre-approving.

v-next/hardhat/src/types/solidity/build-system.ts
kanej: I am happy with the change and your reasoning.
I had assumed that changes in the compilation job/solc input would necessarily mean a cache miss due to the change in the `buildId`/`jobHash`; that getting a cache hit (successfully comparing `jobHash`) __meant__ that the build info file had exactly the same solc input (and from that I inferred compilation job).

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/get-compilation-jobs-cache-hits.ts
alcuadrado: That's correct.

v-next/hardhat/src/types/solidity/build-system.ts
alcuadrado: Your reasoning is correct for isolatedBuilds, in that case, a cache hit implies the exact same build id and compilation job.

In non-isolated ones, we consider a root file to be cached if it and its transitive dependencies haven't changed since the last time it was compiled. There could be other files in that build (due to compilations jobs getting merged) that change, but that won't invalidate the root file's cache.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/get-compilation-jobs-cache-hits.ts
alcuadrado: Note that actually, Base is not a root file. But a transitive dependency of Child

v-next/hardhat/src/types/solidity/build-system.ts
alcuadrado: I think what can make it confusing is that we do check

```
        cacheResult.jobHash !== jobHash ||
```

but that's the pre-merge/individual compilation job. Which only contains the subgraph needed to build the root. That's what we store in the cache, and not the complete build id that was run.

PR #7902, opened 2026-01-27
Generate Typechain types when compiling a subset of the solidity files

### LINKED ISSUE #6311: [Typechain] modify the plugin to generate types for a subset of contracts

alcuadrado:
This PR is a replacement of https://github.com/NomicFoundation/hardhat/pull/7707

### DIFF
--- a/v-next/hardhat-typechain/src/internal/hook-handlers/solidity.ts
+++ b/v-next/hardhat-typechain/src/internal/hook-handlers/solidity.ts
@@ -1,25 +1,50 @@
 import type { HookContext, SolidityHooks } from "hardhat/types/hooks";
+import type {
+  BuildOptions,
+  CompilationJobCreationError,
+  FileBuildResult,
+} from "hardhat/types/solidity";
 
 import { generateTypes } from "../generate-types.js";
 
 export default async (): Promise<Partial<SolidityHooks>> => {
   const handlers: Partial<SolidityHooks> = {
-    async onCleanUpArtifacts(
+    async build(
       context: HookContext,
-      artifactPaths: string[],
+      rootFilePaths: string[],
+      options: BuildOptions | undefined,
       next: (
         nextContext: HookContext,
-        artifactPaths: string[],
-      ) => Promise<void>,
+        nextRootFilePaths: string[],
+        nextOptions: BuildOptions | undefined,
+      ) => Promise<CompilationJobCreationError | Map<string, FileBuildResult>>,
     ) {
+      const result = await next(context, rootFilePaths, options);
+
+      // Skip if build failed (returned an error)
+      if ("reason" in result) {
+        return result;
+      }
+
+      // Skip for test scope (contracts only)
+      if (options?.scope === "tests") {
+        return result;
+      }
+
+      // Clear cache to ensure fresh data after compilation
+      await context.artifacts.clearCache();
+
+      // Get all artifact paths and generate types
+      const allArtifactPaths = await context.artifacts.getAllArtifactPaths();
+
       await generateTypes(
         context.config.paths.root,
         context.config.typechain,
         context.globalOptions.noTypechain,
-        artifactPaths,
+        Array.from(allArtifactPaths),
       );
 
-      return next(context, artifactPaths);
+      return result;
     },
   };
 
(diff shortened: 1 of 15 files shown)


v-next/hardhat-typechain/src/internal/hook-handlers/solidity.ts
```diff
@@ -1,25 +1,50 @@
...
+        nextRootFilePaths: string[],
+        nextOptions: BuildOptions | undefined,
+      ) => Promise<CompilationJobCreationError | Map<string, FileBuildResult>>,
     ) {
+      const result = await next(context, rootFilePaths, options);
+
+      // Skip if build failed (returned an error)
+      if ("reason" in result) {
+        return result;
+      }
```
kanej, line 27: I am tempted to say we add an `isErrorResult` helper for these situations.

v-next/hardhat-typechain/test/index.ts
```diff
@@ -223,4 +228,143 @@ describe("hardhat-typechain", () => {
       assert.equal(await exists(`${process.cwd()}/types`), false);
     });
   });
+
+  describe("types are generated for partial compilation", () => {
+    const projectFolder = "generate-types";
+
+    useFixtureProject(projectFolder);
+
+    it("should generate types for all artifacts when compiling a single file", async () => {
```
kanej:
```suggestion
    it("should generate types for all artifacts from a single file when compiling a single file", async () => {
```
I found this test confusing on the first run through, as we generate types only for those artifacts we have after compilation, not every contract under the contracts folder as I hazily thought of it.

v-next/hardhat-typechain/test/index.ts
```diff
@@ -223,4 +228,143 @@ describe("hardhat-typechain", () => {
...
+      );
+      assert.equal(
+        await exists(
+          path.join(process.cwd(), "types", "ethers-contracts", "B.ts"),
+        ),
+        false,
+      );
+    });
+
+    it("should generate types for all artifacts when types are deleted and cache is valid", async () => {
```
kanej: To check the case here. We have a build of multiple contracts, we remove the types, then rerun a _partial_ build. Because of the first build the cache is primed with all the contracts - hence artifact cleanup leaves all the artifacts in place. When the type gen happens we get types for all contracts (because we have all artifacts).

v-next/hardhat-typechain/test/index.ts
```diff
@@ -223,4 +228,143 @@ describe("hardhat-typechain", () => {
       assert.equal(await exists(`${process.cwd()}/types`), false);
     });
   });
+
+  describe("types are generated for partial compilation", () => {
```
kanej, line 232: Reading through this set of tests I think we should add one for the happy case:

1. Build everything
2. Remove types
3. Build only a single file
4. Check all types exist

I realise this is very close to the `should generate types for all artifacts when compiling a single file`, so maybe it is unnecessary.

kanej (approved):
I left some comments on the tests.
I have done a quick manual test based on the example project, I think we are good.

v-next/hardhat-typechain/src/internal/hook-handlers/solidity.ts
alcuadrado: Agreed, but I'd say we address it when we fix https://github.com/NomicFoundation/hardhat/issues/7748

v-next/hardhat-typechain/test/index.ts
alcuadrado: It was "should generate types for all artifacts when types are deleted and cache is valid", I renamed it to "should generate types for all the artifacts present in the file system, even if a single file is being compiled" in f9736e6f3

v-next/hardhat-typechain/test/index.ts
alcuadrado: I renamed it to: should generate types for the artifacts of the only file being compiled on a fresh build

v-next/hardhat-typechain/test/index.ts
alcuadrado: yes, see https://github.com/NomicFoundation/hardhat/pull/7902/changes#r2746369284

PR #7904, opened 2026-01-28
Suppress pragma+license warning in Solidity test files

### LINKED ISSUE #7894: supporess pragma+license warning in Solidity test files
The solc compiler will give warnings if a Solidity file, including those that are test files, do not include a valid pragma or license.
Hardhat should suppress those warnings when they come from a test file - but allow them coming from core contracts.

We need to be careful in how we identify Solidity test files. The test files being able to exist in the source directory and test directory depending on config.

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/7894

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
@@ -29,6 +29,7 @@ export const SUPPRESSED_WARNINGS: Array<
     message:
       "Natspec memory-safe-assembly special comment for inline assembly is deprecated and scheduled for removal. Use the memory-safe block annotation instead.",
     scope: "specific-file",
+    // Normalize to handle different OS path separators
     filePath: path.normalize("hardhat/console.sol"),
   },
   {
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
@@ -1,0 +1,69 @@
+import path from "node:path";
+
+// Compiler warnings to suppress from build output.
+// Supports two types of suppression rules:
+//
+// 1. scope: 'specific-file' - Suppress warnings from specific file paths
+//    - Use this to suppress known warnings from internal/library files (e.g., console.sol)
+//    - The same warning type will still be shown for user code
+//
+// 2. scope: 'test-files' - Suppress warnings from all test files
+//    - Test files are identified as:
+//      * Files ending in .t.sol (e.g., Counter.t.sol)
+//      * Files inside test/contracts/ directory
+//        (e.g., test/contracts/Example.sol)
+//    - Use this for warnings that are acceptable in test code but not in production code
+//      (e.g., missing SPDX license identifiers or pragma statements)
+export const SUPPRESSED_WARNINGS: Array<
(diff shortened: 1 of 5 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
```diff
@@ -0,0 +1,70 @@
...
+      return errorMessage.includes(rule.filePath);
+    }
+
+    // Check if the message contains a path to a test file
+    // Test files are identified by:
+    // - Ending in .t.sol (e.g., Counter.t.sol)
+    // - Being inside test/contracts/ directory (e.g., test/contracts/Example.sol)
+    return (
+      /\.t\.sol(:|$|\s)/.test(errorMessage) ||
+      /(^|[/\\])test[/\\]contracts[/\\].*\.sol/.test(errorMessage)
```
kanej: I don't think we can get away with this assumption, I think we need to check if the file is in the Solidity test file folder defined in config.

kanej (changes requested):

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
```diff
@@ -0,0 +1,70 @@
...
+//    - Use this to suppress known warnings from internal/library files (e.g., console.sol)
+//    - The same warning type will still be shown for user code
+//
+// 2. scope: 'test-files' - Suppress warnings from all test files
+//    - Test files are identified as:
+//      * Files ending in .t.sol (e.g., Counter.t.sol)
+//      * Files inside test/contracts/ directory
+//        (e.g., test/contracts/Example.sol)
+//    - Use this for warnings that are acceptable in test code but not in production code
+//      (e.g., missing SPDX license identifiers or pragma statements)
```
kanej, line 16: I like this `scope` approach

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/integration/warning-filtering.ts
```diff
@@ -56,49 +93,62 @@ const solidity0833Config: HardhatUserConfig = {
...
-    // Check that the natspec warning WAS printed to console.warn
-    const warningWasShown = consoleWarnMock.mock.calls.some((call) => {
-      const message = String(call.arguments[0] ?? "");
-      return message.includes(NATSPEC_MEMORY_SAFE_WARNING);
+describe("build system - warning filtering", function () {
+  for (const scenario of testScenarios) {
+    it(scenario.name, async (t) => {
+      const consoleWarnMock = t.mock.method(console, "warn", noop);
+
+      await using project = await useTestProjectTemplate(scenario.project);
```
kanej, line 128: Did everyone know about this function except me?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
ChristopherDedominici: Good point! I forgot that the test path could be configured. I changed the logic in this commit: https://github.com/NomicFoundation/hardhat/pull/7904/commits/71d8aed0bbe3292f7af62c9b5f65f414e3f19be0

schaable (approved):

kanej (approved):

PR #7905, opened 2026-01-28
fix(hardhat-ledger): Handle device not ready errors by waiting and retrying

### LINKED ISSUE #4060: [hardhat-ledger] Wait for the ledger to be connected instead of throwing an error
If you send a tx from an address configured with `hardhat-ledger` but the wallet is not connected (or is locked, or the ethereum app is not selected), you'll get an error. Ideally we would do some retry with an exponential backoff, or something like that.

marianfe:
Fixes #4060

## Summary

When a Ledger device is not ready during a signing operation (locked at PIN screen, or Ethereum app not open), the plugin would immediately throw an error. This impacted long running scripts with multiple transactions.

This PR adds automatic retry logic that detects device-not-ready errors, displays an appropriate message, and retries in 30 seconds (up to a max of 30 minutes before giving up).

Handled errors: LockedDeviceError, status code 0x6511 (app not open). Also added a retry during initialization for users to plug in their ledger. This is to ensure a consistent UX with the retry cases during signing.

Added unit tests covering `init()` (connection), `#withConfirmation` (signing), and `#derivePath` (address derivation) code paths.

Also manually tested with the device.

### DIFF
--- a/v-next/hardhat-ledger/src/internal/handler.ts
+++ b/v-next/hardhat-ledger/src/internal/handler.ts
@@ -10,6 +10,8 @@ import type {
 import {
   DisconnectedDevice,
   DisconnectedDeviceDuringOperation,
+  LockedDeviceError,
+  TransportStatusError,
 } from "@ledgerhq/errors";
 import { isEIP712Message } from "@ledgerhq/evm-tools/lib/index";
 import Eth, { ledgerService } from "@ledgerhq/hw-app-eth";
@@ -27,6 +29,7 @@ import {
   hexStringToNumber,
   normalizeHexString,
 } from "@nomicfoundation/hardhat-utils/hex";
+import { sleep } from "@nomicfoundation/hardhat-utils/lang";
 import {
   rpcAddress,
   rpcAny,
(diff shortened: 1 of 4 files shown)


v-next/hardhat-ledger/src/internal/handler.ts
```diff
@@ -333,21 +367,37 @@ export class LedgerHandler {
...
         );
+
+        return this.#derivePath(addressToFindAsBuffer, retryAttempts + 1);
       }
 
+      // Give up - either exhausted retries or other error
       await this.#displayMessage("Derivation failure");
 
+      if (this.#isDeviceNotReadyError(error)) {
+        throw new HardhatError(
```
alcuadrado, line 402: This was already here, but in general you should provide the original error as last param, so that they get chained through [`Error#cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause)

v-next/hardhat-ledger/src/internal/handler.ts
```diff
@@ -45,18 +48,26 @@ import { getYParity } from "./get-y-parity.js";
...
+  public static readonly DEVICE_NOT_READY_RETRY_DELAY_SECONDS = 30;
+  public static readonly MAX_DEVICE_NOT_READY_RETRIES = 60;
 
   readonly #provider: EthereumProvider;
   readonly #displayMessage: (message: string) => Promise<void>;
   readonly #ethConstructor: typeof Eth.default;
   readonly #transportNodeHid: typeof TransportNodeHid.default;
   readonly #cachePath: string | undefined;
+  readonly #delayBeforeRetry: (seconds: number) => Promise<void>;
+  readonly #maxDeviceNotReadyRetries: number;
```
alcuadrado, line 76: The only part I found a bit confusing of this PR is that the connection and non-ready retries have different limits but the same counter. I think there's a small opportunity to simplify it, but I don't have feel strongly about it either.

alcuadrado (commented):
I left two comments. The one about retries is less important. The one about chaining errors is a bit annoying, but it's the kind of thing that has bit us in the past, making things way harder to debug than they should be.

v-next/hardhat-ledger/src/internal/handler.ts
marianfe: one way i'm thinking about this is that the method has a retryAttempts counter which is the number of retries of that particular method. then the limits can be different per exception. but I agree there's room to disambiguate , although not by simplifying but by having a counter object which should have separate counters per exception type. this also helps fix another corner case we found in manual testing (app first is not opened, then it disconnects the device after opening - which requires separate tracking per each retry). 

added a fix.

v-next/hardhat-ledger/src/internal/handler.ts
marianfe: agree, will fix this

alcuadrado (approved):
I like the `RetryState` introduction. It's simpler to understand now.

PR #7906, opened 2026-01-28
Add boilerplate of hardhat-exposed

alcuadrado:
Please do not merge

### DIFF
--- a/v-next/hardhat-exposed-example/src/internal/config.ts
+++ b/v-next/hardhat-exposed-example/src/internal/config.ts
@@ -1,0 +1,71 @@
+import type { HardhatConfig, HardhatUserConfig } from "hardhat/types/config";
+import type { HardhatUserConfigValidationError } from "hardhat/types/hooks";
+
+import path from "node:path";
+
+/**
+ * Default directory name for generated exposed contract wrappers.
+ * This path is relative to the project root.
+ */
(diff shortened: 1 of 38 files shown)


v-next/hardhat-exposed-example/src/internal/config.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @param userConfig - The user's Hardhat configuration object.
+ * @returns An array of validation errors. Empty if config is valid.
+ */
+export async function validatePluginConfig(
+  userConfig: HardhatUserConfig,
+): Promise<HardhatUserConfigValidationError[]> {
+  if (
+    userConfig.paths !== undefined &&
+    userConfig.paths.exposedContracts !== undefined &&
+    typeof userConfig.paths.exposedContracts !== "string"
```
Amxx, line 26: wouldn't that be enough ?
```suggestion
    typeof userConfig.paths?.exposedContracts !== "string"
```

alcuadrado: I think that change would make the `exposedContracts` required.

Maybe a simpler version could be:

```
const configType = typeof userConfig.paths?.exposedContracts;
if (configType !== "string" && configType !== "undefined) {
```

alcuadrado: We actually use zod in our plugins though, I kept it with minimal dependencies here because it's an example

PR #7907, opened 2026-01-28
feat(hardhat-viem): support custom chains via chainDescriptors and network config fallback

### LINKED ISSUE #7763: HardhatError: HHE40000: No network with chain id "9876" found.
### Version of Hardhat

3.0.17

### What happened?

When using ignitions, there is no way to use chains not supported by viem.

Even if we override the network passed in to network.connect() I think ignition still has some dependency to viem and tries to look for chainid in viem network lists. 

It can work by monkey patching the viem but since it is pulled in compiled, patching typescript and adding the new chain is not enough and resulting js files also needs to be monkey patched which defeats the purpose. 

Also just passing the custom object in a network.connect() breaks the hardhat usual --network <definedNetworkName> flow and makes scripts / ignitions only work on set chain. 

```

const gccnet = {
    id: 9876,
    name: "GCCNET",
    network: "gccnet",
    nativeCurrency: {
        name: "GCC",
        symbol: "GCC",
        decimals: 18,
    },
    rpcUrls: {
        default: {
            http: ["https://gccnet.com"],
        },
    },
    blockExplorers: {
        default: {
            name: "explorer",
            url: "https://gccscanner.com",
        },
    },
};

const gccnetChain = defineChain(gccnet)
const { viem, networkName, ignition } = await network.connect(gccnetChain);
const wallets = await viem.getWalletClients(
    {
        chain: gccnetChain,
    }
);


const wallet = wallets[0];

async function main() {
    console.log(`Deploying to ${networkName}... on wallet ${wallet.account!.address}`);

    const { airdropper } = await ignition.deploy(AirdropperModule);
}

```

### Search terms

_No response_

daanporon:
> This description was updated to reflect the current implementation approach. The original version is available in the edit history.

## Summary
When using `hardhat-viem` with a custom chain not in viem's built-in chain list, workflows fail with:

  `HHE40000: No network with chain id "<id>" found`

`hardhat-viem` resolves the viem `Chain` object from `viem/chains`. If the chain ID isn't found there, it throws. Development networks (`hardhat` and `anvil`) are handled separately and are not affected.

Direct usage works because users can explicitly pass a custom chain:

```typescript
const customChain = defineChain({ ... });
const wallets = await viem.getWalletClients({ chain: customChain });
```

The issue appears when `hardhat-viem` is used indirectly by other plugins, such as Ignition, since they don’t have a way to inject a custom chain.

## Implementation
This PR introduces a fallback chain resolution strategy. When a chain ID is not found in `viem/chains`, `hardhat-viem` now falls back instead of throwing:

1. `viem/chains` lookup - existing behavior
2. `chainDescriptors` fallback - builds a viem `Chain` from a matching descriptor (using its name and block explorer, if available)
3. Network config fallback - builds a minimal viem `Chain` from the Hardhat network name

The minimal fallback chain uses `{ name: "Ether", symbol: "ETH", decimals: 18 }` as the default `nativeCurrency` and `{ default: { http: [] } }` for `rpcUrls`, since `hardhat-viem` always overrides the transport.

`chainDescriptors` and `networkName` are passed explicitly from the hook handler through `initializeViem` to the client and contract factories, and finally to `getChain`.

## Usage
No changes are required for most users. Unknown chain IDs are now handled automatically. Advanced cases will be covered by a follow-up PR (see issue [#8024](https://github.com/NomicFoundation/hardhat/issues/8024)).

Closes https://github.com/NomicFoundation/hardhat/issues/7763


schaable:
Hi @daanporon, I’m putting this PR on hold since we created a design doc to confirm this is the direction we want to take. The design doc is currently under review. I’ll continue reviewing this PR once it’s approved.

phpmac:
It hasn't been merged yet. I'm requesting a merge, guys.

v-next/hardhat-viem/test/chains.ts
```diff
@@ -12,34 +18,90 @@ import {
...
+            },
+          }
+        : {}),
+    },
+  };
+}
+
 describe("chains", () => {
   describe("getChain", () => {
-    it("should return the chain corresponding to the chain id", async () => {
```
schaable, line 21: This file has many changes, but the only relevant ones are:

- `should return a custom chain from chainDescriptors when the chain id is not in viem's chain list` it block
- `should fall back to a minimal chain with the network name when the chain id is not in viem's chain list and no descriptor matches` it block
- `createMinimalChain`, `getDefaultBlockExplorer`, and `resolveChain` describe blocks

All other changes are related to wording, test reordering, and adding the new parameters to `getChain`.

schaable (approved):

v-next/hardhat-viem/src/internal/initialization.ts
```diff
@@ -19,24 +20,53 @@ export function initializeViem<ChainTypeT extends ChainType | string>(
   chainType: ChainTypeT,
   provider: EthereumProvider,
   artifactManager: ArtifactManager,
+  chainDescriptors: ChainDescriptorsConfig,
```
schaable, line 23: I'm really tempted to switch to object params in `initializeViem`, the internal helpers, and `getChain`, since the number of parameters has grown too much for positional args. But we can do that in a follow-up PR to avoid adding noise to this one.

schaable:
@kanej will do a final review tomorrow, and then we should be good to go. Thanks for the patience, @daanporon!

.peer-bumps.json
```diff
@@ -26,6 +26,21 @@
...
+    {
+      "package": "@nomicfoundation/hardhat-viem-assertions",
+      "peer": "@nomicfoundation/hardhat-viem",
+      "reason": "Requires the fallback chain resolution for unknown chain ids"
+    },
+    {
+      "package": "@nomicfoundation/hardhat-toolbox-viem",
+      "peer": "@nomicfoundation/hardhat-viem",
+      "reason": "Requires the fallback chain resolution for unknown chain ids"
     }
```
kanej, line 54: I think this implies they should be included in the changeset.

kanej: And can I check, there are no API level changes but the defaulting behaviour has changed and is used from Ignition.

v-next/hardhat-viem/src/internal/chains.ts
```diff
@@ -214,3 +213,65 @@ function isHardhatMetadata(value: unknown): value is HardhatMetadata {
...
+
+/**
+ * Creates a viem chain with the minimum fields required to satisfy the
+ * {@link ViemChain} type.
+ */
+export function createMinimalChain(id: number, name: string): ViemChain {
+  return {
+    id,
+    name,
+    nativeCurrency: { name: "Ether", symbol: "ETH", decimals: 18 },
```
kanej, line 252: Is there any sense in which we can get `nativeCurrency` or `rpcUrls` from the chain descriptors?

.peer-bumps.json
schaable: Correct. If we don’t include Ignition in the peer deps, users would need to manually update `hardhat-viem` for this to work with Ignition.

v-next/hardhat-viem/src/internal/chains.ts
schaable: `rpcUrls` aren't used at all since we rely on a custom transport, so including them doesn't make much sense.

As for `nativeCurrency`, we discussed it with Pato and agreed it falls under advanced use cases, where the user would provide a `viemChain` in the `chainDescriptor` (to be added in a future PR).

.peer-bumps.json
schaable: > I think this implies they should be included in the changeset.

Done in https://github.com/NomicFoundation/hardhat/pull/7907/changes/5e5824c03244be3e15d738f61a97eeac14b6c5b2

kanej (approved):
Looks good from a code perspective; I have double checked the changeset and peerbump as well.

PR #7909, opened 2026-01-29
[WIP] Use native official solc builds for `production` in Linux ARM64

alcuadrado:


v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -24,6 +25,16 @@ import { z } from "zod";
...
+// The first solc version with official ARM64 Linux builds
+export const FIRST_OFFICIAL_ARM64_SOLC_VERSION = "0.8.31";
+
+/**
+ * Determines if a solc version has an official ARM64 Linux build.
+ */
+export function hasOfficialArm64Build(version: string): boolean {
+  return semver.gte(version, FIRST_OFFICIAL_ARM64_SOLC_VERSION);
+}
+
```
alcuadrado, line 37: I think this would be better placed in `v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-info.ts`.

Their tests should also be moved.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -448,7 +448,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
 
       if (solcLongVersion === undefined) {
         const compiler = await getCompiler(solcConfig.version, {
-          preferWasm: buildProfile.preferWasm,
+          preferWasm: solcConfig.preferWasm ?? buildProfile.preferWasm,
```
alcuadrado, line 451: This logic is repeated below, so it should use a helper.

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
alcuadrado, line 400: The name and comment of this is now outdated. It should be something like `missesSomeOfficialNativeBuilds()`, and its documentation should be updated.

It should also be moved to `v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-info.ts`

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -450,6 +452,65 @@ describe("solidity plugin config validation", () => {
...
+        }),
+        [],
+      );
+    });
+
+    it("Should reject invalid preferWasm values in SolcUserConfig", () => {
+      assert.deepEqual(
+        validateSolidityUserConfig({
+          solidity: {
+            compilers: [{ version: "0.8.28", preferWasm: "true" as any }],
```
alcuadrado, line 475: use `@ts-expect-error` instead of `as any`

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -450,6 +452,65 @@ describe("solidity plugin config validation", () => {
...
+        }),
+        [],
+      );
+    });
+
+    it("Should reject invalid path values in SolcUserConfig", () => {
+      assert.deepEqual(
+        validateSolidityUserConfig({
+          solidity: {
+            compilers: [{ version: "0.8.28", path: 123 as any }],
```
alcuadrado, line 502: use @ts-expect-error instead of as any

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -450,6 +452,65 @@ describe("solidity plugin config validation", () => {
...
+          },
+        }),
+        [
+          {
+            message: "Expected string, received number",
+            path: ["solidity", "compilers", 0, "path"],
+          },
+        ],
+      );
+    });
```
alcuadrado, line 512: These two tests were not part of the plan, but they are useful. Let's keep them, but move them to their own `describe` block at the same level as the current one.

alcuadrado:
The linter isn't passing because `pnpm lint:fix` wasn't run.

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -568,4 +606,140 @@ describe("solidity plugin config resolution", () => {
...
+      assert.equal(overrides["contracts/Special.sol"].preferWasm, true);
+    });
+  });
+
+  describe("ARM64 Linux per-compiler preferWasm defaults", {
+    skip: !(os.platform() === "linux" && os.arch() === "arm64"),
+  }, () => {
+    const otherResolvedConfig = { paths: { root: process.cwd() } } as any;
+
+    it("should default preferWasm to true for versions without official ARM64 builds", async () => {
```
alcuadrado, line 654: Given that this is the default build profile, it should default to false.

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -337,10 +350,20 @@ function resolveSolcConfig(
     };
   }
 
+  // Resolve per-compiler preferWasm:
```
alcuadrado, line 353: This logic is incorrect:

Inside the `if`, it should look something like this:

```
resolvedPreferWasm = production && !hasOfficialArm64Build(solcConfig.version);
```

alcuadrado: This case is not tested either.

alcuadrado:
The reason the tests of ignition are failing is the error I commented here: https://github.com/NomicFoundation/hardhat/pull/7909/changes#r2742920825

alcuadrado:
Closing it for future reference

PR #7911, opened 2026-01-29
Add bump-peers CLI tool and GH Actions workflow for managing peer dependency bumps automatically

alcuadrado:
This PR adds a script that addresses the problem where changesets automatically bumps internal peer dependencies too aggressively. It's built to be run after `pnpm changeset version --no-commit`.

The idea of the tool is that we keep a list of peer dependency bumps in a json file `.peer-bumps.json`, that looks something like this

```json
{
  "$schema": "./scripts/peer-bumps.schema.json",
  "excludedFolders": [
    "folders-with-packages-to-ignore",
  ],
  "bumps": [
    {
      "package": "package that needs a bump",
      "peer": "peer dependency to bump",
      "reason": "reason message"
      // no "version", so it bumps it to the new version being released. Optionally, you can add a version and it would set `workspace:^version`
    }
  ]
}
```

If we use this bumps file, there's no manual process to be done on any v3 release PR. We can replicate it for v2.


alcuadrado:
I also added a new workflow with the same patterns as the `no-changeset-needed`, so that we pay attention to the peer bumps file in the PRs.

Also deleted the one that was forcing a manual review

scripts/validate-pull-request-peer-bumps.mjs
```diff
@@ -0,0 +1,70 @@
...
+}
+
+async function hasPeerBumpEntry() {
+  if (process.env.GITHUB_BASE_REF === undefined) {
+    throw new Error("GITHUB_BASE_REF is not defined");
+  }
+
+  const { stdout } = await exec(
+    `git diff --name-only ${process.env.GITHUB_BASE_REF} -- ${peerBumpsFile}`
+  );
```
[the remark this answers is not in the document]
kanej: Claude is suggesting switching to execFile on `git`, avoid the shell and limit the scope for injection.

scripts/bump-peers.mjs
```diff
@@ -0,0 +1,793 @@
...
+// Type Definitions
+// ============================================================================
+
+/**
+ * @typedef {object} PeerBump
+ * @property {string} package
+ * @property {string} peer
+ * @property {string} reason
+ * @property {string} [version]
+ */
```
kanej, line 18: Not for this PR - but I think we swap all these `.mjs` scripts to `.ts` and use Node 24 with type stripping.

scripts/bump-peers.mjs
```diff
@@ -0,0 +1,793 @@
...
+
+/**
+ * @param {string} args
+ * @returns {string}
+ */
+function pnpm(args) {
+  return exec(`pnpm ${args}`);
+}
+
+// ============================================================================
```
kanej: I know it is crazy, but it annoys me that it is 79 characters rather than 80, so it is one character short of my 80 character gutter.

scripts/bump-peers.mjs
alcuadrado: Agreed. I was tempted to do it, but didn't want to kill the reviewers. I think claude will one-shot it.

scripts/bump-peers.mjs
alcuadrado: Consider it done

scripts/bump-peers.mjs
```diff
@@ -0,0 +1,793 @@
...
+// Entry Point
+// ============================================================================
+
+/**
+ * @returns {void}
+ */
+function main() {
```
kanej: For these sorts of scripts I would reverse the order of function declarations so that they can be read top to bottom.

scripts/bump-peers.mjs
```diff
@@ -0,0 +1,793 @@
...
+}
+
+// Styling helpers for consistent formatting
+const fmt = {
+  /** @param {string} name */
+  pkg: (name) => styleText("bold", name),
+  /** @param {string} v */
+  version: (v) => styleText("green", v),
+  /** @param {string} text */
+  dim: (text) => styleText("dim", text),
```
kanej: Seeing it here it makes sense, but in context it threw me.
Maybe `deemphasize` or `dimText`?

scripts/validate-pull-request-peer-bumps.mjs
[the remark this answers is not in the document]
alcuadrado: Fixed it by first using `which` and then execFileSync

scripts/bump-peers.mjs
alcuadrado: Are you saying that we should use a reverse topological order?

scripts/bump-peers.mjs
```diff
@@ -0,0 +1,793 @@
...
+/**
+ * @param {string} packagePath
+ * @param {PackageJson} json
+ * @returns {void}
+ */
+function writePackageJson(packagePath, json) {
+  const filePath = resolve(packagePath, "package.json");
+  const content = JSON.stringify(json, null, 2) + "\n";
+  writeFileSync(filePath, content);
+}
```
kanej: I was worried about this - that we will end up with reordered properties and other unintended changes (and violations of formatting).
But a run of the script makes changes and linting passes.
My mental model is that JSON.parse and JSON.stringify are maintaining the existing order and the script only updates within peerdeps and devdeps.
Put another I am happy with this, and we can do something more complicated if we start seeing issues in the release PR.

kanej (approved):
I only tested on the current branch with the `hardhat-verify` suggestion. We can iterate on this when edge cases come up.

In general I like a self contained script with next to no deps approach.

scripts/bump-peers.mjs
kanej: Yes, yes I am ... you are having too much fun now.

alcuadrado: I already applied the change, but won't push it yet, because it will ruin your review. Let me know when I should.

scripts/bump-peers.mjs
alcuadrado: I was also worried about it tbh. I validated manually a few times. Agreed that we should fix it if/when it becomes a problem.

scripts/bump-peers.mjs
alcuadrado: I saw the approval, so I'll push it

PR #7912, opened 2026-01-30
[hardhat-typechain] Automatically delete the generated types when running hardhat clean

alcuadrado:


.changeset/fluffy-lions-deny.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-typechain": patch
+---
+
+Automatically delete the generated types when running hardhat clean
```
ChristopherDedominici, line 5: Minor: we should link the issue being fixed, or the PR if there is no issue.

Example: https://github.com/NomicFoundation/hardhat/blob/5b10df676be98be40cd2919c001f2776e7569dc6/.changeset/plenty-bikes-sniff.md?plain=1

ChristopherDedominici (approved):

.changeset/fluffy-lions-deny.md
alcuadrado: That's a good point. Unfortunately there isn't a single issue for this. I closed the ones that mentioned it, amongst other things.

PR #7914, opened 2026-01-30
Support eth_getProof

### LINKED ISSUE #3345: Support `eth_getProof`
Hi, I would like eth_getProof to be part of the supported JSON-RPC methods (https://hardhat.org/hardhat-network/docs/reference#json-rpc-methods-support). I am doing something where I am using merkle proofs, which can be found via that endpoint. FWIW, ganache supports this method (https://github.com/trufflesuite/ganache/issues/382).

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/3345

### Error thrown
This expected error is thrown when trying to use `eth_getProof` in forked networks that have local changes:
image

### Hardhat website update
I did not update the Hardhat website, as currently we do not have a section listing all the supported and unsupported RPC methods.

I create this issue to add docs related to supported/not supported RPC methods: https://github.com/NomicFoundation/hardhat-website/issues/207


v-next/hardhat/.gitignore
```diff
@@ -7,6 +7,9 @@
 # pnpm deploy output
 /bundle
 
+# all cache folders
```
ChristopherDedominici: When running the forking tests, files were stored in a `cache` folder at the root of the package, so I added this. I am not sure whether it is possible to configure a different cache path for forked networks

v-next/hardhat/.gitignore
```diff
@@ -7,6 +7,9 @@
 # pnpm deploy output
 /bundle
 
+# cache directory at package root
+/cache
```
ChristopherDedominici: When running the forking tests, files were stored in a cache folder at the root of the package, so I added this. I am not sure whether it is possible to configure a different cache path for forked networks

v-next/hardhat/test/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -205,6 +206,194 @@ describe("edr-provider", () => {
...
+        }
+      });
+
+      it("should throw error on fork with local changes", async () => {
+        // NOTE: Accounts are NOT disabled in the network configuration, so Hardhat modifies the state by
+        // injecting default accounts, causing eth_getProof to fail with a
+        // "proof not supported in fork mode" error.
+
+        const forkedHre = await createHardhatRuntimeEnvironment({
+          networks: {
```
schaable, line 361: Are the forking tests really needed? Forking the network takes time (about 3-5 seconds in my tests), and it could introduce flakiness in CI.

v-next/hardhat/.gitignore
schaable: You could provide the cache path in the `createHardhatRuntimeEnvironment` call and set it under tests (and ignore it here):

```typescript
  before(async function () {
    hre = await createHardhatRuntimeEnvironment({
      paths: {
        cache: "some/cache/path",
      }
    });
  });
```

EDIT: If we remove the forking tests due to my other comment, the .gitignore change can be reverted.

schaable (approved):
Looks good. I left two comments but I’m pre-approving.

v-next/hardhat/test/internal/builtin-plugins/network-manager/edr/edr-provider.ts
ChristopherDedominici: Yes, they are a specific request from the issue, see here: https://github.com/NomicFoundation/hardhat/issues/3345#issuecomment-3822877280

kanej: I think we might look at some funky http level mocking for the entire test suite in the near future, but I agree that this test is the explicit case we were asked to cover.

kanej:
Thanks for the screenshot of the unsupported case Provider error. I think it makes sense from a user experience perspective.

PR #7915, opened 2026-01-30
Add `SolidityHooks#readNpmPackageRemappings`

alcuadrado:
Docs issue: https://github.com/nomicfoundation/hardhat-website/issues/202


alcuadrado:
Its peer bump is in https://github.com/NomicFoundation/hardhat/pull/7916

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
```diff
@@ -397,41 +421,76 @@ export class RemappedNpmPackagesGraphImplementation
...
-      npmPackage.rootFsPath,
-      (f) => path.basename(f) === "remappings.txt",
-      (f) => !f.endsWith("node_modules"),
-    );
+    const defaultBehavior = async (
+      packageName: string,
+      packageVersion: string,
+      packagePath: string,
+    ) => {
+      // Default: read from remappings.txt files (existing implementation)
```
marianfe: minor: this part about "(existing implementation)" looks like a llm comment

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
```diff
@@ -397,41 +421,76 @@ export class RemappedNpmPackagesGraphImplementation
       UserRemappingError[]
     >
   > {
-    const remappingsTxtFiles = await getAllFilesMatching(
-      npmPackage.rootFsPath,
-      (f) => path.basename(f) === "remappings.txt",
-      (f) => !f.endsWith("node_modules"),
-    );
+    const defaultBehavior = async (
```
marianfe: minor: consider making this an async private method of dependency-resolver.ts, rather than a local delegate inside #readPackageRemappings, it reads weird that the entire standalone implementation of reading remappings.txt is a delegate function rather than its own private method...

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
```diff
@@ -397,41 +421,76 @@ export class RemappedNpmPackagesGraphImplementation
...
-        .filter((line) => line !== "")
-        .filter((line) => !line.startsWith("#"));
+    if (this.#remappingsReader !== undefined) {
+      allRemappings = await this.#remappingsReader(
+        npmPackage.name,
+        npmPackage.version,
+        npmPackage.rootFsPath,
+        defaultBehavior,
+      );
+    } else {
```
marianfe: do we intentionally not want to merge the default (remappings.txt) with whatver the remappingsReader provides (so it's either or, never a union of the 2)? 

update: i see, we basically pass default behavior to the remappingsReader, and it can decide to either add them to the results or to discard them or etc. 

to consider: I wonder if there is another pattern we can do here, as the remappingsReader might incorrectly use the defaultBehavior or incorrectly merge the remappings from the default? does the remappingsReader need to actually get the defaultBehavior "reader", or could it just return an object with its own results and an indication of what it wants to do with the default ones (merge/union, discard, ...)?

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
```diff
@@ -397,41 +421,76 @@ export class RemappedNpmPackagesGraphImplementation
...
+      allRemappings = await defaultBehavior(
+        npmPackage.name,
+        npmPackage.version,
+        npmPackage.rootFsPath,
+      );
+    }
 
-      for (const userRemapping of rawUserRemappings) {
+    // Parse and deduplicate remappings
```
marianfe: minor: consider moving deduplication (everything from here down) to its own private method, makes it easier to read the high level intent of this method (and also to unit test deduplication in isolation?)

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
```diff
@@ -1644,6 +1645,398 @@ contr:to-npm=node_modules/dep/contracts`,
...
+        version: "1.0.0",
+        files: {
+          "contracts/A.sol": `import "foo/Bar.sol";`,
+          "remappings.txt": "foo/=src/foo/",
+        },
+      };
+
+      await using project = await useTestProjectTemplate(template);
+
+      // Track hook invocations
```
marianfe, line 1662: minor: this hook mock pattern could maybe be shared by more tests , as it's usually the same.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
```diff
@@ -1644,6 +1645,398 @@ contr:to-npm=node_modules/dep/contracts`,
...
+      await graph.selectBestUserRemapping(dep1FooSol, "alias1/Something.sol");
+
+      // Verify hook was called for dep1 (because it has remappings.txt)
+      const dep1Call = hookCalls.find((call) => call.name === "dep1");
+      assert.ok(dep1Call !== undefined, "Hook should be called for dep1");
+      assert.equal(dep1Call.version, "2.0.0");
+      assert.equal(dep1Call.path, path.join(project.path, "node_modules/dep1"));
+    });
+
+    it("should deduplicate remappings by context+prefix with first occurrence winning", async () => {
```
marianfe, line 1783: intuitively feels to me like more edge cases should be tested for deduplication, not sure if this test is enough? 

also, do we have some actual remappings to pass to a test as a hardcoded json to see that it works well for real remapping path structures (rather than simplified test ones like foo/=src/first/) - just to have a more realistic dataset. maybe for this and other tests.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
```diff
@@ -1644,6 +1645,398 @@ contr:to-npm=node_modules/dep/contracts`,
...
+      };
+
+      await using project = await useTestProjectTemplate(template);
+
+      const mockReader: RemappingsReaderFunction = async (
+        name,
+        version,
+        packagePath,
+        defaultBehavior,
+      ) => {
```
marianfe, line 1903: this is where my concern on the other comment came from: we can add this to unit test, but we cant really enforce the caller's implementation of remappingsReader to correctly match this pattern, they might have a buggy implementation and it's outside our control. 

the test only really tests "allows hook to add additional remappings" because of this test hook we added here to the test locally, that really merges the results lists (so in reality, this test actually tests "hook returns things when it's called")

on the other hand, maybe we are the only consumers with hardhat-foundry so that's fine

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
```diff
@@ -397,41 +421,76 @@ export class RemappedNpmPackagesGraphImplementation
...
+    // Call the wrapper function if provided, otherwise use default
+    let allRemappings: Array<{ remappings: string[]; source: string }>;
 
-      const rawUserRemappings = packageRemappingsTxtContents
-        .split("\n")
-        .map((line) => line.trim())
-        .filter((line) => line !== "")
-        .filter((line) => !line.startsWith("#"));
+    if (this.#remappingsReader !== undefined) {
+      allRemappings = await this.#remappingsReader(
```
marianfe: do we want exception handling for the custom hook? what if it throws etc.?

marianfe (approved):
LGTM with some comments, please take a look, thanks!

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
alcuadrado: There's nothing special about a hook throwing. If the code executing the hook wants to handle the error, it can, but in most cases, like this one, we'd want the process to die.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
alcuadrado: Oops, it is

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
alcuadrado: Good point

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
alcuadrado: This is a patter we used sometimes to avoid having to pass the HookHandler everywhere, otherwise things get too hard to test.

Maybe we can make the `RemappingsReaderFunction` required, and add to its docs that it's expected to be the hook handler run.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
alcuadrado: Good point. Added `#parseAndDeduplicateRemappings` and it looks much better.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
alcuadrado: Not sure if I understand the comment. Do you mean in this file, or in general in the codebase?

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
alcuadrado: Actually, the only consumer is `hardhat` itself, inside the solidity build system. The way plugins can use it is by registring hook handlers, but they never provide a `RemappingsReaderFunction`. That's provided here:

https://github.com/NomicFoundation/hardhat/blob/0fa0f47d15a76dd99db9e60dd43e60ea5ff37da2/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts#L98-L101

which is called by:

https://github.com/NomicFoundation/hardhat/blob/0fa0f47d15a76dd99db9e60dd43e60ea5ff37da2/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/dependency-graph-building.ts#L24-L43

That way we avoid passing the `HookHandler` around, which makes unit testing things really hard.

alcuadrado: I didn't answer one of your concerns: yes, a custom hook handler can decide not to call `next()`, which means that the default behavior won't be run. That's intentional and documented. If a plugin breaks Hardhat people won't use it.

alcuadrado: I also added a comment explaining that in this one test, it's simulating a hook handler that doesn't call next.

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
alcuadrado: > intuitively feels to me like more edge cases should be tested for deduplication, not sure if this test is enough?

There's a more realistic test case in the `hardhat-foundry` test suite.

alcuadrado: > also, do we have some actual remappings to pass to a test as a hardcoded json to see that it works well for real remapping path structures (rather than simplified test ones like foo/=src/first/) - just to have a more realistic dataset. maybe for this and other tests.

I didn't understand this comments. In particular, the json part.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
alcuadrado: I made it receive a default implementation that just calls the default behavior instead, and I believe it's now easier to understand. See: 856802d44 and this other comments https://github.com/NomicFoundation/hardhat/pull/7915#discussion_r2843305668

alcuadrado:
Merging manually here because we are skipping a version of `hardhat-foundry` and that makes the merge queue fail.

PR #7916, opened 2026-01-30
Introduce `@nomicfoundation/hardhat-foundry` plugin

### LINKED ISSUE #5645: Port `hardhat-foundry` to v3
Investigate how we port the `hardhat-foundry` plugin to v3:

1. A standalone plugin providing the same features
2. Updates to Hardhat 3 configuration setup and rules that provide the equivalent support.

This task is likely dependenc on the solc remappings support of the build system.

### LINKED ISSUE #7561: Better compatibility with remappings coming from `foundry.toml`
## Context

While Hardhat has native support for remappings.txt, multiple projects define them in their foundry.toml, and forge autogenerate some of them in the presence of a `foundry.toml`.

In those cases hardhat can fail to compile a project, due to missing remappings.

## Proposal

We need to create a really simple `hardhat-foundry` plugins that loads those remappings to.

### Expected behavior

Whenever hardhat loads the remappings for a package (either the hardhat project itself, or an npm module), it the plugin should use `forge remappings` to load those remappings too.

## Design

We should add a hook to the package loading process, which happens here: https://github.com/NomicFoundation/hardhat/blob/f29d1d804e00e2b13666f040fb85d7b7b71a58ce/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts#L392

Just like we find all the `remappings.txt` files [here](https://github.com/NomicFoundation/hardhat/blob/f29d1d804e00e2b13666f040fb85d7b7b71a58ce/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts#L400-L404), we should look for the `foundry.toml` files (that are not a `node_modules`), and run `foge remappings` in the context of that file's directory.

## Out of scope

- This plugin won't install forge, but fail with a clear message if it isn't installed.
- This plugin won't change any default config of hardhat, like paths.

## Open questions

- Can our [`Remapping`-related types](https://github.com/NomicFoundation/hardhat/blob/f29d1d804e00e2b13666f040fb85d7b7b71a58ce/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/types.ts#L24-L80) accommodate for this? For example, `source` is expected to be a `remapping.txt` absolute path. Can we reuse it? Will the logic and error reporting break?

alcuadrado:
https://github.com/NomicFoundation/hardhat-website/issues/206


v-next/hardhat-foundry/src/internal/foundry/forge.ts
```diff
@@ -0,0 +1,118 @@
...
+import { promisify } from "node:util";
+
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
+import { ensureError } from "@nomicfoundation/hardhat-utils/error";
+import { exists } from "@nomicfoundation/hardhat-utils/fs";
+
+const execAsync = promisify(exec);
+
+const FORGE_TIMEOUT_MS = 500;
+const FORGE_VERSION_TIMEOUT_MS = 500; // 0.5 seconds
```
marianfe, line 12: should these 2 come from the plugin config?

v-next/hardhat-foundry/test/foundry/forge.ts
```diff
@@ -0,0 +1,212 @@
...
+          "Error should have a 'code' property (exit code or error code)",
+        );
+
+        // The error code should be 127 (command not found) or 'ENOENT' (no such file)
+        // On Windows, the error code is 1 when the command is not found
+        // This is what we mock in our tests as FORGE_NOT_INSTALLED scenario
+        const isCommandNotFound =
+          error.code === 127 ||
+          error.code === "ENOENT" ||
+          (process.platform === "win32" && error.code === 1);
```
marianfe, line 148: do we actually run this test on windows in the CI? or is this more like a guardrail in case someone might run our test cases on windows?

v-next/hardhat-foundry/README.md
```diff
@@ -0,0 +1,32 @@
...
+import hardhatFoundry from "@nomicfoundation/hardhat-foundry";
+
+export default defineConfig({
+  plugins: [hardhatFoundry],
+});
+```
+
+## Usage
+
+You don't need to do anything differently, just install the plugin and Hardhat will be able to resolve Solidity remappings for Foundry-based packages, including those installed with npm.
```
marianfe: text copy needs updating (sounds weird with 'differently'): 

You don't need to do anything additionally, just ...
or You only need to install the plugin ..
or You don't need to do anything other than installing the plugin ...

marianfe (approved):
LGTM with some comments, thanks!

v-next/hardhat-foundry/src/internal/foundry/forge.ts
alcuadrado: It feels like the kind of config that nobody would ever change, so I'd default to keep them as constants unless people actually need to tweak them.

v-next/hardhat-foundry/test/foundry/forge.ts
alcuadrado: We run them on windows, linux, and mac.

v-next/hardhat-foundry/README.md
alcuadrado: You are right. That's pretty spanish 😅

I implemented the second one in b07e9dfe1

PR #7918, opened 2026-01-30
fix: maintain original `MinimalEthereumJsVm` instance

Wodann:

---

Community plugins were using the original `MinimalEthereumJsVm` instance, so instead of replacing it, we now change its internal `stateManager` using an `updateProvider` function, which was the only thing that needs an updated provider.

This should fix the secondary issue reported in #7834.


schaable (approved):
LGTM

PR #7925, opened 2026-02-03
Expose etherscan instance

### LINKED ISSUE #7644: Expose `hardhat-verify` capabilities to other Plugins
To support OpenZeppelin's upgrade plugin we need to expose some of underlying logic of the `hardhat-verify` plugin.

We intentionally limited the exposed API of `hardhat-verify` with the intent that we would expand on request (rather than guessing the API we would based it on real usage).

## OpenZeppelin's Usage

The Upgradable contracts plugin in Hardhat 2 enhances the `hre` with a `upgrades` property that provides methods that operates on ethers contract instances to support deploying and maintaining upgradeable contracts.

After deploying an upgradeable contract or a new implementation during an upgrade, the contracts can be verified on `etherscan`.

Currently the OpenZeppelin upgrade plugin imports the following:

```ts
import { Etherscan } from '@nomicfoundation/hardhat-verify/etherscan';
import type { VerificationProvider } from '@nomicfoundation/hardhat-verify/internal/types';
import { verifyContract } from "@nomicfoundation/hardhat-verify/verify";
```

## Hardhat 3 Needed Changes

The `hardhat-verify` plugin has been restructured in Hardhat 3. However we need to support the equivalent of `verifyContract` to allow third-party plugins like OpenZeppelin's to work as before.

## Design Considerations

* Is it enough to expose the currently internal `verifyContract` function: https://github.com/NomicFoundation/hardhat/blob/cc7b3ecc521302a73ae932dd7a616a9c2c309a53/v-next/hardhat-verify/src/internal/verification.ts#L71
* What is the best API for exposing this functionality for use from other plugins?
* How will exposing `verifyContract` work with additional providers?

## Useful Resources

- [OpenZeppelin Hardhat Upgrades Plugin Repo](https://github.com/OpenZeppelin/openzeppelin-upgrades/tree/master/packages/plugin-hardhat)
- [OpenZeppelin Hardhat Upgrades Plugin Docs](https://docs.openzeppelin.com/upgrades-plugins)
- [OpenZeppelin verify code](https://github.com/OpenZeppelin/openzeppelin-upgrades/blob/master/packages/plugin-hardhat/src/verify-proxy.ts)

schaable:
Closes: https://github.com/NomicFoundation/hardhat/issues/7644

Documentation: (readme) was updated with the new API.

### Contract deployment and verification
```typescript
/**
 * NOTE: This script demonstrates manual contract verification using the
 * low-level verifier API. This is NOT the recommended way to verify contracts.
 *
 * For standard contract verification, use the `verifyContract` function instead:
 * https://github.com/NomicFoundation/hardhat/tree/main/v-next/hardhat-verify#programmatic-verification
 *
 * Only use this manual approach if you have a specific scenario that cannot be
 * covered by the standard `verifyContract` function.
 */

import type { BuildInfo } from "hardhat/types/artifacts";

import fsPromises from "node:fs/promises";
import path from "node:path";
import { setTimeout } from "node:timers/promises";

import { network, artifacts } from "hardhat";

const { viem, verifier } = await network.connect("sepolia");

const contract = await viem.deployContract("Counter", [], { confirmations: 5 });
const contractAddress = contract.address;

console.log("Contract deployed at:", contractAddress);

const artifact = await artifacts.readArtifact("Counter");

const buildInfoPath = path.join(
  process.cwd(),
  "artifacts",
  "build-info",
  `${artifact.buildInfoId}.json`,
);
const content = await fsPromises.readFile(buildInfoPath, { encoding: "utf8" });
const buildInfo: BuildInfo = JSON.parse(content);

const guid = await verifier.etherscan.verify({
  contractAddress: contractAddress,
  compilerInput: buildInfo.input,
  contractName: `${artifact.inputSourceName}:Counter`,
  compilerVersion: `v${buildInfo.solcLongVersion}`,
  constructorArguments: "",
});

// wait for 1 second before polling so that Etherscan has time to process
// the verification request
await setTimeout(1000);

const response = await verifier.etherscan.pollVerificationStatus(
  guid,
  contractAddress,
  `${artifact.inputSourceName}:Counter`,
);

console.log("Verification response:", response);
```

### Using `customApiCall` to fetch event logs
```typescript
import { keccak256, toBytes } from "viem";

const { verifier } = await network.connect("sepolia");

const logs = await verifier.etherscan.customApiCall({
  module: "logs",
  action: "getLogs",
  fromBlock: "0",
  toBlock: "latest",
  address: "0x1c7D4B196Cb0C7B01d743Fbc6116a902379C7238",
  topic0: keccak256(toBytes("Transfer(address,address,uint256)")),
});

console.log("Logs:", logs);
```

### DIFF
--- a/v-next/hardhat-verify/src/internal/etherscan.ts
+++ b/v-next/hardhat-verify/src/internal/etherscan.ts
@@ -465,10 +465,8 @@ export class Etherscan implements VerificationProvider {
 
   public async customApiCall(
     params: Record<string, unknown>,
-    options: EtherscanCustomApiCallOptions = {},
+    options: EtherscanCustomApiCallOptions = { method: "GET" },
   ): Promise<EtherscanResponseBody> {
-    const { method = "GET", body = {} } = options;
-
     const queryParams = {
       chainid: this.chainId,
       apikey: this.apiKey,
@@ -477,7 +475,7 @@ export class Etherscan implements VerificationProvider {
 
     let response: HttpResponse;
     try {
-      if (method === "GET") {
+      if (options.method === "GET") {
         response = await getRequest(
           this.apiUrl,
           { queryParams },
@@ -486,7 +484,7 @@ export class Etherscan implements VerificationProvider {
       } else {
         response = await postFormRequest(
           this.apiUrl,
-          body,
+          options.body,
           { queryParams },
           this.dispatcherOrDispatcherOptions,
         );
--- a/v-next/hardhat-verify/src/internal/etherscan.ts
+++ b/v-next/hardhat-verify/src/internal/etherscan.ts
@@ -1,13 +1,16 @@
 import type {
+  EtherscanCustomApiCallOptions,
   EtherscanChainListResponse,
   EtherscanGetSourceCodeResponse,
   EtherscanResponse,
+  EtherscanResponseBody,
+  EtherscanVerifyArgs,
+  LazyEtherscan,
 } from "./etherscan.types.js";
 import type {
   VerificationProvider,
   VerificationResponse,
   VerificationStatusResponse,
-  BaseVerifyFunctionArgs,
   CreateEtherscanOptions,
   ResolveConfigOptions,
 } from "./types.js";
@@ -20,6 +23,7 @@ import type {
   ChainDescriptorsConfig,
   VerificationProvidersConfig,
 } from "hardhat/types/config";
+import type { EthereumProvider } from "hardhat/types/providers";
 
 import { HardhatError } from "@nomicfoundation/hardhat-errors";
 import { toBigInt } from "@nomicfoundation/hardhat-utils/bigint";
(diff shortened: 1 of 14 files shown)


v-next/hardhat-verify/src/internal/etherscan.types.ts
```diff
@@ -1,8 +1,16 @@
+import type { BaseVerifyFunctionArgs } from "./types.js";
+
+export interface EtherscanResponseBody {
+  status: string;
+  message: string;
+  result: any;
```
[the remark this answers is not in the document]
schaable: This is intentional, as it is meant to be consumed by users and we do not want to force narrowing/validation.

v-next/hardhat-verify/src/internal/etherscan.ts
```diff
@@ -462,6 +462,67 @@ export class Etherscan implements VerificationProvider {
...
+        {
+          name: this.name,
+          url: this.apiUrl,
+          errorMessage:
+            error.cause instanceof Error ? error.cause.message : error.message,
+        },
+      );
+    }
+
+    const responseBody =
```
ChristopherDedominici, line 505: Minor, to avoid casting via `as`, you could do:
```
const responseBody: EtherscanResponseBody = await response.body.json();
```

ChristopherDedominici (approved):
I left a very minor comment, other than that, LGTM

v-next/hardhat-verify/src/internal/etherscan.ts
schaable: I think I prefer the assertion in this scenario, as it’s explicit about what we’re doing. If we used a type annotation, readers might think we’re receiving a typed response. The return type should still be type-checked, but that’s for another PR.

PR #7933, opened 2026-02-09
Hardhat 2 eth_getproof support

kanej:
Provide support for `eth_getProof` as a JSON-RPC method in Hardhat 2.

Both Hardhat 2 and Hardhat 3 use the same underlying EDR library to provide `eth_getProof` support. This is based on the v3 version: https://github.com/NomicFoundation/hardhat/pull/7914.

The `eth_getProof` JSON-RPC call is not supported on a locally-mined block following forking. A `ProviderError` is thrown to explain the unsupported case:

```shell
/workspaces/hardhat/examples/hh2/node_modules/hardhat/src/internal/hardhat-network/provider/provider.ts:462
          error = new ProviderError(
                  ^
ProviderError: The action `Proof of locally modified state in fork mode` is unsupported. Obtaining a proof is not supported in fork mode when local changes have been made. E.g. manual changes to an account or a locally mined block.
    at EdrProviderWrapper.request (/workspaces/hardhat/examples/hh2/node_modules/hardhat/src/internal/hardhat-network/provider/provider.ts:462:19)
    at async main (file:///workspaces/hardhat/examples/hh2/scripts/test-get-proof.js:31:20)
```

## Manual Testing 

I have done a round of manual testing and confirmed that EDR returns a readable `ProviderError` (shown above) for the unsupported case.

`eth_getProof` against a forked mainnet gives the same proof as you would get if going directly.

### DIFF
--- a/packages/hardhat-core/package.json
+++ b/packages/hardhat-core/package.json
@@ -103,7 +103,7 @@
   "dependencies": {
     "@ethereumjs/util": "^9.1.0",
     "@ethersproject/abi": "^5.1.2",
-    "@nomicfoundation/edr": "0.12.0-next.22",
+    "@nomicfoundation/edr": "0.12.0-next.23",
     "@nomicfoundation/solidity-analyzer": "^0.1.0",
     "@sentry/node": "^5.18.1",
     "adm-zip": "^0.4.16",
--- a/packages/hardhat-core/test/internal/hardhat-network/jsonrpc/eth-get-proof.ts
+++ b/packages/hardhat-core/test/internal/hardhat-network/jsonrpc/eth-get-proof.ts
@@ -1,0 +1,95 @@
+import { assert } from "chai";
+import { useEnvironment } from "../../../helpers/environment";
+import { useFixtureProject } from "../../../helpers/project";
+import { TASK_COMPILE } from "../../../../src/builtin-tasks/task-names";
+
+describe("eth_getProof", function () {
+  // Set up the test environment - not dependent on the fixture project
+  useFixtureProject("project-0.8.20");
+  useEnvironment();
+
+  it("should return account proof on local network", async function () {
+    await this.env.run(TASK_COMPILE, { quiet: true });
+
+    const accounts = await this.env.network.provider.send("eth_accounts");
+
+    assert.ok(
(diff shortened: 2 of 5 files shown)


packages/hardhat-core/test/internal/hardhat-network/jsonrpc/eth-get-proof.ts
```diff
@@ -0,0 +1,95 @@
+import { assert } from "chai";
+import { useEnvironment } from "../../../helpers/environment";
+import { useFixtureProject } from "../../../helpers/project";
+import { TASK_COMPILE } from "../../../../src/builtin-tasks/task-names";
+
+describe("eth_getProof", function () {
```
kanej, line 6: I added two happy case tests from the original PR. I have opted not to add a forking test - this is covered in EDR and Hardhat 3.

I have done a local manual test to confirm we are getting a `ProviderError` for the unsupported case of `eth_getProof` on a mined block after forking.

packages/hardhat-core/test/internal/solidity/helpers.ts
```diff
@@ -13,14 +13,17 @@ import { getRealPathSync } from "../../../src/internal/util/fs-utils";
...
-    public libraryName?: string
-  ) {
+  constructor(name: string, versionPragmas: string[], libraryName?: string) {
+    this.name = name;
+    this.versionPragmas = versionPragmas;
+    this.libraryName = libraryName;
+
     this.sourceName = `contracts/${name}.sol`;
     this.absolutePath = path.join(projectRoot, "contracts", `${name}.sol`);
   }
```
kanej, line 29: This is a small refactor to avoid issues with type stripping as Node versions change.

packages/hardhat-core/package.json
```diff
@@ -103,7 +103,7 @@
   "dependencies": {
     "@ethereumjs/util": "^9.1.0",
     "@ethersproject/abi": "^5.1.2",
-    "@nomicfoundation/edr": "0.12.0-next.22",
+    "@nomicfoundation/edr": "0.12.0-next.23",
```
kanej, line 106: This is the same version as HH3.

ChristopherDedominici (approved):

PR #7934, opened 2026-02-10
Test how to improve the CI

ChristopherDedominici:
Test PR to explore different solutions for improving flaky tasks in the CI


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -204,7 +203,9 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
     // This is because the mutex blocks access until a compiler has been fully downloaded, preventing any new process
     // from checking whether that version of the compiler exists. Without mutex it might incorrectly
     // return false, indicating that the compiler isn't present, even though it is currently being downloaded.
-    return this.#mutexCompiler.use(async () => {
+    const mutex = new MultiProcessMutex(`compiler-download-${version}`);
```
ChristopherDedominici, line 207: Why the per-version mutex probably matters for CI timeouts:

The `hardhat` package uses NodeJS native test runner, which runs test files in parallel. Multiple test files trigger Solidity compilation, which calls `downloadCompiler()`.

Previously, `downloadCompiler()` used a single shared `MultiProcessMutex("compiler-download")`, a file-based lock at `/tmp/compiler-download.txt`. This meant that across all parallel test workers, only one could download or even check if a compiler exists at a time. Workers needing different compiler versions would still block each other, potentially waiting long enough to hit the test timeout.

The fix makes the mutex version-specific `(compiler-download-${version})`, so workers downloading different versions proceed in parallel. Same-version downloads are still correctly serialized.

alcuadrado:
I think closing this PR now would makes sense, so that we actually test that behavior. The downloads are really fast from the CI, so this doesn't have big performance hit. At the same time, it can hide bugs.

I think all the compiler downloader flakiness was related to the mutex and the bugs fixed in the mutex PR.

PR #7936, opened 2026-02-10
build: upgrade to @nomicfoundation/edr v0.12.0-next.24 (Hardhat 3)

### LINKED ISSUE #7872: Support EIP-7212 precompile at address 0x100 in solidity tests
### Describe the feature

Since Osaka, EIP-7212 adds a precompile for P256 signature verification at address 0x100.

I am able to run tests against that precompile using foundry (in solidity) and hardhat 2 (in javascript). However, when I upgrade to hardhat 3, and run my solidity tests using hardhat 3, it appears that the precompile is not present. 

I've set `networks.default.hardfork = 'osaka'`

### Search terms

precompile 7212

popescuoctavian:

---

Bumped EDR version to [`0.12.0-next.24`](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.24).

### Minor Changes
- NomicFoundation/edr@7bf3576: Added the ability to specify the hardfork when running Solidity tests. Added support for the Osaka hardfork in Solidity tests.
- NomicFoundation/edr@b505164: Added structured errors for unsupported Solidity test cheatcodes

### DIFF
--- a/.changeset/two-numbers-admire.md
+++ b/.changeset/two-numbers-admire.md
@@ -1,0 +1,8 @@
+---
+"hardhat": patch
+---
+
+Bumped EDR version to [`0.12.0-next.24`](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.24).
+
+- NomicFoundation/edr@7bf3576: Added the ability to specify the hardfork when running Solidity tests. Added support for the Osaka hardfork in Solidity tests.
+- NomicFoundation/edr@b505164: Added structured errors for unsupported Solidity test cheatcodes
--- a/.changeset/two-numbers-admire.md
+++ b/.changeset/two-numbers-admire.md
@@ -2,7 +2,4 @@
 "hardhat": patch
 ---
 
-Bumped EDR version to [`0.12.0-next.24`](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.24).
-
-- NomicFoundation/edr@7bf3576: Added the ability to specify the hardfork when running Solidity tests. Added support for the Osaka hardfork in Solidity tests.
-- NomicFoundation/edr@b505164: Added structured errors for unsupported Solidity test cheatcodes
+Fixed missing EIP-7212 precompile in Solidity Tests ([#7872](https://github.com/NomicFoundation/hardhat/issues/7872))
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
@@ -24,6 +24,7 @@ import { hexStringToBytes } from "@nomicfoundation/hardhat-utils/hex";
 import chalk from "chalk";
 
 import { DEFAULT_VERBOSITY, OPTIMISM_CHAIN_TYPE } from "../../constants.js";
+import { hardhatHardforkToEdrSpecId } from "../network-manager/edr/utils/convert-to-edr.js";
 
 import { type Colorizer, formatArtifactId } from "./formatters.js";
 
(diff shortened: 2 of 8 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
kanej: The other invocation of `hardhatHardforkToEdrSpecId` to return a specId in `EDRProvider`, updates the following genesis state as well:

https://github.com/NomicFoundation/hardhat/blob/dbb29e92351c291689a539f68aa1cef1f2ec8a18/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts#L463-L468

v-next/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -158,10 +158,21 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
...
+  // Extract hardfork from the selected network configuration
+  let hardfork: string | undefined;
+  if (hre.globalOptions.network !== undefined) {
+    const networkName = hre.globalOptions.network;
+    const networkConfig = hre.config.networks[networkName];
+    if (networkConfig !== undefined && networkConfig.type === "edr-simulated") {
+      hardfork = networkConfig.hardfork;
+    }
+  }
+
```
kanej, line 170: I think this looks right. Can I check, EDR is responsible for selecting the default hardfork? So if there is not an override from the network, we let EDR decide.

popescuoctavian: We've changed the behavior in EDR to always require a `hardfork` in the `SolidityTestRunnerConfigArgs` (similar to the   provider config). Therefore, Hardhat must control the hardfork. This was done to avoid impact of EDR/REVM defaults to Hardhat behavior.

`resolveHardfork` defaults to the latest hardfork for the chain type when there is no override.

kanej: Ah okay. So we don't have a default hardfork set in config (as we do for other defaults) as it has to be calculated (i.e. are we in optimism?), but it is on the Hardhat side that we resolve to hardfork to use, before passing it to EDR.

popescuoctavian: Yes, that is correct.

.changeset/two-numbers-admire.md
```diff
@@ -0,0 +1,8 @@
+---
+"hardhat": patch
+---
+
+Bumped EDR version to [`0.12.0-next.24`](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.24).
+
+- NomicFoundation/edr@7bf3576: Added the ability to specify the hardfork when running Solidity tests. Added support for the Osaka hardfork in Solidity tests.
+- NomicFoundation/edr@b505164: Added structured errors for unsupported Solidity test cheatcodes
```
kanej:
```suggestion
Fixed missing EIP-7212 precompile in Solidity Tests ([#7872](https://github.com/NomicFoundation/hardhat/issues/7872))
```
This changeset is likely to be the only one in the release. I have went with a message at the Hardhat level.

v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
popescuoctavian: I don't have 100% of the context here, but I believe the genesis state doesn't have any impact on Solidity tests -- we are not exposing any config for that and I can't find any related code to it in Solidity tests related code.

.changeset/two-numbers-admire.md
popescuoctavian: Thanks, I've applied the suggestion

v-next/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
kanej: The genesis state here is being passed as the `localPredeploys` for the `SolidityTestRunnerConfigArgs`.

```ts
/**
   * The predeploys applied in local mode. Defaults to no predeploys.
   * These should match the predeploys of the network in fork mode, so they
   * aren't set in fork mode.
   * The code must be set and non-empty. The nonce and the balance default to
   * zero and storage defaults to empty.
   */
  localPredeploys?: Array<AccountOverride>
```

The predeploys are being set in the code above by `opGenesisState` and `l1GenesisState` based on the latest hardfork in each chain. I suspect the most accurate simulation is to base it off of the hardfork being used, so `opHardforkFromString(resolvedHardfork))` and `l1HardforkFromString(resolvedHardfork)`.

My mental model here is Foundry does not really support anything other than mainnet and forking so its codebase doesn't really consider changing predeploys, but EDR with multi-chain does.

popescuoctavian: That makes sense, thanks for the details! Addressed by dd6daf40b309cfaf0dd71ad3fda26a328a4eb79f.

kanej (approved):
This looks great - thanks @popescuoctavian and @ChristopherDedominici.

I did a quick test against the OpenZeppelin contracts repo and the precompile tests now pass:

Image

Lets get this merged!

PR #7942, opened 2026-02-11
Introduce new inter-process mutex implementation

alcuadrado:
This PR introduces a new implementation of the `MultiProcessMutex`, which inverts the behavior when compared to the previous one.

Instead of considering a lock staled based on the time that the process has been waiting for it, it does it when the process that acquired it no longer exists. This can be really quick, without having to wait much. It's also less error prone.

Please read the jsdoc for more info.

This PR also fixes three related bugs:

1. The `download` function in the request module was always using the same tmp path for every file with the same name, leading to more race-y code. This was only triggered because the locks weren't working (see 3)
2. The `download` function didn't properly wait for the file to be closed before moving it.
3. The compiler downloader could use different paths (based on new tmp dirs each time) for the lock in each process. Leading to them potentially not working.


alcuadrado:
@ChristopherDedominici this is ready to be reviewed now

.changeset/chilled-ads-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-utils": patch
+---
+
+Introduce new inter-process mutex implementation
```
ChristopherDedominici: Minor:
We usually add a link to the issue, if available, or a link to the PR itself. I noticed that John always manually adds them during the release if they are missing, so this should help him.
Example:

```
Introduce new inter-process mutex implementation ([7942](https://github.com/NomicFoundation/hardhat/pull/7942)).
```

v-next/hardhat-utils/src/synchronization.ts
```diff
@@ -1,164 +1,596 @@
...
+
+type StalenessResult =
+  | { isStale: true; metadata: LockMetadata | undefined }
+  | { isStale: false };
+
+type AcquireResult =
+  | { acquired: true }
+  | { acquired: false; reclaimedStaleLock: boolean };
+
+/**
```
ChristopherDedominici, line 57: I think there might be a concurrency issue here, I'll try to explain in here, so you can tell me if you agree:

The release is a two-step operation (delete metadata, then delete directory), and `{ recursive: true }` on `rmdirSync` can cause it to silently destroy another process's lock if that process acquired between the two steps:

```
TIME
 │
 │  PROCESS A (releasing)                          PROCESS B (acquiring)
 │  ─────────────────────                          ──────────────────────
 │  #releaseLock():
 │    rmSync(metadataPath) ✓              (L255)
 │         ↓
 │         ║                               #tryAcquire():
 │         ║                                 mkdirSync(lockPath) → EEXIST  (L295)
 │         ║                               #checkStaleness():
 │         ║                                 #readMetadata() → undefined   (L367→L498)
 │         ║                                 metadata === undefined         (L370)
 │         ║                                 return { isStale: true }       (L374)
 │         ║                               #tryUnlockingStaleLock():
 │         ║                                 rmSync(metadataPath) → ENOENT (L424)
 │         ║                                 rmdirSync(lockPath) ✓         (L435)
 │         ║                               #tryAcquire() (retry via L238→L222):
 │         ║                                 mkdirSync(lockPath) ✓         (L295)
 │         ║                                 #activeLocks.add(lockPath)    (L297)
 │         ║                               #writeMetadata():
 │         ║                                 writeFileSync(metadata) ✓     (L477)
 │         ║                                 return { acquired: true }     (L363)
 │         ║                               #acquireLock() returns           (L224→L225)
 │         ║                               ** B enters critical section **
 │         ↓
 │  #releaseLock() continued:
 │    rmdirSync(lockPath, {recursive:true})         (L266)
 │    → SUCCEEDS because recursive:true
 │      deletes B's metadata AND the directory!
 │    #activeLocks.delete(lockPath)                  (L286)
 │    ** A thinks release is done **
 │
 │                                         ** B's lock is gone, but B doesn't know **
 │
 │                                                         PROCESS C (acquiring)
 │                                                         ──────────────────────
 │                                                         #tryAcquire():
 │                                                           mkdirSync(lockPath) ✓  (L295)
 │                                                         #writeMetadata():
 │                                                           writeFileSync() ✓      (L477)
 │                                                           return { acquired: true }
 │                                                         ** C enters critical section **
 │
 │  ⚠️  B and C are BOTH in the critical section = mutex violated
 ▼
```

Without `{ recursive: true }`, `rmdirSync` on a non-empty directory fails with `ENOTEMPTY`,
which might be the correct signal that the directory was re-acquired by someone else.

v-next/hardhat-utils/src/synchronization.ts
```diff
@@ -1,164 +1,596 @@
...
+        !("hostname" in parsed) ||
+        !("createdAt" in parsed) ||
+        !("platform" in parsed) ||
+        typeof parsed.pid !== "number" ||
+        typeof parsed.hostname !== "string" ||
+        typeof parsed.createdAt !== "number" ||
+        typeof parsed.platform !== "string" ||
+        Number.isSafeInteger(parsed.pid) === false ||
+        parsed.pid < 1 ||
+        Number.isSafeInteger(parsed.createdAt) === false ||
```
ChristopherDedominici, line 617: Minor: the `pid` property includes a check ensuring the number is not less than 1, but a similar check does not exist for the `createdAt` property, e.g.: createdAt cannot be negative

v-next/hardhat-utils/src/synchronization.ts
```diff
@@ -1,164 +1,596 @@
...
     try {
-      log(`Deleting mutex file at path '${this.#mutexFilePath}'`);
-      fs.unlinkSync(this.#mutexFilePath);
+      // Write metadata inside the lock directory
+      MultiProcessMutex.#writeMetadata(lockPath);
     } catch (e) {
-      ensureNodeErrnoExceptionError(e);
+      log("Failed to write metadata, cleaning up and rethrowing");
 
+      MultiProcessMutex.#releaseLock(lockPath);
```
ChristopherDedominici: If this line throws an error, the original error `e` (line 339) is lost

ChristopherDedominici (changes requested):

v-next/hardhat-utils/src/synchronization.ts
alcuadrado: good catch!

alcuadrado: This is the fix i'm implementing:

Instead of deleting the directory with a recursive rmdirSync, delete it without recursive. This should throw NOEMPTY if another process already created its metadata file. If that's the case, we don't consider it an error.

v-next/hardhat-utils/src/synchronization.ts
alcuadrado: good point. This cleanup should be a best-effort. If it fails, we continue with the loop.

ChristopherDedominici (approved):

alcuadrado:
Let's go for the lockfile version instead, because it's simpler to reason about, and the chances of people using hardhat in a network filesystem (AKA NFS) are really low.

PR #7948, opened 2026-02-12
add perf diagnostics to hardhat mocha

kanej:
Adds performance measures around the key subtasks of the mocha test task. Leverages the Node performance API, and logs using Hardhat `debug` logging system.

The performance measures are tracked during the run, and printed at the end as one batch of measures.

To see the performance results of mocha test set the debug ENV var:

```shell
DEBUG="hardhat:mocha:performance" npx hardhat test mocha
```

## Example

Taking the `./v-next/example-project`, then running its Mocha test suite gives:

image

### DIFF
--- a/v-next/hardhat-mocha/package.json
+++ b/v-next/hardhat-mocha/package.json
@@ -45,6 +45,7 @@
   "devDependencies": {
     "@nomicfoundation/hardhat-node-test-reporter": "workspace:^3.0.0",
     "@nomicfoundation/hardhat-test-utils": "workspace:^",
+    "@types/debug": "^4.1.7",
     "@types/mocha": ">=10.0.10",
     "@types/node": "^22.0.0",
     "c8": "^9.1.0",
@@ -59,6 +60,7 @@
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.5",
     "@nomicfoundation/hardhat-zod-utils": "workspace:^3.0.0",
     "chalk": "^5.3.0",
+    "debug": "^4.3.2",
     "tsx": "^4.19.3",
     "zod": "^3.23.8"
   },
--- a/v-next/hardhat-mocha/src/performance.ts
+++ b/v-next/hardhat-mocha/src/performance.ts
@@ -1,0 +1,199 @@
+import { performance } from "node:perf_hooks";
+
+/**
+ * A tracker for measuring the performance of an operation and its subphases.
+ * It leverages Node's performance APIs under the hood.
+ *
+ * NOTE: This interface was built to support `hardhat-mocha`, you should
+ * consider carefully whether its approach applies in your context.
+ *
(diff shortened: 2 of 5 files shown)


v-next/hardhat-mocha/package.json
```diff
@@ -59,6 +60,7 @@
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.5",
     "@nomicfoundation/hardhat-zod-utils": "workspace:^3.0.0",
     "chalk": "^5.3.0",
+    "debug": "^4.3.2",
```
kanej, line 63: This matches the pattern of pulling in `debug` used in other plugins.

v-next/hardhat-mocha/src/task-action.ts
```diff
@@ -18,13 +19,30 @@ import {
...
   noCompile: boolean;
 }
 
+type PerformancePhase =
+  | "Build"
+  | "Get test files"
+  | "Mocha setup"
+  | "Test file loading"
+  | "Test execution"
+  | "Reporting";
```
kanej, line 37: We could get cleverer, but this along with the runtime checks the performance module does for starting/ending gives us what we need for a good sanity check.

v-next/hardhat-mocha/src/performance.ts
```diff
@@ -0,0 +1,197 @@
...
+ * perf.end();
+ *
+ * perf.logInto(performanceLog);
+ * perf.clear();
+ * ```
+ *
+ * @typeParam PhaseT - A string union of the valid phase names e.g.
+ * "Build" | "Reporting"
+ */
+export interface PerformanceTracker<PhaseT extends string> {
```
kanej, line 46: This is built for the particular `hardhat-mocha` use case - it could be considered for moving to `hardhat-utils` so I have tried to apply the standard practices (e.g. jsdocs) here.

Currently we don't intend to move it over - performance measures being typically a bespoke thing.

v-next/hardhat-mocha/src/performance.ts
```diff
@@ -0,0 +1,197 @@
+/**
```
schaable, line 3:
```suggestion
import { performance } from "node:perf_hooks";

/**
```

Minor: Can we explicitly import `performance` here? I know it's globally available, but I'd prefer to avoid any ambiguity.

schaable (approved):
Very clean PR 💚 I left one minor comment, but I’ll leave the decision to you. Pre-approving.

v-next/hardhat-mocha/src/performance.ts
kanej: Applied - I have also opened a PR to add an eslint rule restricting certain globals including `performance`: https://github.com/NomicFoundation/hardhat/pull/7951

PR #7949, opened 2026-02-13
build: upgrade to @nomicfoundation/edr v0.12.0-next.25

### LINKED ISSUE #7963: Add support for -vvvv in JS/TS tests
Currently tracing through `-vvvv` is only available in Solidity Tests. This feature should be included in JS/TS and scripts.

## TODO

- [ ] https://github.com/NomicFoundation/hardhat/pull/7949

Wodann:

---

Upgraded EDR to `v0.12.0-next.25`:

### Minor Changes

- NomicFoundation/edr@3974769: Added `callTraces()` to `Response` object, inclusion of which is configurable through the `includeCallTraces` option on the `ObservabilityConfig`
- NomicFoundation/edr@f4bdc36: Removed `getLatestSupportedSolcVersion` API

  BREAKING CHANGE: A new API `latestSupportedSolidityVersion` was previously introduced to replace the deprecated `getLatestSupportedSolcVersion`. The old API has now been removed. Users should update their code to use `latestSupportedSolidityVersion` instead.

- NomicFoundation/edr@3974769: Removed `traces()` API from the `Response` object
- NomicFoundation/edr@f4bdc36: Added support to the `debug_traceCall` & `debug_traceTransaction` JSON-RPC methods for different tracers (`4byteTracer`, `callTracer`, `flatCallTracer`, `prestateTracer`, `noopTracer`, and `muxTracer`).

  Our API is now aligned with Geth's tracing capabilities.

  BREAKING CHANGE: Memory capture used to be enabled by default on geth, but has since been flipped <https://github.com/ethereum/go-ethereum/pull/23558> and is now disabled by default. We have followed suit and disabled it by default as well. If you were relying on memory capture, you will need to explicitly enable it by setting the `enableMemory` option to `true` in your tracer configuration.

## TO DO
I wasn't clear on standards in Hardhat 3 for this, so I figured it would be easier for a Hardhat developer to quickly implement this:

- [ ] `v-next/hardhat/src/internal/builtin-plugins/solidity-test/formatters.ts` needs to be moved to a package that is shared between Solidity Tests and the JSON-RPC provider
- [ ] A configuration option for the verbosity level should be added that can be configured via CLI (e.g. `-vvvv`) similar to Solidity tests
- [ ] The EDR `ProviderConfig` should receive the `IncludeTraces` configuration, based on the configured verbosity level.
- [ ] The `formatTraces` function should be called in `edr-provider.ts`, depending on the verbosity level

If no one is available, I can do it with some direction on CLI configuration options and preferred file locations.


kanej:
On this question:

```
- [ ] v-next/hardhat/src/internal/builtin-plugins/solidity-test/formatters.ts needs to be moved to a package that is shared between Solidity Tests and the JSON-RPC provider
```

I think we move the formatters up into `network-manager` (both Solidity Test and the Network Manager are built-in so part of the Hardhat package), and then reuse them from the Solidity Test plugin.

So maybe move the formatters to:

```
v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/trace-formatters.ts
```

@schaable, what do you think?

schaable:
> On this question:
> 
> ```
> - [ ] v-next/hardhat/src/internal/builtin-plugins/solidity-test/formatters.ts needs to be moved to a package that is shared between Solidity Tests and the JSON-RPC provider
> ```
> 
> I think we move the formatters up into `network-manager` (both Solidity Test and the Network Manager are built-in so part of the Hardhat package), and then reuse them from the Solidity Test plugin.
> 
> So maybe move the formatters to:
> 
> ```
> v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/trace-formatters.ts
> ```
> 
> @schaable, what do you think?

Agree, but not all exports in that file are related to stack traces. `formatArtifactId` is still specific to `solidity-test`, the `Colorizer` interface seems more generic, and I don't think `formatLogs` is used. Maybe we split it into:

 - `v-next/hardhat/src/internal/builtin-plugins/solidity-test/formatters.ts` (keep `formatArtifactId` here)
 - `v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/trace-formatters.ts` (move `formatTraces` and related helpers here)
 - `v-next/hardhat/src/internal/utils/colorizer.ts` (move `Colorizer` here, unless there's a better place)
 
We can also remove `formatLogs` and its tests. All this could be done as a prep PR ahead of this one.

ChristopherDedominici:
Closed in favor of https://github.com/NomicFoundation/hardhat/pull/7983

PR #7950, opened 2026-02-13
inline task actions - run time checks

### LINKED ISSUE #7560: Inline task actions

ChristopherDedominici:
Same as [7851](https://github.com/NomicFoundation/hardhat/pull/7851), but here checks are mostly executed at runtime. This PR implements the comment from the PR linked above.

Link to the website updated docs PR: https://github.com/NomicFoundation/hardhat-website/pull/214

### Questions
- This PR introduces breaking changes. Should they be added to the changelog and the website documentation, and how should we communicate them to the users?
**ANSWER**: changeset is fine, we're still in beta


v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -694,6 +694,22 @@ Please double check your arguments.`,
...
+        number: 417,
+        messageTemplate: `The task "{task}" defines an "inlineAction", which is not allowed in plugins.`,
+        websiteTitle: "Inline action not allowed in plugins",
+        websiteDescription: `Plugins are not allowed to use inline actions for tasks.
+
+Please define the action in a separate file and reference it.`,
+      },
+      ACTION_ALREADY_SET: {
+        number: 418,
+        messageTemplate: `The action for task "{task}" has already been set. You can only call "setAction" or "setInlineAction" once per task definition.`,
```
ChristopherDedominici, line 707: Self reminder: If we change to "lazy actions", this must be updated

v-next/hardhat/src/internal/core/tasks/builders.ts
```diff
@@ -54,7 +56,11 @@ export class EmptyTaskDefinitionBuilderImplementation
 
 export class NewTaskDefinitionBuilderImplementation<
   TaskArgumentsT extends TaskArguments = TaskArguments,
-> implements NewTaskDefinitionBuilder<TaskArgumentsT>
+  ActionTypeT extends
+    | "LAZY_ACTION"
```
ChristopherDedominici, line 60: LAZY_ACTION or ACTION? This depends if we want to use `setAction` or `setLazyAction`

v-next/hardhat/src/internal/core/tasks/validations.ts
```diff
@@ -23,6 +29,40 @@ export function validateId(id: string | string[]): void {
...
+  if (isPlugin && inlineAction !== undefined) {
+    throw new HardhatError(
+      HardhatError.ERRORS.CORE.TASK_DEFINITIONS.INLINE_ACTION_CANNOT_BE_USED_IN_PLUGINS,
+      { task: formatTaskId(taskId) },
+    );
+  }
+
+  if (action !== undefined && inlineAction !== undefined) {
+    throw new HardhatError(
+      HardhatError.ERRORS.CORE.TASK_DEFINITIONS.ACTION_ALREADY_SET,
```
alcuadrado: This error can be confusing. Maybe we should add a descriptor.

v-next/hardhat/src/internal/core/config-validation.ts
```diff
@@ -350,6 +338,51 @@ export function validateTaskOverride(
   return validationErrors;
 }
 
+function validateActionFields(
```
alcuadrado, line 344: This function should also be able to validate (somehow, with some refactor in the caller) that a plugin shouldn't use an inline action.

v-next/hardhat/src/types/tasks.ts
```diff
@@ -90,37 +101,44 @@ export interface EmptyTaskDefinition {
...
 
   positionalArguments: PositionalArgumentDefinition[];
 }
 
 /**
- * An override of an existing task.
+ * The definition of a new task.
  */
-export interface TaskOverrideDefinition {
+export type NewTaskDefinition = BaseTaskDefinition & TaskAction;
```
alcuadrado, line 121: It's unclear to me if this is a breaking change or not, but we are introducing some anyway, so it's ok

v-next/hardhat/src/internal/core/tasks/validations.ts
alcuadrado: The reason being that here you aren't setting anything. It should be more like "Invalid task definition with both action and lazy action"

alcuadrado (changes requested):
I left two small comments, but otherwise LGTM

v-next/hardhat/src/types/tasks.ts
```diff
@@ -75,6 +75,17 @@ export enum TaskDefinitionType {
   TASK_OVERRIDE = "TASK_OVERRIDE",
 }
 
+export type TaskAction =
+  | { action: LazyActionObject<NewTaskActionFunction>; inlineAction?: never }
```
alcuadrado, line 79: To be more precise: I think we don't need https://github.com/NomicFoundation/hardhat/pull/7974, but instead just rename `action` to `lazyAction` here, in `TaskOverrideAction`, and the builders.

v-next/hardhat/src/types/tasks.ts
```diff
@@ -172,15 +197,38 @@ export interface NewTaskDefinitionBuilder<
   /**
    * Sets the action of the task.
    *
-   * It can be provided as a function, or as a `file://` URL pointing to a file
+   * It must be provided as a `file://` URL pointing to a file
    * that exports a default NewTaskActionFunction.
```
alcuadrado: This is outdated. It has to be a function that looks like this `() => import(...)`

v-next/hardhat/src/internal/core/config-validation.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7950/changes/89b2c28183a09e080e6718e4d6e2f710371dedf0

v-next/hardhat/src/types/tasks.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7950/changes/d29d903d1f439ab56c956a3c870f5a2a1edaff38

v-next/hardhat/src/internal/core/tasks/validations.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7950/changes/0162dff035339a2829f2717ccca29a7b4ddba74e

v-next/hardhat/src/types/tasks.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/7950/changes/7a9a31a0c1b081f468ec0250a49c40f65b8e2653

I only updated the types. So the user still has to use `setAction` and NOT `setLazyAction`. Also, all the errors are still showing `action` and NOT `lazyAction`, except this scenario where the `path` must reflect the property name:
image

.changeset/smooth-socks-dance.md
```diff
@@ -0,0 +1,7 @@
+---
+"@nomicfoundation/hardhat-errors": patch
+"@nomicfoundation/hardhat-verify": patch
```
alcuadrado:
```suggestion
```

This one isn't changed in the PR anymore

alcuadrado (approved):

PR #7951, opened 2026-02-13
chore: add an eslint rule on implicit use of performance

kanej:
There are globals that are added implicitly in Node. We prefer an explicit approach.

This PR adds a new global eslint rule has been added to restrict certain globals, specifically: `performance`, `fetch` and `structuredClone`.

## performance

`perforamance` should be imported explicitly e.g.

```ts
import { performance } from "node:perf_hooks";
```

## fetch

`fetch` should be done with `undici`.

## structureClone

`structureClone` should be avoided, with `deepClone` being the likely replacement from `hardhat-utils`.

### DIFF
--- a/v-next/config/eslint.config.js
+++ b/v-next/config/eslint.config.js
@@ -344,6 +344,24 @@ export function createConfig(
       },
     ],
     "use-isnan": "error",
+    "no-restricted-globals": [
+      "error",
+      {
+        name: "performance",
+        message:
+          "Import from 'node:perf_hooks' explicitly e.g. `import { performance } from \"node:perf_hooks\";`",
+      },
+      {
+        name: "fetch",
+        message:
+          'Consider using `request` from `hardhat-utils e.g. `import { getRequest } from "@nomicfoundation/hardhat-utils/request";`',
+      },
+      {
+        name: "structuredClone",
+        message:
+          'Avoid structuredClone, consider `deepClone` instead e.g. `import { deepClone } from "@nomicfoundation/hardhat-utils/lang";`',
+      },
+    ],
     "no-restricted-imports": [
       "error",
       {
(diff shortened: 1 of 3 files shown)


v-next/hardhat/src/internal/cli/telemetry/sentry/transport.ts
```diff
@@ -177,6 +177,7 @@ async function sendSerializedEnvelopeToSentryBackend(
     .filter(Boolean)
     .join(", ");
 
+  // eslint-disable-next-line no-restricted-globals -- TODO: determine whether this should be using undici
```
kanej, line 180: Should this usage be moved over to `undici`?

v-next/config/eslint.config.js
```diff
@@ -344,6 +344,24 @@ export function createConfig(
...
+      "error",
+      {
+        name: "performance",
+        message:
+          "Import from 'node:perf_hooks' explicitly e.g. `import { performance } from \"node:perf_hooks\";`",
+      },
+      {
+        name: "fetch",
+        message:
+          "Import from 'undici' explicitly e.g. `import { fetch } from \"undici\";`",
```
schaable: I think importing `undici` directly is also not allowed. We should use `hardhat-utils/request` instead.

v-next/hardhat/src/internal/cli/telemetry/sentry/transport.ts
schaable: It seems this should use `hardhat-utils/request`, but we currently don't have a generic POST helper there. We only expose `postJsonRequest` and `postFormRequest`, both of which set a specific `Content-Type` and transform the body accordingly. 

We could add a `postRequest` that doesn't set the header and leaves body handling to the caller. That said, introducing it would likely require additional tests, and I’m not sure it’s worth the effort just to replace this call.

I’d leave the TODO for now and revisit it if we run into more cases that require a generic POST helper.

schaable (approved):
We should update the message in the `fetch` case, but aside from that it looks good.

v-next/config/eslint.config.js
kanej: Cool. I have updated the message to point towards `getRequest` in Hardhat utils.

PR #7954, opened 2026-02-15
chore: add isDeprecated flag to ErrorDescriptor interface

### LINKED ISSUE #7732: Add support for deprecating error descriptors
### Describe the issue

Unused error descriptors can’t be deleted because they’re shown on the website and older Hardhat versions may still throw them. We should add a deprecated flag to error descriptors and update the website to clearly mark them as deprecated and suggest upgrading Hardhat.

jose-compu:
Add optional `isDeprecated` field to ErrorDescriptor so unused error descriptors can be marked as deprecated instead of deleted. The CLI error handler now shows a deprecation notice for deprecated errors.

Closes #7732

---

### Changes

- Added optional `isDeprecated?: true` field to the `ErrorDescriptor` interface in `hardhat-errors`
- Updated the CLI error handler to append a yellow deprecation notice when a deprecated error is thrown (both core and plugin errors)
- Added tests for the new field in `hardhat-errors/test/errors.ts` (4 tests)
- Added tests for the deprecation notice in `hardhat/test/internal/cli/error-handler.ts` (2 tests)

### How it works

When an error descriptor has `isDeprecated: true`, the CLI output includes:

> Note: This error is deprecated and will be removed in a future version of Hardhat. Please consider upgrading Hardhat.

This allows the website to clearly mark deprecated errors and avoids deleting descriptors that older Hardhat versions may still throw.

### DIFF
--- a/.changeset/sharp-pets-tap.md
+++ b/.changeset/sharp-pets-tap.md
@@ -1,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-errors": patch
+"hardhat": patch
+---
+
+chore: add isDeprecated flag to ErrorDescriptor interface
--- a/v-next/hardhat-errors/src/descriptors.ts
+++ b/v-next/hardhat-errors/src/descriptors.ts
@@ -33,6 +33,16 @@ export interface ErrorDescriptor {
    * can use markdown.
    */
   websiteDescription: string;
+
+  /**
+   * `true` if this error descriptor is deprecated.
+   *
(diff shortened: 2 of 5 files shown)


alcuadrado:
Thanks for this PR, @jose-blockchain. Unfortunately, it's not really helpful. The reason is that we don't delete error descriptors, as we use them to generate the error documentation in the website. And we want older versions of Hardhat to still be able to print valid links to deprecated errors

jose-compu:
Ok, got it @alcuadrado , thanks.

PR #7973, opened 2026-02-17
add `connectOnBefore` network connect helper function to `hardhat-mocha`

### LINKED ISSUE #7965: Add network connect helper function to `hardhat-mocha`
The current pattern of `const connection = await network.connect()` at the top level of a mocha test file can lead to poor performance in large test suites.

Mocha loads all test files before starting a run, triggering an evaluation of each file. Hardhat 3's top level network connection pattern leads to a Hardhat in-memory network being created per file before any tests are run.

To avoid this we are adding a new `connectOnBefore` method to allow for a top level network connection in the test suite without incurring the performance penalty:

```ts
connectOnBefore(
    networkOrParams?: NetworkConnectionParams<ChainTypeT> | string,
    closeOnAfter?: boolean,
): NetworkConnection<ChainTypeT>;
```

`connectOnBefore` is a Mocha test helper function that returns a proxy network connection while registering a Mocha `before` and `after` hook in the surrounding `describe`. The `before` hook will resolve the connection before the first test is run. The connection is automatically closed in an `after` hook unless `closeOnAfter` is set to `false`.

```ts

describe("Token", () => {
  const { ethers } = network.mocha.connectOnBefore();
 
  before(async () => {
      counter = await ethers.deployContract("Token");
  });

  it("has the correct total supply", async function () {
    const totalSupply = await token.totalSupply();
    assert.equal(totalSupply, ethers.parseEther("1000000"));
  });
});
```

This is the recommended way to set up a network connection in Hardhat Mocha test suites. The returned proxy allows safe destructuring of `provider`, `ethers`, `networkHelpers`, while deferring the network setup to the actual run of the test.

kanej:
Provide an alternative to the top level `const connection = await network.connect()` pattern in Mocha test suites - as this triggers the creation of an in-memory Hardhat network for every test file on initial load.

Instead the helper function `connectOnBefore()` allows for a proxy connection, that will only be initialized when the test suite is being run:

```ts
import type { Counter } from "../../types/ethers-contracts/contracts/Counter.js";

describe("Counter", function () {
  const { ethers } = network.mocha.connectOnBefore();

  let counter: Counter;
  beforeEach(async function () {
    counter = await ethers.deployContract("Counter");
  });

  it("Should emit the Increment event when calling the inc() function", async function () {
    await expect(counter.inc()).to.emit(counter, "Increment").withArgs(1n);
  });
});
```

Docs: https://github.com/NomicFoundation/hardhat-website/issues/216
Resolves https://github.com/NomicFoundation/hardhat/issues/7965

## TODO

- [x] Add the hook to proved `network.mocha.connectOnBefore()`
- [x] Add tests to cover the new helper
  - [x] Cover standard connect case
  - [x] Add tests to cover `ethers` usage
  - [x] Add tests to cover chai matchers usage
  - [x] Add tests to cover `network-helpers` usage
  - [x] Add unit tests for createProxy
- [x] Update the mocha/ethers template project to use the pattern
- [x] Update the linked issue description and add it to changeset
- [x] Review types
- [x] Review Proxy code
- [x] Review error handling in proxy - pull in HardhatErrors
- [x] Manual testing
- [ ] Add docs PR showing off the pattern in the ethers/mocha testing guide

## Manual testing

- [x] example project mocha tests using `connectOnBefore`
- [x] verdaccio install of mocha project
- [x] run against OpenZeppelin contracts `hardhat3` branch

# Review

The reviewer should review this PR commit at a time. At its heart this change is to:

- Add the `connectOnBefore` helper under `./v-next/hardhat-mocha/src/connect-on-before`
- Add an `hre` hook in `./workspaces/hardhat/v-next/hardhat-mocha/src/hookHandlers/hre.ts`, that sets up `hre.network.mocha.connectOnBefore`
- Add type extensions for the new `mocha` object in `./v-next/hardhat-mocha/src/type-extensions.ts`
- A unit test suite for the proxy under `./v-next/hardhat-mocha/test/create-network-connection-proxy.ts`
- Integration tests running the Mocha test task for a fixture project that leverages `connectOnBefore` under different usage examples
- A cross cutting cleanup of other fixture projects to ensure that the workspace version of Hardhat packages are used consistently (to avoid build errors from this PRs type extensions)

## Local testing

The example project is setup with the new Mocha test pattern:

```shell
cd ./v-next/example-project
pnpm build
pnpm hardhat test mocha ./test/mocha/connect-on-before.ts
```

## Questions

- [x] Does the `hardhat-mocha` need a plugin dependency on the network manager plugin if we are extending the network manager type?
  - No, we are happy that we can rely on built-in plugins being loaded first
- [x] Are we happy with the pnpm-lock update to cleanup the fixture projects

## Peer bump reasoning

There is no need for a peer-bump. `hardhat-errors` is a direct dependency, and the Mocha chai matchers add only tests, there is no change to a used API in `hardhat-mocha`.

## Potential followup work

- [ ] Expand the readme to cover all configuration options e.g. `parallel`
- [ ] Expand the readme with usage suggestions e.g. `connectOnBefore`


alcuadrado:
> Does the hardhat-mocha need a plugin dependency on the network manager plugin if we are extending the network manager type?

Not necessary, because builtin plugins are guaranteed to be loaded before the external ones.

v-next/hardhat-mocha/src/type-extensions.ts
```diff
@@ -22,3 +28,21 @@ declare module "hardhat/types/test" {
...
+  }
+}
+
+export interface MochaNetworkHelpers<
+  ChainTypeT extends ChainType | string = DefaultChainType,
+> {
+  connectOnBefore(
+    networkOrParams?: NetworkConnectionParams<ChainTypeT> | string,
+    closeOnAfter?: boolean,
+  ): BaseNetworkConnection<ChainTypeT>;
```
alcuadrado: Why not?

```suggestion
  ): NetworkConnection<ChainTypeT>;
```

v-next/hardhat-mocha/src/create-connect-on-before.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    // Optionally tear down after tests.
+    after(async function () {
+      if (resolved !== undefined && closeOnAfter) {
+        await resolved.close();
+        resolved = undefined;
+      }
+    });
+
+    // Creates a lazy proxy for a given property path. When a user destructures
```
alcuadrado: I would move this to its own file so that it can be tested independently

alcuadrado:
We should make sure that this works with the network helpers' fixtures

alcuadrado:
> We should make sure that this works with the network helpers' fixtures

This works:

```ts
import type { NetworkConnection } from "hardhat/types/network";

async function deployCounter(connection: NetworkConnection) {
  const counter = await connection.ethers.deployContract("Counter");
  return { counter };
}

describe("Counter foo", async function () {
  const { ethers, networkHelpers } = network.mocha.connectOnBefore();

  it("Should emit the Increment event when calling the inc() function", async function () {
    console.log(networkHelpers);
    const { counter } = await networkHelpers.loadFixture(deployCounter);
    await expect(counter.inc()).to.emit(counter, "Increment").withArgs(1n);
  });

  it("The sum of the Increment events should match the current value", async function () {
    const { counter } = await networkHelpers.loadFixture(deployCounter);

    const deploymentBlockNumber = await ethers.provider.getBlockNumber();

    // run a series of increments
    for (let i = 1; i <= 10; i++) {
      await counter.incBy(i);
    }

    const events = await counter.queryFilter(
      counter.filters.Increment(),
      deploymentBlockNumber,
      "latest",
    );

    // check that the aggregated events match the current value
    let total = 0n;
    for (const event of events) {
      total += event.args.by;
    }

    expect(await counter.x()).to.equal(total);
  });
});
```

v-next/hardhat-mocha/src/type-extensions.ts
kanej: Done

v-next/hardhat-mocha/src/create-connect-on-before.ts
kanej: I have moved it out and added a unit test suite against it.

v-next/hardhat-ethers-chai-matchers/test/fixture-projects/connect-on-before/contracts/Counter.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
```
kanej, line 2: I added tests to the `hardhat-ethers-chai-matchers` package for `connectOnBefore` to avoid a cyclic dependency between this package and `hardhat-mocha`.

v-next/hardhat-ethers-chai-matchers/test/fixture-projects/connect-on-before/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "module",
+  "devDependencies": {
+    "@nomicfoundation/hardhat-ethers": "workspace:*",
+    "@nomicfoundation/hardhat-mocha": "workspace:*",
```
alcuadrado, line 5: it never crossed my mind that this was possible 🤯🤯🤯 awesome!

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,215 @@
...
+
+      if (prop === Symbol.toPrimitive || prop === Symbol.toStringTag) {
+        if (resolved === null || resolved === undefined) {
+          return undefined;
+        }
+
+        return Reflect.get(resolved, prop);
+      }
+
+      if (prop === "then") {
```
alcuadrado, line 41: This is a bit unexpected

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,215 @@
...
+
+      // Not yet resolved — return a nested proxy so destructuring works.
+      // `const { ethers } = connectOnBefore(...)` will capture a proxy here
+      // that defers to `resolved.ethers` once the `before` hook fires.
+      return createNestedProxyForPath(() => {
+        const r = getResolved();
+
+        if (r === undefined) {
+          return undefined;
+        }
```
alcuadrado, line 73: Shouldn't this throw? i think it would be equivalent to accessing a property of `undefined`. it's `undefined[prop]`

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,215 @@
...
+        );
+      }
+
+      return Reflect.set(resolved, prop, value);
+    },
+
+    has(_obj, prop) {
+      const resolved = getResolved();
+
+      if (resolved === null || resolved === undefined) {
```
alcuadrado, line 103: Maybe also throw?

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,215 @@
...
+        return false;
+      }
+
+      return Reflect.has(resolved, prop);
+    },
+
+    ownKeys() {
+      const resolved = getResolved();
+
+      if (resolved === null || resolved === undefined) {
```
alcuadrado, line 115: Maybe also throw?

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,215 @@
...
+        return [];
+      }
+
+      return Reflect.ownKeys(resolved);
+    },
+
+    getOwnPropertyDescriptor(_obj, prop) {
+      const resolved = getResolved();
+
+      if (resolved === null || resolved === undefined) {
```
alcuadrado, line 127: Maybe also throw?

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,215 @@
...
+    },
+  });
+}
+
+// Creates a lazy proxy for a given property path. When a user destructures
+// like `const { ethers } = connectOnBefore(...)`, `ethers` becomes a proxy
+// that defers access until `resolved` is populated.
+function createNestedProxyForPath(getTarget: () => unknown): any {
+  return new Proxy(Object.create(null), {
+    get(_obj, prop) {
```
alcuadrado, line 152: I feel like this should be recursive somehow. Otherwise it won't work for things like:

```
const {networkHelpers:{loadFixture}} = connectConBefore(...)
```

Maybe I'm missing something

alcuadrado: We can decide not to make it recursive, and throw an clear error if someone tries to use it though.

v-next/hardhat-mocha/test/fixture-projects/connect-on-before/test/connect-on-before-with-explicit-hre.ts
```diff
@@ -0,0 +1,34 @@
+import assert from "node:assert/strict";
```
alcuadrado, line 1: There's a t missing in the name of this file. expliciT

alcuadrado (commented):
It looks good, but i'm slightly confused by the proxy logic. Maybe let's do a quick sync call about it?

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
kanej: So my reading of this is that `await` checks if `then` exists and is a function; we don't want the proxy being falsely treated as a Promise.

Looking again, I think at the top level (so network connection and not subproxies) we want to just throw to say this is a mistake:

```ts
describe("my test", async () => {
  const connect = await network.connectOnBefore();
```
Specically `connectOnBefore` assumes it is within a `describe` block, but in Mocha `describe` blocks can't be async.

I have added a test suite and new Hardhat error here: https://github.com/NomicFoundation/hardhat/pull/7973/changes/7208967f3df0f25113e40489be82e0f2691520bc

The question is what to do about the `then` check in nested proxies (so the equivalent line in `createNestedProxyForPath`?

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
kanej: This is covering the initial destructure case immeditaly after the call to `connectOnBefore` i.e.:

```ts
  const { ethers, provider } = connectOnBefore();
```

The underlying connection is `undefined` before the `before`, but we can still create subproxies. The returned `undefined` here is signaling to the subproxies that the connection is unresolved.

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
kanej: Agreed, I have updated `has`, `ownKeys` and `getOwnPropertyDescriptor` at both the top level and nested proxy level.

See https://github.com/NomicFoundation/hardhat/pull/7973/changes/5f0240539dbccadf995a9a17a8f104287a06ee13

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
kanej: We have decided to leave nested `then` as returning undefined.

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
kanej: I have updated the proxy generation to support arbitrary nested proxies through recursion.

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,322 @@
...
+    getOwnPropertyDescriptor(_obj, prop) {
+      const target = getTarget();
+
+      if (target === null || target === undefined) {
+        throw new HardhatError(
+          HardhatError.ERRORS.HARDHAT_MOCHA.CONNECT_ON_BEFORE.USE_BEFORE_HOOK,
+        );
+      }
+
+      return Reflect.getOwnPropertyDescriptor(target, prop);
```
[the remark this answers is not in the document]
kanej: The successful build suggests this is not needed.

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,320 @@
...
+
+        if (t === null || t === undefined) {
+          return undefined;
+        }
+
+        return Reflect.get(t, prop);
+      }
+
+      // `then` must return undefined so the proxy is not mistaken for a
+      // thenable (e.g. when returned from an async context or awaited).
```
alcuadrado:
```suggestion
      // thenable (e.g. when returned from an async context or awaited).
      // Note: This prevents promises as properties of the network connection
      // or its transitive props, but that's ok, as floating promises like that are
      // dangerous to begin with
```

alcuadrado:
I reviewed the PR, with a deeper focus on the proxy. I tried to break it and simplify it as much as possible, and added 5 commits:

1. [f6e7088](https://github.com/NomicFoundation/hardhat/pull/7973/commits/f6e708803ea6ccb7dd209a3f9132b4fa48b26ae2) Normalizes as much as possible how `toString()` and `util.inspect` work. When this wasn't possible, adding tests to act as witnesses as some slightly unexpected behavior.
2. [a086e67](https://github.com/NomicFoundation/hardhat/pull/7973/commits/a086e67a8357f45eb392d635e4438b8b634f3d61) Removes the check that prevented `then` as a nested property, as this one wasn't needed.
3. [db82de3](https://github.com/NomicFoundation/hardhat/pull/7973/commits/db82de3facdb0b247767d5c1253b243e595db2cf) Is a failed attempt to not bind the functions. I thought this was possible, but the tests weren't covering `#private` fields.
4. [36cdb82](https://github.com/NomicFoundation/hardhat/pull/7973/commits/36cdb82ad42eafab33ab4b68a6fe9f4540c5df5b) Corrects the previous commits and improves the tests. Also adding a test acting as a witness of a known limitation (i.e. functions are unexpectedly bound, so you can't change their `this`)
5. [38a22ea](https://github.com/NomicFoundation/hardhat/pull/7973/commits/38a22ea4d09bcb8577ca101f6e90f31fcc6a6599) Binds the functions in the first recursive proxy creation step (i.e. when getting a top level function from an unresolved proxy)

alcuadrado:
> There is no need for a peer-bump. `hardhat-errors` is a direct dependency, and the Mocha chai matchers add only tests, there is no change to a used API in `hardhat-mocha`.

Agreed

alcuadrado (approved):

kanej:
The final steps before merging this will be a run through on the OpenZeppelin test suite and an update to the mocha testing guide in the docs.

v-next/hardhat-mocha/test/fixture-projects/connect-on-before/test/connect-on-before-network-helpers-usage.ts
```diff
@@ -0,0 +1,50 @@
+import assert from "node:assert/strict";
+import { numberToHexString } from "@nomicfoundation/hardhat-utils/hex";
+import { network } from "hardhat";
+
+describe("connectOnBefore network helpers usage", async function () {
```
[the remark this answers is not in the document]
kanej: Agreed - applied.

v-next/hardhat-ethers-chai-matchers/test/fixture-projects/connect-on-before/test/connect-on-before-ethers-chai-matchers-usage.ts
```diff
@@ -0,0 +1,18 @@
+import { expect } from "chai";
+import { network } from "hardhat";
+import type { NetworkConnection } from "hardhat/types/network";
+
+async function deployCounter(connection: NetworkConnection) {
+  const counter = await connection.ethers.deployContract("Counter");
+  return { counter };
+}
+
+describe("connectOnBefore ethers chai matchers usage", async function () {
```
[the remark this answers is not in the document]
kanej: Agreed - applied.

v-next/hardhat-toolbox-mocha-ethers/test/fixture-projects/toolbox/package.json
```diff
@@ -3,8 +3,8 @@
   "private": "true",
   "type": "module",
   "dependencies": {
-    "@types/chai": "*",
-    "chai": "*",
-    "hardhat": "*"
+    "@types/chai": "^4.2.0",
+    "chai": "^5.1.2",
+    "hardhat": "workspace:*"
```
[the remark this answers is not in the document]
kanej: Agreed - applied.

v-next/hardhat-mocha/src/connect-on-before/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,322 @@
...
+
+        // We need to bind functions to the target object on get, because
+        // otherwise the Proxy will be used as `this` (e.g. proxy.method()) and
+        // won't be able to access #private properties
+        if (typeof value === "function") {
+          return value.bind(target);
+        }
+
+        return value;
+      }
```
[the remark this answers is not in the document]
kanej: We have tested this code and are happy with the current level of resilience.

v-next/hardhat-mocha/test/create-network-connection-proxy.ts
```diff
@@ -0,0 +1,906 @@
...
+          },
+        });
+
+        const proxy = createNetworkConnectionProxy(() => resolved);
+
+        const closeFn = proxy.close;
+
+        await closeFn();
+        assert.equal(idDuringClose, 99);
+      });
```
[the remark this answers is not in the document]
kanej: I have changed the test name to:

```
should preserve 'this' context when functions are extracted from the proxy
```

kanej:
We are taking a different approach and enhancing the core API.

PR #7974, opened 2026-02-17
Convert `action` to `lazyAction`

ChristopherDedominici:
This PR is created on top of this one: https://github.com/NomicFoundation/hardhat/pull/7950

In this PR, the **ONLY** changes are the renames from `action` to `lazyAction`.
These affects:
- src code
- tests
- comments
- types

### DIFF
--- a/v-next/example-project/hardhat.config.ts
+++ b/v-next/example-project/hardhat.config.ts
@@ -24,7 +24,7 @@ const exampleEmptyTask = emptyTask("empty", "An example empty task").build();
 
 const exampleEmptySubtask = task(["empty", "task"])
   .setDescription("An example empty subtask task")
-  .setAction(async () => ({
+  .setLazyAction(async () => ({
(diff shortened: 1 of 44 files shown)


v-next/example-project/hardhat.config.ts
```diff
@@ -24,15 +24,15 @@ const exampleEmptyTask = emptyTask("empty", "An example empty task").build();
 
 const exampleEmptySubtask = task(["empty", "task"])
   .setDescription("An example empty subtask task")
-  .setAction(async () => ({
+  .setLazyAction(async () => ({
```
alcuadrado, line 27: oh, this was confusing on my end, sorry. I think from the POV of a user we should change it. I was referring only to the field in the builder and definitions.

alcuadrado: Because this does look like a breaking change without a real benefit otherwise

PR #7975, opened 2026-02-17
Round average gas usage in `--gas-stats`

alcuadrado:
@k06a reported that `--gas-stats` was displaying two decimals of precision, which is meaningless in this context. This PR rounds them instead.

### DIFF
--- a/.changeset/mighty-lilies-kiss.md
+++ b/.changeset/mighty-lilies-kiss.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Round average and mean gas usage in the gas analytics output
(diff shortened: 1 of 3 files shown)


k06a:
I would assume the median doesn’t require any rounding since it’s typically a number from the set.

alcuadrado:
> I would assume the median doesn’t require any rounding since it’s typically a number from the set.

Good point! 😅

alcuadrado:
TIL: Median of an even-sized array is the average of the two middle elements

k06a:
@alcuadrado good point

schaable (approved):

PR #7978, opened 2026-02-19
Install `libusb-1.0-0-dev` on CI

alcuadrado:
This change addresses this kind of sporadic CI failure: https://github.com/NomicFoundation/hardhat/actions/runs/22157700985/job/64073155843

The reason for this is that `node-hid` downloads some prebuilt binaries, which can fail. When it does fail, it falls back to building from source, and in that case, we need `libusb-1.0-0-dev` to be installed.

This is only relevant to Linux, as windows and mac use other usb backends.


kanej (approved):
We should look at using prebaked containers or other approaches at some point.

PR #7979, opened 2026-02-19
Introduce new inter-process mutex implementation (lockfile version)

alcuadrado:
This is an alternative to https://github.com/NomicFoundation/hardhat/pull/7942/changes with a simpler approach, using a lockfile instead of a lock directory. This doesn't work well on network filesystems, but it's much simpler, and probably enough for our needs.


v-next/hardhat-utils/src/synchronization.ts
```diff
@@ -1,164 +1,547 @@
...
+export {
+  IncompatibleHostnameMultiProcessMutexError,
+  IncompatibleMultiProcessMutexError,
+  IncompatiblePlatformMultiProcessMutexError,
+  IncompatibleUidMultiProcessMutexError,
+  InvalidMultiProcessMutexPathError,
+  MultiProcessMutexError,
+  MultiProcessMutexTimeoutError,
+  StaleMultiProcessMutexError,
+} from "./errors/synchronization.js";
```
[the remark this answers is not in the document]
alcuadrado: spurious comment

.changeset/slick-donuts-carry.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Make the solc downloader safe when run by multiple processes ([7946](https://github.com/NomicFoundation/hardhat/pull/7946)).
```
ChristopherDedominici, line 5: A changelog for this PR was already present, so I am just making sure this duplication is expected

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compiler/downloader.ts
```diff
@@ -152,19 +152,18 @@ export class CompilerDownloaderImplementation implements CompilerDownloader {
...
-   */
   constructor(
     platform: CompilerPlatform,
     compilersDir: string,
     downloadFunction: typeof download = download,
   ) {
     this.#platform = platform;
     this.#compilersDir = compilersDir;
+    this.#mutexCompilerList = new MultiProcessMutex(
+      path.join(compilersDir, "compiler-download-list"),
```
ChristopherDedominici: Previously, the mutex was stored in a temporary folder using `os.tmpdir()`, but that is no longer the case. I'm wondering if we are ok in adding lock files directly to the same folder as the compilers. These lock files are removed at the end, but orphans could occur, (e.g.: after a crash).

Do users check this folder manually? If so, it might be worth putting all locks inside a ".locks" folder (`path.join(compilersDir, ".locks", "compiler-download-list`), or maybe using a temporary folder in this case as well?

v-next/hardhat-utils/test/synchronization.ts
```diff
@@ -164,6 +230,932 @@ describe("multi-process-mutex", () => {
...
+      !fs.existsSync(lockPath),
+      "Lock file should be removed after release",
+    );
+  });
+
+  it("should recover a stale lock with a dead PID", async () => {
+    const lockPath = path.join(getTmpDir(), "stale.lock");
+    // Create a fake lock file with metadata pointing to a non-existent PID
+    writeFakeMetadata(lockPath, {
+      pid: 999999999, // Very unlikely to exist
```
ChristopherDedominici, line 256: Could, even if very unlikely, be flaky. What about creating a function to get a PID for a test?
Example (pseudocode)
```
uid = ( Number ( process.getuid?.() ) || 0 ) + 1;
```

ChristopherDedominici:
The logic looks solid 😎, I just left 2 minor comments

v-next/hardhat-utils/test/synchronization.ts
alcuadrado: I think any option has some risk from having a collision. This one is very unlikely (because the number is huge), and it will be evident due to the comment if it ever happen

.changeset/slick-donuts-carry.md
alcuadrado: Yes, I added one changeset per "change" I made in this PR, so that the changelog looks better

ChristopherDedominici (approved):

PR #7982, opened 2026-02-19
Use latest supported solc in coverage instrumentation

alcuadrado:
This PR fixes an integration issue between Hardhat and EDR. Previously, when a compilation job tried to use a version of solc that wasn't fully supported by EDR, the plugin would try to change the solcVersion of the compilation job, instead of just limiting itself to use a different version for instrumentation. Changing the version in a hook is forbidden and throws.

In order to test this, I moved the instrumentation logic to its own module, and test it directly.

Note: This PR also removes a `console.log` that was printed ever time the version was changed.

### DIFF
--- a/.changeset/angry-mugs-bow.md
+++ b/.changeset/angry-mugs-bow.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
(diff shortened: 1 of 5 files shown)


schaable (approved):
LGTM

PR #7988, opened 2026-02-22
Improve compiler selection errors

### LINKED ISSUE #7748: Improve error message when no solc version is compatible with a file

alcuadrado:
Fixes https://github.com/NomicFoundation/hardhat/issues/7748 by:

- Running a better analysis of the error situation and presenting the users with more error cases
- Using file system paths instead of `inputSourceName`s to display the files

It also makes working with results of `SolidityBuildSystem#build` and `SolidityBuildSystem#getCompilationJobs` simpler. /cc @kanej no more `"reason" in ...`

Replaces https://github.com/NomicFoundation/hardhat/issues/7748

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -638,7 +638,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     }
 
     return {
-      success: true as const,
+      success: true,
       compilationJobsPerFile,
       indexedIndividualJobs,
       cacheHits,
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -430,7 +430,6 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     const solcConfigSelector = new SolcConfigSelector(
       buildProfileName,
       buildProfile,
-      dependencyGraph,
     );
 
     let subgraphsWithConfig: Array<
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -244,7 +244,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       options,
     );
 
-    if ("reason" in compilationJobsResult) {
+    if (!compilationJobsResult.success) {
       return compilationJobsResult;
     }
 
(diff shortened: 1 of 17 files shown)


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
```diff
@@ -179,7 +276,7 @@ export class SolcConfigSelector {
         continue;
       }
 
-      visited.add(file);
+      visited = new Set([...visited, file]);
```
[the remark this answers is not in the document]
alcuadrado: That behavior is different. And efficiency doesn't matter as much here, because it's the unhappy case, and we are about to fail the compilation.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
ChristopherDedominici: The docs needs to be updated according to the changes

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -631,7 +636,12 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       }
     }
 
-    return { compilationJobsPerFile, indexedIndividualJobs, cacheHits };
+    return {
+      success: true as const,
```
ChristopherDedominici: minor: I think `as const` can be removed

ChristopherDedominici (approved):
LGTM, only 2 minor comments

PR #7989, opened 2026-02-22
Node network config

alcuadrado:
This PR introduces a new `node` network config, which is present by default, and whose type must always be `edr-simulated`. This last requirement is enforced during the validation of the config.

Depends on: https://github.com/NomicFoundation/hardhat/pull/7805
Docs PR: https://github.com/NomicFoundation/hardhat-website/pull/222

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/node/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/node/task-action.ts
@@ -12,7 +12,6 @@ import { ensureDir, exists } from "@nomicfoundation/hardhat-utils/fs";
 import chalk from "chalk";
 import debug from "debug";
 
-import { DEFAULT_NETWORK_NAME } from "../../constants.js";
 import { isSupportedChainType } from "../../edr/chain-type.js";
 import { BUILD_INFO_DIR_NAME } from "../artifacts/artifact-manager.js";
 import { EdrProvider } from "../network-manager/edr/edr-provider.js";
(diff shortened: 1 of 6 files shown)


v-next/hardhat/src/internal/builtin-plugins/node/task-action.ts
```diff
@@ -39,31 +38,19 @@ const nodeAction: NewTaskActionFunction<NodeActionArguments> = async (
...
+    "`node` network type must be `edr-simulated`",
+  );
 
   const connectionParams: {
     network: string;
     chainType?: ChainType;
     override?: EdrNetworkConfigOverride;
   } = {
-    network,
+    network: "node",
```
schaable: Does this mean we can’t run `npx hardhat node --network myNodeNetwork` anymore?

v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -154,6 +154,65 @@ describe("network-manager/hook-handlers/config", () => {
...
+
+    it("should extend the user config with the node network", async () => {
+      const config: HardhatUserConfig = {};
+      const next = async (nextConfig: HardhatUserConfig) => nextConfig;
+
+      const extendedConfig = await extendUserConfig(config, next);
+      assert.ok(
+        extendedConfig.networks?.node !== undefined,
+        "node network should be defined",
+      );
```
schaable: Minor: this assertion is redundant, since the next one would fail if the node network is `undefined`. I'm fine if you want to keep it, as it makes the intent more explicit.

v-next/hardhat/test/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -217,6 +276,28 @@ describe("network-manager/hook-handlers/config", () => {
...
+          node: {
+            type: "http",
+            url: "http://localhost:8545",
+          },
+        },
+      };
+
+      /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      -- testing invalid network type for js users */
+      const validationErrors = await validateNetworkUserConfig(config as any);
```
schaable: is the type assertion needed?

schaable (approved):
Looks good. Left a question and some minor comments on the tests.

v-next/hardhat/src/internal/builtin-plugins/node/task-action.ts
alcuadrado: That's right. That's ignored. Maybe we should print warning or throw an error

alcuadrado: Or maybe use `node` just by default. Is that possible?

alcuadrado: It wouldn't be possible, because `--network` has a default value of `"default"`

alcuadrado: I guess if users want to be able to override we can add a task option `--network-config` or something like that in the future.

schaable: can't you do?

```
const network =
    hre.globalOptions.network !== undefined
      ? hre.globalOptions.network
      : hre.config.network.node;
```

alcuadrado: You are right!

schaable (approved):

kanej:

PR #7991, opened 2026-02-23
Hardhat 2 plugins deprecation message on load

### LINKED ISSUE #7960: Mock HH2 plugin APIs with better error messages

alcuadrado:
Fixes #7960 

This is done with three kinds of changes:

* Mock the HH2 plugin API and throw a proper error when used.
* For common subpaths that Hardhat 2 plugins imported I
  * Added them if make sense
  * Added an empty module for the ones that don't make sense. The reason an empty module works is that HH2 plugins are CJS, so they lead to a load-time error if a value isn't exported. 

This PR can be reviewed one commit at the time.


v-next/hardhat/src/internal/using-hardhat2-plugin-errors.ts
```diff
@@ -0,0 +1,94 @@
...
+export class UsingHardhat2PluginError extends CustomError {
+  public readonly callerRelativePath: string | undefined;
+  constructor() {
+    const callerPath = getCallerRelativePath();
+
+    let message: string;
+    if (callerPath !== undefined) {
+      message = `You are trying to use a Hardhat 2 plugin in a Hardhat 3 project.
+
+This file is part of the Hardhat 2 plugin calling an API that was removed in Hardhat 3: ${chalk.bold(callerPath)}
```
schaable:
```suggestion
This file is part of a Hardhat 2 plugin calling an API that was removed in Hardhat 3: ${chalk.bold(callerPath)}
```

v-next/hardhat/src/internal/using-hardhat2-plugin-errors.ts
```diff
@@ -0,0 +1,94 @@
...
+
+Please read https://hardhat.org/migrate-from-hardhat2 to learn how to migrate your project to Hardhat 3.
+`;
+    } else {
+      message = `You are trying to use a Hardhat 2 plugin in a Hardhat 3 project.
+      
+Take a look at this stack trace to find the which plugin it is.
```
schaable:
```suggestion
Check the stack trace below to identify which plugin is causing this.
```

v-next/hardhat/src/internal/using-hardhat2-plugin-errors.ts
```diff
@@ -0,0 +1,94 @@
...
+Please read https://hardhat.org/migrate-from-hardhat2 to learn how to migrate your project to Hardhat 3.
+`;
+    }
+
+    super(message);
+    this.callerRelativePath = callerPath;
+  }
+}
+
+export function getCallerRelativePath(depth: number = 5): string | undefined {
```
schaable, line 61:
```suggestion
/**
 * Returns the relative path of the file that called a deprecated Hardhat 
 * plugin API, based on the stack trace. This helps identify which plugin 
 * file is triggering usage of Hardhat 2 APIs in a Hardhat 3 project.
 *
 * @param {number} [depth=5] The stack trace depth to locate the caller's 
 * source file. By default, depth 5 is used because:
 *   0 = message
 *   1 = getCallerRelativePath
 *   2 = UsingHardhat2PluginError constructor
 *   3 = throwUsingHardhat2PluginError
 *   4 = deprecated function
 *   5 = actual caller (the plugin file)
 *
 * @returns {string|undefined} The shortened relative path of the caller file, 
 * or undefined if not found.
 *
 * @example
 * If the stack trace is:
 * // Error
 * //     at getCallerRelativePath (src/internal/using-hardhat2-plugin-errors.ts:34:15)
 * //     at UsingHardhat2PluginError.constructor (src/internal/using-hardhat2-plugin-errors.ts:7:3)
 * //     at throwUsingHardhat2PluginError (src/internal/using-hardhat2-plugin-errors.ts:90:3)
 * //     at deprecatedFunction (plugins/example-plugin/deprecated.js:50:10)
 * //     at main (plugins/example-plugin/index.js:100:5)
 * Calling getCallerRelativePath() returns 'plugins/example-plugin/index.js'
 */
export function getCallerRelativePath(depth: number = 5): string | undefined {
```

I'd add this jsdoc and remove the comment on line 41. The example above is completely made up 😂 It would be nice to use a real example, but when I tried importing a Hardhat 2 plugin in the example project, the stack trace looked different.

v-next/hardhat/src/internal/using-hardhat2-plugin-errors.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    // Default depth=5 assumes: 0=message, 1=getCallerRelativePath, 2=Hardhat2PluginError constructor, 3=throwHardhat2PluginError, 4=deprecated fn, 5=actual caller
+    const lines = stack.split("\n");
+    const callerLine = lines[depth];
+    if (callerLine === undefined) {
+      return undefined;
+    }
+
+    const regex = new RegExp(
```
schaable: minor, but I don't quite like the "part by part" explanation of this regex. I get that it's complex, but this approach doesn't really make it easier to understand for me. That said, it might just be me, so take this with a grain of salt. A higher-level comment explaining its purpose and what it captures might work better, since anyone updating it will still need to spend some time reasoning about the regex anyway.

```typescript
/**
 * Matches a single stack trace line:
 *
 * at FunctionName (path/to/file.ts:10:5)
 * at path/to/file.ts:10:5
 *
 * Captures:
 *  - group 1: file location (without line/column)
 */
const STACK_TRACE_LINE_REGEX =
  /^at (?:.+? \()?([^\(].*?)(?::\d+)?(?::\d+)?\)?$/;
```

v-next/hardhat/src/types/index.ts
```diff
@@ -0,0 +1,14 @@
+export * from "./arguments.js";
```
schaable, line 1: whoops, did we forget to add this file earlier?

v-next/hardhat/src/internal/deprecated-module-imported-from-hardhat2-plugin.ts
```diff
@@ -0,0 +1,8 @@
+// This is an empty module that is used to exported it with a subpath that's
```
schaable, line 1: 🤯 this is very smart!

v-next/hardhat/src/internal/cli/error-handler.ts
```diff
@@ -146,3 +152,15 @@ function getErrorMessages(error: Error): ErrorMessages {
...
+function printUsingHardhat2Error(
+  error: UsingHardhat2PluginError,
+  print: (message: string | Error) => void = console.error,
+): void {
+  print(chalk.red.bold(`Hardhat 3 installation error:`));
+  print("");
+  if (error.callerRelativePath !== undefined) {
+    print(error.message);
+  } else {
+    print(error.stack);
```
schaable, line 164: I'm a bit confused about this one. Aren't we also setting a message in `v-next/hardhat/src/internal/using-hardhat2-plugin-errors.ts` when `error.callerRelativePath === undefined`? Is that message getting lost?

alcuadrado: `.stack` always includes the message as `<ErrorType>: <Message>` as the first line of the stack

v-next/hardhat/src/internal/using-hardhat2-plugin-errors.ts
alcuadrado: Fantastic!

v-next/hardhat/src/internal/using-hardhat2-plugin-errors.ts
alcuadrado: Agreed

schaable (approved):

kanej:

PR #7993, opened 2026-02-24
Fix `hardhat-ethers-chai-matchers`

alcuadrado:
_This PR is a result of investigating https://github.com/NomicFoundation/hardhat/pull/7726 and replaces it_

This PR introduces to three fixes to `hardhat-ethers-chai-matchers`:

1. It remove all the usages of `assertHardhatInvariant` with `assert.fail` with descriptive messages. The `assertHardhatInvariant` function should only be used for things that should *always* be true. It should not be used to validate inputs or return values.
2. It preservers all the `error.cause`s
3. It removes `HardhatError`s from within assertions, as they play nicely with mocha.

It also introduces a small formatting improvement to a `hardhat-mocha` error message.

### DIFF
--- a/v-next/hardhat-ethers-chai-matchers/package.json
+++ b/v-next/hardhat-ethers-chai-matchers/package.json
@@ -63,7 +63,6 @@
     "typescript": "~5.8.0"
   },
   "dependencies": {
-    "@nomicfoundation/hardhat-errors": "workspace:^3.0.5",
     "@nomicfoundation/hardhat-utils": "workspace:^3.0.5",
     "@types/chai-as-promised": "^8.0.1",
     "chai-as-promised": "^8.0.0",
--- a/v-next/hardhat-ethers-chai-matchers/src/internal/matchers/emit.ts
+++ b/v-next/hardhat-ethers-chai-matchers/src/internal/matchers/emit.ts
@@ -131,7 +131,10 @@ export function supportEmit(
 
         return waitForPendingTransaction(tx, contract.runner.provider).then(
           (receipt) => {
-            assertIsNotNull(receipt, "receipt");
+            assertIsNotNull(
+              receipt,
+              "Transaction's receipt cannot be fetched from the network",
+            );
             return onSuccess(receipt);
           },
         );
@@ -183,7 +186,7 @@ const tryAssertArgsArraysEqual = (
     const parsedLog = chaiUtils
       .flag(context, "contract")
       .interface.parseLog(logs[0]);
-    assertIsNotNull(parsedLog, "parsedLog");
+    assertIsNotNull(parsedLog, "Can't parse the first log");
 
     return assertArgsArraysEqual(
       Assertion,
@@ -204,7 +207,7 @@ const tryAssertArgsArraysEqual = (
         const parsedLog = chaiUtils
           .flag(context, "contract")
           .interface.parseLog(logs[index]);
-        assertIsNotNull(parsedLog, "parsedLog");
+        assertIsNotNull(parsedLog, `Can't parse the log index ${index}`);
 
         assertArgsArraysEqual(
           Assertion,
--- a/v-next/hardhat-ethers-chai-matchers/src/internal/matchers/emit.ts
+++ b/v-next/hardhat-ethers-chai-matchers/src/internal/matchers/emit.ts
@@ -74,8 +74,9 @@ export function supportEmit(
         } catch (e) {
           if (e instanceof TypeError) {
             const errorMessage = e.message.split(" (argument=")[0];
-            // eslint-disable-next-line no-restricted-syntax -- keep the original chai error structure
-            throw new AssertionError(errorMessage);
+            const error = new AssertionError(errorMessage);
+            error.cause = e;
+            throw error;
           }
         }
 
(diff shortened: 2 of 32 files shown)


alcuadrado:
I'll fix the chai v5 workflow tomorrow.

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/emit.ts
```diff
@@ -131,7 +131,10 @@ export function supportEmit(
 
         return waitForPendingTransaction(tx, contract.runner.provider).then(
           (receipt) => {
-            assertIsNotNull(receipt, "receipt");
+            assertIsNotNull(
+              receipt,
+              "Transaction's receipt cannot be fetched from the network",
+            );
```
kanej, line 133: This makes sense if we assume chai matchers can never be used against a non-automining chain, are we happy with that assumption?

v-next/hardhat-ethers-chai-matchers/src/internal/utils/balance.ts
```diff
@@ -25,7 +27,22 @@ export async function getBalances(
...
+        try {
+          chaiAssert.fail(
+            "Failed to get the balance of the account " + address,
+          );
+        } catch (e) {
+          ensureError(e);
+          e.cause = cause;
+
+          throw e;
+        }
```
kanej, line 43: Well that is fun.

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/reverted/legacyReverted.ts
```diff
@@ -12,8 +12,8 @@ export function supportLegacyReverted(
       this._obj.catch(() => {});
     }
 
-    throw new HardhatError(
-      HardhatError.ERRORS.CHAI_MATCHERS.GENERAL.DEPRECATED_REVERTED_MATCHER,
+    chaiAssert.fail(
+      "The .reverted matcher has been deprecated. Use .revert(ethers) instead.",
     );
   });
```
kanej, line 18: Does this imply any cleanup of errors like `DEPRECATED_REVERTED_MATCHER`?

v-next/hardhat-ethers-chai-matchers/package.json
```diff
@@ -63,7 +63,6 @@
     "typescript": "~5.8.0"
   },
   "dependencies": {
-    "@nomicfoundation/hardhat-errors": "workspace:^3.0.5",
```
kanej, line 66: We should update `pnpm-lock.json`

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/emit.ts
alcuadrado: I didn't change that assumption here. It was always present. I guess we can create an issue to revisit it if needed. @ChristopherDedominici 

Note that these matchers may assume a few extra things that are only present in dev networks.

v-next/hardhat-ethers-chai-matchers/src/internal/utils/balance.ts
alcuadrado: Bringing back the future to Chai

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/reverted/legacyReverted.ts
alcuadrado: Not really, because otherwise the existing versions will print links that 404

v-next/hardhat-ethers-chai-matchers/package.json
alcuadrado: it's updated, because otherwise all the ci jobs fail at `pnpm install --frozen-lockfile --prefer-offline`

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/emit.ts
ChristopherDedominici: I created this PR (still in draft), to address the issue: https://github.com/NomicFoundation/hardhat/pull/7997

kanej: @ChristopherDedominici how does this work in Hardhat 2, did chai matchers always assume you were running against an automine network?

alcuadrado: Yes, in HH2 it works like that. Much of this code was inherited from Waffle, which worked only with Ganache.

alcuadrado:
Fixed the windows CI

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/emit.ts
alcuadrado: Let's not block this PR on the other one though

kanej (approved):
Leaving the question of supporting non-automine situations aside, this looks good.

Can I check that the trick to improving the missing `await` error message is that chai has some inbuilt support for this, and that by changing across to chai's assertions we enable that?

alcuadrado:
> Can I check that the trick to improving the missing await error message is that chai has some inbuilt support for this, and that by changing across to chai's assertions we enable that?

`hardhat-mocha` adds an `unhandledRejection` listener that prints that if the unhandled rejection's error is an `AssertionError`.

PR #8002, opened 2026-02-24
Deprecate the `latest` tag in all the packages that aren't compatible with Hardhat 3

### LINKED ISSUE #7961: Replace deprecated HH2 plugins with stubs that redirect to new versions/migration documentation

alcuadrado:
This PR replaces all the packages that are not compatible with Hardhat 3 with new major versions of them that intentionally kill Hardhat.

When imported, they print a message explaining that the version that the user installed is not compatible with Hardhat 2 nor 3, and redirects them to the migration guide or to install the `hh2` version.

The reason for this change is that if a user installs Hardhat 3 and a Hardhat 2 package using `latest` for both (e.g. `npm install hardhat @nomicfoundation/hardhat-toolbox`), they can have a really hard time figuring out how to get out of there.

---

Reviewing the diff of this PR doesn't add any value. The end-state is way smaller than the diff, and simpler enough to review just that.

Anyways, as this is a huge PR, I'll explain how I built it: 

* First I removed everything what was obviously not needed. Packages, empty dirs, e2e tests, config files, github workflows, etc
* Updated the root package.json to only keep changesets, and updated the workspace settings
* Manually update `hardhat-chai-matchers`
* Used Claude to replicate the change in the other packages
* Updated the release workflow to use `deprecated-versions` as base branch and `latest` as npm tag
* Updated the root README.md
* Fixed some formatting across all the packages.
* Updated the changesets config
* Added a changeset to release a new major of all the remaining packages
* Made sure that `pnpm version-for-release` still works
* Deleted all the `scripts/` that weren't needed to run the changeset workflow

Finally, I also manually validated that all of these packages have a `hh2` tag published on npm.


schaable (approved):
Reviewed the branch and looks good.

PR #8003, opened 2026-02-25
Upgrade to @nomicfoundation/edr v0.12.0-next.27

ChristopherDedominici:
Upgraded EDR to `v0.12.0-next.27`. 
It jumps from `v .24` to `v .27`.

This PR contains more changes than a standard update, as EDR .25 introduces breaking changes that require updates to the Hardhat logic.


v-next/hardhat/test/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -110,7 +110,7 @@ describe("edr-provider", () => {
       assert.deepEqual(response, {
         failed: false,
         gas: 21000,
-        returnValue: "",
+        returnValue: "0x",
```
ChristopherDedominici, line 113: Before hardhat used edrRpcDebugTraceToHardhat that was converting to "", now we use EDR response directly

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/convert-to-edr.ts
```diff
@@ -274,67 +271,6 @@ export function hardhatMempoolOrderToEdrMineOrdering(
   }
 }
 
-// TODO: EDR should handle this conversion. This is a temporary solution.
```
ChristopherDedominici, line 277: as mentioned in the comment, this function is no longer needed as EDR has now functionalities to replace it

v-next/hardhat/test/internal/builtin-plugins/network-manager/edr/edr-provider.ts
alcuadrado: But this is a breaking change then, isn't it?

alcuadrado: in EDR, I mean

alcuadrado: No, this updated behavior is correct. See this example from quicknode's docs:

```sh
curl https://docs-demo.quiknode.pro/ \ 
-X POST \
-H "Content-Type: application/json" \
--data '{"method":"debug_traceTransaction","params":["0x9e63085271890a141297039b3b711913699f1ee4db1acb667ad7ce304772036b"], "id":1,"jsonrpc":"2.0"}' | jq | tail
```

```     "0000000000000000000500000000000000000000000000000000000000000000",
          "0000000000000000000100000000000000000000000000000000000000000000"
        ]
      }
    ],
    "gas": 107817,
    "failed": false,
    "returnValue": "0x"
  }
}
```

.changeset/calm-eels-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Bumped EDR version to [`0.12.0-next.25`](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.25).
```
alcuadrado, line 5:
```suggestion
Bumped EDR version to [`0.12.0-next.25`](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.25).

BREAKING CHANGE: Memory capture used to be enabled by default on geth, but has since been flipped https://github.com/ethereum/go-ethereum/pull/23558 and is now disabled by default. We have followed suit and disabled it by default as well. If you were relying on memory capture, you will need to explicitly enable it by setting the enableMemory option to true in your tracer configuration.
```

alcuadrado: Let's include that message from EDR

alcuadrado (approved):

PR #8006, opened 2026-02-25
Improve `hardhat-viem` and `hardhat-ethers` autocompletion and type-level errors

alcuadrado:
_This PR is the result of investigating this issue https://github.com/NomicFoundation/hardhat-vscode/issues/714_

Currently `hardhat-viem` and `hardhat-ethers` use functions like this:

```ts
function doSomething<ContractNameT extends string>(
   contractName: ContractNameT,
   // ...
): DoSomethingReturnType<ContractNameT> {
  // ...
}
```

Which has the result of `doSomething` accepting any string, and offering no autocompletion.

This PR changes that so that they use a new `ArtifactContractNames`, which has the actual types of the **compiled** artifacts:

```ts
function doSomething<ContractNameT extends ArtifactContractNames>(
   contractName: ContractNameT,
   // ...
): DoSomethingReturnType<ContractNameT> {
  // ...
}
```

This leads to better autocomplete (it suggest the real contract names, including failing if there are multiple contracts with the same name), and generates a typescript error if the contract doesn't exist.

The only drawback is that if you add a contract, you will get a type error until you run `hardhat build`. Note that this happens in other places too though, like when using a contract abstraction instance. For example, this happens in the `hardhat-viem` sample project.

## Missing work

* This same change hasn't been implemented in the ignition packages, because they don't depend on `hardhat` (only as dev dependency, so they don't have the type. I think an optional peer dependency or something like that can be used. Maybe it can work without any modification to the `package.json`, but that sounds more risky.
* There's a test that fails, but i think it's because of how it was built, not because it should fail.

### DIFF
--- a/.peer-bumps.json
+++ b/.peer-bumps.json
@@ -26,6 +26,16 @@
       "package": "@nomicfoundation/hardhat-typechain",
       "peer": "hardhat",
       "reason": "Uses the new api to detect errors in SolidityBuildSystem#build"
+    },
+    {
+      "package": "@nomicfoundation/hardhat-viem",
+      "peer": "hardhat",
+      "reason": "It now uses ArtifactContractNames"
+    },
+    {
+      "package": "@nomicfoundation/hardhat-viem",
+      "peer": "hardhat",
+      "reason": "It now uses ArtifactContractNames"
     }
   ]
 }
--- a/.peer-bumps.json
+++ b/.peer-bumps.json
@@ -30,12 +30,12 @@
     {
       "package": "@nomicfoundation/hardhat-viem",
       "peer": "hardhat",
-      "reason": "It now uses ArtifactContractNames"
+      "reason": "It now uses StringWithArtifactContractNamesAutocompletion"
     },
     {
-      "package": "@nomicfoundation/hardhat-viem",
+      "package": "@nomicfoundation/hardhat-ethers",
       "peer": "hardhat",
-      "reason": "It now uses ArtifactContractNames"
+      "reason": "It now uses StringWithArtifactContractNamesAutocompletion"
     }
   ]
 }
--- a/v-next/example-project/test/node/example-test.ts
+++ b/v-next/example-project/test/node/example-test.ts
@@ -21,6 +21,8 @@ describe("Example EDR based test", () => {
     });
 
     assert.equal(blockNumberAfter, `0x5`);
+
+    await connection.viem.deployContract("Counter2");
   });
 
   it("should show stack traces when a transaction reverts", async () => {
--- a/v-next/example-project/test/node/example-test.ts
+++ b/v-next/example-project/test/node/example-test.ts
@@ -21,8 +21,6 @@ describe("Example EDR based test", () => {
     });
 
     assert.equal(blockNumberAfter, `0x5`);
-
-    await connection.viem.deployContract("Counter2");
   });
 
   it("should show stack traces when a transaction reverts", async () => {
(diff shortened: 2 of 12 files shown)


alcuadrado:
To show a concrete example of this working in practice, here the autocomplete of vscode shows me all the names of the contracts in the example project:

image

The names are extracted from the compilation artifacts, so if you write a contract name that doesn't exist, or that you haven't built yet, you get this error:

image

Unfortunately, typescript doesn't provide ways to return custom/friendlier error messages.

alcuadrado:
TBD if this should also be applied to `ArtifactManager#readArtifact`, which is a lower-level API, but also uses the `ArtifactMap`.

v-next/hardhat-ethers/src/internal/hardhat-helpers/hardhat-helpers.ts
```diff
@@ -164,7 +169,7 @@ export class HardhatHelpers {
   }
 
   public async getContractAt(
-    nameOrAbi: string | Abi,
+    nameOrAbi: ArtifactContractNames | Abi,
```
olehmisar:
```suggestion
    nameOrAbi: (ArtifactContractNames | (string & {})) | Abi,
```

Have you tried this? It should still provide autocomplete for compiled contracts but not throw an error for arbitrary strings

alcuadrado: Wow, nice trick! 🤯

alcuadrado: I pushed a change that allows that. I guess the question is what's the better guarantee that we should offer:
* strict errors, leading to errors before `hardhat build`, but no runtime errors.
* autocomplete, no error before `hardhat build`, potentially runtime errors.

.peer-bumps.json
```diff
@@ -6,5 +6,16 @@
...
+    {
+      "package": "@nomicfoundation/hardhat-viem",
+      "peer": "hardhat",
+      "reason": "It now uses ArtifactContractNames"
+    },
+    {
+      "package": "@nomicfoundation/hardhat-viem",
+      "peer": "hardhat",
+      "reason": "It now uses ArtifactContractNames"
+    }
```
schaable:
```suggestion
    }
```
This is duplicated

v-next/hardhat/src/types/artifacts.ts
```diff
@@ -17,6 +18,26 @@ export type GetArtifactByName<ContractNameT extends string> =
     ? ArtifactMap[ContractNameT]
     : Artifact;
 
+/**
+ * A type type that represents the contract names (bare and fully qualified) of
```
schaable:
```suggestion
 * A type that represents the contract names (bare and fully qualified) of
```

v-next/example-project/test/node/example-test.ts
```diff
@@ -21,6 +21,8 @@ describe("Example EDR based test", () => {
     });
 
     assert.equal(blockNumberAfter, `0x5`);
+
+    await connection.viem.deployContract("Counter2");
```
schaable: is this needed?

.peer-bumps.json
```diff
@@ -6,5 +6,16 @@
     "v-next/hardhat/templates",
     "v-next/config"
   ],
-  "bumps": []
+  "bumps": [
+    {
+      "package": "@nomicfoundation/hardhat-viem",
```
schaable, line 31:
```suggestion
      "package": "@nomicfoundation/hardhat-ethers",
```

schaable (approved):
Left a few comments, but it looks good. I tested it in the example project using `deployContract` with duplicate contract names, and also verified that it doesn't show errors before building or when passing a string that isn't a contract name.

v-next/example-project/test/node/example-test.ts
alcuadrado: I removed it

PR #8007, opened 2026-02-25
Accept relative paths to node_modules in npm remappings

### LINKED ISSUE #7562: Support relative paths to `node_modules` in the `reamppings.txt` files
## Context

When the target of a reampping starts with `node_modules/` we treat it as a special case, that will be resolved using the npm resolution rules. This is done here: https://github.com/NomicFoundation/hardhat/blob/f29d1d804e00e2b13666f040fb85d7b7b71a58ce/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts#L502

Some projects have relative paths to those `node_modules` though, like `../../node_modules`. See: https://github.com/NomicFoundation/hardhat/issues/6994

## Proposal

We should also support those, as they would still work using npm resolution rules.

Note that we shouldn't support any `node_modules`, but only the ones that are in the root of the project or a parent dir. For example:

- Accept: `./node_modules`, `../node_modules`, `../../node_modules`, `../../a/../node_modules`
- Ignore: `./something/node_modules` — This `node_modules` WON'T be found by the npm resolution rules.

alcuadrado:
Fixes https://github.com/NomicFoundation/hardhat/issues/7562, a requirement to drop the beta label

### DIFF
--- a/.changeset/polite-beds-doubt.md
+++ b/.changeset/polite-beds-doubt.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Accept relative paths to node_modules in npm remappings [#8007](https://github.com/NomicFoundation/hardhat/pull/8007)
--- a/.changeset/polite-beds-doubt.md
+++ b/.changeset/polite-beds-doubt.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Accept relative paths to node_modules in npm remappings [#8007](https://github.com/NomicFoundation/hardhat/pull/8007)
+Fixed the acceptance of relative paths to `node_modules` in npm remappings ([#8007](https://github.com/NomicFoundation/hardhat/pull/8007))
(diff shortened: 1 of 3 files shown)


.changeset/polite-beds-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Accept relative paths to node_modules in npm remappings [#8007](https://github.com/NomicFoundation/hardhat/pull/8007)
```
kanej:
```suggestion
Fixed the acceptance of relative paths to `node_modules` in npm remappings ([#8007](https://github.com/NomicFoundation/hardhat/pull/8007))
```

kanej (approved):
This looks good - by which I mean:

* the tests make sense to me
* a quick manual test in the example project showed remappings being normalized

PR #8008, opened 2026-02-25
Improve multi-compiler support in the build system by adding a SolidityCompilerConfig base interface

marianfe:
## Summary 

This change introduces compiler-agnostic base interfaces: `SolidityCompilerConfig`, `SolidityCompilerUserConfig` and `SolidityCompilerType`, with a new `type` property to distinguish between compilers e.g. when downloading/injecting settings/getting the compiler/invoking compilation. 

Note: This PR also introduces a new pattern to add post-validation hooks after a config has been resolved. We need this because we need to validate that all compiler entries in the config have been 'claimed' by an existing plugin (either the built-in solc, or other plugins) and that no remaining compiler types are 'orphaned' (e.g. due to typos). 

This is the first change in a series of PRs to introduce support for solx in Hardhat 3. In later PRs, the `type` property will be used to split handling of downloads, settings/extra args, and compilation invocation between the current built-in solc pipeline, and a new solx compilation path (which will be made available via a plugin). 

## Details

- SolcUserConfig/SolcConfig extend the base types, preserving backward compatibility. 
- Export spawnCompile() and isSolcConfig() for plugin reuse (will be used in a later PR)
- Widen invokeSolc hook parameter from SolcConfig to SolidityCompilerConfig
- Add zod schema type field for runtime validation

This is a backward compatible change for Hardhat users, and a mostly backward compatible change for plugin authors - it is strictly backward compatible at runtime, although at build time there is a less likely minor type-level breaking change: invokeSolc hook parameter widens from SolcConfig to SolidityCompilerConfig. Existing handlers compile and run without changes. Plugins accessing preferWasm should use the isSolcConfig() type guard. 
- Currently no plugins seem to actually use preferWasm, this property is only used in hardhat core for the current solc pipeline.


v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -682,10 +698,18 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
     );
 
-    assertHardhatInvariant(
-      runnableCompilationJob.solcLongVersion === compiler.longVersion,
-      "The long version of the compiler should match the long version of the compilation job",
-    );
+    // For custom compilers (path set), compare short version only (long version may differ)
+    if (runnableCompilationJob.solcConfig.path !== undefined) {
+      assertHardhatInvariant(
+        runnableCompilationJob.solcConfig.version === compiler.version,
```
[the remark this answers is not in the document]
marianfe: fixed.

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -335,32 +338,54 @@ function resolveSolcConfig(
...
+  const resolvedSettings = deepMerge(
+    defaultSettings,
+    compilerConfig.settings ?? {},
+  );
+
+  // Resolve solc-specific preferWasm if this is a SolcUserConfig
+  if (isSolcUserConfig(compilerConfig)) {
+    // Resolve per-compiler preferWasm:
+    // If explicitly set, use that value.
+    // Otherwise, for ARM64 Linux in production, default to true only for
```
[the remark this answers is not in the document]
marianfe: fixed.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -439,7 +455,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     );
 
     let subgraphsWithConfig: Array<
-      [SolcConfig, DependencyGraphImplementation]
+      [SolidityCompilerConfig, DependencyGraphImplementation]
     > = [];
```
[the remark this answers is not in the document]
alcuadrado: This is a very good point tbh

v-next/hardhat/package.json
```diff
@@ -37,6 +37,9 @@
     "./types/utils": "./dist/src/types/utils.js",
     "./types/solidity": "./dist/src/types/solidity.js",
     "./console.sol": "./console.sol",
+    "./internal/coverage": "./dist/src/internal/builtin-plugins/coverage/exports.js",
+    "./internal/gas-analytics": "./dist/src/internal/builtin-plugins/gas-analytics/exports.js",
+    "./internal/solidity": "./dist/src/internal/builtin-plugins/solidity/exports.js",
```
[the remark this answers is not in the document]
alcuadrado: I wouldn't pay attention to this comment, LGTM

v-next/hardhat/src/internal/builtin-plugins/solidity/type-extensions.ts
```diff
@@ -63,23 +84,28 @@ declare module "../../../types/config.js" {
...
-    compilers: SolcConfig[];
-    overrides: Record<string, SolcConfig>;
+    compilers: SolidityCompilerConfig[];
+    overrides: Record<string, SolidityCompilerConfig>;
   }
 
   export interface SolidityConfig {
     profiles: Record<string, SolidityBuildProfileConfig>;
     npmFilesToBuild: string[];
+    registeredCompilerTypes: SolidityCompilerType[];
```
alcuadrado, line 272: great idea!

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -1,12 +1,27 @@
...
+
 import assert from "node:assert/strict";
 import { describe, it } from "node:test";
 
+// Extend compiler type definitions for testing non-solc types
+declare module "../../../../src/types/config.js" {
+  interface SolidityCompilerTypeDefinitions {
+    solx: true;
+  }
+}
```
alcuadrado: Doing test-only type-extensions like this is not possible. If you do it in a test, it affects all the package. Instead, we test these things with castings. Pretty unfortunate, but we couldn't find a way around it :(

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -753,9 +837,347 @@ describe("solidity plugin config resolution", () => {
...
+        {
+          solidity: {
+            compilers: [{ version: "0.8.28" }],
+          },
+        },
+        otherResolvedConfig,
+      );
+
+      const compiler = resolvedConfig.solidity.profiles.default.compilers[0];
+      assert.equal(compiler.type, undefined);
```
alcuadrado, line 1010: We can also set value here to `"solc"`, despite the type having to be optional due to backwards compatibility. In practice is the same, as plugins will need to check `undefined` and `"solc"`, so feel free to ignore this change.

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
alcuadrado: luckily removing it only implies a few casts, and only in this file

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
```diff
@@ -44,7 +44,9 @@ export class SolcConfigSelector {
    */
   public selectBestSolcConfigForSingleRootGraph(
     subgraph: DependencyGraph,
-  ): { success: true; config: SolcConfig } | CompilationJobCreationError {
+  ):
+    | { success: true; config: SolidityCompilerConfig }
+    | CompilationJobCreationError {
```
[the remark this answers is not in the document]
alcuadrado: This is an interesting point. But the whole selector class works with a single build profile. It will return the right `SolidityCompilerConfig` based either on the version matching logic, or `overrides`.

Copilot isn't aware that we only have a single compiler config per version, so the situation of matching the same version but different types is not possible.

However, we should make sure that we are considering the type when merging compilation jobs.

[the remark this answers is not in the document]
alcuadrado: > However, we should make sure that we are considering the type when merging compilation jobs.

Self-answer: we merge by `SolidityCompilerConfig` identity, not even value, so the `type` is taken into account.

v-next/hardhat/src/internal/builtin-plugins/solidity/hook-handlers/config.ts
```diff
@@ -21,6 +24,53 @@ export default async (): Promise<Partial<ConfigHooks>> => {
 
       return resolveSolidityUserConfig(userConfig, resolvedConfig);
     },
+    validateResolvedConfig: async (resolvedConfig) => {
+      const errors: HardhatUserConfigValidationError[] = [];
```
alcuadrado: nit-picky: can we move this to the same file of `resolveSolidityUserConfig`? that makes testing it simpler. This is explained here: https://hardhat.org/docs/plugin-development/tutorial/config#extending-the-config-validation-and-resolution

alcuadrado: Otherwise this looks good

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -1,12 +1,27 @@
...
+  }
+}
+
+import { isSolcConfig } from "../../../../src/internal/builtin-plugins/solidity/build-system/build-system.js";
 import { missesSomeOfficialNativeBuilds } from "../../../../src/internal/builtin-plugins/solidity/build-system/solc-info.js";
 import {
   resolveSolidityUserConfig,
   validateSolidityUserConfig,
 } from "../../../../src/internal/builtin-plugins/solidity/config.js";
+import configHandlerFactory from "../../../../src/internal/builtin-plugins/solidity/hook-handlers/config.js";
```
alcuadrado: This is one of the reasons for this comment: https://github.com/NomicFoundation/hardhat/pull/8008/changes#r2896986484

v-next/hardhat/test/internal/builtin-plugins/solidity/hooks.ts
```diff
@@ -60,12 +60,12 @@ describe("solidity - hooks", () => {
                   context: HookContext,
                   compiler: Compiler,
                   solcInput: CompilerInput,
-                  solcConfig: SolcConfig,
+                  solcConfig: SolidityCompilerConfig,
```
alcuadrado, line 63: The names of the params can be changed without a breaking change

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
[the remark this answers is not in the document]
alcuadrado: Yes, we would need to add `type` to `CompileCacheEntry`.

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -38,25 +40,64 @@ const commonSolcUserConfigType = z.object({
...
-const singleVersionSolcUserConfigType = solcUserConfigType.extend({
+// Per-compiler config: preferWasm is only allowed for solc (type undefined or "solc")
+const compilerUserConfigType = conditionalUnionType(
+  [
+    [
+      (data) =>
+        !isObject(data) ||
+        !("type" in data) ||
+        data.type === undefined ||
+        data.type === "solc",
```
alcuadrado: I think this is what you meant here

```suggestion
      (data) =>
        isObject(data) &&
        (!("type" in data) || data.type === undefined || data.type === "solc"),
```

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -38,25 +40,64 @@ const commonSolcUserConfigType = z.object({
...
-const solcUserConfigType = z.object({
+const baseCompilerFields = {
+  type: z.string().optional(),
   version: z.string(),
   settings: z.any().optional(),
   path: z.string().optional(),
-  preferWasm: z.boolean().optional(),
+};
+
+const incompatibleVersionFields = {
```
alcuadrado: Maybe `incompatibleSolidityCompilerConfigFields`?

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -38,25 +40,64 @@ const commonSolcUserConfigType = z.object({
...
+        isObject(data) &&
+        "type" in data &&
+        typeof data.type === "string" &&
+        data.type !== "solc",
+      z
+        .object({
+          ...baseCompilerFields,
+          ...incompatibleVersionFields,
+        })
+        .passthrough(),
```
alcuadrado: This isn't needed because we only use Zod to validate types, not to parse the input. The reason for this is that we have our custom resolution step, which is in charge of what Zod calls "parse".

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -38,25 +40,64 @@ const commonSolcUserConfigType = z.object({
...
+    ],
+  ],
+  "Expected a valid compiler configuration",
+);
+
+// Note: this is only to match the setup present in ./type-extensions.ts.
+// preferWasm is allowed here for all types; it is only meaningful for solc
+// (type undefined or "solc") and silently ignored for other compiler types.
+const singleVersionSolcUserConfigType = z.object({
```
alcuadrado: I think the `SingleVersionSolidityUserConfig` doesn't need to always have `preferWasm`. I'll explain this on my proposed type changes.

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -38,25 +40,64 @@ const commonSolcUserConfigType = z.object({
...
   preferWasm: z.boolean().optional(),
 });
 
 const multiVersionSolcUserConfigType = commonSolcUserConfigType.extend({
-  compilers: z.array(solcUserConfigType).nonempty(),
-  overrides: z.record(z.string(), solcUserConfigType).optional(),
+  compilers: z.array(compilerUserConfigType).nonempty(),
+  overrides: z.record(z.string(), compilerUserConfigType).optional(),
   isolated: z.boolean().optional(),
   preferWasm: z.boolean().optional(),
```
alcuadrado: On the contrary, `preferWasm` needs to be here, as we don't have a top-level `type`.

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -335,32 +379,54 @@ function resolveSolcConfig(
   };
 
   if (production) {
```
alcuadrado: I would do:

```suggestion
  if (production && isSolcUserConfig(compilerConfig)) {
```

as this probably only applies to solc.

v-next/hardhat/src/types/hooks.ts
```diff
@@ -93,6 +93,20 @@ export interface ConfigHooks {
...
+   * This hook runs after all plugins have resolved their config. Use it to
+   * validate cross-cutting concerns that require the fully resolved config
+   * (e.g., checking that all compiler types are registered).
+   *
+   * @param resolvedConfig The fully resolved config.
+   * @returns An array of validation errors.
+   */
+  validateResolvedConfig: (
+    resolvedConfig: HardhatConfig,
+  ) => Promise<HardhatUserConfigValidationError[]>;
```
alcuadrado: This should use a new type `HardhatConfigValidationError`

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -624,8 +699,8 @@ describe("solidity plugin config resolution", () => {
       );
 
       const compilers = resolvedConfig.solidity.profiles.default.compilers;
-      assert.equal(compilers[0].preferWasm, true);
-      assert.equal(compilers[1].preferWasm, false);
+      assert.equal((compilers[0] as SolcConfig).preferWasm, true);
```
alcuadrado: These casts are only needed because of the type extension. Once removed, they aren't needed.

v-next/hardhat/src/internal/core/hre.ts
```diff
@@ -252,6 +253,21 @@ export async function resolveUserConfigToHardhatConfig(
...
+    "config",
+    "validateResolvedConfig",
+    [config],
+  );
+  const resolvedConfigValidationErrors = resolvedConfigValidationResults.flat();
+
+  if (resolvedConfigValidationErrors.length > 0) {
+    return {
+      success: false,
+      userConfigValidationErrors: resolvedConfigValidationErrors,
```
alcuadrado: This would lead to the resolved config errors being displayed as user config errors, confusing the users.

For example, if a user, without any plugin, has this user config:

```ts

export default defineConfig({
  solidity: {
    type: "foo" as any,
    version: "0.8.28",
  },
});
```

_ignore the `as any`, this is just an example, but it isn't exclusive about that field_

They would get an error like this:

```
        * Config error in config.solidity.profiles.default.compilers.0.type: Unknown compiler type "foo". Registered types: solc
```

But that path won't be present in their user config.

Instead, we should treat them as a different case.

alcuadrado: Note that this function is also used in the network manager.

v-next/hardhat/src/internal/builtin-plugins/solidity/type-extensions.ts
```diff
@@ -20,10 +20,31 @@ declare module "../../../types/config.js" {
     | MultiVersionSolidityUserConfig
     | BuildProfilesSolidityUserConfig;
 
-  export interface SolcUserConfig {
+  /**
```
alcuadrado, line 16: I struggle to see the big picture here. I don't see how Solidity and Solc types are related to each other, and how the user will upcast/narrow between the different compiler types.

v-next/hardhat/test/internal/builtin-plugins/solidity/hooks.ts
alcuadrado: Just to be clear: this was more of a note/reflection, rather than actionable feedback that needs to be addressed.

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -38,25 +40,64 @@ const commonSolcUserConfigType = z.object({
...
   compilers: incompatibleFieldType("This field is incompatible with `version`"),
   overrides: incompatibleFieldType("This field is incompatible with `version`"),
   profiles: incompatibleFieldType("This field is incompatible with `version`"),
-});
+};
 
-// NOTE: This is only to match the setup present in ./type-extensions.ts
-const singleVersionSolcUserConfigType = solcUserConfigType.extend({
+// Per-compiler config: preferWasm is only allowed for solc (type undefined or "solc")
+const compilerUserConfigType = conditionalUnionType(
```
alcuadrado: I think keeping this zod types/schemas aligned with the actual types can make it easier to understand/maintain them.

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -335,32 +379,54 @@ function resolveSolcConfig(
...
 
   return {
-    version: solcConfig.version,
-    settings: deepMerge(defaultSolcConfigSettings, solcConfig.settings ?? {}),
-    path: solcConfig.path,
-    preferWasm: resolvedPreferWasm,
+    type: compilerConfig.type,
+    version: compilerConfig.version,
+    settings: resolvedSettings,
+    path: compilerConfig.path,
```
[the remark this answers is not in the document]
alcuadrado: This comment is irrelevant, because we have our own resolution. This change looks good to me.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: This logic is now outdated. These maps should be aware of the compiler type.

.changeset/cold-deer-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Introduce multi-compiler abstraction with SolidityCompilerConfig base interface
```
alcuadrado: This should explain what the core new feature is: new plugins can add new compiler types

.changeset/cold-deer-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
```
alcuadrado, line 2: We need a separate changeset for the new config hook.

alcuadrado: https://github.com/NomicFoundation/hardhat/commit/9bab76a2d7d8c7c379fef5ca8d5cbe67460bf4e7

.changeset/cold-deer-buy.md
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/4e2bd4ab59061a1654e8e78ee2841fe1abddefc2

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -79,14 +83,26 @@ import { shouldSuppressWarning } from "./warning-suppression.js";
 const log = debug("hardhat:core:solidity:build-system");
 
 /**
- * Resolves the preferWasm setting for a given solc config, falling back
+ * Returns true if the given compiler config is a SolcConfig.
+ */
+export function isSolcConfig(
```
alcuadrado: If we introduce new types, we should ideally use the new names consistently. This May only make sense with my proposed changes.

alcuadrado: https://github.com/NomicFoundation/hardhat/commit/3c6dcc9c3bdec5b020bead04722689d9fd2030af

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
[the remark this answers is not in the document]
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/c52e07683da1afb14aa2780243cf554ff12afdb8

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/6daf3fff422fd0f09049fa38785005828a08b242

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/b180bb9acb381ebe33c18a585592cb8799086fbe and https://github.com/NomicFoundation/hardhat/commit/5a50239c9ffd37388f5721e686e937b82f1d187c

v-next/hardhat/src/internal/builtin-plugins/solidity/hook-handlers/config.ts
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/ffa3ff6a141a069bd10561ae94ece72b1af8d6ec

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
alcuadrado: Ignore this comment please

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/d4206c56b64f7608c733a35266def1661f1a8055

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
alcuadrado: Done as part of https://github.com/NomicFoundation/hardhat/commit/d4206c56b64f7608c733a35266def1661f1a8055

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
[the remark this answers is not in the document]
alcuadrado: Done as part of https://github.com/NomicFoundation/hardhat/commit/d4206c56b64f7608c733a35266def1661f1a8055

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
alcuadrado: As mentioned above, it's not needed. The reason is that the types I introduced in https://github.com/NomicFoundation/hardhat/commit/8db2c1c19bb5317f64ede4a5780b8f6a8e1fba06 are aware of the `type`, and only have `preferWasm` when it's missing or `"solc"`.

The zod validation introduced in https://github.com/NomicFoundation/hardhat/commit/d4206c56b64f7608c733a35266def1661f1a8055 follow that same logic

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
alcuadrado: My thinking was invalid. Instead, we need to post-validate that the top-level `preferWasm` is only used for `solc`.

https://github.com/NomicFoundation/hardhat/commit/558ec52b262de2328f68b8dc732fdd7b809b0534

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/141b9c138efacb8f63a19305293ebf257fbd12db

v-next/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -335,32 +379,54 @@ function resolveSolcConfig(
...
+    };
+    return solcResolved;
   }
 
   return {
-    version: solcConfig.version,
-    settings: deepMerge(defaultSolcConfigSettings, solcConfig.settings ?? {}),
-    path: solcConfig.path,
-    preferWasm: resolvedPreferWasm,
+    type: compilerConfig.type,
```
alcuadrado: This becomes unreachable (i.e. `compilerConfig` is `never`) once you remove the type-extension from the test. 

https://github.com/NomicFoundation/hardhat/commit/b180bb9acb381ebe33c18a585592cb8799086fbe

v-next/hardhat/src/internal/builtin-plugins/solidity/type-extensions.ts
alcuadrado: I changed the types in https://github.com/NomicFoundation/hardhat/commit/8db2c1c19bb5317f64ede4a5780b8f6a8e1fba06 so that each compiler has its own hierarchy. I also built the general (i.e. not specific to a compiler) types as a mapped type, so that the value of `type` lets typescript narrow to the right hierarchy.

v-next/hardhat/src/internal/core/hre.ts
alcuadrado: Implemented in https://github.com/NomicFoundation/hardhat/commit/98909d4bc28a14ebf416bc7458ac7750823d7e2a and https://github.com/NomicFoundation/hardhat/commit/adbed0af1b21f87ec8be3e5bcbaa6e642140e3a5

I was also forced to update another part of the code to work with the new behavior: https://github.com/NomicFoundation/hardhat/commit/262e3dc7d8f0a8bd0607f641a91e73ec1f67d07e

v-next/hardhat/src/types/hooks.ts
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/98909d4bc28a14ebf416bc7458ac7750823d7e2a

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/ffa3ff6a141a069bd10561ae94ece72b1af8d6ec

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
alcuadrado: https://github.com/NomicFoundation/hardhat/commit/5a50239c9ffd37388f5721e686e937b82f1d187c

v-next/hardhat/test/internal/builtin-plugins/solidity/config.ts
alcuadrado: let's ignore it

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
alcuadrado: This should include information about the compiler type, and done in a careful way to preserve backwards compatibility.

alcuadrado: This should include information about the compiler type, and done in a careful way to preserve backwards compatibility.

alcuadrado: This comment also applies to the `SolidityBuildInfo` interface.

alcuadrado: Way trickier than expected, but https://github.com/NomicFoundation/hardhat/commit/6ed305607adf7a03c89bdbca3dc95de7740eeaa6

alcuadrado: https://github.com/NomicFoundation/hardhat/commit/6ed305607adf7a03c89bdbca3dc95de7740eeaa6

alcuadrado (approved):
🚀

PR #8009, opened 2026-02-26
Add downloadCompilers parallel hook for extensible compiler downloads

marianfe:
## Summary

Create and export two new hooks in the build system, `downloadCompilers` and `getCompiler`, which can later be used for the solx plugin. 
This is part 2 in the chain of PRs to introduce solx support in Hardhat 3. 

The `downloadCompilers` hook is a parallel hook that allows each compiler plugin to handle the downloads for their own `compiler.type`, and we need the `getCompiler` chained hook to be able to correctly set the path to the new downloaded binaries prior to compilation invocation. 

The previous Part 1 PR is here: https://github.com/NomicFoundation/hardhat/pull/8008

## Details 

- Extract solc download logic into a new downloadCompilers parallel hook.
- All registered handlers run concurrently via runParallelHandlers.
- Built-in handler filters for solc configs only (type undefined or "solc") using isSolcConfig()
- Third-party plugins register their own handlers for custom compiler types
- Collect all compiler configs from all build profiles (including overrides) before downloading

Depends on: #8008 (multi-compiler abstraction)

### DIFF
--- a/.changeset/good-eels-buy.md
+++ b/.changeset/good-eels-buy.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Add downloadCompilers parallel hook for extensible compiler downloads
--- a/.changeset/good-eels-buy.md
+++ b/.changeset/good-eels-buy.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Add `downloadCompilers` and `getCompiler` hooks for extensible custom compiler support
+Add `SolidityHooks#downloadCompilers` and `SolidityHooks#getCompiler` hooks for extensible custom compiler support ([#8009](https://github.com/NomicFoundation/hardhat/pull/8009))
--- a/.changeset/good-eels-buy.md
+++ b/.changeset/good-eels-buy.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Add downloadCompilers parallel hook for extensible compiler downloads
+Add `downloadCompilers` and `getCompiler` hooks for extensible custom compiler support
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -58,6 +58,7 @@ import pMap from "p-map";
 
 import { FileBuildResultType } from "../../../../types/solidity/build-system.js";
 import { DEFAULT_BUILD_PROFILE } from "../build-profiles.js";
+import { getSolcCompilerForConfig } from "../solidity-hooks.js";
 
 import {
   getArtifactsDeclarationFile,
@@ -89,20 +90,6 @@ export function isSolcSolidityCompilerConfig(
   config: SolidityCompilerConfig,
 ): config is SolcSolidityCompilerConfig {
   return config.type === undefined || config.type === "solc";
-}
-
-/**
- * Resolves the preferWasm setting for a given compiler config, falling back
- * to the build profile's preferWasm if not set on the compiler.
- */
-function resolvePreferWasm(
-  compilerConfig: SolidityCompilerConfig,
-  buildProfilePreferWasm: boolean,
-): boolean {
-  if (isSolcSolidityCompilerConfig(compilerConfig)) {
-    return compilerConfig.preferWasm ?? buildProfilePreferWasm;
-  }
-  return false;
 }
 
 // Compiler warnings to suppress from build output.
@@ -511,10 +498,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
           "getCompiler",
           [compilerConfig],
           async (_context, cfg) =>
-            getCompiler(cfg.version, {
-              preferWasm: resolvePreferWasm(cfg, buildProfile.preferWasm),
-              compilerPath: cfg.path,
-            }),
+            getSolcCompilerForConfig(cfg, buildProfile.preferWasm),
         );
         longVersion = compiler.longVersion;
         longVersionMap.set(compilerConfig.version, longVersion);
@@ -758,10 +742,7 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       "getCompiler",
       [runnableCompilationJob.solcConfig],
       async (_context, cfg) =>
-        getCompiler(cfg.version, {
-          preferWasm: resolvePreferWasm(cfg, buildProfile.preferWasm),
-          compilerPath: cfg.path,
-        }),
+        getSolcCompilerForConfig(cfg, buildProfile.preferWasm),
     );
 
     log(
--- a/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -1294,16 +1294,21 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     }
 
     for (const job of runnableCompilationJobs) {
+      const compilerType = job.solcConfig.type ?? "solc";
       const solcVersion = job.solcConfig.version;
       const solcInput = await job.getSolcInput();
       const evmVersion =
         solcInput.settings.evmVersion ??
         `Check solc ${solcVersion}'s doc for its default evm version`;
 
-      let jobsPerVersion = jobsPerVersionAndEvmVersion.get(solcVersion);
+      // Group by compiler type + Solidity version to produce separate log
+      // lines for e.g. "solc 0.8.33" vs "solx 0.1.3 (Solidity 0.8.33)".
+      const groupKey = `${compilerType}\0${solcVersion}`;
+
+      let jobsPerVersion = jobsPerVersionAndEvmVersion.get(groupKey);
       if (jobsPerVersion === undefined) {
         jobsPerVersion = new Map();
-        jobsPerVersionAndEvmVersion.set(solcVersion, jobsPerVersion);
+        jobsPerVersionAndEvmVersion.set(groupKey, jobsPerVersion);
       }
 
       let jobsPerEvmVersion = jobsPerVersion.get(evmVersion);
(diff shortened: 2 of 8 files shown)


.changeset/good-eels-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Add `downloadCompilers` and `getCompiler` hooks for extensible custom compiler support
```
alcuadrado:
```suggestion
Add `SolidityHooks#downloadCompilers` and `SolidityHook#getCompiler` hooks for extensible custom compiler support ([#8009](https://github.com/NomicFoundation/hardhat/pull/8009)).
```

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -477,17 +477,23 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     // get longVersion and isWasm from the compiler for each version
     const solcVersionToLongVersion = new Map<string, string>();
```
alcuadrado: As mentioned in the other PR, these maps and all of their usages are now incorrect, as they are mixing up different compiler types. This was addressed in https://github.com/NomicFoundation/hardhat/commit/6daf3fff422fd0f09049fa38785005828a08b242

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -246,7 +246,8 @@ export class CompilationJobImplementation implements CompilationJob {
 
     const jobHash = await createNonCryptographicHashId(preimage);
 
-    return `solc-${this.solcConfig.version.replaceAll(".", "_")}-${jobHash}`;
+    const compilerType = this.solcConfig.type ?? "solc";
```
alcuadrado: Addressed in https://github.com/NomicFoundation/hardhat/commit/6ed305607adf7a03c89bdbca3dc95de7740eeaa6 which is more complex to keep solc formats and ids backwards compatible

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -1032,8 +1038,10 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
   ): Promise<CompilerOutput> {
     const quiet = options?.quiet ?? false;
 
-    // We download the compiler for the build info as it may not be configured
-    // in the HH config, hence not downloaded with the other compilers
+    // Build info recompilation is always solc-only: build info files are
```
alcuadrado, line 1076: I would say that this is an outdated assumption. BuildInfos should include the compiler type, so we should run the same download logic here.

See this change where I added `compilerType` to `SolidityBuildInfo`: https://github.com/NomicFoundation/hardhat/commit/6ed305607adf7a03c89bdbca3dc95de7740eeaa6

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -24,6 +24,7 @@ import {
 } from "@nomicfoundation/hardhat-utils/fs";
 
 import { SolidityBuildSystemImplementation } from "../../../../../src/internal/builtin-plugins/solidity/build-system/build-system.js";
+import createSolidityHookHandlers from "../../../../../src/internal/builtin-plugins/solidity/hook-handlers/solidity.js";
```
alcuadrado, line 27: Small tip: Moving the implementation of the handler into a separate file makes the test nicer.

The way we tend to think about it is this:
- If the handler requires the context, we may keep the implementation there, and use more integration-like tests.
- If the handler is more pure, we separate it and unit test it.

v-next/hardhat/src/internal/builtin-plugins/solidity/index.ts
```diff
@@ -41,6 +41,7 @@ const hardhatPlugin: HardhatPlugin = {
   hookHandlers: {
     config: () => import("./hook-handlers/config.js"),
     hre: () => import("./hook-handlers/hre.js"),
+    solidity: () => import("./hook-handlers/solidity.js"),
```
alcuadrado, line 44: This is a surprisingly nice pattern

v-next/hardhat/test/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: For example, this test file has `hooks.setContext({} as HookContext);` which I don't really like. It assumes that the `HookContext` is not used b any of the tests. I'd rather unit-test the handler.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -1224,16 +1229,21 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
     }
 
     for (const job of runnableCompilationJobs) {
+      const compilerType = job.solcConfig.type ?? "solc";
```
alcuadrado, line 1268: I guess the changes in this file from here on in fact belong to #8008, but it's not important. Just mentioning it for my future reference and clarity.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -1224,16 +1229,21 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
       const solcVersion = job.solcConfig.version;
       const solcInput = await job.getSolcInput();
       const evmVersion =
         solcInput.settings.evmVersion ??
         `Check solc ${solcVersion}'s doc for its default evm version`;
 
-      let jobsPerVersion = jobsPerVersionAndEvmVersion.get(solcVersion);
+      // Group by compiler type + Solidity version to produce separate log
+      // lines for e.g. "solc 0.8.33" vs "solx 0.1.3 (Solidity 0.8.33)".
+      const groupKey = `${compilerType}#${solcVersion}`;
```
alcuadrado, line 1277: I'm more of a fan of multi-level maps than this combined keys, as the type system enforces that you give them some thought. I'm not agains this either.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: We could create a partial SolidityCompilerConfig like this:

```ts
import type { SolidityBuildInfo } from "hardhat/types/solidity";
import type { SolidityCompilerConfig } from "hardhat/types/config";

function getPartialSolidityCompilerConfigFromBuildInfo(
  buildInfo: SolidityBuildInfo,
): SolidityCompilerConfig {
  return {
    // This cast may result in a "type" that no plugin can handle. So we should
    // probably have a way to know if a handler downlaoded a certain compiler,
    // and throw otherwise. This can't be done with parallel one, but can be
    // done with chained ones.
    type: buildInfo.compilerType as SolidityCompilerConfig["type"],
    version: buildInfo.solcVersion,
    settings: buildInfo.input.settings,
  };
}
```

alcuadrado: There's no way to know if the buildinfo was built with wasm solc, but that's ok. We were already doing `preferWasm: false` below

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -693,19 +699,19 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
-        ),
-        compilerPath: runnableCompilationJob.solcConfig.path,
-      },
+    const compiler = await this.#hooks.runHandlerChain(
+      "solidity",
+      "getCompiler",
+      [runnableCompilationJob.solcConfig],
+      async (_context, cfg) =>
+        getCompiler(cfg.version, {
+          preferWasm: resolvePreferWasm(cfg, buildProfile.preferWasm),
```
alcuadrado: I like this. The `buildProfile.preferWasm` shouldn't affect any other compiler type. I actually forbid using it for build profiles with non-solc compilers in https://github.com/NomicFoundation/hardhat/commit/558ec52b262de2328f68b8dc732fdd7b809b0534

v-next/hardhat/test/internal/builtin-plugins/solidity/hooks.ts
```diff
@@ -23,6 +28,38 @@ import { useFixtureProject } from "@nomicfoundation/hardhat-test-utils";
...
+              nextUserConfig: HardhatUserConfig,
+              nextResolveConfigurationVariable: ConfigurationVariableResolver,
+            ) => Promise<HardhatConfig>,
+          ) => {
+            const resolved = await next(
+              userConfig,
+              resolveConfigurationVariable,
+            );
+            // eslint-disable-next-line @typescript-eslint/consistent-type-assertions -- test registers unregistered types
+            resolved.solidity.registeredCompilerTypes.push(...(types as any[]));
```
alcuadrado, line 56: Just for future reference: in actual plugins we return new resolved config objects rather than modifying it. Otherwise understanding what's going on in the presence of multiple plugins gets really tricky.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/compilation-job.ts
alcuadrado: When rebasing the branch `pr-8009-followup` on top of `pr-8008-followup` this change was dropped.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: When rebasing the branch `pr-8009-followup` on top of `pr-8008-followup` this change was dropped.

v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: Reviewed these changes. LGTM

alcuadrado (approved):
LGTM. I created an issue for the only open comment, which can be addressed in the future, if needed: https://github.com/NomicFoundation/hardhat/issues/8081

PR #8012, opened 2026-02-26
Expose `Result` type for task action success/failure signaling

schaable:
## Summary
- Expose `Result` interface that task actions can optionally return to signal success or failure to the CLI
- When `task.run()` returns a `Result` with `success: false`, the CLI sets `process.exitCode = 1`
- Both success and failure can optionally carry a value
- Backward compatible: tasks that return `undefined` or non-`Result` values continue to work unchanged
- Exports the `isResult()` type guard and `successfulResult()` / `errorResult()` factory helpers from `hardhat/utils/result`

## Usage

```typescript
import type { Result } from "hardhat/types/result";
import { successfulResult, errorResult } from "hardhat/utils/result";

// Success without data (e.g. keystore, telemetry)
async function configureAction(): Promise<Result<void, void>> {
  if (conflictingFlags) {
    return errorResult();
  }
  // ...
  return successfulResult();
}

// Success/failure with data (e.g. test runners)
interface TestSummary {
  failed: number;
  passed: number;
}

async function testAction(): Promise<Result<TestSummary, TestSummary>> {
  const summary = await runTests();

  if (summary.failed > 0) {
    return errorResult(summary); // CLI sets exitCode=1, caller still gets data
  }

  return successfulResult(summary);
}
```

Docs: https://github.com/NomicFoundation/hardhat-website/pull/228


v-next/hardhat/package.json
```diff
@@ -38,6 +38,7 @@
     "./types/solidity": "./dist/src/types/solidity.js",
     "./console.sol": "./console.sol",
     "./utils/contract-names": "./dist/src/utils/contract-names.js",
+    "./utils/task-result": "./dist/src/utils/task-result.js",
```
schaable: This could be:
```suggestion
    "./utils/task": "./dist/src/utils/task.js",
```
if we plan to add other task helpers in the future. Or we could use the task namespace.

alcuadrado:
I'll review this better tomorrow, but I think using `taskSuccess` and the other helpers from non-builtin plugins would force us to peer-bump all of them, for little user-value. Other than that, I like the general idea. They can achieve the same type-safety without the helpers though.

v-next/hardhat/src/types/tasks.ts
```diff
@@ -543,6 +543,28 @@ export interface TaskManager {
...
+ * A TaskResult can optionally carry a value, regardless of whether it
+ * represents success or failure. This is useful for tasks like test
+ * runners that need to report summary data even on failure.
+ *
+ * @template ValueT The type of the optional value. Defaults to undefined,
+ * meaning no value is carried.
+ */
+export interface TaskResult<ValueT = undefined> {
+  success: boolean;
+  value?: ValueT;
```
alcuadrado: Why would this be optional? If you want optionality you can return `TaskResult<Foo | undefined>`.

alcuadrado: Also, why not a more general ?

```ts
type Result<ValueT, ErrorT> = { success: true, value: ValueT} | { success: false, error: ErrorT};
```

If we have that we can start using it in more places.

alcuadrado: we actually have that internally `v-next/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/types.ts`

schaable: some successful tasks don't return anything, so requiring it would force them to return `{ success: true, value: undefined }`, which is awkward. And some failed tasks need to return a value (the `test` task and subtasks return the test summary)

schaable: I just checked the Result type in the resolver, we could use that, but I'd merge it with what I have in TaskResult so we can return a value on failure. I'd make the error optional as well, to avoid forcing the task to return an error when it doesn't need to.

v-next/hardhat/package.json
```diff
@@ -38,6 +38,8 @@
     "./types/solidity": "./dist/src/types/solidity.js",
     "./console.sol": "./console.sol",
     "./utils/contract-names": "./dist/src/utils/contract-names.js",
+    "./types/result": "./dist/src/types/result.js",
```
schaable: I'm tempted to export the type from `"./types/utils"` instead of its own path, since grouping them seems logical. wdyt @alcuadrado?

v-next/hardhat/src/utils/result.ts
```diff
@@ -0,0 +1,57 @@
...
+  isError?: (e: unknown) => e is ErrorT,
+): value is Result<ValueT, ErrorT> {
+  if (!isObject(value) || typeof value.success !== "boolean") {
+    return false;
+  }
+
+  if (value.success === true) {
+    return isValue === undefined || isValue(value.value);
+  }
+
```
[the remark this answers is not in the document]
schaable: Good point, done in https://github.com/NomicFoundation/hardhat/pull/8012/commits/0c8d1f9848cbcf13adc2ca14ec0ef7ac1a0d0488

v-next/hardhat/test/utils/result.ts
```diff
@@ -0,0 +1,139 @@
...
+    it("should return true for a failed Result", function () {
+      assert.equal(isResult({ success: false, error: "oops" }), true);
+    });
+
+    it("should return true for a successful Result without a value", function () {
+      assert.equal(isResult({ success: true }), true);
+    });
+
+    it("should return true for a failed Result without an error", function () {
+      assert.equal(isResult({ success: false }), true);
```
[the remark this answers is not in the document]
schaable: Related to previous comment. Fixed in https://github.com/NomicFoundation/hardhat/pull/8012/commits/0c8d1f9848cbcf13adc2ca14ec0ef7ac1a0d0488

v-next/hardhat/src/internal/cli/main.ts
```diff
@@ -217,7 +218,14 @@ export async function main(
 
     log(`Running task "${task.id.join(" ")}"`);
 
-    await Promise.all([task.run(taskArguments), sendTaskAnalytics(task.id)]);
+    const [taskResult] = await Promise.all([
+      task.run(taskArguments),
+      sendTaskAnalytics(task.id),
+    ]);
+
+    if (isResult(taskResult) && !taskResult.success) {
```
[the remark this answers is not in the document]
schaable: Ideally, we set the exitCode in one place (here).

v-next/hardhat/package.json
alcuadrado: yea, i would leave it in `./types/utils`, and maybe its own `exports` only for the helpers.

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -1767,4 +1767,43 @@ GLOBAL OPTIONS:
...
+    it("should set process.exitCode = 1 when task returns Result with success: false and an error", async function () {
+      await runMain("npx hardhat failing-task-with-value");
+      assert.equal(process.exitCode, 1);
+    });
+
+    it("should not set process.exitCode when task returns Result with success: true and no value", async function () {
+      await runMain("npx hardhat succeeding-task-no-value");
+      assert.equal(process.exitCode, undefined);
+    });
+  });
```
alcuadrado, line 1820: nitpicky: maybe worth adding a test that sets the exitCode to something like 123, then runs a successful tasks, and asserts that it doesn't unset it.

v-next/hardhat/src/utils/result.ts
```diff
@@ -0,0 +1,57 @@
...
+import { isObject } from "@nomicfoundation/hardhat-utils/lang";
+
+/**
+ * Creates a successful Result with the given value.
+ *
+ * @param value The value to include in the result.
+ * @returns A Result with success: true and the given value.
+ */
+export function successResult<ValueT = undefined>(
```
alcuadrado: For some reason, this name sounds a bit weird to me. I would have expected `successfulResult`

alcuadrado (approved):
There are some minor comments about naming, exports and a test, but pre-approving.

alcuadrado:
Also: on could create an issue to change all the builtin plugins' tasks to use Result, and update their tests so that they don't depend on `process.exitCode`

alcuadrado:
And we should update the docs of `setAction` and `setInlineAction`, both in the website and the jsdoc.

v-next/hardhat/package.json
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/8012/commits/1c799afe46256ffeac2d89e605f8c70685de0116

v-next/hardhat/test/internal/cli/main.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/8012/commits/a873933a3bbf6eb30dcc8e7a1879c2e60ed83e9f

v-next/hardhat/src/utils/result.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/8012/commits/e35ba8d5af4364e2d3eb665b79385b9a4956c52e

v-next/hardhat/test/internal/cli/main.ts
```diff
@@ -1781,11 +1781,23 @@ GLOBAL OPTIONS:
       assert.equal(process.exitCode, 1);
     });
 
+    it("should override process.exitCode to 1 when task returns Result with success: false", async function () {
+      process.exitCode = 123;
```
alcuadrado, line 1785: These new tests aren't cleaning up `procese.exitCode`. They need a `try/finally`.

v-next/hardhat/src/types/tasks.ts
```diff
@@ -5,6 +5,8 @@ import type {
   PositionalArgumentDefinition,
 } from "./arguments.js";
 import type { HardhatRuntimeEnvironment } from "./hre.js";
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars -- used in JSDoc {@link} */
```
alcuadrado, line 8: This is pretty weird. jsdoc can use import, or we can just remove the `@link`.

v-next/hardhat/test/internal/cli/main.ts
schaable: There's an afterEach that clears the exitcode. Do you think we also need the finally?

v-next/hardhat/src/types/tasks.ts
schaable: ESLint doesn't recognize that the import is used in JSDoc, so it thinks it's unused. I'd keep it, since users don't have an easy way to check the type without the `@link`.

schaable:
> Also: on could create an issue to change all the builtin plugins' tasks to use Result, and update their tests so that they don't depend on `process.exitCode`

This is already done in https://github.com/NomicFoundation/hardhat/pull/8015

> And we should update the docs of `setAction` and `setInlineAction`, both in the website and the jsdoc

Done in https://github.com/NomicFoundation/hardhat/pull/8012/commits/db636ad19293160eba9f59d4510503a47cf80de4 and https://github.com/NomicFoundation/hardhat-website/pull/228

v-next/hardhat/test/internal/cli/main.ts
alcuadrado: oops, i missed that

alcuadrado (approved):

PR #8015, opened 2026-02-28
feat: return typed Result from test runner and telemetry task actions

schaable:
Update test runner (`solidity`, `node`, `mocha`) and `telemetry` task actions to return `Result` instead of setting `process.exitCode`. Define a shared `TestSummary` type and use `isResult` with type guards in the parent test task to aggregate results without type assertions.

This should be reviewed after https://github.com/NomicFoundation/hardhat/pull/8012.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
@@ -4,13 +4,16 @@ import type {
   Task,
   TaskArguments,
 } from "../../../types/tasks.js";
+import type { TestSummary } from "../../../types/test.js";
 
 import {
   assertHardhatInvariant,
   HardhatError,
 } from "@nomicfoundation/hardhat-errors";
+import { isObject } from "@nomicfoundation/hardhat-utils/lang";
 import chalk, { type ChalkInstance } from "chalk";
 
+import { errorResult, isResult, successResult } from "../../../utils/result.js";
 import { HardhatRuntimeEnvironmentImplementation } from "../../core/hre.js";
 
 interface TestActionArguments {
@@ -19,6 +22,10 @@ interface TestActionArguments {
   grep: string | undefined;
   noCompile: boolean;
   verbosity: number;
+}
+
+function isTestSummary(value: unknown): value is TestSummary {
+  return isObject(value);
 }
 
 const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
@@ -49,18 +56,10 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
     hre._coverage.disableReport();
   }
 
-  const testSummaries: Record<
-    string,
-    {
-      failed?: number;
-      passed?: number;
-      skipped?: number;
-      todo?: number;
-      failureOutput?: string;
-    }
-  > = {};
+  const testSummaries: Record<string, TestSummary> = {};
 
   let failureIndex = 1;
+  let hasFailures = false;
   for (const subtask of thisTask.subtasks.values()) {
     const files = getTestFilesForSubtask(subtask, testFiles, subtasksToFiles);
 
@@ -88,14 +87,26 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
 
     if (subtask.options.has("testSummaryIndex")) {
       args.testSummaryIndex = failureIndex;
+    }
 
-      testSummaries[summaryId] = await subtask.run(args);
-      failureIndex += testSummaries[summaryId].failed ?? 0;
-    } else if (summaryId === "mocha") {
-      // mocha doesn't use the testSummaryIndex, but it does return failure & success counts
-      testSummaries[summaryId] = await subtask.run(args);
-    } else {
-      await subtask.run(args);
+    const subtaskResult = await subtask.run(args);
+
+    if (isResult(subtaskResult, isTestSummary, isTestSummary)) {
+      const summary = subtaskResult.success
+        ? subtaskResult.value
+        : subtaskResult.error;
+
+      if (summary !== undefined) {
+        testSummaries[summaryId] = summary;
+
+        if (subtask.options.has("testSummaryIndex")) {
+          failureIndex += summary.failed ?? 0;
+        }
+      }
+
+      if (!subtaskResult.success) {
+        hasFailures = true;
+      }
     }
   }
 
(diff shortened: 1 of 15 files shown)


v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -88,14 +87,26 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
...
+
+        if (subtask.options.has("testSummaryIndex")) {
+          failureIndex += summary.failed ?? 0;
+        }
+      }
+
+      if (!subtaskResult.success) {
+        hasFailures = true;
+      }
     }
```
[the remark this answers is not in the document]
schaable: Test subtasks should always return a `TestSummary` payload as indicated by the return type.

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -21,6 +24,10 @@ interface TestActionArguments {
   verbosity: number;
 }
 
+function isTestSummary(value: unknown): value is TestSummary {
+  return isObject(value);
```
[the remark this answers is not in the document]
schaable: This is probably overkill.

v-next/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -179,20 +176,25 @@ describe("solidity-test/task-action", function () {
...
+    });
+
+    it("should return a success result when all tests pass", async () => {
+      hre = await createHardhatRuntimeEnvironment(hardhatConfigAllTests);
+
+      const result = await hre.tasks
+        .getTask(["test", "solidity"])
+        .run({ noCompile: true });
+      assert.equal(result.success, true);
     });
```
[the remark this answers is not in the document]
schaable: Updated to assert the full object.

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
alcuadrado:
```suggestion
) => Result<void, void> {
```

What about adding, to enforce consistency.

v-next/hardhat/src/types/test.ts
```diff
@@ -5,3 +5,16 @@ export interface HardhatTestUserConfig {}
 /* eslint-disable-next-line @typescript-eslint/no-empty-interface -- Empty
 interface allow plugins to extend the Test configuration for Hardhat. */
 export interface HardhatTestConfig {}
+
+/**
+ * Summary of a test run, containing counts of test outcomes and optional
+ * failure output. All fields are optional because different test runners
+ * may provide different subsets.
+ */
+export interface TestSummary {
```
alcuadrado, line 13: What if we make them non-optional? This should be doable. Given that they are counts, it's easy to set a value of `0` if the feature is not supported in a given test runner.

Of course, failure output should still be optional.

v-next/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -179,20 +176,31 @@ describe("solidity-test/task-action", function () {
       } finally {
         process.env.HH_TEST = hhTest;
         process.env.NODE_ENV = nodeEnv;
-        process.exitCode = exitCode;
       }
     });
 
-    it("should run all the tests and throw if any of them fail", async () => {
+    it("should return an error result when any test fails", async () => {
```
alcuadrado, line 182: This kind of test is now so much better

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -88,14 +87,26 @@ const runAllTests: NewTaskActionFunction<TestActionArguments> = async (
 
     if (subtask.options.has("testSummaryIndex")) {
       args.testSummaryIndex = failureIndex;
+    }
+
+    const subtaskResult = await subtask.run(args);
+
+    if (isResult(subtaskResult, isTestSummary, isTestSummary)) {
```
alcuadrado: I think this could be done backwards compatible by doing something like:

```ts
const summary = isResult(subtaskResult, isTestSummary, isTestSummary)
  ? subtaskResult.success
        ? subtaskResult.value
        : subtaskResult.error
  : isSummary(subtaskResult) ? subtaskResult : undefined; 
 
if (summary !== undefined) {
   // ...
}
```

alcuadrado: The reason to do this is that while we are bumping the peer dependency, that ensures that a new version of the plugin is running with HH >= certain version, but not the other way round. You can still only upgrade HH.

v-next/hardhat/src/types/test.ts
alcuadrado: Note that this is also affected by this comment: https://github.com/NomicFoundation/hardhat/pull/8015/changes#r2874509690 — As the backwards compatibility path should work with optional values.

alcuadrado (changes requested):
I think there are two small improvements that could be made, but in general looks good

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/8015/commits/ec372f6e2c9d3ce1a082946c758fb503b56c2171

v-next/hardhat/src/types/test.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/8015/commits/ec372f6e2c9d3ce1a082946c758fb503b56c2171

v-next/hardhat/src/internal/builtin-plugins/test/task-action.ts
```diff
@@ -24,14 +29,27 @@ interface TestActionArguments {
   verbosity: number;
 }
 
-function isTestSummary(value: unknown): value is TestSummary {
-  return isObject(value);
+// Old plugins may only return { failed, passed } without skipped/todo,
+// so we accept a partial shape and fill defaults in the coordinator.
+interface PartialTestSummary extends Omit<TestSummary, "skipped" | "todo"> {
```
schaable, line 34: Older versions of `hardhat-mocha` didn't include `skipped` and `todo` in `TestSummary`. To keep these versions compatible, we make the fields optional and default them on lines 125-129:

```typescript
      testSummaries[summaryId] = {
        skipped: 0,
        todo: 0,
        ...summary,
      };
```

alcuadrado: good catch

alcuadrado (approved):

PR #8021, opened 2026-03-02
Port the BannerManager to HH3

alcuadrado:


v-next/hardhat/src/internal/cli/banner-manager.ts
```diff
@@ -0,0 +1,236 @@
...
+
+    try {
+      const response = await getRequest(
+        BANNER_CONFIG_URL,
+        {
+          abortSignal:
+            timeout !== undefined ? AbortSignal.timeout(timeout) : undefined,
+        },
+        this.#dispatcher,
+      );
```
schaable, line 142: The third param also allows passing dispatcher options, including the `timeout`. Since `this.#dispatcher` is only used for testing, a simpler approach would be to pass the `timeout` when the dispatcher is `undefined`. I'm not sure we've tested the timeout when passed this way, so I'll leave it up to you whether to keep the original approach or use this one.

```suggestion
      const response = await getRequest(
        BANNER_CONFIG_URL,
        undefined,
        this.#dispatcher ?? { timeout },
      );
```

v-next/hardhat/src/internal/cli/banner-manager.ts
```diff
@@ -0,0 +1,236 @@
...
+      log(
+        `Error requesting banner config: ${
+          error instanceof Error ? error.message : JSON.stringify(error)
+        }`,
+      );
+    }
+  }
+
+  #isBannerConfig(value: unknown): value is BannerConfig {
+    if (typeof value !== "object" || value === null || Array.isArray(value)) {
```
schaable: let's use isObject

schaable (approved):
LGTM. I left two minor comments, feel free to apply them if you agree.

PR #8028, opened 2026-03-04
Fix spell checking workflow

alcuadrado:
This PR fixes the spell checking workflow, as it had two problems:

1. It wasn't reproducible locally, so you could push a PR, and only get an error on Github.
2. It wasn't really spell checking everything.

To fix it, I switch the github action from a cspell one, to use a new `pnpm spellcheck` root script, and properly configured `cspell`. This PR also fixes all the typos that uncovered.


schaable (approved):
LGTM

PR #8034, opened 2026-03-06
Add @nomicfoundation/hardhat-solx plugin

marianfe:
## Summary

Adds `@nomicfoundation/hardhat-solx` plugin that enables adding solx for test builds in Hardhat 3.
The plugin extends `SolidityCompilerType`, `SolidityCompilerConfig`  adding `solx` types, and hooks into `validateUserConfig`, `downloadCompilers` and `getCompiler` to orchestrate the compilation flow for solx. 

The plugin handles supported versions and supported EVM versions, and validates solx configs. 
It requires adding a `solx` build profile that can be used when running build or tests. 

Downloads from `solx-releases-mirror.hardhat.org` with retry logic matching the solc downloader pattern

Built on top of #8008 and #8009.


packages/hardhat-errors/src/descriptors.ts
```diff
@@ -2960,4 +2973,26 @@ Check the error message for more details and verify your foundry.toml configurat
       },
     },
   },
+  HARDHAT_SOLX: {
+    GENERAL: {
+      UNSUPPORTED_PLATFORM: {
+        number: 110000,
+        messageTemplate: `solx is not available for the current platform ({platform}/{arch}).
+
+solx supports: linux/x64, linux/arm64, darwin (macOS), windows/x64.`,
```
alcuadrado, line 3085: Are they universal binaries? Or which platform?

alcuadrado: self-answer: they are universal binaries

v-next/hardhat-solx/src/internal/downloader.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * Returns the deterministic path where a solx binary for the given version
+ * would be cached. This is a pure function — it does not check whether the
+ * binary exists on disk.
+ */
+export async function getSolxBinaryPath(solxVersion: string): Promise<string> {
+  const assetName = getSolxAssetName(solxVersion);
+  const globalCacheDir = await getCacheDir();
+  return path.join(globalCacheDir, `solx-v${solxVersion}`, assetName);
```
alcuadrado: What about this to keep it aligned with where we store the compilers for solc?

```suggestion
  return path.join(globalCacheDir, "compilers-v3", `solx-v${solxVersion}`, assetName);
```

v-next/hardhat-solx/src/internal/downloader.ts
```diff
@@ -0,0 +1,95 @@
...
+  // Return cached binary if it already exists
+  if (await exists(binaryPath)) {
+    log(`Using cached solx binary at ${binaryPath}`);
+    return binaryPath;
+  }
+
+  // Use a mutex to prevent concurrent downloads of the same version
+  const mutex = new MultiProcessMutex(`solx-download-${solxVersion}`);
+
+  return mutex.use(async () => {
```
alcuadrado: I would use the mutex in each retry, giving other processes the opportunity to proceed if needed, so that the mutex-timeout in other processes is less likely to happen.

v-next/hardhat-solx/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@nomicfoundation/hardhat-solx",
+  "version": "0.1.0",
```
alcuadrado:
```suggestion
  "version": "2.0.0",
```

so that the bump of changeset turns it into `3.0.0`, the first version that we use for Hardhat 3 plugins.

alcuadrado: Unless we want to signal that this is some sort of alpha/testing version

.changeset/warm-solx-plugin.md
```diff
@@ -0,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-solx": minor
```
alcuadrado: Depending on the answer to https://github.com/NomicFoundation/hardhat/pull/8034/changes#r2902296959 this may need to become

```suggestion
"@nomicfoundation/hardhat-solx": major
```

.changeset/warm-solx-plugin.md
```diff
@@ -0,0 +1,6 @@
...
+"@nomicfoundation/hardhat-errors": patch
+---
+
+Add @nomicfoundation/hardhat-solx plugin for solx compiler support in test builds
```
alcuadrado:
```suggestion
Add @nomicfoundation/hardhat-solx plugin for solx compiler support in test builds ([#8034]()https://github.com/NomicFoundation/hardhat/pull/8034)
```

We normally add this, but I think it can be automatic. I'll send a PR about it.

v-next/hardhat-solx/README.md
```diff
@@ -0,0 +1,67 @@
...
+  },
+});
+```
+
+## Usage
+
+Run tests using the solx-powered test profile:
+
+```bash
+hardhat test --buildProfile test
```
alcuadrado:
```suggestion
hardhat test --build-profile test
```

packages/hardhat-solx/src/internal/solx-compiler.ts
```diff
@@ -0,0 +1,46 @@
+import type {
+  Compiler,
+  CompilerInput,
+  CompilerOutput,
+} from "hardhat/types/solidity";
+
+import { spawnCompile as defaultSpawnCompile } from "hardhat/internal/solidity";
+
+export class SolxCompiler implements Compiler {
+  public readonly version: string;
```
alcuadrado, line 10: I thought this may accidentally be used as the "Solidity" version somewhere. But I just checked, and it isn't. This makes sense, as the language version is normally obtained from the config.

packages/hardhat-solx/src/internal/solx-compiler.ts
```diff
@@ -0,0 +1,46 @@
...
+  readonly #spawnCompile: typeof defaultSpawnCompile;
+
+  constructor(
+    solxVersion: string,
+    compilerPath: string,
+    extraSettings: Record<string, unknown> = {},
+    spawnCompile: typeof defaultSpawnCompile = defaultSpawnCompile,
+  ) {
+    this.version = solxVersion;
+    this.longVersion = `${solxVersion}+solx`;
```
alcuadrado, line 25: So no commits. I think this is fine for `solx`, but we'd have to improve it for the next compiler.

packages/hardhat-solx/src/internal/hook-handlers/solidity.ts
```diff
@@ -0,0 +1,73 @@
...
+      return next(context, compilerConfig);
+    }
+
+    const solxVersion = SOLIDITY_TO_SOLX_VERSION_MAP[compilerConfig.version];
+    assertHardhatInvariant(
+      solxVersion !== undefined,
+      `No solx version mapping for Solidity ${compilerConfig.version} — this should have been caught by config validation`,
+    );
+
+    const binaryPath = await getSolxBinaryPath(solxVersion);
```
alcuadrado, line 121: I think this should validate that it exists.

packages/hardhat-solx/src/internal/solx-compiler.ts
```diff
@@ -0,0 +1,46 @@
...
+  public async compile(input: CompilerInput): Promise<CompilerOutput> {
+    const args = ["--standard-json", "--no-import-callback"];
+
+    // Merge extra settings into the standard-json input's settings object.
+    // This is how solx-specific settings like LLVMOptimization are passed.
+    const modifiedInput: CompilerInput = {
+      ...input,
+      settings: {
+        ...input.settings,
+        ...this.#extraSettings,
```
alcuadrado, line 39: I don't know where to validate this: are the extra settings top-level like this?

v-next/hardhat-solx/test/integration.ts
```diff
@@ -0,0 +1,85 @@
...
+
+describe("hardhat-solx integration", () => {
+  const projectFolder = "simple";
+
+  useFixtureProject(projectFolder);
+
+  it("resolves plugin config through the HRE", async () => {
+    const baseHhConfig = (
+      await import(`./fixture-projects/${projectFolder}/hardhat.config.js`)
+    ).default;
```
alcuadrado: This also works:

```ts
const baseHhConfig = await importUserConfig(
  await resolveHardhatConfigPath(),
);
```

Those functions are in `hardhat/hre`.

v-next/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,427 @@
...
+  resolveUserConfig,
+});
+
+export async function validateUserConfig(
+  userConfig: HardhatUserConfig,
+): Promise<HardhatUserConfigValidationError[]> {
+  const errors = validateUserConfigZodType(userConfig, solxUserConfigType);
+
+  if (userConfig.solidity !== undefined) {
+    errors.push(...validateSolxCompilerFields(userConfig.solidity));
```
alcuadrado: I think this isn't necessary anymore, as with the changes introduced in `pr-8008-followup` it's now a type-error to include `preferWasm` in a `SolxCompilerUserConfig`.

The only places where a `preferWasm` may still be present is the top-level of a multi version config, or in a build profile. This commits validates that if `preferWasm` is `true` in either of them, all the compilers must be `solc`: https://github.com/NomicFoundation/hardhat/commit/558ec52b262de2328f68b8dc732fdd7b809b0534

v-next/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,427 @@
...
+  const solidity = userConfig.solidity;
+  if (
+    solidity === undefined ||
+    typeof solidity === "string" ||
+    Array.isArray(solidity) ||
+    !("profiles" in solidity) ||
+    solidity.profiles.production === undefined
+  ) {
+    log(
+      "Skipping production profile validation: no explicit production profile configured",
```
alcuadrado: This is incomplete, as `solx` could have been set in `default` and that would be inherited in `production`. Here's a counter example:

```ts
import { defineConfig, task } from "hardhat/config";
import hardhatSolx from "@nomicfoundation/hardhat-solx";

export default defineConfig({
  plugins: [hardhatSolx],
  tasks: [
    task("config", "Prints the solidity config")
      .setInlineAction(async (_, hre) => {
        console.log(JSON.stringify(hre.config.solidity, null, 2));
      })
      .build(),
  ],
  solidity: {
    version: "0.8.33",
    type: "solx",
  },
});
```

I think this need to be post-validated.

v-next/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,427 @@
...
+  }
+
+  return errors;
+}
+
+/**
+ * Check that at least one compiler version in the config is supported by solx.
+ * If the plugin is installed but no versions are compatible, emit an error.
+ */
+function validatePluginIsUseful(
```
alcuadrado: I understand that this is meant to be here for convenience of the users, but I think they will hate it. As small changes in the config will make your project unusable.

v-next/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,427 @@
...
+
+  return errors;
+}
+
+interface CompilerEntryInfo {
+  path: Array<string | number>;
+  entry: { type?: string; settings?: any; version?: string };
+}
+
+function iterateCompilerEntries(
```
alcuadrado: I think it's easier to maintain and understand if we express these validations using zod and keeping them aligned with the builtin plugin ones. It would look something like this:

```ts
// These zod types need to be aligned in shape with the one of the solidity
// builtin plugin, but don't need to revalidate everything.

const SUPPORTED_VERSIONS = Array.from(
  Object.keys(SOLIDITY_TO_SOLX_VERSION_MAP),
);

const supportedVersionsType = z
  .string()
  .refine((val) => SUPPORTED_VERSIONS.includes(val), {
    message: `Solx only support versions: ${SUPPORTED_VERSIONS.join(", ")}`,
  });

const supportedEvmVersionsType = z
  .string()
  .refine((val) => SUPPORTED_SOLX_EVM_VERSIONS.includes(val), {
    message: `Solx only EVM versions: ${SUPPORTED_SOLX_EVM_VERSIONS.join(", ")}`,
  });

const soxSolidityCompilerUserConfigType = z.object({
  settings: z
    .object({
      version: supportedVersionsType,
      evmVersion: supportedEvmVersionsType.optional(),
    })
    .optional(),
});

const solidityCompilerUserConfigType = conditionalUnionType(
  [
    [
      (data) => isObject(data) && "type" in data && data.type === "solx",
      soxSolidityCompilerUserConfigType,
    ],
    [(_data) => true, z.any()],
  ],
  "Expected a valid compiler configuration",
);

const singleVersionSolidityUserConfigType = conditionalUnionType(
  [
    [
      (data) => isObject(data) && "type" in data && data.type === "solx",
      soxSolidityCompilerUserConfigType,
    ],
    [(_data) => true, z.any()],
  ],
  "Expected a valid single-version Solidity configuration",
);

const multiVersionSolidityUserConfigType = z.object({
  compilers: z.array(solidityCompilerUserConfigType).nonempty(),
  overrides: z.record(z.string(), solidityCompilerUserConfigType).optional(),
});

const singleVersionBuildProfileUserConfigType = conditionalUnionType(
  [
    [
      (data) => isObject(data) && "type" in data && data.type === "solx",
      soxSolidityCompilerUserConfigType,
    ],
    [(_data) => true, z.any()],
  ],
  "Expected a valid compiler configuration",
);

const multiVersionBuildProfileUserConfigType = z.object({
  compilers: z.array(solidityCompilerUserConfigType).nonempty(),
  overrides: z.record(z.string(), solidityCompilerUserConfigType).optional(),
});

const buildProfilesSolidityUserConfigType = z.object({
  profiles: z.record(
    z.string(),
    conditionalUnionType(
      [
        [
          (data) => isObject(data) && "version" in data,
          singleVersionBuildProfileUserConfigType,
        ],
        [
          (data) => isObject(data) && "compilers" in data,
          multiVersionBuildProfileUserConfigType,
        ],
      ],
      "Expected an object configuring one or more versions of Solidity",
    ),
  ),
});

const solidityUserConfigType = conditionalUnionType(
  [
    [
      (data) => isObject(data) && "version" in data,
      singleVersionSolidityUserConfigType,
    ],
    [
      (data) => isObject(data) && "compilers" in data,
      multiVersionSolidityUserConfigType,
    ],
    [
      (data) => isObject(data) && "profiles" in data,
      buildProfilesSolidityUserConfigType,
    ],
    [(_data) => true, z.any()],
  ],
  "Expected a version string, an array of version strings, or an object configuring one or more versions of Solidity or multiple build profiles",
);

const solxUserConfigType = z.object({
  solidity: solidityUserConfigType.optional(),
  solx: z
    .object({
      dangerouslyAllowSolxInProduction: z.boolean().optional(),
    })
    .optional(),
});
```

alcuadrado: Note that this validates less things than the existing code, as I already addessed some comments while doing it.

v-next/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,427 @@
...
+      ...compiler,
+      // Only set type: "solx" for versions we support
+      type:
+        compiler.version in SOLIDITY_TO_SOLX_VERSION_MAP
+          ? SOLX_COMPILER_TYPE
+          : compiler.type,
+      // Strip settings for solx entries (solx defaults are injected in SolxCompiler)
+      settings:
+        compiler.version in SOLIDITY_TO_SOLX_VERSION_MAP
+          ? {}
```
alcuadrado: We shouldn't strip the `settings.outputSelection`, those are important for Hardhat to work.

packages/hardhat-solx/test/hook-handlers/solidity.ts
```diff
@@ -0,0 +1,203 @@
...
+import type { SolidityCompilerConfig } from "hardhat/types/config";
+import type { CompilerInput, CompilerOutput } from "hardhat/types/solidity";
+
+import assert from "node:assert/strict";
+import { describe, it } from "node:test";
+
+import { assertRejectsWithHardhatError } from "@nomicfoundation/hardhat-test-utils";
+
+// Helper to create a compiler config
+function createSolidityCompilerConfig(
```
alcuadrado, line 14: these tests feel a bit too hacky/fragile. They feel like integration tests with mocks for the parts that were annoying to create. I think it's better to:

- Unit test the downloader hook handler
- Use `createHardhatRuntimeEnvironment` with a mock plugin in front of this solidity to act as `next`.

Apart from that, the tests themselved look good.

v-next/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,427 @@
...
+    nextResolveConfigurationVariable: ConfigurationVariableResolver,
+  ) => Promise<HardhatConfig>,
+): Promise<HardhatConfig> {
+  const resolvedConfig = await next(userConfig, resolveConfigurationVariable);
+
+  // Register "solx" as a known compiler type
+  resolvedConfig.solidity.registeredCompilerTypes.push(SOLX_COMPILER_TYPE);
+
+  // Add "test" build profile if not already present
+  addTestProfile(resolvedConfig);
```
alcuadrado: In general we'd do something like:

```ts
  return {
    ...resolvedConfig,
    solidity: {
      ...resolvedConfig.solidity,
      profiles: {
        ...resolvedConfig.solidity.profiles,
        test: resolveTestProfile(resolvedConfig.solidity),
      }
    },
    solx: resolveSolxConfig(userConfig.solx),
  };
```

intead of mutating the objects inside of a hook chain.

v-next/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,243 @@
...
+  const resolvedConfig = await next(userConfig, resolveConfigurationVariable);
+
+  return {
+    ...resolvedConfig,
+    solidity: {
+      ...resolvedConfig.solidity,
+      registeredCompilerTypes: [
+        ...resolvedConfig.solidity.registeredCompilerTypes,
+        SOLX_COMPILER_TYPE,
+      ],
```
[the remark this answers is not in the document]
marianfe: i can take a look at fixing this.

packages/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,243 @@
...
+  const errors: HardhatConfigValidationError[] = [];
+
+  // Check that the user defined a "solx" build profile
+  if (resolvedConfig.solidity.profiles.solx === undefined) {
+    errors.push({
+      path: ["solidity"],
+      message:
+        'The hardhat-solx plugin has been installed, but no "solx" build profile was found in the Solidity configuration. Please read the plugin documentation for information on how to create a "solx" build profile.',
+    });
+  }
```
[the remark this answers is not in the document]
marianfe: fixed pr description.

packages/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,243 @@
...
+} from "hardhat/types/hooks";
+
+import { isObject } from "@nomicfoundation/hardhat-utils/lang";
+import {
+  conditionalUnionType,
+  validateUserConfigZodType,
+} from "@nomicfoundation/hardhat-zod-utils";
+import debug from "debug";
+import { z } from "zod";
+
```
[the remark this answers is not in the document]
marianfe: this doesn't seem to apply to config.ts, hook handlers are already lazy loaded AFAICT

packages/hardhat-solx/test/hook-handlers/solidity.ts
```diff
@@ -0,0 +1,203 @@
...
+      // verify via the error that it tries and that path is not mutated.
+      // For a true unit test we'd mock downloadSolx, but for now just
+      // check the path isn't set before the download attempt.
+      const originalPath = configs[0].path;
+
+      try {
+        await hooks.downloadCompilers!(context, configs, true);
+      } catch {
+        // Expected — download fails in test environment
+      }
```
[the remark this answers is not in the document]
marianfe: will look at fixing this where it makes sense, to follow the same pattern like solc downloader tests.

v-next/hardhat-solx/src/internal/downloader.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      try {
+        await download(url, binaryPath);
+
+        // Set executable permission on Unix
+        if (process.platform !== "win32") {
+          await chmod(binaryPath, 0o755);
+        }
+
+        log(`Successfully downloaded solx ${solxVersion}`);
```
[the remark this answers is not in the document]
marianfe: will look at adding a .sha256 check, if the mirror supplies it.

packages/hardhat-solx/src/internal/downloader.ts
```diff
@@ -0,0 +1,108 @@
+import path from "node:path";
+
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
+import { ensureError } from "@nomicfoundation/hardhat-utils/error";
+import { chmod, exists } from "@nomicfoundation/hardhat-utils/fs";
+import { getCacheDir } from "@nomicfoundation/hardhat-utils/global-dir";
+import { download } from "@nomicfoundation/hardhat-utils/request";
+import { MultiProcessMutex } from "@nomicfoundation/hardhat-utils/synchronization";
+import debug from "debug";
```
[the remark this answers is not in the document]
marianfe: this doesn't seem to apply to downloader.ts, this is called by a hook that is already lazy loaded AFAICT

packages/hardhat-solx/src/internal/platform.ts
```diff
@@ -0,0 +1,39 @@
+import os from "node:os";
+
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
+
```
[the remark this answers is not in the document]
marianfe: this doesn't seem to apply to platform.ts, rather index.ts AFAICT

packages/hardhat-solx/src/internal/hook-handlers/solidity.ts
```diff
@@ -0,0 +1,78 @@
...
+    assertHardhatInvariant(
+      await exists(binaryPath),
+      `solx binary not found at ${binaryPath} — downloadCompilers should have been called first`,
+    );
+
+    log(
+      `Creating SolxCompiler for Solidity ${compilerConfig.version} (solx ${solxVersion}) at ${binaryPath}`,
+    );
+
+    return new SolxCompiler(solxVersion, binaryPath, DEFAULT_SOLX_SETTINGS);
```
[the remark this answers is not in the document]
marianfe: i can look into unifying this with the solc behavior, but I think it's lower value for solx.

packages/hardhat-solx/src/internal/hook-handlers/config.ts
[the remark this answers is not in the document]
alcuadrado: please ignore this comment, as those rules were wrong. I updated them in #8088

packages/hardhat-solx/src/internal/downloader.ts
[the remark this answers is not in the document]
alcuadrado: Same as in the other comment. Those rules were wrong, and were causing a performance hit. HH3 already partitions its dependency graph through the plugin system. Really few things need to be dynamically imported.

packages/hardhat-solx/src/internal/downloader.ts
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+
+  throw new HardhatError(
+    HardhatError.ERRORS.HARDHAT_SOLX.GENERAL.DOWNLOAD_FAILED,
+    {
+      version: solxVersion,
+      attempts: DOWNLOAD_RETRY_COUNT.toString(),
+      reason: lastError?.message ?? "unknown error",
+    },
```
alcuadrado, line 181:
```suggestion
    },
    lastError,
```

this keeps the `error.cause` chain, making errors easier to debug

v-next/hardhat-solx/src/internal/hook-handlers/solidity.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (compilerConfig.type !== SOLX_COMPILER_TYPE) {
+      return next(context, compilerConfig);
+    }
+
+    // Honor custom path — skip version lookup and download
+    if (compilerConfig.path !== undefined) {
+      assertHardhatInvariant(
+        await exists(compilerConfig.path),
+        `solx binary not found at ${compilerConfig.path} — the configured path does not exist`,
+      );
```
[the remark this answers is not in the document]
alcuadrado: Agreed with this comment, this should throw a `HardhatError`, as it can be a user mistake, more than an assertion violation.

v-next/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,251 @@
...
+    settings: z
+      .object({
+        evmVersion: supportedEvmVersionsType.optional(),
+      })
+      .passthrough()
+      .optional(),
+  })
+  .passthrough()
+  .refine(
+    (data) => "path" in data || SUPPORTED_VERSIONS.includes(data.version),
```
alcuadrado:
```suggestion
    (data) => SUPPORTED_VERSIONS.includes(data.version),
```

I believe this is what you wanted, because later you have code to explicitly handle the custom path.

packages/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,251 @@
...
+      .optional(),
+  })
+  .passthrough()
+  .refine(
+    (data) => "path" in data || SUPPORTED_VERSIONS.includes(data.version),
+    {
+      message: `Solx only supports versions: ${SUPPORTED_VERSIONS.join(", ")}`,
+      path: ["version"],
+    },
+  );
```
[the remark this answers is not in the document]
alcuadrado: I thin this doesn't make a ton of sense

alcuadrado (approved):
I left some minor comments, but LGTM

packages/hardhat-solx/src/internal/hook-handlers/solidity.ts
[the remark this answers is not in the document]
alcuadrado: This comment is outdated, as your current version does honor the custom path

packages/hardhat-solx/src/internal/platform.ts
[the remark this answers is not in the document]
alcuadrado: ignore it, there are new rules for that

packages/hardhat-solx/src/internal/hook-handlers/config.ts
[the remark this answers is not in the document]
marianfe: I actually went ahead and implemented this suggestion, i'll explain below on your other comment why i think this makes sense.

packages/hardhat-solx/src/internal/downloader.ts
marianfe: fixed!

v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -3014,4 +3027,26 @@ Check the error message for more details and verify your foundry.toml configurat
...
+
+solx supports: linux/x64, linux/arm64, darwin (macOS), windows/x64.`,
+      },
+      DOWNLOAD_FAILED: {
+        number: 110001,
+        messageTemplate: `Failed to download solx {version} after {attempts} attempts: {reason}`,
+        websiteTitle: "solx download failed",
+        websiteDescription: `The solx compiler binary could not be downloaded from GitHub releases.
+
+Check your internet connection and verify that the requested solx version exists.`,
```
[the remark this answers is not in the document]
marianfe: fixed

v-next/hardhat-solx/src/internal/hook-handlers/solidity.ts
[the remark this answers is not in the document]
marianfe: fixed

v-next/hardhat-solx/src/internal/hook-handlers/config.ts
marianfe: I actually intentionally want us to not check supported versions when the compiler custom path is used , and i've added an additional check to make sure the path is not a string.Empty path . the reason is that for custom paths, users might be using a latest nightly build which supports a higher solidity version than the current solx plugin map. e.g. while 0.8.35 is in development in the nightly build and not released yet. i actually have unit tests specifically checking that this works, i think it is a legit use case (using a custom path to a compiler that supports a solidity version that's not officially in the supported map). thoughts?

PR #8053, opened 2026-03-11
Add `connectOnSingleton` to Hardhat Mocha

kanej:
Add a `connectOnSingleton` helper modeled after `connectOnBefore` to `hardhat-mocha`. This helper function supports using a singleton EDR instance across an entire Mocha test suite:

```ts
import type { Counter } from "../../types/ethers-contracts/contracts/Counter.js";

const { provider } = network.mocha.connectToSingleton();

describe("Example", function () { 
  it("gets the block number", async function () {
    const blockNumber = await provider.request({ method: "eth_blockNumber" });
  });
});
```

A network name can be passed (taking the config from the `hardhat.config.ts`):

```ts
const connection = network.mocha.connectToSingleton("localhost");
```

A reduced version of the network params (no overrides) can also be passed:

```ts
const connection = network.mocha.connectToSingleton({ network: "localhost", chainType: "l1" });
```

## API

Signature `network.mocha.connectToSingleton`:

```ts
connectToSingleton<ChainTypeT extends ChainType | string = DefaultChainType>(
  networkOrParams?: SingletonConnectionParams<ChainTypeT> | string,
): NetworkConnection<ChainTypeT>;
```

Where `SingletonConnectionParams` is `{ network?: string; chainType?: ChainTypeT }` — a subset of `NetworkConnectionParams` without `override`.

## Implementation

The helper creates proxy network connection instances per network name and chainType. The proxy is initialized into a real connection in the first test to run using the proxy. There is no explicit cleanup, with network connections are cleaned up once the complete suite has run.

Parallel mode is supported, but the singleton will be scoped per worker negating much of the performance gain.

### DIFF
--- a/v-next/hardhat-errors/src/descriptors.ts
+++ b/v-next/hardhat-errors/src/descriptors.ts
@@ -230,10 +230,10 @@ export const ERROR_CATEGORIES: {
         max: 30099,
         websiteSubTitle: "General errors",
       },
-      CONNECT_ON_BEFORE: {
+      CONNECTION_PROXY: {
         min: 30100,
         max: 30199,
-        websiteSubTitle: "connectOnBefore errors",
+        websiteSubTitle: "Connection proxy errors",
       },
     },
   },
@@ -2130,28 +2130,28 @@ Please try again later.`,
           'You have run your tests twice programmatically and your project is an ESM project (you have `"type": "module"` in your `package.json`, or some of your files have the `.mjs` extension). This is not supported by Mocha yet (https://github.com/mochajs/mocha/issues/2706).',
       },
     },
-    CONNECT_ON_BEFORE: {
+    CONNECTION_PROXY: {
       USE_BEFORE_HOOK: {
         number: 30100,
         messageTemplate: `The network connection proxy can't be used before the \`before\` hook runs. Make sure you only use it inside \`it\`, \`before\`, \`beforeEach\`, etc.`,
         websiteTitle:
           "Used network connection before the Mocha `before` hook ran",
         websiteDescription:
-          "You tried to use a `connectOnBefore` network connection proxy before the Mocha `before` hook has run. Make sure you only use the value inside `it`, `before`, `beforeEach`, or other Mocha hooks.",
+          "You tried to use a network connection proxy (created by helpers like `connectOnBefore` or `connectToSingleton`) before the Mocha `before` hook has run. Make sure you only use the value inside `it`, `before`, `beforeEach`, or other Mocha hooks.",
       },
-      AWAIT_CONNECT_ON_BEFORE: {
+      AWAIT_CONNECTION_PROXY: {
         number: 30101,
-        messageTemplate: `\`connectOnBefore\` must not be awaited. It returns a synchronous proxy — remove the \`await\` keyword.`,
-        websiteTitle: "Awaited connectOnBefore",
+        messageTemplate: `The network connection proxy must not be awaited. It returns a synchronous proxy — remove the \`await\` keyword.`,
+        websiteTitle: "Awaited network connection proxy",
         websiteDescription:
-          "You used `await` on the network connection proxy returned by `connectOnBefore`. This function returns a synchronous proxy that is populated by a Mocha `before` hook. Remove the `await` keyword.",
+          "You used `await` on a network connection proxy. These functions return a synchronous proxy that is populated by a Mocha `before` hook. Remove the `await` keyword.",
       },
       UNSUPPORTED_OPERATION: {
         number: 30102,
         messageTemplate: `This operation is not supported on the network connection proxy.`,
         websiteTitle: "Unsupported proxy operation",
         websiteDescription:
-          "You attempted an operation (such as Object.getPrototypeOf, Object.defineProperty, delete, new, or Object.freeze) on a connectOnBefore network connection proxy. These operations are not supported. Use standard property access and assignment instead.",
+          "You attempted an operation (such as Object.getPrototypeOf, Object.defineProperty, delete, new, or Object.freeze) on a network connection proxy. These operations are not supported. Use standard property access and assignment instead.",
       },
     },
   },
--- a/v-next/hardhat-mocha/src/connect-on-before/create-connect-to-singleton.ts
+++ b/v-next/hardhat-mocha/src/connect-on-before/create-connect-to-singleton.ts
@@ -1,0 +1,117 @@
+import type { SingletonConnectionParams } from "../type-extensions.js";
+import type {
+  ChainType,
+  DefaultChainType,
+  NetworkConnection,
+  NetworkManager,
+} from "hardhat/types/network";
+
+import { createNetworkConnectionProxy } from "./create-network-connection-proxy.js";
+
+interface SingletonEntry {
+  proxy: NetworkConnection;
+  connectPromise: Promise<void> | undefined;
+  resolved: NetworkConnection | undefined;
+}
+
+const DEFAULT_NETWORK_NAME = "default";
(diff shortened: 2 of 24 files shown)


v-next/hardhat-errors/src/descriptors.ts
```diff
@@ -230,10 +230,10 @@ export const ERROR_CATEGORIES: {
         max: 30099,
         websiteSubTitle: "General errors",
       },
-      CONNECT_ON_BEFORE: {
+      CONNECTION_PROXY: {
```
kanej, line 233: I switched the `connectOnBefore` errors to be more generic, as the proxy network connection is used in both `connectOnBefore` and `connectToSingleton`.

v-next/hardhat-mocha/src/connect-on-before/create-connect-to-singleton.ts
```diff
@@ -0,0 +1,117 @@
...
+import { createNetworkConnectionProxy } from "./create-network-connection-proxy.js";
+
+interface SingletonEntry {
+  proxy: NetworkConnection;
+  connectPromise: Promise<void> | undefined;
+  resolved: NetworkConnection | undefined;
+}
+
+const DEFAULT_NETWORK_NAME = "default";
+const DEFAULT_CHAIN_TYPE = "generic";
```
kanej, line 18: These are defined in core but not publicly exported.
I am adding them here so that when neither is defined the key to the singleton entry matches up with expectations during console.logging.

v-next/hardhat-mocha/src/connect-on-before/create-connect-to-singleton.ts
```diff
@@ -0,0 +1,117 @@
...
+        // to avoid race conditions with other suites' before hooks,
+        // hence the then that sets resolved.
+        entry.connectPromise = networkManager
+          .connect(params)
+          .then((connection) => {
+            /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions --
+            Connection is generically typed, and the entry is not */
+            entry.resolved = connection as NetworkConnection;
+          });
+      }
```
kanej, line 90: This two step is needed.
The proxy is returned immediately from `connectOnSingleton`, on the first before we trigger the connection resolution represented by this promise - but the promise setup has to be synchronous to avoid race conditions with other `before` hooks.
As soon as the promise resolves it sets the entry, which is then shared to the proxies.

v-next/hardhat-mocha/src/connect-on-before/create-connect-to-singleton.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      // wait for the connection to be resolved
+      await entry.connectPromise;
```
kanej, line 93: This is either the first before to trigger the connection resolution and awaits the just setup resolution promise.
Or it is the second or nth, in which case the resolution promise already exists and it can just await it.
Once the promise is resolved, the proxy has a reliable connection and the the before can finish.

v-next/hardhat-mocha/src/type-extensions.ts
```diff
@@ -31,16 +31,19 @@ declare module "hardhat/types/test" {
 
 import "hardhat/types/network";
 declare module "hardhat/types/network" {
-  export interface NetworkManager<
-    ChainTypeT extends ChainType | string = DefaultChainType,
-  > {
-    mocha: MochaNetworkHelpers<ChainTypeT>;
```
kanej, line 37: The generic here was wrong.
It has to be at function level, otherwise both `connectOnBefore` and `connectToSingleton` will reject chainTypes other than `generic`.

v-next/hardhat-mocha/src/type-extensions.ts
```diff
@@ -84,8 +87,48 @@ export interface MochaNetworkHelpers<
...
+   * @param networkOrParams - The network to connect to. Can be a network
+   * name string, a {@link SingletonConnectionParams} object for network and
+   * chain type, or omitted to connect to the default network.
+   * @returns A proxy to the {@link NetworkConnection} that resolves lazily
+   * when its properties are accessed inside a test or hook.
+   */
+  connectToSingleton<ChainTypeT extends ChainType | string = DefaultChainType>(
+    networkOrParams?: SingletonConnectionParams<ChainTypeT> | string,
+  ): NetworkConnection<ChainTypeT>;
 }
```
kanej, line 134: The interface is modelled on `network.connect` but bans overrides - as these can't be supported with a singleton per networkName/chainType.

v-next/hardhat-mocha/test/create-connect-to-singleton.ts
```diff
@@ -0,0 +1,164 @@
...
+import { afterEach, beforeEach, describe, it } from "node:test";
+
+import { assertRejects } from "@nomicfoundation/hardhat-test-utils";
+
+import { createConnectToSingleton } from "../src/connect-on-before/create-connect-to-singleton.js";
+
+// We capture the hooks registered by `connectToSingleton` so we can invoke
+// them manually in the test, rather than relying on a real Mocha runtime.
+let capturedBeforeHooks: Array<() => Promise<void>>;
+let originalBefore: typeof globalThis.before;
```
kanej, line 13: This is more magic than I would prefer, but given the importance of the method this is a set of invariants we should test.

v-next/hardhat-mocha/src/connect-on-before/create-connect-to-singleton.ts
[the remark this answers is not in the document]
kanej: We want default/generic if nothing is passed.

alcuadrado, line 18: This is `hre.config.defaultChainType`

alcuadrado: The default network is always "default".

alcuadrado: FYI: `hre.config.defaultChainType` always has the type `DefaultChainType`

v-next/hardhat-mocha/src/connect-on-before/create-connect-to-singleton.ts
```diff
@@ -0,0 +1,117 @@
...
+  DefaultChainType,
+  NetworkConnection,
+  NetworkManager,
+} from "hardhat/types/network";
+
+import { createNetworkConnectionProxy } from "./create-network-connection-proxy.js";
+
+interface SingletonEntry {
+  proxy: NetworkConnection;
+  connectPromise: Promise<void> | undefined;
```
alcuadrado, line 13: This is dangerous, holding a promise like this can lead to an unhandled promise rejection and kill the process.

v-next/hardhat-mocha/src/connect-on-before/create-connect-to-singleton.ts
alcuadrado: I'd use an `AsyncMutex` from `hardhat-utils` instead. That addresses the problem directly, simplifies the code, and avoid the problem of a dangling promise that can lead to an unhandled promise rejection.

I would add this explanation to the the mutex initialization, otherwise this knowledge would be lost.

v-next/hardhat-mocha/src/type-extensions.ts
alcuadrado: good catch

alcuadrado (changes requested):
I left some comments about how promises and race conditions are handled, the rest of it looks good

kanej:
We are taking a different approach and enhancing the core API.

PR #8066, opened 2026-03-17
fix: add validation for changeTokenBalance subject to throw meaningful error (#7316)

### LINKED ISSUE #7316: `TypeError: txResponse.wait is not a function` in `changeTokenBalance.ts:239:38`
### Version of Hardhat

3.0.1

### What happened?

When I run `npx hardhat test mocha` and use `changeTokenBalance` here: https://github.com/dl-solarity/solidity-lib/blob/hh-3/test/finance/vesting/Vesting.test.ts

it fails with the following error:

```
TypeError: txResponse.wait is not a function
    at getBalanceChange (/Users/inter/Desktop/Solarity/solidity-lib/node_modules/@nomicfoundation/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts:239:38)
    at async Promise.all (index 0)
```

1. Clone https://github.com/dl-solarity/solidity-lib.
2. Check out the `hh-3` branch.
3. Go to https://github.com/dl-solarity/solidity-lib/blob/hh-3/test/finance/vesting/Vesting.test.ts and add `.only` to the `describe` (delete `.skip`)
4. Run the `npx hardhat test mocha` command.

### Search terms

changeTokenBalance

aayushbaluni:
## Summary

Fixes #7316.

**Root cause:** The `changeTokenBalance` matcher expects the subject to be a transaction response (or a promise of one), but when users pass a BigInt from `balanceOf()` or other non-transaction values, the code calls `txResponse.wait()` without validation, causing `TypeError: txResponse.wait is not a function`.

**Fix:** Add a validation check in `getBalanceChange` to verify the resolved subject has a `wait` method before calling it, and throw a clear error with guidance on correct usage.

## Changes

- `v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts`: Add validation that the subject is a transaction response (has `wait` method) before calling `txResponse.wait()`, with a helpful error message
- `v-next/hardhat-ethers-chai-matchers/test/matchers/changeTokenBalance.ts`: Add 2 tests for `changeTokenBalance` and `changeTokenBalances` when subject is not a transaction response (e.g. BigInt from balanceOf)

## Testing

- Verified the fix addresses the reported scenario in #7316
- Added 2 tests covering: subject is BigInt from balanceOf for both changeTokenBalance and changeTokenBalances matchers
- Change is minimal and follows existing code patterns
- No unrelated changes included
- All 1603 tests pass

Made with [Cursor](https://cursor.com)

### DIFF
--- a/v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts
+++ b/v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts
@@ -236,7 +236,7 @@ export async function getBalanceChange(
   transaction: TransactionResponse | Promise<TransactionResponse>,
   token: Token,
   account: Addressable | string,
-  matcherName: string = CHANGE_TOKEN_BALANCE_MATCHER,
+  matcherName: string,
 ): Promise<bigint> {
   const txResponse = await transaction;
 
--- a/v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts
+++ b/v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts
@@ -17,7 +17,10 @@ import {
   CHANGE_TOKEN_BALANCE_MATCHER,
 } from "../constants.js";
 import { getAddressOf } from "../utils/account.js";
-import { assertIsNotNull } from "../utils/asserts.js";
+import {
+  assertIsNotNull,
+  assertIsTransactionResponse,
+} from "../utils/asserts.js";
 import { buildAssert } from "../utils/build-assert.js";
 import { preventAsyncMatcherChaining } from "../utils/prevent-chaining.js";
 
@@ -84,7 +87,13 @@ export function supportChangeTokenBalance(
       };
 
       const derivedPromise = Promise.all([
-        getBalanceChange(ethers, subject, token, account),
+        getBalanceChange(
+          ethers,
+          subject,
+          token,
+          account,
+          CHANGE_TOKEN_BALANCE_MATCHER,
+        ),
         getAddressOf(account),
         getTokenDescription(token),
       ]).then(checkBalanceChange);
@@ -123,7 +132,13 @@ export function supportChangeTokenBalance(
 
       const balanceChangesPromise = Promise.all(
         accounts.map((account) =>
-          getBalanceChange(ethers, subject, token, account),
+          getBalanceChange(
+            ethers,
+            subject,
+            token,
+            account,
+            CHANGE_TOKEN_BALANCES_MATCHER,
+          ),
         ),
       );
       const addressesPromise = Promise.all(accounts.map(getAddressOf));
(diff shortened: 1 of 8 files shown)


v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts
```diff
@@ -224,6 +224,18 @@ export async function getBalanceChange(
 ): Promise<bigint> {
   const txResponse = await transaction;
 
+  if (
+    txResponse === null ||
+    txResponse === undefined ||
+    typeof txResponse.wait !== "function"
+  ) {
+    chaiAssert.fail(
+      `The subject of "changeTokenBalance" must be a transaction response (or a promise of one). ` +
```
schaable: The error message mentions `changeTokenBalance`, but it also applies to `changeTokenBalances`. It should reference the correct matcher name. We could pass it into `getBalanceChange` when calling it.

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts
```diff
@@ -224,6 +224,18 @@ export async function getBalanceChange(
...
+        `Received something else (e.g. a BigInt from balanceOf()). ` +
```
schaable: This line is a bit too specific to your scenario. The error would trigger with any non-transaction value.

```suggestion
      `The subject of "changeTokenBalance" must be a transaction response (or a promise of one) ` +
        `but a different value was provided (e.g. the result of a read-only call).` +
```

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts
```diff
@@ -224,6 +224,18 @@ export async function getBalanceChange(
 ): Promise<bigint> {
   const txResponse = await transaction;
 
+  if (
+    txResponse === null ||
+    txResponse === undefined ||
+    typeof txResponse.wait !== "function"
```
schaable:
```suggestion
    !isObject(txResponse) ||
    typeof txResponse.wait !== "function"
```

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts
```diff
@@ -224,6 +224,18 @@ export async function getBalanceChange(
 ): Promise<bigint> {
   const txResponse = await transaction;
 
+  if (
```
schaable: The same validation should also be added to `changeEtherBalance` and `changeEtherBalances`, since they also call `txResponse.wait()`. We could extract it into a helper.

schaable (changes requested):

v-next/hardhat-ethers-chai-matchers/test/matchers/changeTokenBalance.ts
```diff
@@ -733,6 +733,21 @@ describe(
             );
           });
 
+          it("subject is not a transaction response (e.g. BigInt from balanceOf)", async () => {
```
schaable, line 736: `changeEtherBalance` and `changeEtherBalances` are missing this test.

v-next/hardhat-ethers-chai-matchers/src/internal/matchers/changeTokenBalance.ts
```diff
@@ -221,9 +236,12 @@ export async function getBalanceChange(
   transaction: TransactionResponse | Promise<TransactionResponse>,
   token: Token,
   account: Addressable | string,
+  matcherName: string = CHANGE_TOKEN_BALANCE_MATCHER,
```
schaable: I'd remove this default

schaable (changes requested):
Thanks for the changes. I left a few more suggestions, but after that this should be ready to merge.

schaable (approved):
I still need to do a round of manual testing but the code looks good. Thanks @aayushbaluni!

PR #8077, opened 2026-03-23
Cache genesis state

alcuadrado:
This PR caches the creation of the genesis state of EDR providers, and fixes a bug in `network.connect()` which was triggering the re-resolution of the config unnecessarily. The bug was preventing the cache from working.

### DIFF
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
@@ -30,6 +30,12 @@ const noForkingConfigCacheMarkerObject = {};
  * in practice this should rarely happen, except when using network config
  * overrides that generate the same config. These cases should be the minority
  * within a test suite.
+ *
+ * Note: the main reason that we don't use the entire NetworkConfig as cache
+ * key is that EDR initialization path recreates the NetworkConfig object to
+ * override some properties like `allowUnlimitedContractSize`, leading to a
+ * different NetworkConfig reference, despite keeping the references of most
+ * of its properties (including the accounts and forking config) the same.
  */
 const genesisStateAndAccountsCache: WeakMap<
   EdrNetworkAccountsConfig,
--- a/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
+++ b/v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
@@ -1,0 +1,147 @@
+import type {
+  EdrNetworkAccountsConfig,
+  EdrNetworkForkingConfig,
+} from "../../../../types/config.js";
+import type { ChainType } from "../../../../types/network.js";
+
+import {
+  l1GenesisState,
+  l1HardforkFromString,
+  opGenesisState,
+  opHardforkFromString,
+  type AccountOverride,
+} from "@nomicfoundation/edr";
+import { hexToBytes } from "ethereum-cryptography/utils";
+import { addr } from "micro-eth-signer/address";
+
+import { OPTIMISM_CHAIN_TYPE } from "../../../constants.js";
+
+import { hardhatAccountsToEdrOwnedAccounts } from "./utils/convert-to-edr.js";
+
+const noForkingConfigCacheMarkerObject = {};
+
+/**
+ * We cache the genesis state and owned accounts because computing them can be
+ * expensive, especially when run multiple times for the same configuration
+ * (e.g. when creating many EDR connections to the same network config).
+ *
+ * We use mostly references as cache keys, which means that we can miss some
+ * cache hits if an equivalent config is used with a different reference, but
+ * in practice this should rarely happen, except when using network config
+ * overrides that generate the same config. These cases should be the minority
+ * within a test suite.
+ */
+const genesisStateAndAccountsCache: WeakMap<
+  EdrNetworkAccountsConfig,
+  WeakMap<
+    EdrNetworkForkingConfig | typeof noForkingConfigCacheMarkerObject,
+    Map<
+      ChainType,
+      Map<
+        string,
+        {
+          genesisState: Map<Uint8Array, AccountOverride>;
+          ownedAccounts: Array<{ secretKey: string; balance: bigint }>;
+        }
+      >
+    >
+  >
+> = new WeakMap();
+
+export async function getGenesisStateAndOwnedAccounts(
+  accountsConfig: EdrNetworkAccountsConfig,
+  forkingConfig: EdrNetworkForkingConfig | undefined,
+  chainType: ChainType,
+  specId: string,
+): Promise<{
+  genesisState: Map<Uint8Array, AccountOverride>;
+  ownedAccounts: Array<{ secretKey: string; balance: bigint }>;
+}> {
+  const cached = genesisStateAndAccountsCache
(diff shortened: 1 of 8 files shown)


v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
```diff
@@ -0,0 +1,147 @@
...
+    const existingOverride = genesisState.get(account.address);
+    if (existingOverride !== undefined) {
+      // Favor the genesis state specified by the user
+      account.balance = account.balance ?? existingOverride.balance;
+      account.nonce = account.nonce ?? existingOverride.nonce;
+      account.code = account.code ?? existingOverride.code;
+      account.storage = account.storage ?? existingOverride.storage;
+    } else {
+      genesisState.set(account.address, account);
+    }
```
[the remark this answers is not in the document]
alcuadrado: This is intentional, and explained above.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -332,17 +332,32 @@ export class NetworkManagerImplementation implements NetworkManager {
    */
   async #resolveNetworkConfig<ChainTypeT extends ChainType | string>(
     resolvedNetworkName: string,
-    networkConfigOverride: NetworkConfigOverride | undefined = {},
+    networkConfigOverride: NetworkConfigOverride = {},
     resolvedChainType: ChainTypeT,
```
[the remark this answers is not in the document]
alcuadrado: This comment doesn't make sense. It has `= {}`, so passing `undefined` defaulted to `{}`.

For example, this test:

```ts
interface I {
  ii?: number;
}

function test(i: {} | undefined = {}) {
  console.log(i);
}

test();
test(undefined);
```

Prints
```
{}
{}
```

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
```diff
@@ -0,0 +1,147 @@
...
+        string,
+        {
+          genesisState: Map<Uint8Array, AccountOverride>;
+          ownedAccounts: Array<{ secretKey: string; balance: bigint }>;
+        }
+      >
+    >
+  >
+> = new WeakMap();
+
```
kanej, line 57: From my reading of this - we create a WeakMap based off the network accounts. This is in effect going to be scoped to the lifetime of the `hre`; in a test suite this will mainly be the global hre.
We weakmap to the forking config as well, but as we use a "no forking" marker object (the most common use case) which has module scope, for most cases this level will act just as a map.
The next levels are not objects but relevant keys.

My mental model here is that this cache is scoped to the `hre` object with some optimizations, but in general the genesis states will be cached for the lifetime of the HRE.

I think this makes sense.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
```diff
@@ -0,0 +1,147 @@
...
+  specId: string,
+): Promise<{
+  genesisState: Map<Uint8Array, AccountOverride>;
+  ownedAccounts: Array<{ secretKey: string; balance: bigint }>;
+}> {
+  const cached = genesisStateAndAccountsCache
+    .get(accountsConfig)
+    ?.get(forkingConfig ?? noForkingConfigCacheMarkerObject)
+    ?.get(chainType)
+    ?.get(specId);
```
kanej, line 73: I might be tempted to collapse the chainType layer and specId into a single key + map.
I don't feel strongly on this.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  if (cached !== undefined) {
+    return cached;
+  }
```
kanej, line 77: We could use the same mutex trick here that you used for the ContractDecoder to ensure there is only one calculation.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
```diff
@@ -332,17 +332,32 @@ export class NetworkManagerImplementation implements NetworkManager {
...
 
+    if (hasNoOverrides && isChainTypeDefault) {
+      return {
+        ...existingNetworkConfig,
+        /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions --
+          TypeScript can't follow this case, but we are just providing the
+          default */
+        chainType: resolvedChainType as ChainType,
+      };
+    }
```
kanej, line 359: For this PR I think this works.
That it forces an ugly test later that can't rely on reference equality for the network config - that causes me to pause.
I think this suggests that network config within `networkConfigs` should have the chain type set. Or am I missing a subtlety of our config resolution - and the undefined has to be preserved?

Claude suggested memoizing here - but I think that is too complicated to satisfy my desire to get the same networkConfig object.

kanej (approved):
I have left comments - but I don't think any are blockers.

I rebased this onto the `connectOnBefore` branch and ran the OpenZeppelin test suite with one connection per file and everything passed.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
alcuadrado: Your take is mostly correct.

The only case when it isn't is that when there are network config overrides the config gets re-resolved, so its lifetime is independent of the HRE. That means that this cache isn't hit for config overrides, because the references are different, and that those values can be GC'd after the network connection dies.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
alcuadrado: I also found this surprising as well. 

The reason `chainType` is not always resolved is that `undefined` lets you change it on `connect()`.

> Claude suggested memoizing here - but I think that is too complicated to satisfy my desire to get the same networkConfig object.

Not unreasonable to be honest, it would make the cache simpler. So the complexity would be moved somewhere else.

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
alcuadrado: That's a good point.

v-next/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
alcuadrado: This is not the only place where the network config is spread, so I added a better explanation in https://github.com/NomicFoundation/hardhat/pull/8077/changes/93d953fe260f9ae47c174b6dc9d54519377b8f39

v-next/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
alcuadrado: Added it in [db3ad1d](https://github.com/NomicFoundation/hardhat/pull/8077/commits/db3ad1da36f3cabcf1d846c60750bcb74496f928)

PR #8095, opened 2026-03-30
feat: add fallible methods to artifact manager

Wodann:

---

The existing `ArtifactManager` methods always throw, which can be expensive when failures are common, especially if the thrown error is not even used. The new methods avoid throwing and instead return `undefined`.

[The Openzeppelin Contracts](https://github.com/Amxx/openzeppelin-contracts) suffer from this problem, as they have a test suite that auto-generates test vectors for as many artifacts as possible (for more info, see [1](https://github.com/Amxx/openzeppelin-contracts/blob/00780408a72fd41efd3cf1c54addf74d305f037a/test/helpers/storage.js#L36) & [2](https://github.com/Amxx/openzeppelin-contracts/blob/00780408a72fd41efd3cf1c54addf74d305f037a/hardhat/hardhat-oz-contracts-helpers/hook-handlers/hre.ts#L12).

As seen in this profiling run, Hardhat's `editDistance` function takes the 4th largest amount of time. 

image

In total, the `#getFullyQualifiedName` function is consuming about 6.3% of the OZ test time:

image

There are several reasons for this becoming a hot path despite being error handling code:
1. OpenZeppelin causes 6006 calls to `#getSimilarStrings` due to missing artifacts
2. `#getSimilarStrings` is called with 764+ names to compare against
3. `editDistance` is called with names ranging in length from 3 to 39.
4. `editDistance` uses Levenshtein algorithm to determine the edit distance. The standard implementation has a time complexity of `O(m×n)`

After optimisation, we're almost spending no time (i.e. <0.1%) on this code path anymore:

image

The runtime of `npx hardhat test mocha --no-compile` went from 1:31.84 to 1:22.23, a reduction of 9.61s (-10.5%).

The accompanying change in OpenZeppelin Contracts can be found [here](https://github.com/Wodann/openzeppelin-contracts/commit/a0cb3d884555e88e312438fdf143f06776fedb23).

This PR also speeds up the `artifactExists` implementation shipped with Hardhat by not requiring the generation of the (later discarded) error for non-existent artifacts.

# Considerations

- I considered alternatives to the Levenshtein algorithm, like Jaro-Winkler Similarity which has `O(m+n)` time complexity; but given that OpenZeppelin never actually uses the generated error, I thought it would be a better alternative to introduce APIs that signal the possibility of the file not existing.
- An alternative optimisation would be to change the `#throwNotFoundError` to return a cheap intermediate error type that needs to be converted to a full, expensive-to-calculate Hardhat error (with suggestion) at a later time. That intermediate error type would contain the following information:
  ```ts
  {
    allFullyQualifiedNames: ReadonlySet<string>;
    bareNameToFullyQualifiedNameMap: Map<string, ReadonlySet<string>>;
  }
  ```
  Plugins (or users) that directly interface with `ArtifactManager` as a "library API" would then have control over whether to use the error or not. End-user application like Hardhat's CLI would then invoke the error conversion on the intermediate error type.

### DIFF
--- a/packages/hardhat-ethers/test/helpers/artifact-manager-mock.ts
+++ b/packages/hardhat-ethers/test/helpers/artifact-manager-mock.ts
@@ -49,9 +49,37 @@ export class MockArtifactManager implements ArtifactManager {
     return artifact;
   }
 
+  public async tryToReadArtifact<ContractNameT extends string>(
+    contractNameOrFullyQualifiedName: ContractNameT,
+  ): Promise<GetArtifactByName<ContractNameT> | undefined> {
+    const artifactFileName = this.#artifactsPaths.get(
+      contractNameOrFullyQualifiedName,
+    );
+
+    if (artifactFileName === undefined) {
+      return undefined;
+    }
+
+    const artifact = (await import(`./artifacts/${artifactFileName}.ts`))
+      .CONTRACT;
+
+    return artifact;
+  }
+
   public async getArtifactPath(
     _contractNameOrFullyQualifiedName: string,
   ): Promise<string> {
+    throw new HardhatError(
+      HardhatError.ERRORS.CORE.INTERNAL.NOT_IMPLEMENTED_ERROR,
+      {
+        message: "Not implemented in MockArtifactManager",
+      },
+    );
+  }
+
+  public async tryToGetArtifactPath(
+    contractNameOrFullyQualifiedName: string,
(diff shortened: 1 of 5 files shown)


Wodann:
@alcuadrado does #8122 make this PR obsolete?

alcuadrado:
Oops, sorry I forgot to mention that. Yes. We decided to go for the option that doesn't modify the API. Once again, amazing finding!

PR #8103, opened 2026-04-01
refactor: conditionally avoid network handlers

### LINKED ISSUE #8047: Reviewing await usage in the network stack
A potential speed up for JS/TS test suites is to reduce the number of `await`'s in our combined network/plugin paths.

kanej:
Add metadata to each subhandler in `onRequest` specifying which which RPC Methods are used. The handler will be skipped if the request's method does not match.

The goal here is avoid a cycle of async/awaits, instead we do sync checks.

The check is rerun within the handler as a guard and to force an engineer modifying the handler to consider the supported list. This justifies the no additional tests.

Resolves #8047

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
@@ -39,6 +39,15 @@ import { ChainId } from "../chain-id/chain-id.js";
 
 const EXTRA_ENTROPY = false;
 export class LocalAccountsHandler extends ChainId implements RequestHandler {
+  readonly #methods: ReadonlySet<string> = new Set([
+    "eth_accounts",
+    "eth_requestAccounts",
+    "eth_sign",
+    "personal_sign",
+    "eth_signTypedData_v4",
+    "eth_sendTransaction",
+  ]);
+
   readonly #localAccountsHexPrivateKeys: string[];
 
   #addressToPrivateKey: Map<string, Uint8Array> | undefined;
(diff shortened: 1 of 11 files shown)


packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
```diff
@@ -39,6 +39,15 @@ import { ChainId } from "../chain-id/chain-id.js";
...
 export class LocalAccountsHandler extends ChainId implements RequestHandler {
+  readonly #methods: ReadonlySet<string> = new Set([
+    "eth_accounts",
+    "eth_requestAccounts",
+    "eth_sign",
+    "personal_sign",
+    "eth_signTypedData_v4",
+    "eth_sendTransaction",
+  ]);
+
```
kanej, line 50: This list is based on #resolveRequest method - which returns null if no change is required. It is covered by tests, hence this list should be safe.

packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/accounts/local-accounts.ts
```diff
@@ -53,6 +62,27 @@ export class LocalAccountsHandler extends ChainId implements RequestHandler {
...
+    const response = await this.#resolveRequest(jsonRpcRequest);
+    if (response !== null) {
+      return response;
+    }
+
+    await this.#modifyRequest(jsonRpcRequest);
+
+    return jsonRpcRequest;
+  }
+
```
kanej, line 85: My apologies on this one, I moved a private method below it hence the appearence of total change.

The only actual change is the addition of an explicit `isSupportedMethod` check at the top, which is intended to replicate the implicit check wihtin `#resolveRequest`.

alcuadrado (approved):

packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/chain-id/chain-id-handler.ts
```diff
@@ -25,6 +25,13 @@ export class ChainIdValidatorHandler extends ChainId implements RequestHandler {
     this.#expectedChainId = expectedChainId;
   }
 
+  public isSupportedMethod(_jsonRpcRequest: JsonRpcRequest): boolean {
+    // This handler should be executed for every request,
+    // as it needs to validate the chain ID before any other
+    // handler processes the request.
```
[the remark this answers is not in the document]
kanej: I have pulled the conditional logic up into `isSupportedMethod`, and now use that to determine whether to short-circuit on in `handle`.

PR #8106, opened 2026-04-02
perf: avoid mutex if hook categories have been initialized

Wodann:

After prior optimisations, the OpenZeppelin contracts (OZC) repo is spending a significant time inside `HookManagerImplementation::#getPluginHooks`, its call to `exclusiveRun`, and idling (i.e. waiting for `Promise`s to be resolved) during this time.

image

Originally, we locked an `AsyncMutex` before asynchronously retrieving and potentially constructing hooks per category.

To reduce the time spent inside `exclusiveRun`, I attempted an optimisation that checks whether the requested category has already been cached, before locking the `AsyncMutex` and performing an `exclusiveRun` (also called a double-check pattern).

Since multiple `Promise`s can now be trying to acquire the `AsyncMutex` at the same time, this required a second step to avoid lock contention. Since each `Promise` is only interested in modifying a unique cached entry, I instead introduced an `AsyncMutex` per category.

For the OZC repo, this reduced the runtime of `npx hardhat test mocha --no-compile` from 1:31.84 to 1:27.49, a reduction of -4.35s (-4.7%).

Optimisations are reaching a point where the required changes are becoming larger and gains smaller. It's unclear whether the gain for OZC due to these changes would translate to a similar gain (or maybe even loss) in performance for small(er) Hardhat projects.

To that end, I'd recommend implementing benchmarks for multiple HH3 repos before deciding to merge this PR (or not).

### DIFF
--- a/packages/hardhat/src/internal/core/hook-manager.ts
+++ b/packages/hardhat/src/internal/core/hook-manager.ts
@@ -18,9 +18,27 @@ import { AsyncMutex } from "@nomicfoundation/hardhat-utils/synchronization";
 
 import { detectPluginNpmDependencyProblems } from "./plugins/detect-plugin-npm-dependency-problems.js";
 
-export class HookManagerImplementation implements HookManager {
-  readonly #mutex: AsyncMutex = new AsyncMutex();
+class StaticHookHandlers {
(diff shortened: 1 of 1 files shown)


alcuadrado:
Replaced by #8177. Once again, great findings!

PR #8120, opened 2026-04-10
Fix remappings duplication

alcuadrado:
There was a bug in the dependency graph, which could lead to remappings being duplicated, this PR fixes it.

Remappings are stored in the edges of the dependency graph, so that we can use only the necessary remappings when building a subgraph of the entire dependency graph. They are stored as a set within each edge, so they can't be duplicated in the same edge, even when merging graphs with the same edge.

The problem was that we weren't deduplicating across edges. For example, if a graph has an edge between `A` and `B` with remapping `r1`, and an edge between `C` and `D` with remapping `r1`, `r1` would be duplicated.

This PR fixes it by doing a cross-edge deduplication in the `getAllRemappings()` method of the graph.

The way this is structured, the first commit add some regression (integration style) tests, and actual unit tests of the functionality. The second commit fixes it, which is tiny.

### DIFF
--- a/.changeset/yellow-cycles-grow.md
+++ b/.changeset/yellow-cycles-grow.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix remappings duplication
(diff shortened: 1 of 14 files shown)


schaable (approved):
LGTM

PR #8121, opened 2026-04-10
Update the default `outputSelection` setting of solc to decrease the artifacts size

alcuadrado:


packages/hardhat/test/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -541,10 +541,16 @@ describe("CompilationJobImplementation", () => {
         "*": {
           "*": [
             "abi",
-            "evm.bytecode",
```
schaable, line 544: nit: we could use the constant here

packages/hardhat/test/internal/builtin-plugins/solidity/build-system/compilation-job.ts
```diff
@@ -560,7 +566,13 @@ describe("CompilationJobImplementation", () => {
           settings: {
             outputSelection: {
               "*": {
-                "*": ["storageLayout", "storageLayout", "abi", "abi"],
+                "*": [
+                  "storageLayout",
+                  "evm.bytecode.object",
```
schaable, line 571: is this correct?

schaable (approved):
Looks good. I left a nitpick and a question.

packages/hardhat/test/internal/builtin-plugins/solidity/build-system/compilation-job.ts
alcuadrado: What do you mean?

packages/hardhat/test/internal/builtin-plugins/solidity/build-system/compilation-job.ts
alcuadrado: I think it's better to have the list, unfortunately. Because any change to the test will break if we use the constant, because the compilation job dedups and sorts everything.

packages/hardhat/test/internal/builtin-plugins/solidity/build-system/compilation-job.ts
schaable: I'm not sure I follow why `outputSelection` now needs `evm.bytecode.object`. Was this due to an issue in the test? Not a blocker, just trying to better understand the reasoning.

alcuadrado: just to show that it's still present after the dedup and sort. i.e. we don't dedup inside selectors.

packages/hardhat/src/internal/builtin-plugins/solidity/constants.ts
```diff
@@ -0,0 +1,21 @@
...
+        "evm.bytecode.object",
+        "evm.bytecode.opcodes",
+        "evm.bytecode.sourceMap",
+        "evm.deployedBytecode.immutableReferences",
+        "evm.deployedBytecode.linkReferences",
+        "evm.deployedBytecode.object",
+        "evm.deployedBytecode.opcodes",
+        "evm.deployedBytecode.sourceMap",
+        "evm.methodIdentifiers",
+      ],
```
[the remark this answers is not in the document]
alcuadrado: good point, i added tests for this

PR #8122, opened 2026-04-10
Optimize `ArtifactsManager#artifactExists`

alcuadrado:
`ArtifactsManager#artifactExists` use to try to read the artifact, and return `false` when the read threw. That was slow, because the error that `readArtifact` generates includes all the similar contract names, which is a heavy operation. This implementation checks for the existence directly, making it much faster.


packages/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
```diff
@@ -81,22 +81,16 @@ export class ArtifactManagerImplementation implements ArtifactManager {
...
 
-      throw error;
+    if (this.#isFullyQualifiedName(contractNameOrFullyQualifiedName)) {
+      return allFullyQualifiedNames.has(contractNameOrFullyQualifiedName);
     }
+
+    return bareNameToFullyQualifiedNameMap.has(
+      contractNameOrFullyQualifiedName,
+    );
   }
```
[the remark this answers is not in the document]
alcuadrado: good point, done

packages/hardhat/src/types/artifacts.ts
```diff
@@ -92,8 +92,6 @@ export interface ArtifactManager {
    * This function doesn't throw if the name is not unique.
    *
    * @param contractNameOrFullyQualifiedName Contract or fully qualified name.
-   * @throws Throws an error if a non-unique contract name is used,
-   *   indicating which fully qualified names can be used instead.
```
schaable, line 96: Given that this function’s semantics have changed, it would be good to call it out here. When there are multiple contracts with the same name, `artifactExists("Foo")` returning `true` no longer guarantees that `readArtifact("Foo")` will succeed:

```typescript
if (await hre.artifacts.artifactExists("Foo")) {
  const a = await hre.artifacts.readArtifact("Foo"); // MULTIPLE_FOUND error
}
```

This could be confusing for users who expect `readArtifact` not to throw if `artifactExists` returned `true`.

schaable (approved):
Looks good, left a comment regarding documentation but I'm pre-approving it.

packages/hardhat/src/types/artifacts.ts
alcuadrado: I'll add a comment. That's a good point.

Note that I don't think the intentional semantics changed. I did a bit of and this method didn't used to have the `@throws`. I made a mistake in [this commit](https://github.com/NomicFoundation/hardhat/commit/19f9bc52bf99816e5a835d8e39837930b363a077) by adding every `@throws` possible, despite contradicting the description.

ChristopherDedominici (approved):

packages/hardhat/src/types/artifacts.ts
alcuadrado: Addressed in b1e4fcbd7

PR #8123, opened 2026-04-10
Hardhat version in buildinfo

alcuadrado:
https://github.com/NomicFoundation/hardhat-website/pull/252

### DIFF
--- a/.changeset/metal-garlics-study.md
+++ b/.changeset/metal-garlics-study.md
@@ -2,6 +2,6 @@
 "hardhat": patch
 ---
 
-Add a `toolVersionsInBuildInfo` setting to the Solidity Build Profiles, which is `true` by default in the `production` build profile. When enabled, the version of Hardhat is included in the Build Info files.
+Add a `toolVersionsInBuildInfo` setting to the Solidity config, which is `true` by default in the `production` build profile. When enabled, the version of Hardhat is included in the Build Info files.
 
-NOTE: This change can lead to build info ids changing, despite compilation output's bytecodes being identical, especially when using the `production` build profile.
+NOTE: This change can lead to build info ids changing despite the compilation output's bytecodes being identical, especially when using the `production` build profile.
--- a/.changeset/metal-garlics-study.md
+++ b/.changeset/metal-garlics-study.md
@@ -1,0 +1,7 @@
+---
+"hardhat": patch
+---
+
+Add a `toolVersionsInBuildInfo` setting to the Solidity Build Profiles, which is `true` by default in the `production` build profile. When enabled, the version of Hardhat is included in the Build Info files.
+
+NOTE: This change can lead to build info ids changing, despite compilation output's bytecodes being identical, especially when using the `production` build profile.
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
@@ -164,7 +164,7 @@ const singleVersionBuildProfileUserConfigType = conditionalUnionType(
         (!("type" in data) || data.type === undefined || data.type === "solc"),
       solcSolidityCompilerUserConfigType.extend({
         isolated: z.boolean().optional(),
-        includeBuildInfoVersions: z.boolean().optional(),
+        toolVersionsInBuildInfo: z.boolean().optional(),
         ...incompatibleVersionFields,
       }),
     ],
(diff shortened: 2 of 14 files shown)


packages/hardhat/src/internal/builtin-plugins/solidity/type-extensions.ts
```diff
@@ -173,15 +173,17 @@ declare module "../../../types/config.js" {
   export type SingleVersionBuildProfileUserConfig =
     SolidityCompilerUserConfig & {
       isolated?: boolean;
+      toolVersionsInBuildInfo?: boolean;
```
alcuadrado, line 190: Note: We only add this setting to the profiles, and not to every top-level config, as that seems unneeded.

packages/hardhat/src/types/solidity/compilation-job.ts
```diff
@@ -22,6 +23,12 @@ export interface CompilationJob {
    */
   solcLongVersion: string;
 
+  /**
+   * The versions to include in the build info, if any.
+   * When present, these versions influence the build id.
+   */
+  toolVersions?: ToolVersions;
```
alcuadrado, line 30: Added this to the `CompilationJob` because it influences the build info and the build id.

packages/hardhat/src/internal/builtin-plugins/solidity/type-extensions.ts
alcuadrado: this was a bad idea, I'm changing it to make it consistent with the other fields.

alcuadrado: done

packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -484,13 +488,15 @@ function resolveBuildProfileConfig(
...
+  toolVersionsInBuildInfo?: boolean,
 ): SolidityBuildProfileConfig {
   if ("version" in solidityConfig) {
     return {
       compilers: [resolveSolidityCompilerConfig(solidityConfig, production)],
       overrides: {},
       isolated: solidityConfig.isolated ?? production,
       preferWasm: solidityConfig.preferWasm ?? false,
+      toolVersionsInBuildInfo: toolVersionsInBuildInfo ?? production,
     };
```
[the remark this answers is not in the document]
alcuadrado: I decided to add it to `SingleVersionSolidityUserConfig` and `MultiVersionSolidityUserConfig`.

packages/hardhat/src/internal/builtin-plugins/solidity/type-extensions.ts
```diff
@@ -259,6 +275,7 @@ declare module "../../../types/config.js" {
   export interface SolidityBuildProfileConfig {
     isolated: boolean;
     preferWasm: boolean;
+    toolVersionsInBuildInfo?: boolean;
```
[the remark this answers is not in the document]
alcuadrado: Intentionally optional for backwards compatibility, I'll add a comment

.changeset/metal-garlics-study.md
```diff
@@ -0,0 +1,7 @@
+---
+"hardhat": patch
```
schaable, line 2: This should probably be a minor bump.

packages/hardhat/src/types/solidity/solidity-artifacts.ts
```diff
@@ -43,6 +51,14 @@ export interface SolidityBuildInfo {
    */
   readonly compilerType?: string;
 
+  /**
+   * Versions of the different tools used to create this build info.
+   *
+   * Not present if the build profile used to create this build info
+   * had `toolVersionsInBuildInfo` as `false`.
+   */
+  readonly versions?: ToolVersions;
```
schaable: Have you considered using `toolVersions` to match the type and other variables, and to make it more explicit?

schaable (approved):
Looks good. I left some minor suggestions. Pre-approving.

.changeset/metal-garlics-study.md
alcuadrado: Good point. The next release will be a patch because of the split compilation PRs, so I think this one doesn't matter that much though

packages/hardhat/src/types/solidity/solidity-artifacts.ts
alcuadrado: i guess it's a good point

PR #8128, opened 2026-04-12
Add `splitTestsCompilation` solidity setting (1): Configuration updates

alcuadrado:
This PR introduces the new config, it's validation, and resolution

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
@@ -81,6 +81,7 @@ const incompatibleCompilerFields = {
 const commonSolidityUserConfigFields = {
   isolated: z.boolean().optional(),
   npmFilesToBuild: z.array(z.string()).optional(),
+  splitTestsCompilation: z.boolean().optional(),
 };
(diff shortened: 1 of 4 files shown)


packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -209,6 +210,7 @@ const buildProfilesSolidityUserConfigType = z.object({
       "Expected an object configuring one or more versions of Solidity",
     ),
   ),
+  ...commonSolidityUserConfigFields,
```
schaable, line 213: Can I check that I understand this change correctly? You added `splitTestsCompilation` to `commonSolidityUserConfigFields`, but we had accidentally missed adding these fields to `buildProfilesSolidityUserConfigType` before, so you’ve added them now. Is that correct?

schaable (approved):
LGTM

packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
alcuadrado: correct

PR #8138, opened 2026-04-12
Add `splitTestsCompilation` solidity setting (11): closing PR

alcuadrado:
This PR has the last few changes to complete the effort, and do some cleanup work without having to rebase everything.

The PLUGIN_MIGRATION_GUIDE.md is committed so it can be reviewed, but will be delated.

The file `packages/hardhat/src/internal/to-be-deleted.ts` was added so that this PR triggers the full CI, but it should be deleted.

This PR is expected to pass all the checks.


packages/hardhat/src/internal/to-be-deleted.ts
```diff
@@ -0,0 +1 @@
+// Empty file to trigger the full CI
```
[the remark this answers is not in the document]
alcuadrado: this file is a placeholder and will be deleted

PLUGIN_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,151 @@
...
+- `hre.solidity.emitArtifacts(compilationJob, compilerOutput, { scope: "tests" })`
+- `hre.solidity.cleanupArtifacts(rootFiles, { scope: "tests" })`
+
+**Exception:** `hre.solidity.getArtifactsDirectory("tests")` does **not** throw. It returns the main artifacts path (same as `"contracts"`), since it is a read-only query with no side effects.
+
+### `getRootFilePaths({ scope: "contracts" })`
+
+When `splitTestsCompilation` is `false`, this returns **all** build roots — contract roots, test roots, and npm roots — together.
+
+When `splitTestsCompilation` is `true`, it returns contract roots only (unchanged).
```
schaable: The way this is phrased makes it sound like when `splitTestsCompilation` is `true`, npm roots are not returned, which isn’t the case.

```suggestion
When `splitTestsCompilation` is `false`, this returns **all** build roots — contract roots, test roots, and npm roots — together.

When `splitTestsCompilation` is `true`, it returns contract roots and npm roots only (unchanged).
```

PLUGIN_MIGRATION_GUIDE.md
```diff
@@ -0,0 +1,151 @@
...
+  contractRootPaths: string[];
+  testRootPaths: string[];
+}
+```
+
+The arrays reflect the roots actually built, partitioned using `getScope()`.
+
+### Plugin pattern for calling `build`
+
+If your plugin calls `build` and previously passed `noTests: true`, update it to branch on the config. When `splitTestsCompilation` is `false` there is no cheap "contracts-only" pass to fall back to — skipping tests would require an extra partial build — so the flag is only meaningful in split mode:
```
schaable: I think this explanation is a bit confusing. Claude suggestion is something like:

```suggestion
If your plugin calls `build` with `noTests: true`, update it to branch on the config. In unified mode (`splitTestsCompilation: false`), passing `noTests: true` turns the build into a partial build — cleanup and `artifacts.d.ts` regeneration are skipped. Since tests are compiled in the same pass as contracts anyway, there is no performance benefit to excluding them, and the plugin loses full-build semantics it likely depends on.
```

schaable (approved):

Issue #8041, opened 2026-03-09
Improve the filtering out of errors before reporting to Sentry


kanej:
_issue description replaced by @alcuadrado_

We are reporting too many errors to Sentry, and at the same time, we are missing opportunities to display better error messages to the users. In particular, related to migration issues.

To solve these problems we need to:
- [ ] Analyze the list of errors reported to sentry using claude or a throw-away tool created by it. Claude can download all the events using Sentry's api. The main result of the analysis should be classifying them according to the below explanation.
- [ ] Create general policies to classify them.
- [ ] Update the error reporting module so that it doesn't report errors based on policies generated above.
- [x] Update the list of error descriptors so most of them don't have `shouldBeReported: true`. Only errors that represent bugs should have it.
- [ ] Update the CLI's error displaying to show better errors for migration related classes.

## Error classification

I went through a portion of the errors, and I noticed that they tend to fall into these classes:

- `HardhatError`s that should have a descriptor without `shouldBeReported: true`. e.g. `HardhatError: HHE705: The network "hardhat" is not defined in your networks config....`
- Error related to Hardhat 3 migration e.g. `The requested module 'viem' does not provide an export named 'hardhat'`.
- Errors related to CJS to ESM migration e.g. `SyntaxError: The requested module 'viem' does not provide an export named 'hardhat' ...`
- Errors that we know should never be reported, as they are expected from normal workflows. An incomplete list of examples:
  - EDR Unknown error: `UnknownError: Received an unexpected status code from https:/<user-path>`
  - EDR `SolidityError`: `Error: VM Exception while processing transaction: reverted with custom error'
  - `ProviderError`: `ProviderError: Sender doesn't have enough funds to send tx. The max upfront cost is: 259805042880 and the sender's balance is: 0.`
  - Errors thrown by user code, so they shouldn't be reported: originating in user's tasks, task overrides, `mocha`/`node:test` tests, scripts, config loading, and from third-party plugins' hook handler, tasks, and task overrides.
 - Errors worth investigating.

Note that only the last class should be reported to sentry.

Potentially other classes may appear while going through the full list of errors.

## Policies

While doing a dry-run of this, I noticed that there are errors that can be discarded wither based on the class name of the error (i.e. `error.name`), or by matching part of their message. Again, other patterns may araise when looking at the whole list.

The errors thrown by the user code can be detected by inspecting the stack trace, as there are only a few places that yield control to the user code.

[...]

PR #8140, opened 2026-04-13
Error reporting cleanup PR #1

alcuadrado:
This is the first PR to tackle https://github.com/NomicFoundation/hardhat/issues/8041

It is not complete, but as it changes some linter rules, the sooner we get to merge it, the better.

It has these commits:

1. https://github.com/NomicFoundation/hardhat/commit/dafc4d3d81f75ffb7d0ed0fa14e20e74dc48f449: Remove `shouldBeReported` flags from error descriptors that shouldn't have it.
2. https://github.com/NomicFoundation/hardhat/commit/81730fa139615aa24bb702d229ae0c7bcd8a0420: Enable an eslint rule to force us to always use `return await promise` instead of `return promise`.
3. An autofix of the rule described above.
4. https://github.com/NomicFoundation/hardhat/commit/026b8cc07e4d7bc172afd3f8bc013694f1af93f7: Enable an eslint rule that forces every function that returns a promise to be `async`.
5. An autofix of the rule described above.
6. A fix for an unhandled promise.
7. Changests 
8. Small linter fixes

The reason the eslint rules are added is that v8, and hence Node, reports better error stacks if you await the promises and use async functions instead of sync functions that return promises. It doesn't skip the stack frames like that. It's not perfect, because things like `setTimeout()`/`.then()`/etc can still confuse it, but it's a big improvement. Some info here: https://v8.dev/docs/stack-trace-api#async-stack-traces

This may seem random with respect to the error reporting error, but it's important to be able to properly classify what should and should not be reported to sentry.

### DIFF
--- a/.changeset/async-improvements-hardhat-ethers-chai-matchers.md
+++ b/.changeset/async-improvements-hardhat-ethers-chai-matchers.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ethers-chai-matchers": patch
+---
+
+Improve async handling to get better error reports
--- a/.changeset/async-improvements-hardhat-ethers.md
+++ b/.changeset/async-improvements-hardhat-ethers.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ethers": patch
+---
+
+Improve async handling to get better error reports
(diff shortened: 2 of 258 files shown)


alcuadrado:
The comments about `on` and `once` are incorrect. Copilot doesn't realize that the ethers version of those methods return promises. Oh, ethers...

alcuadrado:
Replaced with https://github.com/NomicFoundation/hardhat/pull/8148 to avoid having a huge pr

PR #8141, opened 2026-04-14
Structured Solidity Test Cheatcode Errors

### LINKED ISSUE #7966: Display structured errors
EDR has updated the error messages returned from Solidity Test to provide more structured information around the reason why a particular cheatcode is missing or unsupported.

This has been implemented and released in EDR, but needs to be integrated into Hardhat. See the [design doc](https://www.notion.so/nomicfoundation/Structured-Solidity-Test-Cheatcode-Errors-2ea578cdeaf58090b278dc70222bdc17).

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/7966

Example of error:
image

### DIFF
--- a/.changeset/funny-socks-worry.md
+++ b/.changeset/funny-socks-worry.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Added support for EDR structured Solidity test cheatcode errors.
--- a/packages/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/solidity-stack-trace.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/solidity-stack-trace.ts
@@ -25,10 +25,12 @@ import type {
   InternalFunctionCallStackEntry,
   ContractCallRunOutOfGasError,
   CheatcodeErrorStackTraceEntry,
+  CheatcodeErrorDetails,
 } from "@nomicfoundation/edr";
 
 import {
   StackTraceEntryType,
+  CheatcodeErrorCode,
   stackTraceEntryTypeToString,
   FALLBACK_FUNCTION_NAME,
   RECEIVE_FUNCTION_NAME,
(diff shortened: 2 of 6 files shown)


alcuadrado (approved):
LGTM, but let's wait for @popescuoctavian to see if the output matches his expectation

popescuoctavian:
LGTM, thanks for implementing this! For some context, when a call targets the cheatcode address the inspector tries to ABI-decode the cheatcode address to known cheatcodes -- if that fails with `UnknownSelector` there are 2 cases:

1. Known upstream cheatcode => human-readable cheatcode signature is returned
2. Completely unknown selector (not part of Foundry ABI) => 4-byte selector is returned

I am assuming the `MissingCheatcode` test that produced the output from the screenshot falls under the 2nd case.

@ChristopherDedominici @alcuadrado

popescuoctavian (approved):

ChristopherDedominici:
> LGTM, thanks for implementing this! For some context, when a call targets the cheatcode address the inspector tries to ABI-decode the cheatcode address to known cheatcodes -- if that fails with `UnknownSelector` there are 2 cases:
> 
> 1. Known upstream cheatcode => human-readable cheatcode signature is returned
> 2. Completely unknown selector (not part of Foundry ABI) => 4-byte selector is returned
> 
> I am assuming the `MissingCheatcode` test that produced the output from the screenshot falls under the 2nd case.
> 
> @ChristopherDedominici @alcuadrado

@popescuoctavian  Yes, that's correct, the `MissingCheatcode` case corresponds to case 2 (unknown selector). As you can see in the screenshot, EDR couldn't resolve it to a human-readable signature, so the message shows the raw 4-byte selector (0x7218e329). The key difference this PR introduces is that when EDR provides structured details, we use that to produce a clearer, Hardhat-specific message instead of the generic one. When details is undefined (e.g. older EDR versions), we fall back to the original raw message for backward compatibility.

popescuoctavian:
@ChristopherDedominici that's in line with what I had in mind when I added the structured errors (maximizing backward compatibility). Thanks!

PR #8142, opened 2026-04-14
Make node:test sequential by default

### LINKED ISSUE #8125: Make `node:test` sequential by default
Our `node:test` plugin runs in parallel by default, but that can have unexpected performance hits, and is inconsistent with the mocha plugin. We should switch it to be sequential by default.

In a follow up effort we'll add `--parallel` flags to both plugins.

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/8125

Docs do not need updateds in the `website` repo, as there are no references to `node:test` being executed in parallel.

Manual test ran in `example-project` to validate everything works as expected:
- pnpm hardhat test nodejs
- pnpm hardhat test nodejs --coverage 
- pnpm hardhat test nodejs --gasStats


alcuadrado:
@schaable can you review this one as you have more context about coverage and gas stats?

alcuadrado:
Also snapshots

packages/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -114,7 +114,7 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
     const nodeTestOptions: LastParameter<typeof run> = {
       files,
       only,
-      concurrency: true, // uses `os.availableParallelism() - 1`
```
alcuadrado, line 116: I think we also need `isolation: "none"` to match Mocha's behavior. See: https://nodejs.org/api/test.html#runoptions

alcuadrado: And that's what may impact the integrations the most. /cc @schaable

ChristopherDedominici: I'll add `isolation=none` and see if tests in the CI pass as a first quick dirty test, then I'll investigate more on the actual impact

packages/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -114,7 +114,8 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
...
+      concurrency: false,
+      isolation: "none",
     };
```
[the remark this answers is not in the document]
ChristopherDedominici: `isolation: "none"` is already exercised by the existing integration tests that spawn real Hardhat CLI processes.

.changeset/strange-books-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-node-test-runner": patch
+---
+
+Make `node:test` sequential by default.
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8142/changes/22a7093c7333732938c9c09bd9030fe88510ccdd

packages/hardhat-node-test-runner/test/helpers/run-hardhat.ts
```diff
@@ -0,0 +1,101 @@
...
+  child.stderr.on("data", (chunk: string) => {
+    stderr += chunk;
+  });
+
+  const exitCode = await new Promise<number | null>((resolve, reject) => {
+    child.once("error", reject);
+    child.once("close", (code) => {
+      resolve(code);
+    });
+  });
```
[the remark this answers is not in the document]
ChristopherDedominici: Won't fix. Node's node `--test` runner already enforces a default per-test timeout. If the child process hangs, the outer test will be killed by that mechanism. Adding a redundant timeout inside this helper would just duplicate that responsibility with extra cleanup complexity for no real benefit.

packages/hardhat-node-test-runner/src/task-action.ts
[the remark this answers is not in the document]
ChristopherDedominici: No longer valid

packages/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -88,33 +87,21 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
...
-    const testWorkerDone = new URL(
-      import.meta.resolve(
-        "@nomicfoundation/hardhat-node-test-runner/test-worker-done",
-      ),
-    );
-    imports.push(testWorkerDone.href);
-  }
+    hre.globalOptions.gasStatsJson !== undefined;
 
-  process.env.NODE_OPTIONS = imports
```
ChristopherDedominici, line 108: Since everything now is executed in the same process, there is no need to pass to subprocesses the info

packages/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -88,33 +87,21 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
...
+  }
 
   async function runTests(): Promise<TestSummary> {
     const nodeTestOptions: LastParameter<typeof run> = {
       files,
       only,
-      concurrency: true, // uses `os.availableParallelism() - 1`
+      concurrency: false,
+      isolation: "none",
     };
```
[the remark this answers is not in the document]
ChristopherDedominici: Adding a fixture that deterministically detects parallel vs sequential execution would be fairly complex for the value it provides. The `concurrency: false` flag is a single, straightforward line that's unlikely to be accidentally removed or changed

packages/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -192,6 +170,13 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
 
   const testResults = await runTests();
 
+  await hre.hooks.runHandlerChain(
```
ChristopherDedominici, line 172: Same approach as in mocha test runner

packages/hardhat-node-test-runner/src/task-action.ts
```diff
@@ -88,33 +87,12 @@ const testWithHardhat: NewTaskActionFunction<TestActionArguments> = async (
...
-  imports.push(tsx.href);
-
-  if (
-    hre.globalOptions.coverage === true ||
-    hre.globalOptions.gasStats === true ||
-    hre.globalOptions.gasStatsJson !== undefined
-  ) {
-    const testWorkerDone = new URL(
-      import.meta.resolve(
-        "@nomicfoundation/hardhat-node-test-runner/test-worker-done",
```
schaable, line 102: I believe you also need to remove the `/test-worker-done` export in the package.json and the `packages/hardhat-node-test-runner/src/test-worker-done.ts` as they were only used here.

schaable (approved):
LGTM. Pre-approving.

packages/hardhat-node-test-runner/src/task-action.ts
alcuadrado: let's not remove it, because older versions of the plugin will break

alcuadrado: oh, this is the plugin itself, not hardhat

schaable: Yes, it's used by the plugin itself so it should be safe

ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8142/changes/33e5b9184fc3bf8f40113be7708fad2851f9d108

PR #8168, opened 2026-04-17
Add `from` config tests and rename gas-config tests to network-config

schaable:
Adds test coverage for the `from` field in `HardhatEthersSigner.sendTransaction` and viem's `walletClient.sendTransaction`, and renames the `gas-config` test files to `network-config` in both the `hardhat-ethers` and `hardhat-viem` plugins now that they cover more than just gas configuration.

**hardhat-ethers** — new `from config` describe block covering:
- default signer sends from its own address
- non-first signer sends from its own address
- explicit `from` matching the signer is respected
- explicit `from` mismatching the signer throws `"from address mismatch"` (existing pre-PR behavior from `#sendUncheckedTransaction`)

**hardhat-viem** — new `from config` describe block covering:
- default wallet client sends from its own address
- non-first wallet client sends from its own address

No behavior changes — tests only.

### DIFF
--- a/packages/hardhat-ethers/src/internal/signers/signers.ts
+++ b/packages/hardhat-ethers/src/internal/signers/signers.ts
@@ -252,7 +252,10 @@ export class HardhatEthersSigner implements HardhatEthersSignerI {
 
     const promises: Array<Promise<void>> = [];
 
-    // Make sure the from matches the sender
+    // Mirror stock ethers: ENS-resolve `from` and assert it matches the
+    // signer. Hardhat can be pointed at mainnet/testnets via --network, where
+    // ENS works and users expect ethers-compatible `from` handling. Don't
+    // "simplify" this by letting the node handle the mismatch.
     if (resolvedTx.from !== null && resolvedTx.from !== undefined) {
       const _from = resolvedTx.from;
       promises.push(
--- a/packages/hardhat-ethers/test/network-config.ts
+++ b/packages/hardhat-ethers/test/network-config.ts
@@ -254,6 +254,26 @@ function defineNetworkConfigTests(initEthers: () => Promise<InitResult>) {
     });
   });
 
+  describe("from config", () => {
+    let ethers: HardhatEthers;
+
+    before(async () => {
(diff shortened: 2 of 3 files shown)


packages/hardhat-ethers/src/internal/signers/signers.ts
```diff
@@ -252,26 +251,7 @@ export class HardhatEthersSigner implements HardhatEthersSignerI {
...
-            "transaction",
-            tx,
-          );
-          resolvedTx.from = from;
-        })(),
-      );
-    } else {
-      resolvedTx.from = this.address;
-    }
+    resolvedTx.from = this.address;
```
alcuadrado: Shouldn't this be:

```
    if (resolvedTx.from === null || resolvedTx.from === undefined) {
      resolvedTx.from = this.address;
    }
```

So that by default it respects our config, but doesn't completely break ethers' behavior from the POV of a user?

alcuadrado (changes requested):

packages/hardhat-ethers/src/internal/signers/signers.ts
schaable: you're right! done in https://github.com/NomicFoundation/hardhat/pull/8168/commits/c69215c1f02a6c3d8054651284a6a7d676da889b

alcuadrado (approved):

packages/hardhat-ethers/src/internal/signers/signers.ts
```diff
@@ -252,24 +251,7 @@ export class HardhatEthersSigner implements HardhatEthersSignerI {
...
-            "transaction",
-            tx,
-          );
-          resolvedTx.from = from;
-        })(),
-      );
-    } else {
+    if (resolvedTx.from === null || resolvedTx.from === undefined) {
       resolvedTx.from = this.address;
     }
```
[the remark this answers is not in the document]
schaable: yeah pr description is wrong

alcuadrado (approved):

PR #8175, opened 2026-04-21
Restructure the imports of the network-manager

### LINKED ISSUE #8158: Restructure the imports of the `network-manager`.
The `network-manager` hook handlers load too many things. For example, they load the `edr-provider.ts` module, just to access a constant. 

We should restructure those files so that we only load the things that are strictly needed, when they are needed.

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/8158

# Benchmark

Benchmarked `pnpm hardhat` (no args) in a minimal empty project (a `hardhat.config.ts` containing just `export default {}`), recording wall-clock and EDR load probes.

Methodology: one `pnpm clean && pnpm build` per branch, then 5 measurement invocations; each invocation does 5 warmup + 30 timed runs and reports the mean/median/standard deviation over those 30 samples.

| metric                                       |     main | branch (with changes) |                  Δ |
| -------------------------------------------- | -------: | ----------------: | -----------------: |
| median wall-clock                            | 298.62 ms |       **279.37 ms** | **−19.25 ms** (−6.4%) |
| `@nomicfoundation/edr` native addon loaded   |     true |         **false** |        ✅ eliminated |
| `network-manager/edr/edr-provider.js` loaded |     true |         **false** |        ✅ eliminated |


packages/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
```diff
@@ -16,16 +16,13 @@ import {
...
-  resolveEdrNetwork,
-  resolveHttpNetwork,
-} from "../config-resolution.js";
-import { validateNetworkUserConfig } from "../type-validation.js";
 
 export default async (): Promise<Partial<ConfigHooks>> => ({
   extendUserConfig,
-  validateUserConfig: validateNetworkUserConfig,
+  validateUserConfig: async (userConfig) => {
+    const { validateNetworkUserConfig } = await import("../type-validation.js");
```
alcuadrado: https://github.com/NomicFoundation/hardhat/blob/main/docs/engineering-guidelines.md#gc3-top-level-imports-vs-dynamic-imports ⚠️

alcuadrado:
>  a tiny script does await import("…/dist/…/hook-handlers/config.js") in a fresh Node process and times it with performance.now()

This methodology isn't valid. We always run those config hooks, so it's incomplete. The best way to measure it is running `pnpm hardhat` in an empty project.

alcuadrado (changes requested):

packages/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/config.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8175/changes/53e2c573968f81b343ba9199b782a2fa15bbf488

packages/hardhat/src/internal/builtin-plugins/network-manager/edr/types/hardfork.ts
```diff
@@ -51,9 +45,9 @@ export function getHardforks(chainType: ChainType): string[] {
 }
 
 export function getCurrentHardfork(chainType: ChainType): string {
```
ChristopherDedominici, line 47: Avoids importing @nomicfoundation/edr solely to read the latest hardfork.

packages/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -65,61 +61,6 @@ import { printLine, replaceLastLine } from "./utils/logger.js";
 
 const log = debug("hardhat:core:hardhat-network:provider");
 
-export const EDR_NETWORK_DEFAULT_COINBASE =
```
ChristopherDedominici, line 68: Moved to file `packages/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-constants.ts` so callers can read these defaults without triggering the import of `@nomicfoundation/edr`

packages/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-constants.ts
```diff
@@ -0,0 +1,61 @@
+import type { EdrNetworkHDAccountsConfig } from "../../../../types/config.js";
```
ChristopherDedominici, line 1: See comment: https://github.com/NomicFoundation/hardhat/pull/8175/changes#r3131060214

packages/hardhat/src/internal/builtin-plugins/network-manager/edr/types/hardfork.ts
alcuadrado: that's great!

alcuadrado (approved):

PR #8177, opened 2026-04-21
`HookManager` optimizations

alcuadrado:
This PR extends @Wodann's finding and his PR #8106 to cache more aggressively in the `HookManager`.

Given the amount of caching that this PR adds, analyzing its concurrent behavior can be trickier than before. Luckily Claude came up with a testing pattern that's great, and easy to understand. It uses different promises to run a specific concurrent scenario, and test the results.

This PR also adds a 1-line change optimization that I found while testing it.

PS: I rebased this PR on top of #8179 to be able to continue making progress, but this is effectively independent.

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/hook-handlers/hre.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/hook-handlers/hre.ts
@@ -126,11 +126,11 @@ class LazySolidityBuildSystem implements SolidityBuildSystem {
   }
 
   async #getBuildSystem(): Promise<SolidityBuildSystem> {
-    const { SolidityBuildSystemImplementation } = await import(
-      "../build-system/build-system.js"
-    );
-
     if (this.#buildSystem === undefined) {
+      const { SolidityBuildSystemImplementation } = await import(
+        "../build-system/build-system.js"
+      );
+
       this.#buildSystem = new SolidityBuildSystemImplementation(
         this.#hooks,
         this.#options,
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/hook-handlers/hre.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/hook-handlers/hre.ts
@@ -21,7 +21,14 @@ import type {
   CompilerOutputError,
 } from "../../../../types/solidity/compiler-io.js";
 import type { SolidityBuildInfo } from "../../../../types/solidity.js";
-import type { SolidityBuildSystemOptions } from "../build-system/build-system.js";
+import type {
+  SolidityBuildSystemOptions,
+  SolidityBuildSystemImplementation as SolidityBuildSystemImplementationT,
+} from "../build-system/build-system.js";
+
+let SolidityBuildSystemImplementation:
+  | typeof SolidityBuildSystemImplementationT
+  | undefined;
 
 class LazySolidityBuildSystem implements SolidityBuildSystem {
   readonly #hooks: HookManager;
@@ -126,11 +133,13 @@ class LazySolidityBuildSystem implements SolidityBuildSystem {
   }
 
   async #getBuildSystem(): Promise<SolidityBuildSystem> {
-    if (this.#buildSystem === undefined) {
-      const { SolidityBuildSystemImplementation } = await import(
-        "../build-system/build-system.js"
-      );
+    if (SolidityBuildSystemImplementation === undefined) {
+      const buildSystemModule = await import("../build-system/build-system.js");
+      SolidityBuildSystemImplementation =
+        buildSystemModule.SolidityBuildSystemImplementation;
+    }
 
+    if (this.#buildSystem === undefined) {
       this.#buildSystem = new SolidityBuildSystemImplementation(
         this.#hooks,
         this.#options,
--- a/packages/hardhat/src/internal/core/hook-manager.ts
+++ b/packages/hardhat/src/internal/core/hook-manager.ts
@@ -370,7 +370,7 @@ export class HookManagerImplementation implements HookManager {
       hookName,
     );
 
-    const sequential = chained.slice().reverse();
+    const sequential = chained.toReversed();
 
     handlersByName.set(hookName as string, sequential);
 
--- a/packages/hardhat/src/internal/core/hook-manager.ts
+++ b/packages/hardhat/src/internal/core/hook-manager.ts
@@ -193,7 +193,7 @@ export class HookManagerImplementation implements HookManager {
     // handlerParams and the `next` closure, and call the default implementation
     // directly.
     if (handlers.length === 0) {
-      return defaultImplementation(...handlerParams) as any;
+      return (await defaultImplementation(...handlerParams)) as any;
     }
 
     const numberOfHandlers = handlers.length;
--- a/packages/hardhat/src/internal/core/hook-manager.ts
+++ b/packages/hardhat/src/internal/core/hook-manager.ts
@@ -96,6 +96,10 @@ export class HookManagerImplementation implements HookManager {
       for (const hookCategoryName of Object.keys(plugin.hookHandlers) as Array<
         keyof HardhatHooks
       >) {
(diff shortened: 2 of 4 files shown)


packages/hardhat/src/internal/core/hook-manager.ts
```diff
@@ -23,35 +23,90 @@ export class HookManagerImplementation implements HookManager {
...
+        let pluginsForCategory =
+          this.#pluginsByHookCategory.get(hookCategoryName);
+
+        if (pluginsForCategory === undefined) {
+          pluginsForCategory = [];
+          this.#pluginsByHookCategory.set(hookCategoryName, pluginsForCategory);
+        }
+
+        pluginsForCategory.push(plugin);
+      }
```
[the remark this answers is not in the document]
alcuadrado: addressed

packages/hardhat/src/internal/core/hook-manager.ts
```diff
@@ -96,10 +159,33 @@ export class HookManagerImplementation implements HookManager {
...
+    // Fast path for the common case of no registered handlers: skip building
+    // handlerParams and the `next` closure, and call the default implementation
+    // directly.
+    if (handlers.length === 0) {
+      if (hookCategoryName !== "config") {
+        assertHardhatInvariant(
+          this.#context !== undefined,
+          "Context must be set before running non-config hooks",
+        );
+
```
kanej: This check and the invariant are shared with the `handlerParams` below.

I think we can pull the handlerParams calculation before the "jump straight to default implementation" optimization, and leverage `handleParams` within the optimization.

packages/hardhat/src/internal/core/hook-manager.ts
```diff
@@ -220,14 +306,48 @@ export class HookManagerImplementation implements HookManager {
...
+    }
+
+    const cached = handlersByName.get(hookName as string);
+    if (cached !== undefined) {
+      return cached;
+    }
 
-    const dynamicHooks = await this.#getDynamicHooks(
+    const staticCategories =
+      await this.#getStaticHookHandlerCategories(hookCategoryName);
```
kanej, line 318: Should we consider taking the complexity overhead of splitting out the cache lookup (sync) and the static category resolution here?

packages/hardhat/src/internal/core/hook-manager.ts
```diff
@@ -237,62 +357,70 @@ export class HookManagerImplementation implements HookManager {
...
-
-          if (existingCategory !== undefined) {
-            return existingCategory as Partial<HardhatHooks[HookCategoryNameT]>;
-          }
+      if (recheck !== undefined) {
+        return recheck;
+      }
 
+      const resolved = await Promise.all(
+        plugins.map(async (plugin) => {
```
kanej: I would pull the core of the static handler resolution out into its own method so that `getStaticHookHandlerCategories` remains at one abstraction level; demonstrating the top level caching flow (including the recheck within the mutex).

packages/hardhat/test/internal/core/hook-manager.ts
```diff
@@ -1075,4 +1075,299 @@ describe("HookManager", () => {
...
+        let signalStarted!: () => void;
+        const started = new Promise<void>((resolve) => {
+          signalStarted = resolve;
+        });
+
+        let unblock!: () => void;
+        const gate = new Promise<void>((resolve) => {
+          unblock = resolve;
+        });
+
```
kanej, line 1271: Oh the trickery.

kanej (approved):

packages/hardhat/src/internal/builtin-plugins/solidity/hook-handlers/hre.ts
```diff
@@ -126,11 +126,11 @@ class LazySolidityBuildSystem implements SolidityBuildSystem {
   }
 
   async #getBuildSystem(): Promise<SolidityBuildSystem> {
-    const { SolidityBuildSystemImplementation } = await import(
-      "../build-system/build-system.js"
-    );
-
     if (this.#buildSystem === undefined) {
+      const { SolidityBuildSystemImplementation } = await import(
```
schaable: This is one of the places we need to keep the await import isolated from the initialization. See https://github.com/NomicFoundation/hardhat/pull/8183#discussion_r3126142732

packages/hardhat/src/internal/core/hook-manager.ts
```diff
@@ -96,10 +159,33 @@ export class HookManagerImplementation implements HookManager {
     params: InitialChainedHookParams<HookCategoryNameT, HookT>,
     defaultImplementation: LastParameter<HookT>,
   ): Promise<Awaited<Return<HardhatHooks[HookCategoryNameT][HookNameT]>>> {
-    const handlers = await this.#getHandlersInChainedRunningOrder(
-      hookCategoryName,
-      hookName,
-    );
+    // Synchronous fast path for already cached handlers
```
schaable:
```suggestion
    // Synchronous fast path for already cached handlers. Duplicates the 
    // check in #getHandlersInChainedRunningOrder to avoid a microtask 
    // tick on cache hit
```

Took me a bit to understand why this is duplicated in `#getHandlersInChainedRunningOrder`. I think it's worth expanding the comment. Or elaborating further:

```typescript
// Synchronous fast path for already cached handlers. This duplicates
// the check inside #getHandlersInChainedRunningOrder on purpose:
// calling that async method introduces a microtask tick even on a
// cache hit, whereas a direct Map lookup stays on the current tick.
// That tick matters here because runHandlerChain is on every hook's
// hot path, and this path pairs with the empty-handlers shortcut
// below to dispatch straight to defaultImplementation with no awaits.
```

packages/hardhat/src/internal/core/hook-manager.ts
```diff
@@ -237,62 +357,70 @@ export class HookManagerImplementation implements HookManager {
...
+      return cached;
+    }
+
+    const chained = await this.#getHandlersInChainedRunningOrder(
       hookCategoryName,
       hookName,
     );
 
-    return handlersInChainedOrder.reverse();
+    const sequential = chained.slice().reverse();
```
schaable: nit: 
```suggestion
    const sequential = chained.toReversed();
```

packages/hardhat/src/internal/core/hook-manager.ts
```diff
@@ -321,27 +449,53 @@ export class HookManagerImplementation implements HookManager {
...
 
-      return handler as HardhatHooks[HookCategoryNameT][HookNameT];
+      return resolved;
     });
   }
+
+  #invalidateResolvedHandlersCache<
+    HookCategoryNameT extends keyof HardhatHooks,
+  >(hookCategoryName: HookCategoryNameT) {
+    // Invalidation deletes the outer entry rather than clearing the inner
```
schaable, line 475: 🤯

packages/hardhat/src/internal/core/hook-manager.ts
alcuadrado: good point, I addressed it

packages/hardhat/src/internal/core/hook-manager.ts
alcuadrado: I just tried it and it didn't change much. Just made the code more complex.

packages/hardhat/src/internal/core/hook-manager.ts
alcuadrado: good point, i just did it

packages/hardhat/src/internal/builtin-plugins/solidity/hook-handlers/hre.ts
alcuadrado: I implemented it

packages/hardhat/src/internal/core/hook-manager.ts
alcuadrado: i like the proposal

schaable (approved):

PR #8180, opened 2026-04-21
Optimize `getAllFilesMatching` and `getAllDirectoriesMatching`

alcuadrado:
This PR optimizes the `hardhat-utils` functions `getAllFilesMatching` and `getAllDirectoriesMatching`, by decreasing the number of `syscall`s they have to do. This is done by using `fsPromises.readdir(dirFrom, { withFileTypes: true });`, which returns the the file type of each entry.

For filesystems where this isn't supported, the new implementations fallback to an added `lstat` per entry, which is the previous behavior.

On top of that, this PR optimizes the usage of these functions, by:
- Remove it's usage from `init.ts`
- Improve how `getAllFilesMatching` is used in `template.ts`
- Avoid a duplicated `getAllFilesMatching` when getting the root files in the build system.
- Stop using `getAllFilesMatching` to fetch the `build-info` files, as it's a flat directory.

This PR is based on #8177, but it's independent from it

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
@@ -119,6 +119,8 @@ export class ArtifactManagerImplementation implements ArtifactManager {
       return new Set();
     }
 
+    // The build-info directory is expected to be flat: every build-info file
+    // lives directly under it, so a non-recursive `readdir` is enough.
     const buildInfoIds = (await readdir(buildInfosDir))
       .filter(
         (entry) => entry.endsWith(".json") && !entry.endsWith(".output.json"),
--- a/packages/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
@@ -14,7 +14,7 @@ import {
 import {
   exists,
   getAllFilesMatching,
-  readdir,
+  readdirOrEmpty,
   readJsonFile,
 } from "@nomicfoundation/hardhat-utils/fs";
 
@@ -115,17 +115,14 @@ export class ArtifactManagerImplementation implements ArtifactManager {
   public async getAllBuildInfoIds(): Promise<ReadonlySet<string>> {
     const buildInfosDir = path.join(this.#artifactsPath, BUILD_INFO_DIR_NAME);
 
-    if (!(await exists(buildInfosDir))) {
-      return new Set();
-    }
-
     // The build-info directory is expected to be flat: every build-info file
     // lives directly under it, so a non-recursive `readdir` is enough.
-    const buildInfoIds = (await readdir(buildInfosDir))
-      .filter(
-        (entry) => entry.endsWith(".json") && !entry.endsWith(".output.json"),
-      )
-      .map((entry) => path.basename(entry, ".json"));
+    const buildInfoFiles = await readdirOrEmpty(buildInfosDir);
+
+    const jsonExtensionLength = ".json".length;
+    const buildInfoIds = buildInfoFiles
+      .filter((f) => f.endsWith(".json") && !f.endsWith(".output.json"))
+      .map((entry) => entry.slice(0, entry.length - jsonExtensionLength));
 
     return new Set(buildInfoIds);
   }
--- a/packages/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
@@ -14,6 +14,7 @@ import {
 import {
   exists,
   getAllFilesMatching,
+  readdir,
   readJsonFile,
 } from "@nomicfoundation/hardhat-utils/fs";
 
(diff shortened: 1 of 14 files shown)


packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -1102,10 +1121,13 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
-    const buildInfoFiles = await getAllFilesMatching(buildInfosDir, (f) =>
-      f.startsWith(buildInfosDir + path.sep),
-    );
+    // The build-info directory is expected to be flat: every build-info file
+    // lives directly under it, so a non-recursive `readdir` is enough.
+    const buildInfoFiles = !(await exists(buildInfosDir))
+      ? []
+      : (await readdir(buildInfosDir)).map((entry) =>
+          path.join(buildInfosDir, entry),
+        );
```
schaable: this can be simplified using `readdirOrEmpty`, making just one syscall

packages/hardhat/src/internal/cli/init/init.ts
```diff
@@ -435,17 +434,36 @@ export async function copyProjectFiles(
...
+        );
+
+        if (!(await exists(absoluteWorkspacePath))) {
+          return undefined;
+        }
+
+        // We ignore directories in this clash detection
+        if (await isDirectory(absoluteWorkspacePath)) {
+          return undefined;
+        }
```
schaable, line 457: We could add a `lstatOrUndefined` helper in the `hardhat-utils` to replace the `exists(absoluteWorkspacePath)` + `isDirectory(absoluteWorkspacePath)` pattern

packages/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
```diff
@@ -112,12 +113,21 @@ export class ArtifactManagerImplementation implements ArtifactManager {
...
-    );
+    const buildInfosDir = path.join(this.#artifactsPath, BUILD_INFO_DIR_NAME);
+
+    if (!(await exists(buildInfosDir))) {
+      return new Set();
+    }
+
+    // The build-info directory is expected to be flat: every build-info file
+    // lives directly under it, so a non-recursive `readdir` is enough.
+    const buildInfoIds = (await readdir(buildInfosDir))
```
schaable: this can be simplified using `readdirOrEmpty`, making just one syscall

packages/hardhat/src/internal/cli/init/template.ts
```diff
@@ -65,27 +65,26 @@ export async function getTemplates(
...
+            return false;
+          }
+
+          // .gitignore files are expected to be called gitignore in the templates
+          // because npm ignores .gitignore files during npm pack (see https://github.com/npm/npm/issues/3763)
+          if (path.basename(f) === ".gitignore") {
+            return false;
+          }
+          return true;
+        },
```
schaable, line 79: I'd preserve the comment

schaable (changes requested):
Looks good! There are some follow-up improvements we can make to reduce syscalls even further.

packages/hardhat/src/internal/builtin-plugins/artifacts/artifact-manager.ts
alcuadrado: great idea, i'll clean up and export that function

packages/hardhat/src/internal/cli/init/init.ts
alcuadrado: not sure if worthwhile, as it's only for `init`, right? not on a hot-path

packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
alcuadrado: agreed

packages/hardhat/src/internal/cli/init/template.ts
alcuadrado: but that's only a mention about glob, which has been released for years

alcuadrado: it feels like an outdated todo more than a comment

alcuadrado:
Added tests to `readdirOrEmpty`, exported it, and used it where i found it to make sense in `hardhat`

packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -176,23 +177,15 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
   ): Promise<string[]> {
     const scope = options.scope ?? "contracts";
-    const unified = !this.#options.solidityConfig.splitTestsCompilation;
 
     this.#ensureSplitCompilationModeIfTestsScope(scope);
 
+    const unified = !this.#options.solidityConfig.splitTestsCompilation;
+    const { localContractFiles, sourceTestFiles } =
+      await this.#getSoliditySourcesRootFilePaths();
+
```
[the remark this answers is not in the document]
alcuadrado: not an issue

packages/hardhat/src/internal/cli/init/init.ts
```diff
@@ -435,17 +434,36 @@ export async function copyProjectFiles(
...
+
+        return relativeWorkspacePath;
+      }),
+    )
+  ).filter((relativeWorkspacePath) => relativeWorkspacePath !== undefined);
+
+  const matchingRelativeWorkspacePathsSet = new Set(
+    matchingRelativeWorkspacePaths,
+  );
```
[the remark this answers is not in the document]
alcuadrado: not true

schaable (approved):

schaable:
Looks good. I added a commit which moves the `readdirWithFileTypesOrEmpty` and `isDirectoryDirentAware` helpers to `internal/fs.ts`

Issue #8166, opened 2026-04-17
Lazy load the `GasAnalyticsManager`


alcuadrado:
These `GasAnalyticsManager` is always loaded and not always used.

We should lazy load its actual implementation, following the pattern that we already use in places like the [network-manager's HRE hook handlers](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts).

I'm not entirely sure if that pattern can be cleanly applied, but I'm sure something can be done.

Issue #8165, opened 2026-04-17
Lazy load the `UserInterruptionManager`


alcuadrado:
These `UserInterruptionsManager` always loads some dependencies, and is rarely used.

We should lazy load its actual implementation, following the pattern that we already use in places like the [network-manager's HRE hook handlers](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts).

Issue #8157, opened 2026-04-17
Lazy load the `NetworkConnection#ignition` in `hardhat-ignition-viem` and `hardhat-ignition-ethers`


alcuadrado:
These plugins load Ignition the first time a `NetworkConnection` is created, even if Ignition isn't used.

They should lazy load the actual implementation of `NetworkConnection#ignition`, following the pattern that we already use in places like the [`network-manager`'s HRE hook handlers](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts).

PR #8183, opened 2026-04-22
Lazy-load HRE-created plugin managers

### LINKED ISSUE #8167: Lazy load the `CoverageManager`
These `CoverageManager ` is always loaded and not always used.

We should lazy load its actual implementation, following the pattern that we already use in places like the [network-manager's HRE hook handlers](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/hre.ts).

I'm not entirely sure if that pattern can be cleanly applied, but I'm sure something can be done.

schaable:
## Summary

Several builtin/plugin managers were constructed eagerly at HRE creation time, pulling in their entire dependency graph even on CLI paths that never used them (`hardhat compile`, `hardhat test` without the relevant flag, etc.). This PR defers their loading so users only pay the cost when the feature is actually exercised.

- **coverage** — `CoverageManagerImplementation` is now constructed only when `--coverage` is set.
- **gas-analytics** — `GasAnalyticsManagerImplementation` is now constructed only when `--gas-stats` or `--gas-stats-json` is set.
- **hardhat-ignition-ethers** — `EthersIgnitionHelperImpl` is wrapped behind `LazyEthersIgnitionHelper`; the impl only loads on the first `connection.ignition.deploy(...)`.
- **hardhat-ignition-viem** — `ViemIgnitionHelperImpl` is wrapped behind `LazyViemIgnitionHelper`; the impl only loads on the first `connection.ignition.deploy(...)`.
- **hardhat/interruptions** — `UserInterruptionManagerImplementation` is wrapped behind `LazyUserInterruptionManager`; the impl only load on the first call to `hre.interruptions.*`.

Closes https://github.com/NomicFoundation/hardhat/issues/8167, https://github.com/NomicFoundation/hardhat/issues/8166,  https://github.com/NomicFoundation/hardhat/issues/8157 and https://github.com/NomicFoundation/hardhat/issues/8165.


packages/hardhat-ignition-ethers/src/internal/hook-handlers/network.ts
```diff
@@ -1,9 +1,128 @@
...
+
+  async #getEthersIgnitionHelper(): Promise<
+    EthersIgnitionHelperImplT<ChainTypeT>
+  > {
+    // Note: `await import` must run BEFORE the instance cache check so that
+    // concurrent callers share a single microtask-dedupe point — otherwise
+    // each suspended caller re-enters the branch and constructs its own
+    // impl, so callers end up holding different impl instances and state,
+    // which can cause concurrency issues.
+    if (EthersIgnitionHelperImpl === undefined) {
```
alcuadrado, line 106: This is a great point, and we should check if we aren't doing it in other places

packages/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/hre.ts
```diff
@@ -1,26 +1,36 @@
 import type { HardhatRuntimeEnvironmentHooks } from "../../../../types/hooks.js";
+import type { CoverageManagerImplementation as CoverageManagerImplementationT } from "../coverage-manager.js";
 
-import { CoverageManagerImplementation } from "../coverage-manager.js";
 import { getCoveragePath, setCoverageManager } from "../helpers.js";
```
alcuadrado: Aren't the helpers importing mode things than needed? I think we can also lazy-load them to be safe

alcuadrado: i just checked, and they don't. But i guess if we load them lazily this code will be more robust in the future. Otherwise we may reintroduce an accidental import.

packages/hardhat/src/internal/builtin-plugins/gas-analytics/hook-handlers/hre.ts
```diff
@@ -1,29 +1,41 @@
 import type { HardhatRuntimeEnvironmentHooks } from "../../../../types/hooks.js";
+import type { GasAnalyticsManagerImplementation as GasAnalyticsManagerImplementationT } from "../gas-analytics-manager.js";
 
-import { GasAnalyticsManagerImplementation } from "../gas-analytics-manager.js";
 import { setGasAnalyticsManager } from "../helpers/accessors.js";
```
alcuadrado, line 4: Same here

packages/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/hre.ts
schaable: Not really, the helpers are very lightweight. If you want more safety, we can split them like in gas-analytics and have an `accessors.ts` file for the getter and setter, which is the only thing we need to import from here.

alcuadrado (approved):
Left two minor comments, but looks good to me

packages/hardhat/test/internal/builtin-plugins/gas-analytics/hook-handlers/hre.ts
```diff
@@ -0,0 +1,51 @@
...
+import { createHardhatRuntimeEnvironment } from "../../../../../src/hre.js";
+import { GasAnalyticsManagerImplementation } from "../../../../../src/internal/builtin-plugins/gas-analytics/gas-analytics-manager.js";
+import { getGasAnalyticsManager } from "../../../../../src/internal/builtin-plugins/gas-analytics/helpers/accessors.js";
+
+describe("gas-analytics/hook-handlers/hre — lazy-loading", () => {
+  describe("when --gas-stats is not set", () => {
+    it("does not install _gasAnalytics on the HRE", async () => {
+      const hre = await createHardhatRuntimeEnvironment({});
+
+      assert.equal("_gasAnalytics" in hre && hre._gasAnalytics, undefined);
```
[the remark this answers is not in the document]
schaable: `_gasAnalytics` is a class field on `HardhatRuntimeEnvironmentImplementation` so it's always `in` the instance

packages/hardhat/src/internal/builtin-plugins/gas-analytics/hook-handlers/hre.ts
schaable: Added a comment [here](https://github.com/NomicFoundation/hardhat/pull/8183/changes/34933c391d00c2121f4df062284ab55b1e00e835#diff-c472519f4dc57ee0f6e7e6d2f2c3cc1bf42e7029d1592afe666721bebded3d09)

packages/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/hre.ts
schaable: Split the file in the same way as in gas-analytics and added a comment to the accessors file to help prevent regressions. See https://github.com/NomicFoundation/hardhat/pull/8183/commits/34933c391d00c2121f4df062284ab55b1e00e835

packages/hardhat/src/internal/builtin-plugins/coverage/helpers/compat.ts
```diff
@@ -0,0 +1,37 @@
+import type { HardhatRuntimeEnvironment } from "../../../../types/hre.js";
```
alcuadrado, line 1: shouldn't this new file be in the package#exports?

schaable: No, it is re-exported from `packages/hardhat/src/internal/builtin-plugins/coverage/exports.ts`

alcuadrado: got it

PR #8185, opened 2026-04-23
Replace chalk with util.styleText

### LINKED ISSUE #8160: Replace `chalk` with `util.styleText`
We are currently using `chalk` to format the output we print. 

Node now provides [`util.styleText`](https://nodejs.org/api/util.html#utilstyletextformat-text-options) a native solution for it, so we should uninstall `chalk` from every package and use it instead.

We need to bump the minimum version of Hardhat to `v22.13.0` as well, as that's when it got stabilized. VSCode, Cursor and Antigravity already use newer versions.

ChristopherDedominici:
Replaced by: https://github.com/NomicFoundation/hardhat/pull/8195

Fixes: https://github.com/NomicFoundation/hardhat/issues/8160


packages/hardhat-node-test-reporter/integration-tests/fixture-tests/assert-test/result.txt
```diff
@@ -12,8 +12,8 @@
       [31m8) should match strings[39m
```
ChristopherDedominici, line 12: Comment for all the `txt` files:

`chalk` wraps each line with its own `[31m…[39m` pair. That's why the old snapshot had:
```
[31mAssertionError: …[39m
[31m[39m
[31m1 !== '1'[39m
[31m[39m
```

`util.styleText` emits the opening code once at the start and the closing code once at the very end, so the intermediate newlines/blank lines sit inside a single color span:
```
[31mAssertionError: …

1 !== '1'
[39m
```

packages/hardhat-node-test-reporter/integration-tests/fixture-tests/async-describe-test/result.txt
```diff
@@ -1,54 +1,54 @@
   describe.async
-    [90m[32m✔[39m[90m describe.async.it.only[39m
+    [90m[32m✔[90m describe.async.it.only[39m
```
ChristopherDedominici: `chalk` resets then re-opens the outer; `styleText` switches directly.

```
# before
[90m[32m✔[39m[90m text[39m    ← reset, then re-open gray

# after
[90m[32m✔[90m text[39m         ← direct switch to gray
```

packages/hardhat/src/internal/cli/init/init.ts
```diff
@@ -602,17 +604,22 @@ export async function installProjectDependencies(
...
   }
 }
 
 function showStarOnGitHubMessage() {
   console.log(
-    chalk.cyan("Give Hardhat a star on Github if you're enjoying it! ⭐️✨"),
+    styleText(
+      "cyan",
+      "Give Hardhat a star on Github if you're enjoying it! ⭐️✨",
+    ),
```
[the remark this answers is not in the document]
ChristopherDedominici: I think this was taken personally 😂

Issue #8164, opened 2026-04-17
Redefine how we import Type Extensions in all the plugins so that the imports get removed during compilation, while preserving the type-level extensions.


alcuadrado:
Hardhat plugins use [Type Extensions](https://hardhat.org/docs/plugin-development/explanations/type-extensions) to add functionality to existing types. 

A Type Extension needs to be reachable by TypeScript for it to take place. If it's in your own codebase, you don't need to do anything special, as in that situation TypeScript evaluates all your files.

When you are distributing a package through npm though, only the imported files of your package, and their transitive dependencies are loaded. If you place a `type-extensions.ts` file next to your plugin's `index.ts` file and don't import it in any way, it will be ignored.

Our current recommendation is for plugins to have a top-level 

```ts
import "./type-extension.ts";
```

 in their `index.ts`, which works to force TypeScript to apply the extension. 

Its drawback is that that import is also present in the compiled `index.js`, and run when the plugin is loaded without any added benefit.

Instead, we can use this pattern:

```ts
export type * from "./type-extensions.ts";
```

it still forces typescript to apply the type extensions, but it doesn't generate a runtime import.

For this pattern to work, is extremely important that the `./type-extensions.ts` file doesn't:

1. Produce any side effect
2. Export any type
3. Has a `export type * from "plugin-dependency";` for each plugin you depend on.

The new updated pattern for type extensions should then be:

- Add `export type * from "./type-extensions.ts";` to your plugin's `index.ts`
- Do not run any executable code in your `./type-extensions.ts` file. Only type-level code is valid.
- Do not export any type from your `type-extension.ts`
- Type-extensions don't need to import the modules they are augmenting. This was an incorrect way to force the `.ts` file to be treated as a module. Use `export {};`, or better yet [`"moduleDetection": "force"`](https://www.typescriptlang.org/tsconfig/#moduleDetection).
- Do not export any type from your `index.ts`, otherwise plugins depending on yours will accidentally re-export it.

To do this, we should:
- [ ] Update the plugin author documentation, explaining that they should use the new pattern instead.
- [ ] Update the https://github.com/NomicFoundation/hardhat3-plugin-template (`main` and `tutorial` branches)
- [ ] Update all of our plugins

kanej:
We have implemented multiple improvements to our type extension loading. There is follow on work, but that is captured in https://github.com/NomicFoundation/hardhat/issues/8320

PR #8191, opened 2026-04-24
Change type extensions handling

alcuadrado:
This is the first change to complete this: 

- https://github.com/NomicFoundation/hardhat/issues/8164

It reworked how we handle type-extensions preventing Hardhat from importing every single type files. It currently only imports the ones with `enum`s.

### DIFF
--- a/packages/hardhat/src/types/config.ts
+++ b/packages/hardhat/src/types/config.ts
@@ -1,6 +1,3 @@
-// NOTE: We export the built-in plugin types to load their type extensions
-export type * from "../internal/builtin-plugins/index.js";
-
 /**
  * A configuration variable to be fetched at runtime from
  * different sources, depending on the user's setup.
--- a/packages/hardhat/src/types/config.ts
+++ b/packages/hardhat/src/types/config.ts
@@ -1,6 +1,5 @@
-// NOTE: We import the builtin plugins in this module, so that their
-// type-extensions are loaded when the user imports `hardhat/types/config`.
-import "../internal/builtin-plugins/index.js";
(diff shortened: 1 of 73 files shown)


alcuadrado:
Added 'no docs needed", the docs chnges are in the plugin development guide, and tracked by an issue.

packages/hardhat/src/types/config.ts
```diff
@@ -1,6 +1,3 @@
-// NOTE: We export the built-in plugin types to load their type extensions
-export type * from "../internal/builtin-plugins/index.js";
```
kanej, line 2: This strikes me as a breaking change. Is it that we are intentionally wanting to break here before stable?

kanej: Is this safe because any usage is likely going to have pulled in other export endpoints?

kanej (approved):
I left a question, but I have been through these changes and I don't have anything to add.
My mental model is that this PR should not alter the footprint of the exported types.

packages/hardhat/src/types/config.ts
alcuadrado: technically, it is a breaking change. It's pretty hard to notice it.

You would need to:
1. Have a `hardhat.config.ts` file that doesn't import `hardhat/config`, but `hardhat/types/config`.
2. Not have a `tsconfig.json`, or not have a file that imports `hardhat`, `hardhat/config`, nor `hardhat/hre`
3. Open the config in vscode right without opening a file that imports one of the files mentioned above.

This is because if you have a tsconfig, all the ts files that it refers to (all by default), affect the import of types at once.

If you don't the LSP traverses the imported types as needed.

PR #8194, opened 2026-04-27
feat(test): support excluding tests with --invert flag

### LINKED ISSUE #8184: Support negation of test name filters in Solidity Test
Hardhat 3's Solidity Test support allows for filtering of tests with the `--grep` option:

```shell
npx hardhat test solidity --grep "^testFork_"
```

However, there is currently no way of expressing "everything but testFork_". The underlying Rust crate used intentionally errors if you attempt it:

```
npx hardhat test solidity --grep "^(?!testFork_)"

HardhatError: HHE802: Unhandled EDR error while running Solidity tests: regex parse error:
    ^(?!testFork_)
     ^^^
error: look-around, including look-ahead and look-behind, is not supported
```

One option would be the inclusion of an equivalent of `--grep` but that filters out based on the regex. In Foundry this would be [no-match-test](https://www.getfoundry.sh/config/reference/testing#no-match-test) .

We should examine other test filtering options at the same time e.g.:

* [match-path](https://www.getfoundry.sh/config/reference/testing#match-path) 
* [match-contract](https://www.getfoundry.sh/config/reference/testing#no-match-contract)

JunaidCD:

---

Closes #8184

### Problem

Hardhat supports `--grep` to run specific tests, but there is no way to exclude tests from execution.

### Solution

Add a new `--invert` flag that inverts the `--grep` match, allowing users to run all tests except those matching a pattern.

### Implementation

- Added `--invert` flag to the `test` task
- Passed the flag through task layers to the Mocha runner
- Used Mocha’s built-in `invert` option

### Example

npx hardhat test --grep "TestFork" --invert

Runs all tests except those matching "TestFork".

### Notes

- Tests were run locally but encountered Windows-specific issues related to symlink permissions and mutex locks.
- This appears to be environment-specific and unrelated to the implementation.

### DIFF
--- a/packages/hardhat-mocha/src/index.ts
+++ b/packages/hardhat-mocha/src/index.ts
@@ -25,6 +25,10 @@ const hardhatPlugin: HardhatPlugin = {
         defaultValue: undefined,
       })
       .addFlag({
+        name: "invert",
+        description: "Run tests that do NOT match the grep pattern",
+      })
+      .addFlag({
         name: "noCompile",
         description: "Don't compile the project before running the tests",
       })
--- a/packages/hardhat-mocha/src/task-action.ts
+++ b/packages/hardhat-mocha/src/task-action.ts
@@ -6,7 +6,7 @@ import type { MochaOptions } from "mocha";
 
 import { resolve as pathResolve } from "node:path";
 
-import { HardhatError } from "@nomicfoundation/hardhat-errors";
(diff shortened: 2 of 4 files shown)


kanej:
Hey, I am closing this PR inline with our contribution guidelines: https://github.com/NomicFoundation/hardhat/blob/main/CONTRIBUTING.md#note-about-small-prs-and-airdrop-farming.

The original feature is targeted at Solidity Test, not mocha.

PR #8195, opened 2026-04-27
Replace chalk with util.styleText

### LINKED ISSUE #8160: Replace `chalk` with `util.styleText`

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/8160

#### Review suggestions: do it commit by commit.
There are a lot of files to review, but most of them are just simple replacement of `chalk` being changed into `styleText`


packages/hardhat-keystore/src/internal/ui/user-display-messages.ts
```diff
@@ -13,8 +14,9 @@ export class UserDisplayMessages {
     key: string,
     dev: boolean,
   ): string {
-    return chalk.yellow(
-      `The key "${key}" already exists in the ${getKeystoreType(dev)} keystore. Use the ${chalk.blue.italic("--force")} flag to overwrite it.`,
+    return styleText(
+      "yellow",
+      `The key "${key}" already exists in the ${getKeystoreType(dev)} keystore. Use the ${styleText(["blue", "italic"], "--force")} flag to overwrite it.`,
```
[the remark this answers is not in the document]
alcuadrado: I think this is different with respect to `chalk`, @ChristopherDedominici

alcuadrado:
just a reminder: you need to push the minimum supported node version

.github/workflows/ci.yml
```diff
@@ -367,7 +367,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
-        node: [22.0.0, 24.0.0]
```
ChristopherDedominici, line 370: I think here we forgot to update to `22.10.0`. Or there is a reason I missed on why it was left to `22.0.0`?
Now I aligned the version in all the file

packages/hardhat-node-test-reporter/src/formatting.ts
```diff
@@ -1,12 +1,17 @@
...
 import type { TestEventData } from "./types.js";
 
-import chalk from "chalk";
+import { styleText } from "node:util";
 
 import { formatError } from "./error-formatting.js";
 
-export const INFO_SYMBOL: string = chalk.blue("\u2139");
-export const SUCCESS_SYMBOL: string = chalk.green("✔");
+export function getInfoSymbol(): string {
```
ChristopherDedominici, line 8: Functions instead of consts so that styleText is called at usage time, correctly reflecting the runtime environment (e.g. NO_COLOR, TTY detection).

packages/hardhat-node-test-reporter/integration-tests/fixture-tests/assert-test/result.txt
```diff
@@ -12,7 +12,7 @@
       [31m8) should match strings[39m
```
ChristopherDedominici, line 12: For all the `.txt` files:

I ran the command `node --import tsx/esm integration-tests/index.ts --color --show-output` in the folder `hardhat-node-test-reporter` and compared the colors and messages with `main`, all files matche

packages/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/trace-output.ts
```diff
@@ -107,10 +108,12 @@ export class TraceOutputManager {
...
       const prefix = callTraces.length > 1 ? "Traces from" : "Trace from";
       const coloredPrefix = this.#labelColor(prefix);
-      const styledMethod = failed ? chalk.red(method) : chalk.dim(method);
+      const styledMethod = failed
+        ? styleText("red", method)
+        : styleText("dim", method);
       const header = `${coloredPrefix} ${coloredLabel}: ${styledMethod}`;
 
-      this.#printLineFn(`${header}\n${formatTraces(callTraces, "  ", chalk)}`);
+      this.#printLineFn(`${header}\n${formatTraces(callTraces, "  ")}`);
```
ChristopherDedominici, line 116: I added `styleText` as default in function `printLineFn`, so passing it it's no longer needed

packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
```diff
@@ -161,7 +161,6 @@ export function isTestSuiteArtifact(artifact: Artifact): boolean {
 export function warnDeprecatedTestFail(
   artifact: Artifact,
   sourceNameToUserSourceName: Map<string, string>,
-  colorizer: Colorizer = chalk,
```
ChristopherDedominici, line 164: it was not used

packages/hardhat/src/internal/utils/colorizer.ts
```diff
@@ -1,9 +0,0 @@
-export interface Colorizer {
```
ChristopherDedominici, line 1: All the functions are now called inline directly, this file was a bit of an overkill

.github/workflows/ci.yml
alcuadrado: good call

alcuadrado:
Two small comments:

1) There are still a few mentions of `chalk`:

```
$ rg chalk -l
CONTRIBUTING.md
pnpm-lock.yaml
packages/hardhat-verify/CHANGELOG.md
packages/hardhat-node-test-reporter/integration-tests/index.ts
packages/hardhat-node-test-reporter/README.md
packages/hardhat/CHANGELOG.md
packages/hardhat/src/internal/builtin-plugins/coverage/helpers/accessors.ts
packages/hardhat/src/internal/builtin-plugins/gas-analytics/helpers/accessors.ts
```

The lockfile and the changelogs can be ignored.

2) I'm not convinced that it's safe to romove `colorize.ts`. I would leave it, exporting only the interface, and with a note that it's deprecated. We remove it when we bump major of utils.

Everything else looks good

packages/hardhat-node-test-reporter/src/formatting.ts
alcuadrado: makes sense

packages/hardhat/src/internal/builtin-plugins/network-manager/edr/utils/trace-output.ts
alcuadrado: makes sense

ChristopherDedominici:
> Two small comments:
> 
> 1. There are still a few mentions of `chalk`:
> 
> ```
> $ rg chalk -l
> CONTRIBUTING.md
> pnpm-lock.yaml
> packages/hardhat-verify/CHANGELOG.md
> packages/hardhat-node-test-reporter/integration-tests/index.ts
> packages/hardhat-node-test-reporter/README.md
> packages/hardhat/CHANGELOG.md
> packages/hardhat/src/internal/builtin-plugins/coverage/helpers/accessors.ts
> packages/hardhat/src/internal/builtin-plugins/gas-analytics/helpers/accessors.ts
> ```
> 
> The lockfile and the changelogs can be ignored.
> 
> 2. I'm not convinced that it's safe to romove `colorize.ts`. I would leave it, exporting only the interface, and with a note that it's deprecated. We remove it when we bump major of utils.
> 
> Everything else looks good

@alcuadrado Good catch, I fixed it here: https://github.com/NomicFoundation/hardhat/pull/8195/changes/09178d3fe80ed5760c9588dc62afc3ba2a80b2ac.

Just to be sure I understand: internally I no longer use `Colorize`. Your idea is to keep it as it is, and add a `@deprecated` warning?

alcuadrado:
Yes, because older versions of hardhat may end up with the new version of hardhat-utils, and it's unclear if they will break. 

They `import type` it, and I couldn't find a definite answer if removing the exports is a breaking change or not. But let's play safe.

We only need to keep the export and the interface though.

alcuadrado (approved):

alcuadrado:
wait, i got this wrong! sorry. I thought i was reviewing a hardhat-utils change. You were right. it can be removed. Sorry

PR #8197, opened 2026-04-27
New error classifier and filter

alcuadrado:
This PR introduces two new functions to classify and filter the errors caught by the CLI's `main` and unhandled errors and promise rejections.

The first commit introduces the classier and filter logic and their tests. The classifier first categorizes the error, and then the filter decides if it should be reported based on the error and the category. I tried keeping their tests as realistic (i.e. integration tests) as possible, but I had to unit tests some things, otherwise this would be a huge PR.

The second commit integrates it into the new error reporter module.

The third commit deletes all the old error filtering logic.

Finally, a forth commit fixes some small issues with our sentry initialization. Take a look at the commit for more detail.

I recommend reviewing one commit at the time.


packages/hardhat/src/internal/cli/telemetry/error-classification/filter.ts
```diff
@@ -0,0 +1,140 @@
...
+ */
+export function shouldBeReported(
+  error: Error,
+  category: ErrorCategory,
+): boolean {
+  switch (category) {
+    case ErrorCategory.CJS_TO_ESM_MIGRATION_ERROR:
+    case ErrorCategory.HH2_TO_HH3_MIGRATION_ERROR:
+    case ErrorCategory.TYPESCRIPT_SUPPORT_ERROR:
+    case ErrorCategory.DEVELOPMENT_TIME_ERROR:
```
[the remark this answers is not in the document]
alcuadrado: ignore, we have a linter for that

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
```diff
@@ -0,0 +1,219 @@
...
+    return FrameOrigin.USER_PROJECT;
+  }
+
+  return FrameOrigin.OTHER;
+}
+
+/**
+ * Returns true when `value` starts with any of the supplied prefixes.
+ */
+function startsWithAny(value: string, ...prefixes: string[]): boolean {
```
schaable, line 233: this may be worth moving into `hardhat-utils/string`

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
```diff
@@ -0,0 +1,219 @@
...
+  error: Error,
+  errorClass: abstract new (...args: never[]) => Error,
+): boolean {
+  return error.name === errorClass.name;
+}
+
+/**
+ * Returns true when `value` contains any of the supplied substrings.
+ */
+export function includesAny(
```
schaable, line 74: this may be worth moving into `hardhat-utils/string`

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
```diff
@@ -0,0 +1,219 @@
...
+    ) {
+      return FrameOrigin.FIRST_PARTY;
+    }
+
+    return FrameOrigin.THIRD_PARTY;
+  }
+
+  if (
+    startsWithAny(location, "/", "file://", "[eval]") ||
+    location.match(/^[A-Z]\:\\\\/) !== null
```
schaable:
```suggestion
    /^[A-Za-z]:\//.test(location)
```

packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts
```diff
@@ -0,0 +1,622 @@
...
+  isPluginTaskActionError,
+  isUserTaskActionError,
+  isPluginHookHandlerError,
+];
+
+const ESM_MIGRATION_MARKERS = [
+  "require is not defined in es module scope",
+  "module is not defined in es module scope",
+  "exports is not defined in es module scope",
+];
```
schaable, line 179: These could be collapsed into a single pattern like `"is not defined in ES module scope"`. Also, these two are currently missing (and would be covered by that pattern):
 - `"__dirname is not defined in ES module scope"`
 - `"__filename is not defined in ES module scope"`

And we should also include `"Cannot use import statement outside a module"` for cases where ESM imports are used in a CJS file, which can happen during partial migrations.

packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts
```diff
@@ -0,0 +1,622 @@
...
+function isHH3MigrationError(
+  context: ErrorContext,
+): ErrorCategory.HH2_TO_HH3_MIGRATION_ERROR | undefined {
+  if (
+    context.errorChain.some((candidate) =>
+      hasErrorClassName(candidate, UsingHardhat2PluginError),
+    ) ||
+    context.errorChain.some((candidate) =>
+      /You are trying to use a Hardhat 2 plugin in a Hardhat 3 project/i.test(
+        candidate.message,
```
schaable:
```suggestion
    context.errorChain.some((candidate) =>
      hasErrorClassName(candidate, UsingHardhat2PluginError) ||
      /You are trying to use a Hardhat 2 plugin in a Hardhat 3 project/i.test(
        candidate.message,
      ),
    )
````

File: packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts [hunk 7]
```diff
@@ -0,0 +1,622 @@
...
+  'the requested module "hardhat" does not provide an export named',
+  "the requested module 'hardhat/config' does not provide an export named",
+  'the requested module "hardhat/config" does not provide an export named',
+  "the requested module 'hardhat/plugins' does not provide an export named",
+  'the requested module "hardhat/plugins" does not provide an export named',
+  "the requested module 'hardhat/builtin-tasks/task-names' does not provide an export named",
+  'the requested module "hardhat/builtin-tasks/task-names" does not provide an export named',
+  "the requested module 'hardhat/types/runtime' does not provide an export named",
+  'the requested module "hardhat/types/runtime" does not provide an export named',
+];
```
schaable, line 210: These can also happen in a fresh HH3 project, especially the first one. I’m assuming this is intentional and that we’re only keeping this matcher temporarily, is that correct?

packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts
```diff
@@ -0,0 +1,622 @@
...
+        isNodeFilesystemError(candidate),
+    )
+  ) {
+    return ErrorCategory.FILESYSTEM_INTERACTION_ERROR;
+  }
+}
+
+function isUnknownEdrError(error: Error, context: ErrorContext): boolean {
+  const errorIndex = context.errorChain.indexOf(error);
+  const errorChain =
```
schaable: The error is always in `context.errorChain`, so this should be an assertion. Even if it weren’t, `context.stackFramesByError` is built from `context.errorChain`, so there wouldn’t be a stack frame for this error there, and the `.some` call at 515 would return `false`.

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
```diff
@@ -0,0 +1,219 @@
...
+  return (
+    value !== undefined &&
+    substrings.some((substring) => value.includes(substring))
+  );
+}
+
+/**
+ * Returns a Node-style `code` string from an error or any Error cause.
+ */
+export function getNodeErrorCode(error: Error): string | undefined {
```
schaable: Should this break on cycles and use a fixed max depth, similarly to how `getErrorChain` does?

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
schaable: Claude found a few places where we could reuse this:

| File | Current |
|---|---|
| [solidity-test/helpers.ts:155](packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts#L155) | `name.startsWith("test") \|\| name.startsWith("invariant")` |
| [dependency-resolver.ts:426](packages/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts#L426) | `importPath.startsWith("./") \|\| importPath.startsWith("../")` |
| [error-formatting.ts:384](packages/hardhat-node-test-reporter/src/error-formatting.ts#L384) | `location.startsWith("node:") \|\| location.startsWith("async node:")` |
| [network-helpers.ts:181-182](packages/hardhat-network-helpers/src/internal/network-helpers/network-helpers.ts#L181-L182) | `SUPPORTED_TEST_NETWORKS.some((n) => version.toLowerCase().startsWith(n))` |

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
schaable: Claude found a few places where we could reuse this:

| File | Current |
|---|---|
| [jsonrpc-client.ts:359-360](packages/ignition-core/src/internal/execution/jsonrpc-client.ts#L359-L360) | `error.message.includes("execution reverted") \|\| error.message.includes("invalid opcode")` |

packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts
```diff
@@ -0,0 +1,622 @@
...
+    context.errorChain.some((candidate) => {
+      const code = getTypescriptSupportErrorCode(candidate);
+
+      if (code === undefined) {
+        return false;
+      }
+
+      return (
+        code !== "ERR_UNKNOWN_FILE_EXTENSION" ||
+        includesAny(candidate.message, ".ts", ".mts", ".cts")
```
schaable: These will appear as written in the filesystem. Probably uncommon, but users could use uppercase (".TS", ".MTS", ".CTS"). Worth doing `candidate.message.toLowerCase()` to handle that.

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
```diff
@@ -0,0 +1,219 @@
...
+  allStackFrames: StackFrame[];
+}
+
+/**
+ * Builds the shared derived data used by classification and filtering.
+ *
+ * This keeps stack parsing and cause-chain traversal consistent across
+ * matchers, and avoids recomputing them for every category heuristic.
+ */
+export function createErrorContext(error: Error): ErrorContext {
```
schaable, line 33: Every time we walk the error chain to compare messages, we're calling `toLowerCase` or using case-insensitive regexes. It may be worth adding a `lowercaseMessageByError: Map<Error, string>`. Not for performance, but for consistency and to avoid having to remember to normalize case each time. Claude should be able to add this and simplify all matchers to use `lowercaseMessageByError` instead of `errorChain`, removing all the `toLowerCase` calls and the `i` flag in regexes.

schaable (commented):
I still need to take a pass over the test files, but the rest of the PR looks good to me. Left a few comments.

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
schaable: btw, I don’t mean you need to replace these, I just wanted to confirm it’s worth moving the function to hardhat-utils

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
schaable: btw, I don’t mean you need to replace these, I just wanted to confirm it’s worth moving the function to hardhat-utils

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
alcuadrado: 15e6c7183312e0723f5df3cb27a40c6036061a72

packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts
alcuadrado: 3f3719814fa23840777dd4e54c829c7f10d4393b

packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts
alcuadrado: 893e1432701878dab6e4de488404d97f168b6f19

packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts
alcuadrado: I added docs saying that it's temporary

04d70b3d4a10d9ccee6c84e8696e9e6b6274e355

packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts
alcuadrado: 50329e91b68ad683793dc0e5a60abfff865a785b

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
alcuadrado: 21d6e3104e1b03e3ebee87855c470d432cd0c490

packages/hardhat/src/internal/cli/telemetry/error-classification/classifier.ts
alcuadrado: Fixing the next comment made this outdated

packages/hardhat/src/internal/cli/telemetry/error-classification/helpers.ts
alcuadrado: a3931e1f64983a3efff5b88a858e36819dcd56b8

schaable (approved):
LGTM

PR #8213, opened 2026-04-28
fix: update memory usage when passing coverage

### LINKED ISSUE #8212: OpenZeppelin Contract coverage out of memory issue
The Hardhat 3 migration branch of OpenZeppelin has shown up a memory issue with `--coverage`.

The coverage CI job on our branch of OpenZeppelin/contracts (https://github.com/NomicFoundation/openzeppelin-contracts/actions/runs/24997663272/job/73198824287) false. The CI job runs:

```shell
npx hardhat test --coverage
```

It fails with an out of memory on the Node side. The OpenZeppelin test suite is large and its Solidity Test suite contains multiple invariant and fuzz tests with run values of 5000.

kanej:
Improve memory usage during `--coverage` runs of both Solidity test and JS/TS tests.

We do this by holding less hit count data - swapping an array that is appended to and eventually processed with a map structure where processing happens immediately on count data coming across the EDR NAPI boundary.

Fixes #8212

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
@@ -74,7 +74,7 @@ export class CoverageManagerImplementation implements CoverageManager {
   /**
    * @private exposed for testing purposes only
    */
-  public data: CoverageData = [];
+  public data: Map<string, number> = new Map();
 
   readonly #coveragePath: string;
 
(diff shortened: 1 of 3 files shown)


packages/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
schaable, line 41: The file comment already mentions that EDR doesn’t expose per-statement execution counts yet, so what we’d persist aren’t really “counts” but just how many times EDR happened to notify us about a tag. A `Set<string>` reflects that more accurately and avoids suggesting the values are meaningful. It also ends up slightly smaller on disk.

That said, this is a nit. Happy either way if you’d prefer to keep the `Map` for forward compatibility.

schaable (approved):
Looks good! I left a small comment about counting the data, but I’ll leave the decision to you.

packages/hardhat/src/internal/builtin-plugins/coverage/coverage-manager.ts
kanej: I am going to keep the map to save on the retesting to be honest.
This needs revisiting to get proper line counts, but this will get us the performance win.

PR #8218, opened 2026-04-29
Fix bug in solidity test artifact valdation

alcuadrado:
When running `test solidity --no-compile` without providing explicit test file we were running the validation of artifacts anyway.

This was unnecessary, and lead to some edge cases that we don't even need to handle. For example, test files that don't have any contract are roots, but don't have artifacts.

### DIFF
--- a/cspell.dictionary.txt
+++ b/cspell.dictionary.txt
@@ -123,6 +123,7 @@ txts
 ufixed
 Uints
 Uncontended
+uncompiled
 unidici
 unmined
 unpad
--- a/packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
@@ -141,9 +141,10 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
     ({ edrArtifactsWithMetadata, allBuildInfosAndOutputs } =
       await loadArtifacts(hre.solidity, ["contracts"]));
 
-    // When noCompile, validate selected test roots have compiled artifacts when
-    // they are provided by the user, otherwise we know we have all the compiled
-    // artifacts already so we can skip this validation
+    // When noCompile is enabled, only validate compiled artifacts for test roots
+    // explicitly selected by the user. If no files were specified, skip
+    // validating every discovered test root and run whatever compiled test
+    // suites are available.
     if (noCompile === true && testFiles.length > 0) {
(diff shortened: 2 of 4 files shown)


packages/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -484,6 +484,24 @@ describe("solidity-test/task-action", function () {
       }
     });
 
+    it("should not throw when noCompile is true and no testFiles are provided, even if the tests path covers uncompiled files", async () => {
```
marianfe, line 487: do we have the reverse of this: it("should throw when noCompile and testFiles are provided") ?

cspell.dictionary.txt
```diff
@@ -123,6 +123,7 @@ txts
 ufixed
 Uints
 Uncontended
+uncompiled
```
marianfe, line 126: tiny nit: probably "not compiled" would express the same without adding a dictionary exception

marianfe (approved):

packages/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts
alcuadrado: yes, that one was already present

cspell.dictionary.txt
alcuadrado: good idea

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -141,8 +141,11 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
     ({ edrArtifactsWithMetadata, allBuildInfosAndOutputs } =
       await loadArtifacts(hre.solidity, ["contracts"]));
 
-    // When noCompile, validate selected test roots have compiled artifacts
-    if (noCompile === true) {
+    // When noCompile is enabled, only validate compiled artifacts for test roots
+    // explicitly selected by the user. If no files were specified, skip
+    // validating every discovered test root and run whatever compiled test
+    // suites are available.
+    if (noCompile === true && testFiles.length > 0) {
```
Wodann, line 148: Question: Is there a reason `hardhat compile` doesn't generate all possible test compilation artifacts?

If `hardhat test --no-compile` is called after `hardhat compile`, I would expect the check for compiled artifacts to pass with or without filter.

PR #8223, opened 2026-04-30
Fix tx leak in viem.assertions.emit / emitWithArgs when ABI shape check fails

lsheva:
Both helpers run synchronous `assert.ok` checks against the contract ABI before awaiting `contractFn`. Since viem submits the write as soon as the promise is created, when one of these checks fails (wrong arg count, missing event, ambiguous overload) the in-flight tx is left dangling and ends up mining inside the next test, which has been a fairly nasty source of cross-test state bleed.

This settles `contractFn` first via a small `settle()` helper and only rethrows its rejection after the ABI checks pass, so the existing precedence (ABI errors over tx reverts) stays the same.

The other assertion helpers all await their promise as the first thing inside a `try`, so they aren't affected.

Adds three regression tests that pin the contract promise lifetime down by attaching a settlement flag and asserting it's true after the assertion error escapes — they fail on `v-next` without the source change and pass with it.

### DIFF
--- a/.changeset/fix-viem-assertions-leak-tx.md
+++ b/.changeset/fix-viem-assertions-leak-tx.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-viem-assertions": patch
+---
+
+Fix `emit` and `emitWithArgs` leaking the underlying transaction into the next test when the synchronous ABI shape check failed. These helpers now always settle `contractFn` before any assertion can throw.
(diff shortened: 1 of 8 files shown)


alcuadrado:
Great catch! 

I hope you don't mind I force pushed into your branch. You were using the wrong base branch (`v-next`), so I rebased yours on top of `main`.

I also added some tests to ensure that your changes don't break the async stack traces.

alcuadrado (approved):

PR #8234, opened 2026-05-05
feat: add per-cheatcode suggestions for unsupported cheatcodes

### LINKED ISSUE #8209: feat: add suggestion to unsupported cheatcodes
Currently, unsupported cheatcodes report that they are explicitly unsupported in EDR+Hardhat, compared to Foundry.

In addition to this, it would be useful to allow per-cheatcode suggestions when we explicitly provide an alternative.

E.g. for:

```
    #[cheatcode(group = Utilities, status = Unsupported)]
    function eip712HashType(string calldata bindingsPath, string calldata typeName) external pure returns (bytes32 typeHash);
```

we could suggest:

> Providing a path to a bindings file is not supported. Please use the `eip712HashType(string calldata typeNameOrDefinition)` cheatcode instead.

and even provide a link to Hardhat documentation.

tenderdeve:
When an unsupported cheatcode error is reported, this appends a context-specific suggestion if a known alternative exists. For example, using `eip712HashType(string,string)` now suggests `eip712HashType(string)` instead.

The suggestion map lives in a simple Record and can be extended as more cheatcode alternatives are identified.

Fixes #8209

Tested with the existing CHEATCODE_ERROR test suite plus new cases for suggestion lookup.

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
@@ -28,19 +28,8 @@ const CHEATCODE_SUGGESTIONS: Record<string, string> = {
 
 export function getCheatcodeSuggestion(cheatcode: string): string {
   const suggestion = CHEATCODE_SUGGESTIONS[cheatcode];
-  if (suggestion !== undefined) {
-    return suggestion;
-  }
 
-  // Match by function name prefix for overloaded variants
-  const name = cheatcode.split("(")[0];
-  for (const [key, value] of Object.entries(CHEATCODE_SUGGESTIONS)) {
-    if (key.startsWith(name + "(")) {
-      return value;
-    }
-  }
-
-  return "";
+  return suggestion ?? "";
 }
 
 export function createSolidityErrorWithStackTrace(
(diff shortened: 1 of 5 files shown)


packages/hardhat/src/internal/builtin-plugins/network-manager/edr/stack-traces/stack-trace-solidity-errors.ts
```diff
@@ -21,6 +21,28 @@ import {
   UNRECOGNIZED_FUNCTION_NAME,
 } from "./solidity-stack-trace.js";
 
+const CHEATCODE_SUGGESTIONS: Record<string, string> = {
```
ChristopherDedominici: Thanks for taking this on! One thing though: the linked issue explicitly mentions providing a link to Hardhat documentation as part of the feature, and the current shape (Record<string, string>) doesn't leave room for that

packages/hardhat/src/internal/builtin-plugins/solidity-test/stack-trace-solidity-errors.ts
```diff
@@ -71,7 +73,7 @@ export function getMessageFromLastStackTraceEntry(
       if (stackTraceEntry.details !== undefined) {
         switch (stackTraceEntry.details.code) {
           case CheatcodeErrorCode.UnsupportedCheatcode:
-            return `Cheatcode '${stackTraceEntry.details.cheatcode}' is not supported by Hardhat.`;
+            return `Cheatcode '${stackTraceEntry.details.cheatcode}' is not supported by Hardhat.${getCheatcodeSuggestion(stackTraceEntry.details.cheatcode)}`;
```
ChristopherDedominici: I'd probably add a unit test to cover this change

ChristopherDedominici (dismissed):
Hi @tenderdeve , thanks for the PR! I left a few comments. Also, a short changeset should be added too. Let me know if you have any comment :)

packages/hardhat/src/internal/builtin-plugins/solidity-test/stack-trace-solidity-errors.ts
kanej: I added a test to cover this.

kanej (approved):
LGTM!

PR #8243, opened 2026-05-05
Support EIP-712 cheatcodes in Solidity Test

### LINKED ISSUE #8171: Support EIP-712 cheatcodes in Solidity Test
Currently the EIP-712 cheatcodes are not supported in Hardhat Solidity Test. We should explore adding targeted support for:

```solidity
#[cheatcode(group = Utilities)]
function eip712HashType(string calldata typeNameOrDefinition) external pure returns (bytes32 typeHash);

#[cheatcode(group = Utilities)]
function eip712HashStruct(string calldata typeNameOrDefinition, bytes calldata abiEncodedData) external pure returns (bytes32 typeHash);

#[cheatcode(group = Utilities)]
function eip712HashTypedData(string calldata jsonData) external pure returns (bytes32 digest);
```

ChristopherDedominici:
Link to website docs PR: https://github.com/NomicFoundation/hardhat-website/pull/260

Fixes: https://github.com/NomicFoundation/hardhat/issues/8171

----

# Possible follow-up
[See comment here](https://github.com/NomicFoundation/hardhat/pull/8243#discussion_r3237866804) - the comment is partially addressed [here](https://github.com/NomicFoundation/hardhat/pull/8243/changes/62cf8b9e614e05d215e8c7a949fb912de2daa134)

`getTestFunctionOverrides` (packages/hardhat/src/internal/builtin-plugins/solidity-test/inline-config/index.ts) and `collectEip712CanonicalTypes` both JSON.parse the same build-info outputs from buildInfosAndOutputs. We could cache the parsed outputs in `getTestFunctionOverrides` and reuse them from `collectEip712CanonicalTypes` to avoid re-parsing the overlap.

----

# Question
✅ Answer from EDR team: throwing makes sense: [see here](https://github.com/NomicFoundation/hardhat/pull/8243#issuecomment-4454925210)

The PR can already be merged, but I have 1open question that can be addressed now, as follow-up, or ignored if we think the solution is solid enough.

### Duplicate struct names throw instead of being disambiguated

```solidity
// File A.sol
struct S { uint256 a; }

// File B.sol
struct S { uint256 b; }
```

- Forge: silently renames to `S_0` and `S_1`; user calls `vm.eip712HashType("S_0")`.
- This PR: throws `EIP712_DUPLICATE_STRUCT_NAME` and refuses to run.

The new behavior is arguably better UX, but it's a divergence — projects that worked on forge will fail here. Worth flagging in the migration guide.

----

# Potential EDR issues

[Tracked here](https://github.com/NomicFoundation/hardhat/pull/8243#issuecomment-4453239715)

Three issues rooted in EDR. Hardhat's collector emits canonical type strings that match `forge bind-json` byte-for-byte; EDR's downstream handling diverges.

## 1. Self-referential structs crash the test run

A struct that references itself (directly or transitively) makes EDR abort at startup.

```solidity
struct Node {
  uint256 value;
  Node[] children;
}
```

**Expected:** `vm.eip712HashType("Node")` returns `keccak256("Node(uint256 value,Node[] children)")`.

**Actual:** the whole test run aborts before any test executes:

```
HHE802: Unhandled EDR error while running Solidity tests:
failed to canonicalize EIP-712 type Node(uint256 value,Node[] children):
missing type in type resolution: primary component
```

**Root cause:** EDR's canonical-type resolver expects every member type to appear in the appended-dependency section. For a self-reference, the type appears only as the primary, not in the appended section, and resolution fails.

**Impact:** anyone with a tree, linked list, or recursive routing struct — common in DeFi — hits this on first run. The only workaround is to rename or remove the struct from the include glob.

## 2. `eip712HashStruct` silently rejects the tuple ABI encoding

`abi.encode(field1, field2, ...)` and `abi.encode(structInstance)` are byte-identical for static-only structs, but differ for any struct with a dynamic field (string, bytes, array). EDR only accepts the second form.

```solidity
struct Sample {
  address a;
  string memo;
}

// Works:
CHEATS.eip712HashStruct(
  "Sample",
  abi.encode(Sample({a: address(0xCAFE), memo: "hi"}))
);

// Fails with "buffer overrun while deserializing":
CHEATS.eip712HashStruct(
  "Sample",
  abi.encode(address(0xCAFE), "hi")
);
```

**Root cause:** EDR's cheatcode handler ABI-decodes the buffer as the resolved struct type, which includes the outer offset word for dynamic structs. The tuple form lacks that word, so decoding overruns.

**Impact:** users will write the tuple form first (it works for static structs and reads more naturally), then hit a confusing error the moment they add a string or bytes field. The error message doesn't mention either encoding form.

**Suggested fix:** accept both shapes, or replace the error message with something like `expected abi.encode(structInstance); got abi.encode(field1, field2, ...) — these are not equivalent for structs with dynamic fields`.

## 3. Whitespace in the canonical-type-string overload is rejected

The design doc says `vm.eip712HashType` "will output the canonical type even if the input is malformed with extra whitespaces". EDR's parser does not.

```solidity
// Works:
CHEATS.eip712HashType("Foo(uint256 x)");

// Fails:
CHEATS.eip712HashType("Foo( uint256 x )");
//   vm.eip712HashType: failed to parse EIP-712 canonical type
//   Foo( uint256 x ): no parseable type definition found
```

**Root cause:** EDR's canonical-type-string parser tokenizes strictly and doesn't normalize whitespace around `(`, `)`, or `,`.

**Impact:** divergence from the documented behavior. Forge handles this; users porting test fixtures from Foundry can hit it.

**Suggested fix:** normalize whitespace in the parser before tokenizing.


packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/glob.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  return true;
+}
+
+/**
+ * Returns true if `value` matches at least one of the given glob patterns.
+ */
+function matchesAny(value: string, patterns: string[]): boolean {
+  for (const pattern of patterns) {
+    if (globToRegExp(pattern).test(value)) {
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/a5d409073164cabbe2d1d63354b1d3e5c9b28849

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      const structs = extractStructsFromAst(source.ast, userSourceName);
+      for (const struct of structs) {
+        const key = `${userSourceName}::${struct.name}`;
+
+        if (seen.has(key)) {
+          continue;
+        }
+
+        seen.add(key);
+        collected.push(struct);
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/96a94d7bab164bb060bdee51e4ca80aac469672d

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
```diff
@@ -0,0 +1,219 @@
...
+    const existingHead = headByName.get(struct.name);
+
+    if (existingHead === undefined) {
+      byName.set(struct.name, struct);
+      headByName.set(struct.name, head);
+      sourceByName.set(struct.name, struct.sourcePath);
+      continue;
+    }
+
+    if (existingHead === head) {
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/7a56f67f4c148ca9e7ba6fae338ced399fc6d679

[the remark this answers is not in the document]
ChristopherDedominici: Old comment, this is already fixed

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      }
```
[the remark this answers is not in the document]
ChristopherDedominici: Old comment, already fixed

packages/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/remapped-npm-packages-graph.ts
```diff
@@ -28,15 +28,14 @@ import {
...
 
+import { HARDHAT_PROJECT_INPUT_SOURCE_NAME_ROOT } from "../../../../../types/solidity/solidity-artifacts.js";
 import { UserRemappingErrorType } from "../../../../../types/solidity.js";
 
 import { getNpmPackageName } from "./npm-module-parsing.js";
 import { parseRemappingString, selectBestRemapping } from "./remappings.js";
 import { sourceNamePathJoin } from "./source-name-utils.js";
 import { UserRemappingType } from "./types.js";
 
-const HARDHAT_PROJECT_INPUT_SOURCE_NAME_ROOT = "project";
```
ChristopherDedominici, line 38: I moved it in file `packages/hardhat/src/types/solidity/solidity-artifacts.ts`, so it can be re-used in `packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts`. So we have a single source of truth

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (sources === undefined) {
+      continue;
+    }
+
+    for (const [inputSourceName, source] of Object.entries(sources)) {
+      const userSourceName =
+        inputToUserSource.get(inputSourceName) ?? inputSourceName;
+
+      if (!isPathSelected(userSourceName, include, exclude)) {
+        continue;
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/c71f117b77a3ff30503221f30d456d0309edc66b

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/glob.ts
```diff
@@ -0,0 +1,225 @@
...
+  let escaped = "";
+  for (const ch of body) {
+    if (ch === "\\" || ch === "]") {
+      escaped += `\\${ch}`;
+    } else {
+      escaped += ch;
+    }
+  }
+
+  return `[${negated ? "^" : ""}${escaped}]`;
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/c7812618c0e7672136eee03135912a9c09668071

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/glob.ts
```diff
@@ -0,0 +1,225 @@
...
+      }
+    } else if (c === "?") {
+      regex += "[^/]";
+      i += 1;
+    } else if (c === "[") {
+      const end = findCharClassEnd(pattern, i);
+      if (end !== -1) {
+        regex += translateCharClass(pattern.slice(i + 1, end));
+        i = end + 1;
+        continue;
```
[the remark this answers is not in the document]
ChristopherDedominici: Not true, I added a test to confirm: https://github.com/NomicFoundation/hardhat/pull/8243/changes/c47fe34eb0d1054604033808c6cd0c7b8cf4bfbd

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/glob.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  let escaped = "";
+  for (const ch of body) {
+    if (ch === "\\" || ch === "]") {
+      escaped += `\\${ch}`;
+    } else {
+      escaped += ch;
+    }
+  }
+
```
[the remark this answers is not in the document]
ChristopherDedominici: Duplicated

1) not true
2) already added a test here to prove this: https://github.com/NomicFoundation/hardhat/pull/8243#discussion_r3227632397

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts
```diff
@@ -0,0 +1,116 @@
...
+        )
+          ? inputSourceName.slice(PROJECT_INPUT_SOURCE_NAME_PREFIX.length)
+          : inputSourceName;
+      }
+
+      if (!isPathSelected(userSourceName, include, exclude)) {
+        continue;
+      }
+
+      collected.push(
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/a17ad5971155e1fc10bb84c80879e69fc1311a8d

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
```diff
@@ -0,0 +1,271 @@
...
+
+      const deps = directStructDeps(struct, knownNames);
+      if (deps.some((dep) => !encodable.has(dep))) {
+        encodable.delete(name);
+        changed = true;
+      }
+    }
+  }
+
+  return encodable;
```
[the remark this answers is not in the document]
ChristopherDedominici: Already addressed in PR description

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
```diff
@@ -0,0 +1,271 @@
...
+      continue;
+    }
+
+    if (existingFingerprint === fingerprint) {
+      continue;
+    }
+
+    if (selectedNames !== undefined && !selectedNames.has(struct.name)) {
+      continue;
+    }
```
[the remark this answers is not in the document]
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/e043b885aad33d1405fc39d89ae3ba7651b05c19

packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
```diff
@@ -122,10 +124,13 @@ export async function solidityTestConfigToSolidityTestRunnerConfigArgs({
 
   const shouldAlwaysCollectStackTraces = verbosity > DEFAULT_VERBOSITY;
 
+  // `eip712Types` must be passed as a separate param, not via `config`
+  const { eip712Types: _, ...configWithoutEip712 } = config;
```
schaable: Medium/Low:

This feels a bit off. The reader has to understand why there are two variables, `eip712Types` and `eip712CanonicalTypes`, referring to EIP-712 types, and the comment doesn't help much.

I'd prefer to destructure the config in the task action and pass the config without the types to `solidityTestConfigToSolidityTestRunnerConfigArgs`:

```typescript
const { eip712Types, ...solidityTestConfig } = hre.config.test.solidity;
const testRunnerConfig =
  await solidityTestConfigToSolidityTestRunnerConfigArgs({
    // ...
    config: solidityTestConfig,
    eip712CanonicalTypes,
  });
```

The type of `config` in `SolidityTestConfigParams` can then be updated to `Omit<SolidityTestConfig, "eip712Types">`.

Also, the function name `solidityTestConfigToSolidityTestRunnerConfigArgs` and the type `SolidityTestConfigParams` are a bit misleading, and could use a rename, for example `resolveSolidityTestRunnerConfigArgs` and `SolidityTestRunnerConfigArgsInputs`.

That said, if the boundary between Hardhat and EDR moves, all this code might go away, so the refactor may not be worth it. However, we don't know if or when that's going to happen, so I think it's worth doing it now. Claude can probably one-shot it.

packages/hardhat/src/types/solidity/solidity-artifacts.ts
```diff
@@ -1,5 +1,16 @@
...
+/**
+ * The input source name prefix used for files belonging to the Hardhat
+ * project (as opposed to npm packages, which use `npm/<pkg>@<version>`).
+ *
+ * For example, a project file at `<root>/contracts/Foo.sol` has input
+ * source name `project/contracts/Foo.sol`. Part of the build-info format
+ * contract: changing this value would require a new build-info format
+ * version.
+ */
+export const HARDHAT_PROJECT_INPUT_SOURCE_NAME_ROOT = "project";
```
schaable: High:

I'm not sure this is the best place for this constant. The file is re-exported in `packages/hardhat/src/types/solidity.ts`, which makes this constant part of the public interface, and I don't think that's intentional. It's also a runtime value in a directory that's otherwise type-only.

I suggest moving it to `packages/hardhat/src/internal/builtin-plugins/solidity/constants.ts` which already holds other build system constants.

packages/hardhat/src/internal/builtin-plugins/solidity-test/type-extensions.ts
```diff
@@ -64,6 +64,11 @@ declare module "../../../types/test.js" {
       includePushBytes?: boolean;
       shrinkRunLimit?: number;
     };
+
+    eip712Types?: {
```
schaable, line 68: Minor:

This may be worth resolving to empty arrays, so we can drop these lines in `collectEip712CanonicalTypes`:

```typescript
  const include = config?.include ?? [];
  const exclude = config?.exclude ?? [];
```

Not a big deal. I like to keep resolution separate from other steps, but in this case full resolution would mean converting `eip712Types` into `eip712CanonicalTypes`, which isn't possible without taking a performance hit. This could be a decent middle ground, at least so we don't have to keep handling `undefined` checks when using the resolved config.

Up to you whether you think it's worth applying.

.changeset/clever-foxes-jump.md
```diff
@@ -0,0 +1,6 @@
+---
+# docs: https://github.com/NomicFoundation/hardhat-website/pull/260
+"@nomicfoundation/hardhat-errors": patch
+---
+
+Add error descriptor for duplicate EIP-712 struct names surfaced by the Solidity Test cheatcodes.
```
schaable, line 6: Nit: 

A separate changeset for the hardhat error seems weird. This is part of the feature. I'd merge both changesets and ship only one.

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts
```diff
@@ -0,0 +1,121 @@
...
+  const parsed = buildInfosAndOutputs.map(({ buildInfo, output }) => {
+    const parsedBuildInfo: SolidityBuildInfo = JSON.parse(
+      bytesToUtf8String(buildInfo),
+    );
+    const parsedOutput: SolidityBuildInfoOutput = JSON.parse(
+      bytesToUtf8String(output),
+    );
+
+    return { buildInfo: parsedBuildInfo, output: parsedOutput };
+  });
```
schaable: Parsing all build infos and their outputs is expensive, so we should avoid it whenever possible. There are a couple of tricks we already use in the inline config logic that could help here too:

1. You're parsing all build infos just to build `inputToUserSource`. You can build the same mapping from the `edrArtifactsWithMetadata` array in `task-action.ts` using the `userSourceName` field together with `edrArtifact.id.source`.
2. You're parsing all build info outputs, even ones that don't contain structs. You can use `bytesIncludesUtf8String(buildInfo, "struct ")` to filter outputs that contain structs and only parse those.

There's another optimization we could do, but I think it needs more analysis and should go in a separate PR if we decide to apply it:
While `getTestFunctionOverrides` and `collectEip712CanonicalTypes` do different things, both end up parsing some of the same build info outputs. We could cache the parsed outputs in `getTestFunctionOverrides` and reuse them in `collectEip712CanonicalTypes` whenever the same outputs need to be parsed again. The benefit would be proportional to the overlapping between the two functions, but it's still cheap to do. I still think it's worth creating an issue to track this.

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  const collected: CollectedStruct[] = [];
+  const selectedNames = new Set<string>();
+
+  for (const { output } of parsed) {
+    const sources = output.output.sources;
+    if (sources === undefined) {
+      continue;
+    }
```
schaable: Do you know in which scenarios this can happen? I thought `SolidityBuildInfoOutput.output.sources` was always available. If this is a guard against malformed outputs, it should probably be an invariant.

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    // Build the user-defined value type index *per build info*. solc node ids
+    // are unique only within a single compilation, so pooling the indexes
+    // across build infos would let one compilation's user-defined value type
+    // shadow another at the same numeric id and silently resolve
+    // `referencedDeclaration` to the wrong underlying type. The index still
+    // has to span every source in *this* build (not just include-matched ones)
+    // because a struct in an included file may reference a user-defined value
+    // type defined in a non-included file.
```
schaable: Nit:

This took me a bit to understand, so I think a rewording could help future readers. Not a blocker.

```suggestion
   // Two constraints determine the index's scope:
   //
   // 1. Per build info, not pooled across them. solc assigns node ids
   //    fresh in each compilation, so the same numeric id can mean
   //    different UDVTs in different builds. Pooling would let one
   //    compilation's definition silently overwrite another's at the
   //    same key, mis-resolving `referencedDeclaration`. See the test
   //    "scopes user-defined value type resolution per build info when
   //    node ids collide" for a repro.
   //
   // 2. Whole build info, not narrowed to a subset of sources. A struct
   //    member's `referencedDeclaration` can point at a UDVT defined in
   //    any source within the same compilation, so the index must cover
   //    every source in the build.
```

anaPerezGhiglia:
@ChristopherDedominici regarding the potential EDR issues
1. **Self-referential structs crash the test run**: I opened an issue for this on alloy-rs/core repo. 
   - alloy-rs/core#1102. 
   - There is a linked PR already for it
3. **eip712HashStruct silently rejects the tuple ABI encoding**: I double checked and this behaves the same way as in forge. That means that we wouldn't be exposing HH to claims like "my test used to work on Foundry, but fail on Harhat". However, that being said, if considered useful we can try to improve the information provided to the user about this. EDR _does not know why_ the decoding failed, but it can provide a more verbose error message. Another alternative is to improve the cheatcode documentation in the website. 
4. **Whitespace in the canonical-type-string overload is rejected**: I opened a PR on alloy-rs/core with a possible fix for it
   - alloy-rs/core#1102 

I'll take care of tracking alloy's changes & releases to include these fixes in EDR as soon as they are published. Let me know what you think about point (2)

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
```diff
@@ -0,0 +1,318 @@
...
+  userDefinedValueTypeI: UserDefinedValueTypeIndex = new Map(),
+): string | undefined {
+  if (!isObject(typeName)) {
+    return undefined;
+  }
+
+  switch (typeName.nodeType) {
+    case "ElementaryTypeName": {
+      return typeof typeName.name === "string" ? typeName.name : undefined;
+    }
```
schaable, line 216: For EIP-712, `"uint"` and `"int"` shortforms are not valid canonical types. They should be normalized as:

```text
uint  -> uint256
int   -> int256
```

A better approach would be to use `typeDescriptions.typeString`, but you have to be careful with `address payable`, since the `typeString` is `"address payable"` while the canonical EIP-712 type is `"address"`.

```typescript
case "ElementaryTypeName": {
  const desc = isObject(typeName.typeDescriptions)
    ? typeName.typeDescriptions
    : undefined;
  const typeString =
    typeof desc?.typeString === "string" ? desc.typeString : undefined;
  if (typeString !== undefined) {
    return typeString.endsWith(" payable")
      ? typeString.slice(0, -" payable".length)
      : typeString;
  }
  return typeof typeName.name === "string" ? typeName.name : undefined;
}
```

Tests worth adding with this fix:

```text
uint -> "uint256"
int -> "int256"
byte -> "bytes1"
address payable -> "address"
```

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
```diff
@@ -0,0 +1,318 @@
...
+        : undefined;
+      const typeString =
+        typeof desc?.typeString === "string" ? desc.typeString : "";
+      const match = /\[(\d+)\]$/.exec(typeString);
+
+      if (match !== null) {
+        return `${base}[${match[1]}]`;
+      }
+
+      return `${base}[]`;
```
schaable: If I got this right, this is a defensive path. But returning `${base}[]` here would make the encoding succeed with a wrong type. We should return `undefined`, like in the other cases, to drop the struct.

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
```diff
@@ -0,0 +1,318 @@
...
+        return `${base}[]`;
+      }
+
+      if (
+        isObject(length) &&
+        length.nodeType === "Literal" &&
+        typeof length.value === "string"
+      ) {
+        return `${base}[${length.value}]`;
+      }
```
schaable: Solidity accepts hex literals in array lengths, for example `bytes32[0x100] entries;`. If a user writes a hex literal there, this block would produce a non-canonical form. This block can be removed, since the block below already handles this through the array's `typeString`, which canonicalizes the length.

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
```diff
@@ -0,0 +1,318 @@
...
+    }
+
+    case "ArrayTypeName": {
+      const base = encodeMemberType(typeName.baseType, userDefinedValueTypeI);
+      if (base === undefined) {
+        return undefined;
+      }
+
+      const length = typeName.length;
+      if (length === null || length === undefined) {
```
schaable, line 294: I'd assume `length === undefined` is a defensive check. If it is, I'd let it fall through and return `undefined`.

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
```diff
@@ -0,0 +1,318 @@
...
+      const typeString =
+        typeof desc?.typeString === "string" ? desc.typeString : "";
+
+      if (typeString.startsWith("enum ")) {
+        return "uint8";
+      }
+
+      if (
+        typeString.startsWith("contract ") ||
+        typeString.startsWith("interface ")
```
schaable: Do you know in which scenario `"interface "` is emitted?

schaable (commented):
Nice work on this. I'm still missing some tests and part of `canonicalize.ts`, but I wanted to submit my current findings so you can start reviewing them. I'll follow up with the remaining comments separately.

anaPerezGhiglia:
@ChristopherDedominici I ran a Claude-assisted scan over the Foundry-using repos in [this tracker](https://didactic-memory-zrj11yl.pages.github.io/tracker-export-2026-02-10_18-56-15.html) (~170 repos, pure-Foundry + mixed).
`aave/aave-v4` is the only one calling EIP-712 cheatcodes, and it doesn't rely on Foundry's `<Name>_<i>` rename, no `_i` in its generated `JsonBindings.sol`, no `_i` at any cheatcode call site, no `_i` baked into a canonical typehash. 

Considering that, erroring on duplicate struct names (instead of silently renaming) looks sound for the current usage in the sample.

packages/hardhat/src/types/solidity/solidity-artifacts.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/21822a4d5d75232c0e9974a8285cce8a093c3373

packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/9fc53e148a90801075ab289d1f663113adf0ee4a

However, I did not rename the function `solidityTestConfigToSolidityTestRunnerConfigArgs`, to avoid going outside the scope of this PR, which is already relatively big

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/1feac9d311d3edffb558190736f9f0a4cf1d3769

packages/hardhat/src/internal/builtin-plugins/solidity-test/type-extensions.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/3e7fef7d4de4266ae4c9432d85ce965bafdf562f

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/f545768df60a098e3cd5c20f5ba667323c778c9f

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/f545768df60a098e3cd5c20f5ba667323c778c9f

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/883477f0ee29b4770ecde5c756c56b88b16a4316

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/27191bfd7a9de9a87cfe16939ec0fda3522b92e2

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/76523d586e06621fd9ce1b347354fc989e5122e6

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts
ChristopherDedominici: Partially done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/62cf8b9e614e05d215e8c7a949fb912de2daa134.

Follow-up point added in PR description

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/ast-walker.ts
ChristopherDedominici: Actually I have a follow-up fix on this: https://github.com/NomicFoundation/hardhat/pull/8243/changes/6902946ba4a4aa7ff185f4456ade3cc9bcb3e093.

Reaons: solc omits `length` entirely for dynamic arrays; it isn't emitted as `null`. So `undefined` is needed

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
```diff
@@ -0,0 +1,334 @@
...
+ * `resolve_struct_eip712` returns `None` for any struct containing unsupported
+ * constructs and propagates `None` through the dep graph so dependents are
+ * also dropped.
+ *
+ * When `selectedNames` is provided, only those names are emitted; non-selected
+ * structs still participate in dep resolution so cross-file deps inline correctly.
+ */
+export function canonicalizeStructs(
+  structs: CollectedStruct[],
+  selectedNames?: Set<string>,
```
schaable: `selectedNames` is always passed from `collectEip712CanonicalTypes`, so the optional signature and the `undefined` branches only exist to simplify tests. I'd make it required and have tests pass `new Set(collected.map((s) => s.name))`, or use a small helper, for the "emit everything" case.

ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/b100ffd70696b188d9d22d0bdaea696463d86067

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
```diff
@@ -0,0 +1,334 @@
...
+    let base = member.type;
+    while (true) {
+      const match = /^(.*)\[[^\]]*\]$/.exec(base);
+
+      if (match === null) {
+        break;
+      }
+
+      base = match[1];
+    }
```
schaable: Can't you do `member.type.split("[")[0]` instead?

ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/1101e0cea38e4f6c688e22756756df2d0aa5e64e

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
```diff
@@ -0,0 +1,334 @@
...
+      })
+      .filter((s): s is string => s !== undefined);
+
+    deps.sort();
+
+    const canonical = head + deps.join("");
+    if (!seen.has(canonical)) {
+      seen.add(canonical);
+      result.push(canonical);
+    }
```
schaable: This is not needed as `indexByName` already dedupes by name and distinct names produce distinct heads.

ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/cf58e3a7551749fd1df022b0e0b3462055418140

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
```diff
@@ -0,0 +1,322 @@
...
+        continue;
+      }
+
+      const deps = directStructDeps(struct, knownNames);
+      if (deps.some((dep) => !encodable.has(dep))) {
+        encodable.delete(name);
+        changed = true;
+      }
+    }
+  }
```
schaable, line 98: This block recomputes the deps on every iteration, but the deps are always the same for a given struct. We could precompute them to reduce the complexity, something like:

```typescript
  const depsByName = new Map<string, string[]>();
  for (const [name, struct] of byName) {
    depsByName.set(name, directStructDeps(struct, knownNames));
  }
  
  let changed = true;
  while (changed) {
    changed = false;
    for (const name of [...encodable]) {
      const deps = depsByName.get(name);
      assertHardhatInvariant(deps !== undefined, "...");
      if (deps.some((dep) => !encodable.has(dep))) {
        encodable.delete(name);
        changed = true;
      }
    }
  }
```

schaable (approved):
Left one final comment, but I'm pre-approving. Thanks for bearing with me while I reviewed this 😅

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8243/changes/382df0912c917a31d44344cfab41cdc9e8479b8c

PR #8249, opened 2026-05-05
Add receipt-wait option for hardhat-ethers signer

mmv08:

---

## Motivation

`HardhatEthersSigner.sendTransaction` currently mirrors ethers' `JsonRpcSigner`: after `eth_sendTransaction`, it polls `eth_getTransactionByHash` and resolves once the transaction is visible to the node. It does not wait for a receipt.

That behavior is correct as the default because it preserves ethers-compatible pending-transaction semantics. Some Hardhat tests and user code intentionally rely on receiving a `TransactionResponse` before the transaction is mined, especially when automining is disabled.

At the same time, many existing Hardhat test suites were authored against Hardhat's in-process network, where `eth_sendTransaction` effectively behaves synchronously because the transaction is mined before the call returns. When those suites are pointed at external nodes such as anvil, geth, or reth, a common pattern becomes racy:

```ts
await contract.stateChangingMethod();
const value = await contract.read();
```

On external RPC nodes, `eth_sendTransaction` can return after mempool insertion but before mining, so the read may observe pre-transaction state. This makes otherwise useful cross-client or fork-based test runs fail for reasons unrelated to contract behavior.

## What Changed

This PR adds an opt-in, per-network `hardhat-ethers` configuration option:

```ts
networks: {
  anvil: {
    type: "http",
    url: "http://127.0.0.1:8545",
    ethers: {
      waitForTransactionReceipt: true,
    },
  },
}
```

When enabled for a network, `HardhatEthersSigner.sendTransaction` still returns the same `TransactionResponse`, but only after `provider.waitForTransaction(hash)` has observed a receipt.

## Why This Is Safe

- The default remains `false`, so existing ethers-compatible behavior is unchanged.
- The option is scoped per network, so users can enable it only for external-node compatibility runs.
- The implementation is contained within `@nomicfoundation/hardhat-ethers`; it does not change Hardhat core or the network manager.
- The signer waits through `provider.waitForTransaction(hash)` instead of `transactionResponse.wait()`. This is intentional: `TransactionResponse.wait()` throws for `status === 0` receipts, but callers and chai matchers still need the response so they can inspect reverted transactions themselves.
- Existing pending-transaction workflows remain covered by tests that assert the default behavior does not wait for a receipt.

## Testing

From `packages/hardhat-ethers`:

```bash
pnpm lint
pnpm build
node --import tsx/esm --test --test-reporter=@nomicfoundation/hardhat-node-test-reporter "test/contracts.ts" "test/error-messages.ts" "test/gas-price.ts" "test/hardhat-ethers-provider.ts" "test/hardhat-ethers-signer.ts" "test/index.ts" "test/network-config.ts" "test/no-accounts.ts" "test/plugin-functionalities.ts" "test/provider-events.ts" "test/revert-error-cause.ts" "test/transactions.ts"
```

Result: `266 passing`, `2 skipped`.

### DIFF
--- a/packages/hardhat-ethers/src/internal/hook-handlers/config.ts
+++ b/packages/hardhat-ethers/src/internal/hook-handlers/config.ts
@@ -1,0 +1,94 @@
+import type {
+  ConfigurationVariableResolver,
+  HardhatConfig,
+  HardhatEthersNetworkConfig,
+  HardhatEthersNetworkUserConfig,
+  HardhatUserConfig,
+  NetworkConfig,
+} from "hardhat/types/config";
+import type {
+  ConfigHooks,
+  HardhatUserConfigValidationError,
+} from "hardhat/types/hooks";
+
+import { isObject } from "@nomicfoundation/hardhat-utils/lang";
+import { validateUserConfigZodType } from "@nomicfoundation/hardhat-zod-utils";
+import { z } from "zod";
+
+const ethersUserConfigSchema = z.object({
+  ethers: z
+    .object({
+      waitForTransactionReceipt: z.boolean().optional(),
+    })
+    .optional(),
+});
+
+export default async (): Promise<Partial<ConfigHooks>> => ({
+  validateUserConfig,
+  resolveUserConfig,
+});
+
+export async function validateUserConfig(
+  userConfig: HardhatUserConfig,
+): Promise<HardhatUserConfigValidationError[]> {
+  if (userConfig.networks === undefined) {
+    return [];
+  }
+
+  const networksErrors: HardhatUserConfigValidationError[] = [];
+
+  for (const [networkName, network] of Object.entries(userConfig.networks)) {
+    if (!isObject(network)) {
+      continue;
+    }
+
+    const errors = validateUserConfigZodType(
+      { ethers: network.ethers },
+      ethersUserConfigSchema,
+    ).map((err) => {
+      err.message = `network "${networkName}" - ${err.message}`;
+      return err;
+    });
+
+    networksErrors.push(...errors);
+  }
+
+  return networksErrors;
+}
+
+export async function resolveUserConfig(
+  userConfig: HardhatUserConfig,
+  resolveConfigurationVariable: ConfigurationVariableResolver,
+  next: (
+    nextUserConfig: HardhatUserConfig,
+    nextResolveConfigurationVariable: ConfigurationVariableResolver,
+  ) => Promise<HardhatConfig>,
+): Promise<HardhatConfig> {
+  const resolvedConfig = await next(userConfig, resolveConfigurationVariable);
+
+  const networks: Record<string, NetworkConfig> = {};
+
+  for (const [networkName, networkConfig] of Object.entries(
+    resolvedConfig.networks,
+  )) {
+    networks[networkName] = {
+      ...networkConfig,
+      ethers: resolveHardhatEthersNetworkConfig(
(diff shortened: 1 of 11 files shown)


packages/hardhat-ethers/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,96 @@
...
+} from "hardhat/types/hooks";
+
+import { isObject } from "@nomicfoundation/hardhat-utils/lang";
+
+export default async (): Promise<Partial<ConfigHooks>> => ({
+  validateUserConfig,
+  resolveUserConfig,
+});
+
+export async function validateUserConfig(
```
schaable, line 31: The preferred approach for validation is using `validateUserConfigZodType` from `@nomicfoundation/hardhat-zod-utils`. There's an example in `packages/hardhat-ledger/src/internal/config/validate-ledger-user-config.ts`.

packages/hardhat-ethers/src/type-extensions.ts
```diff
@@ -8,3 +8,35 @@ declare module "hardhat/types/network" {
...
 }
+
+declare module "hardhat/types/config" {
+  export interface HardhatEthersNetworkUserConfig {
+    /**
+     * When enabled, HardhatEthersSigner.sendTransaction waits until the
+     * transaction receipt is available before resolving.
+     *
+     * The default is false, matching ethers' JsonRpcSigner behavior.
+     */
```
schaable, line 29:
```suggestion
    /**
      * When enabled, HardhatEthersSigner.sendTransaction resolves only after the
      * transaction has been mined and its receipt is available (one confirmation),
      * instead of as soon as the transaction is visible to the node.
      *
      * Reverted transactions still resolve to a TransactionResponse (the signer
      * does not throw), so callers and chai matchers can inspect the receipt.
      *
      * The default is false, matching ethers' JsonRpcSigner behavior.
      * 
      * Enable this only for compatibility runs against external nodes (anvil, geth,
      * reth) whose eth_sendTransaction returns before mining. Leave it off otherwise,
      * since it changes pending-transaction timing and waits for mining on every send.
      */
```

packages/hardhat-ethers/src/type-extensions.ts
```diff
@@ -8,3 +8,35 @@ declare module "hardhat/types/network" {
...
+    waitForTransactionReceipts?: boolean;
```
schaable: I'd use the singular `waitForTransactionReceipt`, as we only wait for a single receipt per `sendTransaction` call. It also matches ethers' `provider.waitForTransaction`, which is singular as well.

packages/hardhat-ethers/src/internal/hook-handlers/config.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  return {
+    ...resolvedConfig,
+    networks,
+  };
+}
+
+function resolveHardhatEthersNetworkConfig(
+  userConfig: NetworkUserConfig["ethers"] = {},
```
schaable:
```suggestion
  userConfig: HardhatEthersNetworkUserConfig = {},
```

schaable (changes requested):
Hey @mmv08, thanks for the contribution and for submitting this PR. I've left some comments for you to take a look at.

mmv08:
@schaable thanks for the review! i believe i addressed all the points, would appreciate another round 🙏

schaable (approved):

PR #8250, opened 2026-05-05
Handle no-data execution errors in revert matcher

mmv08:
## Why

The broad `.revert(ethers)` matcher is used to assert that contract execution failed. Some clients and provider stacks report EVM execution failures from `eth_call` and `eth_estimateGas` without revert return data, especially for invalid opcode and out-of-gas style failures. Those failures should satisfy the broad matcher, while reason-specific matchers should continue to require actual revert data.

This follows the existing Ignition behavior of normalizing recognized no-data execution failures.

## What changed

- The broad `.revert(ethers)` matcher now accepts recognized no-data EVM execution failures from call and gas-estimation paths.
- `.revertedWith`, `.revertedWithPanic`, `.revertedWithCustomError`, and `.revertedWithoutReason` remain data-driven and still require actual revert return data.
- The `isKnownEvmExecutionErrorMessage` message classifier is an internal helper of `hardhat-ethers-chai-matchers`. No other packages are changed.
- The message classifier matches existing Hardhat/provider execution-failure formats, including no-reason reverts, invalid opcode, out-of-gas, and explicit EVM exceptional halt names such as `InvalidFEOpcode` and `OutOfGas` (a verbatim copy of EDR's `ExceptionalHalt` enum).
- The classifier avoids treating arbitrary capitalized `EVM error ...` messages as execution failures.
- Matcher utility error handling now uses `unknown` inputs and explicit narrowing instead of `any`-based property access.

## Tests

- Added matcher coverage for no-data call exceptions, provider execution errors, nested provider errors, out-of-gas exceptional halts, unrelated provider errors, and the reason-specific matcher invariant.
- Added unit coverage for `isNoDataExecutionError` and for accepted/rejected provider messages in `isKnownEvmExecutionErrorMessage`.

Verification run from `packages/hardhat-ethers-chai-matchers`:

- `pnpm build`
- `pnpm lint`
- `node --import tsx/esm --test --test-reporter=@nomicfoundation/hardhat-node-test-reporter test/matchers/reverted/*.ts`

### DIFF
--- a/.changeset/hardhat-ethers-chai-matchers-no-data-execution-errors.md
+++ b/.changeset/hardhat-ethers-chai-matchers-no-data-execution-errors.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ethers-chai-matchers": patch
+---
+
+The broad `.revert(ethers)` matcher now recognizes EVM execution failures that some providers report without revert data, such as invalid opcode, out-of-gas, and other exceptional halts. Reason-specific matchers like `.revertedWith` are unchanged and still require revert data.
--- a/.changeset/hardhat-ethers-chai-matchers-no-data-execution-errors.md
+++ b/.changeset/hardhat-ethers-chai-matchers-no-data-execution-errors.md
@@ -2,4 +2,6 @@
 "@nomicfoundation/hardhat-ethers-chai-matchers": patch
 ---
 
-The broad `.revert(ethers)` matcher now recognizes EVM execution failures that some providers report without revert data, such as invalid opcode, out-of-gas, and other exceptional halts. Reason-specific matchers like `.revertedWith` are unchanged and still require revert data.
+The broad `.revert(ethers)` matcher now recognizes EVM execution failures that some providers report without revert data, such as invalid opcode and out-of-gas errors. This makes `.revert(ethers)` behave consistently across providers that omit revert data on these failures. Reason-specific matchers like `.revertedWith` are unchanged and still require revert data.
+
+As a consequence, `.not.to.be.revert(ethers)` now fails for these no-data execution failures, since they are treated as reverts.
--- a/packages/hardhat-ethers-chai-matchers/src/internal/matchers/reverted/utils.ts
+++ b/packages/hardhat-ethers-chai-matchers/src/internal/matchers/reverted/utils.ts
@@ -37,15 +37,13 @@ export function getReturnDataFromError(error: unknown): string {
   }
 
   const errorData = getErrorData(error);
-
   if (errorData === undefined) {
     // eslint-disable-next-line no-restricted-syntax -- re-throw because the error is not related to a reverted transaction
     throw error;
   }
 
   const returnData = getReturnData(errorData);
-
-  if (returnData === undefined || typeof returnData !== "string") {
+  if (returnData === undefined) {
     // eslint-disable-next-line no-restricted-syntax -- re-throw because the error is not related to a reverted transaction
     throw error;
   }
@@ -69,6 +67,16 @@ export function isNoDataExecutionError(error: unknown): boolean {
   );
 }
 
+/**
+ * Returns true when the error is the one ethers throws after a failed
+ * `eth_call` or `eth_estimateGas` that came back without any revert data.
+ *
+ * In that case ethers wraps the failure in a `CALL_EXCEPTION` with no data or
+ * reason and the short message "missing revert data", while keeping the
+ * original provider error under `error.info.error`. We only treat it as a
+ * no-data revert when that original error also looks like a known EVM
+ * execution failure (matching by message, and by code when it has one).
+ */
 function isEthersCallExceptionWithoutData(error: Error): boolean {
   if (!isObject(error)) {
     return false;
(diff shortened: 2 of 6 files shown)


packages/hardhat-ethers-chai-matchers/src/internal/matchers/reverted/utils.ts
```diff
@@ -44,6 +42,106 @@ export function getReturnDataFromError(error: any): string {
...
+  return code === 3 || code === -32000 || code === -32003;
+}
+
+function getJsonRpcErrorProperty(error: Error, key: string): unknown {
+  const nestedError = getObjectProperty(error, "error");
+
+  return getObjectProperty(nestedError, key) ?? getObjectProperty(error, key);
+}
+
+function getObjectProperty(value: unknown, key: string): unknown {
```
schaable: We already have `isObject` in `hardhat-utils/lang` (it narrows to `Record<string | symbol, unknown>`, so you can index without an `as`), and we usually walk untyped values using `isObject` + `in`; see `packages/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts`. Could we use that instead of a custom getter? The predicates could read directly from the error and most of these helpers would disappear.

packages/hardhat-utils/src/eth.ts
```diff
@@ -48,6 +48,54 @@ export function toEvmWord(value: bigint | number): PrefixedHexString {
...
+function isEvmExceptionalHaltMessage(message: string): boolean {
+  const match = /^EVM error:?\s+([A-Z][A-Za-z0-9_]*)\b/.exec(message);
+
+  return (
+    match !== null &&
+    match[1] !== undefined &&
+    EVM_EXCEPTIONAL_HALT_MESSAGES.has(match[1])
+  );
+}
+
```
schaable: Minor: we usually put non-exported functions under `internal` (so `internal/eth.ts` in this case). You can create the file if it doesn't exist.

packages/ignition-core/src/internal/execution/jsonrpc-client.ts
```diff
@@ -355,10 +356,7 @@ export class EIP1193JsonRpcClient implements JsonRpcClient {
 
         // Geth, and potentially other nodes, may return an error without a data
         // field if there was no reason returned
-        if (
-          error.message.includes("execution reverted") ||
-          error.message.includes("invalid opcode")
-        ) {
+        if (isKnownEvmExecutionErrorMessage(error.message)) {
```
schaable: @kanej, can you confirm that extending the "allowlist" on this check is fine?

.changeset/hardhat-ethers-chai-matchers-no-data-execution-errors.md
```diff
@@ -0,0 +1,7 @@
+---
+"@nomicfoundation/hardhat-ethers-chai-matchers": patch
+"@nomicfoundation/hardhat-utils": patch
+"@nomicfoundation/ignition-core": patch
```
schaable: I'd use 3 separate changesets:

```
---
"@nomicfoundation/hardhat-ethers-chai-matchers": patch
---

The broad `.revert(ethers)` matcher now recognizes EVM execution failures that some providers report without revert data, such as invalid opcode, out-of-gas, and other exceptional halts. Reason-specific matchers like `.revertedWith` are unchanged and still require revert data.
```

```
---
"@nomicfoundation/ignition-core": patch
---

Ignition now correctly handles EVM execution failures that providers report without revert data, such as out-of-gas and invalid opcode errors, instead of surfacing them as unexpected errors.

```

```
---
"@nomicfoundation/hardhat-utils": minor
---

Added `isKnownEvmExecutionErrorMessage` to detect provider error messages that indicate an EVM execution failure with no return data.

```

packages/hardhat-ethers-chai-matchers/src/internal/matchers/reverted/utils.ts
```diff
@@ -44,6 +42,106 @@ export function getReturnDataFromError(error: any): string {
   return returnData;
 }
 
+/**
+ * Some JSON-RPC clients report EVM execution failures from eth_call or
+ * eth_estimateGas without return data. These can satisfy the broad revert
+ * matcher, but reason-specific matchers still need actual return data.
+ */
+export function isNoDataExecutionError(error: unknown): boolean {
```
schaable, line 59: Can we unit test this?

packages/hardhat-utils/src/eth.ts
```diff
@@ -48,6 +48,54 @@ export function toEvmWord(value: bigint | number): PrefixedHexString {
...
+    ) ||
+    /^VM Exception while processing transaction: (?:invalid opcode|out of gas|reverted\b|Transaction reverted\b)/i.test(
+      message,
+    ) ||
+    /(?:^|:\s*)invalid opcode\b/i.test(message) ||
+    isEvmExceptionalHaltMessage(message)
+  );
+}
+
+const EVM_EXCEPTIONAL_HALT_MESSAGES = new Set([
```
schaable: minor: could we add a comment pointing at EDR's `ExceptionalHalt` enum as the source of truth? This list is a verbatim copy and can silently drift on future EDR bumps.

schaable (changes requested):
Hi @mmv08, thanks for the PR. I've left a few comments when you get a chance to take a look.

packages/ignition-core/src/internal/execution/jsonrpc-client.ts
kanej: The distinction here is between results that mean "the EVM ran and the call failed deterministically", so that we can journal it, and something unexpected happend, and we are going to immediately halt everything. So understood failure that will repeat, vs unexpected exception.

Most of the additions are pure improvement in this regard. But OutOfGas is an issue. We throw as unexpected for those (as in the user need's to get involved now), and this PR changes that behaviour.

mmv08:
@schaable thanks! this is ready for another review round

packages/ignition-core/src/internal/execution/jsonrpc-client.ts
schaable: @mmv08 thanks for adding the `includeOutOfGas` opt-out. That does cover the `OutOfGas` case, but after syncing with @kanej internally, we'd like to take a different approach here.

`OutOfGas` isn't the only case: several of these exceptional halts should remain "unexpected, user needs to step in" on Ignition's side. So a per-call flag doesn't scale very well, and it would be easy for this to drift on future EDR bumps.

Rather than parametrizing Ignition, let's leave Ignition's behavior as it was and scope this PR to the chai matchers only.

Concretely:

- Revert the `ignition-core` changes, keeping the original `execution reverted` / `invalid opcode` checks and no behavior change.
- Move the helper into `hardhat-ethers-chai-matchers` as an internal helper. With Ignition out, the matcher is the only consumer, so we don't need a `hardhat-utils` export.
- Drop `includeOutOfGas`, since the matcher should treat these as reverts.
- Drop the `hardhat-utils` changeset/bump, leaving just the `hardhat-ethers-chai-matchers` patch.

That keeps the change scoped to `hardhat-ethers-chai-matchers` only. Let me know if anything's unclear.

mmv08: hey @schaable thanks and that makes sense! pr updated

schaable (approved):
I pushed a few minor refactors, but this looks good overall. Thanks for all the work on this one.

PR #8262, opened 2026-05-08
Add agents.md and agentic skills to sample projects

fvictorio:
This PR makes the sample projects more agent-friendly by adding an `AGENTS.md` file and some skills. These skills can also be installed in existing projects using [Vercel's skills CLI](https://www.npmjs.com/package/skills).

### Project initialization

Project initialization adds `AGENTS.md` as any other file, but it also creates a `CLAUDE.md` file. How this is done depends on the OS:

- On Linux/macOS, a symlink pointing to `AGENTS.mD` is used.
- On Windows, a file with just `@AGENTS.md` as its content is created. This lets Claude know it should read that file. The alternative here would be to copy the file entirely, but I prefer to use the reference.

Then a `.agents` directory is created with the Hardhat core skill and the skill corresponding to the toolbox they use. A `.claude` directory is also created:

- On Linux/macOS, a symlink pointing to `.agents` is used.
- On Windows, the skills are copied.

### `AGENTS.md`

`AGENTS.md`, like `README.md`, is a file that's assumed to be modified partially or entirely. For that reason, it's relatively simple, just describing the project structure, pointing to the Hardhat skills, and referencing some docs.

### Skills

Skills live in a new directory, `packages/hardhat/skills`, which is bundled with the Hardhat package. This makes them available to the project initialization logic.

There are three skills:
- A skill for Hardhat core
- A skill for the viem toolbox
- A skill for the mocha+ethers toolbox

Each skill references a Hardhat package in its metadata. This is not being used right now, but I think it's a good idea to model them that way. This also might be useful in the future to ship skills with plugins and allow updating them after updating a plugin. Right now that's purely speculative though.

## Testing this

Deploy to verdaccio and initialize a viem or ethers project. Then start an agent and ask it to do something.


.changeset/six-dancers-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
```
alcuadrado, line 2:
```suggestion
"hardhat": patch
"@nomicfoundation/hardhat-utils": patch
```

I think you need this so that the dependency is bumped

packages/hardhat/src/internal/cli/init/init.ts
```diff
@@ -493,9 +497,124 @@ export async function copyProjectFiles(
...
+  workspace: string,
+  template: Template,
+): Promise<void> {
+  if (!template.files.includes("AGENTS.md")) {
+    return;
+  }
+
+  const deps = {
+    ...template.packageJson.dependencies,
+    ...template.packageJson.devDependencies,
```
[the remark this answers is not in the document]
alcuadrado: not true, spreading undefined is ok

alcuadrado (changes requested):
I think the comments about the tests are somewhat valid, and this is missing the interactive init.

alcuadrado:
Otherwise LGTM

.changeset/six-dancers-tie.md
fvictorio: I thought the same, but when I ran `changesets version` locally with separate changesets for each package, it still bumped `hardhat-utils`'s version in Hardhat. Still, I'll apply this change just in case, because it does make sense.

fvictorio: What bothers me (conceptually) is that this means the change will show up in `hardhat-utils`'s changelog. It doesn't matter in this case because it's not a user-facing package, but it's still weird.

packages/hardhat-utils/src/fs.ts
```diff
@@ -837,11 +837,22 @@ export async function getFileSize(absolutePath: string): Promise<number> {
  * Checks if a file or directory exists.
  *
  * @param absolutePath The absolute path to the file or directory.
```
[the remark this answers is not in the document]
fvictorio: I checked and all the usages of `exists` in this PR pass absolute paths. Ignoring this comment.

packages/hardhat-utils/src/fs.ts
```diff
@@ -837,11 +837,22 @@ export async function getFileSize(absolutePath: string): Promise<number> {
...
+): Promise<boolean> {
+  const followSymlinks = options.followSymlinks ?? true;
+
   try {
-    await fsPromises.access(absolutePath);
+    if (followSymlinks) {
+      await fsPromises.access(absolutePath);
+    } else {
+      await fsPromises.lstat(absolutePath);
+    }
```
fvictorio, line 855: @alcuadrado please review this change. I added this after a copilot feedback comment which made sense to me: using `exists` to check a path before creating a symlink is problematic, because broken symlinks cause an error (exists as it was returned "false" for broken symlinks, but then the symlink creation would fail because there was one already; I confirmed this behavior manually).

I added this `followSymlinks` option and set it to false when using `exists` before creating a symlink.

packages/hardhat/src/internal/cli/init/init.ts
```diff
@@ -661,6 +670,129 @@ export async function copyProjectFilesNonInteractive(
...
+  const agentsMdPath = path.join(workspace, "AGENTS.md");
+  if (!(await exists(agentsMdPath))) {
+    return;
+  }
+
+  const claudeMdPath = path.join(workspace, "CLAUDE.md");
+  if (await exists(claudeMdPath, { followSymlinks: false })) {
+    if (force !== true) {
+      return;
+    }
```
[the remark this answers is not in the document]
fvictorio: Ignoring because this is a huge strecth: someone has to have a directory called `CLAUDE.md` with important, non-committed things, then initialize a Hardhat project there, and then choose to override existing files.

packages/hardhat-utils/src/fs.ts
alcuadrado: good catch

alcuadrado (approved):

PR #8275, opened 2026-05-11
feat(e2e): substitute `${localEnv:NAME}` tokens in scenario env

Wodann:

---

> ℹ️ **NOTE**
>
> This PR was preceded by #8274

## Summary

Lets `scenario.json` `env` values reference host environment variables using the devcontainer-style `${localEnv:NAME}` token. CI can then route an existing secret (e.g. `ALCHEMY_URL`) into whatever variable name a scenario expects (e.g. `FOUNDRY_RPC_URL`) without inventing per-scenario secret names.

## Motivation

CI exposes a small set of shared secrets, but external scenarios expect their _own_ variable names — `FOUNDRY_RPC_URL`, `MAINNET_RPC_URL`, etc. Without substitution, each new scenario forces either a new CI secret or a runner-side rename. Now the scenario declares the mapping itself:

```json
"env": {
  "FOUNDRY_RPC_URL": "${localEnv:ALCHEMY_URL}",
  "MAINNET_RPC_URL": "${localEnv:ALCHEMY_URL}"
}
```

## Syntax

Mirrors the [devcontainer.json variable spec](https://containers.dev/implementors/json_reference/#variables-in-devcontainer-json):

- Token: `${localEnv:NAME}`, where `NAME` matches `[A-Za-z_][A-Za-z0-9_]*`.
- Only `env` _values_ are substituted; keys are never touched.
- Multiple tokens per value are supported; surrounding text is preserved (`"prefix-${localEnv:FOO}-suffix"` works).
- A `${...}` token that isn't `localEnv:` is left literal (no error), matching devcontainer's lenient behavior.

## Unset vs. empty

| State of `process.env.NAME` | Behavior |
|---|---|
| **unset** (`undefined`) | Throw, naming both the scenario file and the missing variable |
| **empty string** (`""`) | Substitute `""` |

Rationale: in CI an unset variable almost always means a misconfigured workflow — failing loudly catches that. Empty is rare but legitimate (e.g. opt-out flags) and gets through.

## Implementation

- New `resolveEnv(env, scenarioPath)` in [scripts/end-to-end/helpers/env.ts](scripts/end-to-end/helpers/env.ts).
- Called once inside `loadScenario` ([scripts/end-to-end/helpers/directory.ts](scripts/end-to-end/helpers/directory.ts)), right after `_readScenarioJson`. Every downstream consumer reads the resolved `scenario.definition.env` — single source of truth, no parallel raw-vs-resolved variants.
- Schema's `env.description` updated to document the syntax; type stays `string` (substitution runs after validation, so `isScenarioDefinition` is unchanged).

## Out of scope

- `${containerEnv:...}`, `${localWorkspaceFolder}`, and the default-value form `${localEnv:NAME:default}`.
- Substitution in non-`env` fields (`defaultCommand`, `repo`, etc.).

## Test plan

`pnpm test:scripts` covers, in [scripts/end-to-end/helpers/env.test.ts](scripts/end-to-end/helpers/env.test.ts):

- [x] Passes through values with no tokens.
- [x] Substitutes a single `${localEnv:NAME}`.
- [x] Multiple tokens in one value.
- [x] Tokens across multiple keys.
- [x] Empty-string host value substitutes as `""`.
- [x] Unset host value throws, with scenario path + variable name in the message.
- [x] Unrecognized `${...}` tokens left literal.
- [x] Keys are not substituted.

Tests stub `process.env` and restore in `afterEach`.

### DIFF
--- a/scripts/end-to-end/helpers/directory.ts
+++ b/scripts/end-to-end/helpers/directory.ts
@@ -2,6 +2,7 @@ import path, { basename, dirname, resolve } from "node:path";
 import type { Scenario, ScenarioDefinition } from "../types.ts";
 import { isScenarioDefinition } from "../schema/scenario-schema.ts";
 import { readFileSync } from "node:fs";
+import { resolveEnv } from "./env.ts";
 
 export function loadScenario(
   e2eCloneDirectory: string,
@@ -14,6 +15,10 @@ export function loadScenario(
     scenarioFilePath,
   );
   const definition = _readScenarioJson(scenarioFilePath);
+
+  if (definition.env !== undefined) {
+    definition.env = resolveEnv(definition.env, scenarioFilePath);
+  }
 
   return {
     id,
--- a/scripts/end-to-end/helpers/env.test.ts
+++ b/scripts/end-to-end/helpers/env.test.ts
@@ -1,0 +1,116 @@
+import { afterEach, beforeEach, describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { resolveEnv } from "./env.ts";
+
+describe("resolveEnv", () => {
+  let originalEnv: NodeJS.ProcessEnv;
+
+  beforeEach(() => {
+    originalEnv = process.env;
+    process.env = {};
+  });
+
+  afterEach(() => {
+    process.env = originalEnv;
+  });
+
+  it("passes through values with no tokens", () => {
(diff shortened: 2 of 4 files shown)


kanej (approved):
Nice feature for keeping down the complexity of scenarios.
Looking at it does make me want to review our security around tokens, but this PR has made no change to those boundaries.

PR #8276, opened 2026-05-11
feat(e2e): add --use-local, --force-checkout, --force-publish

Wodann:

---

> ℹ️ **NOTE**
>
> This PR was preceded by #8275

## Summary

Adds three flags to `pnpm e2e`:

- `--use-local` — bump-and-publish any workspace package changed since its last release tag (via the new `pnpm verdaccio publish --since-release`), then rewrite the scenario's `package.json` to pin `hardhat` / `@nomicfoundation/*` deps to whatever Verdaccio now serves.
- `--force-checkout` — let `git checkout <commit>` overwrite uncommitted changes in the scenario working dir.
- `--force-publish` — opt-in escape hatch for publishing to a Verdaccio instance that's already running.

Together they let local and CI-hosted benchmarks flow through the e2e harness end-to-end without manually crafting changesets or hand-bumping versions. Consumed by `pnpm bench` / `pnpm bench:regression` in later PRs in this stack.

### `--use-local`

Without it, `pnpm e2e` runs scenarios against the most-recent **published** versions of Hardhat packages. To benchmark or regression-test against working-tree changes, the runner needs to:

1. Detect what's changed since the last release.
2. Get those changes into the local Verdaccio under a version pnpm will accept.
3. Make the scenario's `package.json` actually pin those bumped versions (otherwise it'd still resolve the npm-cached release version through Verdaccio's proxy).

The first two steps are `sinceReleasePublish()` (added in #8273). Step 3 is the new `upgradeLocalDependencies`: for each `hardhat` / `@nomicfoundation/*` dep in the scenario's `package.json`, query `GET http://127.0.0.1:4873/<pkg>` and rewrite the dep range to the `dist-tags.latest` Verdaccio reports. Because Verdaccio merges locally-published packages with npm-proxied ones, this returns bumped local versions where available and npm versions for everything else — exactly what install would resolve, but pinned so the scenario can't drift.

### `--force-checkout`

`--use-local` patches the scenario's `package.json` in place. On the next run, `git checkout <commit>` would refuse because of the uncommitted edit. `--force-checkout` adds `-f` so the patched file gets blown away and the scenario starts from a clean state. Necessary in CI's regression-benchmark loop where the same clone is reused across runs on our dedicated GitHub runner.

### `--force-publish`

`--use-local` against an already-running Verdaccio is ambiguous — we'd be silently mutating someone else's registry contents. By default this errors out:

```
A Verdaccio instance is already running. Skipping --use-local's bump-and-publish
step — the scenario will use whatever the running registry already contains.
  To force a fresh bump-and-publish to the running instance, pass --force-publish.
  Or stop the running instance first:
    pnpm verdaccio stop
```

`--force-publish` opts back in for cases where overwriting is the intent (e.g. iterating locally on the same Verdaccio across multiple `pnpm e2e` invocations) or when you want to use a locally published version of EDR.

## `init` flow

```
1. if --use-local AND verdaccio running AND NOT --force-publish → log warning, skip publish
2. start Verdaccio if not running
3. if (started or --force-publish):
     if --use-local → sinceReleasePublish()
     else           → verdaccioPublish()
4. setupScenario    (clone / fetch+checkout / clean / submodules / preinstall)
5. if --use-local → upgradeLocalDependencies()
6. installScenarioDeps  (custom install script or installDependencies)
7. if we started Verdaccio → stop it
```

Steps 4–6 used to be a single `initializeScenario`; they're split because `upgradeLocalDependencies` has to run **after** clone/checkout (so it can edit the freshly checked-out `package.json`) but **before** install (so the install resolves against the patched deps).

## Install-time fixes that came along for the ride

`installDependencies` ([scripts/end-to-end/helpers/install.ts](scripts/end-to-end/helpers/install.ts)) gains an `allowLockfileUpdates: boolean` (true when `--use-local`, since the dep-pinning edit drifts the lockfile):

- pnpm / bun → append `--no-frozen-lockfile`
- yarn berry → append `--no-immutable`
- npm → no flag needed (`npm install` never freezes; only `npm ci` does)

Two install-time env vars are now always set:

- `npm_config_minimum_release_age=0` — pnpm 10+ rejects packages published less than 7 days ago by default; `sinceReleasePublish` publishes bumped versions seconds before install runs, so this would otherwise fail.
- `npm_config_registry=<VERDACCIO_URL>` — Yarn Classic doesn't honor project `.yarnrc.yml` for `registry`, so we force it via env (highest-priority npm config source).

`fetch()` also now takes the target commit SHA and fetches it specifically, so checkout succeeds even when the SHA has been orphaned upstream (force-pushed branch, etc.).

## Test plan

- [x] `pnpm test:scripts` — new [scripts/end-to-end/helpers/install.test.ts](scripts/end-to-end/helpers/install.test.ts) covers all 4 package managers × allow/disallow lockfile updates for `getInstallArgs`.
- [x] Manually used all commands extensively while testing performance regression benchmark

### DIFF
--- a/scripts/end-to-end/helpers/install.ts
+++ b/scripts/end-to-end/helpers/install.ts
@@ -15,6 +15,7 @@ import type { ScenarioDefinition } from "../types.ts";
 export function installDependencies(
   workDir: string,
   packageManager: ScenarioDefinition["packageManager"],
+  allowLockfileUpdates: boolean,
   env?: Record<string, string>,
 ): void {
   writeRegistryConfig(workDir, packageManager);
@@ -30,17 +31,60 @@ export function installDependencies(
 
   logStep("Installing dependencies");
 
-  const installArgs =
-    packageManager === "yarn"
-      ? ["install"]
-      : // Required by bun as it may not pickup the cwd bunfig.toml
-        ["install", `--registry=${VERDACCIO_URL}`];
+  const installArgs = getInstallArgs(
+    packageManager,
+    allowLockfileUpdates,
+    VERDACCIO_URL,
+  );
 
   execFileSync(which(packageManager), installArgs, {
     cwd: workDir,
     stdio: "inherit",
-    env: { ...process.env, ...env, COREPACK_ENABLE_DOWNLOAD_PROMPT: "0" },
+    env: {
+      ...process.env,
+      ...env,
+      COREPACK_ENABLE_DOWNLOAD_PROMPT: "0",
+      npm_config_minimum_release_age: "0",
+      // Yarn Classic doesn't honor project .yarnrc for `registry` on the
+      // GitHub Actions runner, so force the registry via env var. This is
+      // the highest-priority npm config source, so it overrides any
+      // user/global .npmrc that might be present.
+      npm_config_registry: VERDACCIO_URL,
+    },
   });
+}
+
+/**
+ * Build the package-manager-specific `install` args for a Verdaccio-backed
+ * install.
+ *
+ * `--use-local` patches the scenario's package.json, drifting the lockfile.
+ * Since CI defaults to frozen-lockfile mode for pnpm and Yarn Berry, we allow
+ * lockfile updates.
+ */
+export function getInstallArgs(
+  packageManager: ScenarioDefinition["packageManager"],
+  allowLockfileUpdates: boolean,
+  registryUrl: string,
+): string[] {
+  // bun doesn't reliably read cwd `bunfig.toml`, so it needs the `--registry` CLI flag.
+  // npm & pnpm don't strictly need `--registry` but we pass it for redundancy.
+  // yarn is excluded because it rejects `--registry` as a CLI flag.
+  const args =
+    packageManager === "yarn"
+      ? ["install"]
+      : ["install", `--registry=${registryUrl}`];
+
+  if (allowLockfileUpdates) {
+    // npm install never freezes (only `npm ci` does), so it doesn't need any flag.
+    if (packageManager === "pnpm" || packageManager === "bun") {
+      args.push("--no-frozen-lockfile");
+    } else if (packageManager === "yarn") {
+      args.push("--no-immutable");
+    }
+  }
+
+  return args;
 }
 
 function writeRegistryConfig(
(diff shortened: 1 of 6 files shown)


scripts/end-to-end/helpers/install.ts
```diff
@@ -30,19 +31,62 @@ export function installDependencies(
...
 
   execFileSync(which(packageManager), installArgs, {
     cwd: workDir,
     stdio: "inherit",
-    env: { ...process.env, ...env, COREPACK_ENABLE_DOWNLOAD_PROMPT: "0" },
+    env: {
+      ...process.env,
+      ...env,
+      COREPACK_ENABLE_DOWNLOAD_PROMPT: "0",
+      npm_config_minimum_release_age: "0",
```
kanej, line 47: I was wondering whether we could tighten this so that minimum release age was still applied for non-hardhat packages, but as we support yarn and other package managers I think it would be a moot point.

scripts/end-to-end/subcommands/init.ts
```diff
@@ -1,19 +1,69 @@
...
+  publish as verdaccioPublish,
+  sinceReleasePublish,
+} from "../../verdaccio/publish.ts";
 import { stop as verdaccioStop } from "../../verdaccio/stop.ts";
 import type { Scenario } from "../types.ts";
 
+export const ForceCheckout = {
+  Yes: "Yes",
+  No: "No",
+} as const;
```
kanej, line 23: Nit: I would prefer an options object if the goal is to clarify between arguments at the call point.

kanej (approved):

scripts/end-to-end/subcommands/init.ts
Wodann: What would that look like? Something like this:

```
interface Options {
  forceCheckout: bool,
  forcePublish: bool,
}
```

scripts/end-to-end/helpers/install.ts
Wodann: My original attempt was to use the `preinstall.sh` to modify the pnpm configuration file for the third-party repo by insert an exception for HH and EDR.

When this failed, I researched an alternative. `npm_config_minimum_release_age` was the only way to override all package managers.

If you'd like, I can do one of two things (or both):
1. investigate a method to override as many of the package managers by using the scenario's `preinstall.sh` (npm, pnpm and yarn Berry support it according to this [source](https://gist.github.com/mcollina/b294a6c39ee700d24073c0e5a4e93104/)); only opting for `npm_config_minimum_release_age` when per-project configuration is not possible
2. `npm_config_minimum_release_age: "0",` is only necessary when using `--use-local`. We could at least gate the environment variable behind that boolean flag.

WDYT?

PR #8284, opened 2026-05-11
test(e2e): add lidofinance-dual-governance scenario

Wodann:

---

> ℹ️ **NOTE**
>
> This PR was preceded by #8283

Adds the `lidofinance/dual-governance` end-to-end scenario.

### DIFF
--- a/end-to-end/lidofinance-dual-governance/scenario.json
+++ b/end-to-end/lidofinance-dual-governance/scenario.json
@@ -1,0 +1,20 @@
+{
+  "$schema": "../../scripts/end-to-end/schema/scenario.schema.json",
+  "description": "Lido Dual Governance Smart contracts fork to Hardhat 3.",
+  "tags": ["external-repo"],
+  "repo": "popescuoctavian/dual-governance",
+  "commit": "ef72db01b9b69fddc7fefae9da160e465921ec42",
+  "packageManager": "npm",
+  "submodules": true,
+  "env": {
+    "MAINNET_RPC_URL": "${localEnv:ALCHEMY_URL}"
+  },
+  "defaultCommand": "npx hardhat test solidity --grep=\"^(test_|testFuzz_)\"",
+  "benchmark": {
+    "runs": {
+      "coldCompile": 3,
+      "warmCompile": 83,
+      "defaultCommand": 3
+    }
+  }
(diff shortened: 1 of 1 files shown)


kanej (approved):
The default command runs, but there are failing Solidity tests due to missing cheatcodes. We can fix this by supporting better filtering on the tests.

Wodann:
> The default command runs, but there are failing Solidity tests due to missing cheatcodes. We can fix this by supporting better filtering on the tests.

@kanej  That sounds like a regression. Did it happen to be the snapshot-related cheatcodes?

If so, a fix for that should be in the latest EDR release. I can check after https://github.com/NomicFoundation/hardhat/pull/8298 has been merged and re-enable this scenario as well as aave-v4.

kanej:
> > The default command runs, but there are failing Solidity tests due to missing cheatcodes. We can fix this by supporting better filtering on the tests.
> 
> @kanej That sounds like a regression. Did it happen to be the snapshot-related cheatcodes?
> 
> If so, a fix for that should be in the latest EDR release. I can check after #8298 has been merged and re-enable this scenario as well as aave-v4.

On thinking about it, I suspect that the benchmark does the local update, but my standard run is using the repo lockfiles pointing at an older EDR. Maybe a `pre-install` update of the dependency version or ignoring of the lockfile would work.

In general, I want to get to all scenarios running successfully to completion through their default command with:

```
pnpm e2e exec --scenario ./scome/scenario
```

Further more, that you then rerun it with edits with something like:

```
pnpm e2e exec --scenario ./scome/scenario --since-release
```

But to achieve that means either:

* the scenario repo has been fully prepared to allow both
* the scenario setup script (+ preparation) fully allow that

I like the idea of keeping our need to fork and create modified versions of the repo to a minimum (so place more weight on preparation scripts) but I think both approaches should be expected in practice.

More concretely, if you want to re-enable this one for the benchmark, that works. I can take a look at the changes that acceptance testing invocations of `--exec` would need.

Wodann:
I re-enabled the scenario in https://github.com/NomicFoundation/hardhat/pull/8300 and it's passing successfully in CI.

Could it be that the stale local state of your E2E scenario in `/tmp/end-to-end/` is causing this to fail?

PR #8288, opened 2026-05-12
fix: suppress warnings from coverage-generated files

### LINKED ISSUE #8251: Suppress solidity compilation warnings when running with `--coverage`
## Context
When running with `--coverage`, hardhat outputs some compilation warnings that are a direct consequence of instrumenting the code for coverage and adding `coverage.sol` library into the compilation input:

1. Deprecated memory-safe-assembly special comment

   ```
     --> __hardhat_coverage_library_a3e9cfe2-41b4-4a1f-ad9e-ac62dd82979e.sol:11:5:
      |
   11 |     assembly {
      |     ^ (Relevant source part starts here and spans across multiple lines).
   ```
    
2.  Contract size warnings when compiling without coverage emitted no warning

   ```
   Warning: Contract code size is 25002 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on Mainnet
   ```

  - This Is an undesired side-effect of the source-code-based test instrumentation

## Requested change

HH should filter both these warnings all together **when running with  `--coverage`**. 
Analogous to #7842

JunaidCD:
* [x] Because this PR includes a **bug fix**, relevant tests have been included.

---

Resolves #8251

This PR suppresses Solidity compiler warnings originating from coverage-generated files whose paths contain `__NomicFoundationCoverage`.

The implementation adds a minimal early-return check inside `shouldSuppressWarning` in `warning-suppression.ts`, avoiding changes to the existing typed suppression rule structure.

### Changes

* Added suppression for warnings whose paths include `__NomicFoundationCoverage`
* Added positive and negative tests covering the new suppression behavior

### Verification

* `pnpm lint`
* `pnpm build`
* targeted warning-suppression test suite

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
@@ -55,6 +55,10 @@ export function shouldSuppressWarning(
   absoluteSolidityTestsPath: string,
   absoluteProjectRoot: string,
 ): boolean {
+  if (errorMessage.includes("__NomicFoundationCoverage")) {
+    return true;
+  }
+
   // Compute relative path from project root to test directory.
   // Example:
   // absoluteSolidityTestsPath: /workspaces/hardhat-4/packages/example-project/test/contracts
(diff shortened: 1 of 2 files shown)


ChristopherDedominici:
Hi @JunaidCD, thanks for the PR! I took a look, and there were a quite a few additional changes needed on top of what you did ([see comment](https://github.com/NomicFoundation/hardhat/issues/8251#issuecomment-4419546979)). I went ahead and created another PR https://github.com/NomicFoundation/hardhat/pull/8302 so I was able to iterate faster

PR #8290, opened 2026-05-13
fix(solidity-test): skip abstract contracts in test suite discovery

### LINKED ISSUE #8203: Don't run Solidity tests in `abstract` contracts as their own test suite
### Version of Hardhat

3.4.2

### What happened?

Hardhat currently runs test functions from `abstract contract` as their own test suite. This should not be the case.

Suggestion: It looks like Hardhat only [checks function prefix](https://github.com/NomicFoundation/hardhat/blob/c0eb92a957f101253812c2acdb228ebf198b8118/packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts#L151) and doesn't implement any `abstract` checks. I believe that adding a check for `!bytecode.is_empty()` is enough to fix this issue.

AbtractTest.t.sol:
```solidity
import {Test} from "forge-std/Test.sol";

abstract contract AbstractTest is Test {
    function test_abstact() public {
        assertTrue(true);
    }
}

contract ActualTest is AbstractTest {
    function test_actual() public {
        assertTrue(true);
    }
}
```

`npx hardhat test solidity`:

  test/AbstractTest.t.sol:AbstractTest -- suite for `abstract contract` test
    ✔ test_abstact()

  test/AbstractTest.t.sol:ActualTest
    ✔ test_actual()
    ✔ test_abstact()

### Search terms

_No response_

gultekinmakif:

---

## Summary

This PR addresses a bug:
> Hardhat currently runs test functions from abstract contract as their own test suite. This should not be the case.

## Fix

Prevent abstract contracts (& interfaces) from being reported as their own suite, by filtering out artifacts with zero bytecode.

## Context

`isTestSuiteArtifact` decides whether a compiled contract is a Solidity test suite by checking the ABI for function names starting with `test` or `invariant`. Abstract contracts (and interfaces) compile to no bytecode but can still expose `test*` / `invariant*` ABI entries via inheritance, so they were surfaced as standalone phantom test suites alongside the concrete contracts that inherit from them.

This PR adds an early-exit guard at the top of `isTestSuiteArtifact`: if `artifact.contract.bytecode === "0x"`, the artifact has no deployable form and is skipped.

## Tests
Added tests covering all cases related to `isTestSuiteArtifact` function:

- Concrete contract with `test_*` ABI entries + non-empty bytecode: identified as a test suite.
- Concrete contract with no `test*` / `invariant*` ABI entries: NOT identified.
- Abstract contract (`bytecode: "0x"`) with `test_*` ABI entries: NOT identified (regression test for #8203).

## Changeset

`patch` on `hardhat`.

## Notes

* Closes #8203

### DIFF
--- a/.changeset/quiet-pandas-dance.md
+++ b/.changeset/quiet-pandas-dance.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed `abstract` contracts being picked up as test suites by the Solidity test runner (#8203).
--- a/packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
@@ -150,7 +150,9 @@ export async function solidityTestConfigToSolidityTestRunnerConfigArgs({
 
 export function isTestSuiteArtifact(artifact: Artifact): boolean {
   const bytecode = artifact.contract.bytecode;
-  if (bytecode === "" || bytecode === "0x") {
+
+  // Skip abstract contracts and interfaces i.e. those with no bytecode
+  if (bytecode === "" || bytecode === "0x" || bytecode === undefined) {
     return false;
   }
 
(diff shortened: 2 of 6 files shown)


kanej (approved):
Thanks @gultekinmakif. I added a comment and an integration test, but this looks great.

PR #8291, opened 2026-05-13
fix: suppress warnings from the injected coverage library file

### LINKED ISSUE #8251: Suppress solidity compilation warnings when running with `--coverage`

gultekinmakif:

---

## Summary

This PR addresses a bug:
> `--coverage` flag surfaces solc warnings emitted against the synthetic `__NomicFoundationCoverage` library. This is injected by the coverage hook. Users cannot fix these issues, so there is no point of polluting with these warnings.

## Context

When the coverage hook runs, EDR's synthetic library is injected into the solc input. solc compiles it alongside user sources and emits warnings against it:
```
  --> __NomicFoundationCoverage-1fe87c59-dedc-4831-8918-604bc223bbfa.sol:11:5:
```
The build-system warning pipeline has no rule for this file, so the warnings reach stdout. 

## Fix
* Extended `warning-suppression.ts` with a new `coverage-library` scope that drops any warning whose error text contains `COVERAGE_LIBRARY_FILE_NAME`(i.e., `__NomicFoundationCoverage`).

* Additionally, the rules(`SUPPRESSED_WARNINGS`) are split into three sub-arrays`SPECIFIC_FILE_RULES`, `TEST_FILE_RULES`, `COVERAGE_LIBRARY_RULES` — so each scope keeps a narrow type(TS9017).

## Tests

Unit tests added in `packages/hardhat/test/internal/builtin-plugins/solidity/build-system/warning-suppression.ts`. Integration test (`integration/warning-filtering.ts`) updated for the new sub-array exports.

All tests confirmed.

## Changeset

`patch` on `hardhat`.

## Notes

* Closes #8251

### DIFF
--- a/.changeset/suppress-coverage-library-warnings.md
+++ b/.changeset/suppress-coverage-library-warnings.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Suppress Solidity compiler warnings emitted against the synthetic `__NomicFoundationCoverage` library file injected by the coverage hook, so `--coverage` runs no longer surface warnings users cannot fix.
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/warning-suppression.ts
@@ -16,6 +16,13 @@ import { COVERAGE_LIBRARY_FILE_NAME } from "@nomicfoundation/edr";
 //        (e.g., test/contracts/Example.sol)
 //    - Use this for warnings that are acceptable in test code but not in production code
 //      (e.g., missing SPDX license identifiers or pragma statements)
+//
+// 3. scope: 'coverage-library' - Suppress warnings from the injected coverage library file
+//    - The file is identified by source paths containing COVERAGE_LIBRARY_FILE_NAME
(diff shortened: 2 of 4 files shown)


ChristopherDedominici:
Hi @gultekinmakif, thanks for the PR! I took a look, and there were a few additional changes needed on top of what you did. I went ahead and created another PR [here](https://github.com/NomicFoundation/hardhat/pull/8302) so I was able to iterate faster. I've referenced this PR in the description.

PR #8294, opened 2026-05-13
fix(reporter): display viem error causes

### LINKED ISSUE #8237: Display viem error causes in node test runner
The change to not loose viem error causes means we can leverage the additional information in the `node:test` runner.

Before:

Image

Image

After:

Image

gultekinmakif:

---

## Summary
This PR addresses a bug:
> When a viem contract call reverts, the node test reporter prints viem's full wrapper cause chain without surfacing the underlying Solidity stack trace as its own section.

The reporter now detects a `SolidityError` anywhere in an error's cause chain and renders a dedicated `Solidity stack trace:` section, replacing the noisy wrapper layers, matching the **After** screenshot in the linked issue.

## Fix

`formatSingleError` in `hardhat-node-test-reporter` now:

1. Walks the `.cause` chain looking for an error with `name === "SolidityError"`.
2. When found, extracts the Solidity-side frames from that error's `.stack` (the leading callsites prepended by EDR's `Error.prepareStackTrace`) and renders them as a `Solidity stack trace:` section.
3. Suppresses the recursive `[cause]:` rendering of intermediate viem/ethers wrapper layers.
4. Strips `.sol`-location frames from the top-level stack when the section will be emitted, so they don't double-render in the ethers case (where the top-level thrown error is itself the `SolidityError`).

Subsequent commits address edge cases initially missed.

## Tests
Relevant unit tests are added in `test/error-formatting.ts`.

## Notes
* Closes #8237

### DIFF
--- a/.changeset/mad-bears-rule.md
+++ b/.changeset/mad-bears-rule.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-node-test-reporter": patch
+---
+
+Surfaced Solidity stack traces from `SolidityError` causes in the node test reporter. Failures from viem and ethers contract calls now render a dedicated `Solidity stack trace:` section instead of a deeply nested viem cause chain.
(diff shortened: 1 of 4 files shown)


alcuadrado:
I have almost the same change locally 🤯. I'll review this soon

alcuadrado:
I pushed my changes in https://github.com/NomicFoundation/hardhat/pull/8345

Thanks for sending this PR :)

PR #8296, opened 2026-05-13
hardhat-solx: emit solx debugInfo for EDR Solidity stack traces

marianfe:
Adds `evm.{deployed,}Bytecode.debugInfo` output selectors to solx compiles in order to output debug info used for stack traces. 
Drops Solidity 0.8.33 / solx 0.1.3 from the supported map (predates DWARF). 

Companion EDR PR: https://github.com/NomicFoundation/edr/pull/1425

### DIFF
--- a/.changeset/solx-debug-info-output-selection.md
+++ b/.changeset/solx-debug-info-output-selection.md
@@ -1,5 +1,0 @@
----
-"@nomicfoundation/hardhat-solx": minor
----
-
-Introduce support for Solidity stack traces for solx-compiled tests using solx's 0.1.4+ DWARF format. Drops unsupported Solidity version 0.8.33 (solx 0.1.3) due to no debug info.
--- a/.changeset/solx-debug-info-output-selection.md
+++ b/.changeset/solx-debug-info-output-selection.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-solx": minor
+---
+
+Introduce support for Solidity stack traces for solx-compiled tests using solx's 0.1.4+ DWARF format. Drops unsupported Solidity version 0.8.33 (solx 0.1.3) due to no debug info.
--- a/packages/hardhat-solx/src/internal/hook-handlers/config.ts
+++ b/packages/hardhat-solx/src/internal/hook-handlers/config.ts
@@ -23,6 +23,7 @@ import {
   SOLX_COMPILER_TYPE,
   SUPPORTED_SOLX_EVM_VERSIONS,
 } from "../constants.js";
+import { addSolxDebugInfoSelectors } from "../solx-compiler.js";
 
 const log = createDebug("hardhat:solx:hook-handlers:config");
 
@@ -172,10 +173,17 @@ export async function resolveUserConfig(
 ): Promise<HardhatConfig> {
   const resolvedConfig = await next(userConfig, resolveConfigurationVariable);
 
+  // Add solx debugInfo selectors so the cached solcInput, build-info,
+  // and build-ID hash all include them.
+  const profiles = await augmentSolxOutputSelectionInProfiles(
+    resolvedConfig.solidity.profiles,
+  );
+
   return {
     ...resolvedConfig,
     solidity: {
       ...resolvedConfig.solidity,
+      profiles,
       registeredCompilerTypes:
         resolvedConfig.solidity.registeredCompilerTypes.includes(
           SOLX_COMPILER_TYPE,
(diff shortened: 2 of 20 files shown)


.changeset/solx-debug-info-output-selection.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-solx": minor
+---
+
+Introduce support for Solidity stack traces for solx-compiled tests using solx's 0.1.4+ DWARF format. Drops unsupported Solidity version 0.8.33 (solx 0.1.3) due to no debug info.
```
marianfe: TODO: remove the changeset file as discussed in the sync

packages/hardhat-solx/src/internal/solx-compiler.ts
kanej:
```suggestion
    // Merge default solx settings with user settings. User settings take
    // precedence, allowing overrides of viaIR, LLVMOptimization, etc.
    // Note: outputSelection includes solx selectors (added in resolveUserConfig).
const modifiedInput: CompilerInput = {
...input,
settings: {
...this.#extraSettings,
...input.settings,
},
};
```

I think the previous comment is more accurate, `input.settings` takes precendence over `#extraSettings`.
I don't think the `ouputSelection` part is needed, but maybe just as an addition.

packages/hardhat-solx/src/internal/solx-compiler.ts
```diff
@@ -4,8 +4,15 @@ import type {
   CompilerOutput,
 } from "hardhat/types/solidity";
 
+import { deepClone } from "@nomicfoundation/hardhat-utils/lang";
 import { spawnCompile as defaultSpawnCompile } from "hardhat/internal/solidity";
 
+/** solx 0.1.4+ source-mapping selectors (DWARF in `debugInfo`). */
```
kanej:
```suggestion
// Selectors for enabling DWARF `debugInfo` output
```
I don't think we need to mention the version changes.

packages/hardhat-solx/src/internal/solx-compiler.ts
```diff
@@ -44,3 +61,39 @@ export class SolxCompiler implements Compiler {
...
+    : {};
+  const cloned: Record<string, Record<string, string[]>> = await deepClone(
+    seed,
+  );
+
+  // Hardhat normalizes outputSelection to populate `["*"]["*"]` upstream, but
+  // unit tests construct the input directly with `{}`, so make sure the slot
+  // exists before we push.
+  cloned["*"] ??= {};
+  cloned["*"]["*"] ??= [];
```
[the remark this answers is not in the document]
kanej: I think this is unlikely to happen in practice.

packages/hardhat-solx/src/internal/hook-handlers/config.ts
```diff
@@ -190,6 +198,56 @@ export async function resolveUserConfig(
   };
 }
 
+/**
+ * For each compiler entry whose `type === "solx"`, augments
+ * `settings.outputSelection` with the solx debugInfo selectors.
+ * Non-solx entries pass through unchanged.
+ */
+async function augmentSolxOutputSelectionInProfiles(
```
kanej, line 206: At the architecture level, my only real query is whether config setup is the place we expected these defaults, the solx debug info selectors for output, to be added.

Without too much thought I expected `preprocessSolcInputBeforeBuilding` to be where you would add always enabled solx compiler config. But that is a deprecated hook.

@alcuadrado are we missing a hook for altering the config passed into the compiler or do you expect that config resolution should elaborate the compiler config?

kanej (approved):
I have done a code level review and I think this looks good for merging. We can do a round of manual tests in preparation for the release.

My only real question is this one https://github.com/NomicFoundation/hardhat/pull/8296#discussion_r3340582550. I am happy with the resolution in config (hence pre-approving), but I want to confirm this is not a missed requirement of our build hooks.

packages/hardhat-solx/src/internal/hook-handlers/config.ts
alcuadrado: No, actually this is what i was expecting. We also resolve some defaults for solc.

Maybe the framing that's not clearly explained anywhere is that in a previous PR from @marianfe we introduced the concept of a Solidity compiler, and solc is an implementation of that. The config is now a SolidityCompilerConfig, and the type distinguishes which kind of solidity compiler is being used, and specialize the config types.

A way to think about it is that the config types now are a discriminated union by `type`.

kanej: That answers my question. Thanks.

packages/hardhat-solx/src/internal/solx-compiler.ts
marianfe: thanks, fixed

.changeset/solx-debug-info-output-selection.md
marianfe: done

packages/hardhat-solx/src/internal/solx-compiler.ts
marianfe: agree, fixed

marianfe:
thanks for the review & comments, @kanej ! Pato answered https://github.com/NomicFoundation/hardhat/pull/8296#discussion_r3340582550 , and i've fixed the remaining comments , so will go ahead and merge

Issue #8297, opened 2026-05-13
Upgrade to EDR v0.12.0-next.33


Wodann:
[EDR v0.12.0-next.33](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.33) released with the following changeset:

### Minor Changes

- 8e32c14:
  - Added `LocalConfig` to specify configuration for locally mined blockchains.
  - Added `MiningConfig.blockGasLimit` to optionally specify a block gas limit. When omitted, block gas limit enforcement is disabled.
  - Added `ProviderConfig.defaultTransactionGasLimit` to specify a default transaction gas limit to use for RPC call and transaction requests when a `gas` value is not specified.
  - Added `ProviderConfig.network` to specify whether a fork or locally mined blockchain is used.
  - Changed type of `ProviderConfig.transactionGasCap` to also accept `false`. This allows disabling of the transaction gas cap.
  - BREAKING CHANGE: Removed `ProviderConfig.blockGasLimit`. Instead:
    - Use `MiningConfig.blockGasLimit` to set the block gas limit used for mining blocks.
    - Use `ProviderConfig.defaultTransactionGasLimit` to set the default transaction gas limit used for RPC call and transaction requests when a `gas` value is not specified.
    - Use `LocalConfig.genesisBlockGasLimit` to set the block gas limit of the genesis block for locally mined blockchains.
  - BREAKING CHANGE: Removed `ProviderConfig.fork`. Instead, use `ProviderConfig.network` with a `ForkConfig` object.
  - BREAKING CHANGE: Removed `ProviderConfig.initialBlobGas`. Instead, use `LocalConfig.genesisBlobGas` to set the initial blob gas for locally mined blockchains.
  - BREAKING CHANGE: Removed `ProviderConfig.initialDate`. Instead, use `LocalConfig.genesisDate` to set the initial date for locally mined blockchains.
  - Added `SolidityTestRunnerConfigArgs.transactionGasCap` (defaults to the hardfork default).
  - BREAKING CHANGE: Renamed `SolidityTestRunnerConfigArgs.enableTxGasLimitCap` to `SolidityTestRunnerConfigArgs.disableTransactionGasCap` (still defaults to `false`, but the meaning is inverted; i.e. the transaction gas cap is enabled by default).

### Patch Changes

- af5dc7f: Fixed a bug when using a test filter where the `setUp` function would be run for empty test suites
[...]

PR #8298, opened 2026-05-13
build: upgrade to EDR v0.12.0-next.33

Wodann:

---

This PR upgrades to [EDR v0.12.0-next.33](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.12.0-next.33); updating configuration to address breaking changes, while maintaining same behaviour.

Follow-up work is outlined in #8297

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
@@ -34,7 +34,10 @@ import { ensureError } from "@nomicfoundation/hardhat-utils/error";
 import { numberToHexString } from "@nomicfoundation/hardhat-utils/hex";
 
 import { sendErrorTelemetry } from "../../../cli/telemetry/error-reporter/reporter.js";
-import { EDR_NETWORK_REVERT_SNAPSHOT_EVENT } from "../../../constants.js";
+import {
+  EDR_NETWORK_REVERT_SNAPSHOT_EVENT,
+  OPTIMISM_CHAIN_TYPE,
+} from "../../../constants.js";
 import { hardhatChainTypeToEdrChainType } from "../../../edr/chain-type.js";
 import { getGlobalEdrContext } from "../../../edr/context.js";
 import { BaseProvider } from "../base-provider.js";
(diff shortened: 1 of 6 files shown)


packages/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -464,26 +465,47 @@ export async function getProviderConfig(
...
+      ? // OP stack is yet to activate EIP-7825. Slated for Upgrade 19 (activation TBD as of 2026-05-14)
+        networkConfig.blockGasLimit
+      : hardforkGte(
+            networkConfig.hardfork,
+            L1HardforkName.OSAKA,
+            networkConfig.chainType,
+          )
+        ? 16_777_216n // EIP-7825 transaction gas cap
+        : networkConfig.blockGasLimit;
+
```
Wodann, line 496: This logic mirrors the logic that was removed from EDR to maintain backwards compatibility:

```diff
-        let gas_limit = gas.unwrap_or_else(|| {
-            data.transaction_gas_cap()
-                .unwrap_or_else(|| data.block_gas_limit())
-        });
+        let gas_limit = gas.unwrap_or_else(|| data.default_transaction_gas_limit());
```

https://github.com/NomicFoundation/edr/pull/1389/changes#diff-d8c419823e2bff77088e697d09d23b5f5907ebc448323f1af1f13b9cd6c3ec5eL124-L127

It also accounts for the fact that the OP chain type doesn't support EIP-7825 yet.

Wodann: https://github.com/NomicFoundation/hardhat/issues/8297 already mentions that Hardhat 3 might want to expose a configuration option dedicated to setting the `defaultTransactionGasLimit` to make this more transparent to the user.

As explained in [this design doc](https://www.notion.so/nomicfoundation/EDR-API-for-disabling-transaction-gas-cap-block-gas-limit-35a578cdeaf5803a88cff5a299a61dd8), this behaviour mimics Geth's `RpcGasCap` configuration option.

kanej: Thanks for putting in the pre-Osaka/OP support. It makes sense.
I will push the tests against third party repos to the follow up PR adding our `--coverage` changes.

kanej (approved):

PR #8299, opened 2026-05-14
Add Node 26 to v2's CI

alcuadrado:


.github/workflows/hardhat-ci.yml
```diff
@@ -72,7 +72,7 @@ jobs:
       matrix:
         package: ${{ fromJson(needs.list-packages.outputs.packages) }}
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node: [22.0.0, 24.0.0]
+        node: [22.0.0, 24.0.0, 26.0.0]
```
alcuadrado, line 75: I switched this to use 26 and exclude 24

kanej (approved):

PR #8301, opened 2026-05-15
disable gas limit for coverage

kanej:
This PR disables the per transaction gas cap (introduced in Osaka) and the block gas limit when running either JS/TS tests or Solidity Tests with the coverage flag (i.e. `--coverage`).

Solidity Test runs with both disabled by default, whether running with coverage or not. However both were previously enabled by default for JS/TS Tests.

To support that change the following changes have been made within Hardhat config:

```ts
networks: {
  "my-network": {
    blockGasLimit?: number | bigint | false, # now supports false to disable the check
    transactionGasCap?: number | bigint | false # added config option
  }
},
test: {
  solidity: {
    blockGasLimit?: number | bigint | false, # added number to align with network config
    transactionGasCap?: number | bigint | false # added config option
  }
}
```

For JS/TS tests leaving the configuration option out for either `blockGasLimit` or `transactionGasCap`, defaults to the Hardfork's value.

For Solidity Tests leaving the configuration option out disables the check, matching the existing behavior.

## allowUnlimitedContractSize under `--coverage`

In addition, there is one further change. Previously if running `--coverage` then `allowUnlimitedContractSize` would always be automatically enabled. Now it will be enabled unless the user has explicitly set it to false.

This brings all three overrides for `--coverage` into an aligned semantic; coverage will change them unless the user has set a specific value in config.

## Review

> [!NOTE]
> This PR should be reviewed a commit at a time.

## Docs PR

https://github.com/NomicFoundation/hardhat-website/issues/263


.changeset/free-goats-unite.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": minor
+---
+
+Disable transaction gas cap and block gas limit when running tests under `--coverage`.
```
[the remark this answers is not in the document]
kanej: I may review the changesets, but I am happy with this for the moment.

packages/hardhat/src/internal/builtin-plugins/network-manager/config-resolution.ts
```diff
@@ -83,9 +83,12 @@ export function resolveEdrNetwork(
...
       networkConfig.allowBlocksWithSameTimestamp ?? false,
-    allowUnlimitedContractSize:
-      networkConfig.allowUnlimitedContractSize ?? false,
-    blockGasLimit: BigInt(networkConfig.blockGasLimit ?? 60_000_000n),
+    allowUnlimitedContractSize: networkConfig.allowUnlimitedContractSize,
+    blockGasLimit:
+      typeof networkConfig.blockGasLimit === "number" ||
+      typeof networkConfig.blockGasLimit === "bigint"
+        ? BigInt(networkConfig.blockGasLimit)
+        : networkConfig.blockGasLimit,
```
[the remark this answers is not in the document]
kanej: We are intentionally making this change and accept that plugin authors may have to update their plugins.

packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -50,7 +50,8 @@ const solidityTestUserConfigType = z.object({
   blockTimestamp: z.bigint().optional(),
   prevRandao: z.bigint().optional(),
   gasLimit: z.bigint().optional(),
-  blockGasLimit: z.bigint().or(z.literal(false)).optional(),
+  blockGasLimit: z.number().or(z.bigint()).or(z.literal(false)).optional(),
+  transactionGasCap: z.number().or(z.bigint()).or(z.literal(false)).optional(),
```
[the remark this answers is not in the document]
kanej: If we consider this an issue, then I think we tackle all the config options where this would apply in one follow on PR.

packages/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -478,25 +485,21 @@ export async function getProviderConfig(
...
+    hardfork: networkConfig.hardfork,
+    blockGasLimit: genesisBlockGasLimit,
+    transactionGasCap: networkConfig.transactionGasCap,
+  });
 
   return {
     allowBlocksWithSameTimestamp: networkConfig.allowBlocksWithSameTimestamp,
-    allowUnlimitedContractSize: networkConfig.allowUnlimitedContractSize,
+    allowUnlimitedContractSize:
+      networkConfig.allowUnlimitedContractSize ?? false,
```
kanej, line 502: This was previously defaulting at the config stage but is now handled here so the user intent (defined/undefined) threads through and can be analysed for the coverage overrides.

packages/hardhat/src/internal/builtin-plugins/network-manager/edr/edr-provider.ts
```diff
@@ -461,6 +459,15 @@ export async function getProviderConfig(
...
   );
 
+  const genesisBlockGasLimit =
+    typeof networkConfig.blockGasLimit === "bigint"
+      ? networkConfig.blockGasLimit
+      : DEFAULT_EDR_NETWORK_BLOCK_GAS_LIMIT;
+
+  // EDR's mining block gas limit can be disabled in user config explicitly
+  const miningBlockGasLimit =
+    networkConfig.blockGasLimit === false ? undefined : genesisBlockGasLimit;
```
kanej, line 469: EDR has split apart config for each of these use cases. We do not expose them, but instead resolve them here.

packages/hardhat/src/internal/builtin-plugins/network-manager/type-extensions/config.ts
```diff
@@ -111,7 +111,7 @@ declare module "../../../../types/config.js" {
     // EDR network specific
     allowBlocksWithSameTimestamp?: boolean;
     allowUnlimitedContractSize?: boolean;
-    blockGasLimit?: number | bigint;
+    blockGasLimit?: number | bigint | false;
```
kanej, line 114: The option `?` in all of these cases is to allow us to differentiate between `false` and not set. False turning off the check, unset using the relevant Hardfork value.

packages/hardhat/src/internal/builtin-plugins/network-manager/utils/apply-coverage-network-overrides.ts
```diff
@@ -0,0 +1,32 @@
+import type { EdrNetworkConfig } from "../../../../types/config.js";
```
kanej, line 1: I placed this in a new `utils` folder.
I preferred a separate helper and new folder to extending the network manager implementation and then hacking around it in the tests.

packages/hardhat/src/internal/builtin-plugins/network-manager/utils/apply-coverage-network-overrides.ts
```diff
@@ -0,0 +1,32 @@
...
+  config: EdrNetworkConfig,
+  shouldEnableCoverage: boolean,
+): EdrNetworkConfig {
+  if (!shouldEnableCoverage) {
+    return config;
+  }
+
+  return {
+    ...config,
+    allowUnlimitedContractSize: config.allowUnlimitedContractSize ?? true,
```
kanej, line 28: This is a change - instead of defaulting to true, check the user config first.

packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
```diff
@@ -100,8 +100,18 @@ export async function solidityTestConfigToSolidityTestRunnerConfigArgs({
...
+      ? BigInt(config.blockGasLimit)
+      : undefined;
+  const disableBlockGasLimit = blockGasLimit === undefined;
+
+  const transactionGasCap =
+    typeof config.transactionGasCap === "number" ||
+    typeof config.transactionGasCap === "bigint"
+      ? BigInt(config.transactionGasCap)
+      : undefined;
+  const disableTransactionGasCap = transactionGasCap === undefined;
```
kanej, line 116: The intent here is allow the user to set config, but to disable by default in Solidity Test.

packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
alcuadrado: We have a validation helper to make sure that if this is a number, it's a safe number, as in, it can be turned into a bigint without loosing precision.

packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -55,6 +55,7 @@ const solidityTestProfileUserConfigType = z.object({
   prevRandao: z.bigint().optional(),
   gasLimit: z.bigint().optional(),
   blockGasLimit: z.number().or(z.bigint()).or(z.literal(false)).optional(),
+  transactionGasCap: z.number().or(z.bigint()).or(z.literal(false)).optional(),
```
alcuadrado: We can restrict this only to optional false, while EDR fixes the issue on their side.

packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
kanej: Applied

packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
kanej: Applied.

packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -54,7 +54,11 @@ const solidityTestProfileUserConfigType = z.object({
   blockTimestamp: z.bigint().optional(),
   prevRandao: z.bigint().optional(),
   gasLimit: z.bigint().optional(),
-  blockGasLimit: z.bigint().or(z.literal(false)).optional(),
+  blockGasLimit: z.number().or(z.bigint()).or(z.literal(false)).optional(),
+  // TODO: widen back to .number().or(z.bigint()).or(z.literal(false))
+  // once Solidity test runner no longer breaks setUp() when
+  // disableTransactionGasCap is false. See PR #8301.
+  transactionGasCap: z.literal(false).optional(),
```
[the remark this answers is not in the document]
kanej: We are opting to not constrain the types and just rely on validation in this interim period.

alcuadrado (approved):

Issue #1486, opened 2021-05-26
`hh flatten` doesn't support circular dependencies


fvictorio:

Endlessline:
How can one recreate this bug?

fvictorio:
```solidity
// contracts/Foo.sol
import "./Bar.sol";

contract Foo {}
```

```solidity
// contracts/Bar.sol
import "./Foo.sol";

contract Bar {}
```

`hh compile` works, `hh flatten` doesn't.

The solution is probably to get some topological order of the dependency graph and to use that to generate the flattened file, ignoring previously visited files when resolving `import`s.

BitMaster001:
+1

nekooei:
+1

slvrfn:
+1

vinkent420:
+1

nxqbao:
+1

alcuadrado:
I updated this issue explaining why this is more complicated than it seems, and why it may still not work for some usecases. 

This is something low priority to us, but we may consider doing it once Slang is more mature.

liu-zhipeng:
+1

gnarvaja:
+1

danilaplee:
+1

aaitor:
+1

jadrvel:
+1

IsaacCisnerosFT:
+1

xgamermonster9:
Yes, there’s some kind of givеаway going on. I got some Ethereum there yesterday for different wallets. It’s unlikely they’ll take it back from me, but it’s a pity I didn’t have more wallets and couldn’t get more. But maybe it’s for the best.

<img width="369" alt="Image" src="https://github.com/user-attachments/assets/5b983788-1d9d-45df-9185-dfb9584be220" />

<img width="373" alt="Image" src="https://github.com/user-attachments/assets/567fde7a-2777-4d4b-a13e-fd4c7fae960e" />

donotyue:
+1

adamcanray:
+1

PR #8305, opened 2026-05-16
flatten: error on cyclic dependencies instead of silently flattening

### LINKED ISSUE #7611: Flatten doesn't error on circular dependencies
In Hardhat 3, if you have two files:

```solidity
// ./contracts/F.sol
import "./G.sol";

contract F {}
```

```solidity
// ./contracts/G.sol
import "./F.sol";

contract G {}
```
On running `npx hardhat flatten ./contracts/G.sol`, it will output a flattened file, even though there is a circular dependency.

In Hardhat 2 you would see this error:
```shell
npx hardhat flatten ./contracts/F.sol
Error HH603: Hardhat flatten doesn't support cyclic dependencies.

For more info go to https://hardhat.org/HH603 or run Hardhat with --show-stack-traces
```

We should detect the cycle and throw an error rather than giving the impression we support cyclic dependencies. We do not: https://github.com/NomicFoundation/hardhat/issues/1486#issue-902595754

### Search terms

hardhat3 flatten

Kropiunig:
Closes #7611.

## Problem

Hardhat 3's `flatten` task does a plain DFS over the dependency graph and silently ignores cycles. Given two files that import each other:

```solidity
// contracts/F.sol
import "./G.sol";
contract F {}

// contracts/G.sol
import "./F.sol";
contract G {}
```

`npx hardhat flatten ./contracts/G.sol` happily emits a "flattened" file that drops the back edge. Hardhat 2 raised `HH603` for this case (see #1486), so this is a regression for users porting projects to HH3 and a footgun for contract verification.

## Fix

Switch `getSortedFiles` in `packages/hardhat/src/internal/builtin-plugins/flatten/task-action.ts` to a 3-color iterative DFS (WHITE / GRAY / BLACK). A GRAY node reached during traversal is a back edge, so we assemble the cycle from the current path stack and throw a new `HardhatError` descriptor.

Iterative rather than recursive so that:

1. Large projects don't blow the call stack.
2. The path stack is reified, which makes precise cycle reporting cheap (no separate predecessor map).

### Error descriptor

`HHE602 / CORE.BUILTIN_TASKS.FLATTEN_CYCLIC_DEPENDENCY` in `packages/hardhat-errors/src/descriptors.ts`. Slot 602 was the next free number after `INVALID_FILE_PATH` (601). Message lists every file in the cycle, e.g.:

```
Hardhat flatten doesn't support cyclic dependencies.

The following files form a cycle:
contracts/F.sol -> contracts/G.sol -> contracts/F.sol
```

## Tests

Two new fixture projects under `packages/hardhat/test/fixture-projects/flatten-task/`:

- `contracts-cyclic-dependencies/` (2-node cycle: F <-> G)
- `contracts-cyclic-dependencies-three-files/` (3-node cycle: H -> I -> J -> H)

Two new tests under `Cyclic dependencies` in `task-action.ts` verify the descriptor and the exact deterministic cycle string emitted by the sort.

Full flatten suite: `21 passing` (was 19). Adjacent dependency-graph suite unaffected: `48 passing` total when run together. `hardhat-errors` suite: `38 passing`. ESLint clean on the touched files.

## Checklist


kanej:
Hey @Kropiunig, thanks for this PR.

I couldn't push some updates so I have pulled the commits across and opened this PR: https://github.com/NomicFoundation/hardhat/pull/8326

PR #8310, opened 2026-05-18
fix(hardhat): return revert data as hex string for code 3 errors

### LINKED ISSUE #8075: Local network can not parse custom error while Sepolia network works fine.
### Describe the issue

OS: Windows 11
Hardhat: 3.1.11
solidity: 0.8.29
viem: 2.47.5

I am having trouble that local network can not parse my custom error. I deployed contract to Sepolia and code works fine. Contract method reverted and the error returned contains the following:

```typescript
error object: {
  "cause": {
    "details": "execution reverted",
    "metaMessages": [
      "Error: NoActiveGame()",
      ""
    ],
    "shortMessage": "The contract function \"joinGame\" reverted.",
    "version": "2.47.5",
    "name": "ContractFunctionRevertedError",
    "data": {
      "abiItem": {
        "inputs": [],
        "name": "NoActiveGame",
        "type": "error"
      },
      "errorName": "NoActiveGame"
    },
    "raw": "0x256d9840"
...
```

Here NoActiveGame() is the custom error so you can see it works fine.

But when I run it in local network, it can't be parsed. The error returned when calling publicClient.simulateContract() is as following:

```typescript
{
  "cause": {},
  "details": "data2?.slice is not a function",
  "docsPath": "/docs/contract/simulateContract",
  "metaMessages": [
    "Contract Call:",
    "  address:   0x5fbdb2315678afecb367f032d93f642f64180aa3\n  function:  joinGame()\n  sender:    0x70997970C51812dc3A010C7d01b50e0d17dc79C8"
  ],
  "shortMessage": "An unknown error occurred while executing the contract function \"joinGame\".",
```

Here is the full error info returned:

[local.network.error.log](https://github.com/user-attachments/files/26171222/local.network.error.log)

With the above contract error, there is no chance for the frontend application to find useful information to decode the custom error.

Below is my test step:

1. write and compile contract:
```bash
pnpm hardhat build
```

2. start local network:
```bash
pnpm hardhat node
```

3. deploy contract by using a script:

```typescript
import { writeFileSync } from "fs";
import hre from "hardhat";

type HexAddress = `0x${string}`;

function writeContractInfoFile(contractAddress: HexAddress, ownerAddress: HexAddress): void {
  const fileContent: string =
    `export const CONTRACT_ADDRESS = "${contractAddress}";\n` +
    `export const CONTRACT_OWNER_ADDRESS = "${ownerAddress}";\n`;

  writeFileSync("../frontend/app/utils/contractInfo.ts", fileContent, { encoding: "utf-8" });
}

async function main(): Promise<void> {
  // connect via viem client (Hardhat plugin)
  const network = await hre.network.connect();
  const viem = network.viem;
  const walletClients = await viem.getWalletClients();
  const deployer = walletClients[0];
  const ownerAddress: HexAddress = deployer.account.address as HexAddress;

  // deploy contract using viem helper
  const simple = await viem.deployContract("SimpleTexasHoldem", [
    "0x0000000000000000000000000000000000000000", // ETH mode
  ]);
  const contractAddress: HexAddress = simple.address as HexAddress;

  console.log(`SimpleTexasHoldem deployed to: ${contractAddress}`);
  console.log(`Contract owner (deployer): ${ownerAddress}`);

  //write contract info to frontend file
  //writeContractInfoFile(contractAddress, ownerAddress);
}

main().catch((error: unknown): void => {
  console.error(error);
  process.exitCode = 1;
});
```

And then deploy it:

```bash
pnpm hardhat run --network localhost scripts/deploy.ts
```

I also tried Ignition, but got the same failure.

4. Use ABI json file from artifacts in frontend application. When application triggered the method reverted, network has the following error:

```text
eth_call
  Contract call:       SimpleTexasHoldem#joinGame
  From:                0x70997970c51812dc3a010c7d01b50e0d17dc79c8
  To:                  0x5fbdb2315678afecb367f032d93f642f64180aa3

  Error: VM Exception while processing transaction: reverted with an unrecognized custom error (return data: 0x256d9840)
```

I can't see where these steps gone wrong. My application can run well in local network with other contract methods, only that the error returned by the network seems not right. Also, it works ax expected when running "pnpm hardhat test" for verifying custom error like the following code:

```typescript
    it("Should revert if no active game", async () => {
      await game.write.endGame({ account: owner.account });

      await viem.assertions.revertWithCustomError(
        game.write.joinGame({ account: player1.account }),
        game,
        "NoActiveGame"
      );
    });

```

Please help to check, thanks.

Edit: formatted code blocks.

js360000:
## Summary

Fixes #8075 — when a contract reverts on the local Hardhat node, viem
fails with `data2?.slice is not a function` while the identical
transaction works fine on Sepolia.

The JSON-RPC handler currently wraps `error.data` in a
`{ message, txHash, data }` object for **every** failure response,
including execution-revert (code 3). That diverges from the
geth/anvil/EVM-node convention where `error.data` is the raw revert
hex string, and breaks client tooling (viem, ethers, web3.js) that
decodes custom errors by reading and slicing that field directly.

### Before
```jsonc
// hardhat node response
{
  "error": {
    "code": 3,
    "message": "execution reverted",
    "data": { "message": "...", "txHash": "0x...", "data": "0x256d9840" }
  }
}
```

### After
```jsonc
// hardhat node response (matches geth/anvil/sepolia)
{
  "error": {
    "code": 3,
    "message": "execution reverted",
    "data": "0x256d9840"
  }
}
```

## Scope

- **Revert errors only** (`error.code === 3`) get the hex-string shape.
- Hardhat-internal errors (non-revert) keep the rich
  `{ message, txHash, data }` diagnostic wrapper — tooling does not try
  to decode those as revert data, and the wrapper is still useful for
  surfacing internal failures.
- `http-provider.ts` already handles either shape correctly because
  `extractReturnData` checks `typeof error.data === \"string\"` first,
  so upstream revert responses (now hex strings) flow through unchanged.

## Test plan
- [x] Updated the existing `revertWithDataAndTxHash` test in
  `packages/hardhat/test/internal/builtin-plugins/node/json-rpc/handler.ts`
  to assert the new shape and explain the rationale (links back to #8075).
- [x] `node --import tsx/esm --test test/internal/builtin-plugins/node/json-rpc/handler.ts`
  → all 13 tests pass, including the updated revert-shape test.
- [x] `pnpm lint:file` clean (prettier + eslint) on both changed files.
- [x] Manual re-trace through `_handleError` for non-revert paths
  (`plainError`, `methodNotFound`, `topLevelTxHash`, `dataAsString`,
  `dataWithTxHash`, `dataWithData`, `dataWithBoth`,
  `nonProviderErrorWithDataButNoCode3`) — all still produce the rich
  `{ message, txHash, data }` wrapper as before; no behavior change for
  those.

A changeset entry (`patch` to `hardhat`) is included.

### DIFF
--- a/.changeset/fix-json-rpc-revert-data-hex-string.md
+++ b/.changeset/fix-json-rpc-revert-data-hex-string.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix JSON-RPC `error.data` shape for revert errors (code 3): return the raw revert hex string directly instead of a `{ message, txHash, data }` wrapper object. This matches the geth/anvil/EVM-node convention that client tooling (viem, ethers, web3.js) relies on when decoding custom errors. Closes #8075.
--- a/packages/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
@@ -269,15 +269,8 @@ const _handleError = (error: Error): JsonRpcResponse => {
     error = new InternalError(undefined, error);
   }
 
-  // For revert errors (code 3), the JSON-RPC `data` field must be the hex
-  // revert data as a string — this matches the geth/anvil/EVM-node convention
-  // and is what tooling (viem, ethers, web3.js) expects when decoding custom
-  // errors. Returning the rich `{ message, txHash, data }` wrapper here breaks
-  // clients that call `.slice` on `error.data` (see issue #8075).
-  //
-  // For Hardhat-internal errors (non-revert) we keep the rich object so that
-  // diagnostic info (txHash, message) is preserved — those clients don't
-  // attempt to decode the field as hex revert data.
+  // Revert errors (code 3): return raw hex to match the geth/anvil convention
+  // that viem/ethers/web3.js rely on. Other errors keep the wrapper for diagnostics.
   const data: unknown = isRevertError
     ? returnData
     : { message: error.message, txHash, data: returnData };
--- a/packages/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/node/json-rpc/handler.ts
@@ -269,6 +269,19 @@ const _handleError = (error: Error): JsonRpcResponse => {
     error = new InternalError(undefined, error);
   }
 
+  // For revert errors (code 3), the JSON-RPC `data` field must be the hex
+  // revert data as a string — this matches the geth/anvil/EVM-node convention
+  // and is what tooling (viem, ethers, web3.js) expects when decoding custom
+  // errors. Returning the rich `{ message, txHash, data }` wrapper here breaks
+  // clients that call `.slice` on `error.data` (see issue #8075).
+  //
+  // For Hardhat-internal errors (non-revert) we keep the rich object so that
+  // diagnostic info (txHash, message) is preserved — those clients don't
+  // attempt to decode the field as hex revert data.
+  const data: unknown = isRevertError
+    ? returnData
+    : { message: error.message, txHash, data: returnData };
+
   const response: FailedJsonRpcResponse = {
     jsonrpc: "2.0",
     id: null,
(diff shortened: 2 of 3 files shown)


ChristopherDedominici (approved):
Hi @js360000, thanks for the PR! I reviewed and approved it. I took the liberty of shortening the comments a bit to reduce verbosity, but other than that, everything looks good!

PR #8312, opened 2026-05-19
Hardhat typescript 6

alcuadrado:
This PR upgrades the repository and the HH3 sample projects to HH3.

The main motivation of doing if you initialize a new project with Hardhat 3 it doesn't work well in vscode. This fixes it.

The tsconfig files that we initialize in the samples are simpler, as some things (e.g. `strict` and `esModuleInterop`) are now `true` by default.


alcuadrado:
Instead of fighting `ts-node` in the ignition tests I just migrated them to tsx

packages/hardhat/templates/hardhat-3/03-minimal/tsconfig.json
```diff
@@ -1,13 +1,10 @@
-/* Based on https://github.com/tsconfig/bases/blob/501da2bcd640cf95c95805783e1012b992338f28/bases/node22.json */
 {
   "compilerOptions": {
     "lib": ["es2023"],
-    "module": "node16",
-    "target": "es2022",
-    "strict": true,
```
kanej, line 7: We removed strict mode for `minimal` but not the other templates? Is this intentional?

packages/hardhat/templates/hardhat-3/01-node-test-runner-viem/tsconfig.json
```diff
@@ -1,13 +1,12 @@
-/* Based on https://github.com/tsconfig/bases/blob/501da2bcd640cf95c95805783e1012b992338f28/bases/node22.json */
 {
   "compilerOptions": {
     "lib": ["es2023"],
-    "module": "node16",
-    "target": "es2022",
+    "module": "node20",
+    "target": "es2023",
```
kanej, line 5: My read here is that while almost all of `es2024` is supported in Node 22, we will be conservative and only go to es2023?

packages/hardhat/templates/hardhat-3/01-node-test-runner-viem/tsconfig.json
```diff
@@ -1,13 +1,12 @@
-/* Based on https://github.com/tsconfig/bases/blob/501da2bcd640cf95c95805783e1012b992338f28/bases/node22.json */
 {
   "compilerOptions": {
     "lib": ["es2023"],
-    "module": "node16",
-    "target": "es2022",
+    "module": "node20",
```
kanej, line 4: If I follow, we pin the latest module policy based on running in Node 22. We don't go with `nodenext`, as a pinned version ensures future changes in Node 24/26 don't break a project unnecessarily.

packages/hardhat/templates/hardhat-3/01-node-test-runner-viem/tsconfig.json
```diff
@@ -1,13 +1,12 @@
...
+    "target": "es2023",
     "strict": true,
     "esModuleInterop": true,
```
kanej: Can we drop "strict" as it is true be default in ts@6.
Potentially drop `esModuleInterop` as well?

kanej (approved):
I have left a few comments querying the template tsconfig defaults. But I am happy with the wider PR and pre-approving.

packages/hardhat/templates/hardhat-3/03-minimal/tsconfig.json
alcuadrado: Good catch! I removed them in the rest. `strict` is now the default. Same with `esModuleInterop`.

packages/hardhat/templates/hardhat-3/01-node-test-runner-viem/tsconfig.json
alcuadrado: That's right. I'm starting to thing that maybe we should provide `hardhat/tsconfig/node<x>.json` files.

packages/hardhat/templates/hardhat-3/01-node-test-runner-viem/tsconfig.json
alcuadrado: right

packages/hardhat/templates/hardhat-3/01-node-test-runner-viem/tsconfig.json
alcuadrado: done

PR #8326, opened 2026-05-22
Fix/flatten cyclic dependencies additions

### LINKED ISSUE #7611: Flatten doesn't error on circular dependencies

kanej:
Add error on cyclic dependencies instead of silently flattening.

This PR is mainly taken from #8305. Thanks @Kropiunig.

image

Closes #7611


packages/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -297,6 +308,27 @@ function getSortedFiles(dependencyGraph: DependencyGraph): ResolvedFile[] {
   return sortedFiles;
 }
 
+function assertNotOnPath(
+  path: readonly ResolvedFile[],
+  file: Readonly<ResolvedFile>,
```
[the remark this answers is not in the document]
kanej: I have remove the readonly.

packages/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -297,6 +308,27 @@ function getSortedFiles(dependencyGraph: DependencyGraph): ResolvedFile[] {
...
+): void {
+  const cycleStart = path.indexOf(file);
```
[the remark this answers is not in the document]
kanej: Flatten reads from file - this complexity addition does not make sense given the performance impact.

packages/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -273,30 +275,55 @@ function getSortedFiles(dependencyGraph: DependencyGraph): ResolvedFile[] {
     );
   };
 
-  // Depth-first walking
-  const walk = (files: ResolvedFile[]) => {
+  // Depth-first walking. `path` is the chain of files leading to the current
+  // visit; a dep that reappears in it indicates a cycle.
+  const walk = (files: ResolvedFile[], path: readonly ResolvedFile[]) => {
```
schaable: nit: I'd rename `path` to something that describes the variable more clearly, maybe `ancestors` or `importChain`.

packages/hardhat-errors/src/descriptors.ts
```diff
@@ -933,6 +933,17 @@ Please double check your script's path.`,
...
+      FLATTEN_CYCLIC_DEPENDENCY: {
+        number: 602,
+        messageTemplate: `Hardhat flatten doesn't support cyclic dependencies.
+
+The following files form a cycle:
+{cycle}`,
+        websiteTitle: "Hardhat flatten doesn't support cyclic dependencies",
+        websiteDescription: `Hardhat \`flatten\` works by producing a topological order of the dependency graph and concatenating the files in that order. A graph with cycles has no topological order, so the algorithm can't be applied.
+
+If you need to flatten your project, refactor your Solidity sources to remove the cycle. Cyclic imports are normally avoidable and are harder to reason about.`,
```
schaable:
```suggestion
If you need to flatten your project, refactor your Solidity sources to remove the cycle. Cyclic imports are usually avoidable, and removing them tends to make code easier to reason about.`,
```

packages/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -273,30 +275,55 @@ function getSortedFiles(dependencyGraph: DependencyGraph): ResolvedFile[] {
...
 
   const roots = sortBySourceName(dependencyGraph.getRoots().values());
 
-  walk(roots);
+  walk(roots, []);
 
   return sortedFiles;
 }
 
+function assertFileNotOnPath(
```
schaable: nit: a clearer name would be `throwIfCycle`

packages/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
```diff
@@ -273,30 +275,55 @@ function getSortedFiles(dependencyGraph: DependencyGraph): ResolvedFile[] {
...
+      assertFileNotOnPath(file, path);
 
       const dependencies = sortBySourceName(
         Array.from(dependencyGraph.getDependencies(file)).map((d) => d.file),
       );
 
-      walk(dependencies);
+      walk(dependencies, [...path, file]);
 
+      visitedFiles.add(file);
```
schaable, line 297: Moving this line here means `assertFileNotOnPath` is now responsible for breaking the cycle. Not an issue, but we should be careful with future changes to the algorithm to avoid a stack overflow.

packages/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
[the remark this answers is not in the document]
schaable: I was about to make the same comment 😂 probably not an issue unless the dependency chain is huge. Still, claude can probably one-shot this.

schaable (approved):
LGTM! Left a couple of minor comments just to be annoying 😂 nothing blocking

packages/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
kanej: Agreed, `importChain` is now used here and in the subfunction.

packages/hardhat/src/internal/builtin-plugins/flatten/task-action.ts
kanej: Applied.

PR #8331, opened 2026-05-24
chore: remove unused error classes

gultekinmakif:
No reason to keep: `TransactionExecutionError`, `MethodNotSupportedError`, and `InvalidResponseError`

* Keeps: `MethodNotFoundError`, used in [json-rpc handler tests](https://github.com/NomicFoundation/hardhat/blob/243244aa8f434f7dff7d5a4d46aba64e2144cede/packages/hardhat/test/internal/builtin-plugins/node/json-rpc/handler.ts#L204).

> No changesets needed.


kanej (approved):
Thanks @gultekinmakif

PR #8333, opened 2026-05-25
fix: skip full-input retry for non-retryable verification errors

### LINKED ISSUE #7731: Don't retry verification with full input for non-retryable errors
### Describe the issue

Currently, when verification fails with minimal input, we always retry with the full compiler input. However, some errors (like bytecode mismatches or wrong constructor arguments) will fail regardless of which input we use.

We should add logic to detect these non-retryable errors and throw immediately instead of wasting time on a retry that will fail. The list of possible error messages is available in https://docs.etherscan.io/api-reference/endpoint/checkverifystatus.

This logic should also be extended to Blockscout and Sourcify verification providers.

Related to https://github.com/NomicFoundation/hardhat/pull/7577

gultekinmakif:

---

When verification with minimal compiler input fails, `hardhat-verify` **always retries with the full input**.
For some pre-determined errors (bytecode mismatch, wrong constructor args), this wastes time without any expected outcome.

## Context

Between the minimal and full-input attempts, only `compilerInput` (source files) changes.Address, constructor args, contract name, solc version, and compiler settings stay the same.

## Fix

1. Introduce a new lookup table for `non-retryable errors`: NON_RETRYABLE_PATTERNS
2. Introduce a new Hardhat error descriptor: `NON_RETRYABLE_VERIFICATION_ERROR` (HHE80030)
3. After the first `attemptVerification` fail, check the error message against `NON_RETRYABLE_PATTERNS`.
4. If matched, throw `NON_RETRYABLE_VERIFICATION_ERROR` immediately instead of retrying.

## Tests

`packages/hardhat-verify/test/verification.ts`:
- `isNonRetryableVerificationError` returns true for all 5 detected non-retryable patterns.
- Returns false for retryable messages:compilation errors, missing sources, library issues, temporary errors, generic failures, etc.

## Changeset

`patch` on `@nomicfoundation/hardhat-verify` and `@nomicfoundation/hardhat-errors`.

## Notes

I also considered these for an exhaustive list, but any missing item can be raised and added later:

- Etherscan:
  - Compilation errors, missing sources, library not found: retryable (full input adds the missing files)
  - Temporary errors: explicitly says retry
  - Generic "Fail - Unable to verify": no signal, safer to retry
  - Invalid solc version, contract code not found, already verified: already handled upstream before `attemptVerification` runs
- Blockscout: returns only "Fail - Unable to verify".
- Sourcify: error messages vary, could not locate a doc.

### DIFF
--- a/.changeset/red-owls-live.md
+++ b/.changeset/red-owls-live.md
@@ -1,0 +1,6 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+"@nomicfoundation/hardhat-errors": patch
+---
+
+Skip full-input retry for non-retryable verification errors
--- a/packages/hardhat-errors/src/descriptors.ts
+++ b/packages/hardhat-errors/src/descriptors.ts
@@ -3052,6 +3052,9 @@ Retrying with the full compiler input would produce the same result.`,
         websiteTitle: "Non-retryable verification error",
         websiteDescription: `The block explorer returned an error that cannot be resolved by retrying with different compiler input. Common causes include:
 
+- **Bytecode mismatch**: the deployed bytecode does not match the compiled bytecode. This can happen when the contract was compiled with a different build profile than the one used for verification. Try verifying with the correct \`--build-profile\`.
+- **Constructor arguments**: the provided constructor arguments do not match the ones used during deployment.
+
 Retrying will not change the outcome. Fix the underlying cause and try again.`,
       },
     },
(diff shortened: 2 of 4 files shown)


schaable:
Hi @gultekinmakif. I've been working on a refactor on top of your changes, but the branch diverged quite a bit with the new commits and it's getting difficult to continue building on top of this PR cleanly.

I'm going to close this PR in favor of https://github.com/NomicFoundation/hardhat/pull/8349, which contains the refactor and follow-up changes. Thanks again for the contribution and for the work on this.

gultekinmakif:
@schaable no worries at all.
You can let me know if you have any feedbacks in the future PRs.

Issue #8309, opened 2026-05-18
Reduce gaps in compilation and Solidity Test features


kanej:
In migrating projects to Hardhat 3, multiple features have been identified that should be added to Solidity Test to minimize effort in porting. This is an organizational ticket.

PR #8343, opened 2026-05-27
feat(solidity-test): negation filters

### LINKED ISSUE #8184: Support negation of test name filters in Solidity Test

gultekinmakif:

---
## Summary

Adds `--no-match-test` and `--no-match-contract` CLI flags to the `test solidity` task, enabling Foundry-style negation filters for Solidity tests.

## Context

Hardhat 3's `--grep` option filters tests by a regex, but EDR's regex crate does not support look-around, so there is no way to express "everything except X."

Foundry solves this with dedicated `--no-match-test` and `--no-match-contract` flags. This PR brings the same capability to Hardhat.

### Foundry parity

| Foundry | Hardhat | Notes |
|---------|-------------------|-------|
| `--match-test` | `--grep` (existing) | Same concept, different name |
| `--match-contract` | (not in scope) | `--grep` already filters by test name; contract-level positive matching could be a follow-up |
| `--no-match-test` | `--no-match-test` | **New(this PR)** |
| `--no-match-contract` | `--no-match-contract` | **New(this PR)** |

## Fix

- `--no-match-contract <regex>`: Filters out test contracts whose name matches the regex before passing them to EDR.

- `--no-match-test <regex>`: Collects all test function names from surviving contract ABIs, removes those matching the regex, intersects with `--grep` if present, then constructs a positive `testPattern` regex to pass to EDR.

- Invalid or empty patterns throw `HardhatError`.

- When all tests or contracts are excluded, a warning is printed and the task returns early with zero results.

## Example usage

```shell
# Run everything except fork tests
npx hardhat test solidity --no-match-test "^testFork_"

# Run everything except the ForkTest contract
npx hardhat test solidity --no-match-contract "ForkTest"

# Combine with --grep
npx hardhat test solidity --grep "^test" --no-match-test "Fork"
```

## Notes

- Added 8 test cases covering partial exclusion, total exclusion (both flags), invalid regex, and empty pattern cases.
- Closes #8184 (Support negation of test name filters in Solidity Test)
- Adresses #8309 (Reduce gaps in Solidity Test features)

### DIFF
--- a/.changeset/negation-test-filters.md
+++ b/.changeset/negation-test-filters.md
@@ -1,0 +1,5 @@
+---
+"hardhat": minor
+---
+
+Add `--no-match-test` and `--no-match-contract` CLI flags to `test solidity` for excluding tests and contracts by pattern
--- a/packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
@@ -165,6 +165,24 @@ export async function solidityTestConfigToSolidityTestRunnerConfigArgs({
   };
 }
 
+function parseAbi(artifact: Artifact): Abi {
+  return JSON.parse(artifact.contract.abi);
+}
+
+export function getTestFunctionNames(artifact: Artifact): string[] {
+  const names: string[] = [];
+  for (const { type, name } of parseAbi(artifact)) {
+    if (
+      type === "function" &&
+      typeof name === "string" &&
+      (name.startsWith("test") || name.startsWith("invariant"))
+    ) {
+      names.push(name);
+    }
+  }
+  return names;
+}
+
 export function isTestSuiteArtifact(artifact: Artifact): boolean {
   const bytecode = artifact.contract.bytecode;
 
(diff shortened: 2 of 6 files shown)


gultekinmakif:
@schaable 
can we get claude to review some of these PRs, I have some time to work on them this weekend.
appreciated!

kanej:
Hey @gultekinmakif, I am going to close this PR. We will want to do the Solidity test filtering within EDR. I am having the team take a look: https://github.com/NomicFoundation/edr/issues/1491

PR #8346, opened 2026-05-27
Only load hardhat-ledger's internals when at least one account is configured

alcuadrado:


packages/hardhat-ledger/test/integration.ts
```diff
@@ -100,6 +101,33 @@ describe("LedgerHandler", () => {
...
       assert.deepEqual(res, HARDHAT_ACCOUNTS_ADDRESSES);
     });
+
+    it("should skip the ledger onRequest body when no ledger accounts are configured", async () => {
+      // With the mocked network plugin, eth_accounts fails with -32601. If
+      // the ledger plugin's onRequest body ran, it would swallow that error
+      // and return its (empty) list of ledger accounts. Skipping the body
+      // when no ledger accounts are configured means the error surfaces
+      // unchanged.
+      const hre = await createHardhatRuntimeEnvironment({
```
kanej, line 111: I am not a fan of the trickiness of this test. But I don't have a better suggestion, and we definitely want a regression check.

kanej (approved):
I did a quick manual run to confirm the happy case. LGTM.

packages/hardhat-ledger/test/integration.ts
alcuadrado: agreed

PR #8351, opened 2026-05-29
fix: prefer env vars for config variables

### LINKED ISSUE #7596: Environment variables should take precedence when resolving configuration variables
### Describe the issue

If a config var `FOO` is used, and it's set both in the keystore and as an environment variable, the envvar should be used. The keystore shouldn't be even checked if the envvar exists.

BROCCOLO1D:
## Summary
- Makes `LazyResolvedConfigurationVariable` resolve an existing environment variable before invoking configuration variable hooks.
- Adds a regression test covering env-var precedence over hook-provided values.

## Why
Configuration variables can be resolved by plugins such as keystore integrations. Environment variables should consistently take precedence and avoid invoking those hooks when present.

## Changes
- Short-circuit `LazyResolvedConfigurationVariable._getRawValue()` when `process.env[variable.name]` exists.
- Keep the existing hook/default chain for variables not present in the environment.

## Validation
- `pnpm test:file packages/hardhat/test/internal/core/configuration-variables/configuration-variables.ts`
- `pnpm lint:file packages/hardhat/src/internal/core/configuration-variables.ts packages/hardhat/test/internal/core/configuration-variables/configuration-variables.ts`

## Scope
Focused fix for configuration variable resolution precedence.

Closes #7596

### DIFF
--- a/.changeset/proud-hats-fetch.md
+++ b/.changeset/proud-hats-fetch.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Prefer environment variables over configuration variable hooks when resolving configuration variables.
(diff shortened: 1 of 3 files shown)


BROCCOLO1D:

PR #8352, opened 2026-05-29
fix: scope EIP-712 duplicate checks to included sources

### LINKED ISSUE #8344: EIP-712 collector throws `HHE818` for a same-named struct in a source the user didn't include
### Version of Hardhat

3.6.0

### What happened?

`test.solidity.eip712Types.include` in Hardhat defines the include set only controls which struct names are *emitted*. The collector still walks every compiled source's AST and considers every struct it finds for name-collision detection. A struct in a file the user didn't include — and that nothing in the included scope imports — can therefore abort the test run with `HHE818` if its name happens to clash with a selected struct, even though the user has no expectation that the tool would look at that file at all.

Forge's equivalent (`[bind_json].include`) behaves differently: structs in sources outside the include set never enter the registry.

Self-contained dual-toolchain repro (`forge test` passes, `npx hardhat test solidity` throws `HHE818` against the same source set): <https://github.com/anaPerezGhiglia/repro-hh-eip712-include-scope>.

Two small files plus a test:

- `contracts/A.sol` — listed in `eip712Types.include`; defines `library A { struct Order { address user; uint256 amount; } }`.
- `contracts/B.sol` — **not** included, **not** imported by `A.sol` or any test; defines a same-named `struct Order { bytes32 id; bool active; }`.

The repro repo's README has the full versions table, devcontainer setup, and an optional appendix that also validates the name-form path under `forge bind-json`.

## Observed error

```
Error HHE818: Two different EIP-712 struct definitions named "Order" were found:
- contracts/A.sol
- contracts/B.sol
```

Thrown at collection time, before any test runs.

## Expected behavior

`include` should bound which sources contribute structs to the registry. Conflict detection should fire only on selected-vs-selected conflicts (and on unselected conflicts reachable from a selected struct — the existing `reachableFromSelected` deferred-path machinery in `canonicalize.ts` already covers that). An unselected struct in an unimported file should not throw.

## Environment

- `hardhat@3.6.0` (also reproduces on `3.5.1`)
- `@nomicfoundation/edr@0.12.0-next.33`

### Search terms

eip712, EIP-712, HHE818

BROCCOLO1D:
## Summary
- Fixes Solidity Test EIP-712 type collection so a same-named struct in an unselected source doesn't trigger a duplicate-name error.
- Keeps duplicate checks for selected structs and for non-selected structs that are reachable dependencies of selected types.
- Adds a regression test for include-scoped duplicate struct names.

## Why
The `test.solidity.eip712Types.include` setting should bound which source files contribute selected struct definitions. A file outside the include set should not fail the run just because it defines a struct with the same name as a selected struct.

## Changes
- Tracks selected source paths separately from selected struct names.
- Uses selected source paths when ordering and classifying conflicting definitions.
- Adds a changeset for the Hardhat patch release.

## Validation
- `pnpm test:file packages/hardhat/test/internal/builtin-plugins/solidity-test/eip712/index.ts`
- `pnpm test:file packages/hardhat/test/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts`
- `pnpm lint:file packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts packages/hardhat/test/internal/builtin-plugins/solidity-test/eip712/index.ts`

## Scope
Focused Solidity Test EIP-712 collector fix.

Closes #8344

### DIFF
--- a/.changeset/gentle-orders-include.md
+++ b/.changeset/gentle-orders-include.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed Solidity Test EIP-712 type collection so unselected sources don't trigger duplicate struct-name errors.
--- a/packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
@@ -25,14 +25,17 @@ import { HardhatError } from "@nomicfoundation/hardhat-errors";
  * constructs and propagates `None` through the dep graph so dependents are
  * also dropped.
  *
- * Only names in `selectedNames` are emitted; non-selected structs still
+ * Only names from `selectedSourcePaths` are emitted; non-selected structs still
  * participate in dep resolution so cross-file deps inline correctly.
  */
 export function canonicalizeStructs(
   structs: CollectedStruct[],
   selectedNames: Set<string>,
+  selectedSourcePaths: Set<string> = new Set<string>(
+    structs.filter((s) => selectedNames.has(s.name)).map((s) => s.sourcePath),
+  ),
 ): string[] {
-  const byName = indexByName(structs, selectedNames);
+  const byName = indexByName(structs, selectedNames, selectedSourcePaths);
   const knownNames = new Set(byName.keys());
   const encodable = computeEncodable(byName, knownNames);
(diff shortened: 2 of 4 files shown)


BROCCOLO1D:

PR #8355, opened 2026-05-29
fix: tolerate indexed dynamic siblings when decoding event args

### LINKED ISSUE #8338: `readEventArgument` throws `HHE10001` on events with an indexed dynamic param (string/bytes/array/tuple indexed)
### Version of Hardhat

3.5.1

## Summary
  
  `m.readEventArgument(...)` throws `HHE10001` when the emitting event has **any** indexed dynamic param (`string`/`bytes`/array/tuple `indexed`), even when the arg I'm reading is a different, decodable one.
  
  Indexed dynamic types only store their `keccak256` in the log topic, so ethers v6 returns an `Indexed` placeholder (`{ hash, _isIndexed: true }`). ignition-core's value decoder has no branch for it and falls through to the "can't decode" throw. 
  
  ## Root cause

  ethersValueIntoEvmValue (internal/execution/abi.ts) handles string/number/boolean/ethers.Result but has no branch for the ethers.Indexed placeholder, so it falls through to the throw.

  ## Proposed fix

```javascript
  // Indexed dynamic params aren't stored in the log; ethers returns an `Indexed`
  // placeholder exposing only the topic hash. Return the hash instead of throwing.
  if (
    ethersValue !== null &&
    typeof ethersValue === "object" &&
    (ethersValue as ethers.Indexed)._isIndexed === true
  ) {
    return (ethersValue as ethers.Indexed).hash;
  }
```

## Environment

  - `@nomicfoundation/ignition-core` 3.1.6
  - `@nomicfoundation/hardhat-ignition` 3.1.6
  - `hardhat` 3.5.1
  - Node v22.14.0
  
  ## Minimal reproduction
  
  ```solidity
  contract Repro {
      event SuiteDeployed(address indexed token, string indexed salt);

      function emitEvent() external {
          emit SuiteDeployed(address(this), "my-salt");
      }
  }
```

```javascript
  import { buildModule } from "@nomicfoundation/hardhat-ignition/modules";

  export default buildModule("Repro", (m) => {
    const repro = m.contract("Repro");
    const call = m.call(repro, "emitEvent");
    // Reading any arg decodes the whole event → throws on `string indexed salt`.
    const token = m.readEventArgument(call, "SuiteDeployed", "token", { id: "ReadToken" });
    m.contractAt("Repro", token, { id: "ReproAt" }); // consumes the read so it executes
    return { repro };
  });
```

```bash
  npx hardhat ignition deploy ignition/modules/Repro.ts --network localhost fails at Repro#ReadToken:

  Error HHE10001: Hardhat Ignition can't decode ethers.js value of type string: { "_isIndexed": true }
```

### Search terms

_No response_

kanej:
In Ignition, we ban at validation stage the reading of indexed dynamic params (string/bytes/array/tuple `indexed`), as only their hashes are stored in the log, not their values. However, we also decode the entire event tuple even when reading a single arg, so while we do not read indexed dynamic events we do have to handle them during event decoding.

The #8338 bug was that we were not doing this, specifically Ethers v6 returns an `Indexed` placeholder for them and our decoder did not deal with that placeholder, throwing HHE10001.

We now detect the placeholder with `ethers.Indexed.isIndexed` in `ethersValueIntoEvmValue` and return its `.hash`. Reading the indexed- dynamic arg itself is still blocked by validation, so the hash only ever appears as an intermediate value for sibling decoding, but it no longer blocks reads of allowed siblings.

Fixes #8338.

## Reproducing the 8338 locally

Add two files in the example project:

```solidity
// contracts/Repro8338.sol

pragma solidity ^0.8.0;

// Minimal reproduction of https://github.com/NomicFoundation/hardhat/issues/8338
// `m.readEventArgument` on `token` decodes the whole event, and the
// `string indexed salt` placeholder (`{ _isIndexed: true, hash }`) returned by
// ethers v6 makes ignition-core's value decoder throw HHE10001.
contract Repro8338 {
  event SuiteDeployed(address indexed token, string indexed salt);

  function emitEvent() external {
    emit SuiteDeployed(address(this), "my-salt");
  }
}
```

```ts
// ignition/modules/Repro8338.ts
import { buildModule } from "@nomicfoundation/hardhat-ignition/modules";

// Minimal reproduction of https://github.com/NomicFoundation/hardhat/issues/8338
//
// Run with:
//   pnpm hardhat ignition deploy ignition/modules/Repro8338.ts
//
// Expected: deploys cleanly and `token` resolves to the contract address.
// Actual:   fails at Repro8338#ReadToken with
//   Error HHE10001: Hardhat Ignition can't decode ethers.js value of type
//   string: { "_isIndexed": true }
export default buildModule("Repro8338", (m) => {
  const repro = m.contract("Repro8338");
  const call = m.call(repro, "emitEvent");

  // Reading any arg decodes the whole event, so the `string indexed salt`
  // placeholder is what trips the decoder — not the `token` we're reading.
  const token = m.readEventArgument(call, "SuiteDeployed", "token", {
    id: "ReadToken",
  });

  // Consume the read so the future actually executes.
  m.contractAt("Repro8338", token, { id: "ReproAt" });

  return { repro };
});
```

Finally run:

```
pnpm hardhat ignition deploy ./ignition/modules/Repro8338.ts
```

### DIFF
--- a/.changeset/young-lines-design.md
+++ b/.changeset/young-lines-design.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/ignition-core": patch
+---
+
+Fixed `m.readEventArgument` by tolerating indexed dynamic siblings when decoding event args ([#8338](https://github.com/NomicFoundation/hardhat/issues/8338))
--- a/packages/ignition-core/src/internal/execution/abi.ts
+++ b/packages/ignition-core/src/internal/execution/abi.ts
@@ -574,6 +574,21 @@ function ethersValueIntoEvmValue(
     return ethersResultIntoEvmTuple(ethersValue, paramType.components);
   }
 
+  // Indexed dynamic params (string/bytes/array/tuple) aren't stored in the log
+  // — only their keccak256 lives in a topic — so ethers v6 returns an `Indexed`
+  // placeholder for them. Reading the value itself is rejected upstream by
+  // `validateArtifactEventArgumentParams`, so this branch only fires for
+  // sibling params during tuple decoding; returning the topic hash lets the
+  // surrounding event decode without poisoning reads of other args.
+  if (ethers.Indexed.isIndexed(ethersValue)) {
+    assertIgnitionInvariant(
+      ethersValue.hash !== null,
+      "ethers.Indexed.hash must be set for placeholders produced by decodeEventLog",
+    );
+
+    return ethersValue.hash;
+  }
+
   throw new HardhatError(
     HardhatError.ERRORS.IGNITION.GENERAL.UNSUPPORTED_DECODE,
     {
--- a/packages/ignition-core/test/execution/abi.ts
+++ b/packages/ignition-core/test/execution/abi.ts
@@ -1,6 +1,7 @@
 import { HardhatError } from "@nomicfoundation/hardhat-errors";
 import { assertThrowsHardhatError } from "@nomicfoundation/hardhat-test-utils";
 import { assert } from "chai";
+import { ethers } from "ethers";
 import type { Artifact } from "../../src/types/artifact.js";
 
 import type { SolidityParameterType } from "../../src/index.js";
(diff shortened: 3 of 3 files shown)


alcuadrado (approved):
Looks good for a quick fix. We should remove ethers from ignition in the near future.

PR #8367, opened 2026-06-03
Update imports criteria updated

### LINKED ISSUE #8156: Update the imports criteria in the `engineering-guidelines.md` and `CLAUDE.md` and apply the new ones.
Our current import guidelines mention that we are ok with top-level imports from hook handlers. That is not entirely correct. There are two exceptions:

- `ConfigHooks`
- `HardhatRuntimeEnvironmentHooks`

These are always run, so they should also import as few things as possible.

We should update the guidelines to reflect that, and update the codebase accordingly.

On top of that, we should also add a new criteria for `NetworkHooks`. While they aren't always run, there's a common pattern of using them to extend the `NetworkConnection` object. Those extensions tend to load things, even when they are never used. (e.g.  https://github.com/NomicFoundation/hardhat/issues/8157)

alcuadrado:
Replaces https://github.com/NomicFoundation/hardhat/pull/8254 and closes https://github.com/NomicFoundation/hardhat/issues/8156


docs/engineering-guidelines.md
```diff
@@ -105,6 +105,38 @@ Use `await import` only if one of these conditions is met:
 
 Test files are free to use `await import` freely.
 
+### Always-run hook handler factories
+
+`ConfigHooks` and `HardhatRuntimeEnvironmentHooks` factories run on every Hardhat invocation, so they should follow the same criteria as the plugin's `index.ts` file.
+
+### Lazy wrappers for `NetworkConnection` extensions
+
+`NetworkHooks` aren't always run, but `newConnection` handlers that extend `NetworkConnection` (e.g. `connection.foo = …`) should consider lazily initializing their bussiness logic, unless they are virtually always used. The reason for this is that otherwise the first network connection initializes too many unnecessary things. For example, the `hardhat-ethers` plugin doesn't need to be lazily initialized, as most users that install it will use it most of the time after creating a new network. On the contrary, the `hardhat-ignition-ethers` plugin isn't always used, so it should be lazily initialized.
```
schaable:
```suggestion
`NetworkHooks` aren't always run, but `newConnection` handlers that extend `NetworkConnection` (e.g. `connection.foo = …`) should consider lazily initializing their business logic, unless they are virtually always used. The reason for this is that otherwise the first network connection initializes too many unnecessary things. For example, the `hardhat-ethers` plugin doesn't need to be lazily initialized, as most users that install it will use it most of the time after creating a new network. On the contrary, the `hardhat-ignition-ethers` plugin isn't always used, so it should be lazily initialized.
```

docs/engineering-guidelines.md
```diff
@@ -105,6 +105,38 @@ Use `await import` only if one of these conditions is met:
...
+
+Reference pattern for lazy initialization: `packages/hardhat-ignition-ethers/src/internal/hook-handlers/network.ts`. When the wrapper caches both the imported module and an instance built from it, the getter must run `await import(...)` **before** the instance cache check — see the "Ordering in cached lazy getters" section in `docs/engineering-guidelines.md` (under GC3) for the rationale and code shape.
+
+### Ordering in cached lazy getters
+
+This rule applies to any cached dynamic import (condition 6 in the list above) that also caches an instance built from the imported module. The `NetworkConnection` lazy wrapper above is one example, but the same shape comes up elsewhere.
+
+The getter must run `await import(...)` **before** the instance cache check, so concurrent callers share a single microtask-dedupe point — otherwise each suspended caller re-enters the branch, constructs its own implementation, and the callers end up with different instances.
```
schaable, line 122: It may be worth noting that there must be no await between reading this.#instance === undefined and assigning it

CLAUDE.md
```diff
@@ -43,6 +43,12 @@ Use `await import` only if one of these conditions is met:
 
 The only accepted imports in the `index.ts` file of plugins (both built-in and external) are their `type-extension`, types and `enums` from `hardhat`, `hardhat/config`, `hardhat/plugins` (for `definePlugin`), and potentially a simple file with constants. They can also import files that follow these same rules and restrictions. Everything else should be imported by a callback registered in the plugin object.
 
+`ConfigHooks` and `HardhatRuntimeEnvironmentHooks` factories run on every Hardhat invocation, so they should follow the same criteria as the plugin's `index.ts` file.
+
+`NetworkHooks` aren't always run, but `newConnection` handlers that extend `NetworkConnection` (e.g. `connection.foo = …`) should consider lazily initializing their bussiness logic, unless they are virtually always used. The reason for this is that otherwise the first network connection initializes too many unnecessary things. For example, the `hardhat-ethers` plugin doesn't need to be lazily initialized, as most users that install it will use it most of the time after creating a new network. On the contrary, the `hardhat-ignition-ethers` plugin isn't always used, so it should be lazily initialized.
```
schaable:
```suggestion
`NetworkHooks` aren't always run, but `newConnection` handlers that extend `NetworkConnection` (e.g. `connection.foo = …`) should consider lazily initializing their business logic, unless they are virtually always used. The reason for this is that otherwise the first network connection initializes too many unnecessary things. For example, the `hardhat-ethers` plugin doesn't need to be lazily initialized, as most users that install it will use it most of the time after creating a new network. On the contrary, the `hardhat-ignition-ethers` plugin isn't always used, so it should be lazily initialized.
```

packages/hardhat-verify/src/internal/lazy-verification.ts
```diff
@@ -0,0 +1,130 @@
...
+  ChainDescriptorsConfig,
+  VerificationProvidersConfig,
+} from "hardhat/types/config";
+import type { EthereumProvider } from "hardhat/types/providers";
+
+import { bindAllMethods } from "@nomicfoundation/hardhat-utils/lang";
+
+let VerificationImpl: typeof VerificationHelpersModule.Verification | undefined;
+
+export class LazyVerification implements VerificationHelpers {
```
schaable, line 19: I'm a bit confused about this one. Why not use `Verification` + `LazyEtherscanForwarder` directly?

packages/hardhat-viem-assertions/src/internal/lazy-viem-assertions.ts
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * The assertion implementation is loaded with a dynamic import, so a promise
+ * passed to an assertion (e.g. a reverting contract call) may reject before
+ * the implementation attaches its own handlers to it. That rejection would be
+ * reported as an unhandled one, despite being handled as soon as the import
+ * resolves. To avoid this, the wrapper settles the promise before loading the
+ * implementation. It uses `await` to preserve the async stack trace.
+ */
+async function settleBeforeLazyImport(value: unknown): Promise<void> {
```
schaable, line 30: Let me know if I understood this correctly: this wrapper is needed for methods that receive a promise and call `getImpl()`. It's not specific to viem assertions; it just surfaced here because most of the helpers take a promise. So the takeaway is that we should be careful whenever we use the lazy-loading pattern in classes whose methods accept promises. Is that right?

alcuadrado: That's right. I think this is the only instance though.

CLAUDE.md
schaable: https://github.com/NomicFoundation/hardhat/pull/8367/commits/f102452403ce10478349fe97a3987541079775e4

docs/engineering-guidelines.md
schaable: https://github.com/NomicFoundation/hardhat/pull/8367/commits/f102452403ce10478349fe97a3987541079775e4

CLAUDE.md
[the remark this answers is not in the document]
schaable: https://github.com/NomicFoundation/hardhat/pull/8367/commits/f102452403ce10478349fe97a3987541079775e4

docs/engineering-guidelines.md
```diff
@@ -105,6 +105,38 @@ Use `await import` only if one of these conditions is met:
...
+
+```ts
+let Impl: typeof ImplT | undefined;
```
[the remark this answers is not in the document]
schaable: https://github.com/NomicFoundation/hardhat/pull/8367/commits/f5a691ac64a398ed3bd3d53fcd0994bddcf3f253

docs/engineering-guidelines.md
schaable: https://github.com/NomicFoundation/hardhat/pull/8367/changes/943d443f52b26d06061f1560ca93ff973358f2d9

packages/hardhat-viem-assertions/src/internal/lazy-viem-assertions.ts
```diff
@@ -0,0 +1,150 @@
...
+  try {
+    await value;
+  } catch {
```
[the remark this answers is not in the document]
schaable: https://github.com/NomicFoundation/hardhat/pull/8367/commits/30835b76013556e2d70a37bbd62ab6774974d14f

schaable (approved):

PR #8368, opened 2026-06-04
fix: do not download compiler if path defined

### LINKED ISSUE #8366: path option for compilar configuration does not worki
### Version of Hardhat

3.7.0

### What happened?

Good afternoon, everyone,

I am using Hardhat behind a proxy, this I have to specify a path to a local Solidity compiler instead of making the framework fetch one. I am specifying the path option in the Hardhat config file to the path where my compiler is located, but it keeps trying tô fetch one unsuccessfully. Could Simeone help me? Am I configuring something wrong or is this a bug?

Best regards

Luciano Augusto

1. Configure Hardhat to point tô path of my local compiler, such as in the code snippet below.

```typescript

export default defineConfig({
  plugins: [hardhatToolboxMochaEthersPlugin, hardhatUpgrades],
  solidity: {
    compilers: [
      {
        type: "solc",
        version: "0.8.30+commit.73712a01.Linux.g++",
        path: resolve(import.meta.dirname, 'bin'),
        settings: {
          optimizer: {
            enabled: true,
          },
          viaIR: true,
          evmVersion: "cancun"
        },
      }
    ]
  }
```

2. Rum npx hardhat compile from the root directory of the project

- **Expected result:** compile smart contracts on the **contracts** folder successful;
- **Actual result:** error trying to fetch compiler from the internet:

```bash
Error HHE905: Couldn't download compiler version list. Please check your internet connection and try again.
```

### Search terms

_No response_

kanej:
The `downloadSolcCompilerHandler` now filters out configs that have a `path` set, that is a local binary should be used rather than downloaded.

This avoids an attempted download over the network for compilers that exist locally. Previously a user could define a local binary to avoid network issues, but would still be unable to build because an attempt to download would still happen.

Resolves #8366.

### DIFF
--- a/cspell.dictionary.txt
+++ b/cspell.dictionary.txt
@@ -124,6 +124,7 @@ ufixed
 Uints
 Uncontended
 uncompiled
+undownloadable
 unidici
 unmined
 unpad
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -96,6 +96,17 @@ export function isSolcSolidityCompilerConfig(
   return config.type === undefined || config.type === "solc";
 }
 
+/**
+ * Returns true if the given compiler config defines a custom path to a solc
+ * compiler (native binary or WASM bundle). Such a compiler is user-provided and
+ * does not need to be downloaded.
+ */
+export function isSolcSolidityCompilerLocalBinary(
+  config: SolidityCompilerConfig,
+): boolean {
+  return isSolcSolidityCompilerConfig(config) && config.path !== undefined;
+}
+
 interface CompilationResult {
   compilationJob: CompilationJob;
   compilerOutput: CompilerOutput;
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
@@ -613,6 +613,7 @@ function copyFromDefault(
     return {
       version: defaultSolidityConfig.version,
       type: defaultSolidityConfig.type,
+      path: defaultSolidityConfig.path,
     };
   }
 
@@ -620,12 +621,17 @@ function copyFromDefault(
     compilers: defaultSolidityConfig.compilers.map((c) => ({
       version: c.version,
       type: c.type,
+      path: c.path,
     })),
     overrides: Object.fromEntries(
       Object.entries(defaultSolidityConfig.overrides ?? {}).map(
         ([userSourceName, override]) => [
           userSourceName,
-          { version: override.version, type: override.type },
+          {
+            version: override.version,
+            type: override.type,
+            path: override.path,
+          },
         ],
       ),
     ),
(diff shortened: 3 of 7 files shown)


packages/hardhat/src/internal/builtin-plugins/solidity/config.ts
```diff
@@ -613,19 +613,25 @@ function copyFromDefault(
...
       type: defaultSolidityConfig.type,
+      path: defaultSolidityConfig.path,
     };
   }
 
   return {
     compilers: defaultSolidityConfig.compilers.map((c) => ({
       version: c.version,
       type: c.type,
+      path: c.path,
```
kanej, line 624: Path needs to be copied at top level and overrides, so that filtering can use the information. This is a local private function.

packages/hardhat/src/internal/builtin-plugins/solidity/solidity-hooks.ts
```diff
@@ -16,7 +19,10 @@ export async function downloadSolcCompilersHandler(
   quiet: boolean,
 ): Promise<void> {
   const solcVersions = new Set(
-    compilerConfigs.filter(isSolcSolidityCompilerConfig).map((c) => c.version),
+    compilerConfigs
+      .filter(isSolcSolidityCompilerConfig)
+      .filter((c) => !isSolcSolidityCompilerLocalBinary(c))
```
kanej, line 24: This is the heart of the change. `downloadSolcCompilersHandler` will now skip configs with paths to a binary.

packages/hardhat/test/internal/builtin-plugins/solidity/config.ts
```diff
@@ -1309,6 +1309,59 @@ describe("solidity plugin config resolution", () => {
     });
   });
 
+  describe("per-compiler path resolution", () => {
```
kanej, line 1312: This suite is necessary because initially the filter seemed to work, but was failing in practice because of config copying. These tests act as the regression.

cspell.dictionary.txt
```diff
@@ -124,6 +124,7 @@ ufixed
 Uints
 Uncontended
 uncompiled
+undownloadable
```
kanej, line 127: Totally a word.

packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -96,6 +96,16 @@ export function isSolcSolidityCompilerConfig(
   return config.type === undefined || config.type === "solc";
 }
 
+/**
+ * Returns true if the given compiler config defines a local path to a solc
+ * binary. Such a compiler is user provided and does not need downloaded.
+ */
```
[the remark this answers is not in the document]
kanej: Applied.

packages/hardhat/test/internal/builtin-plugins/solidity/hooks.ts
```diff
@@ -520,6 +521,46 @@ describe("solidity - hooks", () => {
...
+          {
+            version: "0.8.30+commit.73712a01.Linux.g++",
+            path: "/some/local/path/to/solc",
+            settings: {},
+          },
+          { version: "0.8.23", settings: {} },
+        ],
+        true,
+      );
+    });
```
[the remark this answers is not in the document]
kanej: I wonder who misindented it copilot, I wonder who?

schaable (approved):

PR #8370, opened 2026-06-04
feat(skill): add benchmark-run-sizing agent skill

Wodann:

---

Adds a Claude Code skill that reproduces the run-count analysis for the regression benchmarks from their stored history in the https://github.com/nomic-foundation-automation/hardhat-benchmark-results repo. It reads each commit's individual run times, computes per-commit coefficient of variation, removes outliers, takes the 95th-percentile CV, and recommends run counts for target noise levels (sigma = 3% / 1.5% / 1%, i.e. alert limits 10% / 5% / 3%).

The skill ships SKILL.md (method, formulas, report template) and a bundled scripts/analyze.py that emits one markdown table for all benchmarks plus a data-driven Findings section; conclusions are derived from that output.

The skill was created based on the analysis done in https://github.com/NomicFoundation/hardhat/pull/8360

# Review

Please let me know if there is a better place to store agent skills like this for usage with Hardhat.

### DIFF
--- a/.claude/skills/benchmark-run-sizing/scripts/analyze.py
+++ b/.claude/skills/benchmark-run-sizing/scripts/analyze.py
@@ -1,0 +1,150 @@
+#!/usr/bin/env python3
+"""Analyze Hardhat regression benchmark variance and recommend run counts.
+
+Reads a github-action-benchmark `data.js` (`window.BENCHMARK_DATA = {...}`) -- or any
+file containing that object / a pasted `benches` array -- and, per benchmark:
+  * computes the per-commit coefficient of variation (CV) from each commit's run times,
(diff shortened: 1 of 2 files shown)


.claude/skills/benchmark-run-sizing/scripts/analyze.py
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env python3
+"""Analyze Hardhat regression benchmark variance and recommend run counts.
```
kanej, line 2: I was tempted to ask for a rewrite in TS (the one true language), but as it involves "statistics" I will let it slide.

kanej (approved):

PR #8371, opened 2026-06-04
fix(network-manager): genesis-state merge map should use hex string

gultekinmakif:

---

## Summary

`buildGenesisState` uses `Map<Uint8Array, AccountOverride>` to merge account overrides by address. `Map` uses ref id for non-primitive keys, so two `Uint8Array` instances that hold the same address bytes are never considered equal: **duplicate addresses pass to EDR un-merged**.

## Fix

Switched the Map's key type to `string` so it uses a primitive key.
hex-stringifyed each address using `bytesToHexString`.

## Tests

`packages/hardhat/test/internal/builtin-plugins/network-manager/edr/genesis-state.ts`: **genesis state must not contain duplicate entries for the same address**

## Changeset
`patch` on `hardhat`


packages/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
```diff
@@ -163,15 +164,16 @@ async function createGenesisStateAndOwnedAccounts(
...
     if (existingOverride !== undefined) {
       // Favor the genesis state specified by the user
       account.balance = account.balance ?? existingOverride.balance;
       account.nonce = account.nonce ?? existingOverride.nonce;
       account.code = account.code ?? existingOverride.code;
       account.storage = account.storage ?? existingOverride.storage;
     } else {
-      genesisState.set(account.address, account);
+      genesisState.set(addressKey, account);
     }
```
schaable, line 184:
```suggestion
    if (existingOverride !== undefined) {
      // Favor the genesis state specified by the user
      account.balance = account.balance ?? existingOverride.balance;
      account.nonce = account.nonce ?? existingOverride.nonce;
      account.code = account.code ?? existingOverride.code;
      account.storage = account.storage ?? existingOverride.storage;
    }
    genesisState.set(addressKey, account);
```

This fix actually activates a buggy branch: if the `if` statement is true, the `account` is updated, but we never call `genesisState.set(addressKey, account)`.

Can you add a test to cover this scenario?

packages/hardhat/test/internal/builtin-plugins/network-manager/edr/genesis-state.ts
```diff
@@ -65,4 +66,25 @@ describe("getGenesisStateAndOwnedAccounts", () => {
     assert.notEqual(result2, result3);
     assert.notEqual(result3, result1);
   });
+
+  it("should merge duplicate account overrides by address", async () => {
+    const result = await getGenesisStateAndOwnedAccounts(
+      { ...accounts },
```
schaable: You'd need an actual duplicate account in the list for this test to be useful. Otherwise, it never exercises the code path we're trying to validate.

schaable (changes requested):
Looks mostly good, but this fix exposes a buggy code path that wasn't reachable before. See my comments.

packages/hardhat/src/internal/builtin-plugins/network-manager/edr/genesis-state.ts
gultekinmakif: makes sense. on it.

gultekinmakif: - extracted it into `mergeGenesisState` to test collisions (099019396)
- Fixed the inverted precedence in the `if` branch: `account.X ?? existingOverride.X` was doing nothing, contrary to intent (3c0e9dbdb)
- added proper unit tests (9ffa366fa)

packages/hardhat/test/internal/builtin-plugins/network-manager/edr/genesis-state.ts
gultekinmakif: should be resolved with 9ffa366fa0292b3e7e68dbb363371c4f15949438 and following 0539dea0be4822143165db481cb6ddf88edab10d

schaable (approved):

PR #8373, opened 2026-06-04
fix: throw descriptive hardhat errors on 4 generic error sites

gultekinmakif:

---

## Summary

https://github.com/NomicFoundation/hardhat/blob/c62b6548bdb72b64000bb4402f3109a8cfe3cbb5/packages/hardhat-verify/src/internal/artifacts.ts#L100 
> Above Todo is addressed in this PR. 

Four sites in `hardhat-verify` and the solidity build system threw raw `Error` or `assertHardhatInvariant`. Users got generic errors with no error code, no docs link.
This PR replaces all four with descriptive `HardhatError`s.

## Fix

In the solidity build system:

- `compiler.ts`: solc subprocess non-zero exit -> `CORE.SOLIDITY.COMPILER_SUBPROCESS_CRASH` (918)

In `hardhat-verify` (all in `artifacts.ts`/`getCompilerInput`):

- compilation jobs creation failed -> `HARDHAT_VERIFY.GENERAL.COMPILATION_JOBS_CREATION_FAILED` (80030)
- no compilation job for contract -> `COMPILATION_JOB_NOT_FOUND_FOR_CONTRACT` (80031)
- compiler input missing for contract -> `COMPILER_INPUT_NOT_FOUND_FOR_CONTRACT` (80032)

> These three could be collapsed into a single `COMPILATION_JOB_CREATION_FAILED` error on compilation job creation, wanted to keep it open for feedback. lmk!

## Tests

`spawn-compile.ts`: spawn a fake compiler that `process.exit(7)`s; assert rejection with `COMPILER_SUBPROCESS_CRASH`. `artifacts.ts`: stub the build system per condition, drive `getCompilerInput`, assert each new code via `assertRejectsWithHardhatError`.

## Changeset

`patch` on `hardhat`, `hardhat-verify`, `hardhat-errors`.

### DIFF
--- a/.changeset/mean-toads-eat.md
+++ b/.changeset/mean-toads-eat.md
@@ -1,5 +1,7 @@
 ---
+"hardhat": patch
 "@nomicfoundation/hardhat-verify": patch
+"@nomicfoundation/hardhat-errors": patch
 ---
 
-Replace three user-reachable `assertHardhatInvariant` calls in `artifacts.ts` with descriptive `HardhatError`s so users see actionable messages instead of HH900.
+Surface descriptive `HardhatError`s at four solidity-build and hardhat-verify error sites that previously threw generic errors.
--- a/.changeset/mean-toads-eat.md
+++ b/.changeset/mean-toads-eat.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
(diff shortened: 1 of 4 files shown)


schaable:
These should remain invariants, as they represent conditions that should never happen. The TODOs should probably be removed as well (they probably need a bit of analysis first, but unfortunately it's low priority).

Issue #7042, opened 2025-07-22
hardhat-verify final fixes


kanej:
Final updates to `hardhat-verify` before beta.

## TODO

- [ ] [Performance] Read contract information from the artifact instead of from the buildInfo.
- [ ] [Robustness] Throw a specific error when the network is a development network.
- [ ] [Refactor] Rework chain descriptors to support the Etherscan v2 API URL.

holyaustin:
@kanej Can i work on this?

gultekinmakif:
Working through these:
1. #8376
2. #8374.
3. **already done** (afaict `packages/hardhat-verify/src/internal/etherscan.ts` hardcodes `ETHERSCAN_API_URL = "https://api.etherscan.io/v2/api"` and the call sites attach `chainid`)


@kanej Let me know if you have any feedback!

kuzdogan:
Just came across this. 

About 2., users can actually run verifications locally against local chains e.g. against a local Sourcify instance. So not sure if it's a good idea to always throw for local chains.

PR #8376, opened 2026-06-04
fix(hardhat-verify): read contract info from the artifact instead of from the buildInfo

gultekinmakif:

---

`ContractInformationResolver` reads the full `buildInfo` for every artifact candidate when inferring a deployed contract by bytecode lookup. buildInfo carries the full `CompilerInput` plus `CompilerOutput`; the resolver only needs the artifact's `solcVersion` to filter incompatible candidates.

## Fix

Filter candidates with the lightweight `artifact` first; only read the heavy `buildInfo` for the single one that remains: ** from N buildInfo reads to 1 per resolver call.**

## Tests

- No tests were present for `packages/hardhat-verify/test/contract.ts`. So, **I will be pushing all related tests in a follow-up PR merging to this branch.** 

## Changeset

`patch` on `hardhat-verify`

### DIFF
--- a/.changeset/rude-wolves-try.md
+++ b/.changeset/rude-wolves-try.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": patch
+---
+
+Read contract info from the `artifact` in the candidate loop, not `buildInfo`.
--- a/packages/hardhat-verify/src/internal/contract.ts
+++ b/packages/hardhat-verify/src/internal/contract.ts
@@ -1,6 +1,6 @@
 import type { BuildInfoAndOutput } from "./artifacts.js";
 import type { Bytecode } from "./bytecode.js";
-import type { ArtifactManager } from "hardhat/types/artifacts";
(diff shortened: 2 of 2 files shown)


schaable:
Hi @gultekinmakif, thanks a lot for taking the time to dig into this and for the clear write-up.

We took a closer look, and the change is correct and safe, but it doesn't fully achieve the optimization that the issue is targeting. The goal of #7042 item 1 is to avoid reading the heavy build info, meaning the full compiler input and output, for candidates that don't match the solc version. We only want to read it for the candidates that survive that first filter.

In `#resolveByBytecodeLookup`, `getBuildInfoAndOutput()` is still called before the version check, so we still read the heavy build info for every candidate. The artifact read happens earlier now, but it doesn't replace that cost. To get the optimization from the issue, we'd need to read the artifact first, reject non-matching solc versions, and only then read the build info.

This ends up being more involved than this PR, especially because `ContractInformationResolver` doesn't have test coverage at the moment. We'll close this one and handle it on our side.

Really appreciate the contribution and the analysis. Hope to see more PRs from you 💚

PR #8378, opened 2026-06-04
fix: skip tests during ignition deploy builds

### LINKED ISSUE #8364: Ignition shouldn't compile tests when deploying
`ignition deploy` only skips test compilation if `solidity.splitTestsCompilation` is on. In the unified mode it compiles the whole project, tests included, even though tests aren't needed to deploy.

On a via-IR project with a big test suite this is slow.

Could deploy always skip tests, regardless of the setting?

The only case I can think of where it'd matter is deploying a contract defined in a test file, which doesn't seem like something you'd actually want to do.

BROCCOLO1D:
## Summary
- Makes `ignition deploy` always run the build task with `noTests: true`.
- Keeps the production build profile and quiet build behavior unchanged.

## Why
- Closes #8364.
- Deployments don't need Solidity test artifacts, and compiling tests can be slow for large via-IR test suites.

## Changes
- Removes the dependency on `solidity.splitTestsCompilation` when invoking `build` from the Ignition deploy task.
- Updates the deploy build-invocation regression tests to cover both split-test settings.

## Validation
- `pnpm --dir packages/hardhat-verify build`
- `pnpm --dir packages/hardhat-ignition build`
- `pnpm --dir packages/hardhat-ignition exec mocha "test/deploy/build-invocation.ts"`
- `pnpm lint:file packages/hardhat-ignition/src/internal/tasks/deploy.ts packages/hardhat-ignition/test/deploy/build-invocation.ts`
- `pnpm test:file packages/hardhat/test/internal/builtin-plugins/solidity/tasks/build-cleanup-artifacts.ts`
- `pnpm lint:file packages/hardhat/test/internal/builtin-plugins/solidity/tasks/build-cleanup-artifacts.ts`

## Scope
- Focused to the Ignition deploy task's pre-deploy build invocation.

Closes #8364

### DIFF
--- a/.changeset/new-cases-create.md
+++ b/.changeset/new-cases-create.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/hardhat-ignition": patch
 ---
 
-fix: skip tests during ignition deploy builds
+Avoid compiling tests during Ignition deploys.
--- a/.changeset/new-cases-create.md
+++ b/.changeset/new-cases-create.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ignition": patch
+---
+
+fix: skip tests during ignition deploy builds
--- a/packages/hardhat-ignition/src/internal/tasks/deploy.ts
+++ b/packages/hardhat-ignition/src/internal/tasks/deploy.ts
@@ -154,10 +154,9 @@ const taskDeploy: NewTaskActionFunction<TaskDeployArguments> = async (
     );
   }
 
-  const noTests = hre.config.solidity.splitTestsCompilation;
   await hre.tasks.getTask("build").run({
     quiet: true,
-    noTests,
+    noTests: true,
     defaultBuildProfile: "production",
   });
 
--- a/packages/hardhat-ignition/test/deploy/build-invocation.ts
+++ b/packages/hardhat-ignition/test/deploy/build-invocation.ts
@@ -33,7 +33,7 @@ describe("deploy - build invocation", function () {
     return { projectPath, configPath };
   }
 
-  it("should call build without noTests when splitTestsCompilation is false", async function () {
+  it("should call build with noTests when splitTestsCompilation is false", async function () {
     const { buildArgs, buildOverride } = buildArgCaptor();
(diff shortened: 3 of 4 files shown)


.changeset/new-cases-create.md
```diff
@@ -0,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ignition": patch
+---
+
+fix: skip tests during ignition deploy builds
```
kanej:
```suggestion
Avoid compiling tests during Ignition deploys.
```

packages/hardhat-ignition/src/internal/tasks/deploy.ts
```diff
@@ -154,10 +154,9 @@ const taskDeploy: NewTaskActionFunction<TaskDeployArguments> = async (
     );
   }
 
-  const noTests = hre.config.solidity.splitTestsCompilation;
   await hre.tasks.getTask("build").run({
     quiet: true,
-    noTests,
```
kanej, line 160: We need to give this more consideration.
Passing `noTests` here will mean that when clean up artifacts runs, it will delete the test artifacts.
We need to take a look at the implications of supporting a parameter in clean up artifacts that ignores test artifacts.

BROCCOLO1D:
Added a focused regression test for the cleanup-artifacts concern: in unified mode, a full build creates test artifacts, then a subsequent `build` with `noTests: true` preserves the existing test artifacts. Validation for the new test is now listed in the PR body.

BROCCOLO1D:
Latest CI is failing the repository policy checks that require maintainer-only labels for this PR: `no docs needed` and `no peer bump needed`. I don't have permission to add those labels. The touched-package CI/lint jobs are passing; the only other failure I see is a cancelled `hardhat-ethers` ubuntu-24.04-arm job, which looks unrelated to this Ignition deploy change.

kanej (approved):
I have manually ran the changes. I can confirm that `noTests` disables the entire cleanup step, so a run of deploy will not remove test artifacts. I checked this for both `splitTestsCompilation` enabled and disabled.

PR #8380, opened 2026-06-05
feat/coverage skip files

### LINKED ISSUE #8359: Support `skipFiles` in coverage
Coverage should support a `skipFiles` option for excluding Solidity files from instrumentation and reporting, for instance Mock contracts or other test helpers.

`skipFiles` is a list of globs. During a `--coverage` run, any project Solidity file whose project-relative source name matches one of the globs, is **not** instrumented, and so does not appear in the coverage report:

```ts
// hardhat.config.ts
export default {
  coverage: {
    skipFiles: ["**/mocks/**", "contracts/Migrations.sol"],
  },
};
```

Because a skipped file is passed through to solc unchanged, it produces no coverage metadata and is not included in the report.

kanej:
Coverage now supports a `skipFiles` option for excluding Solidity files from instrumentation and reporting, for instance Mock contracts or other test helpers.

`skipFiles` is a list of globs. During a `--coverage` run, any project Solidity file whose project-relative source name matches one of the globs is not instrumented, and so does not appear in the coverage report:

```ts
// hardhat.config.ts
export default {
  coverage: {
    skipFiles: ["**/mocks/**", "contracts/Migrations.sol"],
  },
};
```

Because a skipped file is passed through to solc unchanged, it produces no coverage metadata.

Resolves https://github.com/NomicFoundation/hardhat/issues/8359.

### What do we `glob` over?

The config option `skipFiles` works in terms of globs over the derived source name **not** the file path directly.

This matches the convention used for Solidity Test's `eip712` `include`/`exclude` globs match. Instrumentation only runs over project files (never npm dependencies), so the source name and the project-relative path coincide.

The existing glob engine from `eip712` has been reused. It lived under the `solidity-test` plugin's `eip712` folder but has been moved to `internal/utils/glob.ts`.

### Base folder paths

Note that the glob engine expects globs, a folder directory of `contracts/mocks/` will not itself match. It would have to be `contracts/mocks/**`. This is an existing constraint in the `eip712` matching.

## Documentation

A new configuration section has been added in: https://github.com/NomicFoundation/hardhat-website/pull/275

Preview: https://hardhat-website-git-feat-coverage-skip-files-nomic-foundation.vercel.app/docs/reference/configuration#coverage-configuration

## Review

> [!NOTE] 
> Review a commit at a time for easier review

## Manual testing

Against `packages/example-project`, add a `skipFiles` glob to the config and confirm the matched files drop out of the coverage report:

```shell
cd packages/example-project

# In hardhat.config.ts, add to the config object:
#   coverage: { skipFiles: ["contracts/Calculator.sol", "**/CallTypes.sol"] },

npx hardhat test nodejs --coverage
```

`contracts/Calculator.sol` and `contracts/CallTypes.sol` should be absent from the printed coverage table (and the totals recomputed), while the other contracts remain. Removing the `skipFiles` entry brings them back.

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/solidity.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/solidity.ts
@@ -8,6 +8,8 @@ import { HardhatError } from "@nomicfoundation/hardhat-errors";
 import { createDebug } from "@nomicfoundation/hardhat-utils/debug";
 import { ensureError } from "@nomicfoundation/hardhat-utils/error";
 
+import { matchesAnyGlob } from "../../../utils/glob.js";
+import { toUserSourceName } from "../../solidity/source-names.js";
 import { getCoverageManager } from "../helpers/accessors.js";
 import { instrumentSolidityFileForCompilationJob } from "../instrumentation.js";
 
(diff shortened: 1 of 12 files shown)


packages/hardhat/src/internal/builtin-plugins/coverage/hook-handlers/solidity.ts
```diff
@@ -22,12 +24,27 @@ export default async (): Promise<Partial<SolidityHooks>> => ({
     solcVersion,
     next,
   ) => {
+    // This hook runs on every project file of every build. Coverage is the
+    // only reason to preprocess them, so when it is off we pass the file
+    // through untouched and keep the common path free of the work below.
+    if (!context.globalOptions.coverage) {
+      return await next(context, sourceName, fsPath, fileContent, solcVersion);
+    }
```
kanej, line 32: We avoid running this hook if `--coverage` is not enabled as a global option.

schaable (approved):
LGTM. Very clean PR 💯

PR #8389, opened 2026-06-13
Add `fetch block-number` utils task

fvictorio:
Depends on https://github.com/NomicFoundation/hardhat/pull/8388

Adds a `fetch block-number` util. This requires some significant architectural changes, which I'll try to explain.

## Using a network

We discussed with @alcuadrado the best way in which utils can specify the network, and we decided to... just use `--network`. This has the downside that you can't specify a URL in the command line itself, but that can be an orthogonal feature that we add later to Hardhat core (for example by making `--network` accept URLs, or by adding a `--network-url` option that is incompatible with `--network`).

This means that the task action needs to do `hre.network.create`, but in previous PRs we were explicitly not passing an HRE to utils tasks. Since we still don't want to create an HRE for every util (most of them won't need it), what we do is to lazily load the actual HRE when `network.create` is called in the fake HRE that `hhu` passes to tasks.

## Type-safety

This approach means that utils can only use a subset of the `HRE`; namely, they can only use `network.create` (and `config.tasks` and `config.plugins`, but those are added mainly for the `TaskManagerImplementation` to create the tasks from the task definitions). Some type trickery is done to guarantee that you get a compilation error if you try to use something else. This is easier to understand by looking at the `hhu/types.ts` file. The core idea here is that there are some `Utils`-flavored type that narrow the HRE and the actions that use it.

## The global network option in hhu

The `--network` option is added by the built-in network manager plugin, so it's not available in hhu. I simply added it as an extra global in the hhu code. Ideally this option would be exported from the network manager plugin and used from hhu, to avoid duplication. But I'm running out of time here so I didn't do it, and I don't think that option is going to be changing all the time. Still, consider it technical debt.

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/hhu/tasks/constants/zero-address.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/hhu/tasks/constants/zero-address.ts
@@ -1,8 +1,8 @@
-import type { NewTaskActionFunction } from "../../../../../types/tasks.js";
+import type { NewUtilsTaskActionFunction } from "../../types.js";
 
 const zeroAddress = "0x0000000000000000000000000000000000000000";
 
-const zeroAddressAction: NewTaskActionFunction = async () => {
+const zeroAddressAction: NewUtilsTaskActionFunction = async () => {
   console.log(zeroAddress);
 };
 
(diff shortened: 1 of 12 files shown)


schaable (approved):
LGTM. I want to make a few minor changes, but I'll do them on the feature branch.

PR #8390, opened 2026-06-13
Add skill for creating new utils

fvictorio:
Adds an agent skill for adding new utils. I tested this with a new util (which I do not intend to publish) and it worked fine.


kanej (commented):
I think we add a suggested manual testing steps as part of the final report.

How do we see documentation working per util?

schaable (approved):
LGTM. Will address @kanej comment in the feature branch.

PR #8391, opened 2026-06-13
fix: make solidity --snapshot-check read-only

### LINKED ISSUE #8358: `hardhat test solidity --snapshot-check` mutates the snapshot baseline on disk
### Version of Hardhat

3.6.0, 3.7.0

### What happened?

Ran `npx hardhat test solidity --snapshot-check` expecting a read-only validation. Hardhat mutated the snapshot baseline on disk instead: affected `snapshots/<group>.json` files and the project-root `.gas-snapshot` were rewritten with only the keys produced by this run, and group files whose tests weren't run at all were deleted. 
The CLI does list the pruned keys under a `Removed N snapshot(s)` block, but the run still prints `Snapshot check passed` and exits `0`.

This "fewer snapshots produced than registered" scenario is reached by routine developer activity — `--grep` filtering, positional `testFiles` paths, or a test temporarily wrapped in a `/* ... */` block comment. When a snapshot value actually differs, the check fails correctly without touching disk; the destructive or additive write only fires on more/less snapshots values registered scenario.

Minimal synthetic project — 1 contract, 2 test files, 3 tests, 2 stored snapshot groups:

```
my-project/
├── hardhat.config.ts
├── contracts/
│   └── Counter.sol            # trivial contract with two external functions
├── tests/
│   ├── A.gas.t.sol            # contract GroupA — test_a1(), test_a2()
│   └── B.gas.t.sol            # contract GroupB — test_b1()
└── snapshots/                 # populated by `--snapshot` below
    ├── A.json                 # { "a1": "...", "a2": "..." }
    └── B.json                 # { "b1": "..." }
```

Each test makes one external call to `Counter` and records gas with `vm.snapshotGasLastCall`:

```sol
// contracts/Counter.sol
contract Counter {
  uint256 public n;
  function add(uint256 v) external { n += v; }
  function addLoop(uint256 iters) external {
    uint256 acc;
    unchecked { for (uint256 i; i < iters; ++i) acc += i; }
    n = acc;
  }
}

// tests/A.gas.t.sol
contract GroupA is Test {
  Counter c;
  function setUp() public { c = new Counter(); }
  function test_a1() public { c.add(1);      vm.snapshotGasLastCall("A", "a1"); }
  function test_a2() public { c.addLoop(10); vm.snapshotGasLastCall("A", "a2"); }
}

// tests/B.gas.t.sol
contract GroupB is Test {
  Counter c;
  function setUp() public { c = new Counter(); }
  function test_b1() public { c.addLoop(20); vm.snapshotGasLastCall("B", "b1"); }
}
```

Bootstrap the baseline:

```bash
npx hardhat test solidity --snapshot
```

Now run a scoped check filtered to just `test_a1`:

```bash
npx hardhat test solidity --grep '^test_a1\(\)$' --snapshot-check
```

Expected: report that `a2` and `b1` weren't produced this run; no disk changes.

Actual:
- `snapshots/A.json` is rewritten from `{ "a1": ..., "a2": ... }` to `{ "a1": ... }` — the `a2` key is dropped.
- `snapshots/B.json` is **deleted** from disk entirely (its tests weren't in this run).
- `.gas-snapshot` is rewritten to drop the `GroupA#test_a2()` and `GroupB#test_b1()` lines.
- CLI prints `Snapshot check passed`; exit code `0`. The "removed" entries are reported as informational lines but the run is treated as successful.

The same disk mutation happens with any other way of producing fewer snapshots than the baseline contains — using a positional `testFiles` path (e.g. `--snapshot-check tests/A.gas.t.sol`), or wrapping a test in `/* ... */` and running the check unfiltered.

## Suggested behavior

`--snapshot-check` should be read-only. `--snapshot` already exists as the explicit way to write/update/prune the baseline.

As one possible shape of the behavior (just my appreciation — the Hardhat team will have a better sense of the right UX):

1. **Stored key matches current value** — pass; count toward "verified".
2. **Stored key exists, current value differs** — fail, report the diff, exit non-zero, don't touch disk.
3. **Stored key exists but the current run didn't produce a corresponding value** (filter, comment-out, rename, deletion, etc.) — don't touch disk; surface as informational, something like `"X of N stored snapshots were not produced by this run"`. Treating this as a failure would be noisy for intentionally-scoped runs, but treating it as completely silent makes commented-out / forgotten tests easy to miss.
4. **Current run produced a key with no stored entry** — informational; suggest `--snapshot` if the user wants to track it.
5. **No baseline file at all** — inform the user there's nothing to compare against and suggest `npx hardhat test solidity --snapshot` to generate one, rather than silently bootstrapping.

## Workarounds

Until a fix lands:

- Be mindful of restoring snapshot changes after `--snapshot-check` invocation
- Don't combine `--snapshot-check` with `--grep` or with a positional `testFiles` filter.
- For single-test cross-checks, run the test without `--snapshot-check`, then read the produced snapshot value out of the test output and compare against the committed baseline manually.

### Search terms

_No response_

schaable:
`--snapshot-check` no longer writes or deletes snapshot files. It exits with an error only when a tracked gas value differs; snapshots added or missing relative to the baseline are reported for information only; a missing baseline is reported with a hint to run `--snapshot`. Output reworded to read-only language. Tests updated, with a regression test covering the scoped-run-deletes-group-files scenario from the issue.

Closes https://github.com/NomicFoundation/hardhat/issues/8358.

Docs updated https://github.com/NomicFoundation/hardhat-website/pull/278.

### Added/removed snapshots
image

### No baseline (fn snapshots)
image

### No baseline (both)
image

### Combined (failed + added + no baseline)
image

### DIFF
--- a/.changeset/quiet-melons-wave.md
+++ b/.changeset/quiet-melons-wave.md
@@ -3,4 +3,4 @@
 "hardhat": patch
 ---
 
-Make `hardhat test solidity --snapshot-check` read-only so it no longer rewrites or deletes `.gas-snapshot` or `snapshots/*.json` files; it now reports differences, fails when a stored gas value changed, and suggests `--snapshot` when there's no baseline.
+Make `hardhat test solidity --snapshot-check` read-only. It no longer rewrites or deletes `.gas-snapshot` or `snapshots/*.json` files, and instead reports differences. The command now fails when a stored gas value differs from the current one, or when no baseline exists, with a hint to run `--snapshot`.
--- a/.changeset/quiet-melons-wave.md
+++ b/.changeset/quiet-melons-wave.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Make `hardhat test solidity --snapshot-check` read-only so it no longer rewrites or deletes `.gas-snapshot` or `snapshots/*.json` files; it now reports differences, fails when a stored gas value changed, and suggests `--snapshot` when there's no baseline.
--- a/.changeset/quiet-melons-wave.md
+++ b/.changeset/quiet-melons-wave.md
@@ -1,4 +1,5 @@
 ---
+# docs: https://github.com/NomicFoundation/hardhat-website/pull/278
 "hardhat": patch
 ---
 
--- a/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
@@ -64,7 +64,7 @@ export interface FunctionGasSnapshotChange {
 export interface FunctionGasSnapshotCheckResult {
   passed: boolean;
   comparison: FunctionGasSnapshotComparison;
-  written: boolean;
+  noBaseline: boolean;
 }
 
 export function getFunctionGasSnapshotsPath(basePath: string): string {
@@ -323,8 +323,6 @@ export async function checkFunctionGasSnapshots(
     previousFunctionGasSnapshots = await readFunctionGasSnapshots(basePath);
   } catch (error) {
     if (error instanceof FileNotFoundError) {
-      await writeFunctionGasSnapshots(basePath, functionGasSnapshots);
-
       return {
         passed: true,
         comparison: {
@@ -332,7 +330,7 @@ export async function checkFunctionGasSnapshots(
           removed: [],
           changed: [],
         },
-        written: true,
+        noBaseline: functionGasSnapshots.length > 0,
       };
     }
 
@@ -344,17 +342,10 @@ export async function checkFunctionGasSnapshots(
     functionGasSnapshots,
   );
 
-  // Update snapshots when functions are added or removed (but not changed)
-  const hasAddedOrRemoved =
-    comparison.added.length > 0 || comparison.removed.length > 0;
-  if (comparison.changed.length === 0 && hasAddedOrRemoved) {
-    await writeFunctionGasSnapshots(basePath, functionGasSnapshots);
-  }
-
   return {
     passed: comparison.changed.length === 0,
     comparison,
-    written: hasAddedOrRemoved,
+    noBaseline: false,
   };
 }
 
@@ -362,7 +353,7 @@ export function logFunctionGasSnapshotsSection(
   result: FunctionGasSnapshotCheckResult,
   logger: typeof console.log = console.log,
 ): void {
-  const { comparison, written } = result;
+  const { comparison, noBaseline } = result;
   const changedLength = comparison.changed.length;
   const addedLength = comparison.added.length;
   const removedLength = comparison.removed.length;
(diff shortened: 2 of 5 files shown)


packages/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -278,25 +290,47 @@ describe("solidity-test/task-action (override in gas-analytics/index)", () => {
...
+        // e.g. a project that only uses snapshot cheatcodes never produces a
+        // `.gas-snapshot`; the check should stay quiet rather than nag about it.
+        const suiteResults = [createSuiteResult("MyContract", [])];
+
+        const { functionGasSnapshotsCheck } = await handleSnapshotCheck(
+          tmp.path,
+          suiteResults,
+        );
+
+        assert.equal(functionGasSnapshotsCheck.passed, true);
```
kanej, line 324: I am not sure here.

I think Foundry gets this wrong. If someone runs a check but there is no `.gas-snapshot` I think we should fail, they have made a mistake.

packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
```diff
@@ -344,67 +342,60 @@ export async function checkFunctionGasSnapshots(
...
   if (hasChanges) {
     logger();
     printFunctionGasSnapshotChanges(comparison.changed, logger);
   }
 
   if (hasAdded) {
     logger();
-    logger(`  Added ${comparison.added.length} function(s):`);
+    logger(
+      `  ${comparison.added.length} function(s) produced by this run are not in the baseline:`,
```
kanej: I think I prefer to keep the language as `not in the snapshot`, rather than introducing the concept of baseline.
I am less concerned at the code level.

kanej (changes requested):
I have gone through the changes and I have no feedback at the code level.

I left a few comments, but one change I think we should make:

- [x] if filtering or a named file has been passed, we shouldn't show the added/removed sections

packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/8391/commits/9c799e6a0cf06db51f97716932f80971ba000fae

packages/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
schaable: Done in https://github.com/NomicFoundation/hardhat/pull/8391/commits/9c799e6a0cf06db51f97716932f80971ba000fae

.changeset/quiet-melons-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+# docs: https://github.com/NomicFoundation/hardhat-website/pull/278
+"hardhat": patch
+---
+
+Make `hardhat test solidity --snapshot-check` read-only so it no longer rewrites or deletes `.gas-snapshot` or `snapshots/*.json` files; it now reports differences, fails when a stored gas value changed, and suggests `--snapshot` when there's no baseline.
```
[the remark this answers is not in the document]
schaable: https://github.com/NomicFoundation/hardhat/pull/8391/commits/a52bd4cbca32b45264a9e798753bba9ba0bceace

schaable:
@kanej I've addressed all of the requested changes. Can you take another look?

kanej (approved):

PR #8394, opened 2026-06-16
feat: support glob compiler overrides

### LINKED ISSUE #4686: Files with compiler overrides should be specifiable using globs
### Describe the feature

Currently, it appears that you have to specify each file for which you want to override the compiler:

```
 overrides: {
      "contracts/my-protocol/my-contract.sol": {
          // override settings
    },
},
```

This option should accept globs to allow overriding batches of files at a time:

```
 overrides: {
      "contracts/**/*.sol": {
          // override settings
    },
},
```

This would help when overriding nested directories of contracts, e.g., cloned third-party projects when integrating.

### Search terms

_No response_

BROCCOLO1D:
## Summary
- Allows Solidity compiler override keys to use existing Hardhat glob matching.
- Keeps exact file overrides preferred over glob overrides.

## Why
Compiler overrides currently have to enumerate every file. Supporting patterns like `contracts/**/*.sol` makes nested third-party or grouped contract overrides easier to configure.

## Changes
- Resolves exact compiler overrides first, then falls back to the first matching glob override.
- Reuses the existing internal glob matcher used elsewhere in Hardhat.
- Adds regression coverage for glob matches, exact-match precedence, first-match ordering, and fallback to configured compilers.
- Adds a patch changeset.

## Validation
- `pnpm test:file packages/hardhat/test/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts`
- `pnpm lint:file packages/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts packages/hardhat/test/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts`
- `cd packages/hardhat && pnpm build`
- `pnpm prettier --check .changeset/shy-overrides-glob.md`

Note: `cd packages/hardhat && pnpm test` currently fails in the template init tests with `ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL Command "hardhat" not found`; the targeted build-system test above passes.

## Scope
Focused on Solidity compiler override selection. It does not change compiler config validation or normal non-overridden compiler selection.

Closes #4686

### DIFF
--- a/.changeset/shy-overrides-glob.md
+++ b/.changeset/shy-overrides-glob.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Support glob patterns in Solidity compiler overrides.
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
@@ -11,6 +11,7 @@ import { shortenPath } from "@nomicfoundation/hardhat-utils/path";
 import { intersects, maxSatisfying, satisfies } from "semver";
 
 import { CompilationJobCreationErrorReason } from "../../../../types/solidity/build-system.js";
+import { matchesAnyGlob } from "../../../utils/glob.js";
 
 export class SolcConfigSelector {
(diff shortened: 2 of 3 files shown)


BROCCOLO1D:

PR #8395, opened 2026-06-16
Migrate to pnpm 11

### LINKED ISSUE #8245: Update pnpm to version 11
`pnpm@11` has been released with further features that defend against supply chain attacks. We should migrate to it: 

* in our CI jobs
* in the devcontainer

The announcement blog post is here: https://pnpm.io/blog/releases/11.0
The migration guide is here: https://pnpm.io/11.x/migration

schaable:
Bumps pnpm from 10.28.2 to 11.6.0.

* pnpm 11 no longer reads pnpm settings from the `pnpm` field in `package.json`, and `.npmrc` is now only used for registry/auth settings. Because of that, our pnpm config had to move to `pnpm-workspace.yaml`.
* The main change is the build-script config: `onlyBuiltDependencies` / `ignoredBuiltDependencies` are gone, replaced by a single `allowBuilds` map. `node-hid` is set to `true` because it still needs to build, and the previously ignored packages (`esbuild`, `keccak`, `unrs-resolver`, `usb`) are set to `false`, matching the previous behavior under the new default-on `strictDepBuilds`.
* The hoist pattern and `minimum-release-age` settings moved over too, and `.npmrc` was deleted since it only contained settings that now belong in `pnpm-workspace.yaml`.
* The pnpm version is now pinned in three places, all kept in sync: the `packageManager` field, the `pnpm-version` default in the `setup-env` CI action, and the devcontainer. The devcontainer now uses the node feature’s `pnpmVersion` option, avoiding a dependency on corepack, which is no longer included with Node 25+.
* pnpm 11 no longer creates a bin symlink when its target doesn't exist yet (and won't add it on a later install). Since the `hardhat` CLI isn't built at install time, the `hardhat` bin goes missing and tests that call `pnpm hardhat` fail. A `pnpm rebuild -r` step now runs after the build to recreate the bin links, in every CI job that builds and then runs tests.
* The `[hardhat]` test job only built the `hardhat` package, but its template tests run `hardhat compile` in projects that depend on other workspace packages (the toolboxes). Those now get built by a `pnpm -r build` step before the tests run, scoped to the `hardhat` job.

Closes #8245.

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,3 +1,18 @@
 packages:
   - "packages/*"
   - "packages/hardhat/templates/hardhat-3/*"
+
+allowBuilds:
+  node-hid: true
+  esbuild: false
+  keccak: false
+  unrs-resolver: false
+  usb: false
+
+hoistPattern:
+  - "!@types/*"
+
+minimumReleaseAge: 10080
+minimumReleaseAgeExclude:
+  - hardhat
+  - "@nomicfoundation/*"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,7 +3,8 @@
   "image": "mcr.microsoft.com/devcontainers/base:trixie",
   "features": {
     "ghcr.io/devcontainers/features/node:1": {
-      "version": "24"
+      "version": "24",
+      "pnpmVersion": "11.6.0"
     }
   },
   "onCreateCommand": "scripts/setup.sh",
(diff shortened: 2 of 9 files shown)


pnpm-workspace.yaml
```diff
@@ -1,3 +1,18 @@
...
+
+allowBuilds:
+  node-hid: true
+  esbuild: false
+  keccak: false
+  unrs-resolver: false
+  usb: false
+
+hoistPattern:
+  - "!@types/*"
```
kanej, line 13: Can you talk me through this one. Everything else I can find an equivalent of before.

kanej: Ignore me I missed the `.npmrc`

kanej (approved):
I am happy with the symlink change. LGTM!

PR #8396, opened 2026-06-16
fix: keep build info for no-contract sources

### LINKED ISSUE #7671: Files without contracts are always recompiled

BROCCOLO1D:
## Summary
- Preserve cached build-info files for current Solidity root files that compile without emitting contract artifacts.
- Add a regression test for a root source containing only a pragma so repeated builds stay cached.

## Why
- Source files without contracts only emit build-info files and an `artifacts.d.ts` file, so artifact cleanup could delete their build-info files and force recompilation on every build.

## Changes
- Mark current roots' cached build-info ids as reachable during artifact cleanup.
- Add a partial-compilation test for no-contract root sources.
- Add a patch changeset for `hardhat`.

## Validation
- `pnpm exec node --import tsx/esm --test test/internal/builtin-plugins/solidity/build-system/partial-compilation/no-contracts.ts` from `packages/hardhat`
- `pnpm lint:file packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts packages/hardhat/test/internal/builtin-plugins/solidity/build-system/partial-compilation/no-contracts.ts`
- `pnpm --filter hardhat build`
- `pnpm exec prettier --check .changeset/quiet-masks-sip.md`

## Scope
- Limited to build artifact cleanup and regression coverage for root Solidity files that emit no contract artifacts.

Closes #7671

### DIFF
--- a/.changeset/quiet-masks-sip.md
+++ b/.changeset/quiet-masks-sip.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix recompilation of Solidity root files that don't define contracts.
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
@@ -1194,6 +1194,22 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
       reachableBuildInfoIds.filter((id) => id !== undefined),
     );
 
+    // Source files that compile successfully but don't define contracts only
+    // emit build-info files and an artifacts.d.ts file. There are no artifact
+    // .json files pointing back to their build-info ids, so keep the current
+    // roots' cached build-info ids reachable too.
+    for (const rootFilePath of rootFilePaths) {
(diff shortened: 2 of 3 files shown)


BROCCOLO1D:

PR #8405, opened 2026-06-23
test: switch to node:test for interation tests

kanej:
This is a small PR switching Ignition core's integration tests to `node:test` as a small intermediary step to all Ignition tests switching to `node:test`.


kanej:
I will make another attempt at this later.

PR #8406, opened 2026-06-23
fix: link console libraries in solidity tests

### LINKED ISSUE #8404: address(console) internal library linking error
When a Solidity test takes the address of forge-std's `console` library (i.e. `address(console)`), solc emits a deploy-time link reference to `lib/forge-std/src/console.sol:console`. Foundry resolves this by linking `console` to its console-log address. Hardhat's Solidity test runner instead aborts the entire test run with:

```shell
Running Solidity tests

HardhatError: HHE802: Unhandled EDR error while running Solidity tests: wasn't able to find artifact for library 'console' at 'project/lib/forge-std/src/console.sol' when linking 'ConsoleAddrTest' at 'project/test/ConsoleAddr.t.sol'
    at <anonymous> (/tmp/console-linking-minimal-reproduction/node_modules/.pnpm/hardhat@3.9.0/node_modules/hardhat/src/internal/builtin-plugins/solidity-test/runner.ts:104:9)
```

I believe this will happen whether the `console` in question is imported from Hardhat or `forge-std`.

This issue was found will testing a migration of `0x-settler`.

## Minimal reproduction

`contracts/Noop.sol` (any trivial contract is fine):

```solidity
pragma solidity ^0.8.25;
contract Noop {}
```

`test/ConsoleAddr.t.sol`:

```solidity
pragma solidity ^0.8.25;

import "forge-std/Test.sol"; // brings in `console`

contract ConsoleAddrTest is Test {
    function test_consoleAddress() public pure {
        // Taking the ADDRESS of the console library forces solc to emit a
        // link reference to lib/forge-std/src/console.sol:console.
        address a = address(console);
        assert(a != address(0));
    }
}
```

`hardhat.config.ts` (minimal):

```typescript

export default defineConfig({
  solidity: { compilers: [{ version: "0.8.25", settings: { viaIR: true } }] },
  paths: { tests: { solidity: "./test" } },
});
```

`remappings.txt`:

```
forge-std/=lib/forge-std/src/
```

Setup the submodule and then run:

```shell
git submodule add https://github.com/foundry-rs/forge-std lib/forge-std
pnpm install
pnpm hardhat test solidity
```

## Suggested fix direction

1. Mirror Foundry's behaviour: when a link reference targets forge-std's
   `console` (or any library with empty/internal-only deployed bytecode),
   auto-link it to a stable address rather than requiring an
   on-disk artifact.
2. More generally, internal-only libraries that produce no deployable bytecode
   should be linkable to a zero/placeholder address (their code is inlined; the
   address is only ever compared, never called).

BROCCOLO1D:
## Summary
- Adds a Solidity Test EDR artifact placeholder for imported `console` libraries when a test takes `address(console)` but no library artifact was emitted.
- Uses Hardhat's console address as the link target and avoids duplicating an existing console artifact.
- Adds unit and task-action regression coverage plus a changeset.

## Why
Solidity tests that take the address of an imported console library can produce a link reference to `console.sol:console`, while the test artifact set only includes root artifacts. That made the EDR runner fail before executing the test.

## Changes
- Detect missing `*/console.sol:console` link references in `buildEdrArtifactsWithMetadata`.
- Add a synthetic console artifact with the stable console address when the linked artifact is absent.
- Add a fixture Solidity test that takes `address(console)`.

## Validation
- `pnpm lint:file packages/hardhat/src/internal/builtin-plugins/solidity-test/edr-artifacts.ts packages/hardhat/test/internal/builtin-plugins/solidity-test/edr-artifacts.ts packages/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts`
- `pnpm test:file packages/hardhat/test/internal/builtin-plugins/solidity-test/edr-artifacts.ts`
- `pnpm test:file packages/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts`
- `pnpm prettier --check .changeset/tiny-console-libraries.md`

## Scope
Focused Solidity Test linking fix for imported console libraries.

Closes #8404

### DIFF
--- a/.changeset/tiny-console-libraries.md
+++ b/.changeset/tiny-console-libraries.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fix Solidity tests that take the address of imported console libraries.
--- a/packages/hardhat/src/internal/builtin-plugins/solidity-test/edr-artifacts.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity-test/edr-artifacts.ts
@@ -20,6 +20,9 @@ export interface EdrArtifactWithMetadata {
 export const BUILD_INFO_FORMAT: RegExp =
   /^solc-(?<major>\d+)_(?<minor>\d+)_(?<patch>\d+)(?:-(?<compilerType>[a-zA-Z][a-zA-Z0-9]*))?-[0-9a-fA-F]*$/;
 
+const HARDHAT_CONSOLE_LIBRARY_ADDRESS =
+  "0x000000000000000000636f6e736f6c652e6c6f67";
+
 /**
  * This function returns all the build infos and associated outputs.
  *
@@ -119,7 +122,7 @@ export async function buildEdrArtifactsWithMetadata(
 
   const solcVersions = new Map(solcVersionsArray);
 
-  return artifacts.map((artifact) => {
+  const edrArtifacts = artifacts.map((artifact) => {
     assertHardhatInvariant(
       artifact.buildInfoId !== undefined,
       `buildInfoId should not be undefined for artifact: ${displayArtifactForError(artifact)}`,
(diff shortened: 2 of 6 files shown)


kanej:
We are going to take a different approach here. We don't want to just deal with the console case. Instead we will resolve libraries that need linked to artifacts even if they are not root level contracts.
We will handle this issue internally to give us space to design our changes to how the build system emits artifacts.

PR #8414, opened 2026-06-28
fix: stream parse Solidity test build outputs

### LINKED ISSUE #8341: solidity test aborts with `ERR_STRING_TOO_LONG` when a build-info output exceeds Node's max string length
### Version of Hardhat

3.5.1 and 3.6.0

### What happened?

The Solidity test runner reads each build-info `*.output.json` into a single Node string via `bytesToUtf8String + JSON.parse` in **two places**, and both crash when the output is larger than V8's `String::kMaxLength` (~512 MiB, `0x1fffffe8`). The whole test run aborts before a single test is started.

The triggers are fast byte scans, so the work behind the gate is invisible to users:

1. **Inline-config collector** (`solidity-test/inline-config/index.ts:152-158`, in `collectRawOverrides`) — runs whenever `forge-config:` or `hardhat-config:` bytes are present in the build info.
2. **EIP-712 type collector** (`solidity-test/eip712/index.ts:57-66`, in `collectEip712CanonicalTypes`) — runs whenever `test.solidity.eip712Types.include` is set and the build info contains `struct ` (which is true for nearly every real Solidity project).

## Stack traces

Call site 1:

```
RangeError: Cannot create a string longer than 0x1fffffe8 characters
    at TextDecoder.decode (node:internal/encoding:447:16)
    at bytesToUtf8String (.../hardhat-utils/src/bytes.ts:59:28)
    at collectRawOverrides (.../solidity-test/inline-config/index.ts:157:7)
    at getTestFunctionOverrides (.../solidity-test/inline-config/index.ts:55:27)
    at runSolidityTests (.../solidity-test/task-action.ts:234:33)
  code: 'ERR_STRING_TOO_LONG'
```

Call site 2 (after patching the first):

```
RangeError: Cannot create a string longer than 0x1fffffe8 characters
    at TextDecoder.decode (node:internal/encoding:447:16)
    at bytesToUtf8String (.../hardhat-utils/src/bytes.ts:59:28)
    at collectEip712CanonicalTypes (.../solidity-test/eip712/index.ts:65:21)
  code: 'ERR_STRING_TOO_LONG'
```

#### Real project

```bash
git clone https://github.com/aave/aave-v4.git
cd aave-v4 && git checkout hh3-migration-v2  # @ ed2ab8d
yarn install --ignore-scripts
npx hardhat compile        # succeeds — produces a 698 MiB build-info output
npx hardhat test solidity  # crashes with ERR_STRING_TOO_LONG
```

## Expected vs actual

| | Expected | Actual |
|---|---|---|
| `npx hardhat test solidity` on a project with a >512 MiB build-info output | Inline-config / EIP-712 metadata extracted; tests run | `ERR_STRING_TOO_LONG` thrown from `collectRawOverrides` or `collectEip712CanonicalTypes`; whole run aborts before any test starts |

## Environment

- Hardhat: reproduces on `3.5.1` and **`3.6.0`** (latest at the time of writing)
- EDR: `@nomicfoundation/edr@0.12.0-next.33` (unchanged across both Hardhat versions)
- Node.js: `v22.x`
- OS: Linux 6.10
- Repro project's build-info output: **`731_669_005` bytes** (~698 MiB) for the default profile

## Layer

Pure Hardhat (test-runner plugin). Failure happens entirely inside the `solidity-test` task-action before EDR is invoked.

### Search terms

_No response_

goutamadwant:

---

Closes #8341

## Description

The Solidity test runner decoded selected build-info output bytes into one full string before parsing JSON. Very large outputs can exceed V8's maximum string size before JSON parsing starts.

This changes inline config and EIP-712 collection to parse build-info outputs through a streaming JSON parser from `hardhat-utils`, and awaits those collectors from the Solidity test task action. The existing byte-level skip checks remain in place.

Breaking change: No.

## Tests

- `pnpm lint:file packages/hardhat-utils/src/fs.ts packages/hardhat/src/internal/builtin-plugins/solidity-test/inline-config/index.ts packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/index.ts packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts packages/hardhat/test/internal/builtin-plugins/solidity-test/inline-config/index.ts packages/hardhat/test/internal/builtin-plugins/solidity-test/eip712/index.ts`
- `pnpm test:file packages/hardhat/test/internal/builtin-plugins/solidity-test/inline-config/index.ts packages/hardhat/test/internal/builtin-plugins/solidity-test/eip712/index.ts`
- `pnpm test:file packages/hardhat-utils/test/fs.ts`

### DIFF
--- a/packages/hardhat-utils/src/fs.ts
+++ b/packages/hardhat-utils/src/fs.ts
@@ -1,11 +1,9 @@
-import type * as StreamParserJson from "@streamparser/json-node";
 import type * as JsonStreamStringify from "json-stream-stringify";
 import type { FileHandle } from "node:fs/promises";
 
 import fsPromises from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
-import { Readable } from "node:stream";
 import { pipeline } from "node:stream/promises";
 
 import { ensureError, ensureNodeErrnoExceptionError } from "./error.js";
@@ -19,14 +17,11 @@ import {
   IsDirectoryError,
   DirectoryNotEmptyError,
 } from "./errors/fs.js";
+import { parseJsonStream } from "./internal/bytes.js";
 import {
   collectAllDirectoriesMatching,
   collectAllFilesMatching,
 } from "./internal/fs.js";
-
-// We don't load @streamparser/json-node on startup because it's only
-// used for parsing very large JSON payloads.
-let streamParserJson: typeof StreamParserJson | undefined;
 
 // We don't load json-stream-stringify on startup because it's only
 // used by writeJsonFileAsStream for very large JSON objects.
@@ -406,52 +401,6 @@ export async function readJsonFile<T>(absolutePathToFile: string): Promise<T> {
     ensureError(e);
     throw new InvalidFileFormatError(absolutePathToFile, e);
   }
-}
-
-async function parseJsonStream<T>(stream: Readable): Promise<T> {
-  if (streamParserJson === undefined) {
-    streamParserJson = await import("@streamparser/json-node");
-  }
-
-  // NOTE: We set a separator to disable self-closing to be able to use the parser
-  // in the stream.pipeline context; see https://github.com/juanjoDiaz/streamparser-json/issues/47
-  const jsonParser = new streamParserJson.JSONParser({
-    separator: "",
-  });
-
-  const result: T | undefined = await pipeline(
-    stream,
-    jsonParser,
-    async (
-      elements: AsyncIterable<StreamParserJson.ParsedElementInfo.ParsedElementInfo>,
-    ): Promise<any | undefined> => {
-      let value:
-        | StreamParserJson.JsonTypes.JsonPrimitive
-        | StreamParserJson.JsonTypes.JsonStruct
-        | undefined;
-      for await (const element of elements) {
-        value = element.value;
-      }
-      return value;
-    },
-  );
-
-  if (result === undefined) {
-    throw new Error("No data");
-  }
-
-  return result;
-}
-
-/**
- * Parses JSON bytes as a stream. This function should be used when
- * parsing very large JSON payloads.
- *
- * @param bytes The UTF-8 encoded JSON bytes.
- * @returns The parsed JSON object.
- */
-export async function parseJsonBytesAsStream<T>(bytes: Uint8Array): Promise<T> {
-  return await parseJsonStream<T>(Readable.from([bytes]));
 }
 
 /**
--- a/packages/hardhat-utils/src/fs.ts
+++ b/packages/hardhat-utils/src/fs.ts
@@ -25,7 +25,7 @@ import {
 } from "./internal/fs.js";
 
 // We don't load @streamparser/json-node on startup because it's only
-// used by readJsonFileAsStream for very large JSON files.
+// used for parsing very large JSON payloads.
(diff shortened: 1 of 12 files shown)


packages/hardhat-utils/src/fs.ts
```diff
@@ -407,6 +408,49 @@ export async function readJsonFile<T>(absolutePathToFile: string): Promise<T> {
...
 
+/**
+ * Parses JSON bytes as a stream. This avoids materializing the full input as a
+ * single string, which can exceed V8's maximum string size for very large JSON
+ * payloads.
+ *
+ * @param bytes The UTF-8 encoded JSON bytes.
+ * @returns The parsed JSON object.
+ */
+export async function parseJsonBytesAsStream<T>(bytes: Uint8Array): Promise<T> {
```
schaable: This function is exactly `readJsonFileAsStream` without the file handling. Can we reuse it there?

packages/hardhat-utils/src/fs.ts
```diff
@@ -407,6 +408,49 @@ export async function readJsonFile<T>(absolutePathToFile: string): Promise<T> {
   }
 }
 
+/**
+ * Parses JSON bytes as a stream. This avoids materializing the full input as a
+ * single string, which can exceed V8's maximum string size for very large JSON
+ * payloads.
```
schaable:
```suggestion
 * Parses JSON bytes as a stream. This function should be used when
 * parsing very large JSON payloads.
```

schaable (approved):
I left two minor suggestions, but it looks good overall. Thanks @goutamadwant!

Edit: There are also some conflicts. Could you take a look?

goutamadwant:
Thanks @schaable i rebased onto current main and resolved the EIP-712 test conflict also the byte and file APIs now reuse the same stream parser and I applied the suggested documentation wording and I also added direct utility coverage and a patch changeset. let me know if this looks good now!

schaable:
I made some changes to the original PR:

* Reworded the changeset message to make it more user-facing.
* Moved the utility to `bytes`, since `fs` is for filesystem-related utilities and this parses in-memory JSON.
* I then noticed that `parseJsonStream` is much slower than `JSON.parse`, so I refactored the code to use it only as a fallback when `bytesToUtf8String` fails on very large payloads.
* I also noticed that `parseJsonStream` can throw a raw error that callers weren’t handling, so I wrapped it in a `HardhatError` at the call sites.

packages/hardhat-utils/test/bytes.ts
```diff
@@ -218,16 +218,44 @@ describe("bytes", () => {
...
+
+    it("Should fall back to stream parsing when the payload is too large for a single string", async (t) => {
+      const expectedObject = { a: 1, b: 2 };
+      const bytes = new TextEncoder().encode(JSON.stringify(expectedObject));
+
+      // Simulate a payload too large to be decoded into a single string:
+      // `TextDecoder.decode` throws only in that case, so this exercises the
+      // stream-parsing fallback without allocating a multi-gigabyte buffer.
+      const originalDecode = TextDecoder.prototype.decode;
+      t.mock.method(
```
kanej, line 241: Oh the depths of trickery to which we now descend.

kanej (approved):

kanej:

PR #8416, opened 2026-06-29
docs: document type extension import conventions

### LINKED ISSUE #8320: Document the new conventions around Type Extension imports for plugin authors
Following on from https://github.com/NomicFoundation/hardhat/issues/8164.

BROCCOLO1D:
## Summary
- Documents the type-only `export type * from "./type-extensions.js"` pattern for plugin entry-points.
- Captures the constraints that keep `type-extensions.ts` type-only and safe to re-export.

## Why
Plugin authors need the updated convention for making type extensions visible to TypeScript consumers without emitting unnecessary runtime imports.

## Changes
- Expands `docs/engineering-guidelines.md` under the plugin entry-point import rules.

## Validation
- `pnpm spellcheck:file docs/engineering-guidelines.md`

## Scope
Docs-only change; no runtime behavior changes.

Closes #8320

### DIFF
--- a/docs/engineering-guidelines.md
+++ b/docs/engineering-guidelines.md
@@ -26,7 +26,7 @@ This keeps the declaration file reachable to TypeScript consumers without adding
 
 1. `type-extensions.ts` must contain only type-level module augmentation code and no runtime side effects.
 2. `type-extensions.ts` must not export named types. If it needs to be treated as a module, rely on `"moduleDetection": "force"` or use `export {};`.
-3. If the plugin depends on another plugin's type extensions, re-export them with `export type * from "dependency/type-extensions";` from your own `type-extensions.ts`.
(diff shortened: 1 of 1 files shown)


docs/engineering-guidelines.md
```diff
@@ -16,6 +16,19 @@ The entry-point of the plugins are meant to only export a description of the plu
...
+
+```ts
+export type * from "./type-extensions.js";
+```
+
+This keeps the declaration file reachable to TypeScript consumers without adding a runtime import to the compiled `index.js`. To make this safe and predictable:
+
+1. `type-extensions.ts` must contain only type-level module augmentation code and no runtime side effects.
+2. `type-extensions.ts` must not export named types. If it needs to be treated as a module, rely on `"moduleDetection": "force"` or use `export {};`.
+3. If the plugin depends on another plugin's type extensions, re-export them with `export type * from "dependency/type-extensions";` from your own `type-extensions.ts`.
```
kanej:
```suggestion
3. Type-extensions don't need to import the modules they are augmenting. This was an incorrect way to force the .ts file to be treated as a module. Use `export {};`, or better yet ["moduleDetection": "force"](https://www.typescriptlang.org/tsconfig/#moduleDetection).
```

kanej (approved):

PR #8417, opened 2026-06-29
EIP-712 collector throws HHE818 for a same-named struct in a source the user didn't include

### LINKED ISSUE #8344: EIP-712 collector throws `HHE818` for a same-named struct in a source the user didn't include

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/8344

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
@@ -2,6 +2,12 @@ import type { CollectedStruct } from "./ast-walker.js";
 
 import { HardhatError } from "@nomicfoundation/hardhat-errors";
 
+export const SELECTED_CONFLICT_REMEDIATION =
+  "Rename one of the structs, or scope your `test.solidity.eip712Types.include` / `exclude` globs in `hardhat.config.ts` so that only one of them is selected.";
+
+export const DEPENDENCY_CONFLICT_REMEDIATION =
+  "Rename one of the structs. Narrowing `test.solidity.eip712Types.include` / `exclude` won't resolve this: the conflicting definition is pulled in as a dependency of a selected struct, not selected directly.";
+
 /**
  * Produces the flat list of canonical EIP-712 type strings expected by EDR.
  *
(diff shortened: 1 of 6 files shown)


packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
```diff
@@ -2,6 +2,12 @@ import type { CollectedStruct } from "./ast-walker.js";
 
 import { HardhatError } from "@nomicfoundation/hardhat-errors";
 
+export const SELECTED_CONFLICT_REMEDIATION =
+  "Rename one of the structs, or scope your `test.solidity.eip712Types.include` / `exclude` globs in `hardhat.config.ts` so that only one of them is selected.";
+
+export const DEPENDENCY_CONFLICT_REMEDIATION =
+  "Rename one of the structs so the name has a single definition. The clashing name is pulled in as a dependency of a selected struct, so if you don't need that selected struct, deselecting it with your `test.solidity.eip712Types.include` / `exclude` globs in `hardhat.config.ts` also resolves the conflict.";
```
schaable, line 9: These could be collapsed into a single message for simplicity:

`Rename one of the structs, or scope test.solidity.eip712Types.include / exclude so only one definition is selected (including via dependencies).`

That way, we don’t need the extra argument in the descriptor, and the implementation stays more straightforward. The current approach is fine too, so I’ll leave it up to you.

schaable (approved):
LGTM!

packages/hardhat/src/internal/builtin-plugins/solidity-test/eip712/canonicalize.ts
ChristopherDedominici: I think I'm gonna keep the two messages, the dependency case kinda needs to nudge the user toward a struct that isn't even named in the error, and squishing both into one message ends up a bit too vague to do that

kanej:

PR #8422, opened 2026-07-04
HH2: Port hardhat-network provider to the current EDR API and restore trace events

nebasuke:
Draft in preparation of N-API v3 and re-enabling traces for HH2.

# Claude summary

## What

Ports the `hardhat-network` provider glue to the current EDR API and restores the vm trace-event bridge that HH2 plugins (e.g. `solidity-coverage`, `smock`) rely on:

- **Provider glue port**: the `fork`/`initialDate`/top-level `blockGasLimit` provider config is replaced by the `network` union and `mining.blockGasLimit`/`defaultTransactionGasLimit`; local-network genesis block gas limit and time are threaded through `hardhat_reset`; `convertToEdr` drops the converters for removed EDR surface.
- **Trace-event bridge restored**: EDR re-adds raw traces as `Response.traces()` (a method returning napi-v3-shaped items). The previously commented-out bridge is re-enabled: iterate `traces()` and emit `step`/`beforeMessage`/`afterMessage` on the minimal vm, with marshalling updated to the new shapes (`step.opcode.name`, `step.memory`, `tracingMessageResult.execResult`).
- **Call traces enabled**: EDR's `includeCallTraces` defaults to `None`, in which case `traces()` returns no arenas and the bridge silently emits nothing (the solidity-coverage 0%-coverage failure mode). The provider now requests `IncludeTraces.All` at creation.

## Status / blocked on

Draft until an EDR release ships the `Response.traces()` work (NomicFoundation/edr#1301); the `@nomicfoundation/edr` version bump is intentionally not part of this branch yet.

## Testing

- Validated against a local build of the EDR branch: full `hardhat-core` test run for the glue port, and the bridge verified end-to-end (step/beforeMessage/afterMessage counts) by a dedicated spec in EDR's `hardhat-tests` suite.
- The provider source here is kept byte-identical to the `patches/hardhat@2.28.6.patch` EDR applies in its `hardhat-tests` CI, so this branch is exercised continuously on the EDR side.

### DIFF
--- a/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
+++ b/packages/hardhat-core/src/internal/hardhat-network/provider/provider.ts
@@ -67,7 +67,6 @@ import {
   MempoolOrder,
 } from "./node-types";
 import {
-  edrRpcDebugTraceToHardhat,
   edrTracingMessageResultToMinimalEVMResult,
   edrTracingMessageToMinimalMessage,
   edrTracingStepToMinimalInterpreterStep,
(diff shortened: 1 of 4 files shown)


Wodann:
Duplicate of https://github.com/NomicFoundation/hardhat/pull/7985

PR #8429, opened 2026-07-07
Add detectAgentEnvironment util to hardhat-utils

marianfe:
Adds a `detectAgentEnvironment()` util to `@nomicfoundation/hardhat-utils` (exported as `@nomicfoundation/hardhat-utils/environment`) that detects whether the current process is running inside a coding agent environment, based on the environment variables those tools set in the shells they spawn.

```ts
const { isAgent, agents } = detectAgentEnvironment();
// e.g. { isAgent: true, agents: ["claude-code", "codex"] }
```

- **Supported agents:** Currently Claude Code and Codex. Agents advertising themselves only via a generic variable are reported as `"unknown"`. Others (Cursor, Gemini CLI, …) can be added later.
- `agents` reports **all** detected agents: sessions can be nested.
- `isInteractive()` (stdin+stdout are TTYs) is exposed as a separate util: orthogonal to agent detection, but a useful hint of whether the process is running in an automated environment.

**Note:** this detection is cooperative — the markers can be trivially stripped or spoofed, and a human typing commands into an agent's embedded terminal is indistinguishable from the agent itself. It's intended for optimizations such as output tailoring and telemetry only, not for gating behavior.

No consumer is wired up in this PR; that will be added in follow-up PRs.

Verified end-to-end against the real binaries: Claude Code shell, `codex exec` (nested and standalone), scrubbed-env control, and a PTY-allocated run for the interactive case.

### DIFF
--- a/packages/hardhat-utils/package.json
+++ b/packages/hardhat-utils/package.json
@@ -21,6 +21,7 @@
     "./date": "./dist/src/date.js",
     "./debug": "./dist/src/debug.js",
     "./env": "./dist/src/env.js",
+    "./environment": "./dist/src/environment.js",
     "./error": "./dist/src/error.js",
     "./eth": "./dist/src/eth.js",
     "./fast-semver": "./dist/src/fast-semver.js",
--- a/.changeset/detect-agent-environment.md
+++ b/.changeset/detect-agent-environment.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-utils": patch
+---
+
+Add `detectAgentEnvironment` and `isInteractive` utils
--- a/packages/hardhat-utils/src/environment.ts
+++ b/packages/hardhat-utils/src/environment.ts
@@ -1,0 +1,76 @@
+/**
+ * The result of detecting whether the current process is running inside an
+ * AI coding agent environment.
+ */
+export interface AgentEnvironment {
+  /**
+   * True if at least one agent environment was detected.
+   */
+  isAgent: boolean;
+
+  /**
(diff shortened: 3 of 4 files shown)


packages/hardhat-utils/src/agent.ts
```diff
@@ -0,0 +1,75 @@
...
+   */
+  agents: string[];
+
+  /**
+   * True if both stdin and stdout are TTYs.
+   *
+   * This is orthogonal to agent detection, but can be a useful hint of
+   * whether the process is running in an automated environment.
+   */
+  isInteractive: boolean;
```
schaable: I feel this should be a separate function, as it could be useful on it's own and isn't related to agents.

packages/hardhat-utils/package.json
```diff
@@ -12,6 +12,7 @@
   "license": "MIT",
   "type": "module",
   "exports": {
+    "./agent": "./dist/src/agent.js",
```
schaable: We could merge the `agents` and `ci` exports and expose all environment-related functions through a single `environment` entry point. We don't need to do this now though

schaable (approved):
Pre-approving. I left two suggestions, but the exports refactor can be done later since it requires updating all consumers of `isCI`.

packages/hardhat-utils/package.json
marianfe: agree. will leave this one for another time

packages/hardhat-utils/src/agent.ts
marianfe: agree, will refactor

PR #8438, opened 2026-07-13
Fix network options at cli

### LINKED ISSUE #8436: hardhat node ignores all arguments (--chain-id, --fork, etc)
### Version of Hardhat

3.0.10

### What happened?

Running hardhat node and trying to override things like chain-id, fork, etc ignores the command line arguments.
It just tries to use the configured 'node' network. 
While the hardhat node help still lists the options, and they are validated as well, the lines thath actually applied the custom options on network creation were removed from task-action.ts

Thanks!

I've traced to the commit that broke this:  Start looking at line 111 on task-action.ts of this commit https://github.com/NomicFoundation/hardhat/commit/c3e922bed58d5623672f888537321c57b3305ec0

### Search terms

command line arguments broken in hardhat node

kanej:
The `npx hardhat node` command was ignoring a subset of cli options, including `--fork`. This fix now passes all the cli options to the initialization of the EDR network within the standalone network node.

The dropped cli options that are now fixed:

* `--chain-id`
* `--fork`
* `--fork-block-number`

Fixes #8436.

## Reviewing

> [!TIP]
> Review a commit at a time

This PR, pulls out the mapping code for the EDR connection setting into a helper function, adds tests across the helper function, then re-adds the missing overridden connection settings fixing the bug.

## Manual testing

To test this PR locally you can use the `./packages/example-project`.

To test that cli commands are coming through run in one terminal:

```shell
pnpm hardhat node --chain-id 22
```

In another run the script:

```ts
import hre from "hardhat";

const connection = await hre.network.create("localhost");

// Get the chain id
const chainId = await connection.provider.request({ method: "eth_chainId" });

console.log("Chain ID:", Number(chainId));
```

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/node/utils/resolve-node-connection-params.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/node/utils/resolve-node-connection-params.ts
@@ -73,5 +73,7 @@ export function resolveNodeConnectionParams(
     );
   }
 
+  connectionParams.override = networkConfigOverride;
+
   return connectionParams;
 }
--- a/packages/hardhat/src/internal/builtin-plugins/node/utils/resolve-node-connection-params.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/node/utils/resolve-node-connection-params.ts
@@ -1,0 +1,77 @@
+import type { EdrNetworkConfigOverride } from "../../../../types/config.js";
+import type { ChainType } from "../../../../types/network.js";
+
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
+
+import { isSupportedChainType } from "../../../edr/chain-type.js";
+
+export interface NodeConnectionParams {
+  network: string;
+  chainType?: ChainType;
+  override?: EdrNetworkConfigOverride;
+}
+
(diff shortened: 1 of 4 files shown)


packages/hardhat/src/internal/builtin-plugins/node/utils/resolve-node-connection-params.ts
```diff
@@ -0,0 +1,79 @@
+import type { EdrNetworkConfigOverride } from "../../../../types/config.js";
+import type { ChainType } from "../../../../types/network.js";
+
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
+
+import { isSupportedChainType } from "../../../edr/chain-type.js";
+
+export interface NodeConnectionParams {
```
schaable: nit: I'd call this `ResolvedNodeConnectionParams`

schaable (approved):
LGTM!

packages/hardhat/src/internal/builtin-plugins/node/utils/resolve-node-connection-params.ts
kanej: Agreed and applied.

kanej:

PR #8443, opened 2026-07-15
fix(network): reject malformed HD derivation paths in HD_PATH_REGEX

Kropiunig:
## Problem

`HD_PATH_REGEX` in `packages/hardhat/src/internal/builtin-plugins/network-manager/accounts/derive-private-keys.ts` uses a capturing group that begins with an optional-colon literal, `(:?`, instead of the intended non-capturing group `(?:`:

```ts
const HD_PATH_REGEX = /^m(:?\/\d+'?)+\/?$/;
```

Because `:?` matches an optional stray colon at the start of each segment, malformed BIP-32 paths are silently accepted. For example, both of these pass the validator:

- `m:/44'/60'/0'/0`
- `m/44':/60'/0'/0`

## Failing scenario

In `derivePrivateKeys()`, a path that fails `HD_PATH_REGEX.test(hdpath)` is supposed to throw the friendly `HardhatError` `CORE.NETWORK.INVALID_HD_PATH`. Because of the typo, a colon-containing path passes this check and is forwarded to `HDKey.derive()` (ethereum-cryptography / scure), which throws a low-level error such as `Invalid child index: m:` instead of the intended `INVALID_HD_PATH` error. This degrades the error surface for any HD-account configuration (the default account config and any `{ mnemonic, path }` network config) whenever the path is slightly malformed.

## Fix

Change `(:?` to `(?:`, making the group non-capturing as intended:

```ts
const HD_PATH_REGEX = /^m(?:\/\d+'?)+\/?$/;
```

Every valid path (`m/44'/60'/0'/0`, with or without a trailing slash, hardened segments, etc.) matches identically under both forms, so the change rejects only malformed input and preserves all valid behavior. The group was only used inside `.test()`, so making it non-capturing has no other effect.

## Test

Extends the existing `should throw if the path is invalid` test to also assert that colon-containing paths (`m:/44'/60'/0'/0` and `m/44':/60'/0'/0`) are rejected with `INVALID_HD_PATH`. These cases fail before the fix (the paths are accepted and instead surface a low-level derivation error) and pass after it; all pre-existing valid- and invalid-path assertions remain green.

### DIFF
--- a/.changeset/hd-path-ethers-reject.md
+++ b/.changeset/hd-path-ethers-reject.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-ethers": patch
+---
+
+Fixed the HD wallet derivation-path validator so malformed paths containing stray colons (e.g. `m:/44'/60'/0'/0`) are rejected with the `INVALID_HD_PATH` error.
--- a/.changeset/tidy-otters-reject.md
+++ b/.changeset/tidy-otters-reject.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed the HD wallet derivation-path validator so malformed paths containing stray colons (e.g. `m:/44'/60'/0'/0`) are rejected with the `INVALID_HD_PATH` error instead of being forwarded to the key-derivation library, which threw a low-level error.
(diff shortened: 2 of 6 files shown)


kanej (approved):
Thanks @Kropiunig, I applied the same fix to `hardhat-ethers`.

kanej:

PR #8446, opened 2026-07-17
ci: add memory stats and filters for regression benchmark

popescuoctavian:

---

This PR adds peak RSS memory measurements to the regression benchmark stats. 

It also adds the ability to filter which commands/steps and which scenarios run as part of the regression benchmark via glob patterns. Since some steps depend on previous steps being executed (i.e. `test solidity` runs with `--no-compile` and requires `cold compile` to have run beforehand), it adds a dependency list which allows the workflow to only run the required steps. Stats are only emitted for  filtered commands/steps. This is useful in EDR to check solidity test performance (see https://github.com/NomicFoundation/edr/pull/1563).

### DIFF
--- a/scripts/benchmark/helpers/memory.test.ts
+++ b/scripts/benchmark/helpers/memory.test.ts
@@ -1,0 +1,85 @@
+import { describe, it } from "node:test";
+import assert from "node:assert/strict";
+import { mkdtempSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import path from "node:path";
+
+import {
+  formatTimeCommand,
+  gnuTimeAvailable,
+  readPeakRssMb,
+  wrapWithTime,
+} from "./memory.ts";
+
+const tmp = mkdtempSync(path.join(tmpdir(), "mem-test-"));
+const memFileWith = (name: string, contents: string) => {
+  const p = path.join(tmp, name);
+  writeFileSync(p, contents);
+  return p;
+};
+
+describe("formatTimeCommand", () => {
+  it("wraps a lone program without an inner shell", () => {
+    assert.equal(
+      formatTimeCommand(
+        "hyperfine --runs 3 'npx hardhat compile'",
+        "/t/m",
+        false,
+      ),
+      "/usr/bin/time -o '/t/m' -f %M hyperfine --runs 3 'npx hardhat compile'",
+    );
+  });
+
+  it("wraps a shell command under sh -c so operators are covered", () => {
+    assert.equal(
+      formatTimeCommand("printf x >> f && npx hardhat compile", "/t/m", true),
+      `/usr/bin/time -o '/t/m' -f %M sh -c 'printf x >> f && npx hardhat compile'`,
+    );
+  });
+
+  it("escapes single quotes in the wrapped command", () => {
+    assert.equal(
+      formatTimeCommand("echo 'hi'", "/t/m", true),
+      `/usr/bin/time -o '/t/m' -f %M sh -c 'echo '\\''hi'\\'''`,
+    );
+  });
+});
+
+describe("wrapWithTime", () => {
+  it("is a no-op when GNU time is unavailable (returns the command unchanged)", () => {
+    // This sandbox has no /usr/bin/time; assert the graceful fallback.
+    if (!gnuTimeAvailable()) {
+      assert.equal(
+        wrapWithTime("npx hardhat compile", "/t/m", false),
+        "npx hardhat compile",
+      );
+    } else {
+      assert.match(
+        wrapWithTime("npx hardhat compile", "/t/m", false),
+        /^\/usr\/bin\/time /,
+      );
+    }
+  });
+});
+
+describe("readPeakRssMb", () => {
+  it("parses KB and rounds to MB", () => {
+    assert.equal(readPeakRssMb(memFileWith("a", "1268340\n")), 1239);
+  });
+
+  it("ignores a trailing 'Command exited' line (takes the first integer)", () => {
+    const p = memFileWith(
+      "b",
+      "524288\nCommand exited with non-zero status 1\n",
+    );
+    assert.equal(readPeakRssMb(p), 512);
+  });
+
+  it("returns undefined for a missing file", () => {
+    assert.equal(readPeakRssMb(path.join(tmp, "does-not-exist")), undefined);
+  });
+
+  it("returns undefined for non-numeric content", () => {
+    assert.equal(readPeakRssMb(memFileWith("c", "no digits here")), undefined);
+  });
+});
--- a/scripts/benchmark/helpers/memory.test.ts
+++ b/scripts/benchmark/helpers/memory.test.ts
@@ -75,11 +75,11 @@ describe("readPeakRssMb", () => {
     assert.equal(readPeakRssMb(p), 512);
   });
 
-  it("returns undefined for a missing file", () => {
-    assert.equal(readPeakRssMb(path.join(tmp, "does-not-exist")), undefined);
+  it("throws for a missing file", () => {
+    assert.throws(() => readPeakRssMb(path.join(tmp, "does-not-exist")));
   });
 
-  it("returns undefined for non-numeric content", () => {
-    assert.equal(readPeakRssMb(memFileWith("c", "no digits here")), undefined);
+  it("throws for non-numeric content", () => {
+    assert.throws(() => readPeakRssMb(memFileWith("c", "no digits here")));
   });
 });
--- a/scripts/benchmark/helpers/plan.test.ts
+++ b/scripts/benchmark/helpers/plan.test.ts
@@ -43,14 +43,15 @@ function fixture(): Record<string, CommandConfig> {
   };
 }
 
-// Compact view: what each planned command runs and reports.
+// Compact view: what each planned command runs and reports. Step sequences
+// carry `run` (ordered steps to execute) and an `emit` list of reported steps;
+// single commands carry only a boolean `emit`.
 function summarize(plan: ReturnType<typeof planCommands>) {
-  return plan.map((p) => ({
-    name: p.name,
-    runStepNames: p.runStepNames,
-    emit: p.emit,
-    emitSteps: p.emitSteps,
-  }));
+  return plan.map((p) =>
+    "run" in p
+      ? { name: p.name, run: p.run, emit: p.emit }
+      : { name: p.name, emit: p.emit },
+  );
 }
 
 describe("globToRegExp", () => {
@@ -81,26 +82,16 @@ describe("planCommands", () => {
     assert.deepEqual(plan, [
       {
         name: "compile sequence",
-        runStepNames: [
-          "reset files & cache",
-          "cold compile",
-          "edit min",
-          "edit max",
-        ],
-        emit: false,
-        emitSteps: ["cold compile", "edit min", "edit max"],
+        run: ["reset files & cache", "cold compile", "edit min", "edit max"],
+        emit: ["cold compile", "edit min", "edit max"],
       },
       {
         name: "warm compile",
-        runStepNames: undefined,
         emit: true,
-        emitSteps: [],
       },
       {
         name: "test solidity",
-        runStepNames: undefined,
         emit: true,
-        emitSteps: [],
       },
     ]);
   });
@@ -111,15 +102,12 @@ describe("planCommands", () => {
       {
         name: "compile sequence",
         // reset + cold compile pulled in via dependsOn; edit steps skipped.
-        runStepNames: ["reset files & cache", "cold compile"],
-        emit: false,
-        emitSteps: [], // prerequisites only — not reported
+        run: ["reset files & cache", "cold compile"],
+        emit: [], // prerequisites only — not reported
       },
       {
         name: "test solidity",
-        runStepNames: undefined,
         emit: true,
-        emitSteps: [],
       },
     ]);
   });
@@ -129,15 +117,12 @@ describe("planCommands", () => {
     assert.deepEqual(plan, [
       {
         name: "compile sequence",
-        runStepNames: ["reset files & cache", "cold compile"],
-        emit: false,
-        emitSteps: [],
+        run: ["reset files & cache", "cold compile"],
+        emit: [],
       },
       {
         name: "warm compile",
-        runStepNames: undefined,
         emit: true,
-        emitSteps: [],
       },
       // "test solidity" is neither selected nor a prerequisite → dropped.
     ]);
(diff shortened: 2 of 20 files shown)


scripts/benchmark/regression.ts
```diff
@@ -266,6 +317,7 @@ function resolveArgs(argv: string[]): RegressionArgs | undefined {
 function collectScenarios(args: RegressionArgs): ScenarioEntry[] {
   const entries: ScenarioEntry[] = [];
   const invalid: string[] = [];
+  const scenarioRes = compilePatterns(args.scenarios);
```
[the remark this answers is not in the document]
popescuoctavian: Not a concern: scenarios="" is bad usage of the filter.

scripts/benchmark/helpers/plan.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  for (const e of entries) {
+    for (const dep of e.dependsOn ?? []) {
+      if (!byName.has(dep)) {
+        throw new Error(
+          `Entry "${e.name}" dependsOn "${dep}", which is not a command or step in this scenario`,
+        );
+      }
+    }
+  }
```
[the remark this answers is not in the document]
popescuoctavian: Addressed by 952b5cab3c5510337bff765334ab377f0317045a.

scripts/benchmark/helpers/plan.ts
```diff
@@ -0,0 +1,222 @@
...
+      runSet.add(e.name);
+      for (const other of entries) {
+        if (other.index < e.index) {
+          runSet.add(other.name);
+        }
+      }
+    }
+  }
+
+  const plan: PlannedCommand[] = [];
```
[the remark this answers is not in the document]
popescuoctavian: The conservative run of each prior step was removed in 952b5cab3c5510337bff765334ab377f0317045a. The responsibility is on the scenario author to handle dependencies.

scripts/benchmark/helpers/memory.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    assert.equal(
+      formatTimeCommand("printf x >> f && npx hardhat compile", "/t/m", true),
+      `/usr/bin/time -o '/t/m' -f %M sh -c 'printf x >> f && npx hardhat compile'`,
+    );
+  });
+
+  it("escapes single quotes in the wrapped command", () => {
+    assert.equal(
+      formatTimeCommand("echo 'hi'", "/t/m", true),
+      `/usr/bin/time -o '/t/m' -f %M sh -c 'echo '\\''hi'\\'''`,
```
Wodann, line 43: Question: The generated output is confusing me. What do the extra `'`  quotes around `\\` do in:

```
-c 'echo '\\''hi'\\'''
```

scripts/benchmark/helpers/memory.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    assert.equal(readPeakRssMb(p), 512);
+  });
+
+  it("returns undefined for a missing file", () => {
+    assert.equal(readPeakRssMb(path.join(tmp, "does-not-exist")), undefined);
+  });
+
+  it("returns undefined for non-numeric content", () => {
+    assert.equal(readPeakRssMb(memFileWith("c", "no digits here")), undefined);
+  });
```
Wodann: nit: should we throw an error instead of failing silently?

scripts/benchmark/helpers/plan.ts
```diff
@@ -0,0 +1,220 @@
...
+ * steps to report; steps run but not in `emitSteps` are prerequisites only. For
+ * single commands, `emit` says whether to report it (it may run purely as a
+ * prerequisite of a later command).
+ */
+export interface PlannedCommand {
+  name: string;
+  cfg: CommandConfig;
+  runStepNames: string[] | undefined;
+  emit: boolean;
+  emitSteps: string[];
```
Wodann: nit: make this mutually exclusive in the type system; i.e. either: `CommandData { emit: boolean }` or `StepData { run: string[], emit: string[] }`

scripts/benchmark/helpers/plan.test.ts
```diff
@@ -0,0 +1,264 @@
...
+      },
+    ]);
+  });
+
+  it("glob '*compile*' selects compile-named entries across levels", () => {
+    const plan = summarize(planCommands(fixture(), ["*compile*"]));
+    // "cold compile" (a step) and "warm compile" (a command) match; the
+    // "edit min"/"edit max" steps and "test solidity" do not.
+    assert.deepEqual(plan[0].emitSteps, ["cold compile"]);
+    assert.equal(plan.find((p) => p.name === "warm compile")?.emit, true);
```
Wodann: Suggestion: use `!` instead of `?` if it's expected to exist

```suggestion
    assert.equal(plan.find((p) => p.name === "warm compile")!.emit, true);
```

scripts/benchmark/helpers/plan.ts
```diff
@@ -0,0 +1,220 @@
...
+ * selected). There are no implicit prerequisites: an entry with no `dependsOn`
+ * runs in isolation. Declaring what an entry depends on is the scenario author's
+ * responsibility.
+ *
+ * Entries run in declared order; only entries in the resulting run set run, and
+ * only selected entries are reported. Returns `[]` when nothing is selected
+ * (the scenario is skipped before any expensive init).
+ *
+ * Throws if a `dependsOn` names a nonexistent entry, or one declared after the
+ * dependent (dependencies must precede their dependents).
```
Wodann, line 146: Suggestion: Add a line that says: "This also prevents cyclic dependencies."

At first I thought the algorithm didn't prevent cyclic dependencies. It wasn't until I read the test for self-referential types, that I corrected my understanding.

scripts/benchmark/regression.ts
```diff
@@ -474,11 +584,23 @@ function runStepsPhase(
 
       const elapsed = (performance.now() - start) / 1000;
       samples.get(stepName)?.push(elapsed);
+
+      if (emit.has(stepName)) {
+        const rss = readPeakRssMb(memFile(stepName));
+        if (rss !== undefined) {
+          peakRssMb.set(stepName, Math.max(peakRssMb.get(stepName) ?? 0, rss));
+        }
```
Wodann, line 622: suggestion: consider logging a warning that the `rss` could not be read from the output file.

scripts/benchmark/regression.ts
```diff
@@ -474,11 +584,23 @@ function runStepsPhase(
 
       const elapsed = (performance.now() - start) / 1000;
       samples.get(stepName)?.push(elapsed);
+
+      if (emit.has(stepName)) {
+        const rss = readPeakRssMb(memFile(stepName));
+        if (rss !== undefined) {
+          peakRssMb.set(stepName, Math.max(peakRssMb.get(stepName) ?? 0, rss));
```
Wodann: issue: we overwrite prior reported values for `peakRssMb.get(stepName)`. Can we instead `.push` them like we do for `samples` and apply some statistics on the output, adding those to the `extra` field?

scripts/end-to-end/schema/scenario.schema.json
```diff
@@ -102,6 +102,12 @@
                     "type": "string",
                     "minLength": 1,
                     "description": "Shell command to benchmark"
+                  },
+                  "dependsOn": {
```
Wodann, line 106: praise: Nice idea to work around prerequisites!

scripts/benchmark/regression.ts
```diff
@@ -383,34 +445,57 @@ async function runScenario(
...
           loaded.workingDir,
           loaded.definition.env,
-          name,
-          cfg,
+          planned.name,
+          planned.cfg,
+          // runStepNames is always defined for step sequences.
+          new Set(planned.runStepNames ?? Object.keys(planned.cfg.steps)),
+          new Set(planned.emitSteps),
+          scenarioTmpDir,
```
Wodann, line 467: suggestion: I think we are leaving some performance on the table still; although this might be trickier to implement.

I'm thinking of the following scenario:

- `--benchmarks "test*" to signal that we want to run all tests (this is applicable for EDR)
- "test solidity" `dependsOn` `cold compile`

We get:
```
run: ["test solidity", { name: "compile steps", steps: ["cold compile", "reset"] }]
emit: ["test solidity"]
```

Since we're re-running `["cold compile", "reset"]` we are not just re-running them once which is what we really want, but we're re-running them as many times as they are configured to run.

If our emit contained an entry of "compile steps", that would make sense, since we would need to run those steps every time before we reach the specific step that we want to emit. However, if it entails a command like "test solidity", it doesn't care whether it was run once or N times before the "test solidity" command is reached. It just needs to have run at least once.

Therefore, we might be able to further optimise this by using the list of `emit` to smartly optimise the number of samples/runs we do.

Wodann (changes requested):

scripts/benchmark/helpers/memory.test.ts
popescuoctavian: This ensures that single quotes are handled correctly if used in (future) step commands. In this case:

Step command: `echo 'hi'`
Shell wrap: `sh -c '...'`

- Shell wrapping without added quotes: `sh -c 'echo 'hi''` => the outer `sh -c` closes at the first `'` in `echo '`, leaving `hi'` and a dangling `'`.

- Wrapping enforced here: `sh -c 'echo '\''hi'\'''` => close quote, add escaped literal `'`, reopen quote. This is the canonical way to put single quotes inside a single quoted string in Bash (for example, see [this post](https://stackoverflow.com/questions/1250079/how-to-escape-single-quotes-within-single-quoted-strings)).

scripts/benchmark/helpers/memory.test.ts
popescuoctavian: Addressed by a9383104f79f2cdc79fd1467d146d96b709f5d4a.

scripts/benchmark/regression.ts
popescuoctavian: This now checks if GNU time is available and if so, throws an error. See: a9383104f79f2cdc79fd1467d146d96b709f5d4a.

scripts/benchmark/helpers/plan.ts
popescuoctavian: Addressed by 0b69ab36f587d061843bd359f7e4fc4d60b22941.

scripts/benchmark/helpers/plan.test.ts
popescuoctavian: Addressed by 0b69ab36f587d061843bd359f7e4fc4d60b22941.

scripts/benchmark/helpers/plan.ts
popescuoctavian: Good idea! Addressed by 89e18dbb05d61d2ec5e51bcf6758d7848da628d9.

scripts/benchmark/regression.ts
popescuoctavian: That's a good point, thanks for the suggestion! I think this is pretty straightforward for "test solidity", but I'm not sure what is expected for the other cases:

- "warm compile": here I believe it's fine to only execute on "cold compile"
- "edit & compile": I believe that these require reset to properly test the same behavior

We could also only handle test solidity in this way, since that provides significant value across both Hardhat and EDR CIs.

What do you think?

scripts/benchmark/regression.ts
popescuoctavian: Addressed by c0a8bc07fbd3246fb5a04df712077f0184267107 and a2fcea11a6f25c692c5d0b675dca8620ed00601a. The `value` field holds peak RSS value instead of mean, since I believe that makes more sense in this case. `range` is still empty since it would be misleading to print the stddev range alongside max values instead of mean.

Wodann: I like the approach of using max as value!

popescuoctavian:
@Wodann I've merged the main branch into this one to resolve conflicts, but Github doesn't seem to correctly show the changes in the commit diff ([db6b0a602e518822b29e24774d5b2e98628639f5](https://github.com/NomicFoundation/hardhat/pull/8446/changes/db6b0a602e518822b29e24774d5b2e98628639f5)). I'd recommend reviewing the full `regression.ts` diff (that was the file with the conflicts).

scripts/benchmark/regression.ts
```diff
@@ -656,8 +790,32 @@ function toEntry(
...
+    range: "",
+    extra: JSON.stringify({
+      times: rss.times,
+      min: rss.min,
+      max: rss.max,
+      median: rss.median,
+      mean: rss.mean,
+      stddev: rss.stddev,
     }),
   };
```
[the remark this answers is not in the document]
popescuoctavian: Range left empty on purpose to avoid confusion, since value holds `max`.

scripts/benchmark/regression.ts
Wodann: I think it can be generalised to "step" vs "command":

- For a step
  - If it `dependsOn` another step:
    - within its own command: run same number of times
      - the idea of steps is that they are run sequentially, so there is an explicit expectation that step 1 should be followed by step 2—every time you run it.
    - outside of its own command: run the external step once.
      - unlike the prior case, the step can only "observe" the external step running once (the last time) before it was reached; i.e. there would only be a difference between it running and not running.
  - If it `dependsOn` another command: run the external command once.
    -  Same motivation as the prior case
- For a command
  - If it depends on a step: run the step once.
    -  Same motivation as the prior case
  - If it depends on a command: run the command once.
    -  Same motivation as the prior case
 
 
Using this generalisation might help to come up with an algorithm to structure dependent runs. If this change makes sense, feel free to address it in a separate PR.

Wodann (approved):

Wodann:
Thanks for addressing the feedback!

scripts/benchmark/regression.ts
popescuoctavian: I agree, that's a good pattern. I'll address it in a separate PR. Thanks!

Wodann:
> @Wodann I've merged the main branch into this one to resolve conflicts, but Github doesn't seem to correctly show the changes in the commit diff ([db6b0a602e518822b29e24774d5b2e98628639f5](https://github.com/NomicFoundation/hardhat/pull/8446/changes/db6b0a602e518822b29e24774d5b2e98628639f5)). I'd recommend reviewing the full `regression.ts` diff (that was the file with the conflicts).

I had a look. I think we can just use the GNU time command to simplify the code. Otherwise, it looks good.

I did noticed that Bas' changes broke the [dashboard](https://nomic-foundation-automation.github.io/hardhat-benchmark-results/hardhat3/). The same thing might happen for your RSS graphs. I'll have a look at fixing the graphs once everything has been merged.

Issue #10, opened 2018-08-02
> contract binary not set. Can't deploy new instance.


shrugs:
## Issue

I'v encountered this twice when trying to produce bytecode for certain contracts.

```solidity
pragma solidity ^0.4.24;

import "zos-lib/contracts/upgradeability/AdminUpgradeabilityProxy.sol";


contract CustomAdminUpgradeabilityProxy is AdminUpgradeabilityProxy {
  bytes32 private constant ADMIN_SLOT = keccak256("org.zeppelinos.proxy.admin.custom");
}
```

this produces an artifact with valid abi, but an empty string `bytecode` property.

## Expected Behavior

bytecode should be produced

shrugs:
Actually, maybe this isn't expected behavior; truffle doesn't produce this bytecode either (although for some reason I thought it did). Any ideas?

alcuadrado:
Thanks for reporting this! Compiling that contract should produce a bytecode. I digged dipper into this issue and I think it is a bug in `solc`. I'm opening an issue in their repo, and leaving this one open for now.

shrugs:
@alcuadrado actually this was my fault! sorry I forgot to update this issue!

I hadn't explicitly inherited from the parent contract, so no bytecode was produced. it's expected behavior

alcuadrado:
Wait, no, this is the expected behavior 😅

It's easier to understand it with a minimal reproduction source code:

```solidity
pragma solidity 0.4.24;

contract Proxy {
    function _implementation() internal view returns (address);
}

contract UpgradeabilityProxy is Proxy {
    constructor(address _implementation) public {
    }

    function _implementation() internal view returns (address) {
        return address(0);
    }
}

contract CustomAdminUpgradeabilityProxy is UpgradeabilityProxy {
}
```

`UpgradeabilityProxy`'s constructor expects a param, and `CustomAdminUpgradeabilityProxy` doesn't override it, so the contract is abstract. This behavior is documented [here](https://solidity.readthedocs.io/en/latest/contracts.html#arguments-for-base-constructors).

alcuadrado:
Yes, I figured it out when creating that reproduction code.

PR #8449, opened 2026-07-20
fix(ignition): don't treat a replacement tx as confirmed one block early

Kropiunig:
## Problem

In `getNonceSyncMessages` (`@nomicfoundation/ignition-core`), when Ignition detects that a user has **replaced** one of its transactions, it must not proceed until the user's replacement transaction has `requiredConfirmations` confirmations. The guard for this uses the sender's *safe* transaction count (the tx count `requiredConfirmations` blocks in the past):

```ts
if (latestCount > nonce) {
  const hasEnoughConfirmations =
    safeConfirmationsCount !== undefined &&
    safeConfirmationsCount >= nonce; // <-- off by one
  ...
```

`eth_getTransactionCount` returns a **cardinality** — the number of transactions the sender has, i.e. the *next* nonce to use. So the transaction with a given `nonce` is confirmed at the safe block only when `safeConfirmationsCount > nonce`. When `safeConfirmationsCount === nonce`, only the transactions with nonces `0..nonce-1` are confirmed at the safe block; the replacement transaction with `nonce` itself is **not** yet safely confirmed.

Because the check used `>=`, a replacement transaction that had **one fewer than the required number of confirmations** was treated as fully confirmed.

### Concrete failing example

- `requiredConfirmations = 5`, latest block `= 20`, so `confirmedBlockNumber = 20 - 5 + 1 = 16`.
- Ignition sent a tx with `nonce = 10`; the user replaced it (same nonce `10`), and the user's tx was mined in block `19` — that's only `20 - 19 + 1 = 2` confirmations.
- `latestCount` (at block 20) `= 11` → `11 > 10`, so **Case 1** is entered.
- `safeConfirmationsCount` (at block 16) `= 10` (nonces `0..9` are confirmed by block 16; nonce `10` was mined in block 19 > 16, so it is not counted) → `safeConfirmationsCount === nonce === 10`.
- Old code: `10 >= 10` → `hasEnoughConfirmations = true`. Ignition emits `ONCHAIN_INTERACTION_REPLACED_BY_USER`, which clears the nonce and re-sends, **while the user's replacement has only 2 of the required 5 confirmations** and could still be reorged out.

The same `>=` also treats a replacement mined in the *latest* block (a single confirmation) as fully confirmed.

## Fix

Use `>` so the check matches its two sibling comparisons against the same `nonce` in this function — `latestCount > nonce` (Case 1 entry, line above) and `pendingCount > nonce` (Case 2) — both of which already use `>` correctly:

```ts
safeConfirmationsCount > nonce;
```

With the fix, the example above throws `WAITING_FOR_NONCE` and waits until block `23`, where `safeConfirmationsCount` (at block `19`) `= 11 > 10`, i.e. the replacement genuinely has 5 confirmations.

## Test

Adds a regression test to `test/execution/nonce-management/get-nonce-sync-messages.ts` that pins the boundary: `latest = nonce + 1`, `safe = nonce`. It expects `WAITING_FOR_NONCE`.

- **Before** (`>=`): fails at the `latest block #10` iteration — the guard wrongly returns `REPLACED_BY_USER` instead of throwing.
- **After** (`>`): all 38 cases in the file pass.

Existing tests are unaffected — the two nearby cases use `safe = nonce + 1` (proceed) and `safe = nonce - 1` (throw); neither pinned the `safe === nonce` boundary.

Ran locally: `mocha test/execution/nonce-management/get-nonce-sync-messages.ts` → 38 passing.

A Changeset is included (`ignition-core`, patch).

### DIFF
--- a/.changeset/ignition-nonce-sync-confirmation-boundary.md
+++ b/.changeset/ignition-nonce-sync-confirmation-boundary.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/ignition-core": patch
+---
+
+Fix an off-by-one in the nonce-sync confirmation check that could treat a user's replacement transaction as fully confirmed when it had one fewer than the required number of confirmations. A transaction count is a cardinality (the next nonce to use), so the transaction with a given nonce is only confirmed at the safe block when the safe count is strictly greater than the nonce; the check now uses `>` instead of `>=`.
--- a/.changeset/ignition-nonce-sync-confirmation-boundary.md
+++ b/.changeset/ignition-nonce-sync-confirmation-boundary.md
@@ -2,4 +2,4 @@
 "@nomicfoundation/ignition-core": patch
 ---
 
-Fix an off-by-one in the nonce-sync confirmation check that could treat a user's replacement transaction as fully confirmed when it had one fewer than the required number of confirmations. A transaction count is a cardinality (the next nonce to use), so the transaction with a given nonce is only confirmed at the safe block when the safe count is strictly greater than the nonce; the check now uses `>` instead of `>=`.
+Fixed an off-by-one in Ignition's nonce sync that could let a deployment continue when a user's replacement transaction had one fewer than the required number of confirmations.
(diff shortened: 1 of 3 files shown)


Kropiunig:
This is an internal fix in ignition-core (a replacement tx was treated as confirmed one block early) with no user-facing API or docs surface, and a patch changeset is already included. The two remaining required checks look like they need labels: could a maintainer apply "no peer bump needed" and "no docs needed"? I don't have permission to set labels — happy to add entries instead if you'd prefer.

kanej (approved):
This is a really tight PR and description. Thanks @Kropiunig!

PR #8451, opened 2026-07-21
feat: drop Hardhat 2 from init process options

kanej:
The `hardhat --init` process now only provides the option of initializing a new Hardhat 3 project.

Hardhat 2 has been dropped as an option now that Hardhat 3 is stable and Hardhat 2 has an announced end of life (see the [Hardhat 2 end of life policy](https://hardhat.org/docs/reference/hardhat-2-end-of-life) for details).

## Preview

image

## Approach

* Remove the Hardhat 2 templates, and collapse the Hardhat 3 project templates into the now single top level templates directory
* Remove the chooser and options from the CLI flow
* Cleanup analytics but keep sending `hardhat-3` as the version to keep reporting consistent
* other cleanup

## Reviewing

> [!TIP]
> Review a commit at a time

The first commit includes a large number of template file moves and deletions. It is entirely mechanical.

The code edits come next, followed by supporting cleanups in our repo.


schaable (approved):
LGTM!

PR #8452, opened 2026-07-22
Fix typechain mocks bug

kanej:
Typechain is not being passed mock contracts that exist both as source contracts and tests contracts, which breaks JS/TS tests that rely on the type output of the mocks.

Specifically this sources setup (test left to the default of `["test"]`):

```ts
paths: {
  sources: {
    solidity: ["contracts", "test"],
  },
},
```

When the JS/TS tests leverage a mock contract at `./test/mock/my-mock.sol`

The bug is that the scope of a `*.sol` that is under both sources and tests was being assumed to be a test.

This PR changes the scope definition of contracts in `*.sol` files to contract if they are under a path specified in `paths.sources.solidity`. `*.t.sol` files remain tests under all cases.

## Review

> [!TIP]
> Review a commit at a time

## Manual Testing

The following steps will allow you to reproduce the error locally:

1. Within the example project
2. Create a simple contract, and a mock equivalent under `./test/mocks/*`
3. Create a mocha test that uses the mock contract
4. Set the Hardhat config `paths.sources.solidity: ["contracts", "test"]`, you may need to update the tests path as well to `tests: { solidity: "test" }`
5. Build and confirm that the type information comes through in the mocha test (i.e. `tsc -b .`)


packages/hardhat/test/fixture-projects/solidity/unified-sources-tests-overlap-project/hardhat.config.ts
```diff
@@ -0,0 +1,12 @@
+import type { HardhatUserConfig } from "../../../../src/types/config.js";
+
+const config: HardhatUserConfig = {
+  solidity: "0.8.23",
+  paths: {
+    sources: {
+      solidity: ["contracts", "test"],
```
kanej, line 7: This is an unusual setup, but we have seen it in the wild.

packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -153,19 +153,26 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
...
+    // also registered as under a Solidity sources path. Sources takes
+    // precedence (e.g. `test/` is listed in `paths.sources.solidity` to access
+    // `./test/mocks`)
     if (
+      !isUnderSources &&
       fsPath.startsWith(this.#options.solidityTestsPath + path.sep) &&
       fsPath.endsWith(".sol")
     ) {
       return "tests";
     }
```
kanej, line 170: This is the key fix.

packages/hardhat/src/internal/builtin-plugins/solidity/build-system/build-system.ts
```diff
@@ -153,19 +153,26 @@ export class SolidityBuildSystemImplementation implements SolidityBuildSystem {
   }
```
ChristopherDedominici, line 153: If my understanding is correct, if a file is inside a `source` folder, it is treat as a contract, not a test.

The problem is this also triggers when the test folder is inside a sources folder, not just when the same folder is listed in both places.

Example:
someone configures `sources = src/` and `tests = src/test/`. Since `src/test/` is inside `src/`, all their plain `.sol` test files now count as contracts, so Hardhat silently stops running them. No error, no warning, tests just disappear from the run.

Suggested solution:
when a file is under both a sources path and the tests path, let the more specific (deeper) path win. Same folder in both -> sources wins (keeps the mock fix working). Test folder nested inside sources -> tests wins (nothing breaks).

kanej: I hadn't considered that.

I think nesting test folders within source folders is unusual, I would recommend not doing it to be honest. I also think the current rule is easier to reason about; you added the folder to source and it was treated as a source.

I think we keep the rule as it is, but we can revisit if we get reports of issues in real world setups.

ChristopherDedominici (approved):

Issue #8144, opened 2026-04-14
Support contract-level inline config for Solidity tests


popescuoctavian:
### Describe the feature

Hardhat currently only supports inline config at function-level for Solidity tests. However, there are multiple examples of projects that implement contract-level configs, which was intentionally left out in the original design. Hardhat should add support for this.

Suggestion: Read inline config at contract level and apply it for all test functions from the contract. This shouldn't add a lot of overhead in HH (since it already reads inline configs for test functions) and wouldn't require any changes in EDR.

PR #8457, opened 2026-07-23
Remove inline config

### LINKED ISSUE #8117: inline config parsing failure for large code bases
To parse inline config entries we run solc over the code to generate an AST.
Our build system streams json to disk, however our inline parsing does an in-memory json parse that can fail if the parsed string is larger than Node's 512mb limit.

To take as an example [base/contracts](https://github.com/base/contracts) will produced a large build info output that when parsed triggers the Node limit:

```shell
An unexpected error occurred:

Error: Cannot create a string longer than 0x1fffffe8 characters
    at TextDecoder.decode (node:internal/encoding:494:28)
    at bytesToUtf8String (/tmp/end-to-end/base-contracts/node_modules/.pnpm/@nomicfoundation+hardhat-utils@4.0.2/node_modules/@nomicfoundation/hardhat-utils/src/bytes.ts:59:28)
    at collectRawOverrides (/tmp/end-to-end/base-contracts/node_modules/.pnpm/hardhat@3.3.0/node_modules/hardhat/src/internal/builtin-plugins/solidity-test/inline-config/index.ts:157:7)
    at getTestFunctionOverrides (/tmp/end-to-end/base-contracts/node_modules/.pnpm/hardhat@3.3.0/node_modules/hardhat/src/internal/builtin-plugins/solidity-test/inline-config/index.ts:55:27)
    at runSolidityTests (/tmp/end-to-end/base-contracts/node_modules/.pnpm/hardhat@3.3.0/node_modules/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts:172:33)
    at async runSolidityTests (/tmp/end-to-end/base-contracts/node_modules/.pnpm/hardhat@3.3.0/node_modules/hardhat/src/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts:43:5)
    at async Promise.all (index 0)
    at async main (/tmp/end-to-end/base-contracts/node_modules/.pnpm/hardhat@3.3.0/node_modules/hardhat/src/internal/cli/main.ts:221:26)
    at async file:///tmp/end-to-end/base-contracts/node_modules/.pnpm/hardhat@3.3.0/node_modules/hardhat/dist/src/cli.js:21:1 {
  code: 'ERR_STRING_TOO_LONG'
}

If you think this is a bug in Hardhat, please report it here: https://hardhat.org/report-bug
[e2e] Error: Command failed: pnpm hardhat test solidity
 ELIFECYCLE  Command failed with exit code 1.
```

ChristopherDedominici:
Depends on this PR: https://github.com/NomicFoundation/edr/pull/1530

---

###  Problem: contract-level inline config directives are silently ignored

See issue here: https://github.com/NomicFoundation/hardhat/issues/8144

EDR only honors directives placed on a **function**. A directive placed on the
**contract**, which Foundry supports and applies to every test in that
contract, is silently ignored: no error, no warning, the tests just run with
the default configuration.

This works:

```solidity
contract MyTest {
  /// forge-config: default.fuzz.runs = 7
  function testFuzzA(uint256 x) public pure {}
}
```

This is silently ignored:

```solidity
/// forge-config: default.fuzz.runs = 5
contract MyTest {
  function testFuzzA(uint256 x) public pure {}
  function testFuzzB(uint256 x) public pure {}
}
```

### Where to fix

The parsing lives in EDR, not in Hardhat.
Hardhat only passes `testSourcePaths` and `importMappings` to it.
Either EDR should support contract-level directives (Foundry parity), or it
should report them as an error so users aren't misled.


packages/hardhat-errors/src/descriptors.ts
```diff
@@ -1221,57 +1221,8 @@ Remaining test suites: {suites}`,
...
-      INLINE_CONFIG_INVALID_KEY_FOR_TEST_TYPE: {
-        number: 813,
-        messageTemplate: `The inline config key "{key}" in {functionFqn} is not valid for {testType} tests. Valid keys for {testType} tests are: {validKeys}`,
-        websiteTitle: "Inline config key not valid for test type",
-        websiteDescription:
-          "An inline config key was used that does not apply to the type of test function it was attached to. Fuzz test functions (test*) only accept fuzz.* keys and top-level keys, while invariant test functions (invariant*) only accept invariant.* keys and top-level keys.",
-      },
       SELECTED_TEST_FILES_NOT_COMPILED: {
-        number: 814,
+        number: 807,
```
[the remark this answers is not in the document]
ChristopherDedominici: Old code, no longer applicable

packages/hardhat/src/internal/builtin-plugins/solidity-test/runner.ts
```diff
@@ -113,3 +200,84 @@ export function run(
...
+): string {
+  switch (problem.kind) {
+    case "InlineConfigInvalidSolcVersion":
+      return `The source's solc version has no supported grammar, so its inline configuration could not be parsed.`;
+    case "InlineConfigSourceFileNotFound":
+      return `Could not read source file at "${problem.path}": ${problem.reason}.`;
+    case "InlineConfigDirectiveLocation":
+      return `Could not locate a directive of ${problem.contract}.${problem.function} within the source: ${problem.reason}.`;
+  }
+}
```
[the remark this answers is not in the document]
ChristopherDedominici: Old code, no longer applicable

packages/hardhat/src/internal/builtin-plugins/solidity-test/runner.ts
```diff
@@ -113,3 +200,137 @@ export function run(
...
+
+  switch (kind) {
+    case "source":
+      return INLINE_CONFIG_SOURCE_PROBLEM_KINDS.has(problem.kind);
+    case "directive":
+      return INLINE_CONFIG_DIRECTIVE_PROBLEM_KINDS.has(problem.kind);
+    default:
+      return false;
+  }
+}
```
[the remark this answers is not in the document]
ChristopherDedominici: Old code, no longer applicable

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -196,6 +198,32 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
     ({ edrArtifact }) => edrArtifact.id,
   );
 
+  // Maps each test suite's solc source name to its absolute path on disk, so
+  // EDR can read and parse inline test configuration directly from the sources.
+  const testSourcePaths = Object.fromEntries(
+    testSuiteArtifacts.map(({ userSourceName, edrArtifact }) => [
+      edrArtifact.id.source,
```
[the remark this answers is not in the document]
ChristopherDedominici: Fixed: https://github.com/NomicFoundation/hardhat/pull/8457/changes/58db018f7d10fa6447a337e284974bd67c22f543

packages/hardhat/src/internal/builtin-plugins/solidity-test/runner.ts
```diff
@@ -100,6 +100,27 @@ export function run(
...
+      // them by attaching a structured `inlineConfigErrors` array to the
+      // rejection, and its message already renders every problem located by
+      // source, line, and function. Report that as a Hardhat error instead of
+      // as an unhandled EDR error, mapping solc source names back to the
+      // user's source names.
+      if ("inlineConfigErrors" in error) {
+        stream.destroy(
+          new HardhatError(
+            HardhatError.ERRORS.CORE.SOLIDITY_TESTS.INVALID_INLINE_CONFIG,
+            {
```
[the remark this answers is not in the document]
ChristopherDedominici: Fixed: Fixed: https://github.com/NomicFoundation/hardhat/pull/8457/changes/58db018f7d10fa6447a337e284974bd67c22f543

packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
```diff
@@ -163,11 +165,27 @@ export async function solidityTestConfigToSolidityTestRunnerConfigArgs({
...
     eip712CanonicalTypes,
+    testSourcePaths,
+    importMappings,
   };
 }
 
+// Shared suffix of the inline config directive prefixes (`hardhat-config:`
+// and `forge-config:`). Scanning for it once is cheaper than scanning per
+// prefix; false positives are rare and harmless.
+const INLINE_CONFIG_PREFIX_SUFFIX = "-config:";
```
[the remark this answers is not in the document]
ChristopherDedominici: Fixed: https://github.com/NomicFoundation/hardhat/pull/8457/changes/8b39ce7017d6b009aa320f8267baca22aec6abf0

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -196,6 +198,29 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
...
+  //
+  // NOTE: This rebuilds a dependency graph that the preceding `build` already
+  // computed internally, because the build-system API doesn't expose it.
+  const testDependencyGraph = await buildDependencyGraph(
+    testRootPathsToRun.toSorted(),
+    hre.config.paths.root,
+    readSourceFileFactory(hre.hooks),
+    hre.hooks,
+  );
+  const importMappings = buildImportMappings(testDependencyGraph);
```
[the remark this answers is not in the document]
ChristopherDedominici: This was true, but it's addressed now: the graph rebuild is gated behind `buildInfoContainsInlineConfig`, so it only runs when a compiled source actually contains an inline config directive (and only over the test roots, not all contracts). Projects that don't use inline config never pay the cost.

Reusing the graph from build would be the ideal fix, but the build-system API doesn't expose it, so that would require a deeper change.

Probably can be addressed in following PR. If this is agreed, I'll create an issue to track the work

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -198,6 +201,44 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
...
+    )
+  ) {
+    // Maps each test suite's solc source name to its absolute path on disk,
+    // so EDR can read the inline test configuration from the sources.
+    testSourcePaths = Object.fromEntries(
+      testSuiteArtifacts.map(({ userSourceName, edrArtifact }) => [
+        edrArtifact.id.source,
+        resolveFromRoot(hre.config.paths.root, userSourceName),
+      ]),
+    );
```
schaable: Only passing `testSourcePaths` when `buildInfoContainsInlineConfig` is `true` may be redundant, as 
https://github.com/NomicFoundation/edr/pull/1530/changes#diff-f2d91594d3c69fa713639164fb878dbd979c216831245c6a857e10b48dcbc782R944 seems to already do the same filtering. May be worth checking with the EDR team if it makes sense to always pass `testSourcePaths` directly.

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -198,6 +201,44 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
...
+    // NOTE: This rebuilds a dependency graph the preceding `build` already
+    // computed internally, because the build-system API doesn't expose it.
+    const testDependencyGraph = await buildDependencyGraph(
+      testRootPathsToRun.toSorted(),
+      hre.config.paths.root,
+      readSourceFileFactory(hre.hooks),
+      hre.hooks,
+    );
+
+    importMappings = buildImportMappings(testDependencyGraph);
```
schaable: I'm not sure about this, but Claude is saying that leaving `importMappings` undefined doesn't change the output. It may be worth checking with the EDR team what the parameter is for and why it's needed, since passing it requires calling `buildInfoContainsInlineConfig` and `buildImportMappings`, which are both expensive.

packages/hardhat/src/internal/builtin-plugins/solidity-test/runner.ts
```diff
@@ -100,6 +101,21 @@ export function run(
...
+              errors: formatInlineConfigErrorMessage(error.message),
+            },
+          ),
+        );
+        return;
+      }
```
schaable: Each `error.inlineConfigErrors` entry has the location (`sourceName`, `contract`, `function`, `line`) and a typed `problem` with the relevant details, like `key`, `value`, `expected`, `profile`, `directive`, or `testType` (see [https://github.com/NomicFoundation/edr/pull/1530/changes#diff-a5835722ec14ad3343fa1d9dd416cb1459a47a616aced0ec68d2db05add12f95R171](https://github.com/NomicFoundation/edr/pull/1530/changes#diff-a5835722ec14ad3343fa1d9dd416cb1459a47a616aced0ec68d2db05add12f95R171)). Instead of deprecating errors 807–811/813, we could map those directly. Even if we still want to use a single error code for all of them and keep 821, we can get the relevant info from `inlineConfigError` and avoid `formatInlineConfigErrorMessage`. You can get the `userSourceName` from `sourceNameToUserSourceName`.

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
schaable, line 354: This produces the same error twice. `runner.ts` calls `stream.destroy(...)`, which emits `'error'` on `runStream`, then `finished(runStream)` rejects and the catch block logs it. 

That same error also propagates to `testReporterStream`, which is composed from `runStream`. Since `.pipe()` doesn't forward stream errors to the destination, nothing handles that `'error'`, so Node turns it into an `uncaughtException` and Hardhat logs it again. 

For example:

```solidity
// test/contracts/InvalidInline.t.sol
pragma solidity ^0.8.0;

contract InvalidInlineTest {
  /// forge-config: default.not-a-key = 1
  function testFuzzInvalid(uint256 x) public pure {}
}
```

```
$ npx hardhat test solidity

Running Solidity tests

HardhatError: HHE819: Invalid inline test configuration:

  test/contracts/InvalidInline.t.sol:5: InvalidInlineTest.testFuzzInvalid: invalid key `default.not-a-key`
    at <anonymous> (packages/hardhat/src/internal/builtin-plugins/solidity-test/runner.ts:109:11)


Uncaught exception:

HardhatError: HHE819: Invalid inline test configuration:

  test/contracts/InvalidInline.t.sol:5: InvalidInlineTest.testFuzzInvalid: invalid key `default.not-a-key`
    at <anonymous> (packages/hardhat/src/internal/builtin-plugins/solidity-test/runner.ts:109:11)
```

Should be fixable with:

```typescript
testReporterStream.on("error", () => {});   // ignore error, as it's handled by the finished(runStream) catch below

try {
  await finished(runStream);
  await finished(outputStream);
} catch (error) { ... }
```

schaable (changes requested):

packages/hardhat/src/internal/builtin-plugins/solidity-test/runner.ts
ChristopherDedominici: Fixed: https://github.com/NomicFoundation/hardhat/pull/8457/changes/4ab0598a8f8534c854b7d174362bbd08cbfb44d7

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
ChristopherDedominici: Minor note: this happens in node v26 (node v24 does not have the issue). Fixed here: https://github.com/NomicFoundation/hardhat/pull/8457/changes/d70c5853557ea9c537d635e2a18ac7b9f038bb43

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
ChristopherDedominici: Done. `testSourcePaths` is now always passed, and since EDR already skips sources with no directive we dropped the check entirely, along with the import mappings and the dependency-graph rebuild they required.
Fix: https://github.com/NomicFoundation/hardhat/pull/8457/changes/e9d31e683979fe5135d5d4ff2f05c0394fa7e30c

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
ChristopherDedominici: You're right, it's a no-op. EDR only reads directives from the test file's own source and never follows its imports (its own test suite runs with an empty resolver). Removed it along with the dependency-graph rebuild and the `buildInfoContainsInlineConfig` check; the fixture that imports via a non-relative path still passes.
Fix: https://github.com/NomicFoundation/hardhat/pull/8457/changes/e9d31e683979fe5135d5d4ff2f05c0394fa7e30c

packages/hardhat/test/fixture-projects/solidity-test-inline-config/hardhat.config.ts
```diff
@@ -0,0 +1,8 @@
+import HardhatSolidityPlugin from "../../../src/internal/builtin-plugins/solidity-test/index.js";
+import { HardhatUserConfig } from "../../../src/types/config.js";
+
+const config: HardhatUserConfig = {
+  plugins: [HardhatSolidityPlugin],
+};
+
+export default config;
```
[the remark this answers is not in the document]
ChristopherDedominici: false

packages/hardhat-utils/src/bytes.ts
```diff
@@ -75,27 +75,50 @@ export function bytesIncludesUtf8String(
...
+ * algorithm.
+ *
+ * @param haystack The Uint8Array to search in.
+ * @param needle The string whose UTF-8 encoding to search for.
+ * @param fromIndex The byte index to start the search at. Defaults to 0.
+ * @returns The byte index of the first match at or after `fromIndex`, or -1
+ * if there is none. An empty needle matches at `fromIndex` (clamped to the
+ * haystack's bounds), mirroring `String.prototype.indexOf`.
+ */
+export function bytesIndexOfUtf8String(
```
schaable: This is no longer used anywhere. If we keep it, I'd make it internal, but given the limited usage, I'd just roll back the code.

packages/hardhat/test/fixture-projects/solidity-test-inline-config/node_modules/dependency/src/BaseTest.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+// Imported by the test sources through a non-relative path, so the test run
+// exercises the import mappings passed to EDR.
```
schaable: Nit: this is no longer true. It's good to keep the fixture as a regression, but the comment is misleading.

schaable (approved):
Two minor comments, but this looks good to me. Pre-approving.

packages/hardhat-utils/src/bytes.ts
ChristopherDedominici: Done: https://github.com/NomicFoundation/hardhat/pull/8457/changes/ce669506ef5276312235a0250e61911a1c32c271

packages/hardhat/test/fixture-projects/solidity-test-inline-config/node_modules/dependency/src/BaseTest.sol
ChristopherDedominici: Done: https://github.com/NomicFoundation/hardhat/pull/8457/changes/ce669506ef5276312235a0250e61911a1c32c271

packages/hardhat/test/internal/builtin-plugins/solidity-test/inline-config.ts
```diff
@@ -0,0 +1,128 @@
...
+
+import assert from "node:assert/strict";
+import { describe, it } from "node:test";
+
+import { HardhatError } from "@nomicfoundation/hardhat-errors";
+import { useFixtureProject } from "@nomicfoundation/hardhat-test-utils";
+
+import { createHardhatRuntimeEnvironment } from "../../../../src/internal/hre-initialization.js";
+import hardhatConfig from "../../../fixture-projects/solidity-test-inline-config/hardhat.config.js";
+
```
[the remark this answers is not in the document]
ChristopherDedominici: false

PR #8458, opened 2026-07-23
fix: don't copy build outputs from templates during init

schaable:
When you run `hardhat init`, Hardhat copies a template project into the new directory. To determine which files to copy, it scans the template directory, but it only ignored `node_modules`, not other generated directories like `artifacts`, `cache`, and `types`.

This caused two issues:

* If a template directory contained leftover build output, those files were copied into the newly created project.
* Our tests compile templates in place, creating `artifacts`, and clean them up afterward. If the scan happened while those files briefly existed, `init` could try to copy a file that had already been deleted, causing a `FileNotFoundError`.

This change makes `init` ignore generated directories entirely, so only actual template files are copied. It also adds a regression test covering these exclusions.

### DIFF
--- a/.changeset/quick-owls-cheer.md
+++ b/.changeset/quick-owls-cheer.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed project initialization so that compilation and build output directories are never copied from a template into a new project.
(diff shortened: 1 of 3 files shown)


kanej (approved):
Can I approve twice?

PR #8464, opened 2026-07-25
Prefer environment variables when resolving configuration variables

### LINKED ISSUE #7596: Environment variables should take precedence when resolving configuration variables

nishant-uxs:
## Summary
- Resolve config vars from `process.env` inside `LazyResolvedConfigurationVariable` **before** running the `configurationVariables` hook chain
- Ensures env always wins over every plugin (including keystore), matching @alcuadrado’s guidance on #7596
- Adds unit tests that hooks are not invoked when the env var is set (including empty string)

## Test plan
- [x] `node --import tsx/esm --test test/internal/core/configuration-variables/configuration-variables.ts` (15 passing)
- [ ] CI green

Closes #7596


nishant-uxs:
Fixed the `import/order` lint failure in the configuration-variables test (hooks import before hre).

alcuadrado (approved):

nishant-uxs:
Merge-queue CI failed on a transient network flake, not this PR's code: `[hardhat-toolbox-mocha-ethers] ci on windows-latest` couldn't download from `binaries.soliditylang.org` (`read ECONNRESET`). The PR-branch CI itself is green.

Also stripped the UTF-8 BOM from `.changeset/env-var-config-precedence.md` (Copilot note) so frontmatter starts cleanly with `---`.

@alcuadrado whenever convenient, could you re-add this to the merge queue?

PR #8465, opened 2026-07-26
fix: return new request objects from network handlers

### LINKED ISSUE #8090: Adapt network manager handlers to copy rather than mutate requests
The hook handlers for `onRequest` previously ensured that a passed in request would not be modified using a `deepClone`. We have remove this clone as a performance improvement - but this technically alters the ownership commitments of the `onRequest` hook.

So at a high level:

```ts
const myRequest = {...}
const response = await next(myRequest)
doSomething(myRequest)
```
Next previously commited to not mutating myRequest - so `doSomething(myRequest)` was safe. With the removal of clone we are mutating `myRequest` within `next`. In practice this is not currently an issue.

However, we can update the internal handlers applied in `onRequest` so that instead of updating in place, they return a new object.

See https://github.com/NomicFoundation/hardhat/blob/3268b339c4eb64c530e77997d70ecda45e42ca70/v-next/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/network.ts#L45-L78

nishant-uxs:
## Summary
- Network manager request handlers now return shallow-copied JSON-RPC requests when filling gas / sender / local-account fields instead of mutating the caller's object.
- Adds `replaceJsonRpcRequestTx` helper and regression tests that the original request remains unchanged through `onRequest`.

Closes #8090

## Test plan
- [x] `tsx --test` for network-manager gas, accounts, and hook-handler suites (84 passing)

### DIFF
--- a/.changeset/immutable-network-request-handlers.md
+++ b/.changeset/immutable-network-request-handlers.md
@@ -1,0 +1,5 @@
+﻿---
+"hardhat": patch
+---
+
+Return new JSON-RPC request objects from network request handlers instead of mutating the caller's request
--- a/packages/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/network.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/network-manager/hook-handlers/network.ts
@@ -59,9 +59,8 @@ export default async (): Promise<Partial<NetworkHooks>> => {
         });
       }
 
-      // We previously cloned here, but the performance impact is significant.
-      // TODO: ensure the passed in request is not mutated by adapting the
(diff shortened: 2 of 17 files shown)


nishant-uxs:
Fixed the lint failure: tests were using `structuredClone`, which Hardhat forbids via `no-restricted-globals`. Replaced with `await deepClone(...)` from `@nomicfoundation/hardhat-utils/lang`.

kanej:
Hey @nishant-uxs, thanks for the work here. But we are going to need to dig into this one internally.
This is on the hot path, and will need careful benchmarking before we are sure. Indeed recent changes may mean the original request clone is now acceptable.

I have update the issue to try and avoid external contributors getting waylaid again.

PR #8466, opened 2026-07-27
feat: read Forge gas snapshot format in --snapshot-check

### LINKED ISSUE #8357: Optionally read Forge's `.gas-snapshot` format in `--snapshot-check`
## Summary

`npx hardhat test solidity --snapshot-check` can't validate a `.gas-snapshot` produced by `forge snapshot`, even though the gas values are identical, it fails with `HHE803: Invalid format in snapshot file`. Optionally reading Forge's format would smooth Foundry → Hardhat migrations.

## Background

Hardhat reads/writes its whole-suite snapshot at the hardcoded path `<root>/.gas-snapshot` — the same filename Forge uses — but the parser hard-requires a `#` separator where Forge writes `:`:

The separator is the only structural difference; the `(gas: N)` / `(runs:, μ:, ~:)` details are byte-identical, and the gas values are computed identically. So a Forge-format line falls through to `throw … INVALID_SNAPSHOT_FORMAT`.

## Motivation

Projects migrating Foundry → Hardhat may already commit a `.gas-snapshot`; today they must regenerate it under Hardhat before `--snapshot-check` works, with no signal that the gas numbers actually agree. (It would also let repos keep both toolchains share a baseline, though that's a secondary consideration.)

## Proposed enhancement

**Accept both separators on the read path** — relax the parse regexes to allow either `:` or `#`. Hardhat keeps writing its own `#` format. No new config.

> Note: accepting `:` at read time should cover most basic projects, but may not be sufficient on its own for full parity. The two tools seem to differ in how they name entries when the same `ContractName` appears in multiple files — Hardhat appears to use a fully-qualified name in that case, while Forge files use the bare contract name. Flagging for awareness.

### Search terms

gas snapshots, snapshot, --snapshot-check

nishant-uxs:
## Summary
- Accept both Hardhat (#) and Forge (:) separators when parsing .gas-snapshot files during --snapshot-check
- Continue writing snapshots in Hardhat's # format with --snapshot
- Eases Foundry → Hardhat migration without requiring immediate snapshot file regeneration

Closes #8357

## Documentation
No documentation updates are required for this change. It relaxes snapshot parsing on the read path only; existing gas snapshot docs already cover --snapshot-check. Related: https://github.com/NomicFoundation/hardhat-website/issues/212

## Test plan
- [x] All 56 function-gas-snapshots unit tests pass
- [x] Forge-format standard and fuzz lines parse correctly
- [x] FQN contract names with colon separator parse correctly
- [x] Mixed #/: separators in one file parse correctly
- [x] Comparison passes when Forge baseline matches Hardhat run with only separator difference
- [x] Hardhat write path still produces # separator format

### DIFF
--- a/.changeset/read-forge-gas-snapshot-format.md
+++ b/.changeset/read-forge-gas-snapshot-format.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Accept Forge's colon separator when reading `.gas-snapshot` files during `--snapshot-check`, while continuing to write snapshots in Hardhat's `#` format.
--- a/.peer-bumps.json
+++ b/.peer-bumps.json
@@ -2,9 +2,9 @@
   "$schema": "./scripts/peer-bumps.schema.json",
   "excludedFolders": [
     "archive",
+    "packages/config",
     "packages/example-project",
-    "packages/hardhat/templates",
-    "packages/config"
+    "packages/hardhat/templates"
   ],
   "bumps": []
 }
--- a/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
@@ -189,8 +189,9 @@ export function parseFunctionGasSnapshots(
   const lines = stringifiedSnapshots.split("\n");
   const snapshots: FunctionGasSnapshot[] = [];
 
-  const standardTestRegex = /^(.+)#(.+) \(gas: (\d+)\)$/;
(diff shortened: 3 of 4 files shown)


nishant-uxs:
CI metadata checks: this PR needs the \
o peer bump needed\ label (external contributors cannot add labels on upstream). No peer dependency bumps are required for this patch-only \hardhat\ change. Could a maintainer add that label when triaging? (\
o docs needed\ is optional — PR body now links a related hardhat-website issue for the docs check.)

schaable:
Closing this for now, as the review raised a few scenarios that need design decisions and more work. I’ll leave a summary on the issue for future reference, but I think this one should be handled internally by the team.

PR #8468, opened 2026-07-27
feat(bench): measure memory over time via `/proc` sampling

Wodann:

---

The e2e regression benchmarks previously captured only **peak** memory via GNU `time -f %M`. GNU time reads the kernel's `ru_maxrss` once at `wait4()` — an exact peak, but fundamentally incapable of producing a time series, and its `%M` is the max of any *single* process in the tree, not an aggregate.

This PR samples each measured run's process tree from `/proc` every 100 ms, producing a **memory-over-time series per process** (hardhat vs. solc vs. wrappers) alongside the existing metrics, and unifies all benchmark execution (single commands and step sequences, `pnpm bench` and `pnpm bench:regression`) onto one shared in-process runner — dropping both **hyperfine** and **GNU time** as dependencies.

## How measurement works now

One `bash` spawn per measured run captures everything in a single pass (`scripts/benchmark/helpers/runner.ts`):

- **Wall-clock**: `performance.now()` around the child, minus a shell-spawn calibration offset (the wrapped no-op `:` timed 20× at startup, mean subtracted — the same calibration hyperfine applies). Measured offset: ~1.9 ms.
- **CPU**: bash's `time` builtin (`%U %S`) — a reserved word, not a process: zero fork/exec, and it reads the exact child rusage the kernel reports at `wait()`.
- **Memory** (`scripts/benchmark/helpers/mem-series.ts`): every 100 ms, walk `/proc/<pid>/task/*/children` and read each process's `/proc/<pid>/status`. One sample costs ~50–200 µs in the otherwise-idle driver process.
  - `VmRSS` per process, aggregated by label (`hardhat`, `solc-linux-amd64-v0.8.28+…`, `npx`, …; generic script names like `cli.js` resolve to their npm package, including through `.bin` shims) → the series.
  - `VmHWM` — the kernel's *exact* per-process peak-RSS high-water mark (the same counter behind `%M`) → the peak, for free in the same read. It must be tracked during sampling: `/proc/<pid>` vanishes at process exit, and post-mortem rusage is only available to the parent (bash), not the driver.

On non-Linux (no `/proc`) memory entries are skipped with a warning; timing still works.

## Report format

Every measured name now emits four entries (previously three):

| entry | unit | value | extra |
|---|---|---|---|
| `<scenario> / <name>` | s | mean wall-clock | per-run `times`, min/max/median/mean, `peakRssMb` |
| `… (cpu)` | s | mean total CPU (user+system) | per-run totals + min/max/median/mean, mean `user`/`system` split |
| `… (peak RSS)` | MB | highest per-run peak (max single-process `VmHWM`, `%M` semantics) | per-run peaks + statistics |
| `… (mem over time)` **(new)** | MB | median over runs of the run's median tree-total RSS | see below |

The `(mem over time)` value is `median_r( median_s( Σ_p RSS[r][s][p] ) )` — a duration-weighted "typical working set", robust to spikes (which `(peak RSS)` tracks) and outlier runs, suitable for the 110% drift alerting. Note it aggregates the **sum** of the tree, so it can legitimately exceed `(peak RSS)` (max of a **single** process).

Its `extra`:

```json
{
  "representativeRun": 1,
  "seriesGz": "H4sIAAAA…",   // raw series of the run with the median peak
  "runs": [
    { "durationMs": 4139, "peakRssMb": 315, "total": [3, 250, 383, 425, 569] },
    { "durationMs": 4174, "peakRssMb": 315, "total": [1, 247, 367, 424, 565] }
  ]
}
```

`seriesGz` = base64(gzip(JSON of `{ tMs, byProcess: { label: MB[] } }` with every array delta-encoded)). Decode: gunzip (browsers: `DecompressionStream("gzip")`), `JSON.parse`, cumulative sum per array. When compression wouldn't help (tiny runs) the raw table ships as `series` instead, so the payload is never larger than uncompressed. `total` is the `[p0,p25,p50,p75,p100]` summary of each run's tree-total samples.

## Validation

**GNU time equivalence** (gate for deleting the `%M` path):

| case | GNU time | sampler |
|---|---|---|
| single 300 MB hog, peak | 344 MB | 343 MB |
| tree (200 MB + 350 MB children), peak | 395 MB | 394 MB (max-not-sum semantics confirmed) |
| CPU user+sys, 4 runs | mean 0.637 s | mean 0.651 s (ranges overlap; GNU time rounds to 10 ms) |

**Hyperfine equivalence** (gate for dropping hyperfine; 10 runs each, calibrated runner):

| command | hyperfine | this runner | Δ mean |
|---|---|---|---|
| `sleep 0.5` | 501.5 ms ± 0.6 | 501.6 ms ± 0.4 | **+0.1 ms** |
| CPU burn (~430 ms) | 431.1 ms ± 7.0 | 432.3 ms ± 7.3 | +1.2 ms (≪ noise) |
| shell pipeline (~4 ms) | 3.7 ms ± 1.1 | 3.5 ms ± 0.4 | −0.2 ms |

**Compression** (real `ens-verifiable-factory` data, six `(mem over time)` entries, ~40-sample series):

| encoding | total bytes | vs raw |
|---|---|---|
| raw JSON | 4681 | — |
| gzip + base64 | 2784 | −40.5% |
| delta + gzip + base64 (**implemented**) | 2360 | −49.6% |

It won on every entry even at these unfavorably short series lengths; long `test solidity` runs compress much better. The raw fallback guarantees it can never lose.

## Also in this PR

- Single commands (previously hyperfine) gained **per-run** peak RSS and per-run CPU spread — hyperfine only exported one aggregate peak and mean CPU.
- `pnpm bench` keeps its full CLI (`--warmup`, `--prepare`, `--show-output`, `--ignore-failure`) with hyperfine-compatible `--export-json` output, extended with per-run memory series.
- CI workflow and `scripts/setup.sh` no longer install `hyperfine`/`time`.
- Failure output is unchanged: first error line, repro hint, full stdout/stderr.

## Expected one-time effects on the dashboard

- Timing methodology changed (calibrated in-process vs. hyperfine): validated to agree within ~1 ms, so any step in the charts is ≪ the 110% alert threshold.
- `(cpu)` entries gain a real `±` range and per-run statistics in `extra`.

## Follow-up (hardhat-benchmark-results repo)

Consume the new entry: the scalar chart comes free from the existing renderer; add the commit-overlay view (one line per commit, gradient by age) and the commit × time-in-run heatmap — both read the same `seriesGz` payload. Interactive heatmap mockup shared separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


scripts/benchmark/main.ts
```diff
@@ -81,7 +87,7 @@ export async function runBenchmark(benchArgs: BenchArgs): Promise<void> {
   const benchCommand = command ?? scenario.definition.defaultCommand;
   const runs = benchArgs.runs ?? 10;
 
-  if (init) {
+  if (init || !existsSync(scenario.workingDir)) {
```
Wodann, line 89: note: This was previously in `e2eExec`—which we no longer call—so we move the `!existsSync` check up here

PR #8474, opened 2026-07-29
feat: add analytics per unique network chain type

Wodann:

---

Adds a `network` analytics event that records which chain type (`generic`, `l1`, `op`, or a custom plugin-defined value) users configure when creating a network connection.

### Changes

- **New `network` analytics event** — added a variant to the `AnalyticsEvent` union (`src/internal/cli/telemetry/analytics/types.ts`) with `params: { chainType: string }`. The chain type is reported as a raw string so custom plugin-defined chain types are visible too.
- **New `sendNetworkAnalytics(chainType)` sender** (`src/internal/cli/telemetry/analytics/analytics.ts`) — a stateless sibling of `sendTaskAnalytics`/`sendProjectTypeAnalytics` that reuses the existing consent-gated `sendAnalytics` pipeline (telemetry permission check + detached subprocess).
- **Instrumented `NetworkManagerImplementation.create`** (`src/internal/builtin-plugins/network-manager/network-manager.ts`) — after the connection is built (so the *resolved* chain type is known), the event is sent **once per chain type per process** via an instance-level `Set`, and fire-and-forget so it never delays or breaks connection creation. `connect`, `getOrCreate` cache-misses, and `createServer` all funnel through `create`, so they're all covered.

### Performance considerations

- **Bootstrap time is unaffected.** The analytics import in `network-manager.ts` is a static import, which is correct per our import rules because that module is itself lazily loaded (the `hre` hook handler dynamically imports it on first `hre.network.*` use). Verified with a module-load trace: `hardhat --version` does not load `network-manager.js`.
- **Steady-state overhead per `create()` call** is a single `Set.has()` string lookup. The dedupe `Set` is marked before the consent check, so when telemetry is disabled, repeat connections skip even the consent-file read. The subprocess spawn happens at most once per distinct chain type per process.

### Tests

- Payload-shape test for `sendNetworkAnalytics`, mirroring the existing sender tests (`test/internal/cli/telemetry/analytics/analytics.ts`).
- Dedupe test verifying that a second `create()` with the same chain type sends nothing, while a connection with a new chain type sends a second event (`test/internal/builtin-plugins/network-manager/network-manager.ts`).

### Notes for reviewer

Are newly added analytics mentioned in the changesets of Hardhat?

### DIFF
--- a/.changeset/polite-words-take.md
+++ b/.changeset/polite-words-take.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Report chain type to Google Analytics
--- a/packages/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/network-manager/network-manager.ts
@@ -32,6 +32,7 @@ import { exists, readBinaryFile } from "@nomicfoundation/hardhat-utils/fs";
 import { deepMerge } from "@nomicfoundation/hardhat-utils/lang";
 import { AsyncMutex } from "@nomicfoundation/hardhat-utils/synchronization";
 
+import { sendNetworkAnalytics } from "../../cli/telemetry/analytics/analytics.js";
 import { resolveUserConfigToHardhatConfig } from "../../core/hre.js";
 import { isSupportedChainType } from "../../edr/chain-type.js";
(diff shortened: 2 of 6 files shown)


alcuadrado (approved):
pre-approving. This is only missing a patch changeset

Wodann:
Analytics were tested with a local build and validated to be reported to GA

PR #8487, opened 2026-08-04
Add a --tolerance flag to Solidity test gas snapshots

### LINKED ISSUE #8382: Add a `--tolerance` flag to Solidity test gas snapshots
Hardhat's Solidity test gas snapshots (`hardhat test --snapshot` / `--snapshot-check`) currently use exact comparisons. Any gas change, even +1, fails the check and requires re-running `--snapshot`.

Foundry offers a `--tolerance <PERCENT>` option on `forge snapshot --check`. It allows each test's gas usage to drift by up to the specified percentage without failing the check (the snapshot isn't updated automatically and the diff is ignored). The option applies to both standard and fuzz tests and was introduced as a DX improvement: small gas changes are common, and overly strict checks often result in snapshot updates being accepted without much review https://github.com/foundry-rs/foundry/pull/3102.

Hardhat already runs fuzz tests with a fixed default seed, so gas usage is deterministic across runs. This isn't about reducing fuzz-related noise, but about avoiding churn from small intentional or cumulative gas changes. The downside is that a tolerance threshold can allow gas regressions to accumulate unnoticed against an outdated baseline.

The goal of this issue is to gauge interest. If you'd find a tolerance option useful for gas snapshots, please upvote the issue and, if possible, share your use case.

ChristopherDedominici:
Fixes: https://github.com/NomicFoundation/hardhat/issues/8382

Docs updated in PR: https://github.com/NomicFoundation/hardhat-website/pull/291

### DIFF
--- a/.changeset/puny-eels-swim.md
+++ b/.changeset/puny-eels-swim.md
@@ -1,0 +1,7 @@
+---
+# docs: https://github.com/NomicFoundation/hardhat-website/pull/291
+"@nomicfoundation/hardhat-errors": patch
+"hardhat": patch
+---
+
+Added a `--tolerance` flag to Solidity test gas snapshots.
--- a/.changeset/puny-eels-swim.md
+++ b/.changeset/puny-eels-swim.md
@@ -4,4 +4,4 @@
 "hardhat": patch
 ---
 
-Added a `--tolerance` flag to Solidity test gas snapshots.
+Added a `--tolerance` option to `--snapshot-check` in Solidity tests, allowing snapshot values to drift by a given percentage before the check fails.
--- a/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
@@ -250,7 +250,7 @@ export function parseFunctionGasSnapshots(
 export function compareFunctionGasSnapshots(
   previousSnapshots: FunctionGasSnapshot[],
   currentSnapshots: FunctionGasSnapshotWithMetadata[],
-  tolerance: number = 0,
+  tolerance: number,
 ): FunctionGasSnapshotComparison {
   const previousSnapshotsMap = new Map(
     previousSnapshots.map((s) => [
@@ -334,7 +334,7 @@ export function hasGasUsageChanged(
 export async function checkFunctionGasSnapshots(
   basePath: string,
   suiteResults: SuiteResult[],
-  tolerance: number = 0,
+  tolerance: number,
 ): Promise<FunctionGasSnapshotCheckResult> {
   const functionGasSnapshots = extractFunctionGasSnapshots(suiteResults);
 
--- a/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
@@ -53,8 +53,9 @@ export interface FunctionGasSnapshotComparison {
   added: FunctionGasSnapshot[];
   removed: FunctionGasSnapshot[];
   changed: FunctionGasSnapshotChange[];
-  // Diffs within the allowed tolerance: they don't fail the check, but are
-  // surfaced so a passing check doesn't silently hide drift.
+  // Diffs within the allowed tolerance: they don't fail the check. They are
+  // included in the result for programmatic inspection, but are not shown in
+  // the default console output.
   tolerated: FunctionGasSnapshotChange[];
 }
 
--- a/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
@@ -17,7 +17,11 @@ import {
   parseFullyQualifiedName,
 } from "../../../utils/contract-names.js";
 
-import { formatSectionHeader, getUserFqn } from "./helpers/utils.js";
+import {
+  formatSectionHeader,
+  getUserFqn,
+  isWithinTolerance,
+} from "./helpers/utils.js";
 
 export const FUNCTION_GAS_SNAPSHOTS_FILE = ".gas-snapshot";
 
@@ -49,6 +53,9 @@ export interface FunctionGasSnapshotComparison {
   added: FunctionGasSnapshot[];
   removed: FunctionGasSnapshot[];
   changed: FunctionGasSnapshotChange[];
+  // Diffs within the allowed tolerance: they don't fail the check, but are
+  // surfaced so a passing check doesn't silently hide drift.
+  tolerated: FunctionGasSnapshotChange[];
 }
 
 export interface FunctionGasSnapshotChange {
@@ -242,6 +249,7 @@ export function parseFunctionGasSnapshots(
 export function compareFunctionGasSnapshots(
   previousSnapshots: FunctionGasSnapshot[],
   currentSnapshots: FunctionGasSnapshotWithMetadata[],
+  tolerance: number = 0,
 ): FunctionGasSnapshotComparison {
   const previousSnapshotsMap = new Map(
     previousSnapshots.map((s) => [
@@ -252,6 +260,7 @@ export function compareFunctionGasSnapshots(
 
   const added: FunctionGasSnapshot[] = [];
   const changed: FunctionGasSnapshotChange[] = [];
+  const tolerated: FunctionGasSnapshotChange[] = [];
 
   for (const current of currentSnapshots) {
     const key = `${current.contractNameOrFqn}#${current.functionSig}`;
@@ -278,7 +287,7 @@ export function compareFunctionGasSnapshots(
           ? current.gasUsage.gas
           : current.gasUsage.medianGas;
 
-      changed.push({
+      const change: FunctionGasSnapshotChange = {
         contractNameOrFqn: current.contractNameOrFqn,
         functionSig: current.functionSig,
         kind: currentKind,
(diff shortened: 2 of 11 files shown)


packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
```diff
@@ -49,6 +53,9 @@ export interface FunctionGasSnapshotComparison {
   added: FunctionGasSnapshot[];
   removed: FunctionGasSnapshot[];
   changed: FunctionGasSnapshotChange[];
+  // Diffs within the allowed tolerance: they don't fail the check, but are
+  // surfaced so a passing check doesn't silently hide drift.
+  tolerated: FunctionGasSnapshotChange[];
```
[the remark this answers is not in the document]
ChristopherDedominici: Done: https://github.com/NomicFoundation/hardhat/pull/8487/changes/e765fa97b12b8e5eabe3e0a75b01d97fe446dfbf

packages/hardhat/src/internal/builtin-plugins/gas-analytics/snapshot-cheatcodes.ts
```diff
@@ -73,6 +77,9 @@ export interface SnapshotCheatcodesComparison {
   added: SnapshotCheatcode[];
   removed: SnapshotCheatcode[];
   changed: SnapshotCheatcodeChange[];
+  // Diffs within the allowed tolerance: they don't fail the check, but are
+  // surfaced so a passing check doesn't silently hide drift.
+  tolerated: SnapshotCheatcodeChange[];
```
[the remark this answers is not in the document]
ChristopherDedominici: Done: https://github.com/NomicFoundation/hardhat/pull/8487/changes/e765fa97b12b8e5eabe3e0a75b01d97fe446dfbf

packages/hardhat/src/internal/builtin-plugins/gas-analytics/snapshot-cheatcodes.ts
```diff
@@ -316,13 +325,27 @@ export function compareSnapshotCheatcodes(
         seenPreviousEntries.add(key);
         const previousValue = previousSnapshots[name];
         if (previousValue !== currentEntry.value) {
-          changed.push({
+          const change: SnapshotCheatcodeChange = {
             group,
             name,
             expected: Number(previousValue),
             actual: Number(currentEntry.value),
```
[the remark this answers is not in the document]
ChristopherDedominici: This issue was pre-existing, I created this PR on top of this one to address it: https://github.com/NomicFoundation/hardhat/pull/8488

packages/hardhat/src/internal/builtin-plugins/gas-analytics/helpers/utils.ts
```diff
@@ -84,6 +84,20 @@ export function median(values: number[]): number {
...
+ * The division form matches how percentages are computed when reporting
+ * snapshot changes.
+ */
+export function isWithinTolerance(
+  expected: number,
+  actual: number,
+  tolerance: number,
+): boolean {
+  const diff = Math.abs(actual - expected);
+  return diff === 0 || (expected > 0 && (diff / expected) * 100 <= tolerance);
```
schaable: The division can introduce floating-point errors in some cases, causing the function to return `false` for values that are exactly within tolerance:

```typescript
isWithinTolerance(700, 749, 7)         // false even though the difference is 7%
isWithinTolerance(7000, 7049, 0.7)  // false even though the difference is 0.7%
```

I'd avoid the division by cross-multiplying and remove the comment saying the division form is intentional:

```typescript
return diff * 100 <= tolerance * Math.abs(expected);
```

packages/hardhat/src/internal/builtin-plugins/gas-analytics/index.ts
```diff
@@ -19,6 +19,13 @@ const hardhatPlugin: HardhatPlugin = definePlugin({
         description:
           "Check the snapshots match the stored values (Solidity tests only)",
       })
+      .addOption({
+        name: "tolerance",
+        description:
+          "Allowed gas drift percentage for --snapshot-check (Solidity tests only)",
```
schaable: This also applies to snapshot cheatcodes, which aren't necessarily related to gas:

```suggestion
          "Allowed drift percentage for --snapshot-check (Solidity tests only)",
```

.changeset/puny-eels-swim.md
```diff
@@ -0,0 +1,7 @@
+---
+# docs: https://github.com/NomicFoundation/hardhat-website/pull/291
+"@nomicfoundation/hardhat-errors": patch
+"hardhat": patch
+---
+
+Added a `--tolerance` flag to Solidity test gas snapshots.
```
schaable: `--tolerance` is actually an option, and it applies to snapshot cheatcodes as well, not just gas snapshots. I'd also include a brief description of what it does:

```suggestion
Added a `--tolerance` option to `--snapshot-check` in Solidity tests, allowing snapshot values to drift by a given percentage before the check fails.
```

packages/hardhat/src/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -50,6 +51,30 @@ export interface SnapshotCheckResult {
 const runSolidityTests: TaskOverrideActionFunction<
   GasAnalyticsTestActionArguments
 > = async (args, hre, runSuper) => {
+  // Validate the flags before runSuper: it runs the entire test suite, and a
```
schaable, line 54: Good catch 💯

packages/hardhat/src/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -156,14 +176,17 @@ export function logSnapshotResult(
 export async function handleSnapshotCheck(
   basePath: string,
   suiteResults: SuiteResult[],
+  tolerance: number = 0,
```
schaable: This and the other defaults are redundant because the task already defines a default value, so `tolerance` is always set by the time we get here.

packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
```diff
@@ -49,6 +53,10 @@ export interface FunctionGasSnapshotComparison {
   added: FunctionGasSnapshot[];
   removed: FunctionGasSnapshot[];
   changed: FunctionGasSnapshotChange[];
+  // Diffs within the allowed tolerance: they don't fail the check. They are
+  // included in the result for programmatic inspection, but are not shown in
+  // the default console output.
+  tolerated: FunctionGasSnapshotChange[];
```
schaable, line 59: I'm not convinced we should keep this just for test convenience. Claude suggested showing tolerated drift in the output, maybe with something like `"Function gas snapshots: N within tolerance"`, which would give `tolerated` a use outside the tests. Since that would be a design change, I'd check with the team first and open a follow-up issue.

packages/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
```diff
@@ -647,6 +722,147 @@ describe("solidity-test/task-action (override in gas-analytics/index)", () => {
         assert.deepEqual(await readJsonFile(pathB), beforeB);
       });
     });
+
+    describe("with tolerance", () => {
```
schaable, line 739: One case I'd add to `compareSnapshotCheatcodes` is `"0"` -> `"1"` with a large tolerance. Changes from zero should still be reported since the percentage is undefined.

packages/hardhat/src/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
ChristopherDedominici: I'll forward the compliment to Claude 🤖

packages/hardhat/src/internal/builtin-plugins/gas-analytics/index.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8487/changes/22123daa53f026e50e29a8f45dfe223b2c802529

packages/hardhat/src/internal/builtin-plugins/gas-analytics/helpers/utils.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8487/changes/8da9a6edc1adf11a55d4073677d154f123cad561

packages/hardhat/src/internal/builtin-plugins/gas-analytics/function-gas-snapshots.ts
schaable: Doesn't need to be addressed right now.

packages/hardhat/src/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8487/changes/2f902fb72eb0d1ef8954bbf2be75da1b8a345bb5

packages/hardhat/test/internal/builtin-plugins/gas-analytics/tasks/solidity-test/task-action.ts
ChristopherDedominici: Done: https://github.com/NomicFoundation/hardhat/pull/8487/changes/e31fa88f0a9c1857514cd8527bf159355fd042d8

schaable (approved):
LGTM!

kanej:
Released as part of `hardhat@3.13.0`: https://github.com/NomicFoundation/hardhat/releases/tag/hardhat%403.13.0

PR #8491, opened 2026-08-06
rename `hardhat-solx` to `hardhat-slang-solx`

### LINKED ISSUE #8498: rename hardhat-solx to hardhat-slang-solx
The solx compiler backend is part of the broader slang compiler effort, we are renaming the plugin to reflect that reality.

The hardhat-solx plugin becomes hardhat-slang-solx.

kanej:
The solx compiler backend is part of the broader slang compiler effort, we are renaming the plugin to reflect that reality.

The `hardhat-solx` plugin becomes `hardhat-slang-solx`.

In addition to variable, interface and other code level renames, the following user facing changes have been made:

1. Compiler type discriminator (`type: "solx"` → `type: "slangSolx"`)
2. Build profile name (the plugin's config validation requires a profile named the same as the type)
3. Error category (`HARDHAT_SOLX` → `HARDHAT_SLANG_SOLX`) which is safe as the plugin keeps the same category range
4. the top-level config namespace (`HardhatUserConfig.solx` → `.slangSolx`)

```ts
solidity: {
  profiles: {
    slangSolx: {
      type: "slangSolx",
      version: "0.8.34"
    }
  }
}
```

## Review

> [!TIP]
> Review a commit at a time

The commits initially do file level renames, then code level changes (e.g. interfaces and constants), finally config entries and key values.

## Manual testing

Set up `packages/example-project` to compile and test under the renamed plugin:

```shell
cd packages/example-project
pnpm add -D @nomicfoundation/hardhat-slang-solx@workspace:^
```

Add to `hardhat.config.ts`:

```ts
import hardhatSlangSolx from "@nomicfoundation/hardhat-slang-solx";


export default defineConfig({
  // ...

  // Update plugins
  plugins: [..., hardhatSlangSolx]

  // Update profiles
  profiles: {
    // ...
    slangSolx: {
      compilers: [
        { version: "0.8.22" },
        { version: "0.7.1" },
        { version: "0.8.26" },
        { version: "0.8.33" },
      ],
      overrides: {
        "contracts/Rocket.sol": { type: "slangSolx", version: "0.8.34" },
        "contracts/FailingContract.sol": {
          type: "slangSolx",
          version: "0.8.34",
        },
      },
    },
  }
});
```

Then:

```shell
pnpm hardhat build contracts/Counter.sol --build-profile slangsolx
# Compiled 1 Solidity file with slang-solx 0.1.7 (Solidity 0.8.34) (evm target: osaka)

pnpm hardhat test mocha --build-profile slangsolx
```


packages/hardhat/src/internal/builtin-plugins/solidity-test/edr-artifacts.ts
```diff
@@ -18,7 +18,7 @@ export interface EdrArtifactWithMetadata {
 }
 
 export const BUILD_INFO_FORMAT: RegExp =
-  /^solc-(?<major>\d+)_(?<minor>\d+)_(?<patch>\d+)(?:-(?<compilerType>[a-zA-Z][a-zA-Z0-9]*))?-[0-9a-fA-F]*$/;
+  /^solc-(?<major>\d+)_(?<minor>\d+)_(?<patch>\d+)(?:-(?<compilerType>[a-zA-Z][a-zA-Z0-9-]*))?-[0-9a-fA-F]*$/;
```
kanej: This allows us to have a compiler type with a hyphen, so `slang-solx`, so the build info can be `solc-0_8_34-slang-solx-abc123`.

But it depends on us keeping this structure: `solc-${version}-${compilerType}-${jobHash}`. I can see this being a hostage to fortune if we are forced to make a change and need fallback code for the `slang-solx` users.

Alternatives would be replacing `-` with another character like `_`, when writing to file.

packages/hardhat-errors/src/descriptors.ts
```diff
@@ -333,11 +333,11 @@ export const ERROR_CATEGORIES: {
       },
     },
   },
-  HARDHAT_SOLX: {
+  HARDHAT_SLANG_SOLX: {
```
kanej, line 336: I think this is safe. We have not released the plugin, so as long as the range remains the same, we won't affect other error descriptions.

packages/hardhat-slang-solx/src/type-extensions.ts
```diff
@@ -13,23 +13,26 @@ declare module "hardhat/types/solidity" {
 
 declare module "hardhat/types/config" {
   export interface SolidityCompilerTypeDefinitions {
-    solx: true;
+    // eslint-disable-next-line @typescript-eslint/naming-convention -- the compiler type discriminator is kebab-case to match the plugin name
+    "slang-solx": true;
```
kanej: The hyphen here has an unexpected knock on because the compiler type forms part of the file name of artifact outputs.

packages/hardhat-slang-solx/src/type-extensions.ts
```diff
@@ -13,23 +13,26 @@ declare module "hardhat/types/solidity" {
...
 
   export interface SlangSolxSolidityCompilerUserConfig extends CommonSolidityCompilerUserConfig {
-    type: "solx";
+    type: "slang-solx";
   }
 
   export interface SolidityCompilerUserConfigPerType {
-    solx: SlangSolxSolidityCompilerUserConfig;
+    // eslint-disable-next-line @typescript-eslint/naming-convention -- the compiler type discriminator is kebab-case to match the plugin name
+    "slang-solx": SlangSolxSolidityCompilerUserConfig;
```
kanej: This is a usability question. The property in hardhat config will be `"slang-solx"` requiring escaping as a string to support the `-`. On the whole we would try and avoid that in config properties in JS/TS.

kanej:
The `hardhat-node-test-reporter` fail is likely a Node 26 change that is affecting our output. We should tackle that separately to this PR.

packages/hardhat-slang-solx/src/type-extensions.ts
marianfe: @kanej what if we go with slangSolx in the code?

marianfe (approved):
looks good to me, one comment that we can discuss offline, thanks a lot!

packages/hardhat-slang-solx/src/internal/hook-handlers/config.ts
```diff
@@ -264,34 +271,34 @@ export async function validateResolvedConfig(
 ): Promise<HardhatConfigValidationError[]> {
   const errors: HardhatConfigValidationError[] = [];
 
-  // Check that the user defined a "solx" build profile
-  if (resolvedConfig.solidity.profiles.solx === undefined) {
+  // Check that the user defined a "slang-solx" build profile
+  if (resolvedConfig.solidity.profiles["slang-solx"] === undefined) {
```
kanej: We are pinning the expected profile name here. If you don't have a "slang-solx" profile, we error. If you put a "slang-solx" type into a different profile, we error as well.

PR #8500, opened 2026-08-11
Print fuzz test counterexample calldata as hex

SnowingFox:
## Problem

The calldata of fuzz test counterexamples is printed as a comma-separated list of bytes:

```
Counterexample:
  calldata: 62,32,51,179,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
  args: 0
```

instead of the proper hex string `0x3e2033b300000000000000000000000000000000000000000000000000000000`.

## Root cause

In `packages/hardhat/src/internal/builtin-plugins/solidity-test/reporter.ts`, counterexample values are only hex-formatted when `Buffer.isBuffer(value)` is true. EDR returns the counterexample `calldata` (and the `sender`/`address` fields) as a `Uint8Array`, which is not a `Buffer`, so `Buffer.isBuffer` returns false and the raw bytes fall through to default string interpolation (`62,32,51,...`).

## Fix

Check `value instanceof Uint8Array` instead of `Buffer.isBuffer(value)`. `Buffer` is a subclass of `Uint8Array`, so this covers both cases and delegates to `bytesToHexString` to print a `0x`-prefixed hex string. The same pattern (`value instanceof Uint8Array`) is already used elsewhere in this file.

## Test

Added a reporter test that feeds a failing test result whose counterexample has a `Uint8Array` `calldata` and asserts the calldata is printed as `0x3e2033b3...` instead of the raw byte list.

### DIFF
--- a/.changeset/large-grapes-rest.md
+++ b/.changeset/large-grapes-rest.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed the format of fuzz test counterexamples: byte array values such as `calldata`, `sender` and `address` are now printed as hexadecimal strings (e.g. `0x3e2033b3...`) instead of comma-separated byte lists.
--- a/cspell.dictionary.txt
+++ b/cspell.dictionary.txt
@@ -125,6 +125,7 @@ supath
 synthetix
 Synthetix
 triaging
+trippable
 tset
 txts
 ufixed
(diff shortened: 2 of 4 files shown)


kanej (approved):
Thanks @SnowingFox!

PR #8501, opened 2026-08-11
Support better gas estimation to avoid internal OOG errors

### LINKED ISSUE #8399: Support better gas estimation to avoid internal OOG errors
The current `eth_estimateGas` estimates the minimum gas for the top-level call to succeed. For contracts that make non-failing calls to other contracts (e.g. try/catch, unchecked low-level call), the estimate lets the top-level call succeed while an internal call fails with out-of-gas (OOG): valid for the outer call, broken for the inner one.

Hardhat should offer an opt-in estimation mode that avoids internal OOG failures in non-failing calls, ideally without a separate second estimation pass. The default mode should change to the new behaviour.

## EDR

This new gas estimation mode, `NoInternalOutOfGas`, has been implemented in EDR. The normal estimation runs a binary search to find no top OOG. The new mode adds a further condition to the binary search, that no sub-call ends in an out of gas.

When the new `NoInternalOutOfGas` mode is used, and the transaction would hit an out-of-gas error even when executed with the maximum gas available to the request, EDR returns a new `TransactionFailure`, not a fallback value.

## Hardhat

Implementing this feature on the Hardhat side means wiring through `gasEstimationMode` from Hardhat config on the EDR provider to EDR, then integrating the new error handling.

### New EDR Network config option

The new config option should be added to `EdrNetworkUserConfig`:

```ts
export interface EdrNetworkUserConfig {
  type: "edr-simulated";
  // ...
  gasEstimationMode?: "topLevelSuccess" | "noInternalOutOfGas";
}
```

It will be required in `EdrNetworkConfig` with a resolved default of "noInternalOutOfGas" (the new policy). The Zod validations will need enhanced to cover the new options.

The config convert-to-edr.ts will need enhanced to map down to the EDR enumerations.

#### eth_estimateGas

The new `TransactionFailure` error for unavoidable internal OOG errors will need to be expressed at the hardhat level. The error message should be descriptive enough so the end-user understands that an internal call runs out of gas regardless of the gas limit, so no reliable estimate exists.

#### Automatic gas

When a network's gas is set to `auto` (the default), Hardhat estimates using `eth_estimateGas`; the user did not request the estimate so the new error would be a surprise.

In this case the new error should trigger a fallback to `resolveDefaultTransactionGasLimit` not the block-gas-limit. That helper is aware of the proper caps (e.g. Osaka).

## Testing

Tests should be added to cover the config validation and resolution.

An integration test should be added to cover the new TransactionFailure case. If one doesn't exist covering the error reporting for internal OOG calls for `topLevelSuccess`, that should be covered as well.

Manual tests should be run to check the experience of users deving against contracts with internal calls that OOG, checking the errors in both modes.

## Documentation

The config options need to be added to the Hardhat website. We should explain the new algorithm for `noInternalOutOfGas` tersely.

ChristopherDedominici:
Fixes https://github.com/NomicFoundation/hardhat/issues/8399

Link to website PR: https://github.com/NomicFoundation/hardhat-website/pull/292


packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers-array.ts
```diff
@@ -63,10 +63,26 @@ export async function createHandlersArray<
...
+    // configured with, so the fallback always matches the provider's actual
+    // default. Providers that don't expose it (e.g. HTTP connections) get no
+    // fallback, and the handler surfaces the estimation error instead of
+    // guessing a limit that may not suit the remote network.
+    const provider = networkConnection.provider;
+    const fallbackGas =
+      "defaultTransactionGasLimit" in provider &&
+      typeof provider.defaultTransactionGasLimit === "bigint"
+        ? provider.defaultTransactionGasLimit
+        : undefined;
```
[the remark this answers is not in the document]
ChristopherDedominici: `in` narrows unlisted properties since TS 4.9, so this compiles as-is (verified with the repo's typescript 6.0.3) and the suggested cast would be weaker than the runtime check we already do.

packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/gas/multiplied-gas-estimation.ts
```diff
@@ -7,21 +7,30 @@ import {
...
 } from "@nomicfoundation/hardhat-utils/hex";
 
+import { InternalCallOutOfGasError } from "../../../provider-errors.js";
+
 /**
  * This class handles gas estimation for transactions by applying a multiplier to the estimated gas value.
  * It requests a gas estimation from the provider and multiplies it by a predefined gas multiplier, ensuring the gas does not exceed the block's gas limit.
- * If an execution error occurs, the method returns the block's gas limit instead.
+ * If the estimation fails because an internal call runs out of gas regardless of the gas limit, the method returns the fallback gas limit, if one was provided.
+ * If any other execution error occurs, the method returns the block's gas limit instead.
```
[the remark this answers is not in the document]
ChristopherDedominici: https://github.com/NomicFoundation/hardhat/pull/8501/changes/37a658960d392fb060e5560e0faae2fb757de665

packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/gas/multiplied-gas-estimation.ts
```diff
@@ -63,25 +107,31 @@ export abstract class MultipliedGasEstimation {
...
+
+    assertHardhatInvariant(
+      typeof block === "object" &&
+        block !== null &&
+        "gasLimit" in block &&
+        typeof block.gasLimit === "string",
+      `the ${blockTag} block should have a gasLimit`,
+    );
+
+    return hexStringToNumber(block.gasLimit);
```
schaable, line 140: this will fail for anything above `Number.MAX_SAFE_INTEGER`. Consider using `hexStringToBigInt`

.changeset/bold-geckos-agree.md
```diff
@@ -0,0 +1,6 @@
+---
+# docs: https://github.com/NomicFoundation/hardhat-website/pull/292
+"hardhat": minor
+---
+
+Add `gasEstimationMode` config option to EDR networks, featuring a `"noInternalOutOfGas"` mode to prevent misleading `eth_estimateGas` results when internal calls run out of gas.
```
schaable, line 6:
```suggestion
Added the `gasEstimationMode` config option to EDR-simulated networks, defaulting to `"noInternalOutOfGas"`: `eth_estimateGas` now returns a gas limit that also keeps calls internal to the transaction from running out of gas, and fails with a descriptive error when no gas limit can prevent it. Set it to `"topLevelSuccess"` for the previous behavior, where only the top-level call had to succeed.
```

packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/gas/multiplied-gas-estimation.ts
```diff
@@ -52,6 +70,32 @@ export abstract class MultipliedGasEstimation {
...
+      // already does, so reading it is what keeps the fallback mineable. It
+      // caps even on networks that don't enforce a limit, since their headers
+      // still report a value: that only lowers a still-mineable fallback,
+      // whereas skipping the cap where a limit is enforced would get the
+      // transaction rejected.
+      if (
+        error instanceof InternalCallOutOfGasError &&
+        this.#fallbackGas !== undefined
+      ) {
+        const blockGasLimit = BigInt(await this.#fetchBlockGasLimit("pending"));
```
schaable, line 99: If the network doesn't enforce a `blockGasLimit`, this will unnecessarily cap the estimation. Is this the expected behavior?

packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/gas/multiplied-gas-estimation.ts
ChristopherDedominici: True in principle, `hexStringToNumber` throws above 2^53-1, but it's pre-existing code, unreachable on any real chain's gas limit, and fixing it properly means converting the whole estimation path (including line 61 and the float multiplier math) to bigint, not just that one line. So considering that no chain is at risk, I"d skip

packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/gas/multiplied-gas-estimation.ts
ChristopherDedominici: Done here: https://github.com/NomicFoundation/hardhat/pull/8501/changes/b394f586b5b01925cc82663a5b3bfcc7b59edd48

packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers-array.ts
```diff
@@ -63,10 +63,34 @@ export async function createHandlersArray<
...
 
   if (networkConfig.gas === undefined || networkConfig.gas === "auto") {
+    // On EDR networks, if the gas estimation fails because an internal call
+    // runs out of gas regardless of the gas limit, we fall back to the same
+    // default transaction gas limit that the network uses for requests
+    // without a gas field. The EdrProvider exposes the exact value it was
+    // configured with, so the fallback always matches the provider's actual
+    // default. Providers that don't expose it (e.g. HTTP connections) get no
+    // fallback, and the handler surfaces the estimation error instead of
+    // guessing a limit that may not suit the remote network.
```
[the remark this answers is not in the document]
ChristopherDedominici: This is expected

packages/hardhat/src/internal/builtin-plugins/network-manager/request-handlers/handlers/gas/multiplied-gas-estimation.ts
```diff
@@ -52,6 +73,34 @@ export abstract class MultipliedGasEstimation {
...
+        error instanceof InternalCallOutOfGasError &&
+        this.#fallbackGas !== undefined
+      ) {
+        if (!this.#isBlockGasLimitEnforced()) {
+          return numberToHexString(this.#fallbackGas);
+        }
+
+        const blockGasLimit = BigInt(await this.#fetchBlockGasLimit("pending"));
+
+        return numberToHexString(min(this.#fallbackGas, blockGasLimit));
```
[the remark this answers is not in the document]
ChristopherDedominici: See comment: https://github.com/NomicFoundation/hardhat/pull/8501#discussion_r3863471413

schaable (approved):
Looks good! I'd still update the changeset, and I'll leave the copilot comments for you to review.

PR #8510, opened 2026-08-13
feat(hardhat-verify): support an optional apiKey for Blockscout

nikitosing:
Blockscout supports both the keyless per-instance API and the multichain Pro API, which requires an API key. Mirror what the Etherscan provider already does, while keeping the key optional so instances that don't require one continue working unchanged.

- `verify.blockscout.apiKey` accepts a string or a config variable, and resolves to an optional `ResolvedConfigurationVariable`.
- `Blockscout.resolveConfig`/`create` forward the resolved key to the instance, and the key is sent as the `apikey` query param of the `getsourcecode`, `verifysourcecode` and `checkverifystatus` requests. It's omitted entirely when no key is configured.
- An explicitly configured empty key throws EXPLORER_API_KEY_EMPTY, as it does for Etherscan.

Unlike Etherscan, the chain id isn't sent as a query param: the Blockscout instance is addressed by the `apiUrl` of its chain descriptor, which already encodes the chain id when the API needs it (e.g. `https://api.blockscout.com/{chainId}/api`).

---

### DIFF
--- a/.changeset/blockscout-api-key.md
+++ b/.changeset/blockscout-api-key.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-verify": minor
+---
+
+Added support for an optional `apiKey` in the `verify.blockscout` config, which is sent as the `apikey` query param of the Blockscout verification requests. This is required by Blockscout instances that don't expose a keyless API, like the Pro API.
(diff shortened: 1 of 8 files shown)


schaable (approved):
LGTM!

kanej:
This has been released: https://github.com/NomicFoundation/hardhat/releases/tag/%40nomicfoundation%2Fhardhat-verify%403.1.0

PR #8514, opened 2026-08-13
Fix segment-aware containment checks in the Solidity resolver

### LINKED ISSUE #8513: Solidity resolver allows relative imports to cross package boundaries through prefix collisions
### Version of Hardhat

3.13.0 / current `main`

### What happened?

The Solidity dependency resolver checks whether a relative import stays
within its package using a lexical prefix comparison:

```ts
if (!directImport.startsWith(from.package.inputSourceNameRoot)) {
  // ILLEGAL_RELATIVE_IMPORT
}
```

This comparison is not path-segment-aware. For the Hardhat project package, the input source name root is project, so a normalized source name beginning with project2 is incorrectly considered to be inside project.

For example, given these filesystem paths:

```
contracts/sub/A.sol
File.sol
```

and this import in contracts/sub/A.sol:

`import "../../../project2/File.sol";`

the importing file has the input source name:

`project/contracts/sub/A.sol`

Normalizing the import produces:

`project2/File.sol`

However:

`"project2/File.sol".startsWith("project") === true`

The import therefore bypasses the package-boundary check.

The resolver subsequently removes the expected package prefix with:

```shell
fileInputSourceName.substring(
  npmPackage.inputSourceNameRoot.length + 1,
);
```

which turns project2/File.sol into /File.sol. The filesystem path resolution logic then resolves this path relative to the package root. If <projectRoot>/File.sol exists, the resolver returns a file equivalent to:

```shell
inputSourceName: project2/File.sol
fsPath:          <projectRoot>/File.sol
package:         project
```

The same physical Solidity file can therefore be represented by multiple source unit names if it is also resolved normally as project/File.sol.

This is not a filesystem traversal vulnerability: the later filesystem validation still confines the resolved file to the package root. It is a resolver correctness and source-unit aliasing bug that can affect dependency graphs, compiler inputs, build IDs, artifacts, caching, and diagnostics.

There are two related lexical-prefix checks:

- project/node_modules2/File.sol is incorrectly classified as being inside project/node_modules.
- resolveProjectFile("/foo/project2/A.sol") passes the initial check for a project rooted at /foo/project. Later validation rejects it, but it is reported as PROJECT_ROOT_FILE_DOES_NOT_EXIST instead of

- PROJECT_ROOT_FILE_NOT_IN_PROJECT.

1. Create contracts/sub/A.sol.
2. Create File.sol at the project root.
3. Import ../../../project2/File.sol from A.sol.
4. Resolve or compile A.sol.

The import is resolved to <projectRoot>/File.sol with the source name project2/File.sol instead of being rejected.

### Search terms

ILLEGAL_RELATIVE_IMPORT, dependency-resolver, relative import, package boundary, inputSourceNameRoot, startsWith, project2

questfever:
Fixes  https://github.com/NomicFoundation/hardhat/issues/8513

## Description

The Solidity resolver used raw `startsWith` checks to determine whether source names and filesystem paths were contained within a package or project root.

These checks treated sibling names with a shared prefix as descendants. For example, `project2/File.sol` was considered to be inside the `project` source name root. After the package prefix was removed, the resolver could map that source name to `<projectRoot>/File.sol`, allowing one physical Solidity file to be represented by an invalid source unit name.

This PR makes the relevant checks path-segment-aware:

- relative imports must remain within the importing package's source-name root;
- `node_modules2` is no longer mistaken for `node_modules`;
- project root files are checked with `path.relative`, so sibling-prefix paths
  are reported as outside the project.

Regression tests cover all three prefix-collision cases.

---

### DIFF
--- a/.changeset/clean-ravens-resolve.md
+++ b/.changeset/clean-ravens-resolve.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed Solidity resolver containment checks for paths whose names share a prefix with the project root, a package source-name root, or `node_modules`.
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts
@@ -58,6 +58,16 @@ import {
 
 const NPM_PACKAGES_WITH_SIMULATED_PACKAGE_EXPORTS = new Set(["forge-std"]);
 
+function isInputSourceNameWithinRoot(
+  inputSourceName: string,
+  inputSourceNameRoot: string,
+): boolean {
+  return (
+    inputSourceName === inputSourceNameRoot ||
+    inputSourceName.startsWith(`${inputSourceNameRoot}/`)
+  );
+}
+
 export class ResolverImplementation implements Resolver {
   readonly #projectRoot: string;
   readonly #npmPackageGraph: RemappedNpmPackagesGraphImplementation;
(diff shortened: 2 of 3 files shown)


questfever:
@alcuadrado Could you please take a look when you have a chance? Thanks!

alcuadrado (approved):

alcuadrado:
Thanks for this pr!

PR #8521, opened 2026-08-14
test: add lidofinance-core & graph-horizon scenarios to solx benchmark

Wodann:

---

Adds `lidofinance-core-solx`, `lidofinance-vaults-solx`, and `graph-horizon-solx` to the solx regression benchmark, plus the harness changes they need. The goal is a three-way comparison — forge vs hardhat+solc vs hardhat+solx — on repos whose Hardhat 3 migrations removed their Foundry setups, so each scenario reinstates the pre-migration `foundry.toml` for its forge cells. Verified end-to-end locally: every scenario command, both dump-script paths, and table rendering produce the expected output.

### The lido split

Lido core (Hardhat 3 migration branch, lidofinance/core#1935, pinned at `NomicFoundation/lido-core@28bc9f7ff`) is the corpus's mixed-era repo: four legacy compiler trees (0.4.24–0.8.9) plus the modern vaults tree, upstream at 0.8.25 via-IR. No single scenario can both measure the whole repo and compare compilers like-for-like, so it's benchmarked twice.

### `lidofinance-core-solx` — whole-repo adoption cost

Every cell compiles the entire source graph:

- The legacy trees ride on upstream's own solc versions in every profile — the solx cells included; that mixed shape is what adopting solx here would actually look like. Source roots are enumerated from `contracts/` at config load, so a new upstream tree can't silently fall out. Only `test/` stays out (harnesses and fixtures the base happens to declare as a source root), on both tools.
- The modern vaults tree moves to the compiler under test at 0.8.34 (preinstall relaxes its exact pragmas to caret ranges).
- `contracts/upgrade` stays on upstream's own 0.8.25 in every cell: `UpgradeVoteScript.sol` compiles via-IR at 0.8.25 but hits a Yul stack-too-deep from solc 0.8.26 on (0.8.28/0.8.30/0.8.32/0.8.34 all reproduce, at every optimizer-runs value). solx 0.1.7 compiles it at 0.8.34 by spilling the stack, but a solx cell with no solc counterpart is not a comparison. This looks like a via-IR regression on production code worth a solc upstream issue — lido hits it the day they leave 0.8.25.
- forge follows through the vendored toml's per-tree `compilation_restrictions` plus one benchmark-added `<=0.8.34` cap (no global `FOUNDRY_SOLC`: one version can't serve trees that predate it).
- **Via-IR cells only.** The vaults tree is IR-only (stack-too-deep in `SRLib`, plus `RefSlotCache`'s struct-array copy to storage — an `UnimplementedFeatureError` in legacy codegen, for solc and solx alike). The legacy/no-opt rows render as annotated FAILs; that failure is the datum.
- **Read the wall clock accordingly:** the upgrade tree's via-IR unit (~40–50 s of solc 0.8.25 work) is the critical path in every cell, so the wall columns converge on shared ballast — the compiler signal lives in the CPU columns. That, too, is the adoption datum.

Local numbers (mean wall / total CPU, s; 2 runs per cell; every cell additionally compiles the same legacy + upgrade ballast):

| pipeline | hardhat + solc 0.8.34 | hardhat + solx 0.1.7 | forge 1.7.1 + solc 0.8.34 |
|---|---|---|---|
| via-IR | 52.6 / 99.7 | **52.1 / 88.3** | 49.2 / 81.7 |
| legacy | ✗ does not compile | ✗ does not compile | — |

Compiler assignment, verified: hardhat cells compile 37 files @0.4.24, 1 @0.6.11, 5 @0.6.12, 37 @0.8.9, 10 @0.8.25, and 91 @0.8.34 (solc or solx per profile); forge matches per tree, with a 34-file 0.8.25 unit and no newer solc fetched.

### `lidofinance-vaults-solx` — like-for-like comparison

The modern vaults tree only (37 files), identical sources to all three toolchains at 0.8.34 (`FOUNDRY_SRC` scopes forge the same way). This is the standard compiler-benchmark shape, and where the lido ratios are read from.

- Via-IR cells only, for the reasons above; legacy/no-opt rows are the same annotated FAILs.
- `LIDO_BENCH_SOURCES=upgrade` redirects the wrapper config at `contracts/upgrade` (19 files, self-contained via local interfaces) at 0.8.34: solx compiles it, solc reproduces the YulException — rendered as its own "via-IR, upgrade tree" pivot row with solc's FAIL annotated. This datum is structurally impossible in the whole-repo scenario.
- The dump script skips this scenario: its solx dump would be a strict subset of `lidofinance-core-solx`'s.

Local numbers (same machine and format as above):

| pipeline | hardhat + solc 0.8.34 | hardhat + solx 0.1.7 | forge 1.7.1 + solc 0.8.34 |
|---|---|---|---|
| via-IR | 29.3 / 38.3 | **6.8 / 24.9** | 26.7 / 26.8 |
| via-IR, upgrade tree | ✗ does not compile | **3.4 / 5.3** | — |
| legacy | ✗ does not compile | ✗ does not compile | — |

### `graph-horizon-solx`

Graph Horizon (`packages/horizon` of `NomicFoundation/graphprotocol-contracts`, pinned to the Hardhat 3 migration stack tip) — the first pnpm-monorepo scenario. Five hardhat cells (there is no solx no-opt cell anywhere in the matrix, by design) plus two forge cells.

- All commands `cd packages/horizon`; an unmeasured prime step builds the workspace dists the hardhat config imports.
- Hardhat cells are seeded from upstream's **production** profile (optimizer 100, via-IR, cancun) at solc 0.8.34 (the only solx-mapped version; ranged `^0.8.27` pragmas need no patching).
- The forge cells use the pre-migration `foundry.toml` from PR #1, whose `[profile.prod]` matches those settings exactly (`FOUNDRY_PROFILE=prod` for the via-IR cell; the legacy cell sets the same optimizer via env on the optimizer-off default profile). The pin's lint-only toml only contributed its remappings block, which forge can't auto-generate.
- Preinstall works around four monorepo quirks: `engines.pnpm` (enforced unconditionally by pnpm), `minimumReleaseAge: 7200` (would reject Verdaccio's fresh publishes), version-exact `rocketh` patches orphaned by lockfile re-resolution, and the lockfile itself (its npmjs hardhat pin would otherwise satisfy `^3.11.0` and dodge the local build).

Local numbers (mean wall-clock time / total CPU time; 2 runs per cell):

| pipeline | hardhat + solc 0.8.34 | hardhat + solx 0.1.7 | forge 1.7.1 + solc 0.8.34 |
|---|---|---|---|
| via-IR | 292.8 / 320.9 s | **25.3 / 154.1 s** | 291.9 / 318.5 s |
| legacy | 38.9 / 44.0 s | 32.7 / 161.1 s | 31.8 / 34.7 s |

### Harness changes

- Scenarios can declare `workdir` (schema + types): `bench:dump-standard-json` now compiles inside that package directory instead of the repo root. Benchmark commands still run at the repo root and `cd` themselves.
- A dump variant that fails to compile must be declared in `EXPECTED_DUMP_FAILURES` (keyed `<scenario>|<file>`; currently just lido's plugin-mandated but uncompilable legacy solx profile). Declared skips land in `manifest.json` as `skippedVariants` with the partial dump removed, so the corpus says which holes are expected; anything undeclared rethrows and aborts before the manifest is written, so the publish step's completeness gate does its job. `DUMP_SKIPPED_SCENARIOS` additionally skips whole scenarios whose dumps would duplicate another's (`lidofinance-vaults-solx`), and the out dir is created up front so an all-skipped selection still writes a valid manifest.
- `render-solx-tables` grows an `upgrade` cell flag and a "via-IR, upgrade tree" pivot row for the vaults scenario's PASS/FAIL datum (solc's FAIL is a CELL_NOTES annotation; the parity-fallback footnote moves from ² to ³).

### Caveats

- The benchmarked modern trees compile at 0.8.34 while upstream ships 0.8.25 (lido) / 0.8.35 (horizon) — the usual price of solx's single mapped version; the cells pin `evmVersion: cancun`, so the delta is compiler-internal.
- All scenarios need `EVM_DISABLE_MEMORY_SAFE_ASM_CHECK=1`: solx can only auto-spill stack-too-deep when no assembly block looks memory-unsafe, and both repos have unannotated assembly (aave precedent).
- Horizon's solx **legacy** cell peaks at ~10 GB RSS: solx's legacy pipeline requests `legacyAssembly` output from libsolc for every contract, and building those JSON trees for 91 Foundry test units (dependencies embedded recursively) dominates the main process's memory — worth watching on the shared runner, since it scales with test-suite size. A fix is in flight (solx-solidity#174 plus a solx-side change; 9.6 → 3.9 GB on this exact dump, byte-identical output), so the number should improve in a future pinned release.
- The reinstated foundry.toml files are vendored snapshots; if the pins move, the fail-loud preinstall guards catch the drift.
- Workflow timeout raised 180 → 210 min; horizon's two ~5-min via-IR compilers (solc and forge) dominate the added cost, and the widened lido pair adds roughly ten minutes against ample headroom (the last full run finished in 1h53m).
- The gas-compare and replay subsets deliberately stay on the existing scenarios.


end-to-end/graph-horizon-solx/foundry.toml
```diff
@@ -0,0 +1,41 @@
...
+out = 'build'
+libs = ["node_modules"]
+remappings = [
+    "@graphprotocol/=node_modules/@graphprotocol/",
+    "forge-std/=node_modules/forge-std/src/",
+]
+test = 'test'
+cache_path = 'cache_forge'
+fs_permissions = [{ access = "read", path = "./" }]
+solc_version = '0.8.35'
```
nebasuke, line 17: This is correct (based on fork), but confusing. A comment that this is overridden later would help here.

end-to-end/graph-horizon-solx/hardhat.config.solx.ts
```diff
@@ -0,0 +1,105 @@
...
+export default {
+  ...base,
+  plugins: [...base.plugins, hardhatSolx],
+  // The plugin only allows type: "solx" in the profile named "solx"; this
+  // benchmark needs a second solx profile ("solx-via-ir") for the viaIR sweep,
+  // so opt out of that guard. Throwaway benchmark scenario, not production.
+  solx: { dangerouslyAllowSolxInProduction: true },
+  solidity: {
+    npmFilesToBuild: base.solidity.npmFilesToBuild,
+    profiles: {
```
nebasuke, line 81: As discussed offline, let's add solc to the solx profile, rather than cutting part of the source graph, to get more fair benchmarks.

end-to-end/lidofinance-core-solx/foundry.toml
```diff
@@ -0,0 +1,67 @@
+[profile.default]
```
nebasuke, line 1: This could use a comment similar to graph-horizon-solx to say where it comes from: ( lido-core@242beb163),

scripts/benchmark/dump-standard-json.ts
```diff
@@ -177,20 +181,36 @@ function main(): void {
...
+            ...definition.env,
+            ...env,
+            SOLX_STANDARD_JSON_DEBUG: dumpPath,
+          },
+        });
+      } catch {
+        // A variant may be legitimately uncompilable (lidofinance-core-solx's
+        // vaults tree is IR-only, so its plugin-mandated legacy "solx" profile
+        // can't build). Skip the variant — leaving it out of the manifest —
+        // instead of aborting the run, which would lose every other
```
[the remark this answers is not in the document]
nebasuke: Agree with the direction, and I would go one further than an allowlist: also record the skip in `manifest.json` (e.g. a `skippedVariants: {file: reason}` per scenario). The manifest's whole job is "this dump set is complete"; a silent hole and an expected hole are different things, and downstream consumers (the corpus publish, solx-side benchmarks) cannot tell them apart from a missing key. An explicit expected-failure set in the script (mirroring the `CELL_NOTES` pattern in `render-solx-tables.ts`) plus rethrow-on-unexpected keeps regressions loud; the manifest entry makes the corpus self-describing either way.

end-to-end/lidofinance-core-solx/hardhat.config.solx.ts
```diff
@@ -0,0 +1,163 @@
...
+  [key: string]: unknown;
+};
+
+// Seed every profile from upstream's modern-tree entry (optimizer runs 200,
+// viaIR: true, evmVersion cancun — cancun is in solx's supported set).
+const modernEntry = base.solidity.compilers.find((c) => c.version === "0.8.25");
+if (modernEntry === undefined) {
+  throw new Error(
+    "lidofinance-core-solx: no 0.8.25 compiler entry in the base config — the pinned commit may have changed",
+  );
```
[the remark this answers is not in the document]
nebasuke: These wrapper configs are throwaway benchmark fixtures that run inside the pinned external repo's checkout, not hardhat packages — `HardhatError` and the invariant helper are not importable there without making the hardhat workspace a dependency of the scenario, and the existing solx scenarios' wrapper configs throw plain `Error` the same way. The loud message plus refusing to run is the point; standard error classification buys nothing here.

end-to-end/graph-horizon-solx/hardhat.config.solx.ts
```diff
@@ -0,0 +1,105 @@
...
+// the solc profile. The solx optimization level (-O1) and DWARF debug info both
+// come from the hardhat-solx plugin defaults: DWARF is force-emitted, so solx
+// maps sources just as solc does (Hardhat force-emits solc sourceMaps), keeping
+// the comparison apples-to-apples. We intentionally don't override the optimizer
+// here so the benchmark measures the realistic plugin-default config.
+const baseSettings = base.solidity.profiles.production?.settings;
+if (baseSettings === undefined) {
+  throw new Error(
+    "graph-horizon-solx: no production profile in the base config — the pinned commit may have changed",
+  );
```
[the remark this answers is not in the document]
nebasuke: Same as on the lido config: these wrapper configs are throwaway benchmark fixtures running inside the pinned external repo's checkout — `HardhatError` is not importable there without making the hardhat workspace a scenario dependency, and the existing solx scenarios throw plain `Error` the same way. Loud message + refusal to run is the intended behavior.

nebasuke (approved):
The memory regression is slightly different than your Claude determined. See the draft fix here: https://github.com/NomicFoundation/solx-solidity/pull/174

PR generally looks good, but let's add solc compiler(s) the the solx profile where appropriate so we don't cut down the source trees.

nebasuke:
I've left AI assisted comments on Copilot. Feel free to use/ignore.

scripts/benchmark/dump-standard-json.ts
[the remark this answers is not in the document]
nebasuke: Done both ways: variants are skipped only when declared in `EXPECTED_DUMP_FAILURES` (keyed `<scenario>|<file>`, mirroring `CELL_NOTES`) — currently just lido's legacy solx dump — and anything else rethrows with the scenario/file and a pointer to that table, so the run aborts before `writeManifest` and the publish step's manifest gate does its job. Declared skips land in `manifest.json` as `skippedVariants: { file: reason }`, so the corpus says which holes are expected. Verified on the real scenario: the expected skip is recorded and the partial dump removed; flipping the key out of the table aborts the run with no manifest written.

end-to-end/lidofinance-core-solx/foundry.toml
nebasuke: Added a header naming `NomicFoundation/lido-core@242beb163` and stating that the file is verbatim except for one marked `compilation_restrictions` entry — the widened source graph needs forge to pick a compiler per tree instead of a global `FOUNDRY_SOLC`, and the added entry caps solc at 0.8.34 so ranged pragmas resolve where the hardhat cells put them. (The new `lidofinance-vaults-solx` copy carries its own header; that one is verbatim, since its forge cells scope via `FOUNDRY_SRC`/`FOUNDRY_SOLC` instead.)

end-to-end/graph-horizon-solx/foundry.toml
nebasuke: Commented: kept at the vendored 0.8.35, overridden by `FOUNDRY_SOLC=0.8.34` in every forge invocation because that's the only version in hardhat-solx's map.

end-to-end/graph-horizon-solx/hardhat.config.solx.ts
nebasuke: Done for lido — the solx profiles now carry upstream's 0.4.24–0.8.9 compilers next to the solx entry, the `paths.sources` cut is gone, and forge follows through the toml's per-tree restrictions. `contracts/upgrade` needed one more twist: `UpgradeVoteScript.sol` builds via-IR at upstream's 0.8.25 but hits a Yul stack-too-deep from solc 0.8.26 on (0.8.28/0.8.30/0.8.32/0.8.34, every optimizer runs value), so it can't follow the modern tree to 0.8.34 — instead it rides in every cell on upstream's own 0.8.25 (preinstall leaves its exact pragmas alone, which routes both hardhat and forge there). solx 0.1.7 does compile it at 0.8.34, but a solx cell with no solc counterpart isn't a comparison.

One consequence: that tree's via-IR unit is the slowest in every cell, so the whole-repo scenario's wall clocks converge on shared ballast — the compiler difference shows in the CPU columns. That's the honest adoption datum, but it drowned the like-for-like question, so lido is now two scenarios: `lidofinance-core-solx` (whole graph, as above) and `lidofinance-vaults-solx` (identical sources to every compiler at 0.8.34 — the vaults tree, plus a `contracts/upgrade`-at-0.8.34 row where solx compiles and solc's FAIL is the annotated datum). Numbers for both are in the updated PR body.

PR #8522, opened 2026-08-15
fix: remove the Holesky shipped chain descriptor

### LINKED ISSUE #8499: Drop Holesky chain configs
Holešky chain was deprecated on September 2025. Ethereum Foundation announcement [here](https://blog.ethereum.org/2025/09/01/holesky-shutdown-announcement). None of the block explorer listed in https://chainlist.org/chain/17000 is live anymore. 

Let's remove any config associated to the chain.

JasonColapietro:

---

Fixes #8499.

The Ethereum Foundation shut the Holešky testnet down in September 2025
(https://blog.ethereum.org/2025/09/01/holesky-shutdown-announcement). Neither
block explorer we ship for chain id 17000 resolves any more, so the bundled
descriptor is dead config that can only send users somewhere broken. This
removes the 17000 entry from `DEFAULT_CHAIN_DESCRIPTORS`. `L1_CHAIN_TYPE` is
still used by other entries, so nothing else in the file changes.

One test had to move, so it is worth flagging rather than leaving a reviewer to
wonder. `config-resolution.ts`'s "should only override the provided fields" test
used chain 17000 as its fixture for merging a user chain descriptor over a
shipped default, so deleting the default broke it. It now uses OP Sepolia
(11155420), picked for three reasons:

- it still ships a default descriptor, so the merge path is really exercised;
- it carries both an etherscan and a blockscout explorer, matching Holesky's
  shape, so the `blockExplorers` assertion still covers both;
- its default chain type is OPTIMISM, which differs from the `GENERIC_CHAIN_TYPE`
  the fixture overrides it with, so the override assertion still compares two
  different values instead of collapsing into a self-comparison.

Every assertion is preserved as it was; only the identifiers and the fixture's
name string changed. I checked that the migrated test still binds by mutating the
source: removing the `chainType` override branch fails it, and making the default
lookup miss for 11155420 fails it on the `blockExplorers` assertion.

A patch changeset is included.

### DIFF
--- a/.changeset/remove-holesky-chain-descriptor.md
+++ b/.changeset/remove-holesky-chain-descriptor.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Removed the shipped chain descriptor for the Holesky testnet (chain id 17000). The network was shut down in September 2025 and the block explorers bundled for it no longer resolve.
--- a/.changeset/remove-holesky-chain-descriptor.md
+++ b/.changeset/remove-holesky-chain-descriptor.md
@@ -2,4 +2,4 @@
 "hardhat": patch
 ---
 
-Removed the shipped chain descriptor for the Holesky testnet (chain id 17000). The network was shut down in September 2025 and the block explorers bundled for it no longer resolve.
+Remove Holesky testnet (chain id 17000) as a supported chain as it has been shut down.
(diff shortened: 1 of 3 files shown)


.changeset/remove-holesky-chain-descriptor.md
```diff
@@ -0,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Removed the shipped chain descriptor for the Holesky testnet (chain id 17000). The network was shut down in September 2025 and the block explorers bundled for it no longer resolve.
```
kanej:
```suggestion
Remove Holesky testnet (chain id 17000) as a supported chain as it has been shut down.
```

kanej (approved):
Thanks @JasonColapietro!

PR #8526, opened 2026-08-17
Renovate 1 GitHub actions

kanej:
This PR brings Hardhat's GitHub Actions up to date:

1. every `actions/*` action to its current major
2. every pinned third-party action to a current digest
3. the three third-party actions that were still on tags pinned
4. `slackapi/slack-github-action` migrated from v1 to v4

Three actions were being used at **two different versions simultaneously**, which this also fixes:

- `actions/setup-node` at v4 and v5
- `actions/download-artifact` at v4 and v5
- `slackapi/slack-github-action` at v1.26.0 and v2.1.1

### **`actions/*` stay on major tags; third-party actions are pinned by digest.**

This is continuing the existing pattern for the repo, but maybe we should consider completely swapping to SHAs even for github's own actions. This appears to be the EDR move.

### The Slack migration to the v2 API

The break is entirely between v1 and v2: v1 read the webhook URL from a `SLACK_WEBHOOK_URL` environment
variable and inferred what to do with the payload, whereas from v2 the URL is a `webhook` input and the
mode is explicit in `webhook-type`. Everything after that is quiet — v3's only breaking change is its
Node 24 runtime, and v4's is an internal change to proxied fetch configuration.

## Review

> [!TIP]
> Review a commit at a time for easier review

- **Pin the remaining third-party actions** — establishes the convention the rest of the branch follows.
- **`setup-node` → v7** and **`download-artifact` → v8** — each also unifies two versions that were in use
  at once.
- **`upload-artifact` → v7**, **`checkout` → v7**, **`github-script` → v9 and `cache/restore` → v6** —
  version numbers and runtimes only.
- **Refresh the pinned third-party digests** — five of six were behind, one by three majors, which is the
  cost of pinning without a bot to move them.
- **Migrate `slack-github-action` to v4** — the one with a behavioural trap in it.
- **Fix to our own bundle action** - the artifact downloader changed the layout of downloads

### DIFF
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -26,7 +26,7 @@ runs:
       with:
         version: ${{ inputs.pnpm-version }}
     - id: setup-node
-      uses: actions/setup-node@v7
+      uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
       with:
         node-version: ${{ inputs.node-version }}
         cache: ${{ inputs.cache-save == 'true' && 'pnpm' || '' }}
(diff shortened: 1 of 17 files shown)


alcuadrado:
> ### **`actions/*` stay on major tags; third-party actions are pinned by digest.**
> This is continuing the existing pattern for the repo, but maybe we should consider completely swapping to SHAs even for github's own actions. This appears to be the EDR move.

I think it will be better to use the hash everywhere, so that we can enforce it across the org. It can be in a separate pr though.

alcuadrado (approved):

kanej:
I have updated to a hash everywhere. I am going to merge.

PR #8527, opened 2026-08-18
fix: accept hex digit f/F in int, bigint and float CLI argument types

pucedoteth:
## Problem

The `int`, `bigint` and `float` CLI argument types each validate hex input with the same pattern in `packages/hardhat-core/src/internal/core/params/argumentTypes.ts`:

```js
const hexPattern = /^0[xX][\dABCDEabcde]+$/;
```

The character class spells out `ABCDE` and `abcde` — every hex digit except `f`/`F`.

So any hex argument containing an `f` is rejected as invalid:

| value | accepted today |
|---|---|
| `0xa`, `0x0a`, `0xABCDE` | ✅ |
| `0xF`, `0xf` | ❌ |
| `0xff` | ❌ |
| `0xdeadbeef` | ❌ |
| `0xABCDEF` | ❌ |

`f` shows up in about half of all multi-digit hex values, so this rejects a large share of legitimate input — a task taking `--value 0xff` fails with `Invalid value for argument`.

It reads like a hand-enumerated class where the last digit was simply missed.

## Fix

```js
const hexPattern = /^0[xX][\dA-Fa-f]+$/;
```

Applied at all three call sites (`int`, `bigint`, `float`).

Worth noting this matches what Hardhat 3 already does — `packages/hardhat/src/internal/core/arguments.ts` on `main` defines `VALID_HEX_PATTERN = /^0[xX][\dA-Fa-f]+$/`. The rewrite got it right; this brings v2 in line.

## Tests

Extended the existing `argumentTypes` suite rather than adding a new one — five positive cases per type (`0xF`, `0xf`, `0xff`, `0xdeadbeef`, `0xABCDEF`) and one negative (`0xg`) to pin down that widening the class doesn't start accepting non-hex.

Every one of the positive cases fails on `v2` today; the negative cases and the existing `0xa` / `0x0a` / `x123` assertions are unchanged.

## Note

I originally raised this against the wrong base by mistake. This is the same fix rebuilt cleanly on the current `v2` head — `packages/hardhat-core` only exists on this branch, and the bug is still present in all three places there.

### DIFF
--- a/.changeset/hex-f-cli-argument-types.md
+++ b/.changeset/hex-f-cli-argument-types.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+Fixed the `int`, `bigint` and `float` CLI argument types rejecting hexadecimal values containing the digit `f`/`F` (e.g. `0xff`, `0xdeadbeef`).
--- a/packages/hardhat-core/src/internal/core/params/argumentTypes.ts
+++ b/packages/hardhat-core/src/internal/core/params/argumentTypes.ts
@@ -86,7 +86,7 @@ export const int: CLIArgumentType<number> = {
   name: "int",
   parse: (argName, strValue) => {
     const decimalPattern = /^\d+(?:[eE]\d+)?$/;
-    const hexPattern = /^0[xX][\dABCDEabcde]+$/;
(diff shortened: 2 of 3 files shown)


kanej (approved):
Thanks @pucedoteth

PR #8536, opened 2026-08-18
fix: ignore insignificant whitespace in solidity version pragmas

### LINKED ISSUE #8535: Solidity files with whitespace inside the version pragma fail with HHE909, but compile fine with solc
Encountered while trying to set up Sourcify for `hardhat-slang-solx` using Hardhat 3. It seems that Hardhat 3 using `solidity-analyzer` is more strict in parsing pragmas than `solc` itself. 

# Claude summary

### Version of Hardhat

3.13.0

### What happened?

Hardhat 3 refuses to build a file whose version pragma contains whitespace inside the version literal, e.g.:

```solidity
pragma solidity ^ 0.8 .0;
```

Building fails with:

```
Error HHE909: Failed to create compilation job for file ".../Repro.sol" using the build profile "default".
No solc version enabled in this profile is compatible with this file.
```

solc itself accepts this pragma: the version expression is tokenized by the compiler, so whitespace between tokens is insignificant. The same file compiles without errors when passed directly to `solc --standard-json` (verified with 0.8.34).

This is real on-chain code, not a hypothetical: we hit it sweeping Hardhat over Sourcify-verified contracts. In a corpus of 112,587 verified solc 0.8.34 contracts, 57 (~0.05%) contain whitespace inside a pragma version literal — all of them perfect runtime matches, i.e. solc compiled them as-is. Example: chain 97 (BSC testnet), `0xe7979a289e7864f2d90bd09542450c8c5e740d9f`.

**Root cause**

`@nomicfoundation/solidity-analyzer` extracts the pragma expression verbatim:

```js
analyze("pragma solidity ^ 0.8 .0;\ncontract C {}")
// -> { versionPragmas: ["^ 0.8 .0"], imports: [] }
```

`solc-config-selection.ts` then matches it with npm `semver` (`maxSatisfying`/`satisfies`), which tolerates whitespace after an operator but not inside a version:

```js
semver.satisfies("0.8.34", "^ 0.8.0")  // true
semver.satisfies("0.8.34", "^ 0.8 .0") // false — no error, just no match
```

So every enabled compiler silently fails to match and version selection reports HHE909 with a misleading "no compatible solc version" message.

**Expected behavior**

The pragma should match any solc version that solc itself would accept for the file. Since solc tokenizes the version expression, normalizing the extracted pragma before semver matching (e.g. collapsing whitespace around `.` inside version literals, or re-serializing the token stream) would close the gap. That could live either in solidity-analyzer's extraction or at the consumption site in `solc-config-selection.ts`.

Other versions involved: `@nomicfoundation/solidity-analyzer` 0.1.2, npm `semver` 7.8.5, Node.js 24.19.0, Linux x86_64.

`contracts/Repro.sol`:

```solidity
pragma solidity ^ 0.8 .0;
contract Repro {}
```

`hardhat.config.js`:

```js
export default defineConfig({
});
```

`package.json`:

```json
```

`npm install --save-dev hardhat`, then `npx hardhat build` → HHE909.

Baseline showing solc accepts it:

```bash
echo '{"language":"Solidity","sources":{"a.sol":{"content":"// SPDX-License-Identifier: MIT\npragma solidity ^ 0.8 .0;\ncontract C {}"}},"settings":{"outputSelection":{"*":{"*":["evm.bytecode"]}}}}' \
  | solc-0.8.34 --standard-json
# -> no errors, contract compiles
```

### Search terms

HHE909 pragma whitespace version semver no solc version enabled compatible

kcfuler:
Fixes #8535

solc tokenizes a pragma such as `pragma solidity ^ 0.8 .0;` as `^` + `0.8.0`, so the file compiles. Hardhat 3 takes the string `@nomicfoundation/solidity-analyzer` extracts (`"^ 0.8 .0"`) and feeds it to npm `semver` (`satisfies` / `maxSatisfying`). Interior whitespace is significant there, so a configured `0.8.34` does not match and the build reports HHE909.

This normalizes each extracted pragma by collapsing whitespace before every range check. Multiple pragmas are still joined with a space, so the AND between them is unchanged.

On current `main`, the new unit test returns `NO_COMPATIBLE_SOLC_VERSION_WITH_ROOT`. After the change it selects compiler `0.8.34`.

### DIFF
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
@@ -12,6 +12,22 @@ import { intersects, maxSatisfying, satisfies } from "semver";
 
 import { CompilationJobCreationErrorReason } from "../../../../types/solidity/build-system.js";
 
+
+/**
+ * Solidity tokenizes version pragmas, so whitespace inside a literal is
+ * insignificant (`^ 0.8 .0` means `^0.8.0`). npm `semver` does not treat
+ * interior whitespace that way, and `maxSatisfying` then returns null
+ * (Hardhat HHE909) even when a compatible compiler is enabled.
+ */
+function normalizeVersionPragma(pragma: string): string {
+  return pragma.replace(/\s+/g, "");
+}
+
+function joinVersionPragmas(pragmas: readonly string[]): string {
+  return Array.from(new Set(pragmas.map(normalizeVersionPragma))).join(" ");
+}
+
+
 export class SolcConfigSelector {
   readonly #buildProfileName: string;
   readonly #buildProfile: SolidityBuildProfileConfig;
@@ -60,7 +76,7 @@ export class SolcConfigSelector {
       .map(({ content }) => content.versionPragmas)
       .flat(1);
 
-    const versionRange = Array.from(new Set(allVersionPragmas)).join(" ");
+    const versionRange = joinVersionPragmas(allVersionPragmas);
 
     const overriddenCompiler = this.#buildProfile.overrides[userSourceName];
 
@@ -121,7 +137,7 @@ export class SolcConfigSelector {
     compilerVersions: string[],
     overridden: boolean,
   ): CompilationJobCreationError {
-    const rootVersionRange = root.content.versionPragmas.join(" ");
+    const rootVersionRange = joinVersionPragmas(root.content.versionPragmas);
 
     // This logic is pretty different depending if we are dealing with a config
     // override or not. If we are, we have a single compiler option, so things
@@ -147,7 +163,7 @@ export class SolcConfigSelector {
         dependencyGraph,
       )) {
         const depOwnRange =
-          transitiveDependency.dependency.content.versionPragmas.join(" ");
+          joinVersionPragmas(transitiveDependency.dependency.content.versionPragmas);
 
         if (maxSatisfying(compilerVersions, depOwnRange) === null) {
           return {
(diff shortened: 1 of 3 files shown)


packages/hardhat/src/internal/builtin-plugins/solidity/build-system/solc-config-selection.ts
```diff
@@ -12,6 +12,22 @@ import { intersects, maxSatisfying, satisfies } from "semver";
...
 
+
+/**
+ * Solidity tokenizes version pragmas, so whitespace inside a literal is
+ * insignificant (`^ 0.8 .0` means `^0.8.0`). npm `semver` does not treat
+ * interior whitespace that way, and `maxSatisfying` then returns null
+ * (Hardhat HHE909) even when a compatible compiler is enabled.
+ */
+function normalizeVersionPragma(pragma: string): string {
+  return pragma.replace(/\s+/g, "");
```
official-burak, line 23: `pragma.replace(/\s+/g, "")` also glues comparators together. `semver.validRange(">=0.8.0<0.9.0")` is `null`, and `satisfies("0.8.34", that)` is false.

`pragma solidity >=0.4.22 <0.9.0` is the usual form. That range works today. After this change it becomes `>=0.4.22<0.9.0` and HHE909 comes back for files that were fine. The new test only covers `^ 0.8 .0`.

Only strip spaces inside a version literal (around dots). Keep the space between comparators.

`depOwnRange` below still does `versionPragmas.join(" ")` and never goes through `joinVersionPragmas`.

official-burak (commented):
Stripping every space fixes `^ 0.8 .0` but turns a normal two-comparator pragma into an invalid semver range.

kanej:
Thanks @kcfuler, we have reviewed the underlying issue and determined we do not want to fix the issue as it is really a bug in solc.

PR #8542, opened 2026-08-19
fix: suggest relative import for bare-filename sibling Solidity imports

### LINKED ISSUE #8537: Bare-filename imports of local files get a misleading "invalid npm syntax" error and no suggested fix
Other issue I encountered while trying to set up Sourcify for hardhat-slang-solx using Hardhat 3. It seems that Hardhat 3 cannot import by bare filename which was possible in Hardhat 2. 

I think the Claude suggestion of at least fixing the error message to suggest a slight rewrite (to use ./) is the right one. 

# Claude summary
### Version of Hardhat

3.13.0

### What happened?

Importing a sibling project file by bare filename — the classic Remix/solc
direct-import style —

```solidity
import "Storage.sol";
```

fails with:

```
Error HHE902: There was an error while resolving the import "Storage.sol" from "./contracts/Factory.sol":
You are trying to import an npm file but its syntax is invalid.
```

Two problems with how this surfaces:

1. **The message is misleading and suggests no fix.** The user isn't trying to
   import an npm file — their file sits right next to the importer. For direct
   imports *with* a directory separator, Hardhat already detects the
   local-file case and suggests a remapping
   (`DIRECT_IMPORT_TO_LOCAL_FILE` in
   `build-system/resolver/dependency-resolver.ts`) — but that path explicitly
   bails when the import path contains no `/`, so exactly the least-informed
   users get the worst error. Suggesting the relative form
   (`import "./Storage.sol";`) here would be consistent with the existing UX
   and is likely a small change.

2. **There is no workaround other than editing the source.** Remappings can't
   express a prefix without a `/`: an explicit
   `Storage.sol=contracts/Storage.sol` in `remappings.txt` changes nothing
   (verified). That's irrelevant for a user who can just fix the import, but
   it makes such files impossible to build unmodified — which matters for
   tooling that replays verified/historic sources through Hardhat.

To be clear: this is not a request to make bare-filename imports resolve —
HH3's relative-or-npm import design is deliberate and good. The ask is the
error message/suggestion (1), with (2) as a secondary consideration if
slash-less remapping prefixes are cheap to allow.

Scale data point: sweeping Hardhat over 112,587 Sourcify-verified solc 0.8.34
contracts, 64 (0.06%) use bare-filename imports of local files — mostly small
Remix-style projects, i.e. exactly the audience that won't decode "invalid npm
syntax".

`contracts/Storage.sol`:

```solidity
pragma solidity ^0.8.0;
contract Storage { uint x; }
```

`contracts/Factory.sol`:

```solidity
pragma solidity ^0.8.0;
import "Storage.sol";
contract Factory { Storage s; }
```

`hardhat.config.js`:

```js
export default defineConfig({
});
```

`package.json`:

```json
```

`npm install --save-dev hardhat`, then
`npx hardhat build` → HHE902 "trying to import an npm file but its syntax is
invalid". Adding `Storage.sol=contracts/Storage.sol` to `remappings.txt`
changes nothing.

### Search terms

HHE902 bare filename import npm file syntax invalid direct import local file remapping

official-burak:

Fixes #8537.

Hardhat 3 treats `import "Storage.sol"` (Remix-style sibling import, no `./`) as an npm module. That path is not valid npm syntax, so users get HHE902 ("You are trying to import an npm file but its syntax is invalid") even when `Storage.sol` sits next to the importer.

Direct imports that already contain a `/` already get `DIRECT_IMPORT_TO_LOCAL_FILE` with a remapping suggestion. Bare filenames cannot be expressed as a remapping prefix (those require a trailing `/`), so this change:

1. Detects when the import has no `/` and a sibling file exists.
2. Reports the same local-direct-import error, suggesting `import "./Storage.sol"` instead of the npm-syntax message.

If the sibling file does not exist, the previous invalid-npm-syntax error is unchanged.

Tests cover the resolver result and the formatted error message.

### DIFF
--- a/.changeset/bare-filename-local-import.md
+++ b/.changeset/bare-filename-local-import.md
@@ -1,0 +1,5 @@
+---
+"hardhat": patch
+---
+
+When a Solidity file imports a sibling by bare filename (`import "Storage.sol"`), Hardhat now reports that as a local direct import and suggests `import "./Storage.sol"` instead of a misleading invalid-npm-syntax error.
--- a/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity/build-system/resolver/dependency-resolver.ts
@@ -941,6 +941,25 @@ export class ResolverImplementation implements Resolver {
     from: ResolvedFile;
     importPath: string;
   }): Promise<ImportResolutionError | undefined> {
+    // Bare filenames like `import "Storage.sol"` cannot be expressed as a
+    // remapping prefix (those require a trailing `/`). If a sibling file
+    // exists, suggest the relative form instead of the misleading npm-syntax
+    // error.
+    if (!importPath.includes("/")) {
+      const siblingFsPath = path.join(path.dirname(from.fsPath), importPath);
+
+      if (await exists(siblingFsPath)) {
+        return {
+          type: ImportResolutionErrorType.DIRECT_IMPORT_TO_LOCAL_FILE,
+          fromFsPath: from.fsPath,
+          importPath,
+          suggestedRelativeImport: `./${importPath}`,
+        };
+      }
+
+      return undefined;
+    }
+
     let baseDir = path.dirname(from.fsPath);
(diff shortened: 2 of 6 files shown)


kanej:

PR #8547, opened 2026-08-20
fix(hardhat-viem-assertions): compare the sender address case-insensitively in balancesHaveChanged

Kropiunig:

---

### Problem

`balancesHaveChanged` adds the transaction's gas fee back to the sender's balance, so that a caller only has to assert the value actually transferred. It identifies the sender with a case-sensitive comparison:

```ts
const senderAddress = receipt.from;
// ...
if (address === senderAddress) {
  const senderGasFee = receipt.effectiveGasPrice * receipt.gasUsed;
  balance = balance + senderGasFee;
}
```

`receipt.from` is returned lowercased by viem, but the addresses users pass in are frequently checksummed: `walletClient.account.address` is lowercased, while `walletClient.getAddresses()` returns the EIP-55 form (`0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266`), and hardcoded address literals are conventionally checksummed too.

The sender then never matches, the gas fee is not added back, and this ordinary usage fails:

```ts
const [sender] = await bobWalletClient.getAddresses();

await viem.assertions.balancesHaveChanged(
  bobWalletClient.sendTransaction({
    to: aliceWalletClient.account.address,
    value: 3333333333333333n,
  }),
  [{ address: sender, amount: -3333333333333333n }],
);
```

```
For address "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", expected balance to change by
-3333333333333333 (from 10000000000000000000000 to 9999996666666666666667),
but got a change of -3356589192708333 instead.
```

The reported change is off by exactly the gas fee. The existing tests all pass `account.address`, which is already lowercased, which is why this path was never exercised.

### Fix

Lowercase both sides of the comparison, consistent with how addresses are already compared elsewhere in this package (`src/internal/assertions/emit/core.ts:58-60`, `src/internal/predicates.ts:46`).

### Tests

Added a regression test that passes the checksummed sender address. Running `packages/hardhat-viem-assertions/test/internal/assertions/balances-have-changed.ts`:

- before the fix: 6 pass, 1 fail (the new test, with the error quoted above)
- after the fix: 7 pass, 0 fail

`pnpm eslint` and `pnpm prettier --check` are clean for the changed files.

### DIFF
--- a/.changeset/viem-assertions-checksummed-sender.md
+++ b/.changeset/viem-assertions-checksummed-sender.md
@@ -1,0 +1,5 @@
+---
+"@nomicfoundation/hardhat-viem-assertions": patch
+---
+
+Fixed `balancesHaveChanged` not adding the gas fee back to the sender's balance when the sender address is passed in its checksummed form.
--- a/packages/hardhat-viem-assertions/src/internal/assertions/balances-have-changed.ts
+++ b/packages/hardhat-viem-assertions/src/internal/assertions/balances-have-changed.ts
@@ -29,7 +29,9 @@ export async function balancesHaveChanged<
     hash: resolvedTxHash,
   });
 
-  const senderAddress = receipt.from;
+  // `receipt.from` is returned lowercased by viem, while the user-provided
+  // addresses may be checksummed, so we normalize both sides before comparing.
+  const senderAddress = receipt.from.toLowerCase();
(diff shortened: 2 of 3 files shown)


kanej (approved):
Thanks @Kropiunig, looks good.

PR #8588, opened 2026-08-31
ci(bench): notify Slack on perf regressions in main

popescuoctavian:

---

This PR ports the compare/record split from EDR's `hh3-regression-benchmark.yml` into the regression benchmark workflow, so Slack is also notified when a regression is detected on a `main` baseline run.

- Splits the single "Store benchmark result" step into a read-only compare step (`fail-on-alert: true`, absorbed on baseline runs via `continue-on-error` so main stays green) and a record step that always pushes the new baseline (`auto-push: true`, `fail-on-alert: false`).
- Adds a cleanup step between the two, since the compare step's results-repo clone would make the record step's `git clone` fail.
- Upgrades the notify step to also fire on the absorbed regression, appending " — perf regression on main (baseline updated)" to the Slack message, and sets `errors: true` so a rejected delivery fails the step.

Previously a >110% regression on a main push was silently swallowed by `fail-on-alert: false`; now it's recorded as the new baseline **and** reported to Slack. `/bench` and dispatch runs are unaffected.

### DIFF
--- a/.github/workflows/regression-benchmark.yml
+++ b/.github/workflows/regression-benchmark.yml
@@ -94,8 +94,10 @@ jobs:
     runs-on: hardhat-linux-amd64-self-hosted
     timeout-minutes: 180
     permissions:
-      # Lets the short-lived GITHUB_TOKEN used on non-baseline runs read commit
-      # metadata via the REST API (see "Store benchmark result").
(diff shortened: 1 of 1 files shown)


Wodann (approved):
Thanks for backporting this

PR #8590, opened 2026-08-31
feat(solidity-test): add support for Solidity test profiles

schaable:
Adds support for solidity test profiles. 

Design Doc: https://app.notion.com/p/nomicfoundation/Solidity-test-profiles-3bc578cdeaf58000ba6bffbdc1ca735e

### Notes

- Requires https://github.com/NomicFoundation/edr/pull/1715 to be merged, an EDR release and bump. Until that happens, CI will fail.
- Documentation PR: https://github.com/NomicFoundation/hardhat-website/pull/297

### DIFF
--- a/.changeset/solidity-test-profiles.md
+++ b/.changeset/solidity-test-profiles.md
@@ -1,0 +1,9 @@
+---
+# docs: https://github.com/NomicFoundation/hardhat-website/pull/297
+"@nomicfoundation/hardhat-errors": patch
+"hardhat": minor
+---
+
+Added support for Solidity Test Profiles. You can now declare several sets of Solidity test settings under `test.solidity.profiles` and choose between them with the `--test-profile` argument or the `HARDHAT_TEST_PROFILE` environment variable. Each profile is configured independently, and a `default` profile is required.
+
+Inline test configuration directives can now be scoped to a profile, so `/// hardhat-config: ci.fuzz.runs = 10000` only applies when the `ci` profile is selected, while an unprefixed directive applies under every profile.
--- a/packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
@@ -26,7 +26,11 @@ import {
 } from "@nomicfoundation/hardhat-zod-utils";
 import { z } from "zod";
 
-import { DEFAULT_TEST_PROFILE } from "./test-profiles.js";
+import {
+  DEFAULT_TEST_PROFILE,
+  RESERVED_TEST_PROFILE_NAMES,
+  TEST_PROFILE_NAME_PATTERN,
+} from "./test-profiles.js";
 
 // the keccak256 of "built for ethereum"
 export const DEFAULT_FUZZ_SEED =
@@ -136,12 +140,27 @@ const solidityTestProfilesUserConfigType = z.object({
       (profiles) => DEFAULT_TEST_PROFILE in profiles,
       "A `default` profile is required when using `profiles`",
     )
-    .refine(
-      (profiles) =>
-        !(DEFAULT_TEST_PROFILE in profiles) ||
-        Object.keys(profiles).every((name) => name === DEFAULT_TEST_PROFILE),
-      "Only the `default` profile is supported. Other profile names will be supported in a future release.",
-    ),
+    .superRefine((profiles, ctx) => {
+      for (const profileName of Object.keys(profiles)) {
+        if (!TEST_PROFILE_NAME_PATTERN.test(profileName)) {
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            path: [profileName],
+            message:
+              "Invalid profile name. Profile names can only contain letters, numbers, underscores and dashes",
+          });
+          continue;
+        }
+
+        if (RESERVED_TEST_PROFILE_NAMES.includes(profileName)) {
+          ctx.addIssue({
+            code: z.ZodIssueCode.custom,
+            path: [profileName],
+            message: `\`${profileName}\` can't be used as a profile name, as it's an inline test configuration key`,
+          });
+        }
+      }
+    }),
 });
 
 const solidityTestUserConfigType = conditionalUnionType(
--- a/packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
+++ b/packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
@@ -241,42 +241,33 @@ export async function resolveSolidityTestUserConfig(
   testsPath = typeof testsPath === "object" ? testsPath.solidity : testsPath;
   testsPath ??= "test";
 
-  const defaultRpcCachePath = path.join(resolvedConfig.paths.cache, "edr");
+  const rpcCachePath = path.join(resolvedConfig.paths.cache, "edr");
 
   const solidityUserConfig = userConfig.test?.solidity;
-  let profileUserConfig: SolidityTestProfileUserConfig | undefined;
+
+  // The flat config shape is sugar for a single `default` profile.
+  let profilesUserConfig: Record<string, SolidityTestProfileUserConfig>;
   if (solidityUserConfig !== undefined && "profiles" in solidityUserConfig) {
-    profileUserConfig = solidityUserConfig.profiles[DEFAULT_TEST_PROFILE];
-    assertHardhatInvariant(
-      profileUserConfig !== undefined,
-      "default profile must be present when the profiles wrapper user config is supplied",
-    );
+    profilesUserConfig = solidityUserConfig.profiles;
   } else {
-    profileUserConfig = solidityUserConfig;
+    profilesUserConfig = { [DEFAULT_TEST_PROFILE]: solidityUserConfig ?? {} };
   }
 
-  const resolvedForking = resolveSolidityTestForkingConfig(
-    profileUserConfig?.forking,
-    resolveConfigurationVariable,
+  assertHardhatInvariant(
+    DEFAULT_TEST_PROFILE in profilesUserConfig,
+    "default profile must be present when the profiles wrapper user config is supplied",
   );
 
-  const {
-    memoryLimit,
-    fuzz,
-    invariant,
-    eip712Types,
-    ...otherProfileUserConfig
-  } = profileUserConfig ?? {};
-
-  const resolvedDefaultProfile = {
-    rpcCachePath: defaultRpcCachePath,
-    ...otherProfileUserConfig,
-    memoryLimit: memoryLimit !== undefined ? BigInt(memoryLimit) : undefined,
-    fuzz: resolveFuzzConfig(fuzz),
-    invariant: resolveInvariantConfig(invariant),
-    forking: resolvedForking,
-    eip712Types: resolveEip712TypesConfig(eip712Types),
-  };
+  const profiles: Record<string, SolidityTestProfileConfig> = {};
+  for (const [profileName, profileUserConfig] of Object.entries(
+    profilesUserConfig,
+  )) {
+    profiles[profileName] = resolveSolidityTestProfileUserConfig(
+      profileUserConfig,
+      rpcCachePath,
+      resolveConfigurationVariable,
+    );
+  }
 
   return {
     ...resolvedConfig,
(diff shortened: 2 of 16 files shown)


packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -265,6 +297,8 @@ const runSolidityTests: NewTaskActionFunction<TestActionArguments> = async (
         hre.globalOptions.gasStatsJson !== undefined,
       eip712CanonicalTypes,
       testSourcePaths,
+      testProfile: testProfileName,
+      declaredTestProfiles: Object.keys(testProfiles),
```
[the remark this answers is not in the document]
schaable: No need. Profiles are ordered in EDR.

packages/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -126,6 +147,12 @@ describe("solidity-test/task-action", function () {
     hre = await createHardhatRuntimeEnvironment(hardhatConfigAllTests);
   });
 
+  after(function () {
+    if (ambientTestProfile !== undefined) {
+      process.env.HARDHAT_TEST_PROFILE = ambientTestProfile;
+    }
+  });
```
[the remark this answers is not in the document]
schaable: No need to address this now, I'll create a separate PR with a test util for handling env vars which will cover the scenario.

packages/hardhat/test/internal/builtin-plugins/solidity-test/inline-config.ts
```diff
@@ -19,9 +19,37 @@ const hardhatConfigInvalidTests = {
...
+    // the profiles they need, so an ambient value would fail them.
+    ambientTestProfile = process.env.HARDHAT_TEST_PROFILE;
+    delete process.env.HARDHAT_TEST_PROFILE;
+  });
+
+  after(() => {
+    if (ambientTestProfile !== undefined) {
+      process.env.HARDHAT_TEST_PROFILE = ambientTestProfile;
+    }
+  });
```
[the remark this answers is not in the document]
schaable: No need to address this now, I'll create a separate PR with a test util for handling env vars which will cover the scenario.

packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -222,42 +241,33 @@ export async function resolveSolidityTestUserConfig(
...
-    fuzz: resolveFuzzConfig(fuzz),
-    invariant: resolveInvariantConfig(invariant),
-    forking: resolvedForking,
-    eip712Types: resolveEip712TypesConfig(eip712Types),
-  };
+  const profiles: Record<string, SolidityTestProfileConfig> = {};
+  for (const [profileName, profileUserConfig] of Object.entries(
+    profilesUserConfig,
+  )) {
+    profiles[profileName] = resolveSolidityTestProfileUserConfig(
```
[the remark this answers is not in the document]
schaable: Technically correct, but users would have to name a profile `__proto__`, which isn’t realistic and unlikely to happen.

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -51,19 +52,51 @@ interface TestActionArguments {
...
 
+  // Only global options get an environment variable fallback for free, so we
+  // apply the same naming convention by hand for this task option.
+  const testProfileName =
+    testProfile ??
+    process.env[getEnvVariableNameFromGlobalOption("testProfile")] ??
+    DEFAULT_TEST_PROFILE;
+
+  const testProfiles = hre.config.test.solidity.profiles;
+  const selectedTestProfile = testProfiles[testProfileName];
```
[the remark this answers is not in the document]
schaable: Same reasoning as https://github.com/NomicFoundation/hardhat/pull/8590#discussion_r4049905645. Users would have to call a profile "toString" or "constructor". Not worth the added code.

packages/hardhat/src/internal/builtin-plugins/solidity-test/test-profiles.ts
```diff
@@ -1 +1,25 @@
...
+  "invariant",
+  "isolate",
+];
+
+/**
+ * Profile names are used as prefixes in inline config directives, where the
+ * grammar is dot-separated, and on the command line, so we keep them to a
+ * conservative character set.
+ */
+export const TEST_PROFILE_NAME_PATTERN: RegExp = /^[A-Za-z0-9_-]+$/;
```
kanej: I would prefer if we ban `-` and `_` as the first character.

.changeset/solidity-test-profiles.md
```diff
@@ -0,0 +1,9 @@
+---
+# docs: https://github.com/NomicFoundation/hardhat-website/pull/297
+"@nomicfoundation/hardhat-errors": patch
+"hardhat": minor
+---
+
+Added support for Solidity Test Profiles. You can now declare several sets of Solidity test settings under `test.solidity.profiles` and choose between them with the `--test-profile` argument or the `HARDHAT_TEST_PROFILE` environment variable. Each profile is configured independently, and a `default` profile is required.
+
+Inline test configuration directives can now be scoped to a profile, so `/// hardhat-config: ci.fuzz.runs = 10000` only applies when the `ci` profile is selected, while an unprefixed directive applies under every profile.
```
kanej: Can we shorten this to a single line. We can add the doc ref on the release notes itself.

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -51,19 +52,51 @@ interface TestActionArguments {
...
   hre,
 ): Promise<Result<SolidityTestRunResult, SolidityTestRunResult>> => {
   // Set an environment variable that plugins can use to detect when a process is running tests
   process.env.HH_TEST = "true";
 
+  // Only global options get an environment variable fallback for free, so we
+  // apply the same naming convention by hand for this task option.
+  const testProfileName =
+    testProfile ??
+    process.env[getEnvVariableNameFromGlobalOption("testProfile")] ??
```
kanej: The argument here is we want to keep to the keep to the same convention as global options, as it is a pseudo-global option? As opposed to hard coding the name?

schaable: correct. Edit to add more context: we could use a global option, but then it would show up in any --help command. Using the task option is more restrictive, but we lose the automatic env variable fallback. Using `getEnvVariableNameFromGlobalOption` enforces the same naming convention.

packages/hardhat/src/internal/builtin-plugins/solidity-test/index.ts
```diff
@@ -51,6 +57,21 @@ const hardhatPlugin: HardhatPlugin = definePlugin({
...
+        name: "testProfile",
+        description: "The Solidity test profile to use (Solidity tests only)",
+        type: ArgumentType.STRING_WITHOUT_DEFAULT,
+        defaultValue: undefined,
+      })
+      .setAction(async () => ({
+        default: async (args, _hre, runSuper) => {
+          // We don't need to do anything here, as the test task will forward
+          // the arguments to its subtasks.
+          return await runSuper(args);
```
kanej: We have an option to validate the passed test profile here - failing early rather than compiling.
I am not committed to this change. It is a slightly better UX, but more code. I am happy to leave it to you to decide.

packages/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts
```diff
@@ -2,18 +2,18 @@ import type { HardhatRuntimeEnvironment } from "../../../../src/types/hre.js";
...
 import {
   assertRejectsWithHardhatError,
-  createTestEnvManager,
   useFixtureProject,
 } from "@nomicfoundation/hardhat-test-utils";
 
 import { overrideTask } from "../../../../src/config.js";
 import { createHardhatRuntimeEnvironment } from "../../../../src/internal/hre-initialization.js";
 import hardhatConfig from "../../../fixture-projects/solidity-test/hardhat.config.js";
+import { createTestEnvManager } from "../../../utils.js";
```
kanej: Isn't this a util now?

packages/hardhat/src/internal/builtin-plugins/solidity-test/task-action.ts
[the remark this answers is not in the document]
kanej: I often call my profile's toString, also "LITTLE_JOHNNY; DROP TABLE;"

packages/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts
schaable: oops 🤦

packages/hardhat/src/internal/builtin-plugins/solidity-test/index.ts
schaable: Before making the change, I want to double-check one thing. If we validate here, there’s a scenario where behavior would change: the main test task skips a subtask when explicit test files are passed and none belong to it. So `hardhat test test/Foo.ts --test-profile production` never validates the profile today. With the override, it would throw. Same with an ambient profile.

I think the trade-off is worth it, and this seems like the right time to introduce the change since it wouldn’t break any existing code. Wdyt?

packages/hardhat/src/internal/builtin-plugins/solidity-test/test-profiles.ts
schaable: https://github.com/NomicFoundation/hardhat/pull/8590/commits/8b41315d81817362e484e44bc8297fd1dd3773a1

.changeset/solidity-test-profiles.md
schaable: https://github.com/NomicFoundation/hardhat/pull/8590/commits/092c4cb0c75e8243551c7d24b3a450e651bdd500

packages/hardhat/test/internal/builtin-plugins/solidity-test/task-action.ts
schaable: https://github.com/NomicFoundation/hardhat/pull/8590/commits/0fb6c989fc21fa40852bdd1f928ea264f0a0c572

packages/hardhat/src/internal/builtin-plugins/solidity-test/formatters.ts
```diff
@@ -55,8 +55,8 @@ function formatInlineConfigDirectiveProblem(
   switch (problem.kind) {
     case "InlineConfigInvalidSyntax":
       return `missing "=" in "${problem.directive}"`;
-    case "InlineConfigUnsupportedProfile":
-      return `unsupported profile "${problem.profile}". Only the "default" profile is supported`;
+    case "InlineConfigUndeclaredProfile":
+      return `unknown profile "${problem.profile}". Declared profiles: ${problem.declaredProfiles.map((name) => `"${name}"`).join(", ")}`;
```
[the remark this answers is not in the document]
schaable: Still pending. EDR side is not yet released.

packages/hardhat/src/internal/builtin-plugins/solidity-test/helpers.ts
```diff
@@ -162,6 +166,8 @@ export async function solidityTestConfigToSolidityTestRunnerConfigArgs({
       : CollectStackTraces.OnFailure,
     eip712CanonicalTypes,
     testSourcePaths,
+    testProfile,
+    declaredTestProfiles,
```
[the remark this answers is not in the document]
schaable: Still pending. EDR side is not yet released.

packages/hardhat/src/internal/builtin-plugins/solidity-test/index.ts
kanej: If you think the extra code is not going to be a significant complexity addition, lets add it now.

kanej: I am happy with the tradeoff, and prefer to take early fail on on a malformed test profile.

packages/hardhat/src/internal/builtin-plugins/solidity-test/config.ts
```diff
@@ -136,12 +140,27 @@ const solidityTestProfilesUserConfigType = z.object({
       (profiles) => DEFAULT_TEST_PROFILE in profiles,
       "A `default` profile is required when using `profiles`",
```
kanej:
```suggestion
      "The 'default' profile is required when using Solidity test profiles",
```
This is to bring it inline with equivalent error for build profiles.

packages/hardhat/src/internal/builtin-plugins/solidity-test/index.ts
schaable: https://github.com/NomicFoundation/hardhat/pull/8590/commits/d566c0e6b735f8ea3e645c39cb2f8029caee1ed7

packages/hardhat/src/internal/builtin-plugins/solidity-test/test-task-action.ts
```diff
@@ -0,0 +1,27 @@
...
+ * The `test` task forwards its arguments to the subtasks that declare them, so
+ * this override only validates the selected profile. It does that here, rather
+ * than leaving it to the Solidity subtask, because the umbrella task builds the
+ * project before running any subtask: an unknown profile would otherwise be
+ * reported after a full compilation.
+ */
+const validateTestProfile: TaskOverrideActionFunction<
+  TestActionArguments
+> = async (args, hre, runSuper) => {
+  getTestProfile(hre.config, resolveTestProfileName(args.testProfile));
```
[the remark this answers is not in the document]
schaable: This was an accepted trade-off.

kanej (approved):

